PR #13, opened 2017-10-06
Migrate repo URLs

yangshun:
Migrate some old repo URLs to the new repo URLs.

### DIFF
--- a/content/docs/codebase-overview.md
+++ b/content/docs/codebase-overview.md
@@ -66,7 +66,7 @@ The [fbjs repository](https://github.com/facebook/fbjs) exists because React sha
 After cloning the [React repository](https://github.com/facebook/react), you will see a few top-level folders in it:
 
 * [`src`](https://github.com/facebook/react/tree/master/src) is the source code of React. **If your change is related to the code, `src` is where you'll spend most of your time.**
-* [`docs`](https://github.com/facebook/react/tree/master/docs) is the React documentation website. When you change APIs, make sure to update the relevant Markdown files.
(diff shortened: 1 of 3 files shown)


content/docs/codebase-overview.md
```diff
@@ -66,7 +66,7 @@ The [fbjs repository](https://github.com/facebook/fbjs) exists because React sha
 After cloning the [React repository](https://github.com/facebook/react), you will see a few top-level folders in it:
 
 * [`src`](https://github.com/facebook/react/tree/master/src) is the source code of React. **If your change is related to the code, `src` is where you'll spend most of your time.**
-* [`docs`](https://github.com/facebook/react/tree/master/docs) is the React documentation website. When you change APIs, make sure to update the relevant Markdown files.
+* [`docs`](https://github.com/reactjs/reactjs.org/tree/master/content) is the React documentation website. When you change APIs, make sure to update the relevant Markdown files.
```
yangshun: Strictly speaking this line should be removed because it the docs is no longer part of the facebook/react repo.

bvaughn: I think it's actually pretty important to keep this link here. 😄

gaearon: We should make a pass after repo re-organization and rewrite this.

bvaughn: Definitely.

yangshun: Sorry I meant that it should be removed from this section and rephrased to suit the new repo structure. Definitely agree we should keep the link in this page!

PR #35, opened 2017-10-06
fix #16 missing documentation: null/undefined as input value

### LINKED ISSUE #16: Missing documentation: null/undefined as input valu
Originally filed by @DanielSWolf via facebook/react/issues/11031

**Do you want to request a *feature* or report a *bug*?**

I'd like to request documentation. :-)

When you pass `null` as `value` of an input, you get the following warning:

> warning.js:33 Warning: `value` prop on `select` should not be null. Consider using the empty string to clear the component or `undefined` for uncontrolled components.

This warning very clearly states what to do. My problem is, it fails to mention *why* I'm not supposed to pass `null`. Nor could I find any explanation online:

* I found a link to <http://facebook.github.io/react/tips/controlled-input-null-value.html>. The URL (...controlled-input-null-value...) sounds as if this page should explain exactly what I'm looking for. But it merely redirects to <https://reactjs.org/docs/forms.html>, which doesn't mention `null`.
* The issue ["Warn when input's value is null"](https://github.com/facebook/react/issues/5013) states that this warning should be added, but doesn't say why.
* The (open) pull request ["Document that `value={undefined}` is not controlled component"](https://github.com/facebook/react/pull/9849/files) adds the following documentation: "Similarly to not setting the `value` attribute at all, setting it to `null` or `undefined` will make React assume the component is not controlled and may produce a warning." But again, there is no explanation as to *why* React behaves this way.

Naively, I'd expect `null` to be a perfectly valid value, e.g. for a numeric input that is empty. Passing an empty string to indicate the absence of a numeric value seems odd to me. So I'd really like some more information in the documentation!

FannyVieira:


bvaughn:
Thank you for your pull request and welcome to our community. We require contributors to sign our Contributor License Agreement, and we don't seem to have you on file. In order for us to review and merge your code, please sign up at https://code.facebook.com/cla. **If you are contributing on behalf of someone else (eg your employer)**, the individual CLA may not be sufficient and your employer may need the corporate CLA signed.

If you have received this in error or have any questions, please contact us at [cla@fb.com](mailto:cla@fb.com?subject=CLA%20for%20facebook%2Freact%20%2311138). Thanks!

FannyVieira:
sorry for the inconvenience, I already registered

bvaughn:
No worries! Thanks for confirming. 😄

(Sorry for the awkwardness here. I'm still getting the new repo setup. Normally this comment is automated.)

content/docs/controlled-input-null-value.md
```diff
@@ -0,0 +1,19 @@
+---
+id: controlled-input-null-value
+title: Controlled input null value
+permalink: docs/controlled-input-null-value.html
+---
+
+Specifying the value prop on a [controlled component](/docs/forms.html) prevents the user from changing the input unless you desire so.
```
bvaughn: This should probably link to `/docs/forms.html#controlled-components`

content/docs/controlled-input-null-value.md
```diff
@@ -0,0 +1,19 @@
...
+
+You might have run into a problem where value is specified, but the input can still be changed without consent. In this case, you might have accidentally set value to undefined or null.
+
+For example, this code shows this phenomenon; after a second, the text becomes editable.
```
bvaughn: Grammatical nit: There's a lot of unnecessary commas in this content. Maybe we could change it something like?

> You might have encountered a situation where `value` is specified but an input is still editable. In this case you might have accidentally set `value` to `undefined` or `null`.

> The following code demonstrates this. (After a short delay, the input becomes editable.)

FannyVieira:
Thanks, I'll make the necessary changes.  :+1:

FannyVieira:
I noticed that this redirect rule 
`/tips/controlled-input-null-value.html  /docs/forms.html#anchortaghere` 
can not redirect to a specific topic on the page

bvaughn:
> I noticed that this redirect rule
/tips/controlled-input-null-value.html /docs/forms.html#anchortaghere
can not redirect to a specific topic on the page

It should, so long as the anchor tag actually matches one on the page 😄

PR #36, opened 2017-10-06
Make html-to-jsx redirect permanent

bvaughn:
Will verify this works once Netlify has built a preview.


bvaughn:
Verified https://deploy-preview-36--reactjs.netlify.com/html-jsx.html -> http://magic.reactjs.net/htmltojsx.htm

PR #41, opened 2017-10-07
Update implementation-notes.md

hardfist:
prevChild could be CompositeComponent which cause prevChild.node to be undefined.


content/docs/implementation-notes.md
```diff
@@ -771,7 +771,7 @@ We collect DOM operations on children in a list so we can execute them in batch:
     // Finally, unmount any children that don't exist:
     for (var j = nextChildren.length; j < prevChildren.length; j++) {
      var prevChild = prevRenderedChildren[j];
```
yangshun: The indentation is off within this `for` block (only 1 space), would you mind also fixing it in this PR? 😄

hardfist: fixed

PR #43, opened 2017-10-07
Add FAQ pages

### LINKED ISSUE #29: Add an official React FAQ to the docs
> Note: I originally reported this in https://github.com/facebook/react/issues/10740 .  Please see that thread for prior discussion.

I wrote the [Redux FAQ](http://redux.js.org/docs/FAQ.html) last year, and it's been pretty helpful for answering actual frequently asked questions.  It'd be great if we could add a similar FAQ directly to the React docs.

Tim Arney already has a standalone "React FAQ" site at https://reactfaq.site/ that would be a great starting point for a lot of the material.  My [React/Redux links list](https://github.com/markerikson/react-redux-links) is also a useful source of info in general.

I'm opening this up as a place to discuss possible formatting for a React FAQ, as well as possible topics and volunteers to help write it.

Some thoughts, based on things I've seen and suggestions from the original thread:

A lot of the FAQs I see aren't necessarily about React exactly, but they come up enough and are related enough to be worth trying to answer.

#### My initial suggestions:

- A section on AJAX calls
  - How can I make a call? (Point to common libs)
  - Where do I make a call, constructor/cWM/cDM ?
- A section on method binding necessity and approaches
- Function passing and references
  - The difference between passing a reference to a function and calling it
  - How to pass a parameter to a function as a handler ( `onClick={() => someFunc("someParam")}` ) , which as an FAQ is "why is my function being called whenever I re-render?" because they're doing `onClick={someFunc("someParam")}`

#### Suggestions from @satiewaltz 

Due to React's highly unopinionated nature: It'd be great to use it as a small reference to the different styles and approaches people have building React apps. From the component level all the way up to philosophies on architecture.

As well as help people that are new to React accomplish everyday tasks/get over pitfalls be imparted with the full understanding of what is going on behind the scenes.

Here are some ideas on questions that could be apart of a FAQ:

- A section on styling and CSS:
  - What is the difference between inline styles and CSS?
     - What are ways to do JavaScript animation in React?
     - Can I build animations using React's setState()?
  - A section on common performance issues regarding state updates.
  - A section on state and setState():
      - What setState() does exactly?
      - How do I do ordered state updates?
      - What is the difference between passing an object or a function in setState()?
  - A section on architecture:  
    (It doesn't need to be comprehensive. Acts as an outline of recent developments the community has found when building with React over the years but without giving a side to a certain style):
      - Is there a recommended way to organize a React application's file structure? If not, what are some commonly used strategies?
      - How can I make the transition to React from other component libraries?
  - A section on React's internals:
      - What is the Virtual DOM? What Is the difference between it and the Shadow DOM?
      - What is React's new reconciliation algorithm in React 16?

---

*In foresight to the answers to each FAQ in the future: It'd be best to be cautious of not pointing to one method as being particularly superior to another. Unless it's an anti-pattern of course. Due to the many numbers of ways one can build with React. And the differentiating costs and benefits each way can have. ("Common Anti-patterns" sounds like it could be a section of its own perhaps?) Though I could be entirely wrong about this thought of mine.*

alexkrolick:
Adds an FAQ page addressing some of the ideas from #29

Comments welcome!

---

- Port of https://github.com/facebook/react/pull/11001#event-1282173404 to new repo
- Closes #29 
- Covers request for AJAX guide in #83 
- Covers #75 although the main guide should probably also have information on event handling

### DIFF
--- a/content/docs/faq-ajax.md
+++ b/content/docs/faq-ajax.md
@@ -1,0 +1,15 @@
+---
+id: faq-ajax
+title: AJAX and APIs
+permalink: docs/faq-ajax.html
+layout: docs
+category: FAQ
+---
+
+### How can I make an AJAX call?
+
+You can use an AJAX library you like with React. Some popular ones are [Axios](https://github.com/axios/axios), [jQuery AJAX](https://api.jquery.com/jQuery.ajax/), and the browser built-in [window.fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API).
+
+### Where in the component lifecycle should I make an AJAX call?
+
+You should populate data with AJAX calls in the [`componentDidMount`](https://reactjs.org/docs/react-component.html#mounting) lifecycle method. This is so you can use `setState` to update your component when the data is retrieved.
--- a/content/docs/faq-ajax.md
+++ b/content/docs/faq-ajax.md
@@ -58,3 +58,7 @@ class MyComponent extends React.Component {
   }
 }
 ```
+
+### Cancellation
+
+Note that if the component unmounts before an AJAX call is complete, you may see a warning like `cannot read property 'setState' of undefined`. If this is an issue you may want to keep track of inflight AJAX requests and cancel them in the `componentWillUnmount` lifecycle method.
--- a/content/docs/faq-ajax.md
+++ b/content/docs/faq-ajax.md
@@ -13,3 +13,48 @@ You can use an AJAX library you like with React. Some popular ones are [Axios](h
 ### Where in the component lifecycle should I make an AJAX call?
 
 You should populate data with AJAX calls in the [`componentDidMount`](https://reactjs.org/docs/react-component.html#mounting) lifecycle method. This is so you can use `setState` to update your component when the data is retrieved.
+
+### Example: Using AJAX results to set local state
+
+The component below demonstrates how to make an AJAX call in `componentDidMount` to populate local component state. 
+
+The example API returns a JSON object like this:
+
+```
+{
+  items: [
+    { id: 1, name: 'Apples', price: '$2' },
+    { id: 2, name: 'Peaches', price: '$5' }
+  ] 
+}
+```
+
+```jsx
+class MyComponent extends React.Component {
+  constructor(props) {
+    super(props)
+    this.state = {
+      isLoaded: false,
+      error: null,
+      items: [],
+    }
+  }
+  
+  componentDidMount() {
+    fetch('https://api.example.com/items')
+      .then(res => res.json())
+      .then(result => this.setState({items: result.items))
+      .catch(err => this.setState({ err: error }))
+  }
+  
+  render() {
+    const {error, items} = this.state
+    if (err) return <div>{error}</div>
+    return (
+      <ul>
+        { items.map(item => <li>{item.name} {item.price}</li> }
+      </ul>
+    )
+  }
+}
+```
--- a/content/docs/faq-ajax.md
+++ b/content/docs/faq-ajax.md
@@ -8,7 +8,7 @@ category: FAQ
 
 ### How can I make an AJAX call?
 
-You can use an AJAX library you like with React. Some popular ones are [Axios](https://github.com/axios/axios), [jQuery AJAX](https://api.jquery.com/jQuery.ajax/), and the browser built-in [window.fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API).
+You can use any AJAX library you like with React. Some popular ones are [Axios](https://github.com/axios/axios), [jQuery AJAX](https://api.jquery.com/jQuery.ajax/), and the browser built-in [window.fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API).
 
 ### Where in the component lifecycle should I make an AJAX call?
 
--- a/content/docs/faq-ajax.md
+++ b/content/docs/faq-ajax.md
@@ -31,30 +31,46 @@ The example API returns a JSON object like this:
 
 ```jsx
 class MyComponent extends React.Component {
-  constructor(props) {
-    super(props)
-    this.state = {
-      isLoaded: false,
-      error: null,
-      items: [],
-    }
-  }
-  
+  state = {
+    error: null,
+    isLoaded: false,
+    items: []
+  };
+
   componentDidMount() {
-    fetch('https://api.example.com/items')
+    fetch("https://api.example.com/items")
       .then(res => res.json())
-      .then(result => this.setState({items: result.items))
-      .catch(err => this.setState({ err: error }))
+      .then(result =>
+        this.setState({
+          isLoaded: true,
+          items: result.items
+        })
+      )
+      .catch(error =>
+        this.setState({
+          isLoaded: true,
+          error
+        })
+      );
   }
-  
+
   render() {
-    const {error, items} = this.state
-    if (err) return <div>{error}</div>
-    return (
-      <ul>
-        { items.map(item => <li>{item.name} {item.price}</li> }
-      </ul>
-    )
+    const { error, items } = this.state;
+    if (error) {
+      return <div>Error: {error.message}</div>;
+    } else if (!isLoaded) {
+      return <div>Loading ...</div>;
+    } else {
+      return (
+        <ul>
+          {items.map(item => (
+            <li key={item.name}>
+              {item.name} {item.price}
+            </li>
+          ))}
+        </ul>
+      );
+    }
   }
 }
 ```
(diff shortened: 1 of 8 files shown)


bvaughn:
cc @markerikson

content/docs/faq-ajax.md
```diff
@@ -0,0 +1,64 @@
...
+id: faq-ajax
+title: AJAX and APIs
+permalink: docs/faq-ajax.html
+layout: docs
+category: FAQ
+---
+
+### How can I make an AJAX call?
+
+You can use an AJAX library you like with React. Some popular ones are [Axios](https://github.com/axios/axios), [jQuery AJAX](https://api.jquery.com/jQuery.ajax/), and the browser built-in [window.fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API).
```
JasonEtco: There's a typo in the first few words here, should read:

> You can use any

content/docs/faq-build.md
```diff
@@ -0,0 +1,21 @@
...
+No! Check out https://reactjs.org/docs/react-without-jsx.html
+
+### Do I need to use ES6 (+) with React?
+
+No! Check out https://reactjs.org/docs/react-without-es6.html
+
+### How can I write comments in JSX?
+
+```
+ {*/ Comment goes here */}
```
JasonEtco: I believe that the */ is mixed up at the beginning here, it should be:

```
{/* Comment goes here */}
```

content/docs/faq-functions.md
```diff
@@ -0,0 +1,111 @@
...
+
+If you do have performance issues, by all means, optimize!
+
+### Why is my function being called every time the component re-renders?
+
+Make sure you aren't _calling the function_ when you pass it to the component:
+
+```jsx
+render() {
+  {*/ handleClick is called instead of passed as a reference! */}
```
JasonEtco: This should probably be:

```
{/* handleClick is called instead of passed as a reference! */}
```

alexkrolick: Thanks for the review, fixed these typos.

bvaughn:
I _could_ review this but I'd love for @markerikson to take a first pass if he's willing :smile:

kaycebasques:
Some tech writers hold the opinion that FAQs are simply hints of content that need to be better integrated into the core doc experience.

I personally find them useful sometimes, and I understand that it's a common doc pattern, but I suggest also looking into how you can integrate each of these Qs into the core "Docs" and "Tutorials" sections.

Just trying to be helpful and give an outside perspective! I appreciate all of your work and think the docs are great :)

bvaughn:
Friendly ping to @markerikson in case you'd like to take a first pass 😄

markerikson:
I'll try to take a look at it tomorrow.

content/docs/faq-functions.md
```diff
@@ -0,0 +1,171 @@
...
+  
+  handleClick = i => this.setState({ justClicked: i })
+  
+  render () {
+    return (
+      <div>
+        Just clicked: {this.state.justClicked}
+        <ul>
+          { this.state.items.map(i => 
+            <li onClick={() => this.handleClick(i)}>
```
bvaughn: Add missing `key` attribute.

content/docs/faq-functions.md
```diff
@@ -0,0 +1,171 @@
...
+    })
+  }
+  
+  render () {
+    return (
+      <div>
+        Just clicked: {this.state.justClicked}
+        <ul>
+          { this.state.items.map(i => 
+            <li data-i={i} onClick={this.handleClick}>
```
bvaughn: Add missing `key` attribute.

content/docs/faq-state.md
```diff
@@ -0,0 +1,59 @@
...
+category: FAQ
+---
+
+### What does setState do?
+
+`setState()` schedules an update to a component's `state` object. When state changes, the component responds by re-rendering.
+
+### Why is `setState` is giving me the wrong value?
+
+Calls to `setState` are batched, so it is possible to "lose" an update if you call it with the partial object syntax:
```
bvaughn: I'm not sure the wording "_partial object syntax_" is particularly user-friendly. I'm not sure this is much better, but what do you think about this?

> Calls to `setState` may be asynchronous! If you need to reference a value in state when updating it, use an [updater function](/docs/react-component.html#setstate) instead. ([See below](#what-is-the-difference-between-passing-an-object-or-a-function-in-setstate) for an example.)

content/docs/faq-state.md
```diff
@@ -0,0 +1,59 @@
...
+}
+
+handleSomething() {
+  // this.state.count is 1, then we do this:
+  this.incrementCount()
+  this.incrementCount() // count is now 3
+}
+```
+
+https://reactjs.org/docs/react-component.html#setstate
```
bvaughn: Maybe change this to something like:

bvaughn: Maybe change this to something more like:

> [Learn more](/docs/react-component.html#setstate).

content/docs/faq-styling.md
```diff
@@ -0,0 +1,35 @@
...
+
+```jsx
+render() {
+  return <span className="menu navigation-menu">Menu</span>
+}
+```
+
+### Can I use inline styles?
+
+Yes, see https://reactjs.org/docs/dom-elements.html#style
```
bvaughn: For links within the website, I think it's nicer to use a shorter format rather than the full URL. For example:

> Yes! [Read here](/docs/dom-elements.html#style) to learn more.

content/docs/faq-functions.md
alexkrolick: Since I generated the Array in the constructor, what do you suggest for the key? I could change the example to assume it has some other unique attributes I suppose.

content/docs/faq-state.md
alexkrolick: Good point about that phrasing; I'd like to come up with something in between that explains a bit more about why async patches can drop updates. A link to a Codepen might be sufficient.

content/docs/faq-functions.md
bvaughn: That would be fine. I just don't think we should show an example that omits an important, required property like `key`.

markerikson:
Okay, I finally had a chance to look at this.

I don't think I have many nitpick-type comments.  I _do_ feel like a lot of the answers are kind of bare-bones, and could use both more extensive answers and links to relevant additional info.  (Then again, that's also how _I_ view things, and how I tackled the Redux FAQ, so YMMV.)

That said, this clearly doesn't have to be a final unchangeable thing.  I'm inclined to say we get the FAQ in now, and then keep improving things from there.

alexkrolick:
@markerikson I agree, the idea was to get the skeleton out there, and let the community improve it.

alexkrolick:
@bvaughn addressed your comments and rebased from master

bvaughn:
Yeah, I'm down with the idea of iterating on this.

content/docs/faq-styling.md
```diff
@@ -0,0 +1,35 @@
+---
+id: faq-styling
+title: Styling and CSS
+permalink: docs/faq-styling.html
+layout: docs
+category: FAQ
+---
+
+### How do I add CSS classes to components?
```
darrenscerri: "CSS Classes"  are not a thing. I would just say "class attributes".

content/docs/faq-styling.md
```diff
@@ -0,0 +1,35 @@
...
+
+Yes, see the docs on styling [here](/docs/dom-elements.html#style).
+
+### Are inline styles bad?
+
+CSS classes are generally more efficient than inline styles.
+
+### What is CSS-in-JS?
+
+CSS-in-JS refers to a pattern where CSS is written with Javascript, then extracted into a stylesheet.
```
darrenscerri: Not all libraries extract CSS into a stylesheet

content/docs/faq-ajax.md
```diff
@@ -0,0 +1,80 @@
...
+  componentDidMount() {
+    fetch("https://api.example.com/items")
+      .then(res => res.json())
+      .then(result =>
+        this.setState({
+          isLoaded: true,
+          items: result.items
+        })
+      )
+      .catch(error =>
```
gaearon: This is a very problematic pattern. We have about 20 issues about it in different forms in the React repo. The problem is that this swallows exceptions from *rendering* if the component in question throws. We shouldn't be recommending this.

content/docs/faq-ajax.md
```diff
@@ -0,0 +1,80 @@
...
+        </ul>
+      );
+    }
+  }
+}
+```
+
+### Cancellation
+
+Note that if the component unmounts before an AJAX call is complete, you may see a warning like `cannot read property 'setState' of undefined`. If this is an issue you may want to keep track of inflight AJAX requests and cancel them in the `componentWillUnmount` lifecycle method.
```
gaearon: I just noticed this. It's not true—where is this information coming from? If a component unmounts it won't magically lose setState. You can only get this message if you forget to bind the handler.

alexkrolick: What about navigation events?

PR #47, opened 2017-10-07
Add check for unique redirect URLs

### LINKED ISSUE #1: Add CI check for unique redirect URLs
@sophiebits and I recently noticed that [/docs/index.html](https://reactjs.org/docs/index.html) redirected to [/docs/accessibility.html](https://reactjs.org/docs/accessibility.html) rather than [/docs/hello-world.html](https://reactjs.org/docs/hello-world.html). This turned out to be because both [`/docs/hello-word.md`](https://github.com/reactjs/reactjs.org/blob/master/content/docs/hello-world.md) and [`/docs/accessibility.md`](https://github.com/reactjs/reactjs.org/blob/master/content/docs/accessibility.md) declared the redirect rule `redirect_from: "docs/index.html"`. (Presumably this is because one was copied from the other without much thought about the redirect bit.)

We should add a check to [`gatsby-node.js`](https://github.com/reactjs/reactjs.org/blob/master/gatsby-node.js) that throws an error if it detects a duplicate redirect. This will cause the Netlify preview to fail and alert a PR reviewer that there's a problem.

yangshun:
Fixes #1.

It will be faster to do the checks within the redirect registration but I think doing the check in a separate function is cleaner. There aren't too many edges (currently less than 200) so I think performance is not an issue yet.

The build does not fail when errors are thrown because Gatsby catches them I think. I'm not sure how to get around this except using `process.exit()` instead of throwing an error. @bvaughn Any thoughts?

The errors are currently printed like this. Comments welcome!

```sh
⠁ The following paths are being redirected to different permalinks:
  - docs/animation-ja-JP.html → docs/create-fragment.html, docs/animation.html
  - docs/index.html → docs/accessibility.html, docs/hello-world.html
error Plugin default-site-plugin returned an error


  Error: Found the following redirect_froms to be duplicated:
    - docs/animation-ja-JP.html
    - docs/index.html
  
  - gatsby-node.js:95 Object.exports.createPages
    /Users/yangshun/Developer/reactjs.org/gatsby-node.js:95:11
  
  
  - next_tick.js:188 process._tickCallback
    internal/process/next_tick.js:188:7
 ```

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -28,6 +28,9 @@ exports.modifyWebpackConfig = ({config, stage}) => {
 exports.createPages = async ({graphql, boundActionCreators}) => {
   const {createPage, createRedirect} = boundActionCreators;
 
+  // Used to detect and prevent duplicate redirects
+  const redirectToSlugMap = {};
+
   const blogTemplate = resolve('./src/templates/blog.js');
   const communityTemplate = resolve('./src/templates/community.js');
   const docsTemplate = resolve('./src/templates/docs.js');
@@ -114,13 +117,22 @@ exports.createPages = async ({graphql, boundActionCreators}) => {
           redirect = [redirect];
         }
 
-        redirect.forEach(fromPath =>
+        redirect.forEach(fromPath => {
+          if (redirectToSlugMap[fromPath] != null) {
+            console.error(`Duplicate redirect detected from "${fromPath}" to:\n` +
+              `* ${redirectToSlugMap[fromPath]}\n` +
+              `* ${slug}\n`
+            );
+            process.exit(1);
+          }
+
+          redirectToSlugMap[fromPath] = slug;
           createRedirect({
             fromPath: `/${fromPath}`,
             redirectInBrowser: true,
             toPath: `/${slug}`,
-          }),
-        );
+          });
+        });
       }
     }
   });

### CODE REVIEW

File: gatsby-node.js [hunk 1]
```diff
@@ -57,6 +87,16 @@ exports.createPages = async ({graphql, boundActionCreators}) => {
     throw Error(allMarkdown.errors);
   }
 
+  const issues = detectDuplicateRedirectFroms(allMarkdown.data.allMarkdownRemark.edges);
+  if (issues.length) {
+    console.error('The following paths are being redirected to different permalinks:\n' +
+      issues.map(issue => `  - ${issue.fromPath} → ${issue.permalinks.join(', ')}`).join('\n')
+    );
+    throw new Error('Found the following redirect_froms to be duplicated:\n' +
```
yangshun: Might have to change this to `process.exit(1)` if we want the build to halt.

bvaughn: Agreed. Replacing the `throw` with `process.exit(1)` seems like the right thing to do.

gatsby-node.js
```diff
@@ -25,6 +25,36 @@ exports.modifyWebpackConfig = ({config, stage}) => {
...
+      }
+      redirectFromPaths[fromPath].push(slug);
+    });
+  });
+
+  return Object.keys(redirectFromPaths).map(fromPath => ({
+    fromPath,
+    permalinks: redirectFromPaths[fromPath],
+  })).filter(issue => issue.permalinks.length > 1);
+}
```
bvaughn: I think I see your point about the separate function being _clearer_ but it feels a bit "meh" to add another for-each.

What are your thoughts about doing the check inside of `createPages`, right before we call `createRedirect`? We could still use this technique of redirect-to-slug-map and just fail the first time we saw a key that already existed. This would only warn about the _first_ duplicate (where as your approach warns about _all_ duplicates) but once this check is in place, there should never be duplicates to begin with, so presumably if we found one to warn about- it would be the only one.

For example:
```diff
diff --git a/content/docs/accessibility.md b/content/docs/accessibility.md
index 77cc52df0..de08b19e5 100644
--- a/content/docs/accessibility.md
+++ b/content/docs/accessibility.md
@@ -2,6 +2,7 @@
 id: accessibility
 title: Accessibility
 permalink: docs/accessibility.html
+redirect_from: "docs/index.html"
 ---
 
 ## Why Accessibility?
diff --git a/gatsby-node.js b/gatsby-node.js
index ab60095a6..07d93da1b 100644
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -28,6 +28,9 @@ exports.modifyWebpackConfig = ({config, stage}) => {
 exports.createPages = async ({graphql, boundActionCreators}) => {
   const {createPage, createRedirect} = boundActionCreators;
 
+  // Used to detect and prevent duplicate redirects
+  const redirectToSlugMap = {};
+
   const blogTemplate = resolve('./src/templates/blog.js');
   const communityTemplate = resolve('./src/templates/community.js');
   const docsTemplate = resolve('./src/templates/docs.js');
@@ -114,13 +117,23 @@ exports.createPages = async ({graphql, boundActionCreators}) => {
           redirect = [redirect];
         }
 
-        redirect.forEach(fromPath =>
+        redirect.forEach(fromPath => {
+          if (redirectToSlugMap[fromPath] != null) {
+            console.error(`Duplicate redirect detected from "${fromPath}" to:\n` +
+              `* ${redirectToSlugMap[fromPath]}\n` +
+              `* ${slug}\n`
+            );
+            process.exit(1);
+          }
+
+          redirectToSlugMap[fromPath] = slug;
+
           createRedirect({
             fromPath: `/${fromPath}`,
             redirectInBrowser: true,
             toPath: `/${slug}`,
-          }),
-        );
+          });
+        });
       }
     }
   });
```

yangshun: I get what you mean. I was afraid that it would complicate the `createPage` function too much. I'll change it!

yangshun:
@bvaughn Thanks for your feedback and the idea. I literally changed the PR to the code you provided 😂 

Now the errors look like these. Pretty clear IMO 👍 

```
...
⠁ Duplicate redirect detected from "docs/index.html" to:
* docs/accessibility.html
* docs/hello-world.html

```

PR #48, opened 2017-10-07
Tweak sidebar to update navigation highlight on scroll

### LINKED ISSUE #21: Tutorial navigation highlight should update as you scroll
When you click on side-nav items, the page scrolls to that section (with 4px blue span in front of the link text). If you scroll manually to those parts, the side-nav doesn't update to show where you're at.

You can see it on a screenshot, **An Interactive Component** should be active.

Originally reported by @vicodinchik via facebook/react/issues/10984

jxom:
This intends to fix #21.

Basically, as the page scrolls or the hash is changed, it will update the 'active' sidebar nav item accordingly.

### DIFF
--- a/src/templates/components/Sidebar/Section.js
+++ b/src/templates/components/Sidebar/Section.js
@@ -9,88 +9,161 @@
 
 'use strict';
 
-import React from 'react';
+import React, {Component} from 'react';
 import {colors, media} from 'theme';
 import MetaTitle from '../MetaTitle';
 import ChevronSvg from '../ChevronSvg';
 
-// TODO Update isActive link as document scrolls past anchor tags
-// Maybe used 'hashchange' along with 'scroll' to set/update active links
+class Section extends Component {
+  constructor(props, context) {
+    super(props, context);
 
-const Section = ({
-  createLink,
-  isActive,
-  location,
-  onLinkClick,
-  onSectionTitleClick,
-  section,
-}) => (
-  <div>
-    <MetaTitle
-      onClick={onSectionTitleClick}
-      cssProps={{
-        marginTop: 10,
+    this.state = {
+      activeItemId: null,
+      itemTopOffsets: [],
+    };
 
-        [media.greaterThan('small')]: {
-          color: isActive ? colors.text : colors.subtle,
+    this.handleScroll = this.handleScroll.bind(this);
+  }
 
-          ':hover': {
-            color: colors.text,
-          },
-        },
-      }}>
-      {section.title}
-      <ChevronSvg
-        cssProps={{
-          marginLeft: 7,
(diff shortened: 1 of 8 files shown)


src/templates/components/Sidebar/Section.js
```diff
@@ -9,88 +9,161 @@
...
+    this.handleScroll = this.handleScroll.bind(this);
+  }
+
+  componentDidMount() {
+    const {section} = this.props;
+
+    const itemIds = _getItemIds(section.items);
+    this.setState({
+      itemTopOffsets: _getElementTopOffsetsById(itemIds),
+    });
```
bvaughn: Offsets can change when the window is resized. This approach would require recomputing offsets after a "resize" event.

src/templates/components/Sidebar/Section.js
```diff
@@ -9,88 +9,161 @@
...
-        marginBottom: 10,
-
-        [media.greaterThan('small')]: {
-          display: isActive ? 'block' : 'none',
-        },
-      }}>
-      {section.items.map(item => (
-        <li
-          key={item.id}
+class Section extends Component {
```
bvaughn: How do you feel about leaving the `Section` component as the simple, functional component it was before and creating a new component that decorates it and adds this highlighting behavior? I think this might provide a couple of benefits:
* `Section` stays simple and focused on markup.
* Pages like Tutorial could opt into using this scroll-sync decorator while other pages like Docs can skip it and avoid the unnecessary performance cost.

src/utils/createLink.js
```diff
@@ -52,28 +53,36 @@ const createLinkCommunity = ({item, location, section}) => {
...
-  const isActive = isItemActive(location, item);
+const createLinkTutorial = ({
+  item,
+  location,
+  onLinkClick,
+  section,
+  isActive,
+}) => {
+  const active =
+    typeof isActive === 'boolean' ? isActive : isItemActive(location, item);
```
bvaughn: Would be nice if we didn't have to do this in all of the createLink* util methods.

What do you think about just making `isActive` a boolean parameter that's always expected, and:
* In the case of the scroll-sync decorator component, we could do a smarter check and pass along as a prop to `Section`.
* If no such prop is passed then `Section` can just call the default `isItemActive` util itself

jxom:
Thanks for the feedback @bvaughn! 🙏 Will make this changes later today!

jxom:
@bvaughn - what do you reckon?

src/utils/createLink.js
```diff
@@ -18,9 +18,7 @@ import isItemActive from 'utils/isItemActive';
 import slugify from 'utils/slugify';
 import {colors, media} from 'theme';
```
bvaughn: The `isItemActive` import can be removed now that it's no longer being used.

PR #49, opened 2017-10-07
Document deepEquality checks & JSON.stringify in shouldComponentUpdate are a bad idea

### LINKED ISSUE #7: Document that deep equality checks and JSON.stringify() in shouldComponentUpdate() are a bad idea
Inspired by erikras/redux-form/issues/3461

Relocated from facebook/react/issues/11107

carloscuesta:
Closes #7


AlanFoster:
My 0.02:

On a larger legacy application, that wasn't making use of immutable js or redux, it was actually much faster to implement a deep equality check for shouldComponentUpdate than it was to render all children, particularly for low end iPads 🤔

In my experience it seems like the implementation details for shouldComponentUpdate will depend on the nature of your application, your state tree, and that you should measure the performance carefully yourself.

I wonder if it's worth mentioning that a shallow comparison may be a good solution to use in most cases -  if your props/state are shallow, and are not deeply nested objects that may have been directly mutated -
 then provide a link to https://reactjs.org/docs/shallow-compare.html And that a deep comparison may be slower, but necessary for some applications, however you should always measure your particular application

PR #51, opened 2017-10-07
Fixes #44 : Update CodeEditor to display the complete content

### LINKED ISSUE #44: JSX Editor in documentation need more padding bottom
in every LIVE JSX EDITOR which have scroll, last line is hidden.

image

this is not showed `ReactDOM.render(<Timer />, mountNode);`
image

akshatnitd:
The code at the bottom of the CodeEditor in the examples was hidden, due to overflow : 'hidden' to the parent element. Made changes to show complete code. Also, hid the scroll-bar for better UI.

### DIFF
--- a/src/components/CodeEditor/CodeEditor.js
+++ b/src/components/CodeEditor/CodeEditor.js
@@ -131,6 +131,9 @@ class CodeEditor extends Component {
                   '& pre.prism-code[contenteditable]': {
                     maxHeight: '280px !important',
                     outline: 0,
+                    overflow : 'auto',
+                    marginRight: '0 !important',
+                    marginBottom: '0 !important',
                   },
                 }}
                 className="gatsby-highlight">


akshatnitd:
@bvaughn @gaearon Please review :)

src/css/reset.css
```diff
@@ -40,3 +40,11 @@ img {
...
   font-family: source-code-pro, Menlo, Monaco, Consolas, "Courier New", monospace;
 }
+
+pre::-webkit-scrollbar {
+    display: none;
+}
+
+pre::scrollbar {
+    display: none;
+}
```
bvaughn: We shouldn't disable the scrollbar display like this (although I agree with you that browser scrollbars are generally unattractive). Users depend on scrollbars to indicate that there's content below the fold. Hiding the scrollbar can be a confusing user experience. Also, unless users have a trackpad or a mouse-wheel, they would be unable to scroll the content in this case.

akshatnitd: Yes, correct :) Reverting this change and updating the PR.

akshatnitd:
@bvaughn Made the requested changes :)

akshatnitd:
@bvaughn Thanks for merging :) This was my first PR to the org :smile: Many more to come! :fireworks:

bvaughn:
We're excited to hear that!

PR #55, opened 2017-10-07
add shorthand for specifying pixel values in style props to docs

### LINKED ISSUE #27: Documentation about which style numeric values are treated as pixels is missing
This change was originally requested by @Kovensky via facebook/react/issues/10418

This used to exist in the React 15 documentation under the Tips section, but mentions to this support are now missing entirely.

The old documentation page: https://github.com/facebook/react/blob/8cac523beaaacfeae179ca14a1d8a46d82892016/docs/tips/06-style-props-value-px.md

This was removed in commit 455d2d1b48e5cdaeac5d0b4fd92b29b4d52bcaec

xjlim:
Fix #27 

image

### DIFF
--- a/content/docs/reference-dom-elements.md
+++ b/content/docs/reference-dom-elements.md
@@ -99,6 +99,7 @@ function HelloWorldComponent() {
   return <div style={divStyle}>Hello World!</div>;
 }
 ```
+
 Sometimes you _do_ want to keep the CSS properties unitless. Here's a list of properties that won't get the automatic "px" suffix:
 
 - `animationIterationCount`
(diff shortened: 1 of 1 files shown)


content/docs/reference-dom-elements.md
```diff
@@ -88,6 +88,18 @@ function ComponentWithTransition() {
...
+```js
+const divStyle = {
+  height: 10  // rendered as "height:10px"
+};
+
+function HelloWorldComponent() {
+  return <div style={divStyle}>Hello World!</div>;
+}
+```
+
```
bvaughn: I think the [section about which props are unitless](https://github.com/facebook/react/blob/8cac523beaaacfeae179ca14a1d8a46d82892016/docs/tips/06-style-props-value-px.md) is also worth adding just to avoid potential confusion that _all_ numeric style props are stringified like this.

xjlim: Should it be updated to reflect the latest changes in [this file](https://github.com/facebook/react/blob/4131af3e4bf52f3a003537ec95a1655147c81270/src/renderers/dom/shared/CSSProperty.js#L15-L59)(20 of them are not on the list)?

bvaughn: Sure, or maybe just mention the _type_ of props that aren't converted and list a few examples?

xjlim: Alright, I left it as is and include a link to the file.

bvaughn:
Thanks!

PR #60, opened 2017-10-07
Remove marginbottom, fix screen size of media queries for border-radius

thomasyimgit:
This PR fixes the issue here: https://github.com/reactjs/reactjs.org/issues/54

Now it looks like this:
image

* All the border-radius are fixed
* The redundant margin is removed
￼


bvaughn:
Thanks for the PR! PR #58 was submitted before this one though and so I've merged it.

Issue #62, opened 2017-10-08
Thinking In React example loads CodePen script multiple times


danielsbird:
The [Thinking in React](https://github.com/reactjs/reactjs.org/blob/master/content/docs/thinking-in-react.md) doc has three embedded CodePens and each of them load the same script: [one](https://github.com/reactjs/reactjs.org/blame/master/content/docs/thinking-in-react.md#L65), [two](https://github.com/reactjs/reactjs.org/blame/master/content/docs/thinking-in-react.md#L110), [three](https://github.com/reactjs/reactjs.org/blame/master/content/docs/thinking-in-react.md#L136). 

Loading CodePen's script once is enough to make all the pens on that page work.

bvaughn:
Thanks for pointing this out. Looks like a great first-task for someone!

ishankbahl:
Hey @bvaughn, made a pull request for this.

PR #63, opened 2017-10-08
Load CodePen script once

danielsbird:
Fix issue #62 by loading CodePen script only once on the Thinking in React page.


bvaughn:
Thanks! 😄

PR #64, opened 2017-10-08
Update signature of requestAnimationFrame in documentation

hwndept:
# Scope

- Included timestamp into callback for `requestAnimationFrame` to make the function be compatible with [origin implementation](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame)

### DIFF
--- a/content/docs/reference-javascript-environment-requirements.md
+++ b/content/docs/reference-javascript-environment-requirements.md
@@ -27,6 +27,8 @@ React also depends on `requestAnimationFrame` (even in test environments). A sim
 
 ```js
 global.requestAnimationFrame = function(callback) {
-  setTimeout(callback, 0);
+  setTimeout(function() {
+    callback(Date.now());
+  }, 0);
 };
 ```


content/docs/reference-javascript-environment-requirements.md
```diff
@@ -27,6 +27,8 @@ React also depends on `requestAnimationFrame` (even in test environments). A sim
 
 ```js
 global.requestAnimationFrame = function(callback) {
-  setTimeout(callback, 0);
+  setTimeout(function() {
+    callback(Date.now());
```
bvaughn: This isn't quite right. According to the [documentation for `requestAnimationFrame`](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame#Parameters):

> A parameter specifying a function to call when it's time to update your animation for the next repaint. The callback has one single argument, a [`DOMHighResTimeStamp`](https://developer.mozilla.org/en-US/docs/Web/API/DOMHighResTimeStamp), which indicates the current time (the time returned from [`performance.now()`](https://developer.mozilla.org/en-US/docs/Web/API/Performance/now)) for when `requestAnimationFrame` starts to fire callbacks.

`Date.now()` returns the number of milliseconds elapsed since 1 January. `performance.now()` returns the number of milliseconds elapsed since the beginning of the current document's lifetime.

Unfortunately `performance.now` has the same browser support as does `requestAnimationFrame` so if one's not available, chances are that neither will be. A lot of raf polyfills roughly mimic the time returned by `performance.now` using time deltas.

I'm not sure if this level of detail or complexity is really worth mentioning in this part of the docs. I would be more inclined to add a note saying that libs like [`raf`](https://www.npmjs.com/package/raf) are a nice alternative if you want a more accurate polyfill.

hwndept: I agree with you, my implementation doesn't repeat origin one, it's something that follows the callback's interface.  So if somebody's code relies on the argument - exception will not be thrown.

Also, I agree with you, that there is no sense to provide complex polyfill in the example - it's enough just to mention that requestAnimationFrame is required and nice to have to provide a link to some open source implementation, but in the same time current example may cause an error.

bvaughn:
I think this PR is safe to close because it doesn't seem like it's actively being worked on. Happy to re-open and re-review if you have wording suggestions or other improvements though! Thanks!

hwndept:
Sure, not a problem.

Issue #75, opened 2017-10-08
Document how to pass args to event handlers


bvaughn:
This issue was originally reported by @StokeMasterJack via facebook/react/issues/9685

A similar issue was reported by @Frazer via facebook/react/issues/8065

### Text of original issue #9685

Inside a loop it is common to want to pass a param to an event handler:

```
<button onClick={(event) => this.deleteRow(23)}>Delete Row</button>
<button onClick={this.deleteRow.bind(this,23)}>Delete Row</button>
```

That would be a good thing to add to this doc page:

https://facebook.github.io/react/docs/handling-events.html

Also, how to pass args without triggering a re-render in the child component (i.e. creating a new function every time).

sw-yx:
will work on this

bvaughn:
This issue is all yours! 😄

I've added an "in-progress" label so that others will know not to start work on the issue. If you change your mind about the issue, no worries! Just let me know so that I can remove the label and free it up for someone else to claim.

Cheers!

alexkrolick:
This topic was also been suggested for the FAQ in #29, so I included some comments in PR #43 ([diff](https://github.com/reactjs/reactjs.org/pull/43/files?short_path=e475a67#diff-e475a6732d0a9962daab40c49d5f41cb)).

It might be a good idea to merge the docs on this subject, or link from one to the other.

Frazer:
I think @hedgerh suggestion should also be included.

<button name='23' onClick={this.deleteRow}>Delete Row</button>

deleteRow (e) {
  // do some action to (e.target.name)
}

alexkrolick:
@Frazer added to the FAQ version of this topic: https://github.com/reactjs/reactjs.org/pull/43/commits/b2ef5ff1737b339df61d135d39bf0a9e22869775

alexkrolick:
Closed by #81 I think

gfortaine:
@StokeMasterJack @bvaughn There is also a third way through currying (see [How to pass variables to a function reference?](https://stackoverflow.com/a/50631170), thx @esthercuan)  👍 

```javascript
deleteRow = id => event => {
    console.log('row ${id} deleted');
 }

<button onClick={this.deleteRow(23)}>Delete Row</button>
```

Frazer:
Thanks for the interesting example. This creates a copy of the function
every time it is rendered, so I'm guessing wouldn't be good when
performance is an issue.

sw-yx:
[...]

PR #81, opened 2017-10-08
added documentation for passing arguments to event handlers

sw-yx:
hi, addressing issue #75 !

image

### DIFF
--- a/content/docs/handling-events.md
+++ b/content/docs/handling-events.md
@@ -145,5 +145,6 @@ The problem with this syntax is that a different callback is created each time t
 Inside a loop it is common to want to pass a param to an event handler. For example if `i` is the row id:
 
 ```js
+<button onClick={() => this.deleteRow(i)}>Delete Row</button>
 <button onClick={this.deleteRow.bind(this, i)}>Delete Row</button>
 ```
--- a/content/docs/handling-events.md
+++ b/content/docs/handling-events.md
@@ -139,3 +139,17 @@ class LoggingButton extends React.Component {
 ```
 
 The problem with this syntax is that a different callback is created each time the `LoggingButton` renders. In most cases, this is fine. However, if this callback is passed as a prop to lower components, those components might do an extra re-rendering. We generally recommend binding in the constructor or using the class fields syntax, to avoid this sort of performance problem.
+
+## Passing arguments to event handlers
+
+Inside a loop it is common to want to pass a param to an event handler. For example if `i` is the row id:
+
+```js
+<button onClick={() => this.deleteRow(i)}>Delete Row</button> 
+```
+
+or alternatively (especially if you want to avoid triggering a re-render in a child component):
+
+```js
+<button onClick={this.deleteRow.bind(this,i)}>Delete Row</button>
+```
--- a/content/docs/handling-events.md
+++ b/content/docs/handling-events.md
@@ -142,7 +142,7 @@ The problem with this syntax is that a different callback is created each time t
 
 ## Passing arguments to event handlers
 
-Inside a loop it is common to want to pass a param to an event handler. For example if `i` is the row id:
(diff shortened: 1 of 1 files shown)


sw-yx:
sorry bout that. package-lock never used to matter when i was just working on my little repos. removed and recommitted.

content/docs/handling-events.md
```diff
@@ -139,3 +139,17 @@ class LoggingButton extends React.Component {
...
+
+## Passing arguments to event handlers
+
+Inside a loop it is common to want to pass a param to an event handler. For example if `i` is the row id:
+
+```js
+<button onClick={() => this.deleteRow(i)}>Delete Row</button> 
+```
+
+or alternatively (especially if you want to avoid triggering a re-render in a child component):
```
bvaughn: This line isn't actually true ^
```js
function foo() {}
const ctx = {};
const a = foo.bind(ctx);
const b = foo.bind(ctx);
console.log(a == b); // false
```

Binding in the constructor is a way to avoid unnecessary updates but that's not an option when you want to pass in-scope variables like this.

That being said, I think we _should_ still show the `.bind()` form for people who prefer to write vanilla ES5 but it doesn't perform differently in terms of `shouldComponentUpdate` (at least not that I'm aware of).

One way to avoid unnecessary renders would be:
```jsx
class Parent extends Component {
  deleteRow = (index) => {
    // Do stuff...
  };

  render() {
    // Render stuff, including
    <DeleteRowButton index={i} deleteRow={this.deleteRow} />
  }
}

class DeleteRowButton extends Component {
  onClick = () => {
    this.props.deleteRow(this.props.index);
  };

  render() {
    return <button onClick={this.onClick}>Delete Row</button>;
  }
}
```

But all of this ^ is overkill for a simple `<button>`. It would only make sense to do if the component you're rendering is expensive (eg has as lot of nested children or complex logic).

sw-yx: i think the examples conflate the method of passing params to events (where this pattern is valid) with passing params within loops (where this pattern is not valid as you point out). I should be clearer there AND also provide the hint to pass by props as newbies can overlook that simple solution. I agree the rerendering thing is overkill and if the reader is worrying about perf to that extent they probably wouldnt be reading this section of the docs anyway. Was only including it because it was suggested in the issue raised.

alexkrolick: I added some comments about how and when to avoid binds to the FAQ page (#43). It's probably OK to keep it high-level here.

bvaughn: > It's probably OK to keep it high-level here.

Agreed.

sw-yx: Parsimony is a virtue when it comes to docs. I will just delete the problematic line of code so as not to mislead/confuse.

sw-yx:
Pushed and awaiting your review. sorry for the original problematic line (i'm still not super comfortable when it comes to questions of context)

gaearon:
Tweaked in db1262d06f7ec5081886c8257afe0e6c239dd16c.

"Param" is not explained, and people might confuse it with some React term. "Parameter" is more explicit.

Also added reference links.

Finally, these two lines weren't equivalent, and people would likely ask how to get access to the event. So I made them consistent.

Thanks for working on this change!

bvaughn:
Nice changes Dan. Thanks for adding to the section.

content/docs/handling-events.md
devuxer: I understand your thinking, but I actually disagree with the decision you guys ended up making. I came to this thread specifically because I was confused why there is a warning about re-renders for `onClick={(e) => this.handleClick(e)}` but not for `onClick={(e) => this.deleteRow(id, e)}`.

Also, being that I'm a relative newbie to React, I didn't fully understand the ramifications of these potential re-renders. The discussion here actually helped me get a sense of how serious a problem they might be from a performance standpoint.

I understand not wanting to pollute the documentation with too much detail, but it would be nice to put the extra details in something akin to footnotes.

Issue #74, opened 2017-10-08
Improve docs around which lifecycle methods are safe for side effects


bvaughn:
This issue was originally reported by @jedwards1211 via facebook/react/issues/9693

Outcome of discussion on this issue: We should better document that you're not expected to cause side effects that synchronously update state in certain lifecycle methods. [Here's a cheatsheet](https://gist.github.com/bvaughn/923dffb2cd9504ee440791fade8db5f9) that shows which methods are safe and which aren't.

bvaughn:
Marking this in-progress since @jedwards1211 has submitted a PR

bvaughn:
Closed by #88

PR #88, opened 2017-10-08
Caution (in docs) against dispatching redux actions or triggering component updates by any other means in `componentWillUpdate`

jedwards1211:
(moving from https://github.com/facebook/react/pull/9694)


bvaughn:
I believe this corresponds with issue #74 😄

bvaughn:
Thanks!

PR #90, opened 2017-10-09
Fix visible background over border radius in search dropdown

imagentleman:
Currently the algolia search results have a weird and subtle but eye-twitch-inducing artifact, where you can see the white background on the top round borders of the dropdown.

It looks like this:

before

This PR moves the background from the top container to each individual row that must be white in order to achieve the same effect.

It would look like this:

after

### DIFF
--- a/src/css/algolia.css
+++ b/src/css/algolia.css
@@ -263,7 +263,6 @@
 
 .algolia-autocomplete .ds-dropdown-menu [class^="ds-dataset-"] {
   position: relative;
-  background: #fff;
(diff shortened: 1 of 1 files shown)


imagentleman:
@bvaughn , yep i'll be happy to fix any issues. 

Just visited the autogenerated preview of the PR on my phone (https://deploy-preview-90--reactjs.netlify.com/) but at first glance couldn't see the horizontal overflow : S

Can you attach a screenshot or give me a few more details?

PR #93, opened 2017-10-09
Create script for checks (prettier, lint, flow) in CI

### LINKED ISSUE #67: CI should run Lint, Flow, and Prettier checks for every PR
The `package.json` defines a [`check-all`](https://github.com/reactjs/reactjs.org/blob/be1f9a60d80261eb70eded915b266078b606524c/package.json#L71) script that runs Prettier, Lint, and Flow. This script is intended to be run before commit to help catch inconsistent formatting or type errors. However we are not currently running this check as part of CI.

- [x] To fix this, first create a script that does the following (done via #93)
  * Run Prettier and fail if any modifications are made. (This means that the committed code isn't formatted correctly.)
  * Run ESLint and fail if there are any errors.
  * Run Flow and fail if there are any errors.
- [x] Fix all pre-existing errors so the new `yarn ci-check` task passes in `master`. (done via #104)
- [x] Hook this new script into CircleCI to run automatically for every PR. (A React team member will need to help with this step.) (done via #131)

jxom:
This intends to fix #67.

- Added a `check-all:verbose` script, which is similar to `check-all`, however lists errors out on prettier if the files are different from the prettier formatting.
- Added a `prettier:diff` script, which checks for a difference between the prettified files, and source files.

Yay or nay?

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,7 +1,15 @@
 {
+  "extends": [
+    "eslint:recommended",
+    "plugin:react/recommended"
+  ],
   "plugins": [
     "prettier",
     "react"
   ],
   "parser": "babel-eslint",
+  "env": {
+    "node": true,
+    "browser": true
+  }
 }
--- a/package.json
+++ b/package.json
@@ -68,11 +68,12 @@
   },
   "scripts": {
     "build": "gatsby build",
-    "check-all": "yarn prettier && yarn lint && yarn flow",
+    "check-all": "yarn prettier -- --write && yarn lint && yarn flow",
+    "check-all:verbose": "yarn prettier -- --list-different && yarn lint && yarn flow",
     "dev": "gatsby develop -H 0.0.0.0",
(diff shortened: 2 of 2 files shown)


bvaughn:
Some thoughts! 😄 

* I would rename `check-all:verbose` to something like `ci-check` just to make its purpose clearer 😄 
* `lint` isn't failing when there are errors like unused variables- but it should. I've probably just totally misconfigured it. Maybe we should copy over something similar to the React project's lint config? The goal is to get a useful set of lint checks that don't conflict with Prettier's config.

jxom:
@bvaughn - yeah was a bit unsure about what to name it! 

I was wondering why the linter didn’t pick up stuff like that too! Thought there may have been some reason why you configured it that way. 😆

Will check it out later on!

bvaughn:
Nope. I was trying for a minimal linter config and didn't test it thoroughly enough at the time.

Issue #3, opened 2017-10-06
Gracefully recover when ad block prevents Babel from loading


bvaughn:
I've seen reports that certain ad blockers (eg Fair AdBlocker 1.404) prevent Babel from loading which causes a script error and breaks the docs site. We should more robustly handle this case (probably by showing an error on the code-preview side, where we would show script errors).

Note that this issue is related to issue #2 and may be impacted by its implementation.

bvaughn:
Note that this issue was reported by another user but I am unable to reproduce it. Removing Babel entirely (so it isn't loaded) results in the following for me:
![screen shot 2017-10-09 at 9 55 10 am](https://user-images.githubusercontent.com/29597/31349409-048c5aa6-acd8-11e7-9d70-07f0685dfbc2.png)

I think this is good enough. 😄

bvaughn:
Eh, maybe we could improve the error message slightly to mention ad blockers if we detect no Babel. I'll re-open and re-purpose for that.

renanpvaz:
I can reproduce this, perhaps I can send a PR improving the message.

bvaughn:
Great! This issue is all yours! 😄

I've added an "in-progress" label so that others will know not to start work on the issue. If you change your mind about the issue, no worries! Just let me know so that I can remove the label and free it up for someone else to claim.

Cheers!

PR #96, opened 2017-10-09
Added error boundaries to CodeEditor and layout component

ishankbahl:
Fix issue #3 , https://github.com/facebook/react/issues/11008 by adding error boundaries to CodeEditor and layout components. Nudging (using alert()) users to whitelist the site in AdBlocker in the alternative UI.

### DIFF
--- a/src/components/CodeEditor/CodeEditor.js
+++ b/src/components/CodeEditor/CodeEditor.js
@@ -17,6 +17,7 @@ import Remarkable from 'remarkable';
 import {LiveProvider, LiveEditor} from '@gaearon/react-live';
 import {colors, media} from 'theme';
 import MetaTitle from 'templates/components/MetaTitle';
+import ErrorBoundary from '../ErrorBoundary';
 
 const compile = code =>
   Babel.transform(code, {presets: ['es2015', 'react']}).code; // eslint-disable-line no-undef
@@ -285,4 +286,4 @@ class CodeEditor extends Component {
   };
 }
 
-export default CodeEditor;
+export default ErrorBoundary(CodeEditor);
--- a/src/components/ErrorBoundary/ErrorBoundary.js
+++ b/src/components/ErrorBoundary/ErrorBoundary.js
@@ -1,0 +1,39 @@
+/**
+ * Copyright (c) 2013-present, Facebook, Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @emails react-core
+ */
+
+'use strict';
+
+import React, {Component} from 'react';
+
+function ErrorBoundary(WrappedComponent) {
+    return class extends Component{
+        constructor(props, context) {
+            super(props, context);
+
+            this.state = {
+                error: null,
+            };
+        }
+
+        componentDidCatch(error) {
+            this.setState({ error });
+        }
+
+        render() {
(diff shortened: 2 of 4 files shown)


src/components/ErrorBoundary/ErrorBoundary.js
```diff
@@ -0,0 +1,39 @@
...
+        }
+
+        componentDidCatch(error) {
+            this.setState({ error });
+        }
+
+        render() {
+            console.log(this.state);
+            if (this.state.error) {
+                alert('Error, try to whitelist the site in AdBlocker/Cookie Whitelist');
```
bvaughn: I don't think this is the right way to warn people about this. I think a `console.error` log would be reasonable though. Maybe something like:

> If you're using an ad-blocker or have disabled cookies,
> Consider adding this website to the whitelist.

We should also keep in mind that not _all_ errors are caused by this so our wording shouldn't be overly assertive that this is definitely the cause/fix. Users might see an error even if they aren't using adblockers and haven't disabled cookies, and that would be confusing.

src/components/CodeEditor/CodeEditor.js
```diff
@@ -17,6 +17,7 @@ import Remarkable from 'remarkable';
 import {LiveProvider, LiveEditor} from '@gaearon/react-live';
 import {colors, media} from 'theme';
 import MetaTitle from 'templates/components/MetaTitle';
+import ErrorBoundary from '../ErrorBoundary';
```
bvaughn: nit
```js
import ErrorBoundary from 'components/ErrorBoundary';
```

src/layouts/index.js
```diff
@@ -25,6 +25,7 @@ import '../prism-styles';
 import 'glamor/reset';
 import 'css/reset.css';
 import 'css/algolia.css';
+import ErrorBoundary from '../components/ErrorBoundary';
```
bvaughn: nit
```js
import ErrorBoundary from 'components/ErrorBoundary';
```

src/components/ErrorBoundary/ErrorBoundary.js
```diff
@@ -0,0 +1,39 @@
...
+                error: null,
+            };
+        }
+
+        componentDidCatch(error) {
+            this.setState({ error });
+        }
+
+        render() {
+            console.log(this.state);
```
bvaughn: Remove this^ 😄

src/layouts/index.js
```diff
@@ -82,4 +83,4 @@ class Template extends Component {
   }
 }
 
-export default Template;
+export default ErrorBoundary(Template);
```
bvaughn: Errors in the template will actually result in a totally empty/white page. This is probably not great. We should show _something_ as a fallback in the browser (in addition to logging error info to the console).

ishankbahl: What do you suggest as the fallback?
Maybe we can use a modal (using react-modal) here.

bvaughn: We probably need to show a basic error page. It doesn't have to look super pretty. @joecritch or I can pretty it up.

bvaughn:
It's been a while since there was any movement on this PR, and the code it modifies has been changed by subsequent PRs. I'm going to close this for now- but I'm happy to re-review if the PR is updated. Please just ping me by name if there are more changes pushed!

Issue #97, opened 2017-10-09
Remove hard-coded HTML/CSS from index.md


bvaughn:
(This is a tech debt clean-up issue from the legacy Jekyll docs site.)

The [`docs/index.md`](https://github.com/reactjs/reactjs.org/blob/master/content/index.md) file currently contains a lot of HTML- complete with attributes and class names. This file should contain only (or at least predominantly) markdown syntax. We should remove as much of the HTML as possible and instead convert it to regular markdown.

There are 2 things to be aware of while cleaning up this markdown file:
1. The layout of the home page uses horizontal scrolling for the top section. This is not achievable with regular markdown.
1. The bottom sections each have a live code example that ties into the `home.js` template (see below).

The [`templates/home.js`](https://github.com/reactjs/reactjs.org/blob/master/src/templates/home.js) template makes this situation a little more complex, since it currently uses some of the inline HTML elements from `index.md` to mount examples (eg `<div id="helloExample"></div>`). It might be acceptable to leave _only_ these mount nodes in the file and remove all else.

It might also be nice to break index down into a couple of files, although this would also require:
1. Moving `home.js` from [templates](https://github.com/reactjs/reactjs.org/tree/master/src/templates) to [pages](https://github.com/reactjs/reactjs.org/tree/master/src/pages) and creating a graphql query to load these files
2. Updating [`gatsby-node.js`](https://github.com/reactjs/reactjs.org/blob/master/gatsby-node.js) to ignore the files (so they aren't turned into standalone pages. (We have [a similar check](https://github.com/reactjs/reactjs.org/blob/89cf6d20d6dce1db6f3a273795e3f1041d37c55c/gatsby-node.js#L74-L77) for the error decoder page.)

Making Home a page instead of a template would also provide a nice solution for the horizontal scrolling problem mentioned above.

Also, `mountCodeExample` currently does some [really hacky stuff](https://github.com/reactjs/reactjs.org/blob/42372b24473913fef65146b2ec70e0817076e25f/src/utils/mountCodeExample.js#L16-L28). Making Home a page instead of a template could help us move some of that hackiness out and simplify things like showing a placeholder while pre-loading Babel (#2).

Please feel free to chat with @bvaughn if you decide to pick up this issue.

yangshun:
@bvaughn I would like to take up this task

bvaughn:
This issue is all yours! 😄

[...]

PR #100, opened 2017-10-09
Defer babel loading

### LINKED ISSUE #2: Defer loading Babel until it is needed
We currently load Babel via CDN (which is nice) but we load it in the `<head>` of the page (which is not so nice). We should only load it on pages that actually use it (eg pages that use the [`CodeEditor`](https://github.com/reactjs/reactjs.org/blob/master/src/components/CodeEditor/CodeEditor.js)).

Check out the [Babel website's REPL](https://github.com/babel/website/tree/master/js/repl) for an example of similar deferred loading.

renanpvaz:
Created an utility function to asynchronously load Babel and Algolia attempting to fix #2.


src/layouts/index.js
```diff
@@ -28,13 +30,15 @@ import 'css/algolia.css';
...
-      inputSelector: '#algolia-doc-search',
+    loadScript(algoliaURL).then(() => {
+      // Initialize Algolia search.
+      // TODO Is this expensive? Should it be deferred until a user is about to search?
+      // eslint-disable-next-line no-undef
+      docsearch({
+        apiKey: '36221914cce388c46d0420343e0bb32e',
+        indexName: 'react',
+        inputSelector: '#algolia-doc-search',
+      });
```
bvaughn: This will load the Algolia JS for every page. Even though the request will be cached, the browser might still end up wasting cycles processing/reprocessing the script.

Also, ideally, if we're going to defer loading Algolia- we should defer loading it until the user actually starts to search (eg clicks to focus in the search bar). So maybe we could do this async step there instead? Happy to (a) talk more about this or (b) split that part off into a separate, follow-up issue or task.

src/templates/home.js
```diff
@@ -18,13 +18,17 @@ import React, {Component} from 'react';
...
-    mountCodeExample('helloExample', HELLO_COMPONENT);
-    mountCodeExample('timerExample', TIMER_COMPONENT);
-    mountCodeExample('todoExample', TODO_COMPONENT);
-    mountCodeExample('markdownExample', MARKDOWN_COMPONENT);
+    loadScript(babelURL).then(() => {
+      mountCodeExample('helloExample', HELLO_COMPONENT);
+      mountCodeExample('timerExample', TIMER_COMPONENT);
+      mountCodeExample('todoExample', TODO_COMPONENT);
+      mountCodeExample('markdownExample', MARKDOWN_COMPONENT);
+    });
```
bvaughn: What are your thoughts about showing a loading placeholder in the 4 divs in question until Babel has finished loading?

renanpvaz: It's a good idea, I can see that it is flashing once it loads. To fix this should I style the divs at content/index.md?

bvaughn: Probably not. (We actually need to clean up `index.md`, see #97)

I would try maybe temporarily rendering a placeholder into the container from the Home template.

renanpvaz: So you mean that I should use `ReactDOM.render`  like `mountCodeExample` does?

bvaughn: Probably 😄

src/layouts/index.js
renanpvaz: Perhaps this should be a separate issue.

src/layouts/index.js
```diff
@@ -19,6 +19,8 @@ import Flex from 'components/Flex';
 import Footer from 'components/LayoutFooter';
 import Header from 'components/LayoutHeader';
 import {media} from 'theme';
+import loadScript from 'utils/loadScript';
+import {algoliaURL} from 'site-constants';
```
bvaughn: These imports can be removed ^

src/site-constants.js
```diff
@@ -14,5 +14,6 @@
 // the SSR part in node.js we won't have a proper location.
 const urlRoot = 'https://reactjs.org';
 const version = '16.0.0';
+const babelURL = '//unpkg.com/babel-standalone@6.26.0/babel.min.js';
 
-export {urlRoot, version};
+export {urlRoot, version, babelURL,};
```
bvaughn: Run `yarn prettier` to fix this formatting ^ (the trailing comma should be removed)

renanpvaz:
Well, most of the changes are not related to Babel, I was having trouble trying to render the placeholders because of the way that `mountCodeExample` works and the html structure of the examples. Trying not to touch index.md and thinking that you meant a more elaborate placeholder, I decided to remove all the specific styling that `CodeEditor` had for `children` and instead did something similar inside `mountCodeExample`. Besides that, one of the things I was trying to solve is the "raw" appearance (with the backticks) of the examples before `mountCodeExample` is called.

Anyways, if you're unsure about the changes I could do what you envisioned, but I think that the transition would still look rough once Babel loads.

bvaughn:
Thanks for explaining. 😄 

I'd like to avoid making any changes (at least bigs ones) to `CodeEditor` for the time being.

We don't have to use the `mountCodeExample` util to set the placeholder. It's just a convenience wrapper for code that is controlled by the `CodeEditor`. Our placeholders could just be mounted with `ReactDOM.render(...)`

Let's do with a simple placeholder right now (just a string is fine) and myself and/or @joecritch will follow up on it.

bvaughn:
Apologies for forgetting about the hackiness we were doing inside of `mountCodeExample`. Now I understand why you took the original approach you did with this PR.

The home page content and markdown is all pretty janky until #97 gets resolved unfortunately, so 
it's probably okay for now to just live with a less-than-perfect solution.

Ideally we'd show a nice placeholder when loading is in progress and a nice error if loading failed. When I say "nice" I don't mean "fancy" just- something that will prevent the page from re-flowing text drastically when things load in. Unfortunately this isn't possible without _more_ hacking b'c of how our content is put together and what we're doing in `mountCodeExample`.

How do you feel about (for now) further simplifying this to:
```diff
diff --git a/src/templates/home.js b/src/templates/home.js
index 5131557e3..cb32c683c 100644
--- a/src/templates/home.js
+++ b/src/templates/home.js
@@ -21,10 +21,10 @@ import ReactDOM from 'react-dom';
 
 class Home extends Component {
   componentDidMount() {
-    this.renderExamplePlaceholder('helloExample');
-    this.renderExamplePlaceholder('timerExample');
-    this.renderExamplePlaceholder('todoExample');
-    this.renderExamplePlaceholder('markdownExample');
+    renderExamplePlaceholder('helloExample');
+    renderExamplePlaceholder('timerExample');
+    renderExamplePlaceholder('todoExample');
+    renderExamplePlaceholder('markdownExample');
 
     loadScript(babelURL).then(() => {
       mountCodeExample('helloExample', HELLO_COMPONENT);
@@ -34,25 +34,6 @@ class Home extends Component {
     });
   }
 
-  renderExamplePlaceholder(containerId) {
-    ReactDOM.render(
-      <strong
-        css={{
-          transform: 'translate(65%, -5em)',
-          display: 'block',
-
-          [media.lessThan('xlarge')]: {
-            transform: 'none',
-            marginTop: 15,
-            textAlign: 'center',
-          },
-        }}>
-        Loading code example...
-      </strong>,
-      document.getElementById(containerId),
-    );
-  }
-
   render() {
     const {data} = this.props;
     const title = 'React - A JavaScript library for building user interfaces';
@@ -341,6 +322,16 @@ const markdownStyles = {
   },
 };
 
+// TODO Improve this temporarily placeholder as part of 
+// converting the home page from markdown template to a Gatsby
+// page (see issue #2)
+function renderExamplePlaceholder(containerId) {
+  ReactDOM.render(
+    <h4>Loading code example...</h4>,
+    document.getElementById(containerId),
+  );
+}
+
 // TODO Move these hard-coded examples into example files and out of the template?
 // Alternately, move them into the markdown and transform them during build?
 // This could be done via a new Gatsby transform plug-in that auto-converts to runnable REPLs?
```

renanpvaz:
Sure thing! Oh, and thanks for all the patience since this is my first PR with actual code.

bvaughn:
No thanks needed! Thanks for contributing and being so easy to work with and iterate on this stuff 😄

PR #104, opened 2017-10-09
Added fbjs eslint config, fixed lint errors and warnings

jxom:
I decided to remove `eslint:recommended` config and just use `fbjs` to be consistent with the React project... Let me know if this is better!


bvaughn:
> I decided to remove eslint:recommended config and just use fbjs to be consistent with the React project...

I think this is much better! 😄

src/templates/components/MetaTitle/index.js
```diff
@@ -30,7 +30,9 @@ const MetaTitle = ({
       textTransform: 'uppercase',
       letterSpacing: '0.08em',
       ...cssProps,
-    }}>
+    }}
+    role="link"
+    tabIndex="-1">
```
bvaughn: Hm! It looks like `MetaTitle` is only _clickable_ in one instance- inside of `Section` component. Technically we only need `role="button"` and `tabIndex={0}` in that case, and it's misattributed in the others (since it's not clickable).

What do you think about maybe removing `onClick` (and the role and tab index) from this component and moving that big into `Section` instead? Something like...

```diff
diff --git a/src/templates/components/MetaTitle/index.js b/src/templates/components/MetaTitle/index.js
index 165ad7b5b..6d02296aa 100644
--- a/src/templates/components/MetaTitle/index.js
+++ b/src/templates/components/MetaTitle/index.js
@@ -12,27 +12,17 @@
 import React from 'react';
 import {colors} from 'theme';
 
-const MetaTitle = ({
-  children,
-  title,
-  cssProps = {},
-  onClick,
-  onDark = false,
-}) => (
+const MetaTitle = ({children, title, cssProps = {}, onDark = false}) => (
   <div
-    onClick={onClick}
     css={{
       color: onDark ? colors.subtleOnDark : colors.subtle,
-      cursor: onClick ? 'pointer' : null,
       fontSize: 14,
       fontWeight: 700,
       lineHeight: 3,
       textTransform: 'uppercase',
       letterSpacing: '0.08em',
       ...cssProps,
-    }}
-    role="link"
-    tabIndex="-1">
+    }}>
     {children}
   </div>
 );
diff --git a/src/templates/components/Sidebar/Section.js b/src/templates/components/Sidebar/Section.js
index 517a89dd1..8f1044e82 100644
--- a/src/templates/components/Sidebar/Section.js
+++ b/src/templates/components/Sidebar/Section.js
@@ -26,32 +26,39 @@ const Section = ({
   section,
 }) => (
   <div>
-    <MetaTitle
+    <button
       onClick={onSectionTitleClick}
-      cssProps={{
-        marginTop: 10,
-
-        [media.greaterThan('small')]: {
-          color: isActive ? colors.text : colors.subtle,
-
-          ':hover': {
-            color: colors.text,
-          },
-        },
+      css={{
+        appearance: 'none',
+        background: 'transparent',
+        border: 0,
       }}>
-      {section.title}
-      <ChevronSvg
+      <MetaTitle
         cssProps={{
-          marginLeft: 7,
-          transform: isActive ? 'rotateX(180deg)' : 'rotateX(0deg)',
-          transition: 'transform 0.2s ease',
+          marginTop: 10,
 
-          [media.lessThan('small')]: {
-            display: 'none',
+          [media.greaterThan('small')]: {
+            color: isActive ? colors.text : colors.subtle,
+
+            ':hover': {
+              color: colors.text,
+            },
           },
-        }}
-      />
-    </MetaTitle>
+        }}>
+        {section.title}
+        <ChevronSvg
+          cssProps={{
+            marginLeft: 7,
+            transform: isActive ? 'rotateX(180deg)' : 'rotateX(0deg)',
+            transition: 'transform 0.2s ease',
+
+            [media.lessThan('small')]: {
+              display: 'none',
+            },
+          }}
+        />
+      </MetaTitle>
+    </button>
     <ul
       css={{
         marginBottom: 10,
```

src/components/StickyResponsiveSidebar/StickyResponsiveSidebar.js
```diff
@@ -147,7 +147,9 @@ class StickyResponsiveSidebar extends Component {
             boxShadow: '0 0 20px rgba(0, 0, 0, 0.3)',
             [media.lessThan('small')]: smallScreenBottomBarStyles,
           }}
-          onClick={this._openNavMenu}>
+          onClick={this._openNavMenu}
+          role="button"
+          tabIndex="-1">
```
bvaughn: This should be `tabIndex={0}` so it's keyboard accessible.

jxom:
Sounds great, @bvaughn! Will make those changes.

jxom:
@bvaughn - Seems like the button produces a blue outline when clicked (`:focus`). Removing it with `outline: 'none'` will also encounter accessibility issues... What do you reckon should we do? Leave it, or remove the outline completely?

bvaughn:
Yup. The outline is important for keyboard users until/unless we provide a different focus style (cc @joecritch). I think for now we should leave it in.

jxom:
Awesome!

Issue #84, opened 2017-10-08
Bugs and issues umbrella


bvaughn:
We just completely rewrote docs in facebook/react/issues/7864.
They're bound to cause some issues, have some mistakes, etc.
Let's keep this issue as an umbrella for all problems we're discovering after the initial rollout.

## Claimable
- [ ] Mention in default index.html that it is not production-ready
- [ ] Figure out what to do with Community section (https://github.com/facebook/react/pull/6099)

## In-progress
- [x] @annuhdo: Mark `mockComponent` as legacy and unnecessary (react/issues/2499#issuecomment-94044857)
- [x] @ishankbahl: `shallowCompare` on Addons page should recommend `PureComponent` instead

If you'd like to claim any of the above items, leave a comment and we'll update this issue!

Migrated from facebook/react/issues/8035

annuhdo:
Ooh, I'll grab the issue "Mark mockComponent as legacy and unnecessary" assuming it's to change this section: https://reactjs.org/docs/test-utils.html#mockcomponent

bvaughn:
Okay it's yours

ishankbahl:
Made a PR for "`shallowCompare` on Addons page should recommend `PureComponent` instead".

ishankbahl:
@bvaughn the last two issues had been solved, maybe someone forgot to tick them off at https://github.com/facebook/react/issues/8035 .

bvaughn:
Yup! You're right! String refs are mentioned [here](https://reactjs.org/docs/refs-and-the-dom.html#legacy-api-string-refs) and it looks like all of the forms examples show controlled components now. Thanks for pointing this out! Description has been updated.

shawninder:
I added an issue about the docs in the wrong place so it was closed. Hopefully this is the right place: https://github.com/facebook/react/issues/11221

PR #106, opened 2017-10-09
Update site's doc mockComponent()

annuhdo:
Issue #84 Happy Thanksgiving! 🦃 

Not sure if I was supposed to include that it is preferred to use [shallow rendering](https://github.com/facebook/react/issues/2499#issuecomment-110800356) but I added that mockComponent() is now legacy, unnecessary, and a [similar pattern can be implemented directly](https://github.com/facebook/react/issues/11019#issue-261957590) if user wishes to do so.

Please review and let me know if I'm missing anything 💃

image

### DIFF
--- a/content/docs/addons-test-utils.md
+++ b/content/docs/addons-test-utils.md
@@ -112,6 +112,10 @@ mockComponent(
 
 Pass a mocked component module to this method to augment it with useful methods that allow it to be used as a dummy React component. Instead of rendering as usual, the component will become a simple `<div>` (or other tag if `mockTagName` is provided) containing any provided children.
 
+> Note:
+>
+> `mockComponent()` is a legacy method that is no longer needed. A similar method can be implemented directly.
+
 * * *
 
 ### `isElement()`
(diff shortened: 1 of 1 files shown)


content/docs/addons-test-utils.md
```diff
@@ -112,6 +112,10 @@ mockComponent(
 
 Pass a mocked component module to this method to augment it with useful methods that allow it to be used as a dummy React component. Instead of rendering as usual, the component will become a simple `<div>` (or other tag if `mockTagName` is provided) containing any provided children.
 
+> Note:
+>
+> `mockComponent()` is a legacy method that is no longer needed. A similar pattern can be implemented directly.
```
bvaughn: How do you feel about the following alternative wording?

> `mockComponent()` is a legacy API. We recommend using [shallow rendering](/docs/test-utils.html#shallow-rendering) instead.

gaearon: shallow rendering or [`jest.mock()`](https://facebook.github.io/jest/docs/en/tutorial-react-native.html#mock-native-modules-using-jestmock) if you use Jest.

annuhdo: Nice! I like that there's recommendations in this alternate wording -- added a commit :)

PR #112, opened 2017-10-10
Fix missing closing braces in code samples

fuchodeveloper:
Fixed the closing braces for the code samples at https://reactjs.org/docs/lifting-state-up.html#lifting-state-up

### DIFF
--- a/content/docs/lifting-state-up.md
+++ b/content/docs/lifting-state-up.md
@@ -166,6 +166,8 @@ class TemperatureInput extends React.Component {
 
   render() {
     const temperature = this.state.temperature;
+    }
+} 
 ```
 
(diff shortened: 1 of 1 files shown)


gaearon:
These aren't missing braces, they just show a part of the code.
Maybe we can add

```js
// ...
```

instead?

fuchodeveloper:
Ok.
Adding
`// ...`
That would be great.

gaearon:
Mind making the change? Thanks!

fuchodeveloper:
Effected the change

gaearon:
Seems like we fixed this in https://github.com/reactjs/reactjs.org/pull/115.

PR #114, opened 2017-10-10
Fixes #113 Sidebar collapse and findSectionForPath

### LINKED ISSUE #113: Sidebar does not collapse when clicked on Nav title.
* The sidebar navigation links in `Tutorial`, `Community` and `Blog`  does not collapse when clicked.
* The active Title for `Tutorial` and `Blog` is undefined. Turns out `findSectionForPath` is sending `undefined` for them.

fumblehool:
* `findSectionForPath` sends the first link as active page rather than `undefined`.
* Sidebar will collapse if clicked on nav title.  
* Fixes #113

### DIFF
--- a/src/templates/components/Sidebar/Sidebar.js
+++ b/src/templates/components/Sidebar/Sidebar.js
@@ -57,7 +57,7 @@ class Sidebar extends Component {
         {sectionList.map((section, index) => (
           <SectionComponent
             createLink={createLink}
-            isActive={activeSection === section || sectionList.length === 1}
+            isActive={activeSection === section}
             key={index}
             location={location}
             onLinkClick={closeParentMenu}
--- a/src/utils/findSectionForPath.js
+++ b/src/utils/findSectionForPath.js
@@ -16,7 +16,7 @@ import slugify from './slugify';
  * This method specifically works with the nav_*.yml format.
  */
 const findSectionForPath = (pathname, sections) => {
-  let activeSection;
+  let activeSection = sections[0];
   const slugId = pathname.split('/').slice(-1)[0];
 
   sections.forEach(section => {


src/templates/components/Sidebar/Sidebar.js
```diff
@@ -57,7 +57,7 @@ class Sidebar extends Component {
         {sectionList.map((section, index) => (
           <SectionComponent
             createLink={createLink}
-            isActive={activeSection === section || sectionList.length === 1}
+            isActive={activeSection === section}
```
bvaughn: This behavior was intentional ^

src/utils/findSectionForPath.js
```diff
@@ -16,7 +16,7 @@ import slugify from './slugify';
  * This method specifically works with the nav_*.yml format.
  */
 const findSectionForPath = (pathname, sections) => {
-  let activeSection;
+  let activeSection = sections[0];
```
bvaughn: This behavior could cause false positives for pages that are, for example, in the `docs` folder but are intentionally no longer linked to from the side nav.

src/templates/components/Sidebar/Sidebar.js
fumblehool: I think carret (ChevronSvg) gives user a sense of some kind of `dropdown` or `collapse` action if user clicks on it. 
What do you think?

bvaughn: Yeah, I suppose that's fair. But I don't think I understand the purpose/benefit of collapsing a single section menu.

bvaughn:
It's been a while since there were any commits pushed to this branch, so I'm going to close this PR. Happy to pick back up where we left off if new changes are made though! Thanks!

PR #116, opened 2017-10-10
Fixes #25 - added warning to documentation

### LINKED ISSUE #25: Document that you can't rely on React 16 SSR patching up differences
This issue was originally filed by @Tarnadas at facebook/react/issues/10591. There's a lot of interesting discussion there. This is my (@bvaughn) attempt to summarize based on a [recent comment](https://github.com/facebook/react/issues/10591#issuecomment-334500962) by @dustinsoftware:

The [React 16 blog post](https://reactjs.org/blog/2017/09/26/react-v16.0.html) calls out that mismatches are not recommended, but indicates that some changes are OK (e.g. timestamps). This page, and the React docs, could be updated to point out that it's dangerous to have missing nodes on the server render, and they might get re-created with incorrect attributes.

fumblehool:
* Fixes #25


bvaughn:
Maybe @sebmarkbage (or someone else that has more context on SSR+hydration) can review this PR to make sure it's sufficient.

dustinsoftware:
Thanks! This warning is what I had in mind :)

sebmarkbage:
It's a bit confusing wording because the missing nodes won't get the wrong attributes, it's the other nodes that might get the wrong attributes if one sibling is missing.

bvaughn:
Thanks for weighing in @sebmarkbage. Would this be an improvement?

> However, it's dangerous to have missing nodes on the server render as this might cause sibling nodes to be created with incorrect attributes.

sebmarkbage:
Sounds good

bvaughn:
Cool beans. Changed in 8acec80. Thanks again!

PR #118, opened 2017-10-10
Improved the documentation for React.Children.only

### LINKED ISSUE #87: Improve documentation for React.Children.only
This issue was originally reported by @jimfb via facebook/react/issues/4424

Please refer to the original issue for the related discussion thread.

My understanding of the outcome is that the docs and the implementation are slightly misaligned. The docs currently say:
> Returns the only child in `children`. Throws otherwise.

This is somewhat ambiguous though with regard to how it handles an array with only 1 child element. The [actual implementation](https://github.com/facebook/react/blob/4131af3e4bf52f3a003537ec95a1655147c81270/src/isomorphic/children/onlyChild.js#L29-L35) uses a [helper method](https://github.com/facebook/react/blob/4131af3e4bf52f3a003537ec95a1655147c81270/src/isomorphic/classic/element/ReactElement.js#L373-L379) to verify that the child is an object (eg not in an array).

Let's add some clarification to the docs about exactly what this method does.

renatoselenica:
Fixes #87

Changed the documentation for `React.Children.only()` to make its implementation simpler to understand. 

Since a lot of issues related with this method are connected to the fact that one cannot use a return value from `React.Children.map()`, added a note to explain why map does not return values that can be accessed used by `React.Children.only()`. This can easily be tested in a simple implementation by console logging the return value of map to `React.isValidElement()`, which is what the children.only method checks for itself. This will return false. The structure of the children should still be that of an object for the method to accept it and return it. 

I think this note is necessary but I am unsure on the grammatical structure to make it understandable. If you could please check and maybe suggest a way for it to make sense to future readers of the docs.

### DIFF
--- a/content/docs/reference-react.md
+++ b/content/docs/reference-react.md
@@ -176,7 +176,11 @@ Returns the total number of components in `children`, equal to the number of tim
 React.Children.only(children)
 ```
 
-Returns the only child in `children`. Throws otherwise.
+Verifies that `children` has only one child (a React element) and returns it. Otherwise this method throws.
+
+> Note:
+>
+>`React.Children.only()` does not accept the return value of [`React.Children.map()`](#reactchildrenmap) because it is an array rather than a React element.
 
 #### `React.Children.toArray`
 


content/docs/reference-react.md
```diff
@@ -176,7 +176,11 @@ Returns the total number of components in `children`, equal to the number of tim
 React.Children.only(children)
 ```
 
-Returns the only child in `children`. Throws otherwise.
+Verifies `children` has only one child and the child is a React element, then returns it. Throws otherwise.
+
+> Note:
+>
+>`React.Children.only()` does not accept as argument a return value from [`React.Children.map()`](#reactchildrenmap) because the return of the latest is not a React element.
```
bvaughn: What do you think about

> Verifies that `children` has only one child (a React element) and returns it. Otherwise this method throws.

>> Note:
>>
>> `React.Children.only()` does not accept the return value of [`React.Children.map()`](#reactchildrenmap) because it is an array rather than a React element.

renatoselenica: Looks great. I will make the change right now.

PR #119, opened 2017-10-10
Improve writing style in Reconciliation Tradeoffs

rosenfeld:
I've improved my previous changes to what re-render means by following Dan Abramov's
suggestion on the writing style to remove the parenthesis from that paragraph.

Here's the comment for more context:

https://github.com/reactjs/reactjs.org/pull/108#issuecomment-335511421


gaearon:
Awesome. Thanks.

PR #121, opened 2017-10-10
[Documentation] Re-add "Transferring with ... in JSX" to "JSX In-Depth / Spread Attributes"

hellobrian:
resolves https://github.com/reactjs/reactjs.org/issues/109

Re-add [Transferring with `...` in JSX](https://github.com/facebook/react/blob/v15.4.0-rc.3/docs/docs/06-transferring-props.md#transferring-with--in-jsx) docs previously published with React v15-ish.

The overall content has been kept the same but I slightly changed the example to better reflect why using `...other` can be useful (I also wanted to avoid an example that uses a `<div>` as a checkbox).

In the example, I show a PrimaryButton component that can receive an unknown `onClick` prop function. Things like buttons handle a lot of different events -- instead of passing every possible synthetic event down as a prop, making use of `...other` helps to keep the code for `PrimaryButton` from getting bloated.

If it helps, I've also made the example live on code sandbox but didn't include it...I wasn't sure if it was totally needed, and since the docs surface codepen examples exclusively. I don't know if you all want to stay consistent with one example platform.

[Edit React: Transferring with `...` in JSX](https://codesandbox.io/s/j2vzz8j0z5)

### DIFF
--- a/content/docs/jsx-in-depth.md
+++ b/content/docs/jsx-in-depth.md
@@ -232,20 +232,40 @@ In general, we don't recommend using this because it can be confused with the [E
 
 ### Spread Attributes
 
-If you already have `props` as an object, and you want to pass it in JSX, you can use `...` as a "spread" operator to pass the whole props object. These two components are equivalent:
+Spread attributes can be useful when you are building generic containers. However, they can also make your code messy by making it easy to pass a lot of irrelevant props to components that don't care about them. We recommend that you use this syntax sparingly.
+
+An alternative to spreading all of your props is to specifically list out all the properties that you would like to consume, followed by `...other`.
 
 ```js{7}
-function App1() {
-  return <Greeting firstName="Ben" lastName="Hector" />;
-}
-
-function App2() {
-  const props = {firstName: 'Ben', lastName: 'Hector'};
-  return <Greeting {...props} />;
-}
+const { type, children, ...other } = props;
 ```
 
-Spread attributes can be useful when you are building generic containers. However, they can also make your code messy by making it easy to pass a lot of irrelevant props to components that don't care about them. We recommend that you use this syntax sparingly.
+This ensures that you pass down all the props *except* the ones you're consuming yourself.
+
+```js{7}
+const handleClick = () => console.log("handleClick");
+
+const PrimaryButton = props => {
+  const { type, children, ...other } = props;
+  return (
+    <button type={type} {...other}>
+      {children}
+    </button>
+  );
+};
+
+const App = () => {
+  return (
+    <div>
+      <PrimaryButton type="button" onClick={handleClick}>
+        Hello World!
+      </PrimaryButton>
+    </div>
+  );
+};
+```
+
+In the example above, the `...other` object contains `{ onClick: handleClick() }` but not the `type` or `children` props. This makes a component like the `PrimaryButton` more reusable and flexible because it can extract a set of unknown properties like `onClick` in the above example.
 
 ## Children in JSX
 
--- a/content/docs/jsx-in-depth.md
+++ b/content/docs/jsx-in-depth.md
@@ -232,23 +232,36 @@ In general, we don't recommend using this because it can be confused with the [E
 
 ### Spread Attributes
 
+If you already have `props` as an object, and you want to pass it in JSX, you can use `...` as a "spread" operator to pass the whole props object. These two components are equivalent:
+
+```js{7}
+function App1() {
+  return <Greeting firstName="Ben" lastName="Hector" />;
+}
+
+function App2() {
+  const props = {firstName: 'Ben', lastName: 'Hector'};
+  return <Greeting {...props} />;
+}
+```
+
 Spread attributes can be useful when you are building generic containers. However, they can also make your code messy by making it easy to pass a lot of irrelevant props to components that don't care about them. We recommend that you use this syntax sparingly.
 
 An alternative to spreading all of your props is to specifically list out all the properties that you would like to consume, followed by `...other`.
 
-```js{7}
-const { type, children, ...other } = props;
+```js
+const { children, ...other } = props;
 ```
 
 This ensures that you pass down all the props *except* the ones you're consuming yourself.
 
-```js{7}
+```js{6}
 const handleClick = () => console.log("handleClick");
 
 const PrimaryButton = props => {
-  const { type, children, ...other } = props;
+  const { children, ...other } = props;
   return (
-    <button type={type} {...other}>
+    <button {...other}>
       {children}
     </button>
   );
@@ -265,7 +278,7 @@ const App = () => {
 };
 ```
 
-In the example above, the `...other` object contains `{ onClick: handleClick() }` but not the `type` or `children` props. This makes a component like the `PrimaryButton` more reusable and flexible because it can extract a set of unknown properties like `onClick` in the above example.
+In the example above, the `...other` object contains `{ onClick: handleClick(), type: "button" }` but *not* the `children` prop. This makes a component like the `PrimaryButton` more reusable and flexible because it can extract a set of unknown properties like `onClick` in the above example.
 
 ## Children in JSX
 
--- a/content/docs/jsx-in-depth.md
+++ b/content/docs/jsx-in-depth.md
@@ -245,8 +245,7 @@ function App2() {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


content/docs/jsx-in-depth.md
```diff
@@ -232,20 +232,40 @@ In general, we don't recommend using this because it can be confused with the [E
...
+const { type, children, ...other } = props;
+```
 
-function App2() {
-  const props = {firstName: 'Ben', lastName: 'Hector'};
-  return <Greeting {...props} />;
-}
+This ensures that you pass down all the props *except* the ones you're consuming yourself.
+
+```js{7}
```
bvaughn: This line number (7) no longer points to the thing I think you want to highlight. I believe you want line 6 (where the `other` props are actually passed to `<button>`)>

content/docs/jsx-in-depth.md
```diff
@@ -232,20 +232,40 @@ In general, we don't recommend using this because it can be confused with the [E
...
+const handleClick = () => console.log("handleClick");
+
+const PrimaryButton = props => {
+  const { type, children, ...other } = props;
+  return (
+    <button type={type} {...other}>
```
bvaughn: Could we pick a different example? Since in this case `type` would be passed the same regardless of whether you plucked it out of `props` or passed it via `other`.

content/docs/jsx-in-depth.md
```diff
@@ -247,6 +247,39 @@ function App2() {
...
 
+An alternative to spreading all of your props is to specifically list out all the properties that you would like to consume, followed by `...other`.
+
+```js
+const { children, ...other } = props;
+```
+
+This ensures that you pass down all the props *except* the ones you're consuming yourself.
+
+```js{6}
```
hellobrian: I didn't know what this markdown syntax was before, changed it to highlight line 6 like you suggested.

content/docs/jsx-in-depth.md
```diff
@@ -247,6 +247,39 @@ function App2() {
...
+const handleClick = () => console.log("handleClick");
+
+const PrimaryButton = props => {
+  const { children, ...other } = props;
```
hellobrian: > Could we pick a different example? Since in this case `type` would be passed the same regardless of whether you plucked it out of props or passed it via `other`. - @bvaughn

To your [original comment](https://github.com/reactjs/reactjs.org/pull/121#discussion_r143798264), I updated the example removing `type` from this line. I wasn't trying to focus on `type` but I was trying to show how `PrimaryButton` could now receive unknown props like `onClick`

```jsx
<PrimaryButton type="button" onClick={handleClick}>
```

If this still isn't preferrable, I can just use the original example as-is from the original docs ([Transferring with `...` in JSX](https://github.com/facebook/react/blob/v15.4.0-rc.3/docs/docs/06-transferring-props.md#transferring-with--in-jsx))

bvaughn: I think this is better! But it's still a little awkward since you _could_ also pass `children` via the spread operator as well- and `<button type="button">` isn't a valid attribute. These might be confusing to readers. (They might get distracted thinking, "why is that..." instead of focusing on the spread operator.)

How do you feel about this?
```jsx
const handleClick = () => console.log("handleClick");

const PrimaryButton = props => {
  const { type, ...other } = props;
  const className = type === "primary" ? "PrimaryButton" : "SecondaryButton";
  return <button className={className} {...other} />;
};

const App = () => {
  return (
    <div>
      <PrimaryButton type="primary" onClick={handleClick}>
        Hello World!
      </PrimaryButton>
    </div>
  );
};
```

bvaughn: > `<button type="button">` is a valid attribute

You are correct, of course. Sometimes I forget how wonky HTML attributes are 😆 

```js
const { ...other } = props;
return <button {...other} />;
```

I think this maybe leaves too much to the imagination. "Why would I do this? What's the point?" I think we should show at least one prop being plucked out.

hellobrian: 😆Alright, with everything said so far, 

I can pluck `className` instead of `type` to avoid confusion 
(hopefully, I'm using the "pluck" terminology correctly)

The `PrimaryButton` component takes a `className` prop and any number of unknown props through the `...other` object, just as long as they are valid HTML attributes for `<button>`. 

```jsx
const handleClick = () => console.log("handleClick");

const PrimaryButton = props => {
  const { className, ...other } = props;
  return <button className={className} {...other} />;
};

const App = () => {
  return (
    <div>
      <PrimaryButton className="primaryButton" onClick={handleClick}>
        Hello World!
      </PrimaryButton>
    </div>
  );
};
```
Now, the `App` component can return the `PrimaryButton` so that it can receive it's known props (`className`) *and* any unknown props through `...other`, which contains the following when used in `App`: 

```js
// Contained in { ...other }
{ 
  onClick: handleClick(), 
  children: 'Hello World!' 
}
```

bvaughn: I would have the same concern as I mentioned before:
```js
// This...
const PrimaryButton = props => {
  const { className, ...other } = props;
  return <button className={className} {...other} />;
};
// Is the same thing as this...
const PrimaryButton = props => {
  return <button {...props} />;
};
```

content/docs/jsx-in-depth.md
```diff
@@ -245,7 +245,31 @@ function App2() {
...
 ```
 
-Spread attributes can be useful when you are building generic containers. However, they can also make your code messy by making it easy to pass a lot of irrelevant props to components that don't care about them. We recommend that you use this syntax sparingly.
+You can also pick specific props that your component will consume while passing all other props using the spread operator. 
+This ensures that the component consumes the `kind` prop only, and passes down all other props via `...other`.
+
+```js{2}
+const Button = props => {
+  const { kind, ...other } = props;
+  const className = kind === "primary" ? "PrimaryButton" : "SecondaryButton";
```
hellobrian: > In response to: https://github.com/reactjs/reactjs.org/pull/121#discussion_r144336213 (comment thread was getting long)

Thanks for your patience, I think I finally understand what you've been trying to say 😭 
My motivation for re-adding this documentation was to surface how you can pass unknown props via spread attributes. But, I can see now that there are more considerations that need to be made when opting to use this syntax, which is what you were trying to communicate.  

Thanks again for being so thorough with this PR

bvaughn: No problem at all 😄  It's easy to miscommunicate on a PR.

content/docs/jsx-in-depth.md
```diff
@@ -245,7 +245,31 @@ function App2() {
 }
 ```
 
-Spread attributes can be useful when you are building generic containers. However, they can also make your code messy by making it easy to pass a lot of irrelevant props to components that don't care about them. We recommend that you use this syntax sparingly.
+You can also pick specific props that your component will consume while passing all other props using the spread operator. 
+This ensures that the component consumes the `kind` prop only, and passes down all other props via `...other`.
```
bvaughn: nit: I think this sentence may not be necessary. I think it might be nice to just show the example and then explain it below (like you already do) 👍

content/docs/jsx-in-depth.md
```diff
@@ -245,7 +245,31 @@ function App2() {
...
+    <div>
+      <Button kind="primary" onClick={() => console.log("clicked!")}>
+        Hello World!
+      </Button>
+    </div>
+  );
+};
+```
+
+In the example above, the `kind` prop is safely consumed and *is not* passed directly to the `<button>` element in the DOM.
```
bvaughn: Suggestion / nit:

> In the example above, the `kind` prop is safely consumed and *is not* passed on to the `<button>` element in the DOM.

content/docs/jsx-in-depth.md
```diff
@@ -245,7 +245,31 @@ function App2() {
...
+All other props are passed via the `...other` object making this component really flexible. You can see that it passes an `onClick` and `children` props.
+
+Spread attributes can be useful but not to pass invalid HTML attributes to the DOM. They can also make your code messy by making it easy to pass a lot of irrelevant props to components that don't care about them. We recommend using this syntax sparingly.
```
bvaughn: Suggestion:

> Spread attributes can be useful but they also make it easy to pass unnecessary props to components that don't care about them or to pass invalid HTML attributes to the DOM. We recommend using this syntax sparingly.

PR #123, opened 2017-10-10
Restore sidebar transition when switching from/to small screen/"sticky" sidebar

fk:


fk:
> I'm actually not sure if we want this transition 😁

Yeah, actually asked that myself, too. 😉
FWIW I think it feels OK when going from small to medium, but not so much the other way around.

bvaughn:
We should definitely _either_ fix this property name _or_ remove it entirely, I'm just not sure which. (Leaning towards removing entirely but will see if @joecritch agrees)

joecritch:
@bvaughn Yeah, no transitions when resizing are intentional :)

joecritch:
@bvaughn Sorry, I meant the transitions weren’t intentional. But no sweat, I’ll check out master tomorrow to see how it looks ☺️👍

bvaughn:
Please feel free to change it however you'd like Joe (or just shoot me a Hangouts message with clarification and I'll make the change!) 😄

PR #127, opened 2017-10-10
adds returning multiple elements example to jsx-in-depth guide

jamesvclements:
replaces the section about having to wrap multiple elements in a `div` with the example from the [v16 blog](https://reactjs.org/blog/2017/09/26/react-v16.0.html)

image

### DIFF
--- a/content/docs/jsx-in-depth.md
+++ b/content/docs/jsx-in-depth.md
@@ -308,7 +308,19 @@ You can mix together different types of children, so you can use string literals
 </div>
 ```
 
-A React component can't return multiple React elements, but a single JSX expression can have multiple children, so if you want a component to render multiple things you can wrap it in a `div` like this.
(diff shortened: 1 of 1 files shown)


content/docs/jsx-in-depth.md
```diff
@@ -308,7 +308,19 @@ You can mix together different types of children, so you can use string literals
 </div>
 ```
 
-A React component can't return multiple React elements, but a single JSX expression can have multiple children, so if you want a component to render multiple things you can wrap it in a `div` like this.
+A React component can return multiple elements:
```
bvaughn: How do you feel about adding the word "also" like so?
> A React component can also return an array of elements:

jamesvclements: Sounds good to me! I'll update

PR #131, opened 2017-10-11
Add CircleCI config

jxom:
This intends to fix the third and final item for #67.

Basically just runs the `ci-check` script upon every build, as well as caching the project dependencies 🤗.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,13 @@
+version: 2
+jobs:
+  build:
+    docker:
+      - image: circleci/node:latest
+    steps:
+      - checkout
+      - run:
+          name: Install
+          command: npm install
+      - run:
+          name: Check Prettier, ESLint, Flow
+          command: npm run ci-check
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,16 +7,16 @@ jobs:
       - checkout
       - restore_cache:
           keys:
-          - v1-dependencies-{{ checksum "package.json" }}
+          - dependencies-{{ checksum "package.json" }}
           # fallback to using the latest cache if no exact match is found
-          - v1-dependencies-
+          - dependencies-
       - run:
           name: Install
           command: yarn install
       - save_cache:
           paths:
             - node_modules
-          key: v1-dependencies-{{ checksum "package.json" }}
+          key: dependencies-{{ checksum "package.json" }}
       - run:
           name: Check Prettier, ESLint, Flow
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,22 @@
+version: 2
+jobs:
+  build:
+    docker:
+      - image: circleci/node:latest
+    steps:
+      - checkout
+      - restore_cache:
```
bvaughn: Do we need a step to explicitly install Yarn if it's not  already installed? The [React repo's `circle.yml` file](https://github.com/facebook/react/blob/master/circle.yml) has this:
```
  pre:
    # This is equivalent to $TRAVIS_COMMIT_RANGE
    - echo $CIRCLE_COMPARE_URL | cut -d/ -f7
    # install yarn if it's not already installed
    - |
      if [[ ! -e ~/.yarn/bin/yarn || $(yarn --version) != "${YARN_VERSION}" ]]; then
        curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version $YARN_VERSION
      fi
```

Maybe this isn't necessary with the 2.0 API. The 1.0 API docs have [a section about how to enable Yarn](https://circleci.com/docs/1.0/yarn/) which the 2.0 docs don't have. Maybe it just works out of the box now? :grin: Let's try and see!

.circleci/config.yml
```diff
@@ -0,0 +1,22 @@
...
+          keys:
+          - dependencies-{{ checksum "package.json" }}
```
bvaughn: I wonder if we should `checksum "yarn.lock"` instead? Seems like that may be better?

bvaughn:
Looks good so far 👍 

The first build commit (c4aaa15) failed: https://circleci.com/gh/reactjs/reactjs.org/1

This was because a Prettier issue had slipped in. I fixed it (42372b2) and the second build seems happy: https://circleci.com/gh/reactjs/reactjs.org/2

😄

bvaughn:
We've also got Circle building PRs from forks now too (#134, #135)

.circleci/config.yml
jxom: Pretty sure it’s already installed on v2

.circleci/config.yml
jxom: I agree!

.circleci/config.yml
bvaughn: Yeah, seems like that's the case.

PR #136, opened 2017-10-11
Add `componentDidCatch()` document and link

thomasyimgit:
This is about the issue here: https://github.com/reactjs/reactjs.org/issues/120

Another thing that I noticed is that when I clicked on the link using "open in new tag", it works fine. But when I left-click it directly, it redirects me to the bottom of the blog post (as showed below), rather than the top. I thought it could be a bug?

image

### DIFF
--- a/content/docs/reference-react-component.md
+++ b/content/docs/reference-react-component.md
@@ -62,6 +62,12 @@ This method is called when a component is being removed from the DOM:
 
 - [`componentWillUnmount()`](#componentwillunmount)
 
+#### Error Handling
+
+This method is called when there is an error during rendering, in a lifecycle method, or in the constructor of any child component.
+
+- [`componentDidCatch()`](#componentdidcatch)
+
 ### Other APIs
 
 Each component also provides some other APIs:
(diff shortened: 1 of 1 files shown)


content/docs/reference-react-component.md
```diff
@@ -62,6 +62,12 @@ This method is called when a component is being removed from the DOM:
 
 - [`componentWillUnmount()`](#componentwillunmount)
 
+#### Error Handling
+
+An error handler to catch JavaScript errors anywhere in their **child** component tree. In most of the cases, it logs those errors, and display a fallback UI.
```
bvaughn: What do you think about this wording?
> This method is called when there is an error during rendering, in a lifecycle method, or in the constructor of any child component.

content/docs/reference-react-component.md
```diff
@@ -250,6 +256,20 @@ componentWillUnmount()
 
 * * *
 
+### `componentDidCatch()`
+
+```javascript
+componentDidCatch()
```
bvaughn:
```js
componentDidCatch(error, info)
```

content/docs/reference-react-component.md
```diff
@@ -250,6 +256,20 @@ componentWillUnmount()
...
+```
+
+This is a new feature in React 16 to improve error handling on UI crashes.
+
+A React class component becomes an "error boundary" if this method is declared in it. Generally, it catches JavaScript errors when the rendering of its **children** crashes, logs the error and displays a fallback UI.
+
+For more details, check [*Error Handling in React 16*](/blog/2017/07/26/error-handling-in-react-16.html).
```
bvaughn: What do you think of this wording? (Dropping the "this is a new feature" paragraph and then...)

> Error boundaries are React components that **catch JavaScript errors anywhere in their child component tree, log those errors, and display a fallback UI** instead of the component tree that crashed. Error boundaries catch errors during rendering, in lifecycle methods, and in constructors of the whole tree below them.
>
> A class component becomes an error boundary if it defines this lifecycle method.
>
> For more details, see [*Error Handling in React 16*](/blog/2017/07/26/error-handling-in-react-16.html).
>
> **Note**
> Error boundaries only catch errors in the components below them in the tree. An error boundary can’t catch an error within itself.

thomasyimgit:
Thanks Brian. I gotta admit that yours version is much better. This is my first time contributing to documentation. I'll try to be much more cautious next time. Thanks for your generous help again!

thomasyimgit:
Of course, I'm always willing to learn from people! Thanks

Issue #122, opened 2017-10-10
Migrate examples to CodeSandbox


bvaughn:
The docs site currently links to Codepen.io in several places, eg "Try it on CodePen". This is a maintenance burden for a few reasons:
* These pens are versioned separately from the website and so they can get stale.
* Updating them is a hassle. (Currently updates have to be done by @gaearon since he created the pens.)
* OSS users can't contribute fixes without forking the pens, which would make them harder still to update in the future.

CodeSandbox has a feature to [import pens directly from GitHub](https://codesandbox.io/s/github). This sounds great! If our pens lived in GitHub then it would be much easier to keep them synchronized with docs changes. (Also CodeSandbox is slick ❤️ !)

### Open questions

Ideally we could group all of the current pens together into a single folder (eg `sandbox`) and then deep-link to individual example files. However after a brief look at the CodeSandbox feature, I believe it treats an entire git repository as a single project.

### Considerations

It's possible that the levels of traffic the React website sees would cause scaling problems for CodeSandbox. We might want to try this integration in some limited form at first, with reduced traffic.

cc @CompuIves in case you're interested in collaborating on this

CompuIves:
Absolutely, I'd **love** to collaborate and make this happen!

> Ideally we could group all of the current pens together into a single folder (eg sandbox) and then deep-link to individual example files. However after a brief look at the CodeSandbox feature, I believe it treats an entire git repository as a single project.

Both approaches would work. I'd suggest using the same strategy as redux: create an `examples` folder with `create-react-app` projects and link them separately. An example URL would look like this for `examples/portal` folder:

https://codesandbox.io/s/github/reactjs/reactjs.org/tree/master/examples/portal

The advantage of having separate CRA projects is that the user can easily clone/download the CRA project from GitHub/CodeSandbox and continue working locally. We also show what a possible (recommended) project structure is from the beginning.

> It's possible that the levels of traffic the React website sees would cause scaling problems for CodeSandbox. We might want to try this integration in some limited form at first, with reduced traffic.

Since we leverage the browser for 95% of the work I'm fairly positive that we can handle a lot of traffic. Though maybe we come across other bottlenecks, so I agree that it would be safest to test it in a limited form first. Maybe we can move the examples one by one gradually over, evaluating after every update how the servers hold. We could also create a simple script (like `if (Math.random() > 0.5)`) that allows us to sort of 'AB-test' old vs new, and gradually increase it.

[...]

PR #138, opened 2017-10-11
Convert 2 examples to CodeSandbox

CompuIves:
Related to https://github.com/reactjs/reactjs.org/issues/122.

In this PR I converted 2 different examples over to CodeSandbox:

- Uncontrolled Components
- Portals (1 & 2)

[Uncontrolled Components](https://codesandbox.io/embed/github/CompuIves/reactjs.org/tree/codesandbox/examples/uncontrolled-components?codemirror=1)
[Portals 1](https://codesandbox.io/embed/github/CompuIves/reactjs.org/tree/codesandbox/examples/portals-1?codemirror=1)
[Portals 2](https://codesandbox.io/embed/github/CompuIves/reactjs.org/tree/codesandbox/examples/portals-2?codemirror=1)

For Portals I decided to split the example up in multiple files, since it's one of the more advanced examples. I can always merge it into 1 file if that's desired. 

Regarding page size, @gaearon mentioned a good consideration. CodeSandbox has 2 versions: the embed version and the editor version. The embed version is a stripped down version of the editor and built to be small.

I decided to use the embed version for the examples, but it still downloads 1.8MB before render of editor+preview. This is mainly because we use the Monaco editor. To make the bundle size smaller you can give the option to use CodeMirror instead by appending `?codemirror=1` to the url. Then the size before render will be 900KB and it renders quicker, but you'll lose the vscode editor feeling/features. We'll need to decide on that.

Also, the URLs are currently pointing to my branch, because master doesn't contain the examples yet. Maybe it would be cool to automate this so we can generate custom urls based on branch, that way Netlify demos work without effort.


bvaughn:
> To make the bundle size smaller you can give the option to use CodeMirror instead by appending `?codemirror=1` to the url. Then the size before render will be 900KB and it renders quicker, but you'll lose the vscode editor feeling/features. We'll need to decide on that.

IMO the smaller payload size is better than the vscode polish.

I might be jumping the gun on the preview but it looks like they're all failing at the moment:
screen shot 2017-10-11 at 1 48 38 pm

CompuIves:
Updated the code to use `codemirror`! Regarding your error, that's really strange. I can't reproduce it on Chrome, Safari and Firefox with fresh cache. It could be related to sw-cache. What does your console say and which browser are you using?

CompuIves:
I added the possibility to link to 1 specific file with no boilerplate. In that case it will check if the file is in a `create-react-app` project and use those settings. So for example in this scenario:

```
examples/
  rendering-elements/
    package.json
    public/
    src/
      1.js
      2.js
      3.js
```

You could use these urls:

https://codesandbox.io/embed/github/CompuIves/reactjs.org/tree/codesandbox/examples/rendering-elements/src/1.js?codemirror=1
https://codesandbox.io/embed/github/CompuIves/reactjs.org/tree/codesandbox/examples/rendering-elements/src/2.js?codemirror=1
https://codesandbox.io/embed/github/CompuIves/reactjs.org/tree/codesandbox/examples/rendering-elements/src/3.js?codemirror=1

And you would get 3 different projects with the entry file assigned to the linked file.

I also played a bit with the embed API, and created this:

screen shot 2017-10-11 at 22 42 12

(Top is inline, bottom is codesandbox embed)

You can actually specify line highlights in the source using comments (like `eslint` rules) and they work in embeds, this is really cool since your highlights live in your code (you don't have to sync line numbers) and you don't have to keep the inline code of the doc synced with the git repo.

Curious what you think!

bvaughn:
> I added the possibility to link to 1 specific file with no boilerplate. In that case it will check if the file is in a create-react-app project and use those settings.

This looks nice 👍  since it would cut down significantly on the boilerplate for multi-step examples.

> You can actually specify line highlights in the source using comments (like eslint rules) and they work in embeds, this is really cool since your highlights live in your code (you don't have to sync line numbers) and you don't have to keep the inline code of the doc synced with the git repo.

That _is_ pretty cool 😁

bvaughn:
> Updated the code to use `codemirror`! Regarding your error, that's really strange. I can't reproduce it on Chrome, Safari and Firefox with fresh cache. It could be related to sw-cache. What does your console say and which browser are you using?

Bizarre! I tried the links in an incognito window though and they _do_ work there, just not in my main browser.

I wonder what about these examples would cause them to be broken the first time I load them in an incognito browser? (And if others might encounter this same issue...)

CompuIves:
Right! I'm also exploring more options. I'm currently working on a Gatsby Remark plugin that allows you to add links to the source files in the markdown, these then get converted to the actual prism code highlights with a button to turn it into a CodeSandbox embed. The advantage of this is that you have a single source of truth for the js files, you're also not dependent on an external service just to show code and you automatically have a synced sandbox for every file.


> I wonder what about these examples would cause them to be broken the first time I load them in an incognito browser? (And if others might encounter this same issue...)

Most probably it's because of a cached packager response from a couple months ago. Next week we're moving over to the [new packager](https://hackernoon.com/how-we-make-npm-packages-work-in-the-browser-announcing-the-new-packager-6ce16aa4cee6) with a max cache time of 24h, so hopefully that will be resolved by then.

bvaughn:
> I'm currently working on a Gatsby Remark plugin that allows you to add links to the source files...

Very cool! I had been considering something along those lines as well but haven't gotten around to it yet. Glad to hear you're working on it 😁 

> Most probably it's because of a cached packager response from a couple months ago...

Okay. Good to hear. Service-worker problems like this are the worst 😅

CompuIves:
Okies! I tried something new. 

With this version I added a remark plugin that converts:

```
[Try it on CodeSandbox.](source:examples/uncontrolled-components/src/index.js{11,20})
```

to the Prism Highlight of that file with line highlights and a generated link to the codesandbox sandbox embed based on the branch/user/file. You can see it here: https://deploy-preview-138--reactjs.netlify.com/docs/uncontrolled-components.html.

The next thing I'm going to try is putting a React Live component in the place of the code, this can be a challenge since we may need 2 plugins in that case. One for converting the markdown and then one for hooking in the components. There are two considerations:

1. We lose the line highlighting. I think this is pretty big, since it's valuable to show what we mean in many cases.
2. We should find a way to allow the `import` statements to show. I think these statements can be pretty important, as it allows users to directly copy the code locally and it shows the user the recommended way of using React.

Maybe we can open a PR to react-live to add these 2 points.

A second option would be to replace the highlighted code with an embed as soon as you click on edit. This way we keep the line highlights and module imports. This keeps the user on the same page but still allows the user to fiddle. Big downside is that the user first needs to press a button. I think the React Live version would first be better to try out.

Okay, I'll get back to it this evening!

EricSimons:
@CompuIves this is awesome! Were you planning on taking a stab at turning the "Try it on X" into a dropdown menu? Or should I make a PR to this for it?

Also, reg react-live not having the import statements — I actually think it's better for the docs to not have them there, as it's just redundant code that isn't critical to explaining the core concepts at hand. What if we did the opposite approach, where we strip out any import statements before putting them in the react-live editor? That way when the user clicks on the "try it on x" dropdown menu, the import statements will be there in the "real" environment on CSB/SB/etc (which makes more sense from the learners perspective)

CompuIves:
Pushed a new experimentation. You can now define an example like this:

```
[Run the example.](source:examples/uncontrolled-components/src/index.js{11,20}?editorsize=70)
```

The options after the `?` are options like how big (in %) you'd like the editor to be.

That line will first show the code as a Prism.js with line highlights. When you press 'Run example' the Prism.js highlighted embed will turn into a CodeSandbox embed with the same lines highlighted.

It looks like this:

kapture 2017-10-14 at 23 57 17

EricSimons:
@CompuIves this is cool! Have you started working on the react-live version you mentioned in https://github.com/reactjs/reactjs.org/issues/122#issuecomment-336571580? I'd love to help collab on it!  🙌

CompuIves:
Heyo!

> @CompuIves this is cool! Have you started working on the react-live version you mentioned in #122 (comment)? I'd love to help collab on it! 🙌

Didn't start it, mostly because I'm not sure if it's smart to embed live preview on every doc page for bundle size and react-live doesn't have line highlights yet.

---

Okay! I've converted a different page to give a more 'real-life' example of how it could look like. This page: https://deploy-preview-138--reactjs.netlify.com/docs/state-and-lifecycle.html.

All CodePen examples are now linked to a single file in the repo, these files are automatically synced with both the documentation and CodeSandbox. The last example is interesting, because it's referring a single file from a bigger project. If you reference `index.js` directly, it will use that directory as `src` folder. This means that you can show a single file of a bigger project, with the existing imports. This fits the last example more, because you want to show the 3 `Clock`s instead of a whole file.

I also reduced the embed size.

EricSimons:
@bvaughn proposed the solution for the react-live/babel bundle size issue in https://github.com/reactjs/reactjs.org/issues/122#issuecomment-336922176:

> I like the idea of deferring cost until a user chooses to interact with an example (eg by clicking an "try it" link).

^ that's actually the same solution you're using in this PR, right? So we'd just need to swap out the iframe on CSB for react-live

> react-live doesn't have line highlights yet.

Yes, hence why I asked if you wanted to collab on upgrading it :)

EricSimons:
@CompuIves just checked out your [counters example](https://deploy-preview-138--reactjs.netlify.com/docs/state-and-lifecycle.html) and all of the embeds seem to be completely broken on my machine; not sure if related to the service worker issue @bvaughn hit earlier in this thread. Here's a GIF of it:

screenflow

^ lmk if there's anything I can do to help debug on my end

CompuIves:
I added a force refresh for the first 3 examples, because the setIntervals were complaining.

> ^ that's actually the same solution you're using in this PR, right? So we'd just need to swap out the CSB iframe with react-live

Hmm, yep that could work! 

> not sure if related to the service worker issue @bvaughn hit earlier

Yap it is 😄 . I **finally** know exactly what it is! So our packager generates 2 different files: `dll.js` and `manifest.json` to map the modules to the code. We recently refreshed the cache, which in itself is no problem. But some users only cached the manifest.json and not the dll.js, because dll.js is bigger. Because of this the wrong mapping was assumed and you get the issue you see there.

I'm deploying the first part of the fix now which should fix it for everyone. The next part will be deployed this evening, which will fix it for future cache refreshes. We're going full on with the new packager regardless this week, which doesn't have this issue at all.

EricSimons:
Sweet! Okay perfect, sounds like we're on the same page reg react-live — wanna spin up a branch where we could collab on it? Perhaps it would make sense to fork off of this one, since we'll likely want to reuse parts of your Gatsby plugin?

CompuIves:
I first want to know what @bvaughn thinks of this solution. I've already put a lot of time in making this solution perfect with building the remark plugin, adding `editorsize`, new `hidenavigation` and `forcerefresh` functionality, adding line highlights, upgrading our git-extractor to handle direct file urls/subprojects, and improving our embed size to be much smaller.

I've also already explained my thoughts on why this could be eg. a good gateway (https://github.com/reactjs/reactjs.org/issues/122#issuecomment-336764140), so I'm still in favor of this solution.

EricSimons:
Ah gotchya — I just asked @bvaughn in https://github.com/reactjs/reactjs.org/issues/122#issuecomment-337381981 about this + other remaining q's, so I'm hoping his response there will help clear things up and enable us all to work collaboratively 👍 

> I've already put a lot of time in making this solution perfect

I think we've all put a great deal of time into creating our PR's, debating their merits, and helping figure out the right solution for the React team's needs :)

bvaughn:
Hm. Fresh install (`rm -rf .cache && rm -rf node_modules && yarn install && yarn dev`) fails with:
screen shot 2017-10-21 at 1 08 47 pm

**Edit** Looks like it failed because my local Git repo was detached (b'c I checked out `CompuIves/codesandbox`).

content/docs/portals.md
```diff
@@ -46,7 +46,7 @@ A typical use case for portals is when a parent component has an `overflow: hidd
 >
 > It is important to remember, when working with portals, you'll need to make sure to follow the proper accessibility guidelines.
 
-[Try it on CodePen.](https://codepen.io/gaearon/pen/yzMaBd)
+[Try it on CodeSandbox.](https://codesandbox.io/embed/github/CompuIves/reactjs.org/tree/codesandbox/examples/portals-1?codemirror=1)
```
bvaughn: Might be nice to bundle this up as a custom remark transform. That way we could keep the query params and stuff in one place, and just write Markdown like:
```md
[Try it on CodeSandbox](codesandbox=examples/portals-1)
```

**Edit** Looks like this would be similar to the "Run Example" markdown below 👍

bvaughn: I mentioned this because I was wondering if we could, for example, auto-expand the left side menu on multi-file examples b'c it's not obvious there is more than one file when you open a new sandbox- and it would be nice to not have to find-and-replace a bunch of links to do this.

examples/portals-2/src/Child.js
```diff
@@ -0,0 +1,11 @@
...
+
+export default function Child() {
+  // The click event on this button will bubble up to parent,
+  // because there is no 'onClick' attribute defined
+  return (
+    <div className="modal">
+      <button>Click</button>
+    </div>
+  );
+}
```
bvaughn: The portals examples still have a lot of boilerplate. I see that this adds more flexibility for customization, but I suspect it would also end up being super redundant in the case of most of our examples. 🤔

content/docs/state-and-lifecycle.md
```diff
@@ -13,50 +13,13 @@ So far we have only learned one way to update the UI.
...
-    element,
-    document.getElementById('root')
-  );
-}
-
-setInterval(tick, 1000);
-```
-
-[Try it on CodePen.](http://codepen.io/gaearon/pen/gwoJZk?editors=0010)
+[Run example.](source:examples/single-file-examples/src/state-and-lifecycle/state-and-lifecycle-1.js{11}?editorsize=70&forcerefresh=1)
```
bvaughn: Hm. Just tried this one locally and it failed:
> Failed to load resource: the server responded with a status of 422 ()
> GET https://codesandbox.io/api/v1/sandboxes/github/reactjs/reactjs.org/tree/codesandbox/examples/single-file-examples/src/state-and-lifecycle/state-and-lifecycle-1.js

I think that's because the code it's trying to open only exists on your fork of the repo.

This default behavior would make things hard to test for people adding new runnable examples though?

src/components/MarkdownPage/MarkdownPage.js
```diff
@@ -19,108 +19,136 @@ import toCommaSeparatedList from 'utils/toCommaSeparatedList';
...
+    const codeSandboxLinks = document.querySelectorAll(
+      "a[href^='https://codesandbox.io']",
+    );
 
-              <div css={sharedStyles.articleLayout.content}>
-                <div
-                  css={[sharedStyles.markdown]}
-                  dangerouslySetInnerHTML={{__html: markdownRemark.html}}
-                />
+    codeSandboxLinks.forEach(link => {
```
bvaughn: I think we could do this with a custom remark plugin and remove the runtime document querying for better perf.

content/docs/portals.md
bvaughn: This could also have the same issue I mentioned below though, of being difficult to test initially (at least for a reviewer).

src/components/MarkdownPage/MarkdownPage.js
bvaughn: This breaks IE. ☹️ 

`querySelectorAll` returns a `NodeList` rather than an Array, and the `NodeList` doesn't have a `forEach` method in IE.

Changing to a `for-in` statement fixes it, or...a custom remark plugin would be even better.

CompuIves:
Really good feedback, thanks! I'll do some thinking on this tonight 😄

examples/portals-2/src/Child.js
CompuIves: You're right, we can move this already to a version that has less boilerplate (same how `state-and-lifecycle-7` works).

bvaughn:
Happy to chat about this on Twitter or Slack if you'd like to run ideas by someone Ives.

bvaughn:
Closing this PR for now in favor of a more incremental solutions (#245). Let's keep the conversation going here though, either on GitHub or on Twitter directly. I think the ideas shown here are exciting.

Issue #99, opened 2017-10-09
Interactive code samples should support a non-JSX view


bvaughn:
On the old website, the interactive code samples on [the home page](https://reactjs.org/) used to have 3 tabs:
* JSX editor
* Non-JSX editor
* Preview

The non-JSX tab showed what the JSX compiled to (eg `<div/>` => `React.createElement('div')`). This was a useful view for people who are new to JSX or who just feel more comfortable writing regular JavaScript.

We should add this view back to the code editor.

Please work with @joecritch and @bvaughn if you'd like feedback on UI/UX.

bvaughn:
Closed via #141

PR #141, opened 2017-10-11
add non-jsx view to interactive code samples

criticalbh:
#99 

Hello @bvaughn . 

Here is implementation of code interaction view:
image

and once clicked on `VIEW COMPILED JSX` we get this:
image

### DIFF
--- a/src/components/CodeEditor/CodeEditor.js
+++ b/src/components/CodeEditor/CodeEditor.js
@@ -312,7 +312,7 @@ class CodeEditor extends Component {
   }
 
   _onChange = code => {
-    this.setState(this._updateState(code, this.state.showJSX));
+    this.setState(state => this._updateState(code, state.showJSX));
   };
 }
 
--- a/src/components/CodeEditor/CodeEditor.js
+++ b/src/components/CodeEditor/CodeEditor.js
@@ -14,14 +14,13 @@ import Remarkable from 'remarkable';
 import {LiveEditor, LiveProvider} from '@gaearon/react-live';
 import {colors, media} from 'theme';
 import MetaTitle from 'templates/components/MetaTitle';
-import {highlight, languages} from 'prismjs/components/prism-core'
-import 'prismjs/components/prism-javascript'
+import {highlight, languages} from 'prismjs/components/prism-core';
+import 'prismjs/components/prism-javascript';
 
 const compile = code =>
   Babel.transform(code, {presets: ['es2015', 'react']}).code; // eslint-disable-line no-undef
 
-const compileJsxToJS = code =>
-  Babel.transform(code, {presets: ['react']}).code; // eslint-disable-line no-undef
+const compileJsxToJS = code => Babel.transform(code, {presets: ['react']}).code; // eslint-disable-line no-undef
 
 const prism = code => highlight(code, languages.javascript);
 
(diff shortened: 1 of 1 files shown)


src/components/CodeEditor/CodeEditor.js
```diff
@@ -112,7 +118,27 @@ class CodeEditor extends Component {
...
+                  {showCompiledJsx && (
+                    <a
+                      css={{cursor: 'pointer'}}
+                      role="button"
+                      tabIndex="0"
+                      onClick={this._hideCompiledJsxPreview.bind(this)}>
+                      Back to JSX Editor
+                    </a>
+                  )}
+                </MetaTitle>
```
bvaughn: How do you feel about using a checkbox to toggle JSX mode instead of a button? I think it might look better 😄 

screen shot 2017-10-12 at 10 23 20 am

criticalbh:
@bvaughn Thank you very much, I'll jump on it.

src/components/CodeEditor/CodeEditor.js
```diff
@@ -139,7 +165,14 @@ class CodeEditor extends Component {
...
                 className="gatsby-highlight">
-                <LiveEditor onChange={this._onChange} />
+                {!showCompiledJsx && <LiveEditor onChange={this._onChange} />}
+                {showCompiledJsx && (
+                  <pre
+                    className="prism-code"
+                    spellCheck="false"
+                    dangerouslySetInnerHTML={{__html: prism(compiledJsx)}}
+                  />
+                )}
```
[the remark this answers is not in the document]
criticalbh: In this case, compiledES6 would be editable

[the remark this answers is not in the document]
bvaughn: Ah, good point. I guess this wouldn't quite work as proposed. Maybe we should make the ES6 preview not editable?

criticalbh:
@bvaughn I have made both ES5 and ES6 editable. 

There might be small problem when coming back to `JSX view` since what we have changed in JS wont be applied to JSX. 

If this scenario is not acceptable than we can make `JSX compiled view` to be preview only.

src/components/CodeEditor/CodeEditor.js
```diff
@@ -282,7 +312,7 @@ class CodeEditor extends Component {
   }
 
   _onChange = code => {
-    this.setState(this._updateState(code));
+    this.setState(this._updateState(code, this.state.showJSX));
```
bvaughn: We should change this to:

```js
this.setState(state => this._updateState(code, state.showJSX));
```

The benefit of using a callback for `setState` is that it avoids references potentially stale values in `this.state`. This didn't matter before because we weren't references `this.state` but now that we are....we should use the callback form. 😄

src/components/CodeEditor/CodeEditor.js
```diff
@@ -265,12 +286,21 @@ class CodeEditor extends Component {
     this._mountNode = ref;
   };
 
-  _updateState(code) {
+  _updateState(code, showJSX = true) {
     try {
-      return {
-        compiled: compile(code),
+      let newState = {
+        compiled: compileES5(code),
```
bvaughn: Should `code` always go into `newState`? What's the benefit of only adding it conditionally.

criticalbh: I think no because we want to keep code only if we are in JSX view. 

If we mix code from ES5 view to ES6 it wont work any more.

criticalbh:
Yes indeed, it is cost for ES5 being editable. Thank you very much for guidance it was awesome.

criticalbh:
Hello @bvaughn, what are the reasons of this PR being overwritten with https://github.com/reactjs/reactjs.org/commit/2b44cd3df3d91de4d395317b3234ea80286c2a2b ?

bvaughn:
It wasn't overridden? I just tweaked the wording slightly when testing the changes to the home template. (Without those changes, a blocked Babel script just caused the home page to show "Loading examples" forever.)

criticalbh:
Yes I saw the reason, also sometimes after the page is started and you scroll down, new refresh happens with editors being in the place.

But regarding is it overridden, it is. Here is raw file on master branch for `CodeEditor.js` https://raw.githubusercontent.com/reactjs/reactjs.org/master/src/components/CodeEditor/CodeEditor.js and there are no ES5 transpilers.

bvaughn:
Oh whoa! I see what you mean! Looks like something went wrong with the merge. Thank you for pointing that out! I'll fix.

criticalbh:
@bvaughn your welcome, no problems.

bvaughn:
😅  too much context-switching this morning.

Very sorry this happened. I'll get it fixed right away.

criticalbh:
hh I believe, no problems at all, it is good it is noticed on time.

PR #150, opened 2017-10-12
Refactor home page

yangshun:
This PR is part one of the refactoring of the home page issue #97 . I didn't want the PR to become too big so I left out refactoring of the examples section. I wanted to get some opinions on the current approach before starting on the examples section as well.

This PR does a few things:

- Shift `templates/home.js` to `pages/index.js`
- Transfer the content of `content/index.md` into `pages/index.js`
- Break up the home page marketing content into markdown files and load the data via GraphQL
- Remove hardcoded class names from home page marketing section

I have removed most of the hardcoded class names. There are some leftover styles because I have yet to refactor the examples section. I intend to refactor that part in a similar fashion and extract out the text into Markdown files and also load them via GraphQL. Will follow up with another PR for the examples section later.

cc @bvaughn


content/marketing/component-based.md
```diff
@@ -0,0 +1,8 @@
+---
+title: Component-Based
+order: 1
```
yangshun: Used to decide the ordering of the 3-column marketing content. `order` might not be the best name for this and suggestions are welcome! I briefly considered `rank`.

gatsby-node.js
```diff
@@ -60,15 +60,7 @@ exports.createPages = async ({graphql, boundActionCreators}) => {
   allMarkdown.data.allMarkdownRemark.edges.forEach(edge => {
     const slug = edge.node.fields.slug;
 
-    if (slug === '/index.html') {
```
yangshun: No longer need to programatically create a page because the home page is now within the `pages` directory.

src/pages/index.js
```diff
@@ -6,7 +6,7 @@
 
 'use strict';
 
-import ButtonLink from './components/ButtonLink';
+import ButtonLink from '../templates/components/ButtonLink';
```
yangshun: Should `<ButtonLink>` be moved to `src/components`?

src/pages/index.js
```diff
@@ -135,10 +139,137 @@ class Home extends Component {
...
+                )}
+              </div>
+            </section>
+            <hr css={{
+              height: 1,
+              marginBottom: -1,
+              border: 'none',
+              borderBottom: `1 solid ${colors.divider}`,
+            }} />
+            <section className="home-section">
```
yangshun: Will refactor the examples section in a later PR. Please bear with the hardcoding of class names for now.

src/types.js
```diff
@@ -23,6 +23,7 @@ export type Node = {
     next?: string,
     prev?: string,
     title: string,
+    order?: number,
```
yangshun: Not sure whether this is necessary.

content/marketing/component-based.md
```diff
@@ -0,0 +1,8 @@
+---
+title: Component-Based
```
yangshun: I considered putting the title inside the content instead but I think putting it in the frontmatter makes more sense so we leave the formatting of the title outside this file.

src/pages/index.js
bvaughn: Yes

src/types.js
bvaughn: Not really, since the `order` frontmatter field is only used in the GraphQL query (and not in any Flow-covered code).

yangshun:
@bvaughn Have fixed according to your comments. Do have a look again 😄

content/marketing/component-based.md
bvaughn: Sounds reasonable to me.

content/marketing/component-based.md
bvaughn: I think order works for this. Kind of like flexbox order.

gatsby-node.js
bvaughn: I dig it.

src/pages/index.js
```diff
@@ -6,7 +6,7 @@
```
bvaughn: This works for the `/` route but not for `/index.html`

We need to add an explicit redirect to `gatsby-node.js` to handle this:

```js
  // Without this, /index.html won't redirect to root.
  createRedirect({
    fromPath: '/index.html',
    redirectInBrowser: true,
    toPath: '/',
  });
```

src/pages/index.js
```diff
@@ -135,10 +136,158 @@ class Home extends Component {
...
+                border: 'none',
+                borderBottom: `1 solid ${colors.divider}`,
+              }}
+            />
+            <section css={sectionStyles}>
+              <div id="examples">
+                <div className="example">
+                  <h3>A Simple Component</h3>
+                  <p>
+                    React components implement a `render()` method that takes
```
bvaughn: The content for these sections also needs to come from markdown files (so it can be localized).

src/pages/index.js
```diff
@@ -135,10 +136,158 @@ class Home extends Component {
...
+                    XML-like syntax called JSX. Input data that is passed into
+                    the component can be accessed by `render()` via
+                    `this.props`.
+                  </p>
+                  <p>
+                    <strong>
+                      JSX is optional and not required to use React.
+                    </strong>{' '}
+                    Try the{' '}
+                    <a href="http://babeljs.io/repl#?babili=false&browsers=&build=&builtIns=false&code_lz=MYGwhgzhAEASCmIQHsCy8pgOb2vAHgC7wB2AJjAErxjCEB0AwsgLYAOyJph0A3gFABIAE6ky8YQAoAlHyEj4hAK7CS0ADxkAlgDcAfAiTI-hABZaI9NsORtLJMC3gBfdQHpt-gNxDn_P_zUtIQAIgDyqPSi5BKS6oYo6Jg40A5OALwARCHwOlokmdBuegA00CzISiSEAHLI4tJeQA&debug=false&circleciRepo=&evaluate=false&lineWrap=false&presets=react&prettier=true&targets=&version=6.26.0">
```
bvaughn: You know what might be cool? To auto-generate the URL for this from the actual code example (so the two will always stay in-sync). Babel's REPL just uses the [`lz-string` module](https://www.npmjs.com/package/lz-string) to [serialize and deserialize code](https://github.com/babel/website/blob/f34ae222d3f7edf70711e1c849f13418da95b4cb/js/repl/UriUtils.js#L22-L33) so we could copy it. 😁  It isn't likely to change because it's important to Babel to maintain backwards compatibility.

src/pages/index.js
yangshun: Good catch, thanks for this!

yangshun:
Thanks for the review @bvaughn! I fixed the `index.html` redirect issue. 

Regarding the content for the examples, I intend to address and fix them in a follow up PR so that this PR does not get too big. WDYT?

bvaughn:
> As for the refactoring of the marketing content, I think they are good to be merged. WDYT?

I think we should do it all as one PR, to be honest. We have localization/translation efforts going on right now and I don't want to temporarily pull a bunch of home page content out of the pipeline and cause confusion.

Also cc @ericnakagawa about this change.

yangshun:
I see. Ok I'll continue on this branch then 😄

yangshun:
@bvaughn I extracted out the examples content into markdown files following the same approach for the marketing content. I intend to also move the code out into separate files but I'm not too sure how the `mountCodeExample` hacks can be improved on. Could you provide more guidance regarding how to fix the TODOs tagged to `renderExamplePlaceholder` and `mountCodeExample`?

As far as the localization/translation efforts goes, this PR should be ok given that I have extracted out both the marketing and examples content into markdown files.

src/pages/index.js
yangshun: If this link will be part of the markdown, how do we generate the URL for this? Do we generate after render and modify the `href` in the markup?

yangshun:
> You know what might be cool? To auto-generate the URL for this from the actual code example (so the two will always stay in-sync). Babel's REPL just uses the lz-string module to serialize and deserialize code so we could copy it. 😁 It isn't likely to change because it's important to Babel to maintain backwards compatibility.

If this link will be part of the markdown, how do we generate the URL for this? Do we generate after render and modify the href in the markup?

bvaughn:
I dig the general direction this PR is heading 😁  Thanks for driving this!

Could we move the new directories `content/marketing` and `content/examples` into a shared parent folder, `content/home` (just to more clearly organize the content). Particularly given other efforts like #138 and #162, I'm concerned "examples" might be confusing and "marketing" is a bit ambiguous.

> I intend to also move the code out into separate files but I'm not too sure how the mountCodeExample hacks can be improved on. Could you provide more guidance regarding how to fix the TODOs tagged to renderExamplePlaceholder and mountCodeExample?

Maybe we should ignore the TODO about moving the hard-coded examples, for now, at least pending the outcome of #138 and #162.

What I was thinking was that we could move the example JS snippets near the markdown, but if we did that we wouldn't be able to import/require them in the template without either (a) passing them through via markdown (kind of hacky) or (b) overriding the Webpack config to allow it to resolve paths within `content` (since it's outside of source) and both seem kind of blah.

Maybe for now, we should just rename the code snippet constant names to better match the names of their corresponding markdown files.

> If this link will be part of the markdown, how do we generate the URL for this? Do we generate after render and modify the href in the markup?

It's fine to do this as a follow-up. Actually I'd be happy to do it b'c it sounds fun. 😄  The idea though is that we could come up with a new syntax for linking to an example file and then we could maybe write a custom Remark transform plugin that generates the _actual_ link (with lz compression) from it during build time so they were guaranteed to stay in sync.

bvaughn:
Friendly follow up with @yangshun 😄 

Also cc @ericnakagawa

bvaughn:
One last friendly follow up 😄

If I don't hear back in a day or so, I'm going to take over this PR and implement my own requested changes.

yangshun:
@bvaughn Very sorry for the delay! This PR slipped my mind. I shifted the files into `/home/marketing/` and `/home/examples/`. Is there anything else you want me to do in this PR?

bvaughn:
No worries at all. Let me take a look.

bvaughn:
Thank you!

I'm added to this PR by moving the example JavaScript code into `/home/content` as well (see f5aaf4c) and tying it together with a new local plugin, `gatsby-transformer-home-example-code`.

PR #151, opened 2017-10-12
fix(algolia): make logo show up on FF

### LINKED ISSUE #130: Ensure Algolia attribution is present on search dropdown
Hello!

Just wanted to pop in to mention that the Algolia logo in the search dropdown doesn't appear to be visible in Firefox (oddly enough). Screenshot time!

Firefox 56 (OS X): 
image

Check that bottom right corner...

Chrome 61 (OS X):
image

I see that our very own @Haroenv touched this topic in https://github.com/facebook/react/pull/11032. Could be a good place to start looking?

For the sake of completeness, I've also looped in @vjeux via email, which isn't a GitHub Issue but oh well! 😉

Haroenv:
Background images in svg need to have the # encoded to show up if the image type is svg+xml;utf8

fixes #130


Haroenv:
It works in FF for me now

PR #154, opened 2017-10-13
Improve tutorial step three

fernahh:
There is a issue about a guy that has excluded the `src/` folder. To avoid future issues, I have improved step three with:

- `inside` instead of `in`
- added a parentheses with `don't exclude the folder, just the files`

resolves https://github.com/reactjs/reactjs.org/issues/71

### DIFF
--- a/content/tutorial/tutorial.md
+++ b/content/tutorial/tutorial.md
@@ -54,7 +54,7 @@ If you want to do it, here are the steps to follow:
 
 1. Make sure you have a recent version of [Node.js](https://nodejs.org/en/) installed.
 2. Follow the [installation instructions](/docs/installation.html#creating-a-new-application) to create a new project.
-3. Delete all files inside the `src/` folder of the new project (don't exclude the folder, just the files).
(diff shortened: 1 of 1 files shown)


content/tutorial/tutorial.md
```diff
@@ -54,7 +54,7 @@ If you want to do it, here are the steps to follow:
 
 1. Make sure you have a recent version of [Node.js](https://nodejs.org/en/) installed.
 2. Follow the [installation instructions](/docs/installation.html#creating-a-new-application) to create a new project.
-3. Delete all files in the `src/` folder of the new project.
+3. Delete all files inside the `src/` folder of the new project (don't exclude the folder, just the files).
```
bvaughn: > (don't exclude the folder, just the files).

I'm not sure the wording for this is clear. How do you feel about:

> don't delete the folder, just its contents (eg `rm -f src/*`)

fernahh: nice, I'm update this.

Issue #24, opened 2017-10-06
Add Flow types to website source


bvaughn:
Replace `PropTypes` with Flow types for our components in `www` (and make sure `yarn flow` checks these types to protect against regressions).

I got the ball rolling with commit 1de4c66

[This Twitter thread](https://twitter.com/brian_d_vaughn/status/916061335036739584) also has some interesting discussion for auto-generating Flow types for Gatsby's graphql queries.

hd719:
Hey sorry @bvaughn is this now closed?

hd719:
Hey @bvaughn yes I am still kind of new to open source and still learning Flow Types, but yes I would like to help out on this particular issue.

bvaughn:
Okay! Great.

This is a pretty big issue, so my advice is- let's do it a couple files at a time. That way it's easier to review and we don't risk conflicting with every other PR.

hd719:
Sounds good!

tricinel:
I can help as well. Going through a couple of files at time seems like a good idea.

bvaughn:
Okay. I'll mark this issue in-progress.

Communicate here to avoid stepping on each other's toes. 😄

tricinel:
Awesome. I've already opened a PR for it (but forgot to mention the issue number on the commit): https://github.com/reactjs/reactjs.org/pull/156. Sorry about that. Perhaps if it checks out, then it can be merged so that we keep on top of it and avoid conflicts as much as possible?

I'd like to continue working on it though. So I can continue on it next week if that's ok. I'll work through the `src/components` files first.

Do you think it's worth doing the utils as well? Adding flow typechecking there as well?

bvaughn:
> Do you think it's worth doing the utils as well? Adding flow typechecking there as well?

Yeah 😄

wuweiweiwu:
@bvaughn Is this still in progress? I'd love to work on it

bvaughn:
Please feel free to contribute anywhere you see missing types or types that could be improved, @wuweiweiwu. Probably worth checking through open pull requests first to make sure something hasn't already been done though! We usually let first-in win.

wuweiweiwu:
@bvaughn Sounds good!

kanitsharma:
Hey @bvaughn, I'd like to work on it too, is this still in progress?

bvaughn:
I don't know. Looks like there's another open PR (#449) that got dropped because I just haven't had time to review this repo lately :( SOrry

[...]

PR #156, opened 2017-10-13
Switch to flow types; wip related to #24

tricinel:
I've jump-started this and switch to Flow types on a couple of files (I admit it was the easier ones). But will continue to work on these.

### DIFF
--- a/src/components/MarkdownPage/MarkdownPage.js
+++ b/src/components/MarkdownPage/MarkdownPage.js
@@ -19,6 +19,18 @@ import toCommaSeparatedList from 'utils/toCommaSeparatedList';
 import {sharedStyles} from 'theme';
 import createOgUrl from 'utils/createOgUrl';
 
+type Props = {
+  authors: Array<string>,
+  createLink: Function,
+  date?: string,
+  enableScrollSync?: boolean,
+  ogDescription: string,
+  location: Object,
+  markdownRemark: Object,
+  sectionList: Array<Object>,
+  titlePostfix: string,
+};
+
 const MarkdownPage = ({
   authors = [],
   createLink,
@@ -29,17 +41,7 @@ const MarkdownPage = ({
   markdownRemark,
   sectionList,
   titlePostfix = '',
-}: {
-  authors: Array<string>,
-  createLink: Function,
-  date: string,
-  enableScrollSync: boolean,
-  ogDescription: string,
-  location: Object,
-  markdownRemark: Object,
-  sectionList: Array<Object>,
-  titlePostfix: string,
-}) => {
+}: Props) => {
   const hasAuthors = authors.length > 0;
   const titlePrefix = markdownRemark.frontmatter.title || '';
 
--- a/src/components/MarkdownPage/MarkdownPage.js
+++ b/src/components/MarkdownPage/MarkdownPage.js
@@ -2,6 +2,7 @@
  * Copyright (c) 2013-present, Facebook, Inc.
  *
  * @emails react-core
+ * @flow
  */
 
 'use strict';
@@ -10,7 +11,6 @@ import Container from 'components/Container';
 import Flex from 'components/Flex';
 import MarkdownHeader from 'components/MarkdownHeader';
 import NavigationFooter from 'templates/components/NavigationFooter';
-import PropTypes from 'prop-types';
 import React from 'react';
 import StickyResponsiveSidebar from 'components/StickyResponsiveSidebar';
 import TitleAndMetaTags from 'components/TitleAndMetaTags';
@@ -20,7 +20,7 @@ import {sharedStyles} from 'theme';
 import createOgUrl from 'utils/createOgUrl';
 
 const MarkdownPage = ({
-  authors,
+  authors = [],
   createLink,
   date,
   enableScrollSync,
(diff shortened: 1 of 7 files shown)


src/components/MarkdownPage/MarkdownPage.js
```diff
@@ -29,6 +29,16 @@ const MarkdownPage = ({
   markdownRemark,
   sectionList,
   titlePostfix = '',
+}: {
+  authors: Array<string>,
+  createLink: Function,
+  date: string,
+  enableScrollSync: boolean,
```
bvaughn: `date` and `enableScrollSync` weren't marked as required before, and they are ommitted in some cases, so we should make them optional here..

src/components/TitleAndMetaTags/TitleAndMetaTags.js
```diff
@@ -11,7 +11,15 @@ import React from 'react';
...
-const TitleAndMetaTags = ({title, ogDescription, ogUrl}) => {
+const TitleAndMetaTags = ({
+  title,
+  ogDescription,
+  ogUrl,
+}: {
+  title: string,
+  ogDescription: string,
+  ogUrl: string,
+}) => {
```
bvaughn: This may be a subjective nit, but prefer types to be declared separately (for readability), eg:
```js
type Props = {
  title: string,
  ogDescription: string,
  ogUrl: string,
};

const TitleAndMetaTags = ({ title, ogDescription, ogUrl }: Props) => {
```

src/components/MarkdownPage/MarkdownPage.js
tricinel: 👍

src/components/TitleAndMetaTags/TitleAndMetaTags.js
tricinel: No problem with me on that :)

src/components/MarkdownPage/MarkdownPage.js
```diff
@@ -19,8 +19,20 @@ import toCommaSeparatedList from 'utils/toCommaSeparatedList';
 import {sharedStyles} from 'theme';
 import createOgUrl from 'utils/createOgUrl';
 
+type Props = {
+  authors: Array<string>,
+  createLink: Function,
+  date?: string,
+  enableScrollSync?: boolean,
+  ogDescription: string,
+  location: Object,
```
bvaughn: Should be the `Location` type: https://github.com/facebook/flow/blob/v0.56.0/lib/bom.js#L289

src/components/MarkdownPage/MarkdownPage.js
```diff
@@ -19,8 +19,20 @@ import toCommaSeparatedList from 'utils/toCommaSeparatedList';
...
+  markdownRemark: Object,
```
bvaughn: Can we make this a better type? Like `Node` from the [shared `types`](https://github.com/reactjs/reactjs.org/blob/master/src/types.js)?

src/components/MarkdownPage/MarkdownPage.js
```diff
@@ -19,8 +19,20 @@ import toCommaSeparatedList from 'utils/toCommaSeparatedList';
 import {sharedStyles} from 'theme';
 import createOgUrl from 'utils/createOgUrl';
 
+type Props = {
+  authors: Array<string>,
+  createLink: Function,
```
bvaughn: Would be good to add a TODO here to better type this (once we Flow-type `createLink` and the `Section` component)

src/components/MarkdownPage/MarkdownPage.js
```diff
@@ -19,8 +19,20 @@ import toCommaSeparatedList from 'utils/toCommaSeparatedList';
...
+  date?: string,
+  enableScrollSync?: boolean,
+  ogDescription: string,
+  location: Object,
+  markdownRemark: Object,
+  sectionList: Array<Object>,
```
bvaughn: Would be good to add a TODO here to better type this once we Flow-type the `Section` component

src/components/MarkdownPage/MarkdownPage.js
tricinel: I'm at a bit of a loss here, because all your comments relate to the same MarkdownPage component, and they seem to be a bit conflicting...Which of the changes should I do? :)

bvaughn: I left comments under the lines that should be changed. (I think that's how GitHub's PR tool is typically used.)

screen shot 2017-10-18 at 8 43 48 am

So in this case, the comment is referring to the `markdownRemark` type. It's just a plain `Object` but it could be a more strict type- specifically one I've already created in [this file named `types`](https://github.com/reactjs/reactjs.org/blob/master/src/types.js).

bvaughn: I just edited my previous comments to hopefully make what I'm asking for clearer. Please let me know if there's any confusion.

src/components/MarkdownPage/MarkdownPage.js
tricinel: Sorry, I've been out of touch for the past couple of days. Yes, clearer..I think I was confused because it was highlighting three lines (and they were mostly the same three lines)...Meh, anyway. Pushed. Thanks for that Location type. I didn't know flow had one already. Sweet.

tricinel:
Sweet, I'm continuing on this then, I should have most of them covered by mid of next week.

Thanks for the help!

PR #157, opened 2017-10-13
Nested React Components

DotFreelance:
Added information and an example of validating nested React components.

### DIFF
--- a/content/docs/typechecking-with-proptypes.md
+++ b/content/docs/typechecking-with-proptypes.md
@@ -176,15 +176,22 @@ If you want to validate nested elements, such as the children of your component,
 
 ```javascript
 import PropTypes from 'prop-types';
+import SubComponent from './my-custom-components/SubComponent'
 
 class MyComponent extends React.Component {
   render() {
     // This must be exactly one element or it will warn.
-    const children = this.props.children;
+    const child = this.props.children;
     return (
+      // Condition 1
       <div>
-        {children}
+        {child}
       </div>
+      // Condition 2
+      <SubComponent 
+        containerElement={child.props.containerElement}
+        primaryText={child.props.label}
+      />
     );
   }
 }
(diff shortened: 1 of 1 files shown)


bvaughn:
I'm not sure it seems like a good idea for a parent component to validate its children's props like this. Could you provide some examples for why you might want to do this?

DotFreelance:
I would say in the case where any parent cares about the shape of its children when remapping them. In one specific case, the Component accepts all children as is, but may remap those children to a different SubComponent under certain conditions. i.e. a component like a responsive menu that accepts children as is for large-scale, but reduces those children to simpler SubComponent types on small-scale. Does that make sense?

bvaughn:
Thanks for elaborating. I understand. 😄

Maybe we could update the example code to actually show a meaningful example of this? (Currently neither `containerElement` nor `label` props are actually used in the `render` method.)

DotFreelance:
Okay, an update a little more fleshed out. Hopefully that looks good!

content/docs/typechecking-with-proptypes.md
```diff
@@ -169,3 +169,59 @@ ReactDOM.render(
...
+        {child}
+      </div>
+      // Condition 2
+      <SubComponent 
+        containerElement={child.props.containerElement}
+        primaryText={child.props.label}
+      />
+    );
+  }
+}
```
bvaughn: This isn't valid JSX ^ Maybe you could change it to something like:

```jsx
class MyComponent extends React.Component {
  render() {
    // This must be exactly one element or it will warn.
    const child = this.props.children;

    if (someConditional) {
      return (
        <div>
          {child}
        </div>
    	);
    } else {
      return (
        <SubComponent 
          containerElement={child.props.containerElement}
          primaryText={child.props.label}
        />
      );
    }
  }
}
```

BartoszKlonowski:
Hello @DotFreelance! Looks like this PR adds changes to already outdated part of documentation.
Let me close this, but please feel free to create another PR for any open issues.
Thank you!

PR #158, opened 2017-10-13
Mention ad blockers when Babel is not defined

### LINKED ISSUE #3: Gracefully recover when ad block prevents Babel from loading

renanpvaz:
Closes #3.

### DIFF
--- a/src/components/CodeEditor/CodeEditor.js
+++ b/src/components/CodeEditor/CodeEditor.js
@@ -177,6 +177,7 @@ class CodeEditor extends Component {
                     padding: 10,
                   }}>
                   {error.message}
+                  {!window.Babel && ' (try checking your ad blocker if you have one).'}
                 </pre>
               </div>
             )}


src/components/CodeEditor/CodeEditor.js
```diff
@@ -177,6 +177,7 @@ class CodeEditor extends Component {
                     padding: 10,
                   }}>
                   {error.message}
+                  {!window.Babel && ' (try checking your ad blocker if you have one).'}
```
renanpvaz: Should it also ask the user to whitelist the website?

renanpvaz:
Sure, so the special message should be shown instead of "Babel is not defined", right?

bvaughn:
Yeah, that's what I was thinking- we could check this state-flag and render some custom text _instead of_ `error.message`

Issue #164, opened 2017-10-16
Improve focus style for search input and nav sidebar headers


damaera:
this look not good for me,
![image](https://user-images.githubusercontent.com/5365957/31600067-3d46a84a-b27f-11e7-9a34-4807c689d0c2.png)

IMO, it need more padding, or maybe when focus, no need to show the outline

damaera:
in tree menu, this happens too.
![image](https://user-images.githubusercontent.com/5365957/31600503-f13dd96c-b280-11e7-814c-b0de8909c914.png)

bvaughn:
The focus outline is important for usability. We can't remove it unless we provide some other, equally noticeable focus style (which we just haven't designed yet).

I agree that the outline is a bit snug for the search input, but I don't think it's all that bad.

The awkward nav sidebar focus sizing is worse though. It's caused by the recent change to wrap the styled header `<div>` in a clickable `<button>`. The div wasn't written with the assumption that it would be wrapped, and the button includes the vertical padding + margin of the div.

I've fixed this (in commit 71bc3d9) to move the margin to the button itself so the alignment is centered.

bvaughn:
I'm going to repurpose this issue to be about creating an alternate focus style that can replace the focus outline.

bvaughn:
Check out PR #167 for a potential improvement.

bvaughn:
Closed via #167 and commit 71bc3d9.

el1f:
Honestly I think that the sidenav would use the same love that the topnav received, with something like a dedicated focus style. Can I work on that?

bvaughn:
Please feel free to submit a PR, @el1f

PR #167, opened 2017-10-16
Header focus style improvements

bvaughn:
Relates to issue #164

People don't like default browser focus outlines a lot of the time, because they're ugly. But some sort of focus style is important for accessibility. So this PR takes a stab at replacing the default header focus style with a custom one.

I think most links and buttons (in the body) are fine with default styles, but maybe we could improve the look and feel of the header?

### Before
header-hover-focus-before

### After
header-hover-focus-after

cc @joecritch, @damaera

### DIFF
--- a/src/components/LayoutHeader/Header.js
+++ b/src/components/LayoutHeader/Header.js
@@ -47,6 +47,12 @@ const Header = ({location}) => (
             marginRight: 10,
             height: '100%',
             alignItems: 'center',
+            color: colors.brand,
+
+            ':focus': {
+              outline: 0,
+              color: colors.white,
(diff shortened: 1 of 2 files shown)


bvaughn:
I chatted about this briefly with Jesse from the accessibility team here and my take-away is that this is probably an acceptable change, given its scope.

She also pointed me toward https://github.com/WICG/focus-ring (demo [here](https://wicg.github.io/focus-ring/demo/)) which is a pretty cool proposal. We could always look at doing a more comprehensive change using it if we decide to.

In the meanwhile, I feel good about merging this as-is.

PR #169, opened 2017-10-17
Fix missing `extends` for custom elements

yuheiy:


bvaughn:
Cool, thanks 👍 

CI is failing for all branches at the moment, so it's unrelated.

PR #175, opened 2017-10-17
Add more details to the `Error boundaries` doc

thomasyimgit:
This is a document improvement concerning the issue here. https://github.com/reactjs/reactjs.org/issues/145

What I've done:
1. add a explicit note about which of the scenarios that error boundaries don't work with
2. add a short code sample about how to use try/catch in event handlers

Some of my thoughts:

I added the code sample in the `try/catch` section and changed its title from `Why Not Use try/catch?` to `How About try/catch?`. Because I thinks that piece of code(event handler) still is using `try/catch` as an approach to handling errors. So it doesn't seem good if we use the original title.

Wording is truly hard! If you find anything inappropriate, please leave your comment below. Thanks!

### DIFF
--- a/content/docs/error-boundaries.md
+++ b/content/docs/error-boundaries.md
@@ -13,6 +13,13 @@ A JavaScript error in a part of the UI shouldn’t break the whole app. To solve
 
 Error boundaries are React components that **catch JavaScript errors anywhere in their child component tree, log those errors, and display a fallback UI** instead of the component tree that crashed. Error boundaries catch errors during rendering, in lifecycle methods, and in constructors of the whole tree below them.
 
+> Note
+> 
+> Error boundaries do **NOT** work in the following scenarios:
+> * Event Handlers (See below ["How About Try / Catch"](#how-about-trycatch) for details about dealing with errors in event handlers)
+> * Asynchronous code (e.g. `setTimeout` or `requestAnimationFrame` callbacks)
+> * Server Side Rendering
+
 A class component becomes an error boundary if it defines a new lifecycle method called `componentDidCatch(error, info)`:
 
 ```js{7-12,15-18}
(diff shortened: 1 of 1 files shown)


content/docs/error-boundaries.md
```diff
@@ -13,6 +13,13 @@ A JavaScript error in a part of the UI shouldn’t break the whole app. To solve
 
 Error boundaries are React components that **catch JavaScript errors anywhere in their child component tree, log those errors, and display a fallback UI** instead of the component tree that crashed. Error boundaries catch errors during rendering, in lifecycle methods, and in constructors of the whole tree below them.
 
+> Note
+> 
+> Error boundaries do **NOT** work in the following scenarios:
+> * Event Handlers (See below ["How About Try / Catch"](#how-about-trycatch) for details about dealing with errors in event handlers)
+> * Asynchronous code
```
bvaughn: Could you elaborate on what is meant by this? I worry a reader could interpret this to mean that error boundaries won't work for async React.

Maybe just something like "(eg `setTimeout` or `requestAnimationFrame` callbacks)"

content/docs/error-boundaries.md
```diff
@@ -129,6 +136,34 @@ However, React components are declarative and specify *what* should be rendered:
 
 Error boundaries preserve the declarative nature of React, and behave as you would expect. For example, even if an error occurs in a `componentDidUpdate` hook caused by a `setState` somewhere deep in the tree, it will still correctly propagate to the closest error boundary.
 
+However, a typical event handler inside a React component could leverage `try` / `catch` to deal with errors occured during event handling.
```
bvaughn: How about a slight wording tweak?

> However, an event handler inside of a React component could leverage `try` / `catch` to deal with errors that occur during event handling.

content/docs/error-boundaries.md
thomasyimgit: I didn't dive into the source code of componentDidCatch and am not sure about this. The original post didn't say anything about async code with error boundaries. Thanks for pointing out.

content/docs/error-boundaries.md
thomasyimgit: I agree with this. The word `typical` is somewhat confusing.

thomasyimgit:
@bvaughn I've refactored the wording according to your comment. Thanks for pointing out. Much appreciated. 😄

PR #176, opened 2017-10-18
Tweak glossary

gaearon:
Just rewriting a little bit more to my liking.

### DIFF
--- a/content/docs/reference-glossary.md
+++ b/content/docs/reference-glossary.md
@@ -122,7 +122,7 @@ class Welcome extends React.Component {
 
 A component needs `state` when some data associated with it changes over time. For example, a `Checkbox` component might need `isChecked` in its state, and a `NewsFeed` component might want to keep track of `fetchedPosts` in its state.
 
-The most important difference between `state` and `props` is that `props` are passed from a parent component, but `state` is managed by the component itself. A component cannot change its `props`, but it can change its `state`.
+The most important difference between `state` and `props` is that `props` are passed from a parent component, but `state` is managed by the component itself. A component cannot change its `props`, but it can change its `state`. To do so, it must call `this.setState()`. Only components defined as classes can have state.
 
 For each particular piece of changing data, there should be just one component that "owns" it in its state. Don't try to synchronize states of two different components. Instead, [lift it up](/docs/lifting-state-up.html) to their closest shared ancestor, and pass it down as props to both of them.
 
--- a/content/docs/reference-glossary.md
+++ b/content/docs/reference-glossary.md
@@ -7,38 +7,58 @@ permalink: docs/glossary.html
 
 ---
 
-## Single Page Application
-A single page application is an application that loads a single HTML page and all the necessary assets (CSS, Images & JavaScript) required for the application to run. Any interactions with the page or subsequent pages do not require a round trip to the server which means the page is not reloaded.
+## Single-page Application
 
+A single-page application is an application that loads a single HTML page and all the necessary assets (such as JavaScript and CSS) required for the application to run. Any interactions with the page or subsequent pages do not require a round trip to the server which means the page is not reloaded.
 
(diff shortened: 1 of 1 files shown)


content/docs/reference-glossary.md
```diff
@@ -7,38 +7,58 @@ permalink: docs/glossary.html
...
+## Single-page Application
 
+A single-page application is an application that loads a single HTML page and all the necessary assets (such as JavaScript and CSS) required for the application to run. Any interactions with the page or subsequent pages do not require a round trip to the server which means the page is not reloaded.
 
-## ES6/ES2015/ES7/ES2016/ES8/ES2017
- These acronyms all refer to the most recent versions of the ECMAScript Language Specification standard, which the JavaScript language is an implementation of. It includes many additions to the previous versions such as: arrow functions, classes, template literals, `let` and `const` statements. You can learn more about specific versions [here](https://en.wikipedia.org/wiki/ECMAScript#Versions).
+Though you may build a single-page application in React, it is not a requirement. React can also be used for enhancing small parts of existing websites with additional interactivity. Code written in React can coexist peacefully with markup rendered on the server by something like PHP, or with other client-side libraries. In fact, it's exactly how React is being used at Facebook.
```
bvaughn: nit suggestion:

> In fact, this is exactly how React is being used at Facebook.

content/docs/reference-glossary.md
```diff
@@ -7,38 +7,58 @@ permalink: docs/glossary.html
...
+const name = 'Clementine';
+ReactDOM.render(
+  <h1 className="hello">My name is {name}!</h1>,
+  document.getElementById('root')
+);
+```  
 
 ## [Elements](/docs/rendering-elements.html)
+
 React elements are the building blocks of React applications. One might confuse elements with a more widely known concept of "components". Elements are what components are "made of". An element describes what you want to see on the screen. React elements are immutable.
```
bvaughn: I still think the wording here is a bit vague:
> Elements are what components are "made of".

I wonder if we could improve it by changing it to something like:
> Elements are a template for creating components.

content/docs/reference-glossary.md
```diff
@@ -59,43 +79,88 @@ class Welcome extends React.Component {
...
+
+For components defined as classes, use `this.props.children`:
+
+```js
+class Welcome extends React.Component {
+  render() {
+    return <p>{this.props.children}</p>;
+  }
+}
+```
```
bvaughn: Nice! ^

content/docs/reference-glossary.md
```diff
@@ -59,43 +79,88 @@ class Welcome extends React.Component {
...
 
  ## [Controlled](/docs/forms.html#controlled-components) vs. [Uncontrolled Components](/docs/uncontrolled-components.html)
+
 React has two different approaches to dealing with form inputs. 
 
 An input form element whose value is controlled by React is called a *controlled component*. When a user enters data into a controlled component a change event handler is triggered and your code decides whether the input is valid (by re-rendering with the updated value). If you do not re-render then the form element will remain unchanged.
 
-An *uncontrolled component* works like form elements do outside of React. When a user inputs data into a form field (an input box, dropdown, etc) the updated information is reflected without React needing to do anything.
+An *uncontrolled component* works like form elements do outside of React. When a user inputs data into a form field (an input box, dropdown, etc) the updated information is reflected without React needing to do anything, but you can't force the field to have a certain value.
```
bvaughn: nit suggestion:

> When a user inputs data into a form field (an input box, dropdown, etc) the updated information is reflected without React needing to do anything. This also means that you can't force the field to have a certain value.

content/docs/reference-glossary.md
gaearon: I just removed this sentence since it's duplicating the same thought as in first sentence and then after the snippet.

PR #177, opened 2017-10-18
Add etymological background to "props".

kevmo:
I have heard/seen multiple people ask what "props" stands for, and this is the most logical place to explain it in the documentation.

### DIFF
--- a/content/docs/components-and-props.md
+++ b/content/docs/components-and-props.md
@@ -30,7 +30,7 @@ function Welcome(props) {
 }
 ```
 
-This function is a valid React component because it accepts a single "props" object argument with data and returns a React element. We call such components "functional" because they are literally JavaScript functions.
+This function is a valid React component because it accepts a single "props" (which stands for properties) object argument with data and returns a React element. We call such components "functional" because they are literally JavaScript functions.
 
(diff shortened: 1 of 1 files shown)


gaearon:
Thanks!

content/docs/components-and-props.md
```diff
@@ -30,7 +30,7 @@ function Welcome(props) {
 }
 ```
 
-This function is a valid React component because it accepts a single "props" object argument with data and returns a React element. We call such components "functional" because they are literally JavaScript functions.
+This function is a valid React component because it accepts a single "props" (which stands for properties) object argument with data and returns a React element. We call such components "functional" because they are literally JavaScript functions.
```
bvaughn: Disclaimer: My knowledge of grammar is mediocre.

If the meaning of "props" is unclear then I like the idea of explaining it to people. This sentence seems to read kind of awkwardly now though, given the placement of parens.

I wonder if this might be clearer?

> This function is a valid React component because it accepts a single "props" parameter (short for "properties") with data and returns a React element.

bvaughn: Whoops. Looks like this was merged while I was typing 😅

Issue #171, opened 2017-10-17
Explore adding top-level "API" link to header nav


bvaughn:
[This discussion thread on Twitter](https://twitter.com/LedZ3pp/status/919923967787458565) provided feedback that the reference section of the docs may be too buried given the new website's accordion sidebar.

@gaearon proposed we could do something similar to what [Jest](http://facebook.github.io/jest/) and add an "API" link to the header that directly opens the first page of the reference section (eg [React Top-Level API](https://reactjs.org/docs/react-api.html)).

Unfortunately this is a little more complicated than it would seem at a glance because of:
* [How we determine which header link gets the "active" style](https://github.com/reactjs/reactjs.org/blob/e826e307e0c129a478546c79030518a9ee311389/src/components/LayoutHeader/Header.js#L117). (Jest side-steps this issue by not applying "active" header link styles.)
* A symmetry we try to maintain between the [location of markdown content](https://github.com/reactjs/reactjs.org/tree/e826e307e0c129a478546c79030518a9ee311389/content) (eg content/blog, content/docs) and the top-level nav structure (eg "Blog", "Docs")

We _could_ move things around so that reference links have a unique URL prefix, but we'd need to leave backwards-compatible redirects in place so that no bookmarked links were broken.

Unfortunately we can't rely on anything dynamic (like markdown frontmatter tags) for this because the header and footer are part of the "static" Gatsby layout. (Technically we could move them into the templates but this doesn't seem ideal.)

KyleAMathews:
You could just make the active style programmatic right? Just say any reference doc pages make the API header link active.

Perhaps removing the accordion is an option too?

bvaughn:
> You could just make the active style programmatic right? Just say any reference doc pages make the API header link active.

We _could_ but that's what this section from the description was referencing:

> Unfortunately we can't rely on anything dynamic (like markdown frontmatter tags) to help with the "active" link styling because the header and footer are part of the "static" Gatsby layout. (Technically we could move them into the templates but this doesn't seem ideal.)

Short of moving the header out of the layout _or_ hard-coding a bunch of URLs in it- how would you propose we do this?

KyleAMathews:
>hard-coding a bunch of URLs

done! https://github.com/reactjs/reactjs.org/blob/master/content/docs/nav.yml :-)

bvaughn:
Reading between the lines, what you're suggesting seems a bit janky to me.

[...]

PR #178, opened 2017-10-18
Add top level API link to header nav

ismamz:
In reference to: #171 

captura de pantalla 2017-10-18 a la s 16 26 11

- Reference group in Docs sidebar was removed
- Reference content was moved to `content/api`
- Added "API" link to header nav
- `content/api` pages, redirect from `content/docs`

This is how it looks on mobile:

@bvaughn It's everything ok?

### DIFF
--- a/content/api/dom-elements.md
+++ b/content/api/dom-elements.md
@@ -3,8 +3,9 @@ id: dom-elements
 title: DOM Elements
 layout: docs
 category: Reference
-permalink: docs/dom-elements.html
+permalink: api/dom-elements.html
 redirect_from:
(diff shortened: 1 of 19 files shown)


bvaughn:
Hi @ismamz. Whether or not we add the top-level link is still being discussed on #171 😄

Let's see what the consensus is, and if we decide to add it then I'll review this PR!

ismamz:
Ah ok! I understand. I assumed that there was consensus. You can see how it would look! 😄

bvaughn:
No worries, and much appreciated 😄

Issue #14, opened 2017-10-06
Website displays empty page when cookies are disabled in Firefox


bvaughn:
When using a Firefox with "[Cookie Whitelist](https://addons.mozilla.org/en-US/firefox/addon/cookie-whitelist-with-buttons/)" plugin (and enabled, preventing cookie storage), loading https://reactjs.org/ flashes the page's content then removes all content in the body, producing an empty white page with the following content:

```
<body>
  <div id="___gatsby">
  </div>
</body>
```

When loading the page the content initially loads, flashes, and then a security exception is thrown in the console and all content is removed from the page:

```
DOMException [SecurityError: "The operation is insecure."
code: 18
nsresult: 0x80530012
location: https://reactjs.org/commons-2034d1df9ebd047740fd.js:4]
```

If the "Cookie Whitelist" is set to allow temporary cookies, the page loads normally. 

Copied from facebook/react/issues/11015

tylercrosse:
I experienced a similar issue of the page flashing rendered content and then going blank. The chrome extension Fair AdBlockr was responsible and whitelisting the site solved the problem. (summarized from original description filed on [facebook/react#11015](https://github.com/facebook/react/issues/11015) )

On the blank page the react dev tools only show these 4 components & their children:

```html
<t code ="class HelloMessage extends React.Component { ren...">...</t>
<t code ="class Timer extends React.Component { constructo..">...</t>
<t code ="class TodoApp extends React.Component { construc..">...</t>
<t code ="class MarkdownEditor extends React.Component { c..">...</t>
```

I also get the following errors logged to the console

```
react-dom.production.min.js:187 ReferenceError: docsearch is not defined
    at t.componentDidMount (component---src-layouts-index-js-673385803e28183fd482.js:2)
    at commitLifeCycles (react-dom.production.min.js:169)
    at n (react-dom.production.min.js:180)
    at u (react-dom.production.min.js:183)
    at c (react-dom.production.min.js:184)
    at h (react-dom.production.min.js:188)
    at m (react-dom.production.min.js:187)
    at Object.updateContainer (react-dom.production.min.js:248)
    at react-dom.production.min.js:254
    at Object.unbatchedUpdates (react-dom.production.min.js:190)
d @ react-dom.production.min.js:187
n @ react-dom.production.min.js:181
u @ react-dom.production.min.js:183
c @ react-dom.production.min.js:184
h @ react-dom.production.min.js:188
m @ react-dom.production.min.js:187
[...]

PR #186, opened 2017-10-19
Disable Search Box if CDN is down (#9)

SadPandaBear:
Should help fixing #9 and be a potentially help for #14.


src/components/LayoutHeader/Header.js
```diff
@@ -13,6 +13,7 @@ import React from 'react';
 import {colors, fonts, media} from 'theme';
 import {version} from 'site-constants';
 import ExternalLinkSvg from 'templates/components/ExternalLinkSvg';
+import DocSearch from 'components/DocSearch';
```
bvaughn: Organizational nit: This new `DocSearch` component can just live within the `src/components/LayoutHeader` folder since it's only referenced there.

Particularly since there are styles defined on the `DocSearch` component (eg `flex: '0 0 auto'`) that are dependent on its parent.

src/components/DocSearch/DocSearch.js
```diff
@@ -0,0 +1,110 @@
...
+import {colors, media} from 'theme';
+
+class DocSearch extends Component {
+  state = {
+    disabled: false,
+  };
+  componentDidMount() {
+    // Initialize Algolia search.
+    // TODO Is this expensive? Should it be deferred until a user is about to search?
+    // eslint-disable-next-line no-undef
```
bvaughn: Might be nice but might also be unnecessary. Feel free to leave it for a follow-up, or if you're interested in maybe doing it- measure the perf first to see if it's actually worth doing?

src/components/DocSearch/DocSearch.js
```diff
@@ -0,0 +1,110 @@
...
+        indexName: 'react',
+        inputSelector: '#algolia-doc-search',
+      });
+    } else {
+      this.setState({disabled: true});
+    }
+  }
+
+  render() {
+    const {disabled} = this.state;
```
bvaughn: I think we should just render `null` if search is unavailable. That's less confusing than a disabled search input I think. What do you think?

src/components/DocSearch/DocSearch.js
```diff
@@ -0,0 +1,110 @@
...
+    // TODO Is this expensive? Should it be deferred until a user is about to search?
+    // eslint-disable-next-line no-undef
+    if (window.docsearch) {
+      window.docsearch({
+        apiKey: '36221914cce388c46d0420343e0bb32e',
+        indexName: 'react',
+        inputSelector: '#algolia-doc-search',
+      });
+    } else {
+      this.setState({disabled: true});
```
bvaughn: Might be nice to add a `console.warn` that search failed to load and so it's being disabled.

src/components/DocSearch/DocSearch.js
SadPandaBear: Oh, sorry, I though you meant to disable the input when you said "3. Render the component in a disabled state if so". 😥 😃

src/components/DocSearch/DocSearch.js
SadPandaBear: Maybe not, I just thought of removing the complexity from the Template and just let the DocSearch do everything, mostly because of its own setState behavior.

SadPandaBear:
Turns out I think this PR is more related to #9, since #14 is still happening.

bvaughn:
> since #14 is still happening.

What's the error / stack trace you're seeing in the console after these changes have been made?

src/components/DocSearch/DocSearch.js
bvaughn: No worries at all. I was originally thinking the same, but then I tried this out and thought it might be better just to not show the input at all 😁

SadPandaBear:
> What's the error / stack trace you're seeing in the console after these changes have been made?

The same from before:

screenshot_1

bvaughn:
Ah, I see. I thought before that you were getting an error about `docsearch` being undefined but I see that was actually [reported by another person](https://github.com/reactjs/reactjs.org/issues/14#issuecomment-334832093).

SadPandaBear:
Turning the Cookie button "On" works for me, (which is bizarre since #14 is related to enabling it).

I'm not sure if it helps, but we it's good to consider that [Cookie Whitelist](https://addons.mozilla.org/en-US/firefox/addon/cookie-whitelist-with-buttons/) is Legacy add-on:

screenshot_1

SadPandaBear:
I'm changing the PR name to point to #9.

PR #191, opened 2017-10-21
Update top-level folders explanation

flaviolivolsi:
Following PR #11252 which had major changes regarding the top-level folders structure.


gaearon:
Nice. I made a few more updates too.

Issue #192, opened 2017-10-22
Broken link in the blog


gaearon:
https://reactjs.org/blog/2014/10/16/react-v0.12-rc1.html#react-elements

> We talked about this topic in depth earlier this week. 

Leads nowhere.

bvaughn:
Just to clarify: This task is to fix the broken link to point to `/blog/2014/10/14/introducing-react-elements.html` (eg fix the "introducting" typo)

emmafallancy:
Can I voluntarily take this?

bvaughn:
This issue is all yours< @emmafallancy 😄

I've added an "in-progress" label so that others will know not to start work on the issue. If you change your mind about the issue, no worries! Just let me know so that I can remove the label and free it up for someone else to claim.

Cheers!

emmafallancy:
Yeay!!! Thanks! 🎉

brijeshprasad89:
@emma let me know if you change your mind

gaearon:
Hmm. The issue I see is that the link points to:

https://blog/2014/10/14/introducing-react-elements.html

Which isn't a real URL.

emmafallancy:
@gaearon Yup. There was a typo there. Also, I got question regarding the related matter as well. I've made a PR just now at #193 and a question there too.

gaearon:
Thanks for the PR! I merged it as is but I want to understand two things:

1. Why existing `redirect_from` that you mentioned didn't work. On the old website it seemed to work (and this is why we didn't have this issue earlier).

2. Why, if the website can't find something, it redirects to such a strange URL (with `blog` domain) rather than to our 404 page.

I think we need to keep the issue open until we figure out both of these things. Would you like to continue looking into it?

emmafallancy:
@gaearon yes, I would like to give it a try. Didn't dig further about this just now. Only just found out about the behaviour in question 2. I brought up question 1 is because since we fix the problem for the broken link, was wondering if we ever still need the redirect.

emmafallancy:
@gaearon I've made a PR #194 which leads to the issue in question 2 that you have just now.

bvaughn:
Fixed via #192 and #194

PR #194, opened 2017-10-22
Fix the blog slug html creation that leads to improper redirect

emmafallancy:
@gaearon I've drill down to the question of why it's not redirecting to a proper redirect, especially on `blog`, which you highlighted in #192 . It appears that the creation of slug in `gatsby-node.js` in the root folder contains a forward slash, which I would say create a file with a forward slash in the beginning. Having the forward slash in the beginning seems to have cause a redirect to with `blog` as the domain.

Based on the issue before, having this url -> `<domain>/blog/2014/10/14/introducting-react-elements.html` will redirect to having `blog` as the domain. The fix will resolve it to redirect into `<domain>/blog/2014/10/14/introducing-react-elements.html`.

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -189,7 +189,7 @@ exports.onCreateNode = ({node, boundActionCreators, getNode}) => {
           const day = match[3];
           const filename = match[4];
 
-          slug = `/blog/${year}/${month}/${day}/${filename}.html`;
+          slug = `blog/${year}/${month}/${day}/${filename}.html`;
 
           const date = new Date(year, month - 1, day);
 
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -127,7 +127,7 @@ exports.createPages = async ({graphql, boundActionCreators}) => {
           // A leading "/" is required for redirects to work,
           // But multiple leading "/" will break redirects.
           // For more context see github.com/reactjs/reactjs.org/pull/194
-          const toPath = slug.startsWith('/') ? slug : `/${slug}`
+          const toPath = slug.startsWith('/') ? slug : `/${slug}`;
 
           redirectToSlugMap[fromPath] = slug;
           createRedirect({
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -124,6 +124,9 @@ exports.createPages = async ({graphql, boundActionCreators}) => {
             process.exit(1);
           }
 
+          // A leading "/" is required for redirects to work,
+          // But multiple leading "/" will break redirects.
+          // For more context see github.com/reactjs/reactjs.org/pull/194
           const toPath = slug.startsWith('/') ? slug : `/${slug}`
 
           redirectToSlugMap[fromPath] = slug;
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -124,11 +124,13 @@ exports.createPages = async ({graphql, boundActionCreators}) => {
             process.exit(1);
           }
 
+          const toPath = slug.startsWith('/') ? slug : `/${slug}`
+
           redirectToSlugMap[fromPath] = slug;
           createRedirect({
             fromPath: `/${fromPath}`,
             redirectInBrowser: true,
-            toPath: `/${slug}`,
+            toPath,
           });
         });
       }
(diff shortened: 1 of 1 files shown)


gatsby-node.js
```diff
@@ -189,7 +189,7 @@ exports.onCreateNode = ({node, boundActionCreators, getNode}) => {
           const day = match[3];
           const filename = match[4];
 
-          slug = `/blog/${year}/${month}/${day}/${filename}.html`;
+          slug = `blog/${year}/${month}/${day}/${filename}.html`;
```
gaearon: If we don't need slash here why do we need it on line 206 below?

emmafallancy: I've also notice that most permalink does not contain any forward slash at the beginning of the URL. I could not have find any possible way that any of the pages that falls into line 206. With the blog having the forward slash in front previously had causes an extra forward slash prepended. In line 206, I would not risk saying anything about it but if so, there could also be an extra slash prepended if it goes to line 206, based on the redirect of the `blog`.

screen shot 2017-10-23 at 12 35 55 am

emmafallancy: @gaearon sorry, my bad. For line 206, that only applies to `index.md` in the root content folder. If somehow no _permalink_ is being supplied in any of the `md` file content other than `index.md` in the root content folder, it will give an error based on the slug being created. And because `index.md` does not have any parent to its _relativePath_ and it is a parent of itself, it is necessary for having the forward slash prepended in line 206.

emmafallancy: Maybe we can have something like below for line 206 to prevent unnecessary forward slash to be added if it's not the root `index.md` or coming from having its parent node like `blog`, `docs` for the slug creation. Given if there's a forward slash in the beginning and it does not come from root `index.md`, it will give error...

```js
if (relativePath.match(/.*\/index.md$/)) {
  // create slug with no forward slash in the beginning
} else {
  // create slug with forward slash in the beginning
  slug = `/${relativePath.replace('.md', '.html')}`;
}
```

Just a thought and findings that I made. Happy to hear what you have in mind about this PR 😃 @gaearon

bvaughn: This change (removing the "/" prefix) breaks the main Docs redirect.

To reproduce, click on "Blog" in the top nav, which should redirect to the most recent blog entry, but now/instead redirects to "/blog/blog/2017/09/26/react-v16.0.html" (note the double "blog/blog" at the beginning).

bvaughn: I just built the site locally (`yarn build`) and looked at the generated `public/_redirects` file (the file that's [used by Netlify for redirects](https://www.netlify.com/docs/redirects/)). Looking for this specific redirect I see:
```
/blog/2014/10/14/introducting-react-elements.html  //blog/2014/10/14/introducing-react-elements.html  302
```

The double slash before "blog" shouldn't be there. That problem only occurs for this specific redirect. I _think_ that's because our redirect-creation loop always prepends a "/" before the `slug`:
```js
          createRedirect({
            fromPath: `/${fromPath}`,
            redirectInBrowser: true,
            toPath: `/${slug}`,
          });
```

I think the fix to this might just be to conditionally prefix:
```diff
diff --git a/gatsby-node.js b/gatsby-node.js
index 9613217f8..6846788f5 100644
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -114,11 +114,13 @@ exports.createPages = async ({graphql, boundActionCreators}) => {
             process.exit(1);
           }
 
+          const toPath = slug.startsWith('/') ? slug : `/${slug}`
+
           redirectToSlugMap[fromPath] = slug;
           createRedirect({
             fromPath: `/${fromPath}`,
             redirectInBrowser: true,
-            toPath: `/${slug}`,
+            toPath,
           });
         });
       }
```

That change corrects the entry in the `public/_redirects` file _and_ fixes the (local) redirect logic in the dev server.

bvaughn:
Thanks! 👍 This change seems solid to me.

emmafallancy:
@bvaughn awesome 😄 thanks for pointing out and I wasn't aware about the redirect loop. Thanks for helping me out on it 😃

PR #199, opened 2017-10-23
Tweak Lifecycle Docs

gaearon:
In response to some common confusion points I saw.

### DIFF
--- a/content/docs/reference-react-component.md
+++ b/content/docs/reference-react-component.md
@@ -141,9 +141,13 @@ constructor(props)
 
 The constructor for a React component is called before it is mounted. When implementing the constructor for a `React.Component` subclass, you should call `super(props)` before any other statement. Otherwise, `this.props` will be undefined in the constructor, which can lead to bugs.
 
-The constructor is the right place to initialize state. If you don't initialize state and you don't bind methods, you don't need to implement a constructor for your React component.
+Avoid introducing any side-effects or subscriptions in the constructor. For those use cases, use `componentDidMount()` instead.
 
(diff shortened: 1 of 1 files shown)


content/docs/reference-react-component.md
```diff
@@ -178,7 +184,11 @@ This is the only lifecycle hook called on server rendering. Generally, we recomm
 componentDidMount()
 ```
 
-`componentDidMount()` is invoked immediately after a component is mounted. Initialization that requires DOM nodes should go here. If you need to load data from a remote endpoint, this is a good place to instantiate the network request. Setting state in this method will trigger a re-rendering.
+`componentDidMount()` is invoked immediately after a component is mounted. Initialization that requires DOM nodes should go here. If you need to load data from a remote endpoint, this is a good place to instantiate the network request.
+
+Calling `setState()` in this method will trigger an extra re-rendering, but it is guaranteed to flush during the same tick. This guarantees that even though the `render()` will be called twice in this case, the user won't see the intermediate state. Use this pattern with caution because it often causes performance issues. It can, however, be necessary for cases like modals and tooltips when you need to measure a DOM node before rendering something that depends on its size or position.
```
bvaughn: nit "an extra re-rendering" could just be "an extra rendering"

content/docs/reference-react-component.md
```diff
@@ -178,7 +184,11 @@ This is the only lifecycle hook called on server rendering. Generally, we recomm
...
+
+This method is a good place to set up any subscriptions. If you do that, don't forget to unsubscribe in `componentWillUnmount()`.
```
bvaughn: I wonder if this line might be better off up with the previous paragraph, right after mentioning this method is a good place to instantiate network requests?

PR #200, opened 2017-10-23
Expand on recommended hydrate() usage

gaearon:
Partly addresses https://github.com/reactjs/reactjs.org/issues/25.
See yet another thread: https://github.com/facebook/react/issues/11336.


content/docs/reference-react-dom.md
```diff
@@ -70,7 +70,11 @@ ReactDOM.hydrate(
 )
 ```
 
-Same as [`render()`](#render), but is used to hydrate a container whose HTML contents were rendered by [`ReactDOMServer`](/docs/react-dom-server.html). React will attach event listeners while preserving as much of the existing DOM as possible. For best results, you should try to render the same content on the server as on the client, with as few differences as possible.
+Same as [`render()`](#render), but is used to hydrate a container whose HTML contents were rendered by [`ReactDOMServer`](/docs/react-dom-server.html). React will attempt to attach event listeners to the existing markup.
+
+React expects that the rendered content is identical between the server and the client. It can patch up differences in text content (such as timestamps), but you should treat mismatches as bugs and fix them. React reports any mismatches as warnings during hydration in development. There are no guarantees that attribute differences will be patched up in case of mismatches. This is important for performance reasons because in most apps, mismatches are rare, and so validating all markup would be prohibitively expensive.
```
bvaughn: Suggestion:

Maybe change:
> React reports any mismatches as warnings during hydration in development.

To:
> In development mode, React warns about mismatches during hydration.

gaearon:
Tweaked a bit, mind looking again?

bvaughn:
Why not? They pay me to do this. 😁

bvaughn:
I think the new paragraph is nice too. 👍

Issue #201, opened 2017-10-23
ReactDOM.findDOMNode should be updated with React 16 behavior


ferdaber:
Back in the dark times of when a render function was only able to return a single element, `ReactDOM.findDOMNode` was predictable in that it will only return one and only one element.

With the introduction of being able to return multiple fragments in React 16, the public API docs should be reflected to clarify that it returns the first fragment in the array.

I'd be happy to submit a PR if folks think this is a good idea.

gaearon:
Sounds great. It can also return null for empty components.

bvaughn:
Thanks for reporting this issue @ferdaber! It's all yours! 😄

I've added an "in-progress" label so that others will know not to start work on the issue. If you change your mind about the issue, no worries! Just let me know so that I can remove the label and free it up for someone else to claim.

Cheers!

bvaughn:
Whoops. This was fixed via #202 and just never closed.

PR #202, opened 2017-10-23
feat(react-dom-reference): update findDOMNode API

ferdaber:
This PR modifies the documentation for the `ReactDOM.findDOMNode` API to accommodate the new features in React 16. See #201

### DIFF
--- a/content/docs/reference-react-dom.md
+++ b/content/docs/reference-react-dom.md
@@ -97,7 +97,7 @@ ReactDOM.findDOMNode(component)
 ```
 If this component has been mounted into the DOM, this returns the corresponding native browser DOM element. This method is useful for reading values out of the DOM, such as form field values and performing DOM measurements. **In most cases, you can attach a ref to the DOM node and avoid using `findDOMNode` at all.**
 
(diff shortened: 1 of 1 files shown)


content/docs/reference-react-dom.md
```diff
@@ -95,7 +95,9 @@ Remove a mounted React component from the DOM and clean up its event handlers an
 ```javascript
 ReactDOM.findDOMNode(component)
 ```
-If this component has been mounted into the DOM, this returns the corresponding native browser DOM element. This method is useful for reading values out of the DOM, such as form field values and performing DOM measurements. **In most cases, you can attach a ref to the DOM node and avoid using `findDOMNode` at all.** When `render` returns `null` or `false`, `findDOMNode` returns `null`.
+If this component has been mounted into the DOM, this returns the corresponding native browser DOM element. This method is useful for reading values out of the DOM, such as form field values and performing DOM measurements. **In most cases, you can attach a ref to the DOM node and avoid using `findDOMNode` at all.**
+
+When `render` returns `null` or `false`, `findDOMNode` returns `null`. When `render` returns a non-empty string value or an empty string value, `findDOMNode` will return the primitive string value or an empty text node, respectively. As of React 16, multiple fragments may be returned by `render`, in which case `findDOMNode` will return the first non-null element in the array.
```
bvaughn: > When `render` returns a non-empty string value or an empty string value, `findDOMNode` will return the primitive string value or an empty text node, respectively.

ferdaber: You're right. I was testing this out and was confused by the behavior myself. Turns out that Chrome console will log the string value of a text node if you call it with `console.log`.

gaearon:
Thanks!

PR #203, opened 2017-10-24
Useless caveat about root element in a component

edimitchel:
With new version of React (v16), many root element are acceptable.


gaearon:
I wouldn't quite agree. This note is to warn people against writing

```js
return (
  <div />
  <div />
)
```

which feels intuitive to many people but gives an obscure parse error.

If you remove this note please add something else instead that calls this out. I've seen people repeatedly do this mistake, over and over.

bvaughn:
The messaging before was incorrect and confusing (in my opinion).

I don't think I share the concern about the invalid syntax above, but trusting your judgement- I don't think the previous wording was sufficient to prevent it. 😄

Would be happy to review a PR that adds a new warning for that sort of case though!

PR #213, opened 2017-10-26
Make error boundary doc more in style with the post

gaearon:
I find the clarifications different in style from the original post. Tried to make them look more in line with it.


content/docs/error-boundaries.md
```diff
@@ -136,37 +137,42 @@ However, React components are declarative and specify *what* should be rendered:
 
 Error boundaries preserve the declarative nature of React, and behave as you would expect. For example, even if an error occurs in a `componentDidUpdate` hook caused by a `setState` somewhere deep in the tree, it will still correctly propagate to the closest error boundary.
 
-However, an event handler inside of a React component could leverage `try` / `catch` to deal with errors that occur during event handling.
```
gaearon: This sounded a bit like an ad for try/catch :-)

gaearon:
I’ll just get it in. I think it’s better this way.

PR #215, opened 2017-10-27
Update test-utils to use refs correctly

sericaia:
Update test-utils according to the new way of using refs: https://reactjs.org/docs/refs-and-the-dom.html


gaearon:
Thanks!

Issue #86, opened 2017-10-08
Improve documentation for input type=file


bvaughn:
This issue was originally reported by @kevinsimper via facebook/react/issues/3981

Nowhere on the react documentation is there a reference on how to handle `<input type="file"/>`

It is a speciel field that can not be serialized, and I think it deserve some kind of mention, maybe just a link to some other resource or if it is just like a normal input, just saying that.

I think it might be worth mentioning eg you can't actually programmatically set the value of a file input which is a pretty big deal- it means you're more susceptible to issues with reconciliation. You can only read. You also have to read from `node.files` which is just how those work even outside React, but still might be worth mentioning.

vishalvrv9:
Hello is this issue still open and can I take this up? This is referenced to issue#218 (which is also open)
Sorry I am a little lost (new to open source contributions) but I think I can do this. Thanks.

bvaughn:
This issue is all yours, @vishalvrv9! 😄

I've added an "in-progress" label so that others will know not to start work on the issue. If you change your mind about the issue, no worries! Just let me know so that I can remove the label and free it up for someone else to claim.

Cheers!

bvaughn:
Resolved via #323

Thanks @vishalvrv9 !

PR #218, opened 2017-10-28
doc: added documentation for input type=file

el1f:
Added some explaination on how to implement this form type in React as suggested in #86

### DIFF
--- a/content/docs/forms.md
+++ b/content/docs/forms.md
@@ -190,6 +190,61 @@ Overall, this makes it so that `<input type="text">`, `<textarea>`, and `<select
 ><select multiple={true} value={['B', 'C']}>
 >```
 
+## The file input Tag
+
+In HTML, a `<input type="file">` tag lets the user choose one or more files from their devices storage, allowing the app to handle them to accomplish it's goals.
+
+```html
+<input type="file">
+```
+
+In React a `<input type="file">` works similarly to a normal `<input/>` with some major differences. This input is read-only so programmatically setting the value is impossible.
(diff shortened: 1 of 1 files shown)


content/docs/forms.md
```diff
@@ -190,6 +190,61 @@ Overall, this makes it so that `<input type="text">`, `<textarea>`, and `<select
...
+
+    this.handleFileChange = this.handleFileChange.bind(this);
+    this.handleSubmit = this.handleSubmit.bind(this);
+  }
+
+  handleFileChange(e) {
+    e.preventDefault();
+
+    this.setState({
+      file: e.target.files[0]
```
tsriram: Don't think the file needs to be saved in state, unless I'm overlooking something. Probably having a `ref` to the file input and use it to access the file(s) in submit handler (or wherever required) would be a better option IMHO. See if this makes sense - https://codepen.io/tsriram/pen/qVNoZL?editors=0011

el1f: Yeah, using the file by ref is definitly a betetr solution. My bad. Will change it ASAP

bvaughn: Friendly ping 😄

el1f: Sorry guys, been away for a while and didn't notice the progress of the issue. Seems like @vishalvrv9 has worked on a fix of my PR based on @tsriram's suggestion. Should I go ahead and fix mine or would it better to merge his?

el1f: Yeah, had some issues so I couldn't track down my PR after submitting it but the man definitly deserves some attention for his efforts ;)

PR #220, opened 2017-10-28
Document interaction between native and React events

### LINKED ISSUE #68: Document (non-)interaction between native and synthetic events
This issue was originally reported by @srittau as facebook/react/issues/10284

### Text of the original issue:

I have been trying to find out why my React event handlers were not able to prevent event handlers on document. After hours of searching and experimenting, I stumbled across issue #7094, which explains it.

Unfortunately, the documentation at https://facebook.github.io/react/docs/handling-events.html gives the impression that React events work exactly like native events, apart from wrapping them in synthetic events, and could therefore interact with them. Please document that this is not the case.

The same applies to the documentation at https://facebook.github.io/react/docs/events.html, especially the first paragraph of the overview. A short remark on this page would be helpful as well.

srittau:
Closes #68.

I tried to make the wording as concise and clear as possible, although there is certainly room for improvement. The wording in both files is slightly different to fit the style of the respective document better.

### DIFF
--- a/content/docs/handling-events.md
+++ b/content/docs/handling-events.md
@@ -58,6 +58,10 @@ Here, `e` is a synthetic event. React defines these synthetic events according t
 
 When using React you should generally not need to call `addEventListener` to add listeners to a DOM element after it is created. Instead, just provide a listener when the element is initially rendered.
 
+> Note:
+>
+> For performance reasons, React events are not guaranteed to be installed on the element they are created on. This means that using `stopPropagation` from event handlers installed outside of React might not be able to stop React events from occurring and vice versa. 
+
 When you define a component using an [ES6 class](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Classes), a common pattern is for an event handler to be a method on the class. For example, this `Toggle` component renders a button that lets the user toggle between "ON" and "OFF" states:
 
 ```js{6,7,10-14,18}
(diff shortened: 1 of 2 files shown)


content/docs/handling-events.md
```diff
@@ -58,6 +58,10 @@ Here, `e` is a synthetic event. React defines these synthetic events according t
 
 When using React you should generally not need to call `addEventListener` to add listeners to a DOM element after it is created. Instead, just provide a listener when the element is initially rendered.
 
+> Note:
+>
+> For performance reasons, React events are not guaranteed to be installed on the element they are created on. This means that using `stopPropagation` from event handlers installed outside of React might not be able to stop React events from occurring and vice versa.
```
bvaughn: I'm not sure about the wording "_installed on_" or "_installed outside of_".

srittau: I was looking for a good word here. I am not happy with "installed" myself, but it was the best I could come up with. Suggestions welcome!

bvaughn: (Sorry. Hit enter before I intended to.) What do you think about something like this?

> Note:
>
> React uses event delegation for performance reasons. Generally this is not observable but it is worth noting if you also use the browser's event system directly (via `addEventListener`). For example, calling `stopPropagation()` on a synthetic event will not prevent a native event handler from being executed.

bvaughn: (I'm also not sure about the wording here. Just offering an idea.)

bvaughn: Friendly ping 😄

srittau: @bvaughn Sorry for the late reply. I like your version better than mine, although I needed a bit to understand what "observable" meant in this context. I was looking for a technical meaning. But that is bikeshedding territory.

bvaughn: No problem. Want to take a stab at updating the wording then?

srittau:
I am sorry for dropping this. With React 17, this seems obsolete, closing.

PR #223, opened 2017-10-30
Add note about Function.name to error boundary doc

### LINKED ISSUE #221: Document requirements for component names in error boundary stack traces
Should the requirements for meaningful component names in error boundary stack traces be documented in either the [JavaScript Environment Requirements](../blob/master/content/docs/reference-javascript-environment-requirements.md) (with the other polyfills) or the [Error Boundaries](../blob/master/content/docs/error-boundaries.md) section (as it's specific to these and polyfilling isn't the only option)?

To quote @Kovensky:
> IE11 does not support Function#name, so all your functions (and thus components) are unnamed unless you either set displayName or polyfill it with a getter in Function.prototype. core-js provides such a polyfill.

Ref: facebook/react#11353

wKovacs64:
For the sake of brevity, I intentionally did not include a block of example code with a polyfill like the one in the [JavaScript Environment Requirements](https://reactjs.org/docs/javascript-environment-requirements.html) section. Let me know if you feel one is necessary/valuable and I can add one.

Closes #221.

### DIFF
--- a/content/docs/error-boundaries.md
+++ b/content/docs/error-boundaries.md
@@ -117,6 +117,10 @@ You can also see the filenames and line numbers in the component stack trace. Th
 
 If you don’t use Create React App, you can add [this plugin](https://www.npmjs.com/package/babel-plugin-transform-react-jsx-source) manually to your Babel configuration. Note that it’s intended only for development and **must be disabled in production**.
 
+> Note
+> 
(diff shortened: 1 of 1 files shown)


gaearon:
I'm wary of recommending the whole `core-js`. It's huge. Adding it just to get better dev-only messages in IE 11 seems like a net UX loss to me.

gaearon:
https://github.com/reactjs/reactjs.org/pull/224

bvaughn:
> I'm wary of recommending the whole `core-js`. It's huge.

I agree, but we recommend it a couple of other places so this didn't seem like the best place to mention it.

gaearon:
If you find good Map/Set replacements let's get rid of it in other places too. IMO we should be as granular as we can.

Kovensky:
@gaearon if you want to import _only_ the `Function#name` polyfill (instead of everything), you can do it via `import 'core-js/fn/function/name'`

gaearon:
Does it work well in practice? With other polyfills we noticed that even individual imports drag half of core-js into the bundle.

Kovensky:
In this specific case it looks mostly safe. By following the `require` callse, it looks like it does happen to include an `Object.defineProperty` polyfill, including a check for IE8's broken `defineProperty`, but those are the only extras.

gaearon:
Sounds fine to change to that then. Although I don't care strongly. IMO either is probably fine.

PR #224, opened 2017-10-30
Suggest a smaller Function.name polyfill

gaearon:
See https://github.com/reactjs/reactjs.org/pull/223#issuecomment-340586955.
Seems fairly popular: https://npm-stat.com/charts.html?package=function.name-polyfill&from=2016-10-30&to=2017-10-30

### DIFF
--- a/content/docs/error-boundaries.md
+++ b/content/docs/error-boundaries.md
@@ -119,7 +119,7 @@ If you don’t use Create React App, you can add [this plugin](https://www.npmjs
 
 > Note
 > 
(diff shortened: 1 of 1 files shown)


wKovacs64:
Yeah, I was primarily following suit from the other section with regards to `core-js`. Personally, I only `import 'core-js/es6/function'` rather than the whole thing but `function.name-polyfill` is still smaller. Good call. 👍

bvaughn:
I've poked around a few of the `core-js` polyfills before, and even the standalone ones you get from doing the full imports like that include more boilerplate than is strictly necessary.

wKovacs64:
2.1K gzipped vs 643 bytes, according to Import Cost! 😮

PR #230, opened 2017-11-01
Use `qs` to parse query params in ErrorDecoder

keyanzhang:
When a website uses PHP's `http_build_query` to rebuild a query string, it changes `?invariant=149&args[]=input` to `?invariant=149&args[0]=input` (there's an extra array index `0`). 

More context: http://php.net/manual/en/function.http-build-query.php#77377 (and internal link: https://fburl.com/n6jc0g3g, https://fburl.com/hm2rnd44)

IIRC there wasn't an easy way to add an external library to our documentation site when I was working on the original ErrorDecoder, so I hand-rolled a simple query string parser that couldn't handle this case. Now we have a separate repo so I think it makes more sense to use `qs` for parsing.

Test Plan:
- `yarn run dev`
- The following URLs should all work fine:
  - http://localhost:8000/docs/error-decoder.html?invariant=142&args[]=Foo&args[]=Bar&args[]=Yoooo (this is the one that `error-codes` builds)
  - http://localhost:8000/docs/error-decoder.html?invariant=142&args[0]=Foo&args[1]=Bar&args[2]=Yoooo

### DIFF
--- a/flow-typed/qs.js
+++ b/flow-typed/qs.js
@@ -1,0 +1,5 @@
+declare module 'qs' {
+  declare module.exports: {
+    parse: (str: string, opts: Object) => Object;
+  };
+}
--- a/flow-typed/qs.js
+++ b/flow-typed/qs.js
@@ -1,5 +1,5 @@
 declare module 'qs' {
   declare module.exports: {
-    parse: (str: string, opts: Object) => Object;
+    parse: (str: string, opts: Object) => Object,
   };
 }
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
     "glamor": "^2.20.40",
     "hex2rgba": "^0.0.1",
     "prettier": "^1.7.4",
+    "qs": "^6.5.1",
     "remarkable": "^1.7.1",
     "request-promise": "^4.2.2",
     "rimraf": "^2.6.1",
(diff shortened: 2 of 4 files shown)


bvaughn:
Hey @keyanzhang! Thanks for this fix!

The dependency on `qs` adds an additional 2.15 kB to the total gzipped size of the error decoder bundle. It probably covers a lot more edge-cases than our home-brewed parse, but I wonder if you considered a more incremental fix like...

```diff
diff --git a/src/components/ErrorDecoder/ErrorDecoder.js b/src/components/ErrorDecoder/ErrorDecoder.js
index 3e638d617..782c75de9 100644
--- a/src/components/ErrorDecoder/ErrorDecoder.js
+++ b/src/components/ErrorDecoder/ErrorDecoder.js
@@ -50,8 +50,8 @@ function parseQueryString(location) {
     const query = decodeURIComponent(queries[i]);
     if (query.indexOf('invariant=') === 0) {
       code = query.slice(10);
-    } else if (query.indexOf('args[]=') === 0) {
-      args.push(query.slice(7));
+    } else if (query.indexOf('args[') === 0) {
+      args.push(query.slice(query.indexOf(']=') + 2));
     }
   }
```

sophiebits:
Or like:

```
vr match = /^args\[\d*\]=(.*)$/.exec();
```

bvaughn:
I'm going to finish this PR out by keeping the newly-added Flow types and the class-to-functional component improvements, but removing the added `qs` dependency. Thanks for kicking this off, Keyan!

bvaughn:
Dpme om d23c8c9

PR #236, opened 2017-11-02
Interactive code snippets: first draft of Klipse integration

viebel:
Let's try to have interactive code snippets inside the reactjs documentation and tutorials.
It will be very helpful for newcomers.

Here is a first draft of an integration of [Klipse](https://github.com/viebel/klipse) (as it was suggested in https://github.com/reactjs/reactjs.org/issues/26) using [gatsby-plugin-klipse](https://github.com/ahmedelgabri/gatsby-plugin-klipse) written by @ahmedelgabri.

It looks like this:

jsx-interactive

You can see the live version [here](https://deploy-preview-236--reactjs.netlify.com/docs/jsx-in-depth.html): Enjoy the interactivity!!

There are a few issues:
1. When using `~~~klipse-transpile-jsx` instead of `<pre><code class="gatsby-code-klipse-transpile-jsx"></code></pre>`, the snippet is embedded in a `.gatsby-highlight` that has a black background
2. One has to escape the html tags inside the snippet so instead of `<`, we have to write `&lt'` which is quite cumnbersome
3. Currently the klipse javascript tag is added to all the pages. We need to find a way to selectively add it to a page. See https://github.com/gatsbyjs/gatsby/issues/2681
4. CircleCi has failed: it looks like a network issue.

### DIFF
--- a/content/docs/introducing-jsx.md
+++ b/content/docs/introducing-jsx.md
@@ -18,6 +18,15 @@ It is called JSX, and it is a syntax extension to JavaScript. We recommend using
 
 JSX produces React "elements". We will explore rendering them to the DOM in the [next section](/docs/rendering-elements.html). Below, you can find the basics of JSX necessary to get you started.
 
+
+<pre class="hidden"><code class="gatsby-code-klipse-eval-js" data-external-libs="https://cdnjs.cloudflare.com/ajax/libs/react/15.4.1/react-with-addons.js, https://cdnjs.cloudflare.com/ajax/libs/react/15.4.1/react-dom.js">
+!!React
+</code></pre>
+
+<div>
+Hello World
+</div>
+
 ### Embedding Expressions in JSX
 
 You can embed any [JavaScript expression](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Expressions_and_Operators#Expressions) in JSX by wrapping it in curly braces.
@@ -64,6 +73,29 @@ function getGreeting(user) {
   return <h1>Hello, Stranger.</h1>;
 }
 ```
+
+Hello
+~~~klipse-eval-js
+2+3
+~~~
+
+
+Simple
+
+~~~klipse-render-jsx
+&lt;p&gt;Hello World &lt;/p&gt;
+~~~
+
+Again:
+
+~~~klipse-jsx
+function getGreeting(user) {
+  if (user) {
+    return &lt;h1&gt;Hello, {formatName(user)}!&lt;/h1&gt;;
+  }
+  return &lt;h1&gt;Hello, Stranger.&lt;/h1&gt;;
+}
+~~~
 
 ### Specifying Attributes with JSX
 
@@ -160,3 +192,6 @@ We will explore rendering React elements to the DOM in the next section.
 >**Tip:**
 >
 >We recommend using the ["Babel" language definition](http://babeljs.io/docs/editors) for your editor of choice so that both ES6 and JSX code is properly highlighted. This website uses the [Oceanic Next](https://labs.voronianski.com/oceanic-next-color-scheme/) color scheme which is compatible with it.
+
+
+
--- a/content/docs/jsx-in-depth.md
+++ b/content/docs/jsx-in-depth.md
@@ -31,6 +31,23 @@ React.createElement(
 )
 ```
 
+Here is an interactive version of the jsx snippet:
+~~~klipse-transpile-jsx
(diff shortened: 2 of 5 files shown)


bvaughn:
Thank you for submitting this proof of concept PR. 😄

This component seems functionally similar to the interactive [`CodeEditor` component](https://github.com/reactjs/reactjs.org/blob/master/src/components/CodeEditor/CodeEditor.js) we use on the [home page](https://reactjs.org/). (It has a "JSX?" toggle that can be used to show what the JSX compiles to.) I suppose it's different in that:
* This embeds code that's defined directly within the markdown- which is nice and light-weight.
* It doesn't actually _execute_ the code, only compiles it, which is also nice in some cases.

But it also comes at a pretty large cost; the [`klipse_plugin.min.js`](https://storage.googleapis.com/app.klipse.tech/plugin_prod/js/klipse_plugin.min.js) dependency [this plug-in adds](https://github.com/ahmedelgabri/gatsby-plugin-klipse/blob/master/src/gatsby-ssr.js) is 226 kB gzipped _and_ it also requires Babel which adds another 217 kB gzipped. Due to its size, it also takes a pretty long time to parse and execute (even on my relatively new Macbook Pro):

screen shot 2017-11-06 at 3 48 19 pm

One the one hand, we already load Babel for the examples on the home page, _but_ we only load it if we're actually going to use it. Scoping Klipse to load only for pages that use it would help a bit, but I think the added weight is still too much.

This is a cool proof of concept. Thanks again for sharing it! I think it's best to pass on it for now though, for the reasons mentions above. Let's talk more if you have questions or if I'm misunderstood or misrepresented anything in my initial review. 😄

viebel:
Hi @bvaughn !

Thanks for your feedback.
I understand your concerns about the size and the execution time of the Klipse plugin.
As you mentioned, we could reduce it a bit by making sure that Klipse is not loaded on all pages and not to load the babel plugin from Klipse - as it is already on the page.


More generally, I'd like to better understand your concern: do you think that the user experience is impacted by the time it takes to load and evaluate the script? I've tried to load the page a couple of times on my MacBook pro and I was not able to feel any page freeze or inconsistent UI.

Do you have a different user experience?

bvaughn:
> More generally, I'd like to better understand your concern: do you think that the user experience is impacted by the time it takes to load and evaluate the script? I've tried to load the page a couple of times on my MacBook pro and I was not able to feel any page freeze or inconsistent UI.

Yes, definitely. On the one hand, most people viewing this website are likely developers and so probably have pretty decent hardware. _But_ many are still probably running on machines that are less powerful than your MacBook Pro and maybe in areas with poorer connectivity. A janky loading experience (eg content that shifts and changes after a couple of seconds as the page loads in, and JS parsing that locks the main thread and makes the browser feel unresponsive) is a super frustrating user experience. (Flowtype's REPL is a great example of the latter.)

viebel:
What about adding a button that will allow the user to make the code snippets interactive?
There won't be any performance hit for standard users and this will allow curious users to play with the code snippets...

alexkrolick:
Would it be possible to server-side render the interactive snippets?

viebel:
@alexkrolick interesting idea...

bvaughn:
> What about adding a button that will allow the user to make the code snippets interactive?

I'm not convinced that there is a lot of added value in showing what JSX compiles to all throughout the docs, so long as we cover it in a few places (like the home page and dedicated sections like ["React without JSX"](https://reactjs.org/docs/react-without-jsx.html).

Maybe others feel differently? If so they should weigh in! 😄

viebel:
> I'm not convinced that there is a lot of added value in showing what JSX compiles to all throughout the docs

@bvaughn in some places it is useful to show what JSX compiles to while in other places it is more pertinent to show the result of the rendering. Klipse supports both.

The added value of Klipse is that it allows the user to modify the code of the documentation and see how it affects the output. Some kind of "learning by playing" 😄

bvaughn:
I understand the value of playing with the code and seeing the output. That's why we have the interactive examples on the home page. I think that the majority of our code snippets _within_ the docs wouldn't be runnable though. They're almost always just showing small concepts.

That's why I assumed the only functionality we'd actually get to use for them was the JSX-to-plain-JS transform.

viebel:
Some parts of the doc contain interesting runnable code snippets. 
See for instance a "Klipsification" of an old version of the article about controlled components: http://blog.klipse.tech/react/2016/12/20/react-controlled-components.html

PR #237, opened 2017-11-02
Update reference-react.md

jedwards1211:
The docs don't mention that `key` and `ref` can be overwritten with `cloneElement(element, {key: ..., ref: ...})` and `children` can also be overwritten with `cloneElement(element, {children: ...})`.

### DIFF
--- a/content/docs/reference-react.md
+++ b/content/docs/reference-react.md
@@ -103,15 +103,13 @@ React.cloneElement(
 )
 ```
 
-Clone and return a new React element using `element` as the starting point. The resulting element will have the original element's props with the new props merged in shallowly. New children will replace existing children. `key` and `ref` from the original element will be preserved.
+Clone and return a new React element using `element` as the starting point. The resulting element will have the original element's props with the new props merged in shallowly. New children, if passed as `...children` or otherwise `props.children`, will replace existing children. `key` and `ref` from the original element will be preserved unless `props` contains `key` or `ref` (which will overwrite the `key` or `ref` in the cloned element, respectively).
 
 `React.cloneElement()` is almost equivalent to:
 
 ```js
 <element.type {...element.props} {...props}>{children}</element.type>
 ```
-
-However, it also preserves `ref`s. This means that if you get a child with a `ref` on it, you won't accidentally steal it from your ancestor. You will get the same `ref` attached to your new element.
 
 This API was introduced as a replacement of the deprecated `React.addons.cloneWithProps()`.
 


content/docs/reference-react.md
```diff
@@ -103,16 +103,14 @@ React.cloneElement(
...
-Clone and return a new React element using `element` as the starting point. The resulting element will have the original element's props with the new props merged in shallowly. New children will replace existing children. `key` and `ref` from the original element will be preserved.
+Clone and return a new React element using `element` as the starting point. The resulting element will have the original element's props with the new props merged in shallowly. New children, if passed as `...children` or otherwise `props.children`, will replace existing children. `key` and `ref` from the original element will be preserved unless `props` contains `key` or `ref` (which will overwrite the `key` or `ref` in the cloned element, respectively).
 
 `React.cloneElement()` is almost equivalent to:
 
 ```js
 <element.type {...element.props} {...props}>{children}</element.type>
 ```
 
-However, it also preserves `ref`s. This means that if you get a child with a `ref` on it, you won't accidentally steal it from your ancestor. You will get the same `ref` attached to your new element.
```
bvaughn: I think this is a useful sentence. Maybe we could just tweak the wording slightly?

> Except that it also preserves `ref`s by default. This means that if you get a child with a `ref` on it, you won't accidentally steal it from your ancestor. You will get the same `ref` attached to your new element.

content/docs/reference-react.md
```diff
@@ -103,16 +103,14 @@ React.cloneElement(
 )
 ```
 
-Clone and return a new React element using `element` as the starting point. The resulting element will have the original element's props with the new props merged in shallowly. New children will replace existing children. `key` and `ref` from the original element will be preserved.
+Clone and return a new React element using `element` as the starting point. The resulting element will have the original element's props with the new props merged in shallowly. New children, if passed as `...children` or otherwise `props.children`, will replace existing children. `key` and `ref` from the original element will be preserved unless `props` contains `key` or `ref` (which will overwrite the `key` or `ref` in the cloned element, respectively).
```
bvaughn: The new wording is informative but it reads a little confusingly to me. How do you feel about just slightly tweaking what was there before?

> Clone and return a new React element using `element` as the starting point. The resulting element will have the original element's props with the new props merged in shallowly. New children will replace existing children. `key` and `ref` from the original element will be preserved unless they are explicitly overridden.

content/docs/reference-react.md
jedwards1211: I think this sentence is very unclear though.  Is "child with a `ref` on it" referring to a child of the cloned element or the cloned element itself?

content/docs/reference-react.md
jedwards1211: Hmmm, how can we be precise without being confusing?  "explicitly overridden" doesn't make it clear that they can be overridden by mixing `key` or `ref` in with the other props passed to `cloneElement`.  To me that's not obvious unless explicitly stated because `key` and `ref` are like pseudoprops.

And as far as children goes, "new children will replace existing children" doesn't clarify that existing children can be preserved -- it seems to imply that if `...children` is not provided, then the cloned element will have no children.

The fact that `children` can be overridden via `props.children` is also not obvious at all.   I'm trying to fully and correctly document all behavior of `cloneElement`, which really isn't that trivial...

content/docs/reference-react.md
bvaughn: You could replace, "child with a `ref` on it" with something more like, "if you clone an element that has a `ref`".

jedwards1211: but up above the code example it already says "`key` and `ref` from the original element will be preserved."

rickhanlonii:
Sorry for the long delay here. Since this PR was created against the old site I'm going to close this, but it looks like the new docs explain this already: https://react.dev/reference/react/cloneElement#overriding-props-of-an-element

PR #239, opened 2017-11-03

kinyaying:
1、Declare the defaultProps alone. And it is obeying AirBNB's styleguide.
2、Someone is confused about is it OK to put defaultProps as static props inside React class. Here I explain this question.

### DIFF
--- a/content/docs/typechecking-with-proptypes.md
+++ b/content/docs/typechecking-with-proptypes.md
@@ -148,10 +148,6 @@ MyComponent.propTypes = {
 You can define default values for your `props` by assigning to the special `defaultProps` property:
 
 ```javascript
-const defaultProps = {
-  name: 'Stranger'
-};
-
 class Greeting extends React.Component {
   render() {
     return (
@@ -161,7 +157,9 @@ class Greeting extends React.Component {
 }
 
 // Specifies the default values for props:
-Greeting.defaultProps = defaultProps;
+Greeting.defaultProps = {
+  name: 'Stranger'
+};
 
 // Renders "Hello, Stranger":
 ReactDOM.render(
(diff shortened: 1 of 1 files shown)


content/docs/typechecking-with-proptypes.md
```diff
@@ -157,9 +161,7 @@ class Greeting extends React.Component {
 }
 
 // Specifies the default values for props:
-Greeting.defaultProps = {
-  name: 'Stranger'
-};
+Greeting.defaultProps = defaultProps;
```
bvaughn: I don't understand the motivation behind this change. Compatibility with AirBnb's style-guide seems a bit arbitrary and not necessarily something we need to aim for within the examples shown here.

[AirBnb's own documentation](https://github.com/airbnb/javascript/tree/master/react) shows both formats:
```js
const propTypes = {
  id: PropTypes.number.isRequired,
  url: PropTypes.string.isRequired,
  text: PropTypes.string,
};

const defaultProps = {
  text: 'Hello World',
};

class Link extends React.Component {
  static methodsAreOk() {
    return true;
  }

  render() {
    return <a href={this.props.url} data-id={this.props.id}>{this.props.text}</a>;
  }
}

Link.propTypes = propTypes;
Link.defaultProps = defaultProps;
```
```js
// good
function SFC({ foo, bar, children }) {
  return <div>{foo}{bar}{children}</div>;
}
SFC.propTypes = {
  foo: PropTypes.number.isRequired,
  bar: PropTypes.string,
  children: PropTypes.node,
};
SFC.defaultProps = {
  bar: '',
  children: null,
};
```

This change would also be internally inconsistent with some of the examples above which declare `propTypes` inline. (Why would we treat `defaultProps` differently?)

I tend to think we should revert this bit. 😄

content/docs/typechecking-with-proptypes.md
```diff
@@ -168,4 +170,23 @@ ReactDOM.render(
 );
 ```
 
+You can also put the defaultProps as static props inside React class.
```
bvaughn: How do you feel about combining this paragraph and the one below it into something more like this:

> If you are using a Babel transform like [`transform-class-properties`](https://babeljs.io/docs/plugins/transform-class-properties/), you can also declare `defaultProps` as static property within a React component class. This syntax has not yet been finalized though and will require a compilation step to work within a browser. For more information, see the [class fields proposal](https://github.com/tc39/proposal-class-fields).

content/docs/typechecking-with-proptypes.md
kinyaying: Thank you for your reply, and I will update this commit.

content/docs/typechecking-with-proptypes.md
kinyaying: Thank you for you help. It is a big challenge for a non-native English speaker to write in English.

PR #250, opened 2017-11-06
Alias on webpack production configuration

carboleda:
Despite configuring the UglifyJsPlugin plugin, the message "This page is using the development build of React" continues to appear. Solution found in https://stackoverflow.com/a/37980737/3167601


gaearon:
This is not a right solution. If you have this problem please file an issue on React repo and we will help you determine why your build is not configured correctly.

PR #252, opened 2017-11-07
Community Section Update

mateoholman:
- Removed outdated/broken/invalid links
- Added new content / resources
- Added a "tools" section
- Organized layout and links within each subsection


bvaughn:
Why is there a 10,399 line long `yarn-error.log` file in this PR? 😄 Seems unintentional?

Also, you have a couple of conflicts you might want to rebase and resolve

mateoholman:
Oh boy, that error log was totally unintentional! :stuck_out_tongue_closed_eyes: Sorry about that! I believe I resolved the conflicts. I rebased and squashed some commits to try and tidy things up. This is my first attempt at a respectable contribution, so don't hesitate to point anything else out that looks wacky! :dizzy_face:

gatsby-node.js
```diff
@@ -231,7 +231,9 @@ exports.onCreateNode = ({node, boundActionCreators, getNode}) => {
         // This should only happen for the partials in /content/home,
         // But let's log it in case it happens for other files also.
         console.warn(
-          `Warning: No slug found for "${relativePath}". Falling back to default "${slug}".`,
+          `Warning: No slug found for "${
+            relativePath
+          }". Falling back to default "${slug}".`,
```
bvaughn: I would not expect to see white-space only changes like this. Did you run `yarn prettier` before committing? 😄

package.json
```diff
@@ -49,6 +49,7 @@
     "gatsby-transformer-remark": "^1.7.2",
     "gatsby-transformer-sharp": "^1.6.1",
     "glamor": "^2.20.40",
+    "graphql": "0.10.5",
```
bvaughn: I wouldn't expect to see this change, or any of the changes to the `yarn.lock` file below, given this PR mostly touches markdown files. Did you mean to commit them?

bvaughn:
Looks like the `yarn.lock` file has snuck back in again. 😁

Maybe just try saving [this one from GitHub](https://raw.githubusercontent.com/reactjs/reactjs.org/master/yarn.lock) over the one in your branch and committing it?

mateoholman:
Hmmm. I reverted back as requested to remove the changes mentioned in the package.json file. My yarn.lock file has conflicts due to a number of dependencies being behind master, is that normal? Also, I ran yarn check-all & yarn prettier before the commit, but it still looks like some of those white-space only changes are showing up? Should I change them back manually & commit? Thanks for your patience @bvaughn :smile:

bvaughn:
> Hmmm. I reverted back as requested to remove the changes mentioned in the package.json file. My yarn.lock file has conflicts due to a number of dependencies being behind master, is that normal?

This commit should only be adding markdown files, so no changes should be needed in either `package.json` nor `yarn.lock` files. (Normally we only update a lock file in its own commit, so we can carefully test any changes made by updated dependencies and revert them if needed.)

You probably accidentally updated the file somehow. (Not sure.) But it should be safe to revert it to exactly match what's in master, and then `rm -rf node_modules && yarn install` again to make sure you've got exactly the same dependencies as what's checked in.

> I ran yarn check-all & yarn prettier before the commit, but it still looks like some of those white-space only changes are showing up? Should I change them back manually & commit?

Really not sure what's going on here. Running `yarn prettier` on master doesn't update any of the files your diff changes. You can go ahead and revert those to match master as well, though it would be nice to know why your Prettier command isn't working. (Maybe it has something to do with your lock file?)

If I save your modifications locally and run `yarn prettier`, it undoes all of your whitespace changes FWIW.

mateoholman:
That didn't seem to work either. I'll revert the commit and try grabbing the files from master so they match and try again.

mateoholman:
Phew, that seemed to have done the trick with the white-space issues. It still shows the yarn.lock with changes, but I believe it is the same as the master branch yarn.lock file now? :grin:

bvaughn:
I think you're right that the contents of `yarn.lock` are the same now, in which case what's probably happened is that your branch is just out of date with `master`. (`yarn.lock` in your branch is different than `master` at the time your branch was created, so Git thinks you've changed it- even though in this case it happens to match what's in `master` presently.)

One way to avoid this sort of thing is to either run `git rebase master` or `git merge master` from your branch occassionally, to keep it up to date. This is a good habit anyway b'c it minimizes the risk of conflict when you create a PR or try to actually merge your branch back into `master` later. 😄

bvaughn:
There are many ways you can recover from this situation, but maybe the simplest one in this case is just to make a new branch 😄 Here's one way you could do it:

```bash
# Update your local copy of master
git fetch origin
git checkout master
git pull origin master

# Update your branch to be in-sync with master
git checkout mateoholman-community-section
git merge master

# Dump a patch of the differences between your branch and master
git diff master > ~/Desktop/diff

# Make a new branch from master
git checkout master
git checkout -b mateoholman-community-section-new

# Patch your changes onto this new, clean branch
git apply ~/Desktop/diff

# Visually verify that no unexpected changes are here (like yarn.lock)
# Then commit the changes
git status
git add -- .
git commit -m "<your commit message here>"

# Push your new, clean branch and create a PR from it
git push mateoholman mateoholman-community-section-new
```

I think GitHub will let you update this PR to point to the new branch once you have done this. Worst case you can close this PR and open a new one and CC me on it.

mateoholman:
I couldn't change this PR to point to a new target (I can only change the base I think), so I'll open a new one and CC you. Thanks again for all your help and patience with this PR -- I won't make the same mistakes in future PR's! :relieved:

PR #253, opened 2017-11-07
Small error with context.md

whw19950510:
Line 25 actually you want to write "are familiar", is that right?


bvaughn:
I can see the confusion here, but I think "aren't" is the correct word. This section of the docs is basically saying- "don't use `context` before you've become familiar with other, existing options because those options probably work better for what you're doing anyway" 😄

PR #261, opened 2017-11-08
Link doesn't work

Sedaj:


bvaughn:
The link in question seems to work fine for me. 😄 Maybe try again? Perhaps the goshakkk.name domain was temporarily down?

Sedaj:
just interesting configuration of our work network

PR #267, opened 2017-11-09
Fix webpack `DefinePlugin` usage example

ogonkov:
Shorten form is [preferable](https://webpack.js.org/plugins/define-plugin/#feature-flags)


ogonkov:
Ugh, seems i need to make pull request on their side aswell

bvaughn:
Maintaining docs is hard work 😄 I'm sure they just overlooked it.

PR #275, opened 2017-11-11
Update Portal Example: Render Modal children when Modal is inserted in the DOM tree

### LINKED ISSUE #272: Portal example can cause issues if children access the DOM on componentDidMount
In the Portal example, the portal container is mounted in the DOM in the `Modal`'s `componentDidMount` method

```
componentDidMount() {
  modalRoot.appendChild(this.el);
}
```

This means that when the portal is being mounted, children are initially mounted on a detached node and their `componentDidMount` is called when the component's node is not yet inserted in the DOM. Therefore, any DOM method that requires nodes to be mounted (getting node dimensions, focusing an element, etc.) and props like `autoFocus` will not work.

```
class Child extends React.Component {
  componentDidMount() {
    setTimeout(() => {
      let node = ReactDOM.findDOMNode(this);

      console.log(node.clientWidth);
      // 1980

      console.log(document.body.contains(node));
      // true
    }, 0);

    let node = ReactDOM.findDOMNode(this);

    console.log(node.clientWidth);
    // 0

    console.log(document.body.contains(node));
    // false
  }
  // The click event on this button will bubble up to parent,
  // because there is no 'onClick' attribute defined
  render() {
    return (
      <div className="modal">
        {/* Input will not focus */}
        <input autoFocus />
        <button>Click</button>
      </div>
    );
  }
}
```

This can be fixed by either moving Modal's `appendChild` method to `componentWillMount` or using the pattern used by https://github.com/tajo/react-portal/blob/master/src/Portal.js:

```
class Modal extends React.Component {
  componentWillUnmount() {
    if (this.el) {
      modalRoot.removeChild(this.el);
    }
  }

  render() {
    if (!this.el) {
      this.el = document.createElement("div");
      modalRoot.appendChild(this.el);
    }
    return ReactDOM.createPortal(this.props.children, this.el);
  }
}
```

darrenscerri:
Fixes #272 

Not sure if this is clear enough. Maybe it might be better to move the `Modal` implementation to a separate section along with a beefier explanation?

Also noticed that there is a Modal implementation in a Codepen linked from the Portals page (https://codepen.io/gaearon/pen/yzMaBd) that may cause the mentioned issues.

### DIFF
--- a/content/docs/portals.md
+++ b/content/docs/portals.md
@@ -65,7 +65,7 @@ This includes event bubbling. An event fired from inside a portal will propagate
 
 A `Parent` component in `#app-root` would be able to catch an uncaught, bubbling event from the sibling node `#modal-root`.
 
-```js{20-23,34-41,45,53-55,62-63,66}
+```js{29-34,45-52,56,64-66,73-75,77}
 // These two containers are siblings in the DOM
 const appRoot = document.getElementById('app-root');
 const modalRoot = document.getElementById('modal-root');
@@ -74,10 +74,19 @@ class Modal extends React.Component {
   constructor(props) {
     super(props);
     this.el = document.createElement('div');
+    // We need to keep track of when the portal element
+    // is inserted in the DOM tree since we should only
+    // render the modal's children on a mounted element
+    this.state = {
+      mounted: false
+    };
   }
 
   componentDidMount() {
     modalRoot.appendChild(this.el);
+    this.setState({
+      mounted: true
+    });
   }
 
   componentWillUnmount() {
@@ -86,7 +95,9 @@ class Modal extends React.Component {
 
   render() {
     return ReactDOM.createPortal(
-      this.props.children,
+      // This will allow any children's 'componentDidMount()'
+      // to be called on a mounted node
+      this.state.mounted ? this.props.children : null,
       this.el,
     );
   }
--- a/content/docs/portals.md
+++ b/content/docs/portals.md
@@ -65,7 +65,7 @@ This includes event bubbling. An event fired from inside a portal will propagate
 
 A `Parent` component in `#app-root` would be able to catch an uncaught, bubbling event from the sibling node `#modal-root`.
 
-```js{29-34,45-52,56,64-66,73-75,77}
+```js{28-31,42-49,53,61-63,70-71,74}
 // These two containers are siblings in the DOM
 const appRoot = document.getElementById('app-root');
 const modalRoot = document.getElementById('modal-root');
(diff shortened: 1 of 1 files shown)


content/docs/portals.md
```diff
@@ -86,7 +95,9 @@ class Modal extends React.Component {
 
   render() {
     return ReactDOM.createPortal(
-      this.props.children,
+      // This will allow any children's 'componentDidMount()'
+      // to be called on a mounted node
```
bvaughn: There's no point to call `createPortal` with a `null` value. Maybe it would be nicer to do something like this:
```js
class Modal extends React.Component {
  _container = document.createElement('div');

  // Keep track of when the modal element is added the DOM
  state = {
    mounted: false
  };

  componentDidMount() {
    modalRoot.appendChild(this._container);

    this.setState({
      mounted: true
    });
  }

  componentWillUnmount() {
    modalRoot.removeChild(this._container);
  }

  render() {
    // Don't render the portal until the component has mounted,
    // So the portal can safely access the DOM.
    if (!this.state.mounted) {
      return null;
    }

    return ReactDOM.createPortal(
      this.props.children,
      this._container,
    );
  }
}
```

Of course, now that I think about this, this implementation _also_ has the potential to cause problems. For example, imagine if someone used this `Modal` component like so:
```jsx
// `Child` in this example is a class component
class Parent extends React.Component {
  _childRef = null;

  componentDidMount() {
    // Someone might expect _childRef to be set
  }

  render() {
    return (
      <Modal>
        <Child ref={this._setRef} />
      </Modal>
    );
  }

  _setRef = ref => {
    this._childRef = ref;
  }
}
```

darrenscerri: The ref issue is analogous to the `Child`'s `componentDidMount` issue in the sense that it still doesn't respect certain assumptions, so I think the example in the PR would be a no-go. I cannot think of a 100% correct solution that would mitigate these issues without introducing side-effects in the `constructor`, `componentWillMount` or `render` methods of `Modal`.

Honestly, the current `Modal` implementation wouldn't cause any issues in the majority of use cases, unless it's used in a way that its children need to perform any DOM actions that assume a mounted DOM (measuring dimensions, checking ancestors, etc.), or having `autoFocus` in any of its descendants. Still, this breaks the unwritten (and justified) assumption that a `componentDidMount` is invoked when the component is in fact mounted and available in the DOM tree. This also may introduce inconsistencies in components as they may behave differently if mounted in a `Modal` or not.

Maybe the best way forward is to add a note in the docs highlighting these possible concerns?

bvaughn: > Maybe the best way forward is to add a note in the docs highlighting these possible concerns?

Agreed! Given the trade-offs we've discussed, maybe the best option at the moment (unless someone else can think of a better one?) is just to add a note inline to the existing example, as you say.

bvaughn: (Would you like to submit a new PR that adds that comment? Or update this one to do that instead?)

darrenscerri: I can update this one and then squash commits when merging if that's ok.

bvaughn: Sounds great!

darrenscerri:
Thanks a lot 😄 👍

PR #285, opened 2017-11-14
Update static-type-checking.md

tranvansang:
The document did not specify where to edit `babel` config.

It is confused for beginner

### DIFF
--- a/content/docs/static-type-checking.md
+++ b/content/docs/static-type-checking.md
@@ -24,10 +24,9 @@ yarn add --dev babel-preset-flow
 npm install --save-dev babel-preset-flow
 ```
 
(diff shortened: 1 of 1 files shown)


content/docs/static-type-checking.md
```diff
@@ -24,7 +24,7 @@ yarn add --dev babel-preset-flow
 npm install --save-dev babel-preset-flow
 ```
 
-Then add `flow` to your Babel presets config.
+Then add `flow` to your Babel presets config in`.babelrc`.
```
bvaughn: There should be a space between "in`.babelrc`."

I'm going to just modify this myself directly.

PR #296, opened 2017-11-17
Parallelize some yarn scripts

clemmy:
Preparation to make it possible to do something like:

```
npm-run-all --parallel install translations --serial build
```

since `yarn install` and downloading translations should be able to be done in parallel. Not sure if 
`npm-run-all` is the way to go though since we need `yarn install` to run first in order to have `npm-run-all` available as a binary...

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,6 +78,7 @@
     "check-all": "npm-run-all prettier --parallel lint flow",
     "ci-check": "npm-run-all prettier:diff --parallel lint flow",
     "dev": "gatsby develop -H 0.0.0.0",
+    "flow": "flow",
     "lint": "eslint .",
     "netlify": "yarn install && yarn build",
     "prettier": "prettier --config .prettierrc --write \"{gatsby-*.js,{flow-typed,plugins,src}/**/*.js}\"",
--- a/package.json
+++ b/package.json
@@ -13,7 +13,6 @@
     "eslint-plugin-relay/graphql": "0.10.5"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -75,8 +75,8 @@
   },
   "scripts": {
     "build": "gatsby build",
-    "check-all": "yarn prettier && yarn lint && yarn flow",
-    "ci-check": "yarn prettier:diff && yarn lint && yarn flow",
+    "check-all": "npm-run-all prettier --parallel lint flow",
```
bvaughn: Hm. Did this command work for you locally?

I needed to add a `flow` script entry before it would work for me.

```
$ yarn check-all
...
ERROR: Task not found: "flow"

$ npm run ci-check
...
ERROR: Task not found: "flow"
```

Fixed by:
```diff
diff --git a/package.json b/package.json
index 94e16032b..7ebc70ae4 100644
--- a/package.json
+++ b/package.json
@@ -78,6 +78,7 @@
     "check-all": "npm-run-all prettier --parallel lint flow",
     "ci-check": "npm-run-all prettier:diff --parallel lint flow",
     "dev": "gatsby develop -H 0.0.0.0",
+    "flow": "flow",
     "lint": "eslint .",
     "netlify": "yarn install && yarn build",
     "prettier": "prettier --config .prettierrc --write \"{gatsby-*.js,{flow-typed,plugins,src}/**/*.js}\"",
```

bvaughn: I'm just going to add this myself 😁

PR #305, opened 2017-11-20
Add debounce and throttle examples for event handlers

alexkrolick:


content/docs/faq-functions.md
```diff
@@ -177,3 +177,52 @@ class Alphabet extends React.Component {
...
+class Searchbox extends React.Component {
+  handleChange = event => {
+    event.persist()
+    this._handleChangeDebounced(event)
+  };
+
+  _handleChangeDebounced = debounce(event => {
+      this.props.onChange(event.target.value)
+    }, 250)
+  };
```
bvaughn: Syntax error ^

Remove the `}`

content/docs/faq-functions.md
```diff
@@ -177,3 +177,52 @@ class Alphabet extends React.Component {
...
+```
+
+#### Debounce
+
+```jsx
+import { debounce } from 'lodash'
+
+class Searchbox extends React.Component {
+  handleChange = event => {
+    event.persist()
```
bvaughn: Maybe we should add an inline comment here pointing to:
https://reactjs.org/docs/events.html#event-pooling

Also might be nicer to just pass `event.target.value` to the debounced handler, rather than the whole event, so we don't tie up pooled events unnecessarily.

content/docs/faq-functions.md
```diff
@@ -177,3 +177,52 @@ class Alphabet extends React.Component {
...
 ```
+
+### How can I prevent a function from being called too quickly or too many times in a row?
+
+If you have an event handler such as `onClick` or `onScroll` and want to prevent the callback from being fired too quickly, you can wrap the handler with a utility such as [`_.debounce`](https://lodash.com/docs#debounce) or [`_.throttle`](https://lodash.com/docs#throttle). For a given `delay`, say `100ms`, debouncing calls the handler after activity stops for that amount of time; throttling prevents the handler from being called more than once per `delay`. See a visualization [here](http://demo.nimius.net/debounce_throttle/).
+
+#### Throttle
+
+```jsx
+import { throttle } from 'lodash'
```
bvaughn: Let's show:
```js
import throttle from 'lodash.throttle';
```
So we don't encourage depending on all of Lodash for a one-off function.

content/docs/faq-functions.md
```diff
@@ -177,3 +177,52 @@ class Alphabet extends React.Component {
...
+  render() {
+    return <button onClick={this.handleClick}>Load More</button>
+  }
+}
+```
+
+#### Debounce
+
+```jsx
+import { debounce } from 'lodash'
```
bvaughn: Let's show:
```js
import debounce from 'lodash.debounce';
```
So we don't encourage depending on all of Lodash for a one-off function.

content/docs/faq-functions.md
alexkrolick: Sure. That actually took me a while to figure out 😄 .

content/docs/faq-functions.md
```diff
@@ -201,18 +201,17 @@ class LoadMoreButton extends React.Component {
...
 
 ```jsx
-import { debounce } from 'lodash'
+import debounce from 'lodash.debounce'
 
 class Searchbox extends React.Component {
   handleChange = event => {
     event.persist()
-    this._handleChangeDebounced(event)
+    this._handleChangeDebounced(event.target.value)
```
bvaughn: Now that you're doing this, you no longer need to call `event.persist()` 😄

alexkrolick: Ah right... I think the event.persist() example is important. It's easy to forget and causes quite the headache when debouncing forms. What do you think about reverting this change and adding a note about capturing `value` as an alternative to `.persist` above?

content/docs/faq-functions.md
```diff
@@ -177,3 +177,53 @@ class Alphabet extends React.Component {
...
+
+#### Debounce
+
+```jsx
+import debounce from 'lodash.debounce'
+
+class Searchbox extends React.Component {
+  handleChange = event => {
+    event.persist() // https://reactjs.org/docs/events.html#event-pooling
+    this._handleChangeDebounced(event.target.value)
```
bvaughn: To clarify a bit: You only need to do 1 of the following:

1. Synchronously read the value from `event.target.value` and store it for later, or...
1. Call `event.persist()` so that React doesn't recycle the pooled event.

I think the former is better since we don't really need to freeze the whole event just to read that one value. I don't think we should show both because people are likely to assume you _need_ to do both and just copy-paste that pattern.

I would suggest changing this function to the something like:

```js
handleChange = event => {
  // React pools events, so we read the value before debounce.
  // Alternately we could call `event.persist()`
  // For more info see reactjs.org/docs/events.html#event-pooling
  this._handleChangeDebounced(event.target.value);
};
```

alexkrolick: Sounds good. At some point if there a lot of questions about it Event Pooling might need its own FAQ - "Why do am I getting `cannot read property value of null` in an event handler?"

bvaughn:
Ooh, sorry. I hadn't noticed that you had the "Note" above the example before. That's fine too. 😄  Either way.

alexkrolick:
I like the inline example. It explains the problem and 2 possible fixes. If people have Event Pooling issues I think we can do what I suggested above and add an FAQ for that topic.

content/docs/faq-functions.md
```diff
@@ -177,3 +177,69 @@ class Alphabet extends React.Component {
...
+class LoadMoreButton extends React.Component {
+  componentWillUnmount() {
+    this._handleClick.cancel();
+  }
+
+  render() {
+    return <button onClick={this._handleClick}>Load More</button>;
+  }
+
+  _handleClick = throttle(() => {
```
gaearon: This uses a syntax that isn't part of any ES standard. So far we've avoided such syntax anywhere else in the documentation except the page specifically explaining it (with an "Experimental" warning).

Are we making an exception here? Or do we intentionally want to start using unstable syntax elsewhere in the docs (e.g. on the homepage)? How do we decide when to use it?

gaearon: I'll send a PR since it's not essential here.

bvaughn: Good question, Dan.

This example uses the class public fields proposal (currently in Stage 2) because of its clarity.

I was under the impression we used this syntax elsewhere (like on the ["React Without ES6"](https://reactjs.org/docs/react-without-es6.html) and ["Component State"](https://reactjs.org/docs/faq-state.html) pages) but it's not very common.

bvaughn: > I'll send a PR since it's not essential here.

Sounds good to me 👍

gaearon: On https://reactjs.org/docs/react-without-es6.html, it specifically calls out:

```js
  // WARNING: this syntax is experimental!
  // Using an arrow here binds the method:
```

and mentions:

>you may enable the **experimental** Class Properties syntax proposal with Babel:

https://reactjs.org/docs/faq-state.html is also a new page so it isn't consistent with the original docs either.

alexkrolick: There is an explanation for the class-fields-for-bound-functions on this same FAQ page, so it seems like fair game. But if it clears things up, sure, bind in constructor would be fine.

bvaughn: Yes, I saw the note on the ES6 page. And as I said above, I'm fine with removing the example.

alexkrolick: As a side note, what do you think of _defining_ bound functions in the constructor instead of binding the prototype methods?

```js
class Comp extends React.Component {
  constructor(props) {
    super(props)
    this.handleFoo = () => { /* method definition, already bound */ }
  }
}
```

gaearon: I'm all for mentioning it :-) Perhaps we could have a "how to write more concise code" page where we explain all those tricks. But IMO code from doc pages need to be copy-pasteable into browsers, fiddles, editors, etc, without knowledge of special proposals.

gaearon:
To be clear, didn't mean to criticise the work here—I love that we now have the FAQ! Just want to make sure we keep the docs consistent from now on. (They used to be inconsistent in the past, and reaching consistency took some effort.)

I agree a better way to do this would be to create CONTRIBUTING file instead of me ranting over merged PRs. :-)

bvaughn:
A `CONTRIBUTING.md` file would probably help. Let's work together to create one. 😄 

I suspect that a high level of consistency will always require judicious, experienced gate keepers. I'm obviously very familiar with the website code, but I'm still learning some things like this WRT the docs content. I appreciate your help.

alexkrolick:
Contributing guide PR - https://github.com/reactjs/reactjs.org/pull/311

PR #308, opened 2017-11-20
Minor FAQ updates

gaearon:
Removes unnecessary experimental syntax, and then adds a bunch of small things I wanted to add.

### DIFF
--- a/content/docs/faq-ajax.md
+++ b/content/docs/faq-ajax.md
@@ -48,7 +48,7 @@ class MyComponent extends React.Component {
           this.setState({
             isLoaded: true,
             items: result.items
-          })
+          });
         },
         // Note: it's important to handle errors here
         // instead of a catch() block so that we don't swallow
@@ -57,7 +57,7 @@ class MyComponent extends React.Component {
           this.setState({
             isLoaded: true,
             error
-          })        
+          });
         }
       )
   }
--- a/content/docs/faq-ajax.md
+++ b/content/docs/faq-ajax.md
@@ -31,27 +31,35 @@ The example API returns a JSON object like this:
 
 ```jsx
 class MyComponent extends React.Component {
-  state = {
-    error: null,
-    isLoaded: false,
-    items: []
-  };
+  constructor(props) {
+    super(props);
+    this.state = {
+      error: null,
+      isLoaded: false,
+      items: []
+    };
+  }
 
   componentDidMount() {
     fetch("https://api.example.com/items")
       .then(res => res.json())
-      .then(result =>
-        this.setState({
-          isLoaded: true,
-          items: result.items
-        })
+      .then(
+        (result) => {
+          this.setState({
+            isLoaded: true,
+            items: result.items
+          })
+        },
+        // Note: it's important to handle errors here
+        // instead of a catch() block so that we don't swallow
+        // exceptions from actual bugs in components.
+        (error) => {
+          this.setState({
+            isLoaded: true,
+            error
+          })        
+        }
       )
-      .catch(error =>
-        this.setState({
-          isLoaded: true,
-          error
-        })
-      );
   }
 
   render() {
@@ -59,7 +67,7 @@ class MyComponent extends React.Component {
     if (error) {
       return <div>Error: {error.message}</div>;
     } else if (!isLoaded) {
-      return <div>Loading ...</div>;
+      return <div>Loading...</div>;
     } else {
       return (
         <ul>
--- a/content/docs/faq-functions.md
+++ b/content/docs/faq-functions.md
@@ -25,13 +25,13 @@ There are several ways to make sure functions have access to component attribute
 ```jsx
 class Foo extends Component {
   constructor () {
-    this.handleClick = this.handleClick.bind(this)
+    this.handleClick = this.handleClick.bind(this);
   }
   handleClick() {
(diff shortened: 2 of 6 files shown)


alexkrolick:
@gaearon my bad, do you think you could add a CONTRIBUTING guide for what syntax is OK where and what linter rules are preferred?

content/docs/faq-ajax.md
```diff
@@ -31,35 +31,43 @@ The example API returns a JSON object like this:
...
+      .then(
+        (result) => {
+          this.setState({
+            isLoaded: true,
+            items: result.items
+          })
+        },
+        // Note: it's important to handle errors here
+        // instead of a catch() block so that we don't swallow
+        // exceptions from actual bugs in components.
```
alexkrolick: IMO this needs more explanation - most people do not know the second argument to  `.then` is an error handler.

Very good point though, did not think of that.

gaearon: Happy to take as a further PR :-)

content/docs/faq-functions.md
```diff
@@ -149,28 +164,33 @@ Alternately, you can use DOM APIs to store data needed for event handlers. Consi
...
+    this.handleClick = this.handleClick.bind(this);
+    this.state = {
       justClicked: null,
       letters: Array.from({length: 26}, (_, i) => String.fromCharCode(A + i))
+    };
   }
-  
-  handleClick = event => {
+
+  handleClick(e) {
```
gaearon: Intentionally using `e` in these examples. When it's typed as `event` it's easy to forget the `event` argument and accidentally use the `window.event` global that is provided by the browser for IE compatibility (and has nothing to do with React).

gaearon:
I'll get this in—happy to follow up if something's wrong.

content/docs/faq-functions.md
```diff
@@ -90,6 +90,25 @@ Generally speaking, yes, it is OK, and it is often the easiest way to pass param
...
+```
+
+```js
+var method = obj.method();
+method();
+```
+
+Binding methods helps ensure that the second snippet works the same way as the first one.
+
+With React, typically you only need to bind the methods you *pass* to other components. For example, `<button onClick={this.handleClick}>` passes `this.handleClick` so you want to bind it. However, it is unnecessary to bind the `render` method or the lifecycle methods: we don't pass them to other components.
```
alexkrolick: Great explanation

content/docs/faq-ajax.md
migueloller: @gaearon, would you mind explaining why exceptions would be swallowed when using `catch`? From my understanding of how promises work, it would seem that `catch` would actually catch the error, but if there's an error in the error callback passed (lines 56-61), then _that_ would be swallowed.

gaearon: My point was that

```js
something
  .then(() => this.setState(...))
  .catch(() => this.setState(...))
```

is very easy to mess up because usually people try to `catch` exceptions from `something` and don't realize exception in the first `setState` would *also* trigger second `setState`. If your component throws *rendering*, by that point you have nothing to `setState` on because the tree has been unmounted. So the original error often ends up ignored and not logged. React 16 made this better by always logging rendering errors (even when caught) but it's still a problem.

migueloller: Ahh, I see! Then perhaps something like this could be clearer?

```jsx
    fetch("https://api.example.com/items")
      .then(res => res.json())
      .then(
        (result) => {
          this.setState({
            isLoaded: true,
            items: result.items
          })
        },
        (error) => {
          this.setState({
            isLoaded: true,
            error
          })        
        }
      )
      .catch(
        (error) => {
          // If your component throws an exception in `render`
          // it has to be caught here.
        }
      )
```

Or this:

```jsx
    fetch("https://api.example.com/items")
      .then(res => res.json())
      .then(
        (result) => {
          this.setState({
            isLoaded: true,
            items: result.items
          })
        }
      )
      .catch(
        (error) => {
          this.setState({
            isLoaded: true,
            error
          })        
        }
      )
      .catch(
        (error) => {
          // If your component throws an exception in `render`
          // from the previous `catch` `setState` call it has
          // to be handled here.
        }
      )
```

Because I believe the way it is right now, if the `setState` in line 57 causes an exception in the component, that exception will be swallowed.

I wouldn't mind opening a PR, I know you're likely super busy with other stuff.

migueloller: Here's a sandbox to play around with this: https://codesandbox.io/s/7zw59n0130

gaearon: >if the setState in line 57 causes an exception in the component, that exception will be swallowed

It will be an uncaught rejection, wouldn't it? That's a different thing from being swallowed. Uncaught rejections are reported by Chrome (and throw in Node).

alexkrolick: @migueloller I opened an issue to discuss this topic further with a bit more visibility (outside of this closed PR) - care to take the discussion over here: https://github.com/reactjs/reactjs.org/issues/309?

migueloller: @alexkrolick, thanks! That's a good idea 😄

magbicaleman: I vote for adding this discussion as a read more link in this particular doc page.

PR #313, opened 2017-11-21
Update content: more idiomatic use of uncontrolled forms;

everdimension:
Hi!

I think using uncontrolled components doesn't mean you have to immediately reach for `refs`. React docs themselves call `refs` an escape hatch.

We can do without them much more elegantly.

* First, we should always give `name` attributes to our inputs. I think it's a bad habit not to do this and I've always loved the react docs because they promote good practices.
* Second, once we do that, we can [easily access](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement#Properties) any form value by that name.

---

Food for thought (or for discussion):
Accessing the input by its name is not the only way to read form values. There are other ways:

1. Use [FormData](https://developer.mozilla.org/en-US/docs/Web/API/FormData) API. I guess although it's a perfectly fine and a robust way to read form data, it has too many nuances which may complicate understanding for beginners.

2. We can still use `onChange` callbacks and update state whenever inputs' values change. Currently the docs page may be misunderstood and give an impression that the `onChange` prop won't work. But of course it does.

Also the codepen example should be updated, so I left a note for now.

### DIFF
--- a/content/docs/uncontrolled-components.md
+++ b/content/docs/uncontrolled-components.md
@@ -6,11 +6,11 @@ permalink: docs/uncontrolled-components.html
 
 In most cases, we recommend using [controlled components](/docs/forms.html) to implement forms. In a controlled component, form data is handled by a React component. The alternative is uncontrolled components, where form data is handled by the DOM itself.
 
-To write an uncontrolled component, instead of writing an event handler for every state update, you can [use a ref](/docs/refs-and-the-dom.html) to get form values from the DOM.
+To write an uncontrolled component, instead of writing an event handler for every state update, you can read form values from the [form](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement#Properties) itself.
 
 For example, this code accepts a single name in an uncontrolled component:
 
-```javascript{8,17}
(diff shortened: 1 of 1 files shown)


content/docs/uncontrolled-components.md
```diff
@@ -6,28 +6,29 @@ permalink: docs/uncontrolled-components.html
...
 class NameForm extends React.Component {
   constructor(props) {
     super(props);
     this.handleSubmit = this.handleSubmit.bind(this);
   }
 
   handleSubmit(event) {
-    alert('A name was submitted: ' + this.input.value);
     event.preventDefault();
+    const name = event.target.elements.name.value;
```
everdimension: Perhaps this can be expanded to a couple of lines to make it clearer for everyone:

```javascript
const form = event.target;
const nameInput = form.elements.name;
const name = nameInput.value;
```

BartoszKlonowski:
Hello @everdimension, thank you for the pull request!
Unfortunately this is pretty outdated now as the latest docs do not have that page and those examples you modify.
If you would like to focus on some topics please open new PR, or take one of open issues.
Thank you!

PR #317, opened 2017-11-21
Adding requestAnimationFrame to rate limiting section

### LINKED ISSUE #312: Adding requestAnimation throttling
This is a great new resource: https://reactjs.org/docs/faq-functions.html

[This section](https://reactjs.org/docs/faq-functions.html#how-can-i-prevent-a-function-from-being-called-too-quickly-or-too-many-times-in-a-row) states that people can use `throttle` or `debounce` to reduce the amount of times a function is called. There is another alternative that we have been getting great results from: `requestAnimationFrame` throttling:

https://github.com/alexreardon/raf-schd

Rather than sampling on a time basis - you sample based on `requestAnimationFrame`. This lets the browser decide how many times to actually call the function to optimise for render performance.

```js
import rafSchedule from 'raf-schd';

class ScrollListener extends React.Component {
  constructor(props) {
    super(props);
    this.nextFrameId = null;
    this.handleScroll = this.handleScroll.bind(this);
    this.scheduleUpdate = rafSchedule(this.emitUpdate.bind(this));
  }

  handleScroll(e) {
    this.nextFrameId = this.scheduleUpdate({ x: e.clientX, y: e.clientY });
  }

  emitUpdate(point) {
     this.props.update(point);
  }

  componentWillUnmount() {
    cancelAnimationFrame(this.nextFrameId);
  }

  render() {
    return (
      <textarea
        onScroll={this.handleScroll}
      />
    );
  }
}
```

I would be happy to add the technique to the docs if you think it worth while.

Cheers

alexreardon:
Closes #312

### DIFF
--- a/content/docs/faq-functions.md
+++ b/content/docs/faq-functions.md
@@ -304,9 +304,11 @@ class Searchbox extends React.Component {
 
 #### `requestAnimationFrame` throttling
 
-[`requestAnimationFrame`](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame) is a way of queuing a function to be executed in the browser at the optimal time for rendering performance. A function that is queued with `requestAnimationFrame` will fire in the next frame. The browser will work hard to ensure that there are `60` frames per second (`60 fps`). However, if the browser is unable to it will naturally *limit* the amount of frames in a second. For example, a device might only be able to handle `30 fps` and so you will only get `30` frames in that second. Using `requestAnimationFrame` for throttling is a useful technique in that it prevents you from doing more than `60 updates` in a second. If you are doing `100` updates in a second this creates additional work for the browser that the user will not see anyway.
+[`requestAnimationFrame`](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame) is a way of queuing a function to be executed in the browser at the optimal time for rendering performance. A function that is queued with `requestAnimationFrame` will fire in the next frame. The browser will work hard to ensure that there are 60 frames per second (60 fps). However, if the browser is unable to it will naturally *limit* the amount of frames in a second. For example, a device might only be able to handle 30 fps and so you will only get 30 frames in that second. Using `requestAnimationFrame` for throttling is a useful technique in that it prevents you from doing more than 60 updates in a second. If you are doing 100 updates in a second this creates additional work for the browser that the user will not see anyway.
 
-> Using this technique will only capture the last published value in a frame. You can see an example of how this optimization works on [`MDN`](https://developer.mozilla.org/en-US/docs/Web/Events/scroll)
+>**Note:**
+>
+>Using this technique will only capture the last published value in a frame. You can see an example of how this optimization works on [`MDN`](https://developer.mozilla.org/en-US/docs/Web/Events/scroll)
 
 ```jsx
 import rafSchedule from 'raf-schd';
@@ -314,31 +316,27 @@ import rafSchedule from 'raf-schd';
 class ScrollListener extends React.Component {
   constructor(props) {
     super(props);
+
     this.handleScroll = this.handleScroll.bind(this);
 
-    // create a new function that will schedule updates
-    this.scheduleUpdate = rafSchedule(this.emitUpdate.bind(this));
+    // Create a new function to schedule updates.
+    this.scheduleUpdate = rafSchedule(
+      point => this.props.onScroll(point)
+    );
   }
 
   handleScroll(e) {
-    // When we receive a scroll event, we are going to schedule an update
-    // If we receive many updates within a single frame duration we will
-    // only be publishing the latest value.
-
+    // When we receive a scroll event, schedule an update.
+    // If we receive many updates within a frame, we'll only publish the latest value.
     this.scheduleUpdate({ x: e.clientX, y: e.clientY });
   }
 
-  emitUpdate(point) {
-    this.props.onScroll(point);
-  }
-
   componentWillUnmount() {
-    // As we are unmounted we can clear the next animation frame
+    // Cancel any pending updates since we're unmounting.
     this.scheduleUpdate.cancel();
   }
 
   render() {
-    // Creating a scroll container around a big image
     return (
       <div
         style={{ overflow: 'scroll' }}
--- a/content/docs/faq-functions.md
+++ b/content/docs/faq-functions.md
@@ -222,15 +222,15 @@ class Alphabet extends React.Component {
 
 If you have an event handler such as `onClick` or `onScroll` and want to prevent the callback from being fired too quickly, then you can limit the rate at which callback is executed. This can be done by using:
 
-- **throttling**: sample changes based on a time based frequency ([`_.throttle`](https://lodash.com/docs#throttle))
-- **debouncing**: publish changes after a period of inactivity ([`_.debounce`](https://lodash.com/docs#debounce))
-- **`requestAnimationFrame` throttling**: sample changes based on [`requestAnimationFrame`](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame) ([`raf-schd`](https://github.com/alexreardon/raf-schd))
+- **throttling**: sample changes based on a time based frequency (eg [`_.throttle`](https://lodash.com/docs#throttle))
+- **debouncing**: publish changes after a period of inactivity (eg [`_.debounce`](https://lodash.com/docs#debounce))
+- **`requestAnimationFrame` throttling**: sample changes based on [`requestAnimationFrame`](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame) (eg [`raf-schd`](https://github.com/alexreardon/raf-schd))
 
-See [this visualization](http://demo.nimius.net/debounce_throttle/) for a comparison of `throttle` and `debounce` functions
+See [this visualization](http://demo.nimius.net/debounce_throttle/) for a comparison of `throttle` and `debounce` functions.
 
 > Note:
 >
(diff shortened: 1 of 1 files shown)


content/docs/faq-functions.md
```diff
@@ -200,31 +200,37 @@ class Alphabet extends React.Component {
...
+- **debouncing**: publish changes after a period of inactivity ([`_.debounce`](https://lodash.com/docs#debounce))
+- **`requestAnimationFrame` throttling**: sample changes based on [`requestAnimationFrame`](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame) ([`raf-schd`](https://github.com/alexreardon/raf-schd))
+
+See [this visualization](http://demo.nimius.net/debounce_throttle/) for a comparison of `throttle` and `debounce` functions
 
 > Note:
 >
-> Both `_.debounce` and `_.throttle` provide a `cancel` method to cancel delayed callbacks. You should either call this method from `componentWillUnmount` _or_ check to ensure that the component is still mounted within the delayed function.
+> `_.debounce` and `_.throttle` provide a `cancel` method to cancel delayed callbacks. You should either call this method from `componentWillUnmount` _or_ check to ensure that the component is still mounted within the delayed function. Similarly, `raf-schd` returns a `frameId` that you can use to cancel the next animation frame when the component unmounts with `cancelAnimationFrame`
```
bvaughn: I've used [`raf`](https://www.npmjs.com/package/raf) several times before. It also implements a `.cancel` method (similar to Lodash debounce and throttle).

Maybe we could change this new section to something like:

> Similarly, NPM modules like `raf` and `raf-schd` provide mechanisms to cancel a callback scheduled for the next animation frame.

content/docs/faq-functions.md
```diff
@@ -200,31 +200,37 @@ class Alphabet extends React.Component {
...
 ```
 
 ### How can I prevent a function from being called too quickly or too many times in a row?
 
-If you have an event handler such as `onClick` or `onScroll` and want to prevent the callback from being fired too quickly, you can wrap the handler with a utility such as [`_.debounce`](https://lodash.com/docs#debounce) or [`_.throttle`](https://lodash.com/docs#throttle). See [this visualization](http://demo.nimius.net/debounce_throttle/) for a comparison of the two.
+If you have an event handler such as `onClick` or `onScroll` and want to prevent the callback from being fired too quickly, then you can limit the rate at which callback is executed. This can be done by using:
+
+- **throttling**: sample changes based on a time based frequency ([`_.throttle`](https://lodash.com/docs#throttle))
+- **debouncing**: publish changes after a period of inactivity ([`_.debounce`](https://lodash.com/docs#debounce))
+- **`requestAnimationFrame` throttling**: sample changes based on [`requestAnimationFrame`](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame) ([`raf-schd`](https://github.com/alexreardon/raf-schd))
```
bvaughn: I like this change ^ 😄 

Maybe we could improve clarity though that the NPM modules we link to are only _one_ option (not the only one). Maybe something like:

- **Debouncing**: Publish changes after a period of inactivity (eg [`_.debounce`](https://lodash.com/docs#debounce))
- **Timer-based throttling**: Sample changes based on a time based frequency (eg [`_.throttle`](https://lodash.com/docs#throttle))
- **`requestAnimationFrame` throttling**: Sample changes based on [`requestAnimationFrame`](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame) (eg [`raf`](https://www.npmjs.com/package/raf), [`raf-schd`](https://www.npmjs.com/package/raf-schd))

content/docs/faq-functions.md
```diff
@@ -200,31 +200,37 @@ class Alphabet extends React.Component {
...
+
+See [this visualization](http://demo.nimius.net/debounce_throttle/) for a comparison of `throttle` and `debounce` functions
```
bvaughn: nit: Missing ending "."

content/docs/faq-functions.md
```diff
@@ -295,3 +301,58 @@ class Searchbox extends React.Component {
   }
 }
 ```
+
+#### `requestAnimationFrame` throttling
+
+[`requestAnimationFrame`](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame) is a way of queuing a function to be executed in the browser at the optimal time for rendering performance. It is very similar to `setTimeout` except it takes no time argument. A function that is queued with `requestAnimationFrame` will fire in the next frame. The browser will work hard to ensure that there are `60` frames per second (`60 fps`). However, if the browser is unable to it will naturally *limit* the amount of frames in a second. For example, a device might only be able to handle `30 fps` and so you will only get `30` frames in that second. Using `requestAnimationFrame` for throttling is a useful technique in that it prevents you from doing more than `60 updates` in a second. If you are doing `100` updates in a second this creates additional work for the browser that the user will not see anyway.
```
bvaughn: I'm not sure about this sentence:
> It is very similar to `setTimeout` except it takes no time argument.

`requestAnimationFrame` is only similar in that it executes code asynchronously. Maybe we should just remove this sentence to avoid potentially confusing people?

content/docs/faq-functions.md
```diff
@@ -295,3 +301,58 @@ class Searchbox extends React.Component {
...
+
+```jsx
+import rafSchedule from 'raf-schd';
+
+class ScrollListener extends React.Component {
+  constructor(props) {
+    super(props);
+    // keep a record of the id of the next animation frame
+    this.nextFrameId = null;
+    this.handleScroll = this.handleScroll.bind(this);
```
bvaughn: nit: Maybe swap lines 317-318 and 319 so the comment is more clearly only referring to line 318?

content/docs/faq-functions.md
```diff
@@ -295,3 +301,58 @@ class Searchbox extends React.Component {
...
+      <div
+        style={{ overflow: 'scroll' }}
+        onScroll={this.handleScroll}
+      >
+        <img src="/my-huge-image.jpg" />
+      </div>
+    );
+  }
+}
+```
```
bvaughn: I think it might be better for us to use `raf` in this example instead of `raf-schd`.

`raf-schd` looks nice! But `raf` mirrors the behavior of the standardized `requestAnimationFrame` API, and I think it's a good thing for us to show the more standard approach here to avoid potentially confusing users (eg by mixing `raf-schd` and `cancelAnimationFrame`).

content/docs/faq-functions.md
alexreardon: Sounds good

content/docs/faq-functions.md
alexreardon: `raf` is just a `requestAnimationFrame` polyfill and does not contain the logic for queuing calls in a frame. I included `raf-schd` as it is a similar abstraction to a `throttle` or `debounce` function. I could hand write the `requestAnimationFrame` throttle logic, but that seems strange given that we do not do so for `throttle` or `debounce` in this case

alexreardon: It would look something like this which is a bit messy:

https://github.com/alexreardon/raf-schd#without-raf-schd

content/docs/faq-functions.md
alexreardon: As mentioned below, `raf` is just a polyfill and is not a higher level utility for sampling with `requestAnimationFrame`

content/docs/faq-functions.md
alexreardon: yeah - or i'll see if i can get it a bit clearer

alexreardon:
I can make the updates you suggested. However, before I do it would be good to get clarity around the `raf-schd` questions I put forward. 

Cheers!

content/docs/faq-functions.md
bvaughn: > `raf` is just a `requestAnimationFrame` polyfill

Sure, that's true, but I kind of like it for that reason.

> I could hand write the `requestAnimationFrame` throttle logic, but that seems strange given that we do not do so for `throttle` or `debounce` in this case

That's a good point. I think the thing that nagged me a little bit with `raf-schd` was the mix of _that_ module with the built-in `cancelAnimationFrame`.

bvaughn: I'm curious- did you ever considered adding a `.cancel` property to the `ResultFn` so as to not have to mix the native and non-native calls?

```js
result.cancel = () => {
  cancelAnimationFrame(frameId);
};
```

alexreardon: I did, but that would modify the original function. My thought was to keep it super pure and not modify the wrapped function. However, for consistency with other libs I would consider adding it

alexreardon: I am leaning towards adding it to `raf-schd`. It will then make these examples all read very similar

bvaughn: Wouldn't it just be a property attached to the inline `result` function you created? It wouldn't actually modify the callback a user passed in right? (I agree, that would be bad.)

alexreardon: The original function is not modified - the issue is the flow type. All good, I have been thinking about it for a while. I think having a `.cancel` brings it in line with other utilities. 

Assuming it did add a `.cancel` is there any outstanding issues?

bvaughn: Hm. You said:
> I did, but that would modify the original function.

I was just pointing out that the original function (the `fn` callback) wouldn't be modified? Maybe I just misinterpreted what you said. 😄 

No. I think adding a `.cancel` method would alleviate my concerns.

alexreardon: https://github.com/alexreardon/raf-schd/pull/7

I will update this pr with the updated syntax

content/docs/faq-functions.md
alexreardon: Not needed any more as we do not need to know the frameId

alexreardon:
@bvaughn please take another look and let me know what you think

Cheers

content/docs/faq-functions.md
alexreardon: Now released is `2.1.0`

alexreardon:
Cheers @bvaughn !

PR #319, opened 2017-11-21
Update lifting-state-up.md

### LINKED ISSUE #314: Image has an incorrect aspect ratio
Noticed this in https://reactjs.org/docs/lifting-state-up.html#lessons-learned:

screen shot 2017-11-21 at 00 52 42

Not sure if this is just a single image or if there's a pattern.

Raincal:
Fix #314


gaearon:
Do we have more places where we do this?

gaearon:
Oops, just saw https://github.com/reactjs/reactjs.org/issues/314#issuecomment-345927929. Never mind.

PR #322, opened 2017-11-21
Add mention of custom-elements-es5-adapter

robdodson:
Added a note that including the es5 adapter fixes the issue of transpiled custom elements not working in browsers like Chrome and Safari. I think the previously linked GitHub thread provides a lot of useful background information but might also be confusing for some folks and leave them wondering if custom elements will work at all in their app. Things should be all good after they include the adapter.

### DIFF
--- a/content/docs/web-components.md
+++ b/content/docs/web-components.md
@@ -58,3 +58,4 @@ customElements.define('x-search', XSearch);
 >Note:
 >
 >This code **will not** work if you transform classes with Babel. See [this issue](https://github.com/w3c/webcomponents/issues/587) for the discussion.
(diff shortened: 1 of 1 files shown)


gaearon:
I had no idea this exists. Thanks!

robdodson:
I actually just discovered https://github.com/github/babel-plugin-transform-custom-element-classes which may be a better fit for Babel users. @bvaughn @gaearon would you be ok with me submitting another PR to mention that instead of the custom elements adapter?

gaearon:
Many React users are using CRA which offers no way to add/change Babel plugins. If ES5 adapter "just works" without Babel changes, maybe it would work best? I don't know :-)

robdodson:
Oh good point! Ok then I guess we can leave it as is 😁

PR #333, opened 2017-11-23
fixes #332, remove section related unsupported `react-addons-perf`

### LINKED ISSUE #332: Remove the section about `react-addons-perf`
`react-addons-perf` is no longer supported from React 16 onwards. Still, the main documentation has a page related to it (https://reactjs.org/docs/perf.html) and it might cause confusion to users who can easily overlook the warning on top and start following the code example. 

Relevant link:
- https://stackoverflow.com/questions/46578145/module-not-found-cant-resolve-react-dom-lib-reactperf-in-node-modules-reac

vinitkumar:


gaearon:
Hmm. Now that I see the PR this doesn't make as much sense to me.

This page is unreachable from the main website, is it not? So you can only find it by search. I think it's fair to keep the reference for people still using 15 as long as it's not linked from the main site (which it is not). 

The disclaimer about it being unsupported is also there. Is it too easy to miss? We could make it more prominent.

vinitkumar:
@gaearon Well, you are not wrong. It is reachable only through search engines. I will close this PR then.

gaearon:
What do you think about moving the "note" into its own section with the header saying "Not Supported as of React 16"? So that it's more prominent.

PR #334, opened 2017-11-23
Change Object.assign to spread operator

hermesnetto:


bvaughn:
The spread operator is shown in the section above. I believe the intent of this section is to show a different approach. Combining the spread operator + `delete` feels a bit too contrived. It makes more sense in the context of `Object.assign` though.

I think maybe it's best to the example as-is in this case, so we'll at least show an ES5 / IE-compatible approach. 😄 Thanks though!

PR #336, opened 2017-11-24
[Discussion] Embed animated examples in markdown

bvaughn:
Relates to gatsbyjs/gatsby/pull/3012

This PR shows how we could use the new `gatsby-remark-embed-snippet` plug-in I made to extract some of our more complex examples from markdown to external files that could be formatted by Prettier. The new `gatsby-remark-embed-snippet` plug-in also supports a simpler way of specifying highlighted lines (see [here](https://github.com/bvaughn/reactjs.org/blob/5369ebee11492631e21cd9222de1e5e8ed969830/examples/animation/invalid-unmounted-example.js#L9-L12)) that I'd like to see added to the `gatsby-remark-prismjs` plug-in as well so that we could use this for inline code-comments too. (I've proposed this to Kyle.)

I don't think we should move _all_ (or even most) examples from markdown to the `examples` folder- but I do think it would be nice to be able to do for some of the larger examples, or the ones that we repeat throughout the site in multiple places.

I've temporarily embedded the `gatsby-remark-embed-snippet` plug-in in this PR temporarily (since it hasn't yet been merged to Gatsby). I just updated a single markdown file, `content/docs/addons-animation.md`, for example. You can see the resulting output here: [deploy-preview-336--reactjs.netlify.com/docs/animation.html](https://deploy-preview-336--reactjs.netlify.com/docs/animation.html)

### DIFF
--- a/content/docs/addons-animation.md
+++ b/content/docs/addons-animation.md
@@ -27,48 +27,7 @@ import ReactCSSTransitionGroup from 'react-addons-css-transition-group'; // ES6
 var ReactCSSTransitionGroup = require('react-addons-css-transition-group'); // ES5 with npm
 ```
 
-```javascript{31-36}
-class TodoList extends React.Component {
-  constructor(props) {
-    super(props);
-    this.state = {items: ['hello', 'world', 'click', 'me']};
-    this.handleAdd = this.handleAdd.bind(this);
-  }
-
-  handleAdd() {
-    const newItems = this.state.items.concat([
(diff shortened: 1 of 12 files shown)


BartoszKlonowski:
Hello @bvaughn!
I don't think this is any longer required as the react.dev has it's examples embedded in whole different manner making this PR no longer valid.
Let me close it, but thank you for the work and contributing to the docs!

PR #338, opened 2017-11-24
Add pragmatic advice on file structure

gaearon:
Figured I'd write this

### DIFF
--- a/content/docs/faq-structure.md
+++ b/content/docs/faq-structure.md
@@ -35,7 +35,7 @@ profile/
   ProfileAPI.js
 ```
 
-The definition of what a "feature" is not universal, and it is up to you to choose the granularity. If you can't come up with a list of top-level folders, you can ask the users of your product what major parts it consists of, and use their mental model as a blueprint.
+The definition of a "feature" is not universal, and it is up to you to choose the granularity. If you can't come up with a list of top-level folders, you can ask the users of your product what major parts it consists of, and use their mental model as a blueprint.
 
 #### Grouping by file type
 
--- a/content/docs/faq-structure.md
+++ b/content/docs/faq-structure.md
@@ -8,18 +8,67 @@ category: FAQ
 
 ### Is there a recommended way to structure React projects?
 
+React doesn't have opinions on how you put files into folders. That said there are a few common approaches popular in the ecosystem you may want to consider.
+
+#### Grouping by features or routes
+
 One common way to structure projects is locate CSS, JS, and tests together inside folders grouped by feature or route.
 
 ```
-FeatureA
+common/
+  Avatar.js
+  Avatar.css
+  APIUtils.js
+  APIUtils.test.js
+feed/
   index.js
-  ComponentA.js
-  ComponentA.css
-  ComponentA.test.js
-  Helper.js
-  Helper.test.js
-FeatureB
+  Feed.js
+  Feed.css
+  FeedStory.js
+  FeedStory.test.js
+  FeedAPI.js
+profile/
   index.js
-  ComponentB.js
-  ComponentB.test.js
+  Profile.js
+  ProfileHeader.js
+  ProfileHeader.css
+  ProfileAPI.js
 ```
+
+The definition of what a "feature" is not universal, and it is up to you to choose the granularity. If you can't come up with a list of top-level folders, you can ask the users of your product what major parts it consists of, and use their mental model as a blueprint.
+
+#### Grouping by file type
+
+Another popular way to structure projects is to group similar files together, for example:
+
+```
+api/
+  APIUtils.js
+  APIUtils.test.js
+  ProfileAPI.js
+  UserAPI.js
+components/
+  Avatar.js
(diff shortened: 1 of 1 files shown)


content/docs/faq-structure.md
```diff
@@ -8,18 +8,67 @@ category: FAQ
...
+React doesn't have opinions on how you put files into folders. That said there are a few common approaches popular in the ecosystem you may want to consider.
+
+#### Grouping by features or routes
+
 One common way to structure projects is locate CSS, JS, and tests together inside folders grouped by feature or route.
 
 ```
-FeatureA
+common/
+  Avatar.js
```
kentcdodds: Any chance I could sway things from PascalCase to kebab-case? I've seen many people (especially beginners) have issues with [case-sensitivity across operating systems](http://www.informit.com/articles/article.aspx?p=1144082&seqNum=4) and [with git](https://stackoverflow.com/questions/17683458/how-do-i-commit-case-sensitive-only-filename-changes-in-git). I've found it best to avoid these issues altogether by encouraging all lower case and using kebab-case to separate words. I'd love it if the recommended approach in React's docs helped encourage this as well to avoid casing issues for people following the suggestion.

Ethan-Arrowood:
Gave it a read through and it all sounds good to me! Tried looking for big spelling/grammar mistakes too and nothing popped out. 👍

content/docs/faq-structure.md
gaearon: I think that's going to be a hard sell for most people since it doesn't match component names.

Ethan-Arrowood: Interesting! I've haven't ran into case-sensitivity issues in my own development yet, but this does sound like a very real concern. I would support using kebab-case, and maybe even explaining why it is recommended over PascalCase.

content/docs/faq-structure.md
```diff
@@ -8,18 +8,67 @@ category: FAQ
...
   index.js
-  ComponentB.js
-  ComponentB.test.js
+  Profile.js
+  ProfileHeader.js
+  ProfileHeader.css
+  ProfileAPI.js
+```
+
+The definition of what a "feature" is not universal, and it is up to you to choose the granularity. If you can't come up with a list of top-level folders, you can ask the users of your product what major parts it consists of, and use their mental model as a blueprint.
```
paulkoegel-wessels: There's an "is" missing here -> "The definition of what a feature **is** **is** not universal"

mxstbr:
LGTM :100:

content/docs/faq-structure.md
nickdima: What will `index.js` consist of in this case?

olingern: > The definition of what a "feature" is not universal

The definition of what a "feature" **_can be_** is not universal ?

content/docs/faq-structure.md
olingern: I think kebab-case is a little less intuitive on how to import without opening the file up, i.e.
is `my-component` imported as `MyComponent`, `myComponent`, or `mycomponent`

I think if you maintain the casing and the component name to match, it makes it easier to not constantly reference the file.

erikras: Kebab-case is for npm packages. PascalCase for local components. My 0.02€.

content/docs/faq-structure.md
```diff
@@ -8,18 +8,67 @@ category: FAQ
...
+  Profile.js
+  ProfileHeader.js
+  ProfileHeader.css
 ```
+
+Some people also prefer to go further, and separate components into different folders depending on their role in the application. For example, [Atomic Design](http://bradfrost.com/blog/post/atomic-web-design/) is a design methodology built on this principle. Remember that it's often more productive to treat such methodologies as helpful examples rather than strict rules to follow.
+
+#### Avoid too much nesting
+
+There are many pain points associated with deep directory nesting in JavaScript projects. It becomes harder to write relative imports between them, or to update those imports when the files are moved. Unless you have a very compelling reason to use a deep folder structure, consider limiting yourself to a maximum of three or four nested folders within a single project. Of course, this is only a recommendation, and it may not be relevant to your project.
```
damassi: What about for those using absolute paths?

gaearon: That's exactly the kind of detail I don't want to get into :-) If you know how to set this up and understand the tradeoffs of using symlinks vs non-standard setup that doesn't work with other tooling, you already know more than this section

content/docs/faq-structure.md
KyleAMathews: I fairly regularly see Gatsby issues where people can't deploy from linux CI because of casing issues not discovered on their mac.

amackintosh: @kentcdodds One viable alternative you could do is encourage ES Lint.

We won't get into the savagery of `eslint-config-airbnb`, but I think you could recommend `import/no-unresolved` which throws error when the import declaration doesn't match the underlying filesystem.

PR #341, opened 2017-11-25
Added FAQ for Syncing State and props

jyash97:
Added about usage of State inside FAQ.

### DIFF
--- a/content/docs/faq-state.md
+++ b/content/docs/faq-state.md
@@ -6,6 +6,18 @@ layout: docs
 category: FAQ
 ---
 
+### What Should Go in State?
+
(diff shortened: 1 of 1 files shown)


content/docs/faq-state.md
```diff
@@ -6,6 +6,18 @@ layout: docs
 category: FAQ
 ---
```
alexkrolick: Adding a quick review here to get the ball rolling, sorry if it's lacking details:

- "What Should Go in State?" is a good way to frame this topic
- "What does setState do?" should remain at the top to give context to the other sections
- Might make sense to add "What is the difference between state and props" with a link to main docs on those topics, while explaining that both cause a re-render when they change. Some of the content below could be pulled into this section to focus more on the subjects of computed values and syncing props and state.

jyash97: Thanks for the review. Let me come up with a updated version and inform you. 
@alexkrolick

jyash97:
@alexkrolick Please have a look at the changes.

PR #344, opened 2017-11-26
Updated FAQ - Discuss CSS-in-JS tradeoffs

jyash97:
Added a note that CSS-in-JS is not mandatory adn How to import css files.

### DIFF
--- a/content/docs/faq-styling.md
+++ b/content/docs/faq-styling.md
@@ -42,7 +42,7 @@ CSS classes are generally more efficient than inline styles.
 
 "CSS-in-JS" refers to a pattern where CSS is composed using JavaScript instead of defined in external files. Read a comparison of CSS-in-JS libraries [here](https://github.com/MicheleBertoli/css-in-js).
 
-_Note that this functionality is not a part of React, but provided by third-party libraries._ React does not have an opinion about how styles are defined; if in doubt, define your styles in a separate `*.css` file as usual and refer to them using [`className`](/docs/dom-elements.html#classname).
+_Note that this functionality is not a part of React, but provided by third-party libraries._ React does not have an opinion about how styles are defined; if in doubt, a good starting point is to define your styles in a separate `*.css` file as usual and refer to them using [`className`](/docs/dom-elements.html#classname).
 
 ### Can I do animations in React?
 
--- a/content/docs/faq-styling.md
+++ b/content/docs/faq-styling.md
@@ -40,9 +40,7 @@ CSS classes are generally more efficient than inline styles.
 
 ### What is CSS-in-JS?
 
-CSS-in-JS refers to a pattern where CSS is written with Javascript, then extracted into a stylesheet.
-
-[Comparison of CSS-in-JS Libraries](https://github.com/MicheleBertoli/css-in-js)
+"CSS-in-JS" refers to a pattern where CSS is composed using JavaScript instead of defined in external files. Read a comparison of CSS-in-JS libraries [here](https://github.com/MicheleBertoli/css-in-js).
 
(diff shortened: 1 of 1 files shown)


content/docs/faq-styling.md
```diff
@@ -8,6 +8,12 @@ category: FAQ
 
 ### How do I add CSS classes to components?
 
+Import the CSS file in which the styles are scoped:
+
+```jsx
+import from './path/to/Filename.css';
```
alexkrolick: This is doesn't work out of the box, it requires special loader for Webpack. Overall, it's probably best not to entangle Webpack in the FAQ unless absolutely necessary. I think mentioning importing CSS with `css-loader` as an alternative to CSS-in-JS below, however, would be appropriate, maybe mentioning CSS modules and their benefits (like being able to use SASS/LESS/PostCSS directly, interop with CSS tooling, etc.)

jyash97: Yes we can add Css modules instead.
There's a article on styling in React.May be we can add this link I found it useful:
https://codeburst.io/4-four-ways-to-style-react-components-ac6f323da822

content/docs/faq-styling.md
```diff
@@ -40,7 +46,7 @@ CSS classes are generally more efficient than inline styles.
 
 ### What is CSS-in-JS?
 
-CSS-in-JS refers to a pattern where CSS is written with Javascript, then extracted into a stylesheet.
+CSS-in-JS refers to a pattern where CSS is written with Javascript, then extracted into a stylesheet.It is `not mandatory` to use CSS-in-JS.
```
bvaughn: Nit: Missing a spacer in between "stylesheet.It" Also the use of back-ticks for emphasis doesn't fit with our formatting elsewhere.

This sentence feels a bit incomplete. What do you think about expanding it slightly?

> CSS-in-JS refers to a pattern where CSS is written with Javascript, then extracted into a stylesheet.

> It is _not mandatory_ to use CSS-in-JS. You may define your styles in separate `*.css` files or inline (using the `style` attribute) if you prefer.

gaearon: Even "it is not mandatory" sounds too light to me, as if it implies we still recommend it (but don't force people to use it). Whereas I don't even think we recommend it.

jyash97: Yes I too feel that we need to expand it slightly. Let me come up with newer version.@gaearon @bvaughn

jyash97:
Hey @bvaughn @gaearon @alexkrolick, Check this commit I have done the required changes

alexkrolick:
Thanks for the update @jyash97.

What do you think of adding some more context to the explanation? Something like:

---

### What is CSS-in-JS?

"CSS-in-JS" refers to a pattern where CSS is composed using JavaScript instead of defined in external files. Read a comparison of CSS-in-JS libraries [here](https://github.com/MicheleBertoli/css-in-js).

_Note that this functionality is not part of React, but provided by third-party libraries._ React does not have an opinion about how styles are defined; if in doubt, plain stylesheets referenced using `className` are a good place to start.

jyash97:
Yes we can add something like this, it would be more appropriate. I think we should also add inline styles usage as we have a part mentioning about inline styles in our document.

@alexkrolick
We must make a note wherever we have used style so as not to confuse readers.

alexkrolick:
> We must make a note wherever we have used style so as not to confuse readers.

https://github.com/reactjs/reactjs.org/pull/346

jyash97:
That's great.Let me update this commit also

jyash97:
Hey @alexkrolick see this changes

jyash97:
@alexkrolick Thanks for the changes and the help.

content/docs/faq-styling.md
bvaughn: Good point Dan!

I like the wording that you ended up with, @jyash97. Great iteration.

PR #345, opened 2017-11-27
Add release jsx fragment blog post

clemmy:
First draft of the blog post for the upcoming React v16.2 release.

Todos:

- [x] Ensure CRA release link is correct
- [ ] Add a CodePen link for live demo

**Do not merge yet!**

### DIFF
--- a/content/blog/2017-11-28-fragments-in-react-v16.2.md
+++ b/content/blog/2017-11-28-fragments-in-react-v16.2.md
@@ -3,201 +3,308 @@ title: "Fragments in React v16.2"
 author: [clemmy]
 ---
 
-Today, we're excited to improved support for fragments in React! We are introducing a new `Fragment` export, which allows
-which allows users to render *static* fragments using familiar JSX element syntax.
+React 16.2 is now available! The biggest addition is improved support for returning multiple children from a component's render method. We call this feature *fragments*:
+
+Fragments look like empty JSX tags. They let you group a list of children without adding extra nodes to the DOM:
+
+```js
+render() {
+  return (
+    <>
+      <ChildA />
+      <ChildB />
+      <ChildC />
+    </>
+  );
+}
+```
+
+## What Are Fragments?
+
+A common pattern is for a component to return a list of children. An example is rendering some text with a link inside. Take this example HTML:
+
+```html
+Text with
+<a href="/foo">multiple</a>
+<a href="/bar">links</a>
+inside it.
+```
+
+Prior to version 16, the only way to acheive this in React was by wrapping the children in an extra element, usually a `div` or `span`:
+
+```js
+render() {
+  return (
+    // Extraneous span element :(
+    <span>
+      Text with
+      <a href="/foo">multiple</a>
+      <a href="/bar">links</a>
+      inside it.
+    </span>
+  );
+}
+```
+
+To address this limitation, React 16.0 added support for [returning an array of elements from a component's `render` method](https://reactjs.org/blog/2017/09/26/react-v16.0.html#new-render-return-types-fragments-and-strings). Instead of wrapping the children in a DOM element, you can put them into an array:
 
 ```jsx
-import React, { Fragment } from 'react';
+render() {
+ return [
+    "Text with",
+    <a key="link1" href="/foo">multiple</a>,
+    <a key="link2" href="/bar">links</a>,
+    "inside it."
+ ];
+}
+```
 
+However, this has some confusing differences from normal JSX:
+
+- Children in an array must be separated by commas.
+- Children in an array must have a key to prevent React's [key warning](https://reactjs.org/docs/lists-and-keys.html#keys).
+- Strings must be wrapped in quotes.
+
+To provide a more consistent authoring experience for fragments, React now provides a first-class `Fragment` component that can be used in place of arrays.
+
+```jsx{3,8}
 render() {
   return (
     <Fragment>
-      <li>Look ma</li>
-      <li>a list!</li>
+      Text with
+      <a href="/foo">multiple</a>
+      <a href="/bar">links</a>
+      inside it.
     </Fragment>
   );
 }
 ```
 
-We've also included a new short syntax that is already supported by [some tools](#tooling-support):
+You can use `<Fragment />` the same way you'd use any other element, without changing the way you write JSX. No commas, no keys, no quotes.
 
-```jsx
+The Fragment component is available on the main React object:
+
+```js
+const Fragment = React.Fragment;
+
+<Fragment>
+  <ChildA />
+  <ChildB />
+  <ChildC />
+</Fragment>
+
+// This also works
+<React.Fragment>
+  <ChildA />
+  <ChildB />
+  <ChildC />
+</React.Fragment>
+```
+
+## JSX Fragment Syntax
+
+Fragments are a common pattern in our codebases at Facebook. We anticipate they'll be widely adopted by other teams, too. To make the authoring experience as convenient as possible, we're adding syntactical support for fragments to JSX:
+
+```jsx{3,8}
 render() {
   return (
     <>
-      <li>Look ma</li>
-      <li>a list!</li>
+      Text with
+      <a href="/foo">multiple</a>
+      <a href="/bar">links</a>
+      inside it.
     </>
   );
 }
 ```
 
-### Why Fragments?
+In React, this desugars to a `<React.Fragment/>` element, as in the example from the previous section. (Non-React frameworks that use JSX may compile to something different.)
 
-React 16 added the [ability to return an array of elements from a component's `render` method](https://reactjs.org/blog/2017/09/26/react-v16.0.html#new-render-return-types-fragments-and-strings). This allowed developers to prevent extraneous markup with `div`s that pollute the DOM. However, it was inconvenient for users who were writing static JSX because the syntax is completely different from how they would usually write JSX:
+Fragment syntax in JSX was inspired by prior art such as the `XMLList() <></>` constructor in [E4X](https://developer.mozilla.org/en-US/docs/Archive/Web/E4X/E4X_for_templating). Using a pair of empty tags is meant to represent the idea it won't add an actual element to the DOM.
 
-```jsx
-render() {
-  // keys and commas are needed when returning an array in a render
-  return [
-    <span key="first-span">I am</span>,
-    <span key="second-span">static</span>
-  ];
-}
-```
+### Keyed Fragments
 
-So, we added the `Fragment` component to React! `Fragment`s let developers return multiple elements from `render` methods using a familiar syntax that resembles normal JSX elements:
+Note that the `<></>` syntax does not accept attributes, including keys. 
+
+If you need a keyed fragment, you can use `<React.Fragment />` directly. An use case for this is mapping a collection to an array of fragments -- for example, to create a description list:
 
 ```jsx
-import React, { Fragment } from 'react';
-
-render() {
-  // keys and commas are no longer needed
-  return <Fragment>
-    <span>I am</span>
-    <span>static</span>
-  </Fragment>;
-}
-```
-
-We anticipate that the above will be a common use case, so we added a shorthand syntax for it:
-
-```jsx
-render() {
+function Glossary(props) {
   return (
-    <>
-      <span>I am</span>
-      <span>static</span>
-    </>
+    <dl>
+      {props.items.map(item => (
+        // Without the `key`, React will fire a key warning
+        <Fragment key={item.id}>
+          <dt>{item.term}</li>
+          <dd>{item.description}</li>
+        </Fragment>
+      )}
+    </dl>
   );
 }
 ```
 
-#### The Syntax
-
-Inspired by some prior art such as the `XMLList() <></>` constructor in [E4X](https://developer.mozilla.org/en-US/docs/Archive/Web/E4X/E4X_for_templating) and the existing `<></>` syntax in [Reason React](https://reasonml.github.io/reason-react/), the React team concluded on having a convenient syntax for JSX fragments that looks like `<>...</>`. This captures the idea of a fragment: It is a pair of empty tags which enforces the notion that it won't render to an actual element in the DOM.
-
-The `<></>` syntax also makes it easier to refactor since it looks just like a JSX element. For example, consider swapping the parents in the following code snippets:
-
-```jsx
-// with a parent wrapper
-<div>
-  <span>foo</span>
-  bar
-  <span>baz</span>
-</div>
-
-// with arrays (where commas are necessary and text needs to be surrounded by quotes)
-[
-  <span>foo</span>,
-  'bar',
-  <span>baz</span>
-]
-
-// with jsx fragment syntax
-<>
-  <span>foo</span>
-  bar
-  <span>baz</span>
-</>
-```
-
-#### The Export
-
-After the release of React 16, people in the React community have created solutions in order to address the problem of static lists, such as [react-aux](https://github.com/gajus/react-aux). For `Fragment`s however, the React team has decided to take ownership of the export, since there are some semantics specific to Fragments that aren't possible without core library changes. These changes have some performance wins and make reconciliation a bit more intuitive.
-
-[For the curious, you can read more about that here.](https://github.com/facebook/react/pull/10783)
-
-We are very thankful to [Gajus Kuizinas](https://github.com/gajus/) and other contributors who prototyped this in open source.
-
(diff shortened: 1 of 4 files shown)


content/blog/2017-11-28-fragments-in-react-v16.2.md
```diff
@@ -0,0 +1,180 @@
...
+```
+yarn upgrade @babel/core @babel/preset-react # for yarn users
+npm update @babel/core @babel/preset-react # for npm users
+```
+
+If you are using Babel with [Webpack](https://webpack.js.org/), then no additional steps need to be done because [babel-loader](https://github.com/babel/babel-loader) will be using your peer-installed version of Babel.
+
+Unfortunately, support for Babel 6.x is not available, and there are currently no plans to back-port.
+
+#### Typescript
```
RyanCavanaugh: "Type**S**cript"

DanielRosenwasser: And the same for the below too

content/blog/2017-11-28-fragments-in-react-v16.2.md
```diff
@@ -0,0 +1,180 @@
...
+For Typescript users, please upgrade to [version 2.6.2](https://github.com/Microsoft/TypeScript/releases/tag/v2.6.2). Note that this is important even if you are already on version 2.6.1, since support was added in 2.6.2.
+
+Upgrade to the latest with the command
+
+```
+yarn upgrade typescript # for yarn users
+npm update typescript # for npm users
+```
+
+Also, editors may complain about errors in `.tsx` and `.js/.jsx` files using the new syntax. These errors can be safely ignored, or configured for [VS Code](https://code.visualstudio.com/Docs/languages/typescript#_using-newer-typescript-versions) and [Sublime Text](https://github.com/Microsoft/TypeScript-Sublime-Plugin/#note-using-different-versions-of-typescript).
```
DanielRosenwasser: How about

```text
These errors can be safely ignored, but editor support for JSX Fragments are already available with updates for [Visual Studio 2015](https://www.microsoft.com/en-us/download/details.aspx?id=48593), [Visual Studio 2017](https://www.microsoft.com/en-us/download/details.aspx?id=55258), and [Sublime Text via Package Control](https://packagecontrol.io/packages/TypeScript).
Visual Studio Code will be updated soon, but [can be configured to use TypeScript 2.6.2 and later](https://code.visualstudio.com/Docs/languages/typescript#_using-newer-typescript-versions).
```

content/blog/2017-11-28-fragments-in-react-v16.2.md
```diff
@@ -0,0 +1,181 @@
...
+      <li>Look ma</li>
+      <li>no keys!</li>
+    </>
+  );
+}
+```
+
+### Why Fragments?
+
+When React released version 16, the ability to render an array of elements from a component's `render` method was added. This allowed developers to prevent extraneous markup with `div`s that pollute the HTML. However, it was inconvenient because every element in the array had to be explicitly keyed. Previously, in order to render an array of elements, the following code would be necessary:
```
gaearon: >When React released version 16, the ability to render an array of elements from a component's `render` method was added.

This reads a bit clumsy to me (too much passive voice!)

Maybe:

>The React 16 release included the new ability to return an array from a component's `render` method.

?

content/blog/2017-11-28-fragments-in-react-v16.2.md
```diff
@@ -0,0 +1,181 @@
...
+```jsx
+render() {
+  return [
+    <span key="first-span">I am</span>,
+    <span key="second-span">static</span>
+  ];
+}
+```
+
+However, the keys are only necessary for React's reconciliation algorithm when child elements are re-arranged and React needs to figure out which item in one render corresponds to in a subsequent render. If the array is static, then keys aren't necessary and only add extra noise to the code. So, let's omit them:
```
gaearon: Let's be consistent about "rearrange" vs "re-arrange". I'd pick "rearrange" everywhere.

content/blog/2017-11-28-fragments-in-react-v16.2.md
```diff
@@ -0,0 +1,181 @@
+---
+title: "Fragments in React v16.2"
+author: [clemmy]
+---
+
+Today, we're excited to bring the new `Fragment` export to React, which allows users to render **static** fragments without the unnecessary keying of each element in the fragment!
```
gaearon: This paragraph is slightly confusing because it doesn't mention the new *syntax* but shows the new syntax. Maybe:

>Today, we’re excited to bring the new `Fragment` export to React, which allows users to render **static** fragments without the unnecessary keying of each element in the fragment!
>
>```js
>render() {
>  return (
>    <React.Fragment>
>      <li>Look ma</li>
>      <li>no keys!</li>
>    </React.Fragment>
>  );
>}
>```
>
>There is also a new short syntax for it that is already supported by [some tools](#how-to-use-it):
>
>```js
>render() {
>  return (
>    <>
>      <li>Look ma</li>
>      <li>no keys!</li>
>    </>
>  );
>}
>```

content/blog/2017-11-28-fragments-in-react-v16.2.md
```diff
@@ -0,0 +1,181 @@
...
+As mentioned above, the most common way to use fragments in React is to simply use `<></>` tags. Generally, in the Javascript compiler phase of your build pipeline, the `<>...</>` will be de-sugared into a `createElement` call with `React.Fragment` as the type.
+
+Take a look at the section on [How to Use it](#how-to-use-it) to see how to update your tooling to support JSX fragments.
+
+#### With Attributes and Keys
+
+Note that with the `<></>` syntax, attributes are not allowed to be passed. We made the decision to disallow attributes with fragments since they're semantically different from React elements. However, a fragment can be keyed by using the explicit fragment syntax:
+
+```jsx
+import { Fragment } from 'react';
```
gaearon: Can we remove this and just use `React.Fragment`? Thinking about ES6 imports here adds unnecessary overhead to an already new concept.

content/blog/2017-11-28-fragments-in-react-v16.2.md
```diff
@@ -0,0 +1,181 @@
...
+
+### Example Usages
+
+As mentioned above, the most common way to use fragments in React is to simply use `<></>` tags. Generally, in the Javascript compiler phase of your build pipeline, the `<>...</>` will be de-sugared into a `createElement` call with `React.Fragment` as the type.
+
+Take a look at the section on [How to Use it](#how-to-use-it) to see how to update your tooling to support JSX fragments.
+
+#### With Attributes and Keys
+
+Note that with the `<></>` syntax, attributes are not allowed to be passed. We made the decision to disallow attributes with fragments since they're semantically different from React elements. However, a fragment can be keyed by using the explicit fragment syntax:
```
gaearon: >Note that with the `<></>` syntax, attributes are not allowed to be passed.

Maybe:

>Note that the `<></>` syntax does not allow attributes on fragments.

content/blog/2017-11-28-fragments-in-react-v16.2.md
```diff
@@ -0,0 +1,181 @@
...
+
+After the release of React 16, people in the React community have created solutions in order to address the problem of static lists, such as [react-aux](https://github.com/gajus/react-aux). For `Fragment`s however, the React team has decided to take ownership of the export, since there are some semantics specific to Fragments that aren't possible without core library changes. These changes have some performance wins and make reconciliation a bit more intuitive.
+
+[For the curious, you can read more about that here.](https://github.com/facebook/react/pull/10783)
+
+Still, we are very thankful to the open source contributors who worked on solving this in the ecosystem.
+
+Additionally, there is a [pragma option available in babel-plugin-transform-react-jsx](https://github.com/babel/babel/tree/master/packages/babel-plugin-transform-react-jsx#pragmafrag) that allows the Babel compiler to de-sugar the `<>...</>` syntax to a custom identifier.
+
+### Example Usages
```
gaearon: Maybe just "Examples"?

content/blog/2017-11-28-fragments-in-react-v16.2.md
```diff
@@ -0,0 +1,181 @@
...
+
+#### The Export
+
+After the release of React 16, people in the React community have created solutions in order to address the problem of static lists, such as [react-aux](https://github.com/gajus/react-aux). For `Fragment`s however, the React team has decided to take ownership of the export, since there are some semantics specific to Fragments that aren't possible without core library changes. These changes have some performance wins and make reconciliation a bit more intuitive.
+
+[For the curious, you can read more about that here.](https://github.com/facebook/react/pull/10783)
+
+Still, we are very thankful to the open source contributors who worked on solving this in the ecosystem.
+
+Additionally, there is a [pragma option available in babel-plugin-transform-react-jsx](https://github.com/babel/babel/tree/master/packages/babel-plugin-transform-react-jsx#pragmafrag) that allows the Babel compiler to de-sugar the `<>...</>` syntax to a custom identifier.
```
gaearon: Maybe replace "Additionally" with "If you use JSX with something other than React, " to better explain the context of this sentence.

content/blog/2017-11-28-fragments-in-react-v16.2.md
```diff
@@ -0,0 +1,181 @@
...
+
+### Example Usages
+
+As mentioned above, the most common way to use fragments in React is to simply use `<></>` tags. Generally, in the Javascript compiler phase of your build pipeline, the `<>...</>` will be de-sugared into a `createElement` call with `React.Fragment` as the type.
+
+Take a look at the section on [How to Use it](#how-to-use-it) to see how to update your tooling to support JSX fragments.
```
gaearon: It is pretty confusing to me that this section is called "Example Usages" but there are no examples in it at all, and it asks me to follow to a different section. This breaks the flow. Can we avoid the jumps forward, and also rename this section if it doesn't actually contain examples?

content/blog/2017-11-28-fragments-in-react-v16.2.md
```diff
@@ -0,0 +1,181 @@
...
+
+In the future, `Fragment` may also support more attributes, such as event handlers.
+
+### Live Demo
+
+You can experiment with JSX fragment syntax on this [CodePen](#) using [React v16.2.0](https://github.com/facebook/react/releases/tag/v16.2.0) and [Babel v7](https://github.com/babel/babel).
+
+### How to Use it
+
+JSX fragments are still very new and some tools in the ecosystem may not be ready for it yet, so please be patient! However, we've tried our best to make it as simple as possible to upgrade from several popular toolchains:
```
gaearon: "some tools" -> "many tools"? We worked with some folks, but "some tools" makes it sound like those who are lagging behind are responsible for it. It's a *really* new syntax, we shouldn't expect tool authors to handle it yet.

content/blog/2017-11-28-fragments-in-react-v16.2.md
```diff
@@ -0,0 +1,181 @@
...
+or if you are using the [react preset](https://www.npmjs.com/package/@babel/preset-react):
+
+```
+yarn upgrade @babel/core @babel/preset-react # for yarn users
+npm update @babel/core @babel/preset-react # for npm users
+```
+
+If you are using Babel with [Webpack](https://webpack.js.org/), then no additional steps need to be done because [babel-loader](https://github.com/babel/babel-loader) will be using your peer-installed version of Babel.
+
+Unfortunately, support for Babel 6.x is not available, and there are currently no plans to back-port.
```
gaearon: "to back-port" -> "to backport it" (I think that's a more common spelling)

content/blog/2017-11-28-fragments-in-react-v16.2.md
```diff
@@ -0,0 +1,181 @@
...
+```
+yarn upgrade flow-bin # for yarn users
+npm update flow-bin # for npm users
+```
+
+to update Flow to the latest version.
+
+#### Others
+
+For other tools, please check with the corresponding documentation to check if there is support available. However, it's not a problem at all if you want the functionality of `<></>` without your linters and tooling complaining! You can always start with `<Fragment></Fragment>` and do a code-mod later to replace it with `<></>` when the appropriate support is available.
```
gaearon: "code-mod" -> "codemod"

content/blog/2017-11-28-fragments-in-react-v16.2.md
clemmy: @sebmarkbage 's thought is that `React.Fragment` is quite long and will be harder to sell.

content/blog/2017-11-28-fragments-in-react-v16.2.md
```diff
@@ -0,0 +1,203 @@
...
+
+#### With Attributes and Keys
+
+Note that the `<></>` syntax does not allow attributes on fragments. We made the decision to disallow attributes with fragments since they're semantically different from React elements. However, a fragment can be keyed by using the explicit fragment syntax. A common use case for this is grouping an item list into pairs:
+
+```jsx
+import React, { Fragment } from 'react';
+
+items.map(item => (
+  <Fragment key="item.key">
```
sebmarkbage: This should be `key={item.key}` otherwise it has the same string key for all items.

content/blog/2017-11-28-fragments-in-react-v16.2.md
```diff
@@ -0,0 +1,203 @@
...
+
+Additionally, there is a [pragma option available in babel-plugin-transform-react-jsx](https://github.com/babel/babel/tree/master/packages/babel-plugin-transform-react-jsx#pragmafrag) that allows the Babel compiler to de-sugar the `<>...</>` syntax to a custom identifier.
+
+### Usage
+
+As mentioned above, the most common way to use fragments in React is to simply use `<></>` tags. Generally, in the Javascript compiler phase of your build pipeline, the `<>...</>` will be de-sugared into a `createElement` call with `React.Fragment` as the type.
+
+Take a look at the section on [tooling support](#tooling-support) to see how to update your tooling to support JSX fragments.
+
+#### With Attributes and Keys
```
sebmarkbage: Drop the `With`. It's cleaner.

content/blog/2017-11-28-fragments-in-react-v16.2.md
```diff
@@ -0,0 +1,203 @@
...
+}
+```
+
+So, we added the `Fragment` component to React! `Fragment`s let developers return multiple elements from `render` methods using a familiar syntax that resembles normal JSX elements:
+
+```jsx
+import React, { Fragment } from 'react';
+
+render() {
+  // keys and commas are no longer needed
```
sebmarkbage:
```
// keys and commas are not needed in static JSX content
```
To be more precise?

content/blog/2017-11-28-fragments-in-react-v16.2.md
```diff
@@ -0,0 +1,203 @@
...
+  return <Fragment>
```
sebmarkbage: Should this be in a parenthesis with this line on the second line so that the comparison to the syntax below is equivalent?

clemmy: Yes, I'll update all the code blocks to have consistent styling with parentheses around JSX elements.

content/blog/2017-11-28-fragments-in-react-v16.2.md
```diff
@@ -0,0 +1,203 @@
+---
+title: "Fragments in React v16.2"
+author: [clemmy]
+---
+
+Today, we're excited to improved support for fragments in React! We are introducing a new `Fragment` export, which allows
+which allows users to render *static* fragments using familiar JSX element syntax.
+
+```jsx
+import React, { Fragment } from 'react';
```
gaearon: Would this work for TS users? AFAIK they have to enable a special mode to get imports like this working. This is a thorny issue: https://github.com/facebook/react/issues/11503. That's why I previously suggested to remove all imports from this blog post and just use `React.Fragment`. People will import React they way they already do, and there's no need to confuse them with a syntax that might not work.

gaearon: It also doesn't match the Flow recommendation. Note they change default React import to a namespace import as part of conversion: https://flow.org/en/docs/react/components/. I just don't think we should confuse people with import syntax here. The blog post is to about it.

content/blog/2017-11-28-react-v16.2.0-fragment-support.md
```diff
@@ -0,0 +1,325 @@
...
+
+[Prettier](https://github.com/prettier/prettier) will have support for fragments in their upcoming [1.9 release](https://github.com/prettier/prettier/pull/3237).
+
+### ESLint
+
+JSX Fragments are supported by [ESLint](https://eslint.org/) via [babel-eslint](https://github.com/babel/babel-eslint)! If you are not currently using it, then install it:
+
+```bash
+# for yarn users
+yarn add eslint@3.x babel-eslint@7
```
gaearon: Why are we suggesting to use a pretty old version of ESLint?

Issue #343, opened 2017-11-26
Bundling/Code Splitting Guide


thejameskyle:
I think we should be doing a better job encouraging React apps to do code splitting. Would an "Advanced" guide with the following structure be accepted?

- Bundling - What is it?
- Code Splitting - What is it?
- `import()` - Using it to code split
- Libraries:
  - React Loadable - Loading components/error states/etc
  - Other libs?

Most of the page will assume you're using Create React App or have Webpack setup already, it'll mention that you'll need one of the two and link to the Webpack docs.

bvaughn:
That sounds great to me, James. I'd love to see such a PR.

gaearon:
Fixed by https://github.com/reactjs/reactjs.org/pull/348.

PR #348, opened 2017-11-27
Init Code-Splitting docs

thejameskyle:
re #343

### DIFF
--- a/content/docs/code-splitting.md
+++ b/content/docs/code-splitting.md
@@ -44,9 +44,10 @@ console.log(add(16, 26)); // 42
 > 
 > Your bundles will end up looking a lot different than this.
 
-If you're using [Create React App](https://github.com/facebookincubator/create-react-app)
-or a similar tool you will have a Webpack setup out of the box to bundle your
-app. If you aren't, you'll need to setup bundling yourself. See the
+If you're using [Create React App](https://github.com/facebookincubator/create-react-app), [Next.js](https://github.com/zeit/next.js/), [Gatsby](https://www.gatsbyjs.org/), or a similar tool you will have a Webpack setup out of the box to bundle your
+app.
+
+If you aren't, you'll need to setup bundling yourself. For example, see the
 [Installation](https://webpack.js.org/guides/installation/) and
 [Getting Started](https://webpack.js.org/guides/getting-started/) guides on the
 Webpack docs.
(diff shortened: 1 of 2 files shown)


content/docs/code-splitting.md
```diff
@@ -0,0 +1,205 @@
...
+});
+
+const About = Loadable({
+  loader: () => import('./routes/About'),
+  loading: Loading,
+});
+
+const App = () => (
+  <Router>
+    <Route path="/" component={Home}/>
```
mjackson: We're going to fix this soon, but right now this route config won't give you what you probably expect because `path="/"` does *not* try to match the entire URL pathname, so `Home` will actually render at the `/about` URL as well. This behavior was intended to make it easier for people to nest routes, but it backfires in the `/` case 🤦‍♂️  You can read more about it [here](https://github.com/ReactTraining/react-router/issues/4958).

Anyway, the fix for this example would be to either a) use a more specific `path` (e.g. `path="/home"`) or b) use `<Route exact path="/" .../>`, which will match the URL exactly.

content/docs/code-splitting.md
```diff
@@ -0,0 +1,205 @@
...
+
+```js
+function add(a, b) {
+  return a + b;
+}
+
+console.log(add(16, 26)); // 42
+```
+
+> Note: Your bundles will end up looking a lot different than this.
```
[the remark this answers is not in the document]
thejameskyle: Webpack bundles generally look a lot different than that too

content/docs/code-splitting.md
thejameskyle: whoops, thanks

content/docs/code-splitting.md
```diff
@@ -0,0 +1,205 @@
...
+  loader: () => import('./OtherComponent'),
+  loading: () => <div>Loading...</div>,
+});
+
+const MyComponent = () => (
+  <LoadableOtherComponent/>
+);
+```
+
+React Loadable helps you rendering
```
yangshun: I think there's a small typo here. Should it be "helps you render"/"helps with rendering"?

> React Loadable helps you rendering loading states, error states...

content/docs/code-splitting.md
```diff
@@ -0,0 +1,205 @@
...
+
+When using [Babel](http://babeljs.io/) you'll need to make sure that Babel can
+parse the dynamic import syntax but is not transforming it. For that you will need [babel-plugin-syntax-dynamic-import](https://yarnpkg.com/en/package/babel-plugin-syntax-dynamic-import).
+
+## Libraries
+
+### React Loadable
+
+[React Loadable](https://github.com/thejameskyle/react-loadable) provides you
+with a nice API for introducing code splitting into your app where-ever you
```
yangshun: "where-ever" sounds odd. Would "wherever" be better?

gaearon:
Thank you so much! I’m sorry we left this hanging.

PR #349, opened 2017-11-27
Render Props

thejameskyle:
I saw there was already a page on Higher-Order Components but nothing on render props, so I figured I'd start one.

Since I already had the website running I started this, but I figured it's better to ask before I do a bunch of work since I'd also like input from other people on the important points to make here.

Rough outline:

- Providing values/functions to children
- Higher-Order Components vs render props
- Converting Higher-Order Components to render props

cc @ryanflorence @mjackson since you two seem like you're on a mission on this topic and teach this content every day

### DIFF
--- a/content/docs/nav.yml
+++ b/content/docs/nav.yml
@@ -56,6 +56,8 @@
       title: Web Components
     - id: higher-order-components
       title: Higher-Order Components
+    - id: render-props
+      title: Render Props
     - id: integrating-with-other-libraries
       title: Integrating with Other Libraries
     - id: accessibility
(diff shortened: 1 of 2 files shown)


bvaughn:
A section like this would be a great addition to the docs!

I'm a fan of using function children too, eg
```js
<OtherComponent>
  {data => <h1>Hello {data.target}</h1>}
</OtherComponent>
```

Maybe we can at least show an example of that on this page somewhere? 😄

mjackson:
Just FYI, [I had already submitted a PR](https://github.com/facebook/react/pull/10741) for a doc about Render Props to facebook/react before this repo was created. I was planning on resurrecting it here.

thejameskyle:
Cool, closing this in favor of #355

PR #353, opened 2017-11-27
Unify the annotation of components in docs

MoeSattler:
Hi!

I am not really sure if this is a valid PR. The reason for this PR was, that I was confused about the annotation. 
In some places I found `<div>`, in other places `<div />`. Same goes for React components. 

Since I couldn't recognise what pattern you guys followed, I just made this PR. This will change all `<*>` to `<* />` in the docs.

### DIFF
--- a/content/docs/accessibility.md
+++ b/content/docs/accessibility.md
@@ -42,7 +42,7 @@ Note that all `aria-*` HTML attributes are fully supported in JSX. Whereas most 
 ## Accessible Forms
 
 ### Labeling
-Every HTML form control, such as `<input>` and `<textarea>`, needs to be labeled accessibly. We need to provide descriptive labels that are also exposed to screen readers.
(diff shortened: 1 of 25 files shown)


MoeSattler:
Just signed the cla.

alexkrolick:
Thanks for the PR @MoeSattler. I'll try to clarify:

The closing bracket is required for self-closing tags (such as `<br />` or `<hr />`)  and can be used as a shorthand if a component has no children (`<div></div>` is equivalent to `<div />`).

Most of the doc examples are _not_ self-closing, but if you think the usage is confusing a PR to help clear up the difference would be welcome.

For more info see https://reactjs.org/docs/jsx-in-depth.html

MoeSattler:
Hi @alexkrolick, thank you for the clarification!

So when you talk about a component without a specific context, like [here](https://github.com/reactjs/reactjs.org/blame/master/content/docs/components-and-props.md#L92), would you you always use the self-closing tag?

threepointone:
Closing this, since we seem to have settle on not using the closing mark in our docs. Thanks for the PR!

PR #355, opened 2017-11-27
Add doc about Render Props

mjackson:
This PR is a re-submission of [a PR that I had previously made to facebook/react](https://github.com/facebook/react/pull/10741) but that was closed when the website moved to reactjs/reactjs.org. Sorry it took so long 😅

I'm planning on adding one more section that discusses the trade-offs when using a render prop with `React.PureComponent`, which was the source of some concern in the earlier submission.

/cc @thejameskyle who submitted #349 earlier today which discusses the same concept

### DIFF
--- a/content/docs/render-props.md
+++ b/content/docs/render-props.md
@@ -222,3 +222,66 @@ function withMouse(Component) {
 ```
 
 So using a render prop makes it possible to use either pattern.
+
+## Caveats
+
+### Be careful when using Render Props with React.PureComponent
+
+Using a render prop can negate the advantage that comes from using [`React.PureComponent`](https://reactjs.org/docs/react-api.html#reactpurecomponent) if you create the function inside a `render` method. This is because the shallow prop comparison will always return `false` for new props, and each `render` in this case will generate a new value for the render prop.
+
+For example, continuing with our `<Mouse>` component from above, if `Mouse` were to extend `React.PureComponent` instead of `React.Component`, our example would look like this:
+
+```js
+class Mouse extends React.PureComponent {
+  // Same implementation as above...
+}
+
+class MouseTracker extends React.Component {
+  render() {
+    return (
+      <div>
+        <h1>Move the mouse around!</h1>
+
+        {/*
+          This is bad! The value of the `render` prop will
+          be different on each render.
+        */}
+        <Mouse render={mouse => (
+          <Cat mouse={mouse} />
+        )}/>
+      </div>
+    );
+  }
+}
+```
+
+In this example, each time `<MouseTracker>` renders, it generates a new function as the value of the `<Mouse render>` prop, thus negating the effect of `<Mouse>` extending `React.PureComponent` in the first place!
+
+To get around this problem, you can sometimes define the prop as an instance method, like so:
+
+```js
+class MouseTracker extends React.Component {
+  constructor(props) {
+    super(props);
+
+    // This binding ensures that `this.renderTheCat` always refers
+    // to the *same* function when we use it in render.
+    this.renderTheCat = this.renderTheCat.bind(this);
+  }
+
+  renderTheCat(mouse) {
+    return <Cat mouse={mouse} />
+  }
+
+  render() {
+    return (
+      <div>
+        <h1>Move the mouse around!</h1>
+        <Mouse render={this.renderTheCat} />
+      </div>
+    );
+  }
+}
+```
+
+In cases where you cannot bind the instance method ahead of time in the constructor (e.g. because you need to close over the component's props and/or state) you should extend `React.Component` instead.
--- a/content/docs/render-props.md
+++ b/content/docs/render-props.md
@@ -6,6 +6,18 @@ permalink: docs/render-props.html
 
 The term ["render prop"](https://cdb.reacttraining.com/use-a-render-prop-50de598f11ce) refers to a simple technique for sharing code between React components using a prop whose value is a function.
 
+A component with a render prop takes a function that returns a React element and calls it instead of implementing its own render logic.
+
+```jsx
+<DataProvider render={data => (
+  <h1>Hello {data.target}</h1>
+)}/>
+```
+
+Libraries that use render props include [React Router](https://reacttraining.com/react-router/web/api/Route/Route-render-methods) and [Downshift](https://github.com/paypal/downshift).
+
+In this document, we’ll discuss why render props are useful, and how to write your own.
+
 ## Use Render Props for Cross-Cutting Concerns
 
 Components are the primary unit of code reuse in React, but it's not always obvious how to share the state or behavior that one component encapsulates to other components that need that same state.
--- a/content/docs/render-props.md
+++ b/content/docs/render-props.md
@@ -203,7 +203,7 @@ More concretely, **a render prop is a function prop that a component uses to kno
 
 This technique makes the behavior that we need to share extremely portable. To get that behavior, render a `<Mouse>` with a `render` prop that tells it what to render with the current (x, y) of the cursor.
 
-One interesting thing to note about render props is that you can implement most [higher-order components](/react/docs/higher-order-components.html) (HOC) using a regular component with a render prop. For example, if you would prefer to have a `withMouse` HOC instead of a `<Mouse>` component, you could easily create one using a regular `<Mouse>` with a render prop:
+One interesting thing to note about render props is that you can implement most [higher-order components](/docs/higher-order-components.html) (HOC) using a regular component with a render prop. For example, if you would prefer to have a `withMouse` HOC instead of a `<Mouse>` component, you could easily create one using a regular `<Mouse>` with a render prop:
 
 ```js
 // If you really want a HOC for some reason, you can easily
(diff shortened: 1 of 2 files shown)


toddsby:
I was under the impression this was an anti-pattern due to performance considerations. Thoughts @acdlite @gaearon @vjeux I know Dan has weighed in briefly on twitter but I couldn't find a write up on this.

mjackson:
@toddsby The main concern with using a render prop is when you use them with pure components, as I've already [outlined in the "Caveats" section](https://github.com/reactjs/reactjs.org/pull/355/commits/af78367ec3285a4d937afbf44913f6a38e6d1719). If you think there's something else I can add there, I'm all ears.

content/docs/render-props.md
```diff
@@ -0,0 +1,287 @@
+---
+id: render-props
+title: Render Props
+permalink: docs/render-props.html
+---
+
+The term ["render prop"](https://cdb.reacttraining.com/use-a-render-prop-50de598f11ce) refers to a simple technique for sharing code between React components using a prop whose value is a function.
```
alexkrolick: It would be nice to kick off this page with a simple example and motivation to frame the conversation. The HOC doc starts with:

> Concretely, a higher-order component is a function that takes a component and returns a new component.
> `const EnhancedComponent = higherOrderComponent(WrappedComponent);`
> ...explain at high-level where and when HOCs are useful...

Paraphrasing the example code from @thejameskyle's #349, how about this: 

---

_The term ["render prop"](https://cdb.reacttraining.com/use-a-render-prop-50de598f11ce) refers to a simple technique for sharing code between React components using a prop whose value is a function._

A component with a render prop takes a function that returns a React element and calls it instead of implementing its own render logic.

```jsx
<DataProvider render={data => (
  <h1>Hello {data.target}</h1>
)}/>
```

By inverting control of the render method to the consumer of the component, render props provide a flexible way to encapsulate and share logic between components.

Libraries that use render props include [React Router](https://reacttraining.com/react-router/web/api/Route/Route-render-methods) and the enhanced input library [Downshift](https://github.com/paypal/downshift).

In this document, we’ll discuss why render props are useful, and how to write your own.

content/docs/render-props.md
```diff
@@ -0,0 +1,287 @@
...
+      <div>
+        <h1>Move the mouse around!</h1>
+        <Mouse render={this.renderTheCat} />
+      </div>
+    );
+  }
+}
+```
+
+In cases where you cannot bind the instance method ahead of time in the constructor (e.g. because you need to close over the component's props and/or state) you should extend `React.Component` instead.
```
alexkrolick: How about mentioning some other ways of writing render props at the end?

---

## Using Props Other Than `render`

Although the examples above use `render`, any prop can be used to implement the render prop pattern. For example, the component below passes a function as the `children` prop:

```jsx

<WithTitleAsync url="/my/api/books/123">
  { ({ title }) => <h1>{title}</h1> }
</WithTitleAsync>

```

```jsx

class WithTitleAsync extends React.Component {
  constructor(props) {
    super(props);
    this.state = { title: null };
  }

  componentDidMount() {
    fetch(this.props.url)
      .then(
        ({ title }) => { this.setState({ title }) },
        (error) => { this.setState({ error }) }
      )
  }

  render () {
    if (this.state.error) {
      return <span>Title Not Found</span>;
    } else if (!this.state.title) {
      return <span>Getting Title...</span>
    } else {
      return this.props.children(this.state.title);
    }
  }
};

WithTitleAsync.propTypes = {
  children: PropTypes.func.isRequired,
};
```

## Optimizing Render Props

If you've benchmarked your app and discovered that a render prop declaration is a performance bottleneck, here are some techniques to optimize:

- Move inline functions to be declared outside the render.

**Before: Inline Render Prop**
```jsx
class ABC extends React.Component {
  render() {
    <MyComponent render={() => <h1>Hello World</h1>}
  }
}
```

**After: Declared Outside Class**
```jsx
class ABC extends React.Component {
  render() {
    return <MyComponent render={helloWorld} />
  }
}

function helloWorld() {
  return <h1>Hello World</h1>
}
```

**After: Using Getters**
```jsx
class ABC extends React.Component {
  get helloWorld() {
    return <h1>Hello World</h1>
  }

  render() {
    return <MyComponent render={this.helloWorld} />
  }
}
```

- Memoize the render function using a utility like [Recompose](https://github.com/acdlite/recompose) or [Lodash.Memoize](https://lodash.com/docs/4.17.4#memoize)

content/docs/render-props.md
```diff
@@ -0,0 +1,287 @@
...
+    return (
+      <div>
+        <h1>Move the mouse around!</h1>
+        <Mouse render={this.renderTheCat} />
+      </div>
+    );
+  }
+}
+```
+
```
alexkrolick: _Moving this to a separate comment_

Here's an idea for an optimization section. I don't think it's entirely necessary for the first pass and could certainly be added later.

---

## Optimizing Render Props

If you've benchmarked your app and discovered that a render prop declaration is a performance bottleneck, here are some techniques to optimize:

- Move inline functions to be declared outside the render.

**Before: Inline Render Prop**
```jsx
class ABC extends React.Component {
  render() {
    <MyComponent render={() => <h1>Hello World</h1>}
  }
}
```

**After: Declared Outside Class**
```jsx
class ABC extends React.Component {
  render() {
    return <MyComponent render={helloWorld} />
  }
}

function helloWorld() {
  return <h1>Hello World</h1>
}
```

**After: Using Getters**
```jsx
class ABC extends React.Component {
  get helloWorld() {
    return <h1>Hello World</h1>
  }

  render() {
    return <MyComponent render={this.helloWorld} />
  }
}
```

- Memoize the render function using a utility like [Recompose](https://github.com/acdlite/recompose) or [Lodash.Memoize](https://lodash.com/docs/4.17.4#memoize)

timwis:
Thanks for this write-up! This pattern makes things simpler :)

alexkrolick:
Is anyone intending on adding a review for this? If not I propose merging and iterating via PR.

cc @mjackson @thejameskyle @gaearon

mjackson:
I'd like to add a few of the edits you proposed, @alexkrolick, and then I'll go ahead and merge. Thanks for the review 😅

content/docs/render-props.md
ron23: Memoizing will solve the PureComponent caveat mentioned in the doc wouldn't it?

content/docs/render-props.md
```diff
@@ -0,0 +1,287 @@
...
+To get around this problem, you can sometimes define the prop as an instance method, like so:
+
+```js
+class MouseTracker extends React.Component {
+  constructor(props) {
+    super(props);
+
+    // This binding ensures that `this.renderTheCat` always refers
+    // to the *same* function when we use it in render.
+    this.renderTheCat = this.renderTheCat.bind(this);
```
ron23: I think we often see the usage of arrow functions instead of binding in the constructor. 
Is there any particular reason not to use an arrow function?

mjackson: Yes. The whole point of this example is that we create the function only once instead of dynamically creating a function on every render. Unless you're suggesting I use an arrow function here in the constructor?

ron23: No, I meant Instead of this:
```
class MouseTracker extends React.Component {
  constructor(props) {
    super(props);

    // This binding ensures that `this.renderTheCat` always refers
    // to the *same* function when we use it in render.
    this.renderTheCat = this.renderTheCat.bind(this);
  }

  renderTheCat(mouse) {
    return <Cat mouse={mouse} />
  }

  render() {
    return (
      <div>
        <h1>Move the mouse around!</h1>
        <Mouse render={this.renderTheCat} />
      </div>
    );
  }
}
```

*Maybe* this makes more sense/more consistent?
```
class MouseTracker extends React.Component {
   // `this.renderTheCat` always refers to the *same* function when we use it in render.
  renderTheCat = (mouse) => {
    return <Cat mouse={mouse} />
  }

  render() {
    return (
      <div>
        <h1>Move the mouse around!</h1>
        <Mouse render={this.renderTheCat} />
      </div>
    );
  }
}
```

alexkrolick: The docs intentionally try to stick to standard syntax instead of proposal-stage. Class properties are getting there but aren't standard just yet 😄 

ron23: @alexkrolick my bad. I wasn't aware it isn't standard yet (with babel, the difference between standard and non standard is blurry). Thanks for the clarification.

mjackson: FWIW @ron23 I always write my code in the way you're suggesting here and never actually write out my `constructor`s, so I feel you. But ya, the React docs tend to favor `.bind` at present.

content/docs/render-props.md
Kovensky: (nit: your "Using Getters" example passes react elements as the render value, not a function)

content/docs/render-props.md
koba04: > // This binding ensures that `this.renderTheCat` always refers
   // to the *same* function when we use it in render.

Is there any case that `this.renderTheCat` refers a *different* function in render without this?
I know that `this` in the function may not be a MouseTracker's instance, but I guess the reference to the function is always the same.

PR #359, opened 2017-11-28
Amend link to flux-chat example

heytonytan:
Noticed that the flux-chat example link was dead. Fix by replacing with the right link.

As a side note, not sure why the flux chat example is in the flow folder instead of the flux folder. Assuming that it was appropriated to illustrate Flow concepts.

### DIFF
--- a/content/blog/2014-07-30-flux-actions-and-the-dispatcher.md
+++ b/content/blog/2014-07-30-flux-actions-and-the-dispatcher.md
@@ -41,6 +41,6 @@ Problems arise, however, if we have circular dependencies. That is, if Store A n
 Example Chat App
 ----------------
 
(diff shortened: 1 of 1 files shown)


gaearon:
Seems okay.. I guess. Flux folks probably don't maintain this example anymore so they deleted it in favor of [other examples](https://github.com/facebook/flux/tree/master/examples). Let me ask someone who's maintainig it.

kyldvs:
Yeah we deprecated the old flux-chat example. If you are looking for just a standard TodoMVC example we maintain you can use: https://github.com/facebook/flux/tree/master/examples/flux-todomvc, if you want a more complex example use: https://github.com/facebook/flux/tree/master/examples/flux-async, or you can just link to the example folder: https://github.com/facebook/flux/tree/master/examples

threepointone:
Closing this, thanks for the effort!

PR #370, opened 2017-11-29
Move the style note to be more prominent

gaearon:
Also use our standard formatting of "Note"s instead of an ad-hoc one, and remove mention of "best practices" since the term doesn't mean much.


gaearon:
cc @alexkrolick: let's use this style for notes. Also let's check that links work :-)

PR #374, opened 2017-11-30
Update lifting-state-up.md

budarin:
```js
return output.toFixed(3);
```
is simpler than

```js
  const rounded = Math.round(output * 1000) / 1000;
  return rounded.toString();
```


budarin:
I see the difference - toFixed always produces 3 digits even if is ended with 0

PR #375, opened 2017-11-30
Add FAQ page about dispatching browser events

### LINKED ISSUE #352: Document simulating browser events
Maybe for the FAQ. It's commonly asked how to simulate events in React. We have Simulate helpers but they only trigger the React events without the browser behavior. In some cases that's not enough. We also have SimulateNative helpers but they're undocumented and I think we should get rid of them (https://github.com/facebook/react/issues/11656).

What we should do IMO is to document how to use browser APIs to dispatch normal browser events. Surprisingly many people have no idea (including myself a few weeks ago). For example: https://github.com/facebook/react/issues/11111#issuecomment-347209267. (It's also commonly missed that the node needs to be in the document for this to work with React.)

radubrehar:
Resolves #352 by adding a new page in the FAQ that describes usage of browser APIs to programatically dispatch events

### DIFF
--- a/content/docs/faq-browser-events.md
+++ b/content/docs/faq-browser-events.md
@@ -1,0 +1,91 @@
+---
+id: faq-browser-events
+title: Dispatching Browser Events
+permalink: docs/faq-browser-events.html
+layout: docs
+category: FAQ
+---
+
+### How do I programatically dispatch browser events?
+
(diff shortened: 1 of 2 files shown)


eps1lon:
Thanks for your work. However, the underlying issue was resolved in the meantime. See https://github.com/reactjs/reactjs.org/issues/352#issuecomment-678797498 for more information.

PR #378, opened 2017-11-30
Included yarn prettier-all command as prerequisite

abiduzz420:
When I made my PR to facebook/react, there were lot of formatting changes in the code. Running just `yarn prettier` did not solve the issue. So I also had to execute `yarn prettier-all` which solved the formatting issue in my PR.

### DIFF
--- a/content/docs/how-to-contribute.md
+++ b/content/docs/how-to-contribute.md
@@ -83,7 +83,7 @@ The core team is monitoring for pull requests. We will review your pull request 
 3. If you've fixed a bug or added code that should be tested, add tests!
 4. Ensure the test suite passes (`yarn test`). Tip: `yarn test --watch TestName` is helpful in development.
 5. Run `yarn test-prod` to test in the production environment. It supports the same options as `yarn test`.
-6. Format your code with [prettier](https://github.com/prettier/prettier) (`yarn prettier`).
+6. Format your code with [prettier](https://github.com/prettier/prettier) (`yarn prettier`) and also run `yarn prettier-all`.
 7. Make sure your code lints (`yarn lint`). Tip: `yarn linc` to only check changed files.
 8. Run the [Flow](https://flowtype.org/) typechecks (`yarn flow`).
 9. If you haven't already, complete the CLA.
--- a/content/docs/how-to-contribute.md
+++ b/content/docs/how-to-contribute.md
@@ -83,10 +83,11 @@ The core team is monitoring for pull requests. We will review your pull request 
 3. If you've fixed a bug or added code that should be tested, add tests!
 4. Ensure the test suite passes (`yarn test`). Tip: `yarn test --watch TestName` is helpful in development.
 5. Run `yarn test-prod` to test in the production environment. It supports the same options as `yarn test`.
-6. Format your code with [prettier](https://github.com/prettier/prettier) (`yarn prettier`) and also run `yarn prettier-all`.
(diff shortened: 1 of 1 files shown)


clemmy:
Running `prettier-all` usually isn't necessary unless there is upstream code on master that doesn't adhere to the code style. `prettier-all` is slower, since it runs the entire codebase through `prettier`, whereas `yarn prettier` only runs against the files you change. Can you change the wording to better reflect this? 🙂

abiduzz420:
Yeah sure @clemmy

bvaughn:
Hm. I agree with what Clement said. Needing to run the slower `prettier-all` target should only be necessary in some edge cases- like a PR slipped through because we didn't wait for CI to finish before merging it _or_ because we changed our Prettier configuration somehow.

> If you have set custom settings for code formatting then you should run the command.

I think the newly-added note is confusing or maybe misleading. For example, even if you specify some custom IDE formatting rules that have updated files you've touched, running `yarn prettier` should be sufficient to undo these custom settings since the files will be in your staging area (if they were actually modified).

Could you elaborate on the case you're concerned about?

abiduzz420:
@bvaughn Yeah I agree that running `prettier-all` is slow. My whole point is based on the fact that when I ran `prettier`, it did not really help in adhering to the code style.
Let me give a few examples from this commit

1. Coding style in facebook/react: no whitespaces between the `{ }`
```
// Before running prettier-all, there where whitespace included between { } introduced due to my editor
import type { ReactNativeType } from './ReactNativeTypes';
// After I ran the cmd
import type {ReactNativeType} from './ReactNativeTypes';
```
2. Coding style in facebook/react: Comma at the end of parameter list
```
// Before running prettier-all, the commas were removed at the end of parameter list
ReactGenericBatching.injection.injectFiberBatchedUpdates(ReactNativeFiberRenderer.batchedUpdates)
// After I ran the cmd
ReactGenericBatching.injection.injectFiberBatchedUpdates(ReactNativeFiberRenderer.batchedUpdates,)
```
_Running just `prettier` did not solve my issue, hence I reported the issue._
I agree we need to restructure the sentence a bit. Can we specify in the prerequisite to run `prettier-all` only after you have made final changes to the PR ?

bvaughn:
My point was that running `yarn prettier` and `yarn prettier-all` _should be equivalent_ if we properly enforce that everyone runs `yarn prettier` before a brach is merged. (You shouldn't need to worry about changes outside of your staged changed.)

The only exception should be after upgrading the version of Prettier or modifying its config files, in which case the person doing that should (hopefully) know to reformat the entire project.

So this approach feels like addressing a symptom without addressing the underlying cause. If code is slipping through unformatted, I'd rather identify how (eg via Git blame) and prevent it from happening again rather than to suggest _everyone_ runs a slower running command before submitting a PR.

abiduzz420:
@bvaughn I agree with this point:
> So this approach feels like addressing a symptom without addressing the underlying cause. If code is slipping through unformatted, I'd rather identify how (eg via Git blame) and prevent it from happening again rather than to suggest everyone runs a slower running command before submitting a PR.

bvaughn:
Cool. Maybe you can help us track down things like this if they happen again then? Would be great to pinpoint how a specific change slipped through. 😄 

I think for now I'll close this diff though, b'c I think the goal of having committers just running `yarn prettier` is still a good one.

PR #381, opened 2017-12-01
a11y-doc: Add Fragments and expand Focus Control.

AlmeroSteyn:
I have added some information to the accessibility doc:

React Fragments is an incredible tool to assist in creating semantic accessible HTML and therefore I thought it is worthwhile to mention that in the accessibility docs as it will help stop base a11y errors in React apps:

So to set the scene I added a brief section on Semantic HTML:
image

In real world applications I ran into `ref` issues due to HOCs such as `withRouter` and `connect`. So I expanded the Focus Control section with info on how to use a callback prop with refs in these cases:

image

Let me know what you think.

Ping @jessebeach :-)

### DIFF
--- a/content/docs/accessibility.md
+++ b/content/docs/accessibility.md
@@ -39,6 +39,48 @@ Note that all `aria-*` HTML attributes are fully supported in JSX. Whereas most 
 />
 ```
 
+## Semantic HTML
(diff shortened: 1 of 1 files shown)


gaearon:
LGTM, thanks!

AlmeroSteyn:
It's only a pleasure. Thank you for merging!

PR #382, opened 2017-12-01
Correct words

XuXiaowen:
Correct words

#<Issue>


gaearon:
It's not the next section though.

Maybe "in the next two sections"?

bvaughn:
Is it the word "next" that's throwing you?

I don't think the way it was before is really problematic. 😄  But would something like "in the following sections" work better?

bvaughn:
I think the wording is probably okay as-is. 😄

PR #386, opened 2017-12-01
Add possible element types that accept propTypes

LAITONEN:
Specified that PropTypes validation could also be applied to stateless or functional components, which was not clearly stated.


bvaughn:
Maybe you could move this note to after the example? So the paragraph flows more smoothly?

LAITONEN:
@bvaughn You mean at the bottom of the page or below the first code excerpt?

bvaughn:
Below the example that it's currently above, so that the note doesn't distract from the primary point the paragraph is making. :)

LAITONEN:
@bvaughn Done!

content/docs/typechecking-with-proptypes.md
```diff
@@ -30,6 +30,8 @@ Greeting.propTypes = {
 };
 ```
 
+In this example, we are using a class component, but the same functionality could also be applied to the so-called stateless or functional components.
```
eps1lon, line 33:
```suggestion
In this example, we are using a class component, but the same functionality could also be applied to function components, or components returned from [`React.memo`](https://reactjs.org/docs/react-api.html#reactmemo) or [`React.forwardRef`](https://reactjs.org/docs/react-api.html#reactforwardref).
```

We no longer use the term functional or stateless components (misleading due to hooks). Added memo and forwardRef as other examples.

content/docs/typechecking-with-proptypes.md
```diff
@@ -30,6 +30,8 @@ Greeting.propTypes = {
 };
 ```
 
+In this example, we are using a class component, but the same functionality could also be applied to function components, or components returned from [`React.memo`](https://reactjs.org/docs/react-api.html#reactmemo) or [`React.forwardRef`](https://reactjs.org/docs/react-api.html#reactforwardref).
```
eps1lon, line 33:
```suggestion
In this example, we are using a class component, but the same functionality could also be applied to function components, or components created by [`React.memo`](https://reactjs.org/docs/react-api.html#reactmemo) or [`React.forwardRef`](https://reactjs.org/docs/react-api.html#reactforwardref).
```

eps1lon:
@LAITONEN Thanks!

PR #387, opened 2017-12-02
Document how highlight updates works in the optimizing performance pa…

### LINKED ISSUE #360: Document using "Highlight Updates" in React DevTools
As of React 16 it now works correctly and highlights components that re-render. We should mention this in "Optimizing Performance".

sanjsanj:
…ge - closes #360

#360


content/docs/optimizing-performance.md
```diff
@@ -180,6 +180,34 @@ Note that **the numbers are relative so components will render faster in product
 
 Currently Chrome, Edge, and IE are the only browsers supporting this feature, but we use the standard [User Timing API](https://developer.mozilla.org/en-US/docs/Web/API/User_Timing_API) so we expect more browsers to add support for it.
 
+## Visualize Component Updates with React DevTools' Highlight Updates Feature
+
+<img src="https://media.giphy.com/media/xUOxfgXNvzy37XlueY/giphy.gif" style="max-width:100%; margin-top:20px;" alt="React DevTools Highlight Updates example" />
```
gaearon: I think we might want to re-record a gif without recommendation to use Redux. Since in this context it appears as if it's *us* recommending it :-)

content/docs/optimizing-performance.md
```diff
@@ -180,6 +180,34 @@ Note that **the numbers are relative so components will render faster in product
...
+
+- Install one of the React DevTools packages above
+- Go in to the console and select the React tab
+- Enable the Highlight Updates option
+
+<img src="https://i.imgur.com/tUBa5k7.png" style="max-width:100%;" alt="How to enable highlight updates" />
+
+- Interact with your page and you should see colored borders momentarily appear around any element that has redrawn.
+
+The color of the border represents how many redraws have occurred on a 'heat' scale, with cold being good and hot being bad.
```
gaearon: The best is not seeing any color at all (which means it didn't re-render). I think we need to better explain that re-renders are expected, and that you can use `PureComponent` (as described below) to skip the unnecessary ones. This tool helps you see when you see *unexpected* updates. I think the workflow could be clearer.

We also need to be clear "re-render" doesn't mean updating the DOM.

content/docs/optimizing-performance.md
sanjsanj: Sure thing.

content/docs/optimizing-performance.md
sanjsanj: Cool, I'll add that in.

gaearon: In other words our rough recommendation is that you should see colors around things that actually change on the screen. If there's borders flashing around things that don't change, this might mean you could optimize that tree by skipping its rendering.

sanjsanj: I see now that I wasn't at all clear about a workflow for using this, my bad, I'll reword it!

Just want to be clear about re-rendering not updating the DOM.  Would I be right in saying that in this case we mean that we are measuring unnecessary calls on the `render()` method in the virtual DOM in components that do not change, and thus do not cause a re-render of the actual DOM?

And this happens because a component uses the `setState()` method or props/state it is subscribed to change?

sanjsanj:
@gaearon Apologies I know you must be really busy but just a reminder to have a look at my edits when you get a chance please, cheers!

sanjsanj:
@gaearon Happy new year!  Can I get a re-review on this PR please?

content/docs/optimizing-performance.md
```diff
@@ -185,13 +185,31 @@ Currently Chrome, Edge, and IE are the only browsers supporting this feature, bu
...
 
+You can now visualize these re-renders of the virtual DOM with React DevTools:
+
+- [Chrome Browser Extension](https://chrome.google.com/webstore/detail/react-developer-tools/fmkadmapgofadopljbjfkapdkoienihi?hl=en)
+- [Firefox Browser Extension](https://addons.mozilla.org/en-GB/firefox/addon/react-devtools/)
+- [Standalone Node Package](https://www.npmjs.com/package/react-devtools)
+
+In the developer console select the **Highlight Updates** option in the **React** tab
+
+<img src="https://i.imgur.com/tUBa5k7.png" style="max-width:100%; margin-top:10px;" alt="How to enable highlight updates" />
```
gaearon: I made a slightly higher quality version of this, please use it instead:

https://i.imgur.com/DhSxLIk.png

It would be great if you could add it to the website instead of relying on imgur. See how we did with `../images/blog/react-perf-chrome-timeline.png` above in this file.

content/docs/optimizing-performance.md
```diff
@@ -185,13 +185,31 @@ Currently Chrome, Edge, and IE are the only browsers supporting this feature, bu
...
 
 When a component's props or state change, React decides whether an actual DOM update is necessary by comparing the newly returned element with the previously rendered one. When they are not equal, React will update the DOM.
 
+You can now visualize these re-renders of the virtual DOM with React DevTools:
+
+- [Chrome Browser Extension](https://chrome.google.com/webstore/detail/react-developer-tools/fmkadmapgofadopljbjfkapdkoienihi?hl=en)
+- [Firefox Browser Extension](https://addons.mozilla.org/en-GB/firefox/addon/react-devtools/)
+- [Standalone Node Package](https://www.npmjs.com/package/react-devtools)
+
+In the developer console select the **Highlight Updates** option in the **React** tab
```
gaearon: please add colon at the end of this sentence

content/docs/optimizing-performance.md
```diff
@@ -185,13 +185,31 @@ Currently Chrome, Edge, and IE are the only browsers supporting this feature, bu
...
+
+<img src="https://i.imgur.com/tUBa5k7.png" style="max-width:100%; margin-top:10px;" alt="How to enable highlight updates" />
+
+Interact with your page and you should see colored borders momentarily appear around any elements that have redrawn.
+
+You should expect to correctly see redraws on components that are updating in the DOM, but you shouldn't see them around components that are not updating.  So if you see flashing colors around components that should not be updating then you know you have a needless `render()`.
+
+Below is an example of an [app](https://highlight-demo.firebaseapp.com/) setup specifically to demonstrate a very poorly optimized redraw, everything is being re-rendered in the virtual DOM even though only the text input component needs to be re-rendered.  More info can be found in this [blog post](https://blog.logrocket.com/make-react-fast-again-part-3-highlighting-component-updates-6119e45e6833) from [Ben Edelstein](https://blog.logrocket.com/@edelstein).
```
gaearon: Let's remove the app link, it's not *that* useful and can go down at some point

gaearon: We try to avoid "virtual DOM" in the documentation except one or two cases. Let's say:

>Below is an example of an app setup specifically to demonstrate a poorly optimized redraw. Even though only the text input changes, the components outside it are being re-rendered. Even though React only updates the changed DOM nodes, re-rendering still takes some time. In many cases it's not a problem, but if the slowdown is noticeable, we can fix it by avoiding reconciliation.

content/docs/optimizing-performance.md
```diff
@@ -185,13 +185,31 @@ Currently Chrome, Edge, and IE are the only browsers supporting this feature, bu
...
+
+<img src="https://i.imgur.com/dGIGC5p.gif" style="max-width:100%; margin-top:20px;" alt="React DevTools Highlight Updates example" />
```
gaearon: Let's host this on our server too.

sanjsanj:
@gaearon Thanks for the feedback!  Requested changes have now been made.

gaearon:
Did a minor rewording. This should be good. Thank you!

sanjsanj:
Excellent, cheers @gaearon !

PR #389, opened 2017-12-02
Move the methods above render

aamirafridi:

#<Issue>


alexkrolick:
I believe the docs convention is to put lifecycle methods first, including `render`, and then everything else.

cc @bvaughn this is based on the tweaks you made to some of the FAQs.

If this is something that should be enforced we can also add it to the contribution guide and possibly ESLint.

threepointone:
Closing, thanks for the PR!

PR #392, opened 2017-12-02
Add Fragment Documentation

clemmy:


acdlite:
I think the content you've copied from the blog post needs to be reframed a bit. It's written toward an audience of experienced React developers. But many people who read the Fragments doc will be using React for the first time, and won't care about limitations that existed before 16.2 — in fact, it may just serve to confuse them.

Instead of "here's how you used to return multiple children, and here's how you do it now," let's reframe it as "here's why you might need to return multiple children, and here's how you use fragments to do it."

clemmy:
Updated copy to be more beginner-friendly @acdlite

content/docs/fragments.md
```diff
@@ -0,0 +1,144 @@
+---
+id: fragments
+title: Fragments
+permalink: docs/fragments.html
+---
+
+A common pattern in React is for a component to return multiple elements. React provides first-class `Fragment` support that lets you group a list of children without adding extra nodes to the DOM.
```
acdlite: "React provides first-class `Fragment` support that lets you" -> "Fragments let you"

acdlite: Let's remove the first sentence since it's repeated in the "Motivation" section

content/docs/fragments.md
```diff
@@ -0,0 +1,144 @@
...
+---
+
+A common pattern in React is for a component to return multiple elements. Fragments let you group a list of children without adding extra nodes to the DOM.
+
+Fragments look like empty JSX tags:
+
+```js
+render() {
+  return (
+    <>
```
gaearon: Since neither Create React App nor CodePen (two tools we suggest in docs) support this syntax it seems odd to me that we’ve chosen to use it here.

I would prefer that we use React.Fragment everywhere on this page except a small footnote about the new syntax until it’s mainstream.

PR #405, opened 2017-12-08
components instead of elements

amrlabib:
Should this be elements or components  ? as mentioned in the first few sections, elements are just the building blocks of components like `div` or `span`, and here i think `<Contact />` and `<Chat>` are both components right  ?

#<Issue>


gaearon:
No, elements are anything you get by calling `React.createElement()`. JSX is sugar for that call.

Both `<div />` and `<App />` are elements. `App` is a component.

PR #406, opened 2017-12-09
Parentheses instead of parens

inderpreetsingh:
Just a small typo, I think. 
Unless parens is an accepted short form of parentheses? 
*(Edit: just found out that it is)*

I ended up reading it as parents and kept finding what "parents" did it add. 

**Edit:** Just came across http://www.dictionary.com/browse/parens, but would still advocate the use of the full word "parentheses". If it tripped me off, I believe it might trip others as well.


gaearon:
Sounds good

inderpreetsingh:
yay! thanks :+1:

PR #416, opened 2017-12-12
Added rmw-shell to tools-starter-kits.md

TarikHuber:
Added rmw-shell to tools-starter-kits.md


gaearon:
We've removed this page.

PR #424, opened 2017-12-13
Add "Why JSX"

gaearon:
We "lost" this during our last docs rewrite. Our original writing felt less confident, and we figured it's time we stop being apologetic about it.

However in practice I still see reasonable people get turned off by JSX, and pointing them at Pete's talk generally helps. So let's add this back in with an unapologetic tone?

### DIFF
--- a/content/docs/introducing-jsx.md
+++ b/content/docs/introducing-jsx.md
@@ -18,6 +18,16 @@ It is called JSX, and it is a syntax extension to JavaScript. We recommend using
 
 JSX produces React "elements". We will explore rendering them to the DOM in the [next section](/docs/rendering-elements.html). Below, you can find the basics of JSX necessary to get you started.
 
+### Why JSX?
+
+React embraces the fact that rendering logic is inherently coupled with other UI logic: how events are handled, how the state changes over time, and how the data is prepared for display.
+
+Instead of artificially separating *technologies* by putting markup and logic in separate files, React [separates *concerns*](https://en.wikipedia.org/wiki/Separation_of_concerns) with loosely coupled units called "components" that contain both. We will come back to components in a [further section](/docs/components-and-props.html), but if you're not yet comfortable putting markup in JS, [this talk](https://www.youtube.com/watch?v=x7cQ3mrcKaY) might convince you otherwise.
+
+React [doesn't require](/docs/react-without-jsx.html) using JSX, but most people find it helpful as a visual aid when working with UI inside the JavaScript code. It also allows React to show more useful error and warning messages.
+
+With that out of the way, let's learn the basics of JSX!
+
 ### Embedding Expressions in JSX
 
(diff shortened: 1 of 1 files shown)


content/docs/introducing-jsx.md
```diff
@@ -18,6 +18,16 @@ It is called JSX, and it is a syntax extension to JavaScript. We recommend using
 
 JSX produces React "elements". We will explore rendering them to the DOM in the [next section](/docs/rendering-elements.html). Below, you can find the basics of JSX necessary to get you started.
 
+### Why JSX?
+
+React embraces the fact that rendering logic is inherently coupled with other UI logic: how events are handled, how the state changes over time, and how the data is prepared for display.
+
+Instead of artificially separating *technologies* by putting markup and logic in separate files, React [separates *concerns*](https://en.wikipedia.org/wiki/Separation_of_concerns) with loosely coupled units called "components" that contain both. We will come back to components in a [further section](/docs/components-and-props.html), but if you're not yet comfortable putting markup in JS, [this talk](https://www.youtube.com/watch?v=x7cQ3mrcKaY) might convince you otherwise.
```
clemmy: How about

> Instead of artificially separating *technologies* by putting markup and logic in separate files, React [separates *concerns*](https://en.wikipedia.org/wiki/Separation_of_concerns) ~~with~~ **into** loosely coupled units called "components" that contain both. We will come back to components in a [further section](/docs/components-and-props.html), but if you're not yet comfortable putting markup in JS, [this talk](https://www.youtube.com/watch?v=x7cQ3mrcKaY) might convince you otherwise.

Something about saying ```this talk might convince you otherwise``` sounds a little off-putting to me. Maybe, just ```if you're not yet comfortable..., check out this talk```?

gaearon: I dunno. That's how I replied to folks on Twitter in the past and it's been effective :-) But I'm open to other wordings if you feel strongly—although I'm not the best person to review since I'm not a native speaker.

alexkrolick: What I think this means is that React is redefining "concerns" to mean "pieces of app logic" _instead of_ "layout, state, and mutation". How about:

> Instead of artificially separating _technologies_ by putting markup and logic in separate files, React separates _business concerns_ into loosely coupled units called "components" that contain both. We will come back to components in a further section, but if you're not yet comfortable putting markup in JS, this talk might convince you otherwise.

gaearon: I wouldn't put "business concerns" there. Sounds very enterprisey. "Separation of concerns" is a well-known CS principle, even though people don't always agree what it means.

Is is okay for this paragraph to sting a little bit and be slightly controversial. Those who won't be persuaded won't be anyway. But I think this argument is strong enough to sway the people who are on the fence.

alexkrolick: Doesn't have to be "business concerns"; I agree that it has an enterprisey connotation. But "separate concerns **into** components" is not grammatical; I'd stick with `with` or `using`.

gaearon: Oh I see. Yeah let's leave it :-)

PR #430, opened 2017-12-15
[WIP] Add a post about tooling

gaearon:
Work in progress. Sending a PR so I can read it on an actual website.

@bvaughn Leaving this for you to fill in about the release tool (closer to an end maybe?) A screenshot would be sweet :-)


gaearon:
Resubmitted as https://github.com/reactjs/reactjs.org/pull/434 so Brian can edit

PR #431, opened 2017-12-15
Update handling-events.md

jordanpapaleo:
Provided an example of event handling using currying

### DIFF
--- a/content/docs/handling-events.md
+++ b/content/docs/handling-events.md
@@ -155,7 +155,7 @@ In both cases, the `e` argument representing the React event will be passed as a
 
 ### Event handlers with Currying
 
-We can accomplish the above without re-binding every on render cycle by using function currying with arrow functions.  To do this we will need to slightly update our event handler to provide the function curry and then the handler to invoke the outer function passing our argument and returning the inner function to be used by the event.
+We can also accomplish event handling by using function currying with arrow functions.  To do this we will need to slightly update our event handler. When the handler is invoked by the render function, the outer function is called which passes our arguments in.  The handler also returns an inner function.  The inner function is what is called when the given event is executed.  The inner function has access to the arguments in the outer function.
 
 ```js
 class LoggingButton extends React.Component {
--- a/content/docs/handling-events.md
+++ b/content/docs/handling-events.md
@@ -152,3 +152,21 @@ Inside a loop it is common to want to pass an extra parameter to an event handle
 The above two lines are equivalent, and use [arrow functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) and [`Function.prototype.bind`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Function/bind) respectively.
 
 In both cases, the `e` argument representing the React event will be passed as a second argument after the ID. With an arrow function, we have to pass it explicitly, but with `bind` any further arguments are automatically forwarded.
+
+### Event handlers with Currying
+
+We can accomplish the above without re-binding every on render cycle by using function currying with arrow functions.  To do this we will need to slightly update our event handler to provide the function curry and then the handler to invoke the outer function passing our argument and returning the inner function to be used by the event.
+
+```js
+class LoggingButton extends React.Component {
+  handleClick = (i) => (e) => {
+    console.log(`button ${i} was clicked`);
+  }
+
+  render() {
+    return (
+      <button onClick={this.handleClick(i)}Click me</button>
+    );
+  }
+}
+```


content/docs/handling-events.md
```diff
@@ -152,3 +152,21 @@ Inside a loop it is common to want to pass an extra parameter to an event handle
 The above two lines are equivalent, and use [arrow functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) and [`Function.prototype.bind`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Function/bind) respectively.
 
 In both cases, the `e` argument representing the React event will be passed as a second argument after the ID. With an arrow function, we have to pass it explicitly, but with `bind` any further arguments are automatically forwarded.
+
+### Event handlers with Currying
+
+We can accomplish the above without re-binding every on render cycle by using function currying with arrow functions.  To do this we will need to slightly update our event handler to provide the function curry and then the handler to invoke the outer function passing our argument and returning the inner function to be used by the event.
```
alexkrolick: This is a good example, but a bit misleading since the curried function _does_ return a new function every time, same as `.bind`.

jordanpapaleo: Should I rework the description a bit?

bvaughn: Or you could just shift the focus away from the _function creation_ and instead focus this section on the _syntax_. A lot of people prefer fat-arrows for their syntax. 😄

jordanpapaleo:
Hey Everyone - I was on holiday but back now.  Thanks for the comments.  I updated the description of the function currying.  Let me know if its up to par for the docs.

jordanpapaleo:
@bvaughn @alexkrolick 

Hi Guys - 

Is there anything else I can do for this PR?

Thanks!

content/docs/handling-events.md
```diff
@@ -152,3 +152,21 @@ Inside a loop it is common to want to pass an extra parameter to an event handle
...
+
+```js
+class LoggingButton extends React.Component {
+  handleClick = (i) => (e) => {
+    console.log(`button ${i} was clicked`);
+  }
+
+  render() {
+    return (
+      <button onClick={this.handleClick(i)}Click me</button>
```
bvaughn: Where does `i` come from? Please show a complete example (e.g. a list of buttons) to avoid confusion.

Also, this example is missing the closing ">" between `onClick` and "Click me"

content/docs/handling-events.md
```diff
@@ -152,3 +152,21 @@ Inside a loop it is common to want to pass an extra parameter to an event handle
...
+    );
+  }
+}
```
bvaughn: nit: I prefer putting this type of code example into the ["examples" folder](https://github.com/reactjs/reactjs.org/tree/master/examples) (e.g. `examples/handling-events/event-handlers-with-currying.js`) and then embedding it within the markdown, like so:
```md
`embed:handling-events/event-handlers-with-currying.js`
```

This way Prettier will format it (and warn about any syntax errors like the one I mentioned above).

content/docs/handling-events.md
```diff
@@ -152,3 +152,21 @@ Inside a loop it is common to want to pass an extra parameter to an event handle
...
 
 In both cases, the `e` argument representing the React event will be passed as a second argument after the ID. With an arrow function, we have to pass it explicitly, but with `bind` any further arguments are automatically forwarded.
+
+### Event handlers with Currying
+
+We can also accomplish event handling by using function currying with arrow functions.  To do this we will need to slightly update our event handler. When the handler is invoked by the render function, the outer function is called which passes our arguments in.  The handler also returns an inner function.  The inner function is what is called when the given event is executed.  The inner function has access to the arguments in the outer function.
+
+```js
+class LoggingButton extends React.Component {
+  handleClick = (i) => (e) => {
```
gaearon: This is using an experimental syntax. We intentionally don't do this except for a few sections where we describe this syntax. It is confusing to people who are using plain ES6, and also is not guaranteed to end up in the spec.

jordanpapaleo: Fair enough.  Would you guys like me to just close this then?

devuxer: @gaearon,

I get your point, but as it stands now, the documentation is confusing and incomplete.

It says arrow syntax can cause unnecessary re-renders of child components, then immediately afterward, it shows an example using that flawed syntax to pass an argument to an event handler.

This literally begs the question, how do you avoid unnecessary re-renders *and* pass an argument?

Do you have to be an advanced user to want to know that? I don't think so.

If currying is the answer, it should be in this documentation. You don't have to explain currying, just provide an example and potentially link to different page that explains the concept.

So, I strongly urge you not to close this. This should definitely be fixed.

alexkrolick: There are quite a lot of examples in the FAQ: https://reactjs.org/docs/faq-functions.html


We should probably just link there.

devuxer: A link to an example would be a fine solution to me.

PR #433, opened 2017-12-15
Update `React.PureComponent` reference.

sznowicki:
Maybe I'm somehow weird, but I lived for some time with an idea that `React.Component` does implement `shouldComponentUpdate` with deep comparison. `React.PureComponent` is faster because it compares shallowly.

The changes I propose should avoid this "implicit" kind of way of thinking, since it would explicitly state that `React.Component` does not compare anything.

#<Issue>

### DIFF
--- a/content/docs/reference-react.md
+++ b/content/docs/reference-react.md
@@ -71,9 +71,9 @@ See the [React.Component API Reference](/docs/react-component.html) for a list o
 
 ### `React.PureComponent`
 
-`React.PureComponent` is exactly like [`React.Component`](#reactcomponent), but implements [`shouldComponentUpdate()`](/docs/react-component.html#shouldcomponentupdate) with a shallow prop and state comparison.
+`React.PureComponent` is exactly like [`React.Component`](#reactcomponent), but implements [`shouldComponentUpdate()`](/docs/react-component.html#shouldcomponentupdate) with a shallow prop and state comparison, which [`React.Component`](#reactcomponent) is missing completely. 
 
-If your React component's `render()` function renders the same result given the same props and state, you can use `React.PureComponent` for a performance boost in some cases.
+If your React component's `render()` function renders the same result given the same props and state, you can use `React.PureComponent` for a performance boost in some cases. 
 
 > Note
 >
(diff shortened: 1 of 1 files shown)


content/docs/reference-react.md
```diff
@@ -71,7 +71,7 @@ See the [React.Component API Reference](/docs/react-component.html) for a list o
 
 ### `React.PureComponent`
 
-`React.PureComponent` is exactly like [`React.Component`](#reactcomponent), but implements [`shouldComponentUpdate()`](/docs/react-component.html#shouldcomponentupdate) with a shallow prop and state comparison.
+`React.PureComponent` is exactly like [`React.Component`](#reactcomponent), but implements [`shouldComponentUpdate()`](/docs/react-component.html#shouldcomponentupdate) with a shallow prop and state comparison, which [`React.Component`](#reactcomponent) is missing completely.
```
gaearon: Maybe it would be clearer if it was something like:

>[PC] is similar to [C]. The difference between them is that [C] doesn't implement [sCU], but [PC] ...

?

sznowicki: Makes total sense for me. Updated my PR.

content/docs/reference-react.md
```diff
@@ -70,8 +70,7 @@ See the [React.Component API Reference](/docs/react-component.html) for a list o
 * * *
 
 ### `React.PureComponent`
-
-`React.PureComponent` is exactly like [`React.Component`](#reactcomponent), but implements [`shouldComponentUpdate()`](/docs/react-component.html#shouldcomponentupdate) with a shallow prop and state comparison, which [`React.Component`](#reactcomponent) is missing completely. 
+`React.PureComponent` is similar to [`React.Component`](#reactcomponent). The difference between them is that [`React.Component`](#reactcomponent) doesn't implement [`shouldComponentUpdate()`](/docs/react-component.html#shouldcomponentupdate), but `React.PureComponent` implements it with a shallow prop and state comparison.
```
alexkrolick: I would leave out the part about React.Component doesn't implement sCU, as it could be confused with saying that it can't/shouldn't. It should be enough to say: 

>`React.PureComponent` implements `shouldComponentUpdate` with a shallow prop and state comparison.

bvaughn: It sounds like this was the main reason Szymon opened the PR in the first place. (He had an incorrect impression about `Component` implementing a deep-comparison `shouldComponentUpdate` when it actually doesn't implement one at all.)

PR #435, opened 2017-12-16
Tweak link/code styles

sophiebits:
- Tone down link background
- Tone down code background
- Shrink font -- I'd prefer 15px Menlo and 17px Consolas but we'll meet
in the middle at 16px where both look reasonable, unlike 17px Menlo
which was comically large.

before-after


bvaughn:
I like 👍

PR #442, opened 2017-12-19
react-test-render/Cannot read property 'ReactCurrentOwner' of undefined

davidnguyen179:
**issue:**
[React-Test-Renderer@16 breaks trying to use internal property](https://github.com/facebook/react/issues/11197)

**solution**
[https://github.com/facebook/react/issues/11197#issuecomment-340341457](https://github.com/facebook/react/issues/11197#issuecomment-340341457)


gaearon:
If we do this, we should add a similarly formatted note to all packages, not just this one.

I would also prefer this format (which we highlight):

```
>Note
>
>(text)
```

content/docs/reference-test-renderer.md
```diff
@@ -19,6 +19,8 @@ This package provides a React renderer that can be used to render React componen
 
 Essentially, this package makes it easy to grab a snapshot of the platform view hierarchy (similar to a DOM tree) rendered by a React DOM or React Native component without using a browser or [jsdom](https://github.com/tmpvar/jsdom).
 
+**Note:** make sure `react-test-renderer` version is the same with `react`, `react-dom` version.
```
bvaughn: What about something like:

```md
> Note
>
> To prevent potential incompatibilities, ensure that `react-test-renderer` and other `react` packages all use the same version (eg `react@16.2.0`, `react-dom@16.2.0`, `react-test-renderer@16.2.0`).
```

As Dan mentioned in the main PR, this format will style differently and be more noticeable:
screen shot 2017-12-20 at 9 10 14 am

bvaughn: We can also add this note in other places too, since it's pretty generic. 😄

davidnguyen179: @bvaughn  It is so clean and easy to understand. 😄

bvaughn: Ok 😄 Want to update this PR to use that wording then?

davidnguyen179: Sure I have just updated the new styles & words. 😁

davidnguyen179: any suggestion to put this to another place in document?

davidnguyen179: I also suggest to put this to: https://reactjs.org/tutorial/tutorial.html before "Help, I’m Stuck!" section

bvaughn: Looks great!

Maybe we could also add it somewhere on pages like:
* /docs/react-dom.html
* /docs/shallow-renderer.html
* /docs/test-utils.html

Not sure if it's necessary to list specified things on the "Help, I'm Stuck!" section since there are so many potential causes. Seems better to refer to other resources (like it does).

gaearon:
Let’s add it to all top-level reference pages.

davidnguyen179:
hi @gaearon & @bvaughn!

I agree with your idea. How about add it to https://reactjs.org/docs/installation.html at the end of `Installing React` section?

gaearon:
Sounds good to me.

davidnguyen179:
Hi @gaearon !

I updated it.

content/docs/installation.md
```diff
@@ -89,6 +89,10 @@ npm install --save react react-dom
 
 Both Yarn and npm download packages from the [npm registry](http://npmjs.com/).
 
+>  Note:
+>
+>  To prevent potential incompatibilities, all `react` packages should use the same version (eg `react@16.2.0`, `react-dom@16.2.0`, `react-test-renderer@16.2.0`).
```
gaearon: Please "e.g.", not "eg". Even better, "for example" or "such as".

I also don't think we should hardcode `16.2.0` there, people will just use that. Any other ideas how we can communicate this?

davidnguyen179: I suggest:
+ e.g. => such as
+ remove version of package

**It is going to be**
To prevent potential incompatibilities, all `react` packages should use the same version (such as `react`, `react-dom`, `react-test-renderer`, ...).

davidnguyen179: what do you think about it?
@gaearon

bvaughn: > I also don't think we should hardcode 16.2.0 there, people will just use that. Any other ideas how we can communicate this?

Hm. Really? I guess that's possible, but I think the "eg" or "for example" prefix should make that clear.

gaearon: >Hm. Really?

Maybe I just have very low expectations for people reading the docs. :-) But if there's a way to phrase something that's more easily copy-pasteable by someone who barely knows any English, it's usually better.

bvaughn: Yeah. Just seems like everything else that comes to mind might be confusing in different ways. Maybe just?

> To prevent potential incompatibilities, all `react` packages should use the same version. (This includes `react`, `react-dom`, `react-test-renderer`, etc.)

gaearon: Sounds good to me

davidnguyen179: I have just updated with the words from @bvaughn

PR #444, opened 2017-12-20
deprecate selected attribute

git-pwn:

#<Issue>


donysukardi:
Should add deprecation node instead of removing the whole section on `selected`, as the attribute is practically still supported by React.

Perhaps something like,

### selected

The `selected` attribute is supported by `<option>` components. You can use it to set whether the component is selected.

>Note
>
>Use the `defaultValue` or `value` props on `<select>` instead of setting `selected` on `<option>` as the support for `selected` will be removed in the future release of React.

eps1lon:
Closing in favor of #3224

PR #450, opened 2017-12-20
Add installation pre-hook for dev mode

alexkrolick:
Makes it easier to clone and start developing locally instantly


bvaughn:
I'm reluctant to make this change, because it would add ~5.5 seconds (based on a relatively new Macbook Pro) to dev-server startup time. Assuming you've run `yarn dev` before (and have a cache) startup takes ~11 seconds.

Most people (probably) don't re-run this command too often, but I frequently find myself having to start+restart when debugging plugins, and I think the added time would be annoying.

alexkrolick:
That's fine, it's something we've added to the local dev setup at work and I've gotten used to never typing `yarn install` 😄

bvaughn:
😆 Makes sense.

Issue #451, opened 2017-12-20
"next" footer links don't pull title from source


alexkrolick:
Instead of pulling the title attribute from `next` target in nav metadata, the link generator seems to just strip the file extension and `-`. This looks weird if the title is different than the title generated from the link.

https://github.com/reactjs/reactjs.org/blob/184a7ef9476c524df0ff82362257c02bfbe8698d/src/templates/components/NavigationFooter/NavigationFooter.js#L89

EDIT: found an existing TODO for this 😄 

https://github.com/reactjs/reactjs.org/blob/184a7ef9476c524df0ff82362257c02bfbe8698d/src/components/MarkdownPage/MarkdownPage.js#L125-L132

raunofreiberg:
@alexkrolick Is there someone working on it already? If not, I'd love to take a crack at it.

alexkrolick:
Sure, it's all yours @raunofreiberg

raunofreiberg:
Gave it a shot at #688. Would appreciate it if you took a look :-) @alexkrolick

PR #452, opened 2017-12-20
Break out Installation into top-level doc section

alexkrolick:
- No more jumping over the installation section with links pointing to the "first" Quick Start page
- Get rid of some of that custom footer code that hardcodes "installation" as a special case
- Installation is important, may as well have its own section 😛 

image

### DIFF
--- a/content/docs/add-react-to-a-new-app.md
+++ b/content/docs/add-react-to-a-new-app.md
@@ -10,7 +10,7 @@ The easiest way to get started on a new React project is by using a starter kit.
 
 > Note:
 > 
-> Check out the [Try React](/docs/try-react.html) page if you are looking for light-weight environments to experiment with React.
+> Check out the [Try React](/docs/try-react.html) page if you are looking for lightweight environments to experiment with React.
 > The options on this page are designed to help bootstrap single-page applications with linting, testing, bundling, production optimizations, and more.
 
 ## Create React App
--- a/content/docs/add-react-to-a-new-app.md
+++ b/content/docs/add-react-to-a-new-app.md
@@ -44,4 +44,4 @@ See the Community section for a list of [starter kits](/community/starter-kits.h
 
 ## Advanced
 
-For advanced configuration, see [Installing React](/docs/installation-existing-app.html#installing-react) in the next section.
+If you prefer to configure a project manually, see [Installing React](/docs/add-react-to-an-existing-app.html#installing-react) in the next section.
--- a/content/docs/add-react-to-a-new-app.md
+++ b/content/docs/add-react-to-a-new-app.md
@@ -8,6 +8,11 @@ next: add-react-to-an-existing-app.html
 
 The easiest way to get started on a new React project is by using a starter kit.
 
+> Note:
+> 
+> Check out the [Try React](/docs/try-react.html) page if you are looking for light-weight environments to experiment with React.
+> The options on this page are designed to help bootstrap single-page applications with linting, testing, bundling, production optimizations, and more.
+
 ## Create React App
 
 [Create React App](http://github.com/facebookincubator/create-react-app) is the best way to start building a new React single page application. It sets up your development environment so that you can use the latest JavaScript features, provides a nice developer experience, and optimizes your app for production. You’ll need to have Node >= 6 on your machine.
--- a/content/docs/add-react-to-a-new-app.md
+++ b/content/docs/add-react-to-a-new-app.md
@@ -1,7 +1,9 @@
 ---
-id: installation-new-app
-title: Adding React to a New Application
-permalink: docs/installation-new-app.html
+id: add-react-to-a-new-app
+title: Add React to a New Application
+permalink: docs/add-react-to-a-new-app.html
+prev: try-react.html
+next: add-react-to-an-existing-app.html
 ---
 
 The easiest way to get started on a new React project is by using a starter kit.
--- a/content/docs/installation-try-react.md
+++ b/content/docs/installation-try-react.md
@@ -1,0 +1,21 @@
+---
+id: installation-try-react
+title: Try React
+permalink: docs/installation-try-react.html
+redirect_from:
+  - "docs/installation.html"
+  - "download.html"
+  - "downloads.html"
+  - "docs/tooling-integration.html"
+  - "docs/package-management.html"
+  - "docs/language-tooling.html"
+  - "docs/environments.html"
+---
+
+React is flexible and can be used in a variety of projects. You can create new apps with it, but you can also gradually introduce it into an existing codebase without doing a rewrite.
+
+## Trying Out React
+
+If you're just interested in playing around with React, you can use CodePen. Try starting from [this Hello World example code](http://codepen.io/gaearon/pen/rrpgNB?editors=0010). You don't need to install anything; you can just modify the code and see if it works.
+
+If you prefer to use your own text editor, you can also [download this HTML file](https://raw.githubusercontent.com/reactjs/reactjs.org/master/static/html/single-file-example.html), edit it, and open it from the local filesystem in your browser. It does a slow runtime code transformation, so don't use it in production.
--- a/content/docs/try-react.md
+++ b/content/docs/try-react.md
@@ -31,9 +31,9 @@ If you prefer to use your own text editor, you can also [download this HTML file
 
 - Try the [Tutorial](/tutorial/tutorial.html) for a hands-on practical example.
 
-### Install React
+### Complete Development Environment
 
-If you are ready to add React to your app, check out the install guides:
+If you are ready to build a full application with React, check out the install guides. These setups are designed to get you up and running with a gives you a great developer experience and a production-ready setup with linting, testing, and optimizations built-in; however, they require more time to setup amd install. The lighter-weight solutions above are the best fit if you are new to React or just experimenting.
 
-- **[Add React to a New App](/docs/add-react-to-a-new-app.html)** Create a new app with a starter kit like [Create React App](http://github.com/facebookincubator/create-react-app).
-- **[Add React to an Existing App](/docs/add-react-to-a-new-app.html)** Add React to a build system or larger app.
+- [Add React to a New App](/docs/add-react-to-a-new-app.html) Create a new app with a starter kit like [Create React App](http://github.com/facebookincubator/create-react-app).
+- [Add React to an Existing App](/docs/add-react-to-a-new-app.html) Add React to a build system or larger app.
--- a/content/docs/try-react.md
+++ b/content/docs/try-react.md
@@ -13,7 +13,7 @@ redirect_from:
   - "docs/environments.html"
 ---
 
-Try React online or set up your local dev environment
+Try React online or set up your local development environment.
 
 ## Online
 
--- a/content/docs/try-react.md
+++ b/content/docs/try-react.md
@@ -35,5 +35,5 @@ If you prefer to use your own text editor, you can also [download this HTML file
 
 If you are ready to build a full application with React, check out the install guides. These setups are designed to get you up and running with a great developer experience and a production-ready setup with linting, testing, and optimizations built-in; however, they require more time to set up amd install. The lighter-weight solutions above are the best fit if you are new to React or just experimenting.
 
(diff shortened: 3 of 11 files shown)


content/docs/installation-try-react.md
```diff
@@ -0,0 +1,21 @@
...
+  - "docs/installation.html"
+  - "download.html"
+  - "downloads.html"
+  - "docs/tooling-integration.html"
+  - "docs/package-management.html"
+  - "docs/language-tooling.html"
+  - "docs/environments.html"
+---
+
+React is flexible and can be used in a variety of projects. You can create new apps with it, but you can also gradually introduce it into an existing codebase without doing a rewrite.
```
bvaughn: I'm not sure this lead paragraph (and the below header) make sense anymore, given what this page has been remade into.

Maybe we should rename the page title to something like "Try it on CodePen" and leave out this paragraph entirely?

content/docs/installation-try-react.md
```diff
@@ -0,0 +1,21 @@
...
+title: Try React
+permalink: docs/installation-try-react.html
+redirect_from:
+  - "docs/installation.html"
+  - "download.html"
+  - "downloads.html"
+  - "docs/tooling-integration.html"
+  - "docs/package-management.html"
+  - "docs/language-tooling.html"
+  - "docs/environments.html"
```
bvaughn: The new Installation pages should have prev/next frontmatter links 😄

alexkrolick: See #451 - the lack of sync between the titles and the ids causes the next/prev to appear as "Installation New Apps", etc

content/docs/installation-try-react.md
alexkrolick: Sure, although, how about adding more than Codepen? The Codesandbox template for React is pretty mature and quite popular, and we could link out to other content like the editable examples on the home page.

alexkrolick:
> It feels odd though for "Installation" to come after "Quick Start".

Yeah, a little, but it does feel a bit weird that lots of links to Quick Start appear to be skipping a section if that's where you land from the home page.

content/docs/installation-try-react.md
bvaughn: Not sure I understand. The next/prev titles are currently generated based on the URL you put in the next/prev frontmatter (so it's something we control, even if it's not synced).

I think the new section should still have next/prev links to feel more consistent with the others.

content/docs/installation-try-react.md
bvaughn: Sure. That'd be totally fine. CodeSandbox comes to mind as another excellent solution.

content/docs/installation-try-react.md
alexkrolick:
```
//  installation-existing-app.md
id: installation-existing-app
title: Add React to an Existing Application
```

```
// elsewhere
next: installation-existing-app.html
```

Results in:

image

Instead of:

image

This is because of #451 where the url is parsed instead of looking up the title in the TOC data structure. I am planning on fixing that and adding next/prev once it is done, but **I can certainly adjust the filenames for now so that it looks correct**.

bvaughn: > so it's something we control, even if it's not synced

You chose to name the new file `installation-existing-app.md`. So I'm just saying, before #451 is resolved, we could work around this issue by just naming it `add-react-to-an-existing-app.md` instead. 😄

bvaughn:
I think that's okay, or at least preferable to this. I wonder if people actively _look_ for installation- they'll expect it to come first (and may miss it entirely if it's collapsed below an expanded list of items.)

content/docs/try-react.md
```diff
@@ -0,0 +1,22 @@
...
+  - "docs/environments.html"
+---
+
+React is flexible and can be used in a variety of projects. You can create new apps with it, but you can also gradually introduce it into an existing codebase without doing a rewrite.
+
+## Trying Out React
+
+If you're just interested in playing around with React, you can use CodePen. Try starting from [this Hello World example code](http://codepen.io/gaearon/pen/rrpgNB?editors=0010). You don't need to install anything; you can just modify the code and see if it works.
+
+If you prefer to use your own text editor, you can also [download this HTML file](https://raw.githubusercontent.com/reactjs/reactjs.org/master/static/html/single-file-example.html), edit it, and open it from the local filesystem in your browser. It does a slow runtime code transformation, so don't use it in production.
```
gaearon: I think we should add a section here that points to the next two sections and explains in which case you'll want which. It's too easy to miss them otherwise.

alexkrolick: Working on this 😛

alexkrolick: Pushed a change - added "Local Development" section to the Try React page

alexkrolick:
@bvaughn 
- Moved Installation to top
- Added prev/next by changing filenames
- Added Codesandbox

@gaearon 
- Reorganized and added links to the other install guides

content/docs/try-react.md
```diff
@@ -0,0 +1,33 @@
...
+
+If you're just interested in playing around with React, you can use an online code playground. Try a Hello World template on [CodePen](codepen://hello-world) or [CodeSandbox](https://codesandbox.io/s).
+
+## HTML Template
+
+If you prefer to use your own text editor, you can also [download this HTML file](https://raw.githubusercontent.com/reactjs/reactjs.org/master/static/html/single-file-example.html), edit it, and open it from the local filesystem in your browser. It does a slow runtime code transformation, so don't use it in production.
+
+## Install React
+
+- **[Add React to a New App](/docs/add-react-to-a-new-app.html)** Create a new app with a starter kit like [Create React App].(http://github.com/facebookincubator/create-react-app)
```
gaearon: The link Markdown is broken

content/docs/try-react.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Online
+
+If you're just interested in playing around with React, you can use an online code playground. Try a Hello World template on [CodePen](codepen://hello-world) or [CodeSandbox](https://codesandbox.io/s).
+
+## HTML Template
+
+If you prefer to use your own text editor, you can also [download this HTML file](https://raw.githubusercontent.com/reactjs/reactjs.org/master/static/html/single-file-example.html), edit it, and open it from the local filesystem in your browser. It does a slow runtime code transformation, so don't use it in production.
+
+## Install React
```
gaearon: This makes it sound like it's part of *this* page. IMO this makes it worse. I was thinking something that more obviously implies it's an optional *next* step, not something you have to do.

content/docs/try-react.md
```diff
@@ -0,0 +1,33 @@
...
+
+- **[Add React to a New App](/docs/add-react-to-a-new-app.html)** Create a new app with a starter kit like [Create React App].(http://github.com/facebookincubator/create-react-app)
+- **[Add React to an Existing App](/docs/add-react-to-a-new-app.html)** _Advanced configuration._ Use React with a build system like Webpack.
```
gaearon: I don't think this description is very helpful.

alexkrolick: I'm not sure the dichotomy is correct here - "Add React to A New App" links to CRA primarily, and "Existing App" is all about build systems and CDN links. It makes sense to promote CRA as the starting point, but I think it is confusing to talk about the next page as only for existing apps. It's more of an advanced guide, or maybe a bundling/build guide. The CDN links could also probably be more prominent, with the caveat that they aren't very useful without Babel.

What do you think?

gaearon: I understand what you're saying. My point is just that 

>_Advanced configuration._ Use React with a build system like Webpack.

is not very meaningful to someone visiting the page for the first time. What advanced configuration? What is webpack? Does the first link not use webpack? etc I'm open to other phrasings of the same idea, this sentence just read as confusing to me.

>I think it is confusing to talk about the next page as only for existing apps

In a way, yes. But on the other hand it provides a clear entry point for each use case. If you're ready to configure things on your own you already know enough that you don't need either of these sections.

alexkrolick: Just changed the description to "Add React to a build system or larger app." --- how does that sound? Can iterate on the naming of the page later.

Also added a "Next Steps" header above the Quick Start and Install sections to make it clear that those aren't part of Trying React, but ideas for what to do next if you're set up.

content/docs/try-react.md
```diff
@@ -0,0 +1,37 @@
...
+
+## HTML Template
+
+If you prefer to use your own text editor, you can also [download this HTML file](https://raw.githubusercontent.com/reactjs/reactjs.org/master/static/html/single-file-example.html), edit it, and open it from the local filesystem in your browser. It does a slow runtime code transformation, so don't use it in production.
+
+## Next Steps
+
+### Quick Start
+
+Head over to the [Quick Start](/docs/hello-world.html) section to try some examples.
```
gaearon: I think it's a bit misleading to call QS section that (since it doesn't have that many useful "examples").

Maybe it would be better to direct people either to Tutorial (for those looking for a hands-on practical introduction) or to QS (but frame it as a step-by-step introduction to React concepts).

gaearon:
I'm really liking these changes. Left one last nit I have.

content/docs/try-react.md
alexkrolick: Linking to the Tutorial is a great idea.

alexkrolick:
- Pulled CDN links to the top level
- Added link to tutorial page

Care to take one more pass?

😄

content/docs/try-react.md
```diff
@@ -0,0 +1,39 @@
...
+  - "docs/installation.html"
+  - "download.html"
+  - "downloads.html"
+  - "docs/tooling-integration.html"
+  - "docs/package-management.html"
+  - "docs/language-tooling.html"
+  - "docs/environments.html"
+---
+
+Try React online or set up your local dev environment
```
gaearon: "development environment"

Also let's end with period like we do elsewhere?

content/docs/try-react.md
```diff
@@ -0,0 +1,39 @@
...
+  - "docs/environments.html"
+---
+
+Try React online or set up your local development environment.
+
+## Online
+
+If you're just interested in playing around with React, you can use an online code playground. Try a Hello World template on [CodePen](codepen://hello-world) or [CodeSandbox](https://codesandbox.io/s).
+
+## HTML Template
```
gaearon: Maybe "Minimal HTML Template" to make it clearer what this is about

content/docs/try-react.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Next Steps
+
+### Quick Start
+
+- Head over to the [Quick Start](/docs/hello-world.html) section for a step-by-step introduction to React concepts.
+
+- Try the [Tutorial](/tutorial/tutorial.html) for a hands-on practical example.
+
+### Install React
```
gaearon: Again, this sounds like it's a required step because of the title.

How about: "Powerful Development Environment" or "Complete Development Environment". Maybe also change the sentence to explain when/why you'd want that.

The goal is to explain this step takes time but gives you a great developer experience and a production-ready setup with linting, testing, and optimizations built-in. We have a chance to preempt "omg why is it 100mb" knee-jerk responses here by being explicit that it's not just for 'hello world' but more IDE-like.

content/docs/add-react-to-a-new-app.md
```diff
@@ -0,0 +1,42 @@
+---
+id: add-react-to-a-new-app
+title: Add React to a New Application
+permalink: docs/add-react-to-a-new-app.html
+prev: try-react.html
+next: add-react-to-an-existing-app.html
+---
+
+The easiest way to get started on a new React project is by using a starter kit.
```
gaearon: Related to https://github.com/reactjs/reactjs.org/pull/452/files#r158572397: maybe it's worth pointing people looking for something lightweight to "Try React" from this page. I'd love if we could make it clearer that this is a fully-featured setup that isn't *required* for using React but covers common needs for a *single-page application* including linting, testing, optimizations, etc.

alexkrolick: 👍  Added a note here

content/docs/try-react.md
alexkrolick: 👍

gaearon:
I think this is great! Hopefully nothing breaks 😄 
Thank you so much for reorganizing this!

PR #323, opened 2017-11-21
doc: added documentation for input type=file

vishalvrv9:
Have added some explanation on how to implement the form input type=file in React as suggested in #86. Please review and revert :smile: 

Have used a ref to the file input and used it to access the file(s) in submit handler instead of saving the file in the state. (Was suggested as a better solution in #218)

PR #455, opened 2017-12-21
gatsby-remark-code-repls-modification

vishalvrv9:
This PR is in continuation to #323 
Have added the required code in the [`examples` directory](https://github.com/reactjs/reactjs.org/tree/master/examples) which goes by the name `input-type-file.js` and pointed it to codepen using the gatsby-remark-code-repls plugin.

As you would notice, I created a new branch and submitted a PR. Since I'm new to this, I was curious if it is a good practice to create a new branch every time one wants to submit a PR or should I just have pushed the changes in the previous branch? :thinking:


vishalvrv9:
@bvaughn Can't figure out what went wrong with circleCI tests.

PR #458, opened 2017-12-22
Clarify Component capital letter caveat example.

uberfabulous:
Added "...the capital 'W' in..." to help reader focus their attention on the capital 'W' within `Welcome` for minor clarity improvement.

#<Issue>

### DIFF
--- a/content/docs/components-and-props.md
+++ b/content/docs/components-and-props.md
@@ -87,9 +87,9 @@ Let's recap what happens in this example:
 
 >**Caveat:**
 >
->Always start component names with a capital letter.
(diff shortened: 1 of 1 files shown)


gaearon:
I don’t think this is clearer.

Maybe something like “`<div />` starts with a lowercase “d” so React treats it as a DOM tag, but [...]”

alexkrolick:
How about:

> Always start component names with a capital letter. React assumes components starting with lowercase letters are DOM tags.
>
> For example, `<div />` represents an HTML `div` tag, but `<Welcome />` represents a component and requires `Welcome` to be in scope.

PR #469, opened 2017-12-26
Change "If you aren't familiar with Redux" para in context warnings

vannio:
In the context (hehe) of the sentence, `are` made more sense to me. Know how to use a state management lib? Then avoid using context.

### DIFF
--- a/content/docs/context.md
+++ b/content/docs/context.md
@@ -22,7 +22,7 @@ The vast majority of applications do not need to use context.
 
 If you want your application to be stable, don't use context. It is an experimental API and it is likely to break in future releases of React.
 
-If you aren't familiar with state management libraries like [Redux](https://github.com/reactjs/redux) or [MobX](https://github.com/mobxjs/mobx), don't use context. For many practical applications, these libraries and their React bindings are a good choice for managing state that is relevant to many components. It is far more likely that Redux is the right solution to your problem than that context is the right solution.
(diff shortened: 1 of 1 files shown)


alexkrolick:
I believe the gist of the sentence is that if you shouldn't try to solve the problems Redux solves using context. I.e., if you want to manage state globally, try Redux first.

Sorry that that isn't clear! Does that make sense or do you still think the wording should be changed?

vannio:
@alexkrolick That makes sense, although it sounds like we are saying the same thing! I still think the word needs changing

alexkrolick:
Not quite - what it's trying to say is:
>  Don't use `context` for shared global state, because you will run into the `shouldComponentUpdate` issue. Redux and similar libraries have solved/worked around this issue, so try those out.

vannio:
I'm not disagreeing with the intention of the message.  This as a _sentence structure_ doesn't make sense:
> If you aren't familiar with `A`, don't use `B`. 

There's no implication of what to use instead of `B` because of the double negative.

> If you **are** familiar with `A`, don't use `B`, use `A` instead.

alexkrolick:
The idea is "if you aren't familiar with Redux, _check it out_ to see if it solves your problem better than `context`". Not to remind people familiar with Redux to use it.

The wording could be better. Maybe:

> If you are looking to manage global app state, don't use context, use an library like Redux or MobX.

Familiarity with the library shouldn't be an issue.

threepointone:
Since this, context has become a stable API, so this isn't relevant any more. But thank you for the PR!

Issue #246, opened 2017-11-06
Improve embedded coding examples


bvaughn:
Follow-up for #245:
* Import more of Dan's Codepen examples to the [new `codepen` folder](https://github.com/reactjs/reactjs.org/tree/master/codepen).
* Create a custom markdown tag (eg `[codepen:foo/bar](baz)`) that generates our Codepen links with default "Try it on Codepen" text. We could use the [`gatsby-remark-use-jsx`](https://github.com/reactjs/reactjs.org/tree/master/plugins/gatsby-remark-use-jsx) plugin as a template to follow for this.
* Generated HTML could also include a link to view the file directly in GitHub. (Some people have expressed concerns that Codepen's interface is not great for screen readers.)

bvaughn:
I'm not sure that adding direct links to view example source files in GitHub is worth doing. (I'm not sure it will actually address the accessibility concern.) So I'm going to hold off on doing this unless someone who knows more about this than I do confirms it as a worthwhile course of action.

bvaughn:
I'm going to combine the 2 Gatsby plug-ins I created recently (`gatsby-remark-babel-repl-link` and `gatsby-remark-codepen-examples`) into a single plug-in (`gatsby-remark-code-repls`) that can be shared upstream with Gatsby. This way others can use it too. I've talked with Kyle about this and he thinks it will be useful to others as well. Once it's been merged, I'll update this repo to use the new shared plug-in rather than the 2 local ones.

micpap25:
Hi @bvaughn , did the PR from @danielsbird fully resolve this issue? If it did, I think this can be closed. If not, could you let me know what needs to be done to finish this issue? Thank you.

PR #470, opened 2017-12-28
Move examples of rendering elements from CodePen into ./examples

danielsbird:
Move two examples in the _Rendering Elements_ Quickstart article from CodePen to the `examples` directory. This follows up on the work in #246.

### DIFF
--- a/content/docs/rendering-elements.md
+++ b/content/docs/rendering-elements.md
@@ -43,7 +43,7 @@ ReactDOM.render(
 );
 ```
 
-[Try it on CodePen.](http://codepen.io/gaearon/pen/rrpgNB?editors=1010)
+[](codepen://rendering-elements/render-an-element)
 
 It displays "Hello, world" on the page.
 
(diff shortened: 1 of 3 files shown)


danielsbird:
@bvaughn Can you please take a look at this when you have an opportunity. I think you would be the best person to review since you did #246.

content/docs/rendering-elements.md
```diff
@@ -72,7 +72,7 @@ function tick() {
 setInterval(tick, 1000);
 ```
```
bvaughn: We can also replace this whole inline example above!

content/docs/rendering-elements.md
```diff
@@ -43,7 +43,7 @@ ReactDOM.render(
 );
 ```
```
bvaughn: We can also replace the inline example above by embedding the new snippet too, like we do for the ["file" input type](https://reactjs.org/docs/forms.html#the-file-input-tag). 😁 

This is possible using the [`gatsby-remark-embed-snippet` plugin](https://www.gatsbyjs.org/packages/gatsby-remark-embed-snippet/)

Check out PR #323 for another example of this.

danielsbird:
Thanks for the suggestions! I will implement them this weekend.

danielsbird:
All done @bvaughn. Can you please take another look?

Issue #337, opened 2017-11-24
Document that syncing props and state is usually a bad idea


gaearon:
We used to have a "tip" about this on the old website (a few years ago), but not anymore.

I think maybe we need to revive that section because too many people attempt to derive state from props—either in constructor or cWRP—without realizing they could just do this work in render. In 90% of cases that's exactly what they need to do.

See also https://mobile.twitter.com/EntriaTech/status/934139022204653569.

jyash97:
Hey @gaearon  can I take this? Could you tell what to we need to add ?

alexkrolick:
@jyash97 the issue is yours! 

I've added an "in-progress" label so that others will know not to start work on the issue. If you change your mind about the issue, no worries! Just let me know so that I can remove the label and free it up for someone else to claim.

As far as what to add, how about starting with a disclaimer in [State and Lifecycle](https://reactjs.org/docs/state-and-lifecycle.html) and [FAQ: State](https://reactjs.org/docs/faq-state.html)?

jyash97:
a084d924fa955cf1713f9776fbccb509dcd1ce03

bvaughn:
Hey @jyash97. The outcome of #341 didn't actually end up related to this issue, but rather to #351, so I'm going to mark this issue as no longer in progress. If you'd like to send another PR that implements the suggestion from this issue though, let us know and we'll mark it back in progress. 😄

jyash97:
@bvaughn
Can you have a look at the first commit of #341, I added topics like what should come in State, is this the good way of tackling this issue?

bvaughn:
At a glance the section "_What Shouldn’t Go in State?_" seems like what this issue is asking for.

jyash97:
Okay let me try one more time to update this content

jyash97:
@bvaughn  How about adding this note to the State and Lifecycle or in FAQ:State:

_Note_ not to confuse with state and props as `state` should only contain the minimal amount of data needed to represent your UI's state and `props` are used to pass data to the child component. 

`state` should not contain _duplicated data from props_ ; Try to use props as the source of truth where possible. One valid use to store props in state is to be able to know its previous values, because props can change over time.

bvaughn:
I think that's just another explanation of why state and props are _different_ but this issue is asking for more of a warning about why syncing props to state is generally unnecessary.

jyash97:
Is second paragraph a good way to handle this issue and adding with a warning.

[...]

PR #477, opened 2018-01-01
syncing props and state is usually a bad idea

jyash97:
Update over the issue #337

### DIFF
--- a/content/docs/faq-state.md
+++ b/content/docs/faq-state.md
@@ -22,6 +22,17 @@ Here are some good resources for further reading on when to use `props` vs `stat
 
 In rare cases, it’s okay to initialize state based on [`props`](/docs/components-and-props.html). One valid use to store props in state is to be able to know its previous values, because props can change over time. This effectively “forks” the props and sets the state with the initial props.
 
+In rare cases, it's okay to initialize state based on props. One valid use to store props in state is to be able to know its previous values, because props can change over time. This effectively "forks" the props and sets the state with the initial props. Here's an example of a valid `React.Component` subclass constructor:
+
+```js
+constructor(props) {
+  super(props);
+  this.state = {
+    color: props.initialColor
+  };
+}
+```
+
 Beware of this pattern, as state won't be up-to-date with any props update. Instead of syncing props to state, you often want to [lift the state up](/docs/lifting-state-up.html#lifting-state-up) instead.
 
 If you "fork" props by using them for state, you might also want to implement [`componentWillReceiveProps(nextProps)`](/docs/react-component.html#componentwillreceiveprops) to keep the state up-to-date with them. But lifting state up is often easier and less bug-prone.
--- a/content/docs/faq-state.md
+++ b/content/docs/faq-state.md
@@ -20,11 +20,11 @@ Here are some good resources for further reading on when to use `props` vs `stat
 
 ### Is synching state and props a bad idea?
 
-In rare cases, it’s okay to initialize state based on [`props`](/docs/components-and-props.html). This effectively “forks” the props and sets the state with the initial props.
+In rare cases, it’s okay to initialize state based on [`props`](/docs/components-and-props.html). One valid use to store props in state is to be able to know its previous values, because props can change over time. This effectively “forks” the props and sets the state with the initial props.
 
-Beware of this pattern, as state won't be up-to-date with any props update. Instead of syncing props to state, you often want to [lift the state up](/docs/lifting-state-up.html) instead.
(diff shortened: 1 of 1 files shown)


content/docs/faq-state.md
```diff
@@ -18,6 +18,14 @@ Here are some good resources for further reading on when to use `props` vs `stat
 * [Props vs State](https://github.com/uberVU/react-guide/blob/master/props-vs-state.md)
 * [ReactJS: Props vs. State](http://lucybain.com/blog/2016/react-state-vs-pros/)
 
+### Is synching state and props a bad idea?
+
+In rare cases, it’s okay to initialize state based on [`props`](/docs/components-and-props.html). This effectively “forks” the props and sets the state with the initial props.
+
+Beware of this pattern, as state won't be up-to-date with any props update. Instead of syncing props to state, you often want to [lift the state up](/docs/lifting-state-up.html) instead.
```
bvaughn: Link to anchor tag:
```
[lift the state up](/docs/lifting-state-up.html#lifting-state-up)
```

content/docs/faq-state.md
```diff
@@ -18,6 +18,14 @@ Here are some good resources for further reading on when to use `props` vs `stat
...
+
+If you "fork" props by using them for state, you might also want to implement [`componentWillReceiveProps(nextProps)`](#componentwillreceiveprops) to keep the state up-to-date with them. But lifting state up is often easier and less bug-prone.
```
bvaughn: This link is wrong. Should point to:
```
[`componentWillReceiveProps(nextProps)`](/docs/react-component.html#componentwillreceiveprops)
```

jyash97:
Hey @bvaughn have updated the links and added a example

jyash97:
#337

content/docs/faq-state.md
```diff
@@ -18,6 +18,14 @@ Here are some good resources for further reading on when to use `props` vs `stat
 * [Props vs State](https://github.com/uberVU/react-guide/blob/master/props-vs-state.md)
 * [ReactJS: Props vs. State](http://lucybain.com/blog/2016/react-state-vs-pros/)
 
+### Is synching state and props a bad idea?
+
```
gaearon: I think it's not very clear from this description what "syncing" means.

I think we should show a three-line code example first and *then* explain why it's problematic.

jyash97:
@gaearon  
I have added the code  example

gaearon:
This shows the “good” example. I think we should show the “bad” example and how to fix it first.

jyash97:
I do not get the context of "good" and "bad" example but for fixing it we can add componentWillReceiveProps.
@gaearon

gaearon:
Bad code is when somebody does

```
constructor(props) {
  super(props);
  this.state = { isOn: props.isOn };
}

handleClick() {
  this.setState(prevState => ({
    isOn: !prevState.isOn
  });
}

render() {
  return (
    <button onClick={() => this.handleClick()>
      The switch is {this.state.isOn ? 'on' : 'off'}
    </button>
  );
}
``` 

“Good” example removes the state and shows how to lift it up to the parent instead.

jyash97:
Okay Got it, will take the same example and add description,

jyash97:
Hey @gaearon @bvaughn 
I have added the _bad example_ with its solution and I have added the full code. Please reduce its length as I am not getting which code would be right to display.
Hoping this time I did what was required.

content/docs/faq-state.md
```diff
@@ -18,6 +18,75 @@ Here are some good resources for further reading on when to use `props` vs `stat
...
+In rare cases, it's okay to initialize state based on props. One valid use to store props in state is to be able to know its previous values, because props can change over time. This effectively "forks" the props and sets the state with the initial props.
+
+Here's an example of what not to do:
```
gaearon: It's a bit confusing we start with "in rare cases it's okay" but then the example shows when it's not okay. Let's change to start with the "bad" section and only mention the good use case at the end.

jyash97:
@gaearon I  have deleted some lines which might confuse users.

jyash97:
@gaearon @bvaughn  I have removed the lines which might confuse users.

jyash97:
@gaearon  Should we close #337 or any update needed?

gaearon:
Why did you close? We don't currently have capacity to review all PRs but we'd get to it eventually.

jyash97:
Let me reopen it again

rickhanlonii:
Sorry for the long delay here. Since this PR was created against the old site I'm going to close this, but I think this use case is covered in the new docs for hooks in You Might Not Need an Effect: https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

PR #478, opened 2018-01-01
Changed the broken links 

ppatel221:
Changed the Broken links to link the appropriate docs or headers in the doc folder.


alexkrolick:
Thanks for update, but this isn't quite right - the links should have `.html` extensions so that they point to the right pages _when built_, not in the Markdown source. Here's the preview of your branch so you can see what happens to the links you changed: https://deploy-preview-478--reactjs.netlify.com/docs/add-react-to-a-new-app.html

PR #485, opened 2018-01-05
Fix a couple broken links on React without ES6 page

henryqdineen:
There were a couple of links that were broken


gaearon:
Sweet, thank you!

PR #486, opened 2018-01-05
Properly nest components in router

HeroProtagonist:
Hi!

I was checking out the code-splitting section and making a sample app with the examples. The route based example causes this error `Uncaught Error: A <Router> may have only one child element`. 

I nested them in a `div` to allow the example to run

### DIFF
--- a/content/docs/code-splitting.md
+++ b/content/docs/code-splitting.md
@@ -41,7 +41,7 @@ console.log(add(16, 26)); // 42
 ```
 
 > Note:
-> 
+>
 > Your bundles will end up looking a lot different than this.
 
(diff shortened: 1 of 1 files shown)


gaearon:
Maybe it should be a `Switch`? https://github.com/ReactTraining/react-router/issues/4928#issuecomment-292434315

cc @mjackson

HeroProtagonist:
@gaearon you are right, I updated the PR. For the simple example here a `<div>` seems good, but if the route say `/about/me` existed it would be matching multiple without using `<Switch>`

mjackson:
Yep, looks good :)

PR #489, opened 2018-01-06
Changes check order in 'Tutorial - Decalring a Winner'

Damaged-Organic:
It looks like it's better to check whether square is already filled before checking for winner with much heavier function. No big deal here, but seems like common courtesy.


BartoszKlonowski:
Hello @Damaged-Organic! It looks like your PR modifies content that is already long time ago removed.
Let me close this, but please do not hesitate to create new PRs for other open issues.
Thank you!

PR #491, opened 2018-01-06
Add Post component since unclear

ajgoldenwings:
In the lists-and-keys.html document, was going to replace the _content_ variable but component did not exist and was unclear until creating it by hand.

### DIFF
--- a/content/docs/lists-and-keys.md
+++ b/content/docs/lists-and-keys.md
@@ -253,7 +253,7 @@ Keys serve as a hint to React but they don't get passed to your components. If y
 ```js{12,13}
 function Post(props) {
   return (
-    <div key={props.id}>
+    <div id={props.id}>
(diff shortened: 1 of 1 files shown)


content/docs/lists-and-keys.md
```diff
@@ -250,12 +250,22 @@ ReactDOM.render(
 
 Keys serve as a hint to React but they don't get passed to your components. If you need the same value in your component, pass it explicitly as a prop with a different name:
 
-```js{3,4}
+```js{12,13}
+function Post(props) {
+  return (
+    <div key={props.id}>
```
bvaughn: This `key` attribute isn't necessary and won't actually do anything.

ajgoldenwings: Your right. I ment <div id={props.id}>. I think this will make things more clear.

PR #495, opened 2018-01-08
Clarify preferred "props default to true" expr

davidjb:
The current text is ambiguous without reading further into the paragraph -- until I re-read this text, I was trying to avoid both types of syntax because it seemed to imply that it wans't recommended (eg "props default to true" --> "we don't recommend using this").

https://github.com/facebook/react/pull/10777


threepointone:
Actually, we *do* suggest the latter form, not the former. The 'this' is referring to the heading "If you pass no value for a prop". I tweaked your PR to be clearer. Thank you for the effort!

gaearon:
“We don’t recommend not passing” is awkward. “We recommend always passing the value explicitly” would be better. :-)

PR #503, opened 2018-01-09
Add link to tests for further study

rdickert:
These are some excellent tests. I find them to be actually more informative than the docs in several cases. It would have saved me time if this link had been here.


BartoszKlonowski:
Hello @rdickert!
Unfortunately this PR is now outdated as react-test-renderer becomes deprecated and the page you modify in this PR no longer exists in the latest docs.
If you would still like to contribute you are most welcome to open a new PR to any of open issues.
Thank you!

PR #507, opened 2018-01-10
Add back the explanation of React's onChange vs. DOM "input" event.

mattmccutchen:
Fixes https://github.com/facebook/react/issues/3964 again after the
information was lost in 809e282.

FWIW, re the necessity of this change, as a new React user reading the [Forms page](https://reactjs.org/docs/forms.html), I didn't understand how the example could possibly work until I searched the web for the relationship between `onChange` and `onInput` and found [this thread](https://stackoverflow.com/questions/38256332/in-react-whats-the-difference-between-onchange-and-oninput), which links to the issue.

### DIFF
--- a/content/docs/forms.md
+++ b/content/docs/forms.md
@@ -66,7 +66,7 @@ class NameForm extends React.Component {
 
 [Try it on CodePen.](https://codepen.io/gaearon/pen/VmmPgp?editors=0010)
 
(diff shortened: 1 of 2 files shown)


TimothyDJewell:
Additionally, the [form events section of Synthetic Events](https://reactjs.org/docs/events.html#form-events)
https://github.com/reactjs/reactjs.org/blob/main/content/docs/reference-events.md
Still contains:
> For more information about the onChange event, see [Forms](https://reactjs.org/docs/forms.html).

But the Forms page doesn't actually contain data about `onChange`.

BartoszKlonowski:
Closing as per issue is already closed as well and PR addresses the legacy docs.
Thank you!

PR #519, opened 2018-01-13
Fixed Possible Typo and Listed Points for Legibility

vegtelenseg:
Tweaked the wording of the Context title, and listed the reasons to **NOT** use context for better readability.


threepointone:
Context has since become a stable API. Thanks for the PR!

PR #520, opened 2018-01-13
Update fragments example to use JSX syntax

### LINKED ISSUE #518: Update react-component.html for React 16.2's improved fragment support
Hi!

React 16.2.0 [improved support for fragments](https://reactjs.org/blog/2017/11/28/react-v16.2.0-fragment-support.html), and #392 added some documentation in this repo for them.

However I've spotted another reference to fragments that is still using the array syntax - that I think should probably be updated too?
https://reactjs.org/docs/react-component.html#fragments

gnapse:
Closes #518

### DIFF
--- a/content/docs/reference-react-component.md
+++ b/content/docs/reference-react-component.md
@@ -133,7 +133,7 @@ render() {
 >
 > Don't forget to [add keys](/docs/lists-and-keys.html#keys) to elements in a fragment to avoid the key warning.
 
-The same can also be accomplished using [fragments](/docs/fragments.html), which don't require keys for static items:
+Since [React 16.2.0](/blog/2017/11/28/react-v16.2.0-fragment-support.html), the same can also be accomplished using [fragments](/docs/fragments.html), which don't require keys for static items:
 
 ```javascript
 render() {
--- a/content/docs/reference-react-component.md
+++ b/content/docs/reference-react-component.md
@@ -129,23 +129,23 @@ render() {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


gaearon:
How do you feel about showing both? We should still document that arrays "work" even if they're not the most ergonomic way to do it.

gnapse:
Definitely, was thinking about that too after I replaced it.

content/docs/reference-react-component.md
```diff
@@ -122,13 +122,27 @@ You can also return multiple items from `render()` using an array:
...
-    <li key="B">Second item</li>,
-    <li key="C">Third item</li>,
+    <li key="A">First item</li>,
+    <li key="B">Second item</li>,
+    <li key="C">Third item</li>,
   ];
 }
 ```
 
+The same can also be accomplished using [fragments](/docs/fragments.html):
```
gaearon: ... which don't require keys for static items:

(and then let's remove keys)

alexkrolick: Should there be a note here "New in React 16.2"?

gnapse: The "New in React 16.2" I assumed was already covered in the linked document about details of this feature. Didn't want to attempt to cover too much detail here.

gaearon: We can say "Since React 16.2.0, ..."

"React 16.2.0" could be a link to the blog post

gaearon:
Thanks!

gnapse:
Thank you for the opportunity and the guidance!

PR #522, opened 2018-01-14
Document that setState callback preserves this context (closes #492)

### LINKED ISSUE #492: Documentation should state that React preserves setState callback this context
I just recently found out that `setState`'s callback calls with the correct `this` context per [this line](https://github.com/facebook/react/blob/1637b43e27c40c73f9489603145f9bb1d0ece618/packages/react-reconciler/src/ReactFiberUpdateQueue.js#L352) in the React source. Nowhere in the documentation is it state that React internally calls callbacks woth context. This should be made clear as many assume it *doesn't* leading to unnecessary measures such as arrow functions, explicit binding, etc.

gnapse:
Closes #492

### DIFF
--- a/content/docs/reference-react-component.md
+++ b/content/docs/reference-react-component.md
@@ -299,7 +299,7 @@ A class component becomes an error boundary if it defines this lifecycle method.
 For more details, see [*Error Handling in React 16*](/blog/2017/07/26/error-handling-in-react-16.html).
 
 > Note
-> 
+>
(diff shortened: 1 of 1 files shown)


Beraliv:
@gnapse change topic to deploy it again with the bot, it would be great to have it in the docs.

gnapse:
@Beraliv not sure what you mean by changing the topic to deploy with bot. Can't this just be merged already?

Beraliv:
@gnapse I think it can, but thought bot is responsible for that. Wanted to check if the bot follows changes in PR and deploy after that.

PR #526, opened 2018-01-15
Update conferences

icarlossz:
I change the 'React Day Berlin' event from Upcoming Events to Past Conferences because it's already happened.


gaearon:
Thanks!

icarlossz:
You're welcome!

PR #527, opened 2018-01-15
Updated docs to include that VSCode supports it

Emad-salah:
I've edited the documentation for JSX fragments to include that VSCode supports the JSX Fragments syntax as of the v1.19 update


alexkrolick:
Thank you!

PR #529, opened 2018-01-16
[Improvement] Remove highlight line in blue

WeiChiaChang:
React


alexkrolick:
This is there for accessibility for keyboard navigation - the padding seems a bit off though; care to fix that instead?

WeiChiaChang:
@alexkrolick yes you're right, I should close it 😄

PR #531, opened 2018-01-17
Documentation for handling multiple options with <select>

### LINKED ISSUE #524: Improve documentation for <select> controlled component
In the React documentation for forms, there is only a footnote on using the <select> controlled component with the `multiple` attribute. I recently spent some time getting it to work for my use case and I feel having better documentation would have sped things up. My use case involved populating the `<option>` elements from JSON data being returned from the server, for a `select` component with the `multiple` attribute.

I am happy to write the documentation and raise a PR :)

nupurgrover:
Closes #524 

Screenshot:
image


content/docs/forms.md
```diff
@@ -182,13 +182,35 @@ class FlavorForm extends React.Component {
...
-><select multiple={true} value={['B', 'C']}>
->```
+### Handling multiple options with `<select>`
+
+It is also possible to select multiple options within a `<select>` tag, by enabling the `multiple` attribute and passing an array into the `value` attribute. This array should contain the corresponding `value` attributes of the `<option>` tags inside the `<select>` tag.
+
+The `handleChange` method from the above example would then change to get the selected values from `event.target.selectedOptions`.
+
+```javascript{4,10-12,24}
+handleChange(event) {
```
yangshun: The indentation is inconsistent for the rest of this page, which is 2 spaces for both HTML and JS. Do fix your markup accordingly (not sure if Prettier does that for you).

Also, you can simplify `handleChange` to:

```js
const selectedFlavors = Array.from(event.target.selectedOptions).map(option => option.value);
this.setState({value: selectedFlavors});
```

content/docs/forms.md
```diff
@@ -182,13 +182,35 @@ class FlavorForm extends React.Component {
...
+      return option.value;
+    });
+    this.setState({value: selectedFlavors});
+  }
+```
+
+The `<select>` inside the render would also change to pass an array to the `value` attribute:
+
+```javascript{4,10-12,24}
+<select value={this.state.value} onChange={this.handleChange} multiple={true}>
```
yangshun: 2 spaces indentation.

content/docs/forms.md
nupurgrover: @yangshun Done. I was not sure if using arrow functions was OK for the docs or not, that is why decided to write it using `function`

nupurgrover:
@yangshun I made changes to correct the indentation. Do let me know if there's anything else I should change, thanks! :)

yangshun:
I'm not from the React core team. Just a random commenter passing by 😄

gdennie:
The documentation has not been revised. Spent a while reinventing this solution. :(

nupurgrover:
I actually forgot I wrote this. Have just updated the branch. Will comment on the original issue and see if anyone from the core team would be willing to merge.

gaearon:
Sorry — we're a bit overwhelmed and haven't been looking at docs PRs closely.

content/docs/forms.md
```diff
@@ -182,6 +182,33 @@ class FlavorForm extends React.Component {
...
+  <option value="grapefruit">Grapefruit</option>
+  <option value="lime">Lime</option>
+  <option value="coconut">Coconut</option>
+  <option value="mango">Mango</option>
+</select>
+```
+
+[Try it on CodePen.](https://codepen.io/nupgrover/pen/RxeqLQ?editors=0010)
+
 > Note
```
gaearon: This note below now looks confusing because we've just discussed it. Maybe move it above, and then expand on how to change it dynamically?

gaearon:
Seems like `selectedOptions` isn't supported in IE? Could be confusing.
https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/selectedOptions

Any other ideas:?

content/docs/forms.md
nupurgrover: True, I'll make the change

nupurgrover:
@gaearon Oh, completely missed that! No, I don't have any ideas (except mentioning it explicitly in the docs) but I'll try and find a workaround and update this PR.

rickhanlonii:
Sorry for the delay, this is covered in the new docs here: https://react.dev/reference/react-dom/components/select#enabling-multiple-selection

rickhanlonii:
Actually that page doesn't explain how to do a _controlled_ component with `multiple={true}`. Feel free to submit a new PR to add that.

PR #536, opened 2018-01-17

hippodippo:
If there is a specific reason why the code did not use object destructuring then apologies.


threepointone:
This seems unnecessary, but thank you for the effort! All new code uses destructuring, and there's no harm in leaving this the way it is.

PR #538, opened 2018-01-19
update webpack version

jmny:
see https://github.com/gilbarbara/react-redux-saga-boilerplate/blob/master/README.md


eps1lon:
Thanks for your work. However, this page was removed in the meantime in #2535.

PR #540, opened 2018-01-20
Add next and previous links to Advanced Guides

Ediiik:

#<Issue>


eps1lon (approved):
Thanks. I wasn't sure if these should actually have prev/next pointers since the articles don't have an explicit ordering via numbers (like the articles in "Main concepts"). But we have prev/next links for "Installation" articles as well without explicit numbering.

gaearon:
FWIW it was intentional not to have them for the reason you mentioned. Their order is alphabetical.

I don’t mind too much either way and I agree there’s some inconsistency overall.

PR #542, opened 2018-01-22
Change of function setState signature.

mohitsharma1750:

#<Issue>


yangshun:
That syntax is correct, if a parameter is optional, the comma should be within the square brackets. Refer to [setTimeout](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setTimeout) documentation for example syntax.

PR #545, opened 2018-01-22
Add Object destruction in example code

orinamio:
Changed this:
---
>const scale = this.state.scale;
>const temperature = this.state.temperature;
---
to this:
---
>const { scale, temperature } = this.state;
---
and replaced others in example codes.

#<Issue>


threepointone:
This seems unnecessary, but thank you for the effort! All new code uses destructuring, and there's no harm in leaving this the way it is.

PR #547, opened 2018-01-23
Update handling-events.md

githubshrek:
This seems a much simpler way to reverse the toggle state.

### DIFF
--- a/content/docs/handling-events.md
+++ b/content/docs/handling-events.md
@@ -71,9 +71,7 @@ class Toggle extends React.Component {
   }
 
   handleClick() {
-    this.setState(prevState => ({
-      isToggleOn: !prevState.isToggleOn
-    }));
+    this.setState({isToggleOn: !this.state.isToggleOn});
   }
 
   render() {


githubshrek:
I think that there are 2 main approaches to take. Either you decide that we can live with a simplified view of setting state (since we are in the context of quick start and we don't need to trouble the beginner with such subtleties), OR you decide that it is essential for the beginner to understand this whole need to pass in a function, in which case you add this to the description.

How about doing both? Start the example in a simple way (with the state simply reversed), and then add a small section where the example is improved to use a function in setState? This would have the advantage that the beginner gets a gentle introduction, which then gets a little more complicated.

As it is, the "prevState => ({isToggleOn: !prevState.isToggleOn})" is just black magic on the page without any explanation.

alexkrolick:
It's really quite important that the functional form be used for toggles. If newcomers come to expect state to be updated synchronously they will end up with lots of frustrating bugs.

If it's confusing, let's link to the FAQ: https://reactjs.org/docs/faq-state.html#why-is-setstate-giving-me-the-wrong-value

Dan just updated it with some clarifications about this.state.

PR #548, opened 2018-01-23
Don't say simply

jameshfisher:
Created with https://github.com/jameshfisher/dont-say-simply


jameshfisher:
@gaearon thanks for taking the time to read through this, and to call out some of the mistakes! I've manually edited several parts, and opened some issues for me to fix at https://github.com/jameshfisher/dont-say-simply/issues. Automated removal is not a simple task! I now think this PR is sensible - any objections?

content/blog/2014-02-15-community-roundup-16.md
```diff
@@ -33,7 +33,7 @@ James Padosley wrote a short post on the basics (and merits) of React: [What is
 
 Taylor Lapeyre ([@taylorlapeyre](https://twitter.com/taylorlapeyre)) wrote another nice [introduction to React](http://words.taylorlapeyre.me/an-introduction-to-react).
 
-> React expects you to do the work of getting and pushing data from the server. This makes it very easy to implement React as a front end solution, since it simply expects you to hand it data. React does all the other work.
+> React expects you to do the work of getting and pushing data from the server. This makes it very easy to implement React as a front end solution, since it expects you to hand it data. React does all the other work.
```
gaearon: Here "just" or "only" would be more appropriate. Otherwise the sentence loses its intended meaning.

content/blog/2016-09-28-our-first-50000-stars.md
```diff
@@ -156,7 +156,7 @@ This work was eventually abandoned but was revived after React made its way into
 
 
 > Jordan Walke:
-> For the first step, I propose that we do the easiest, yet most general transformation possible. My suggestion is to simply map xml expressions to function call expressions.
```
gaearon: This is a quote, doesn't seem like we should be editing that.

content/docs/jsx-in-depth.md
```diff
@@ -282,7 +282,7 @@ You can put a string between the opening and closing tags and `props.children` w
 <MyComponent>Hello world!</MyComponent>
 ```
 
-This is valid JSX, and `props.children` in `MyComponent` will simply be the string `"Hello world!"`. HTML is unescaped, so you can generally write JSX just like you would write HTML in this way:
+This is valid JSX, and `props.children` in `MyComponent` will be the string `"Hello world!"`. HTML is unescaped, so you can generally write JSX just like you would write HTML in this way:
```
gaearon: Maybe "will just be the string"?

BartoszKlonowski:
Hello @jameshfisher!
I don't think this pull request is any longer needed. When I check for `simply` in the codebase of the docs, I see only two occurrences which sounds OK to me.
Let me close this PR, but please feel free to rebase and reopen if you find it still valid.
Thank you!

PR #550, opened 2018-01-24
Fix grammar on react-api reference doc

iAmWillShepherd:
Hello, noticed there was an extra `a` floating around so 🔥  it.


alexkrolick:
Thanks!

PR #551, opened 2018-01-24
Tweak state FAQ

gaearon:


content/docs/faq-state.md
```diff
@@ -31,38 +33,74 @@ incrementCount() {
...
+  // Let's say `this.state.count` starts at 0.
   this.incrementCount();
-  this.incrementCount(); // count is now 3
+  this.incrementCount();
+  this.incrementCount();
+
+  // If you read `this.state.count` now, it would still be 0.
+  // But when React re-renders the component, it will be 3.
+  
+  // If you need to know the pending state now, keep it in a local variable.
```
alexkrolick: How about an example of doing this?

```js
let newCount

this.setState(prevState => {
  const oldCount = prevState.count
  newCount = oldCount + 1
  return  { count: newCount }
})
```

alexkrolick: Ah, merged. Will make a PR.

gaearon: This particular example won't work because the updater might be applied later. I don't mind having an example but it'll probably need to be its own subsection.

content/docs/faq-state.md
```diff
@@ -6,23 +6,25 @@ layout: docs
...
+[`props`](/docs/components-and-props.html) (short for "properties") and [`state`](/docs/state-and-lifecycle.html) are both plain JavaScript objects. While both hold information that influences the output of render, they are different in one important way: `props` get passed *to* the component (similar to function parameters) whereas `state` is managed *within* the component (similar to variables declared within a function).
 
 Here are some good resources for further reading on when to use `props` vs `state`:
 * [Props vs State](https://github.com/uberVU/react-guide/blob/master/props-vs-state.md)
 * [ReactJS: Props vs. State](http://lucybain.com/blog/2016/react-state-vs-pros/)
 
 ### Why is `setState` giving me the wrong value?
 
+In React, both `this.props` and `this.state` represent the *rendered* values, i.e. what's currently on the screen.
```
alexkrolick: This is a really good clarification 😀

PR #557, opened 2018-01-29
Minor change reg assign this.state

imanushree:
* Updated state and lifecycle section to read this.state can be assigned in constructor or
   as a class variable.

### DIFF
--- a/content/docs/state-and-lifecycle.md
+++ b/content/docs/state-and-lifecycle.md
@@ -341,7 +341,7 @@ Instead, use `setState()`:
 this.setState({comment: 'Hello'});
 ```
 
-The only place where you can assign `this.state` is the constructor.
+You can assign `this.state` in the constructor or directly inside the class as a class variable.
 
 ### State Updates May Be Asynchronous
 


anushreesubramani:
@alexkrolick Can you please take a look ?

alexkrolick:
Hmm. Class props aren't standard yet... but it might be ok mention them here.

I also think it would be good to keep the word _only_ - it's important that assigning directly to `this.state` can only safely be used for initialization.

alexkrolick:
@gaearon do you think it's work document the class prop syntax here? It's getting pretty common in the wild and undeclared state is a common source of bugs as people start omitting the constructor in favor of bound class props.

anushreesubramani:
@alexkrolick thanks for the review! I was actually not aware that class props were not a standard yet. I saw them being used in several courses on eggheadio, blogs, twitter etc. and thought it was a standard and hence gave the PR as I couldn't find it in the docs.

content/docs/state-and-lifecycle.md
```diff
@@ -341,7 +341,7 @@ Instead, use `setState()`:
 this.setState({comment: 'Hello'});
 ```
 
-The only place where you can assign `this.state` is the constructor.
+You can assign `this.state` in the constructor or directly inside the class as a class variable.
```
mbrowne: Minor feedback: I think the more correct term is "class property" (rather than "class variable").

mbrowne:
Class properties are now a stage 3 proposal and the semantics for public properties at least are pretty well established: https://github.com/tc39/proposal-class-fields.

Maybe according to the contributing guidelines, you could say it's still too early to talk about them in the React docs, but actually they are already mentioned elsewhere, for example here:

> If you are using the experimental public class fields syntax, you can use class fields to correctly bind callbacks
https://reactjs.org/docs/handling-events.html

And actually that suggestion is arguably much more potentially problematic than the suggestion of using a class property to set initial `state`. In the future it looks like a better alternative to arrow functions in class properties to bind methods will be to use a decorator for this (but the Babel plugin for the latest decorators proposal is still in development). For details, see https://github.com/tc39/proposal-decorators/issues/54#issuecomment-369813610

PR #559, opened 2018-01-29
Update lists-and-keys.md

githubshrek:
I think the word "root" is misleading here. For me the root implies a single item, and the <li> here is more of a leaf. I think we can avoid all confusion simply by removing the word.


alexkrolick:
Alternate phrasing - ~~~"root"~~~ "wrapped" or "underlying"

PR #560, opened 2018-01-30
Add The React Podcast to podcasts list

mjackson:
React30 has merged with The React Podcast and is now producing new episodes again.


mjackson:
See also https://twitter.com/ReactPodcast/status/958323917688573952

PR #563, opened 2018-01-31
Updated Gatsby to move manifest from head to the end of body.

### LINKED ISSUE #179: Defer window.webpackManifest if possible?
https://reactjs.org/ has a 17 KB `window.webpackManifest` in a `<script>` tag in the `<head>`. Since it's in the `<head>`, it's render-blocking and the page can't render until after it loads. Is it possible to move this down underneath the content (eg. to the bottom of the `<body>`) so that it's not render-blocking?

### Edit by @bvaughn:

Once gatsbyjs/gatsby/issues/2541 has been resolved, the resolution to this issue will just be:
* Update to the latest release of Gatsby (`yarn upgrade-interactive`)
* Make sure things still build correctly on a fresh install (`rm -rf ./.cache && rm -rf ./node_modules && yarn install && yarn build`)
* Do quick smoke test to make sure there's no new regression (`yarn dev` and test localhost:8000)

isaacchien:
Fixes #179.

Upgraded Gatsby from 1.9.147 to 1.9.175. I'll add a speed test after I get the Netlify build.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -55,9 +55,40 @@
     lodash "^4.2.0"
     to-fast-properties "^2.0.0"
 
+"@types/history@*":
+  version "4.6.2"
+  resolved "https://registry.yarnpkg.com/@types/history/-/history-4.6.2.tgz#12cfaba693ba20f114ed5765467ff25fdf67ddb0"
+
 "@types/node@*":
   version "8.0.50"
   resolved "https://registry.yarnpkg.com/@types/node/-/node-8.0.50.tgz#dc545448e128c88c4eec7cd64025fcc3b7604541"
+
+"@types/react-router-dom@^4.2.2":
+  version "4.2.3"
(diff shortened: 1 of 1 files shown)


gaearon:
It seems like we also block on 

```html
<script src="//unpkg.com/docsearch.js@2.4.1/dist/cdn/docsearch.min.js"></script>
```

Can we not?

isaacchien:
I'll try not blocking on that too.

Here's a visual comparison of before and after updating Gatsby: https://www.webpagetest.org/video/compare.php?tests=180131_PY_3ec9e26f101b00c77e024c6fd2220f43,180131_2M_1f6ed711d933fd76e64eff9c3c13ff24

gaearon:
Sounds great. Let's get this in, and you can do the other update separately. This is awesome work!

PR #564, opened 2018-01-31
Add non-blocking docsearch

isaacchien:
Simply moves docsearch script to the end of body. Not sure if async or defer should be used instead.

Continuing #563.


gaearon:
What happens if you try to search before it loads?

isaacchien:
I throttled the network and manually entered text to see. Results aren't presented for text entered before the page loads until the next character is entered. Same goes for https://reactjs.org/.

gaearon:
Sounds good. Thanks!

PR #572, opened 2018-02-01
Note about special props in PropTypes typechecking

ranyitz:
👋 Hey,

I wanted to add a note regarding defining special props in component's `propTypes`.

While the warning is leading the user towards a usage of `this.props.key` it might also happen during PropTypes validation in development builds, which is a bit harder to figure out using the current explanation. Maybe adding this note could spare some debugging time 😉

### DIFF
--- a/content/warnings/special-props.md
+++ b/content/warnings/special-props.md
@@ -6,6 +6,4 @@ permalink: warnings/special-props.html
 
 Most props on a JSX element are passed on to the component, however, there are two special props (`ref` and `key`) which are used by React, and are thus not forwarded to the component.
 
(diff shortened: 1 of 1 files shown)


gaearon:
Let's make this note shorter and less prominent? I don't think most users bump into this in practice. Maybe just a single sentence as the last paragraph would be sufficient.

ranyitz:
@gaearon Thanks for the quick response!

What do you say about:

> __Note:__ You will also see this warning if you define `key` or `ref` PropTypes (using [PropTypes typechecking](https://reactjs.org/docs/typechecking-with-proptypes.html#proptypes)). React will try to  access them during props validation phase in the development build.

PR #573, opened 2018-02-02
Add Loadable Components to Code Splitting doc

neoziro:
Since [React Loadable](https://github.com/thejameskyle/react-loadable) is not a standard and does not provide any support (issues are disabled), I think it is good to give users an alternative: [Loadable Components](https://github.com/smooth-code/loadable-components).


threehams:
This can probably be closed, since on latest `loadable-components` is the recommendation for server-side code splitting.

neoziro:
🎉

PR #576, opened 2018-02-03
Update README for fork and clone

isaacchien:
This update is meant for people new to contributing like me. From my understanding, forking and then cloning is normally expected, so I figured I'd make this small change to prevent any possible confusion for new contributors in the future.


alexkrolick:
Makes sense 👍

PR #579, opened 2018-02-03
Add a link for translation

koba04:
I saw some issues about translating `reactjs.org` so I think it makes sense to add a link of Crowdin into README.md.

Thanks!


alexkrolick:
Thank you, there are lots of translation requests 😄

koba04:
Thanks!

PR #582, opened 2018-02-04
Make error message bold on the decoder page

gaearon:
Maybe this will make it a bit more visible?
Inspired by https://github.com/facebook/react/issues/12119


gaearon:
I think maybe it's a bit less elegant but it's more visible. I like it.

screen shot 2018-02-04 at 9 32 45 pm

alexkrolick:
Maybe move the error above the fine print?

gaearon:
I'm not sure how to do it without making the layout weird. (e.g. giant intro text shouldn't come after the error)

Happy to review a separate PR that does it though.

gaearon:
I'm digging this:

screen shot 2018-02-04 at 22 37 45

alexkrolick:
Yeah, that's fine too.

PR #587, opened 2018-02-05
16.3 release blog draft

bvaughn:
# TODO
- [ ] Add "learn more" links for lifecycle changes, `StrictMode`, and context API.
- [ ] Change name of file (date) to match release date.


content/blog/2018-02-07-react-v-16-3.md
```diff
@@ -0,0 +1,49 @@
...
+author: [bvaughn]
+---
+
+This release includes a new class component lifecycle (`getDerivedStateFromProps`), a new `StrictMode` component, and an official context API!
+
+For the past few months, the React team has been working on support for asynchronous rendering. We are excited about the new features this will enable, and we've learned that some changes will be required to the way we write React components.
+
+We're releasing version 16.3 primarily so that open source maintainers can start incorporating these changes into their libraries well in advance of the next major release. **However, you should not feel pressured to make changes to your application code yet.** We respect semver and will not ship breaking changes in a minor version!
+
+Read on to learn more about the release!
```
gaearon: Second sentence with an exclamation point in a sequence, maybe remove one of them?

content/blog/2018-02-07-react-v-16-3.md
```diff
@@ -0,0 +1,49 @@
+---
+title: "React v16.3.0: New lifecycles and context"
+author: [bvaughn]
+---
+
+This release includes a new class component lifecycle (`getDerivedStateFromProps`), a new `StrictMode` component, and an official context API!
+
+For the past few months, the React team has been working on support for asynchronous rendering. We are excited about the new features this will enable, and we've learned that some changes will be required to the way we write React components.
```
gaearon: Is it worth to linking to anything at all here for "asynchronous rendering"? Maybe Tomo's F8 talk. The first thought of anyone reading this will be "but I don't need asynchronous rendering". Which is maybe okay given further communication but I wish we could give more info right here.

content/blog/2018-02-07-react-v-16-3.md
```diff
@@ -0,0 +1,49 @@
...
+>
+> `StrictMode` checks are run in development mode only; they do not impact the production build.
+
+Although it is not possible for strict mode to catch all problems (eg mutation), it can help with many. If you see warnings in strict mode, those things will likely cause bugs for async rendering.
+
+In version 16.3, `StrictMode` helps with two things: identifying unsafe lifecycles and detecting unexpected side effects. Additional functionality will be added with future releases of React.
+
+[Learn more about the `StrictMode` component here.](#)
+
+## Official Context API
```
gaearon: I'd put this section fist because it's more exciting

content/blog/2018-02-07-react-v-16-3.md
```diff
@@ -0,0 +1,49 @@
...
+
+For example, with the current API, it is too easy to block the initial render with non-essential logic. In part this is because there are too many ways to accomplish a given task, and it can be unclear which is best. We've observed that the interrupting behavior of error handling is often not taken into consideration and can result in memory leaks- (something that will also impact the upcoming async rendering mode). The current class component API also complicates other efforts, like our work on a React compiler.
+
+Many of these issues are exacerbated by a subset of the component lifecycles (`componentWillMount`, `componentWillReceiveProps`, and `componentWillUpdate`). For this reason, we have decided to deprecate those methods.
+
+> **Note:**
+>
+> Deprecation warnings will be enabled in version 16.4, **but the deprecated lifecycles will continue to work until version 17**. After version 17, only the new "UNSAFE_" lifecycles will work.
```
gaearon: + "We have created an [automatic script] that prefixes the lifecycle names with the "UNSAFE_" prefix"

content/blog/2018-02-07-react-v-16-3.md
```diff
@@ -0,0 +1,49 @@
...
+
+Read on to learn more about the release!
+
+## Component Lifecycle Changes
+
+React's class component API has been around for years with little change. However, as we add support for more advanced features (such as [error boundaries](/docs/react-component.html#componentdidcatch) and the upcoming async rendering mode) we stretch this model in ways that it was not originally intended.
+
+For example, with the current API, it is too easy to block the initial render with non-essential logic. In part this is because there are too many ways to accomplish a given task, and it can be unclear which is best. We've observed that the interrupting behavior of error handling is often not taken into consideration and can result in memory leaks- (something that will also impact the upcoming async rendering mode). The current class component API also complicates other efforts, like our work on a React compiler.
```
gaearon: Odd punctuation: `leaks- (somethin`

content/blog/2018-02-07-react-v-16-3.md
```diff
@@ -0,0 +1,49 @@
...
+
+Many of these issues are exacerbated by a subset of the component lifecycles (`componentWillMount`, `componentWillReceiveProps`, and `componentWillUpdate`). For this reason, we have decided to deprecate those methods.
```
gaearon: Maybe:

>Many of these issues are exacerbated by a subset of the component lifecycles (`componentWillMount`, `componentWillReceiveProps`, and `componentWillUpdate`). They also happen to be the same methods that people get most confused about. For these reasons, we are going to deprecate those methods in favor of better alternatives.

>We recognize that many components use those methods, so the deprecations won’t happen overnight. At Facebook, we maintain more than 50 thousand React components, and we can't tell our engineers to rewrite them either. The migration will be as gradual as possible, and provide escape hatches.

content/blog/2018-02-07-react-v-16-3.md
```diff
@@ -0,0 +1,49 @@
...
+## `StrictMode` Component
+
+`StrictMode` is a tool for highlighting potential problems in an application. Like `Fragment`, `StrictMode` does not render any visible UI. It simply activates additional checks and warnings for its descendants.
+
+
+> **Note:**
+>
+> `StrictMode` checks are run in development mode only; they do not impact the production build.
+
+Although it is not possible for strict mode to catch all problems (eg mutation), it can help with many. If you see warnings in strict mode, those things will likely cause bugs for async rendering.
```
gaearon: "e.g."

content/blog/2018-02-07-react-v-16-3.md
```diff
@@ -0,0 +1,49 @@
...
+
```
gaearon: I think mentioning "mutation" as a problem is going to seriously worry MobX users. What is our messaging here?

content/blog/2018-02-07-react-v-16-3.md
```diff
@@ -0,0 +1,49 @@
...
+In version 16.3, `StrictMode` helps with two things: identifying unsafe lifecycles and detecting unexpected side effects. Additional functionality will be added with future releases of React.
```
gaearon: "for async rendering" — we still haven't explained what it is

content/blog/2018-02-07-react-v-16-3.md
gaearon: Also we have not established what "UNSAFE_" lifecycles are by this point. The person reading this will not know that they correspond to cWM/cWU/cWRP since those are supposed to be deprecated. It's not obvious.

content/blog/2018-02-07-react-v-16-3.md
bvaughn: Okay. I had linked to something like this in an earlier draft of the Quip, but was unsure about it since we had links (below) to an "update on async rendering". I didn't want to link to that here (so early) but I was unsure about linking to an outdated doc/talk too.

I can add a link to Tom's F8 talk though. That seems reasonable.

gaearon: Duplication doesn't hurt. If by the time the blog post comes out, we'll have the "update on async rendering" ready, we can link to it the first time we mention async rendering in the post too. It's just important that people who don't know what it is can immediately learn more without having to dig through the whole post.

content/blog/2018-02-07-react-v-16-3.md
bvaughn: I'm not sure how much to overlap with this doc and the other, more detailed one.

There's a tendency to move more and more into this one. Hm. Will think about it.

content/blog/2018-02-07-react-v-16-3.md
bvaughn: Should have been an em dash. Will fix.

content/blog/2018-02-07-react-v-16-3.md
gaearon: The goal is to prevent FUD spreading like fire. People who spread FUD don't research the problem, they stop at the first sentence that looks "bad". Here's an alternative wording that avoids this pitfall:

>Deprecation warnings will be enabled in version 16.4, **but the deprecated lifecycles will continue to work until version 17**. After version 17, it will still be possible to use them, but they will be aliased with an `UNSAFE_` prefix to indicate that they might cause issues. We have prepared an automated script that does this.

content/blog/2018-02-07-react-v-16-3.md
bvaughn: Not sure. I can soften this? ("certain types of mutations")

content/blog/2018-02-07-react-v-16-3.md
bvaughn: There will be an "update on async rendering" blog post released along with this that this will link into. I intentionally don't want to put _too much_ about that into this, because I want this to be a digestible summary. I can add a link.

content/blog/2018-02-07-react-v-16-3.md
bvaughn: Ok

bvaughn:
Added a section about the `createRef` API and update the `StrictMode` section to mention the string ref detection.

content/blog/2018-02-07-react-v-16-3.md
```diff
@@ -3,7 +3,7 @@ title: "React v16.3.0: New lifecycles and context"
 author: [bvaughn]
 ---
 
-This release includes a new class component lifecycle (`getDerivedStateFromProps`), a new `StrictMode` component, and an official context API!
+This release includes a new class component lifecycle (`getDerivedStateFromProps`), a new `StrictMode` component, an official context API, and a better way for managing refs!
```
gaearon: “Better” might imply you have to convert from functional refs.

Maybe “and a new ergonomic ref API”.

content/blog/2018-02-07-react-v-16-3.md
```diff
@@ -17,6 +17,15 @@ For many years, React has offered an experimental API for context. Although it w
 
 [Learn more about the new context API here.](#)
 
+## `createRef` API
+
+Previously, React provided two ways for managing refs: the legacy string ref API and the callback API. Although the string ref API was the more convenient of the two, it had [several downsides](https://github.com/facebook/react/issues/1373) and so our official recomendation was to [use the callback form instead](https://reactjs.org/docs/refs-and-the-dom.html#legacy-api-string-refs).
+
+Version 16.3 adds a new option for managing refs that offers the convenience of a string ref without any of the downsides:
+`embed:16-3-release-blog-create-ref.js`
+
```
gaearon: I would still add

>Callback refs will continue to be supported alongside this API for the foreseeable future. You don’t need to change the code using callback refs to the new API. Callback refs are slightly more flexible so they will stay as an advanced feature.

examples/16-3-release-blog-create-ref.js
```diff
@@ -0,0 +1,19 @@
+class MyComponent extends React.Component {
+  // highlight-next-line
+  _divRef = React.createRef();
```
gaearon: Nit: we don’t use the underscore convention elsewhere in the docs.

bvaughn: Gotcha. Long time habit.

content/blog/2018-02-07-react-v-16-3.md
bvaughn: I dig it.

bvaughn:
Rebased on master with the new Prettier config width.

bvaughn:
> I'd still prefer if we said "legacy lifecycles" rather than "deprecated lifecycles" until we actually enable the deprecation warnings. I don't feel strongly about this so deferring to your judgement.

Hm. I already changed the only occurrence of "deprecated lifecycles" to "legacy lifecycles". Do you see one I missed?

bvaughn:
Oh whoops, I was searching the other branch. Nevermind 😄

content/blog/2018-02-07-react-v-16-3.md
```diff
@@ -0,0 +1,76 @@
+---
+title: "React v16.3.0: New lifecycles and context"
+author: [bvaughn]
+---
+
+This release includes a new class component lifecycle (`getDerivedStateFromProps`), a new `StrictMode` component, an official context API, and a new ergonomic ref API!
+
+For the past few months, the React team has been working on support for [asynchronous rendering](#). We are excited about the new features this will enable, and we've learned that some changes will be required to the way we write React components.
```
sophiebits: s/some changes will be/some long-term changes will be/ maybe?

content/blog/2018-02-07-react-v-16-3.md
```diff
@@ -0,0 +1,76 @@
...
+
+We're releasing version 16.3 primarily so that open source maintainers can start incorporating these changes into their libraries well in advance of the next major release. **However, you should not feel pressured to make changes to your application code yet.** We respect semver and will not ship breaking changes in a minor version!
```
sophiebits: Honestly this paragraph right here reads to me as a little alarmist since you haven't even told me what you're changing. Can we move this down into the lifecycle section maybe?

content/blog/2018-02-07-react-v-16-3.md
```diff
@@ -0,0 +1,76 @@
...
+
+Read on to learn more about the release.
+
+## Official Context API
+
+For many years, React has offered an experimental API for context. Although it was a powerful tool, our advice was mostly to [avoid using it](/docs/context.html#why-not-to-use-context) because of potential problems with the API. We've always intended to replace the experimental API with a better one, and as of version 16.3 the new API is available!
+
+> **Note**
+>
+> The old context API will keep working for all React 16 minor releases, so you will have time to migrate.
```
sophiebits: "all React 16.x releases"? In case "minor" isn't universally clear.

content/blog/2018-02-07-react-v-16-3.md
```diff
@@ -0,0 +1,76 @@
...
+
+For the past few months, the React team has been working on support for [asynchronous rendering](#). We are excited about the new features this will enable, and we've learned that some changes will be required to the way we write React components.
+
+We're releasing version 16.3 primarily so that open source maintainers can start incorporating these changes into their libraries well in advance of the next major release. **However, you should not feel pressured to make changes to your application code yet.** We respect semver and will not ship breaking changes in a minor version!
+
+Read on to learn more about the release.
+
+## Official Context API
+
+For many years, React has offered an experimental API for context. Although it was a powerful tool, our advice was mostly to [avoid using it](/docs/context.html#why-not-to-use-context) because of potential problems with the API. We've always intended to replace the experimental API with a better one, and as of version 16.3 the new API is available!
```
sophiebits: Most of the why not context was actually focused on people using it as an alternative for props and it was discouraged because it can make code hard to follow, not because of its instability. Can we just say like "Although it was a powerful tool, its use was discouraged, in part because the API had inherent problems in its design from the start, and we always intended to replace the experimental API with a better one. In version 16.3, we're introducing a new context API that supports static type checking and deep updates as well as being more efficient."

content/blog/2018-02-07-react-v-16-3.md
```diff
@@ -0,0 +1,76 @@
...
+
+Version 16.3 adds a new option for managing refs that offers the convenience of a string ref without any of the downsides:
+`embed:16-3-release-blog-create-ref.js`
+
+> **Note:**
+>
+> Callback refs will continue to be supported in addition to the new `createRef` API.
+>
+> You don't need to replace callback refs in your components. They are slightly more flexible, so they will remain as an advanced feature.
+
```
sophiebits: What's the upgrade path? Is there a codemod?

content/blog/2018-02-07-react-v-16-3.md
```diff
@@ -0,0 +1,76 @@
...
+Many of these issues are exacerbated by a subset of the component lifecycles (`componentWillMount`, `componentWillReceiveProps`, and `componentWillUpdate`). These also happen to be the lifecycles that cause the most confusion within the React community. For these reasons, we are going to deprecate those methods in favor of better alternatives.
+
+We recognize that this change will impact many existing components. (At Facebook, we maintain more than 50,000 React components, and we can't tell our engineers to rewrite them either.) Because of this, the migration path will be as gradual as possible, and will provide escape hatches.
+
+> **Note:**
+>
+> Deprecation warnings will be enabled in version 16.4, **but the legacy lifecycles will continue to work until version 17**.
+>
+> After version 17, it will still be possible to use them, but they will be aliased with an "UNSAFE_" prefix to indicate that they might cause issues. We have also prepared an [automated script to rename them](https://github.com/reactjs/react-codemod#rename-unsafe-lifecycles) in existing code.
```
sophiebits: s/After version 17/Even in version 17/

content/blog/2018-02-07-react-v-16-3.md
```diff
@@ -0,0 +1,76 @@
...
+>
+> You don't need to replace callback refs in your components. They are slightly more flexible, so they will remain as an advanced feature.
+
+[Learn more about the new `createRef` API here.](#)
+
+## Component Lifecycle Changes
+
+React's class component API has been around for years with little change. However, as we add support for more advanced features (such as [error boundaries](/docs/react-component.html#componentdidcatch) and the upcoming [async rendering mode](#)) we stretch this model in ways that it was not originally intended.
+
+For example, with the current API, it is too easy to block the initial render with non-essential logic. In part this is because there are too many ways to accomplish a given task, and it can be unclear which is best. We've observed that the interrupting behavior of error handling is often not taken into consideration and can result in memory leaks—(something that will also impact the upcoming async rendering mode). The current class component API also complicates other efforts, like our work on a React compiler.
```
sophiebits: remove em dash

work on a React compiler -> "work on prototyping a React compiler"? don't want people to get hopes up too much. link it to https://twitter.com/trueadm/status/944908776896978946 too tho?

content/blog/2018-02-07-react-v-16-3.md
bvaughn: I think some reassurance that we won't break anything in a minor release belongs near the top, where people are most likely to see it.

Maybe I can shorten and soften the wording slightly to make it less alarmist?

bvaughn: Maybe something more like this:
---

For the past few months, the React team has been working on support for asynchronous rendering. We are excited about the new features it will enable.

We’ve also learned that some long-term changes will be required to the way we write React components. However, we respect semver and **will not ship breaking changes in a minor version!**

content/blog/2018-02-07-react-v-16-3.md
bvaughn: No codemod. I proposed one initially, but there's a codemod couldn't properly handle this case:
```js
class App extends React.Component {
  render() {
    return (
      <SomeOtherComponent
        renderThing={() =>
          <div ref='someRef' />
        }
      />
    );
  }
}
```

I guess we _could_ still write a codemod that bailed out on that case, but it didn't seem like there was a lot of support for one when I mentioned it during last week's sync.

content/blog/2018-02-07-react-v-16-3.md
bvaughn: Love it.

content/blog/2018-02-07-react-v-16-3.md
gaearon: Sorry if I was unclear; I *am* very supportive of the codemod that handles simple cases correctly, as long as it's solid in the bailouts.

bvaughn: I wasn't thinking of you specifically, just the general impression I got was a lack of interest.

I'm not necessarily volunteering to do it myself, but I'll add it to the quip as a non-16.3-release-blocking step. 😄

examples/16-3-release-blog-post/context-example.js
```diff
@@ -0,0 +1,26 @@
+// highlight-next-line
+const ThemeContext = React.createContext('light');
```
alexkrolick: should this be `theme` instead of `light`?

alexkrolick: (No, it should not 😛)

content/blog/2018-02-07-react-v-16-3.md
```diff
@@ -0,0 +1,81 @@
...
+Version 16.3 introduces a new context API that is more efficient and supports both static type checking and deep updates.
+
+> **Note**
+>
+> The old context API will keep working for all React 16.x releases, so you will have time to migrate.
+
+Here is an example illustrating how you might inject a "theme" using the new context API:
+`embed:16-3-release-blog-post/context-example.js`
+
+[Learn more about the new context API here.](#)
```
alexkrolick: `/docs/context.html#api` with #665

bvaughn: Ah, right. Thanks~

content/blog/2018-03-20-react-v-16-3.md
```diff
@@ -0,0 +1,96 @@
+---
+title: "React v16.3.0: New lifecycles and context API"
+author: [bvaughn]
+---
+
+This release includes a new class component lifecycle (`getDerivedStateFromProps`), a new `StrictMode` component, an official context API, a new ergonomic ref API, and a ref-forwarding API!
+
+For the past few months, the React team has been working on support for [asynchronous rendering](/blog/2018/03/01/sneak-peek-beyond-react-16.html). We are excited about the new features it will enable.
+
+We've also learned that some long-term changes will be required to the way we write React components. However, we respect semver and **will not ship breaking changes in a minor version**!
```
ivikash: Can we hyperlink the `semver` with https://semver.org/

content/blog/2018-03-20-react-v-16-3.md
```diff
@@ -0,0 +1,96 @@
...
+> The old context API will keep working for all React 16.x releases, so you will have time to migrate.
+
+Here is an example illustrating how you might inject a "theme" using the new context API:
+`embed:16-3-release-blog-post/context-example.js`
+
+[Learn more about the new context API here.](/docs/context.html#api)
+
+## `createRef` API
+
+Previously, React provided two ways for managing refs: the legacy string ref API and the callback API. Although the string ref API was the more convenient of the two, it had [several downsides](https://github.com/facebook/react/issues/1373) and so our official recomendation was to use the callback form instead.
```
ivikash: typo `recomendation` -> `recommendation`

content/docs/strict-mode.md
```diff
@@ -0,0 +1,88 @@
...
+
+When strict mode is enabled, React compiles a list of all class components using the unsafe lifecycles, and logs a warning message with information about these components, like so:
+
+![](../images/blog/strict-mode-unsafe-lifecycles-warning.png)
+
+Addressing the issues identified by strict mode _now_ will make it easier for you to take advantage of async rendering in future releases of React.
+
+### Warning about legacy string ref API usage
+
+Previously, React provided two ways for managing refs: the legacy string ref API and the callback API. Although the string ref API was the more convenient of the two, it had [several downsides](https://github.com/facebook/react/issues/1373) and so our official recomendation was to [use the callback form instead](https://reactjs.org/docs/refs-and-the-dom.html#legacy-api-string-refs).
```
ivikash: typo recomendation -> recommendation

content/blog/2018-03-20-react-v-16-3.md
ivikash: nitpick: Previously, React provided two ways for managing refs -> `Previously, React provided two ways of managing refs`

bvaughn:
Thanks~

bvaughn:
I've added a new page, entitled ["Forwarding Refs"](https://deploy-preview-587--reactjs.netlify.com/docs/forwarding-refs.html). I've also updated the reference section to include [`React.forwardRef`](https://deploy-preview-587--reactjs.netlify.com/docs/react-api.html#reactforwardref) and updated the [HOC caveats section](https://deploy-preview-587--reactjs.netlify.com/docs/higher-order-components.html#refs-arent-passed-through) too.

examples/16-3-release-blog-post/create-ref-example.js
```diff
@@ -0,0 +1,14 @@
+class MyComponent extends React.Component {
+  // highlight-next-line
+  divRef = React.createRef();
```
brunolemos: Maybe this should be called `inputRef`? PR: https://github.com/bvaughn/reactjs.org/pull/2

bvaughn: D'oh

content/docs/nav.yml
```diff
@@ -74,6 +76,8 @@
       title: Accessibility
     - id: code-splitting
       title: Code-Splitting
+    - id: strict-mode
+      title: StrictMode
```
gaearon: Maybe “Strict Mode” as a title? It seems like this section uses concept names rather than API names for titles. Eg we have “forwarding refs” not “forwardRef”.

examples/context/motivation-problem.js
```diff
@@ -0,0 +1,34 @@
...
+    );
+  }
+}
+
+class Message extends React.Component {
+  render() {
+    // highlight-range{1-3}
+    // The Message component must take `color` as as prop to pass it
+    // to the Button. Using context, the Button could connect to the
+    // color context on its own.
```
gaearon: I got stuck reading this example for 10 seconds looking for usage of context. The second sentence refers to context as if *we're already using it here*. I find this confusing.

I would prefer if it described *how it works in this example*, and explained the pain point. And the comment in the solution would describe the solution and why it solves the pain point.

examples/context/motivation-solution.js
```diff
@@ -0,0 +1,43 @@
+// highlight-range{1}
+const ColorContext = React.createContext();
```
gaearon: Do you think we could stick to our `ThemeContext` example?
I think `ColorContext` is a little bit less clear (why would a color need to be global? what color?)

(Note: I know it's based on old docs.)

content/docs/context.md
```diff
@@ -4,220 +4,92 @@ title: Context
...
+  - [Dynamic Context](#dynamic-context)
+- [Legacy API](#legacy-api)
 
-The vast majority of applications do not need to use context.
 
-If you want your application to be stable, don't use context. It is an experimental API and it is likely to break in future releases of React.
+## Motivation
 
-If you aren't familiar with state management libraries like [Redux](https://github.com/reactjs/redux) or [MobX](https://github.com/mobxjs/mobx), don't use context. For many practical applications, these libraries and their React bindings are a good choice for managing state that is relevant to many components. It is far more likely that Redux is the right solution to your problem than that context is the right solution.
+Context is designed to relieve the pain of passing props down through a deeply nested component tree. For example, in the code below we manually thread through a color prop in order to style the Button and Message components:
```
gaearon: I anticipate that people new to React will assume from this paragraph that you need to use context as soon as you pass a prop down more than a few levels. I don't think we want to encourage that.

I think this should be more specific and say that context is for data that can be considered "global" for a tree of React components, and immediately give a few examples: currently authenticated user, current theme, current language.

content/docs/context.md
```diff
@@ -4,220 +4,92 @@ title: Context
...
 
-      <Route exact path="/" component={Home} />
-      <Route path="/about" component={About} />
-      <Route path="/topics" component={Topics} />
-    </div>
-  </Router>
-);
+```js
+<Consumer>
+ { value => { /* render something based on the context value */ } }
```
gaearon: People will try to write

```js
{ value => { <div>hi</div> } }
```

Maybe let's remove the second curlies and just leave

```js
<Consumer>
  {value => /* render something based on the context value */}
</Consumer>
```

Also nit: no need for space before `value`.

content/docs/context.md
```diff
@@ -4,220 +4,92 @@ title: Context
...
 
-## Referencing Context in Lifecycle Methods
-
-If `contextTypes` is defined within a component, the following [lifecycle methods](/docs/react-component.html#the-component-lifecycle) will receive an additional parameter, the `context` object:
-
-- [`constructor(props, context)`](/docs/react-component.html#constructor)
-- [`componentWillReceiveProps(nextProps, nextContext)`](/docs/react-component.html#componentwillreceiveprops)
-- [`shouldComponentUpdate(nextProps, nextState, nextContext)`](/docs/react-component.html#shouldcomponentupdate)
-- [`componentWillUpdate(nextProps, nextState, nextContext)`](/docs/react-component.html#componentwillupdate)
+Requires a [function as a child](/docs/render-props.html#using-props-other-than-render). This function receives the current context value and returns a React node. It will be called whenever the Provider's value is updated.
```
gaearon: "whenever the Provider's value is updated" is a bit vague. We need to be very explicit that new and old Provider value is compared with `Object.is` algorithm, and whenever it changes, all consumers are re-rendered.

examples/context/motivation-solution.js
gaearon: Looking again, I see an example below uses Theme. It would be great to have a consistent example throughout this doc instead of using two very similar but different examples. Can you unify them?

content/docs/context.md
gaearon: I think we should add a section "When to Use Context" that is very clear about intended use cases for this feature, as well as use cases where we *don't* recommend it (e.g. don't just use it to avoid passing props a few levels down).

examples/context/theme-detailed-theme-context.js
```diff
@@ -0,0 +1,15 @@
...
+  },
+  dark: {
+    foreground: '#000000',
+    background: '#eeeeee',
+  },
+};
+
+// highlight-range{1-3}
+export const ThemeContext = React.createContext(
+  themes.dark
```
gaearon: By this point I forgot what the argument means. Maybe add `// default value`?

examples/context/theme-detailed-app.js
```diff
@@ -0,0 +1,30 @@
...
+          ? themes.light
+          : themes.dark,
+    }));
+  };
+
+  render() {
+    //highlight-range{2,6}
+    return (
+      <ThemeContext.Provider value={this.state.theme}>
+        <ThemedButton onClick={this.toggleTheme}>
```
gaearon: This example makes me feel like `ThemedButton` has to be a direct child of `ThemeContext.Provider`. I think it would be clearer if there was a component like

```js
function Toolbar(props) {
  return (
    <ThemedButton onClick={props.onThemeChange}>Change Theme</ThemedButton>
  );
}
```

in the middle

content/docs/context.md
```diff
@@ -4,220 +4,92 @@ title: Context
...
 
-  render() {
-    return this.props.children;
-  }
-}
-
-MediaQuery.childContextTypes = {
-  type: PropTypes.string
-};
-```
```
gaearon: There needs to be a section on [how to reference context in lifecycle methods](https://github.com/facebook/react/issues/12397#issuecomment-375501574). It can be deduced from this article but people will initially feel it's gross so we need to be explicit that it's a recommended pattern. Additionally it's important to show how it works with ref forwarding.

gaearon: There should also be an explanation on how to consume more than one context. Even if it links to a gist.

content/docs/context.md
```diff
@@ -4,220 +4,92 @@ title: Context
...
+## Legacy API
 
-The problem is, if a context value provided by component changes, descendants that use that value won't update if an intermediate parent returns `false` from `shouldComponentUpdate`. This is totally out of control of the components using context, so there's basically no way to reliably update the context. [This blog post](https://medium.com/@mweststrate/how-to-safely-use-react-context-b7e343eff076) has a good explanation of why this is a problem and how you might get around it.
+> The legacy context API was deprecated in React 16.3 and will be removed in version 17.
```
gaearon: It wasn't deprecated. It is just legacy for now. We'll call it deprecated when we actually add deprecation warnings, and this will happen in a future minor release.

alexkrolick:
@gaearon see https://github.com/bvaughn/reactjs.org/pull/4/files for context doc updates

examples/forwarding-refs/log-props-after.js
```diff
@@ -0,0 +1,26 @@
+function logProps(WrappedComponent) {
+  class LogProps extends React.Component {
+    componentWillReceiveProps(nextProps) {
```
nihgwu: I think we'd better use `componentDidUpdate` instead as `componentWillReceiveProps` is deprecated then

examples/forwarding-refs/log-props-before.js
```diff
@@ -0,0 +1,16 @@
+// highlight-next-line
+function logProps(WrappedComponent) {
+  class LogProps extends React.Component {
+    componentWillReceiveProps(nextProps) {
```
nihgwu: same here

examples/forwarding-refs/log-props-after.js
bvaughn: Nice catch! Updated.

examples/context/forwarding-refs.js
```diff
@@ -0,0 +1,12 @@
...
+    {children}
+  </button>
+);
+
+// highlight-range{1,3}
+export default React.forwardRef((props, ref) => (
+  <ThemeContext.Consumer>
+    {theme => <Button {...props} theme={theme} ref={ref} />}
+  </ThemeContext.Consumer>
+));
```
bvaughn: This example doesn't make sense, because `Button` is a functional component. I'll update it!

examples/context/motivation-problem.js
```diff
@@ -0,0 +1,23 @@
+const ThemedButton = props => {
+  //highlight-range{1}
+  return <Button theme={props.theme} />;
+};
+
+// An intermediate component
+const Toolbar = props => {
```
gaearon: We try to use the functional declaration syntax for functional components since it's familiar to more people. Would you mind doing that too?

```js
function Toolbar(props) {
  // ..
}
```

examples/context/motivation-solution.js
```diff
@@ -0,0 +1,33 @@
+// Create a theme context, defaulting to light theme
+// highlight-next-line
+const ThemeContext = React.createContext('light');
+
+const ThemedButton = props => {
+  // highlight-range{1,3-5}
+  // The ThemedButton receives the theme from context
+  return (
+    <ThemeContext.Consumer>
+      {theme => <Button theme={theme} />}
```
gaearon: You probably want `<Button {...props} theme={theme} />` here.

content/docs/context.md
```diff
@@ -4,220 +4,110 @@ title: Context
...
+</Consumer>
+```
 
-> Note:
->
-> As of React 16, `componentDidUpdate` no longer receives `prevContext`.
+A React component that subscribes to context changes.
 
-## Referencing Context in Stateless Functional Components
+Requires a [function as a child](/docs/render-props.html#using-props-other-than-render). The function receives the current context value and returns a React node. All consumers are re-rendered whenever the Provider value changes. Changes are determined by comparing the new and old values using [`Object.is`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is).
```
gaearon: Maybe "using the same algorithm as `Object.is`" so that people don't think it has to be polyfilled.

And let's link directly to Description (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is#Description) because the API of Object.is isn't relevant.

examples/context/motivation-solution.js
```diff
@@ -0,0 +1,33 @@
...
+    </ThemeContext.Consumer>
+  );
+};
+
+// An intermediate component
+const Toolbar = props => {
```
gaearon: Please use function declaration syntax

examples/context/multiple-contexts.js
```diff
@@ -0,0 +1,47 @@
...
+            <ProfilePage user={user} theme={theme} />
+          )}
+        </UserContext.Consumer>
+      )}
+    </ThemeContext.Consumer>
+  );
+};
+
+class App extends React.Component {
+  static propTypes = {
```
gaearon: This is experimental syntax. Please don't use it in stable docs. I'd also say propTypes aren't necessary for this doc.

examples/context/theme-detailed-app.js
```diff
@@ -0,0 +1,46 @@
...
+const Toolbar = props => {
+  return (
+    <ThemedButton onClick={props.changeTheme}>
+      Change Theme
+    </ThemedButton>
+  );
+};
+
+class App extends React.Component {
+  state = {
```
gaearon: This is experimental syntax. Please don't use it in stable docs.

examples/context/theme-detailed-app.js
```diff
@@ -0,0 +1,46 @@
...
+    theme: themes.light,
+  };
+
+  toggleTheme = () => {
```
gaearon: This is experimental syntax. Please don't use it in stable docs.

examples/context/theme-detailed-app.js
```diff
@@ -0,0 +1,46 @@
...
+  };
+
+  render() {
+    //highlight-range{1-3}
+    // The ThemedButton button inside the ThemeProvider
+    // uses the theme from state while the one outside uses
+    // the default dark theme
+    //highlight-range{3-5,7}
+    return (
+      <div>
```
gaearon: Can we come up with some semantic elements here? I understand they're not very relevant but React apps often gets criticized for using divs everywhere, and I'd prefer we don't invite this criticism with our own doc examples.

examples/context/theme-detailed-themed-button.js
```diff
@@ -0,0 +1,19 @@
+import {ThemeContext} from './theme-context';
+
+class ThemedButton extends React.Component {
```
gaearon: This could be a functional component, right? I think maybe we should use functional components in cases where a class isn't necessary.

examples/reference-react-forward-ref.js
```diff
@@ -0,0 +1,24 @@
...
+      // highlight-range{1-3}
+      return (
+        <WrappedComponent ref={forwardedRef} {...rest} />
+      );
+    }
+  }
+
+  // Intercept the "ref" and pass it as a custom prop, e.g. "forwardedRef"
+  // highlight-range{1-3}
+  function enhanceForwardRef(props, ref) {
```
gaearon: I find this name to be pretty confusing. I'd prefer anonymous functions if we can't give them good names that mean something and aren't just a mashup of other concepts around them.

content/docs/context.md
```diff
@@ -4,220 +4,110 @@ title: Context
...
-Don't do it.
+**themed-button.js**
+`embed:context/theme-detailed-themed-button.js`
 
-React has an API to update context, but it is fundamentally broken and you should not use it.
+**app.js**
+`embed:context/theme-detailed-app.js`
 
-The `getChildContext` function will be called when the state or props changes. In order to update data in the context, trigger a local state update with `this.setState`. This will trigger a new context and changes will be received by the children.
+## Consuming Multiple Contexts
```
gaearon: There should be a paragraph there explaining that to keep context re-rendering fast, React needs to make each context consumer a separate node in the tree. Because we'll get the question.

content/docs/context.md
```diff
@@ -4,220 +4,110 @@ title: Context
...
 
-```javascript
-import PropTypes from 'prop-types';
+`embed:context/multiple-contexts.js`
```
gaearon: After this example, we should point out that if two or more context values are often used together, you might want to consider creating your own render prop component that provides both. Then link to [this utility](https://github.com/FormidableLabs/react-context-composer/) as a simple way to do it.

content/docs/context.md
```diff
@@ -4,220 +4,110 @@ title: Context
...
 
-class MediaQuery extends React.Component {
-  constructor(props) {
-    super(props);
-    this.state = {type:'desktop'};
-  }
+## Accessing Context in Lifecycle Methods
```
gaearon: Again, needs an intro before the example. Explain that this is a relatively common use case. And that instead of adding context to every lifecycle method, you just need to pass it as a prop, and then work with it just like you'd normally work with a prop.

content/docs/context.md
```diff
@@ -4,220 +4,110 @@ title: Context
...
+`embed:context/lifecycles.js`
 
-  componentDidMount() {
-    const checkMediaQuery = () => {
-      const type = window.matchMedia("(min-width: 1025px)").matches ? 'desktop' : 'mobile';
-      if (type !== this.state.type) {
-        this.setState({type});
-      }
-    };
+## Forwarding Refs to Context Consumers
```
gaearon: Need to briefly explain the problem with previous example here.

content/docs/context.md
```diff
@@ -4,220 +4,110 @@ title: Context
...
-MediaQuery.childContextTypes = {
-  type: PropTypes.string
-};
-```
+## Legacy API
 
-The problem is, if a context value provided by component changes, descendants that use that value won't update if an intermediate parent returns `false` from `shouldComponentUpdate`. This is totally out of control of the components using context, so there's basically no way to reliably update the context. [This blog post](https://medium.com/@mweststrate/how-to-safely-use-react-context-b7e343eff076) has a good explanation of why this is a problem and how you might get around it.
+> Note
+> 
+> React previously shipped with an experimental context API. The old API will be supported in all 16.x releases, but applications using it should migrate to the new version. The legacy API will be removed in React 17. Read the [legacy context docs here](/docs/legacy-context.html).
```
gaearon: "in a future major React version", let's not be explicit here (we aren't actually sure yet)

content/docs/context.md
```diff
@@ -4,220 +4,110 @@ title: Context
...
-  }
-}
+**app.js**
+`embed:context/forwarding-refs-app.js`
 
-MediaQuery.childContextTypes = {
-  type: PropTypes.string
-};
-```
+## Legacy API
```
gaearon: Before "Legacy API", we need to add a "Caveats" section.

There, we should mention that pattern like `<Provider value={{ something: 'something' }}>` re-renders all consumers every time the provider parent re-renders because the `value` object is referentially different every time. If you didn't intend this, put the `value` into the parent component's `state` so that it doesn't change unintentionally between re-renders.

examples/reference-react-forward-ref.js
bvaughn: Let's do something like what I did in c41bb29 ?

bvaughn:
Btw @alexkrolick ~ would you like to respond to Dan's context feedback, or would you like me to pick it up?

examples/reference-react-forward-ref.js
gaearon: Sounds good to me

content/docs/context.md
alexkrolick: Would it make sense to put this in the Consumer API section?

gaearon: Maybe. What do you think? I generally prefer to separate "minimal necessary" information to use the API from all the details (that they user won't bump into later on), but maybe it's okay to do in one place. The main reason I suggested a separate section is because I wanted to have a "good" and "bad" example (even if they're one liners), and I felt it would they would steal the focus from the API reference if we put them there.

alexkrolick: OK, how about linking to Caveats/Gotchas after mentioning the Object.is comparison?

examples/context/theme-detailed-app.js
alexkrolick:
```jsx
  <React.Fragment>
    <ThemeContext.Provider value={this.state.theme}>
      <Toolbar changeTheme={this.toggleTheme} />
        </ThemeContext.Provider>
    <Contents>
      <ThemedButton />
    </Contents>
  </React.Fragment>
```

?

content/docs/context.md
gaearon: Sounds goode

bvaughn:
Added the new `getSnapshotBeforeUpdate` lifecycle to 16.3 release notes and added an example in the `React.Component` reference docs.

alexkrolick:
@bvaughn @gaearon Think I covered this round of context comments at https://github.com/bvaughn/reactjs.org/pull/5

Feel free to just edit directly - you have a better sense of where it needs to go.

bvaughn:
Thanks Alex. I've merged your PR into this branch.

content/docs/context.md
```diff
@@ -4,220 +4,130 @@ title: Context
...
-import PropTypes from 'prop-types';
+**app.js**
+`embed:context/forwarding-refs-app.js`
 
-class MediaQuery extends React.Component {
-  constructor(props) {
-    super(props);
-    this.state = {type:'desktop'};
-  }
+## Caveats
```
gaearon: Before Caveats, I think we should also add a section that explains how to create a higher order component for consuming context. This is a common sentiment and I think we should proactively address it: https://twitter.com/DankoKozar/status/977469702204162048?s=20

bvaughn: I'll add this.

bvaughn: Context comes before the Higher-Order Components reference page, but I'll cross link.

gaearon: Yeah, cyclical dependencies are pretty much unavoidable in the "Advanced" section

bvaughn:
I tweaked the note about what belongs in state. In general, I think we might want to consider re-writing this section in the near future- to reduce the endorsement of putting values on the instance (since this causes async trickiness). I don't consider that a 16.3 blocker though.

content/docs/state-and-lifecycle.md
```diff
@@ -251,7 +251,7 @@ Note how we save the timer ID right on `this`.
 
 While `this.props` is set up by React itself and `this.state` has a special meaning, you are free to add additional fields to the class manually if you need to store something that is not used for the visual output.
 
-If you don't use something in `render()`, it shouldn't be in the state.
+State is generally used to store values that are referenced by `render()` (for visual output) or by `getDerivedStateFromProps()` (to compare next and previous props values).
```
gaearon: By this point the user doesn’t know what `getDerivedStateFromProps` is. I don’t think it’s necessary to introduce it here. It just creates a sense of uneasiness (“did I miss something? Do I need to read the doc on it now?”) and potentially point the reader into a wrong direction (learning to sync props before learning to lift them up).

I understand what you were trying to do but maybe there’s another way?

bvaughn: That's fair. I had considered linking to the docs page for the method but that would take people out of the flow.

Maybe just delete the note entirely?

gaearon: Sounds good.

The above sentence can also change to

>While `this.props` is set up by React itself and `this.state` has a special meaning, you are free to add additional fields to the class manually if you need to store something that doesn’t participate in the data flow (like a timer ID).

content/docs/context.md
```diff
@@ -134,6 +136,16 @@ One issue with the render prop API is that refs don't automatically get passed t
 
 ## Caveats
 
+### `Consumer` Inline Function as a Child
+
```
gaearon: Bye bye

PR #590, opened 2018-02-06
Update gatsby

gaearon:
Fixes https://github.com/reactjs/reactjs.org/issues/589


rutzfutz:
I just tested the preview with Safari on an iPhone X running iOS 11.2.5 and the navigation bug is fixed. Well done!

gaearon:
Sweet, thank you so much for verifying!

PR #596, opened 2018-02-06
An update on async rendering

bvaughn:
Work in progress...

[Formatted "update on async rendering"](https://deploy-preview-596--reactjs.netlify.com/blog/2018/03/27/update-on-async-rendering.html)

### TODO before merging
- [x] Add a note below the [`getSnapshotBeforeUpdate ` example](https://deploy-preview-596--reactjs.netlify.com/blog/2018/03/27/update-on-async-rendering.html#reading-dom-properties-before-an-update) stating explicitly whether the polyfill supports this method.
- [x] Update the ["Open source project maintainers" section](https://deploy-preview-596--reactjs.netlify.com/blog/2018/03/27/update-on-async-rendering.html#open-source-project-maintainers) to list support for `getSnapshotBeforeUpdate` _only if_ we agree to move forward with reactjs/react-lifecycles-compat/pull/1.

PR #600, opened 2018-02-07
Increase Prettier line-width for example JS

bvaughn:
Builds on top of PR #596. (I made this a separate PR since all of the modifications were automated.)

This significantly improves the readability of JS examples on Desktop. It causes some line-wrapping on smaller mobile devices (screenshots below) but I think the code is still readable there.

screen shot 2018-02-07 at 8 20 07 am

screen shot 2018-02-07 at 8 20 16 am


gaearon:
(This would need to be merged into your branch, right?)

bvaughn:
Yeah, I plan on merging all 3 of these branches (the two blog posts and now this one) before merging, because I need to cross-link them and stuff.

gaearon:
Another option is to apply this on top of master separately. Then rebasing those PRs on top.

PR #601, opened 2018-02-08
Update refs-and-the-dom.md

iOiurson:
Clearer explanation on refs callbacks firing, see #12183

#12183


eps1lon:
Thanks for your work. It seems like github didn't properly reference the underlying issue which is why we missed this PR. #1069 already applied the same change.

PR #604, opened 2018-02-08
Simplify by making the input field "number"

githubshrek:
If we make the input field of type="number" we can simplify the example a lot. This is good from two points of view:
a) it's the right thing to do, to make a number field of type number, and let the field itself handle the type restriction to numeric
b) it will help the user to focus on the real complexity of the example (which is moving the state up), since it removes a lot of the irrelevant clutter

### DIFF
--- a/content/docs/lifting-state-up.md
+++ b/content/docs/lifting-state-up.md
@@ -263,7 +263,7 @@ class Calculator extends React.Component {
   render() {
     const scale = this.state.scale;
     const temperature = this.state.temperature;
-    const celsius = scale === 'f' ? ? toCelsius(temperature);
+    const celsius = scale === 'f' ? toCelsius(temperature);
     const fahrenheit = scale === 'c' ? toFarenheit(temperature);
 
     return (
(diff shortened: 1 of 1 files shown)


gaearon:
I think there was a reason why we did this but I struggle to recall it. There might've been some bug with controlled numeric inputs. Maybe it got fixed.

gaearon:
Does the same sequence of steps that I do in the GIF at the end of the post work with this change?

alexkrolick:
The old code might have been there to format the results:
> and it keeps the output rounded to the third decimal place

Doesn't seem super important to the overall example, though.

githubshrek:
Incidentally, I have to thank the ReactJS documentation in fact for teaching me about the new with HTML5 "number" input type. The first time I had ever seen it was on the Quick Start -> Forms page, in the Reservation example.

githubshrek:
I'm still quite new to the process here. Are these changes this planned to go live?

BartoszKlonowski:
Dears! Looking at the latest docs, I see that the changes introduced in this PR are no longer needed, as there's no such code and page now in the docs.
Let me close this PR as outdated.

PR #608, opened 2018-02-10
Update static-type-checking.md to fix the installation of Flow

aneequrrehman:
The second command won't work until the "flow" is added to scripts.


gaearon:
Are you sure? I think yarn will attempt to delegate to the Flow binary if the script with the same name is missing. Which is what we need.

aneequrrehman:
@gaearon Well, you are right. `yarn` does that, but `npm` doesn't.

gaearon:
Thanks for explaining.

PR #620, opened 2018-02-16
setState callback semantics

jordanpapaleo:
The `setState` callback does not pass previousState as an argument, it passes the current state, and uses it to set the next state.

I brought this up to Dan on [twitter](https://twitter.com/dan_abramov/status/964296359758221312) and I think we agreed that the rename of the props here is correct in the documentation.

NA

### DIFF
--- a/content/docs/state-and-lifecycle.md
+++ b/content/docs/state-and-lifecycle.md
@@ -362,8 +362,8 @@ To fix it, use a second form of `setState()` that accepts a function rather than
 
 ```js
 // Correct
-this.setState((prevState, props) => ({
-  counter: prevState.counter + props.increment
+this.setState((currentState, currentProps) => ({
+  counter: currentState.counter + currentProps.increment
 }));
 ```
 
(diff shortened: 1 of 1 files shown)


jordanpapaleo:
@gaearon Here is a fix for what I brought up in the docs today :) . Ill look through the rest of the docs for more uses of it.

yangshun:
@jordanpapaleo Good stuff! I think both #621 and #622 should be within this PR so that all mentions of `prevState` will be edited to `currentState` in one commit. In case the commits don't land at the same time, it would be confusing because one place mentions `prevState` and the other mentions `currentState`.

Also, the deploy preview failed, you may want to check it out.

jordanpapaleo:
Thanks!  So I edited these just from the link in the docs.  Im not sure how I would combine these PRs.  I do see the value of having them all merge together.

As for 622, can someone re run the build.  It looks like a `node-gyp` thing and I feel like I have fixed these by rerunning the build.  The PR is just text changes and I do not see anything that would cause an error.

PR #621, opened 2018-02-16
setState callback semantics

jordanpapaleo:
setState callback does not pass prevState, it passes currentState

See PR #620


eps1lon:
This has been resolved in the meantime.

We're now calling it simply `state`:

```js
incrementCount() {
  this.setState((state) => {
    // Important: read `state` instead of `this.state` when updating.
    return {count: state.count + 1}
  });
}
```

-- https://reactjs.org/docs/faq-state.html#what-is-the-difference-between-passing-an-object-or-a-function-in-setstate

PR #623, opened 2018-02-16
Document setting initial state with class property

ligaz:
Document initial state initialization with class property.
Add notes about its experimental syntax.

### DIFF
--- a/content/docs/react-without-es6.md
+++ b/content/docs/react-without-es6.md
@@ -71,6 +71,15 @@ class Counter extends React.Component {
 }
 ```
 
+Alternatively if your initial state does not depend on `props` you can use the **experimental** syntax for class property initializer:
+```javascript
+class Counter extends React.Component {
+  state = {message: 'Hello!'};
+  // ...
+}
+```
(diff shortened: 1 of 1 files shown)


gaearon:
I don’t think this is a good place to put it. This page is about React *without* modern syntax. We only mention ES6 version here as a reference point for comparison, but don’t actually want to show more modern syntax on this page. Does that make sense?

ligaz:
It makes sense but this is the first page that I ended up while trying to find is this already documented. Actually once I issue the PR I found out that it is  in `react-without-es6.md`.

Here is my workflow:
* Go to [https://reactjs.org](https://reactjs.org/).
* Type state in the search box.
* Look at the results:
screen shot 2018-02-16 at 17 51 49

* Click on `Setting the Initial State` because this is exactly what I'm interested in. 

Should I move this into `state-and-lifecycle.md`?

ligaz:
@gaearon How should we proceed with this one?

alexkrolick:
There's already a PR to add that to State and Lifecycle: https://github.com/reactjs/reactjs.org/pull/557/files

Let's run with that one.

ligaz:
@alexkrolick This PR #557 does not have an example and according to guidelines it should mention that class properties is an experimental feature.

alexkrolick:
Correct, that's why it hasn't been merged. Feel free to collaborate.

PR #626, opened 2018-02-18
Clarify callback parameters

reccanti:
Clarifies the callback parameters for React.Children.map() and React.Children.forEach() by noting that they can also contain an optional `index` parameter as a second argument.

I'm not sure if the way I formatted the parameters in the function is correct. I noticed that before, `thisArg` was wrapped in brackets. Was that to signify it was optional? Feel free to correct me on that.

Preview of the change:
image

### DIFF
--- a/content/docs/reference-react.md
+++ b/content/docs/reference-react.md
@@ -154,15 +154,15 @@ Verifies the object is a React element. Returns `true` or `false`.
 #### `React.Children.map`
 
 ```javascript
-React.Children.map(children, function[(thisArg)])
+React.Children.map(children, function(thisArg, [index]))
 ```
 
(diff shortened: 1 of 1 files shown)


content/docs/reference-react.md
```diff
@@ -154,15 +154,15 @@ Verifies the object is a React element. Returns `true` or `false`.
 #### `React.Children.map`
 
 ```javascript
-React.Children.map(children, function[(thisArg)])
+React.Children.map(children, function(thisArg, [index]))
```
yangshun: The comma for optional parameters should be within the square bracket - 

```
React.Children.map(children, function(thisArg [,index]))
```

Issue #629, opened 2018-02-20
Formatting Issue on Fragments Page


shahzaibkhalid:
On Fragments Page, enlist in Advanced Guides in React Docs, (ref. https://reactjs.org/docs/fragments.html) there's a formatting issue as highlighted in the attached image screenshot.
<img width="1397" alt="screen shot 2018-02-20 at 9 17 07 pm" src="https://user-images.githubusercontent.com/18225515/36435608-e6327674-1683-11e8-9aeb-3492c11084ab.png">

gaearon:
Seems intentional but I agree it looks silly in a sentence. Please send a PR with your suggestion?

shahzaibkhalid:
Yes sure. I will send a PR.

shahzaibkhalid:
@gaearon Please review the PR.

harish-sethuraman:
![Screenshot 2022-01-02 at 11 45 12 AM](https://user-images.githubusercontent.com/32865581/147867806-5dc8512e-9c87-49d8-9f18-50f0b633f99e.png)
Seems like the text has been changed and the issue is no longer reproducible. Sorry for the late turnaround time.

Thanks 😄

PR #630, opened 2018-02-20
fix formatting issue on Fragments section

shahzaibkhalid:
Added a fix for formatting issue on Fragments section
Existing reference issue is: 

#629

### DIFF
--- a/content/docs/fragments.md
+++ b/content/docs/fragments.md
@@ -66,7 +66,7 @@ results in a `<Table />` output of:
 </table>
 ```
 
-So, we introduce `Fragments`.
+So, we introduce `Fragment`'s usage as below.
 
 ## Usage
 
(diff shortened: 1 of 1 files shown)


content/docs/fragments.md
```diff
@@ -66,7 +66,7 @@ results in a `<Table />` output of:
 </table>
 ```
 
-So, we introduce `Fragment`s.
+So, we introduce `Fragments`.
```
alexkrolick: Let's kill the monospace completely here to avoid confusion

yvesgurcan: I agree! "`Fragments`" isn't valid code in this context, so not having any kind of special formatting is probably better. If we still wanted to have the word stand out, though, we could say: "`Fragment` components". What do you think, @alexkrolick?

harish-sethuraman:
Screenshot 2022-01-02 at 11 45 12 AM
Seems like the text has been changed and the issue is no longer reproducible. Sorry for the late turnaround time.

Thanks for sending PR

PR #631, opened 2018-02-20
add Reframe to "Starter Kits"

brillout:
How about creating a new section "Frameworks"? Next.js, Gatsby, and Reframe are frameworks and not really starter kits.

Also, what does it take to become "Recommended by the React Team"? :eyes:

Reframe is new but I'm committed to maintain and further develop it.


gaearon:
We've removed this page.

Issue #594, opened 2018-02-06
Accessibility: Add a way to escape the jsx docblocks using the keyboard


knittingcodemonkey:
On the homepage of https://reactjs.org/, there are LIVE JSX EDITORs. For keyboard-only users, once you tab into them, how does one get out?

EarthSchlange:
I would like to take on this issue if possible.

alexkrolick:
@EarthSchlange great, it's yours! If you change your mind, just let us know and we'll remove the `in-progress` label.

PR #635, opened 2018-02-22
Fixed Accessiblity Issue jsx docblocks - 594

EarthSchlange:
set ignoreTabKey to true so that keyboard only users can tab out of jsx docblocks fixing issue #594


alexkrolick:
Netlify broke, but pulled locally to confirm.

PR #637, opened 2018-02-22
Add React.createRef doc

trueadm:
This PR adds the new `React.createRef()` documentation to the refs page in the documentation.

### DIFF
--- a/content/docs/refs-and-the-dom.md
+++ b/content/docs/refs-and-the-dom.md
@@ -94,12 +94,13 @@ React will assign the `value` property with the DOM element when the component m
 
 When the `ref` attribute is used on a custom component declared as a class, the `ref` object receives the mounted instance of the component as its `value`. For example, if we wanted to wrap the `CustomTextInput` above to simulate it being clicked immediately after mounting:
 
-```javascript{4,12}
+```javascript{4,7,12}
 class AutoFocusTextInput extends React.Component {
   constructor(props) {
     super(props);
     this.textInput = React.createRef();
   }
+
   componentDidMount() {
     this.textInput.value.focusTextInput();
   }
@@ -124,7 +125,7 @@ class CustomTextInput extends React.Component {
 
 **You may not use the `ref` attribute on functional components** because they don't have instances:
 
-```javascript{1,8,12}
+```javascript{1,8,13}
 function MyFunctionalComponent() {
   return <input />;
 }
@@ -167,7 +168,7 @@ function CustomTextInput(props) {
         onClick={handleClick}
       />
     </div>
-  );  
+  );
 }
 ```
 
@@ -211,7 +212,7 @@ This works even though `CustomTextInput` is a functional component. Unlike the s
 
 Another benefit of this pattern is that it works several components deep. For example, imagine `Parent` didn't need that DOM node, but a component that rendered `Parent` (let's call it `Grandparent`) needed access to it. Then we could let the `Grandparent` specify the `inputRef` prop to the `Parent`, and let `Parent` "forward" it to the `CustomTextInput`:
 
-```javascript{4,12,21,25}
+```javascript{4,12,20,24}
 function CustomTextInput(props) {
   return (
     <div>
@@ -227,7 +228,6 @@ function Parent(props) {
     </div>
   );
 }
-
 
 class Grandparent extends React.Component {
   constructor(props) {
--- a/content/docs/refs-and-the-dom.md
+++ b/content/docs/refs-and-the-dom.md
@@ -33,22 +33,10 @@ Your first inclination may be to use refs to "make things happen" in your app. I
 
 >**Note:**
 >
->The examples below have updated to use the new `React.createRef()` API, introduced in React 16.3.
+>The examples below have updated to use the `React.createRef()` API introduced in React 16.3.
 
 
-Refs can be created using `React.createRef()` and are commonly assigned to a component class instance in the constructor or via a property initializer.
-
-```javascript{4}
-class MyComponent extends React.Component {
-  constructor(props) {
-    super(props);
-    this.myRef = React.createRef();
-  }
-}
-```
-
-Refs can then be attached to React elements via the `ref` attribute by passing the property on the class instance as the value to that attribute.
-
+Refs can be created using `React.createRef()` and are commonly assigned to an instance property when a component is constructed.
 
 ```javascript{4,7}
 class MyComponent extends React.Component {
@@ -62,9 +50,11 @@ class MyComponent extends React.Component {
 }
 ```
 
+Refs can then be attached to React elements via the `ref` attribute by passing the property on the class instance as the value to that attribute.
+
 When the `ref` attribute is used on an HTML element, the `ref` created in the constructor with `React.createRef()` receives the underlying DOM element as its `value` property. For example, this code uses a `ref` to store a reference to a DOM node:
 
-```javascript{5,11,21}
+```javascript{5,12,22}
 class CustomTextInput extends React.Component {
   constructor(props) {
     super(props);
@@ -75,6 +65,7 @@ class CustomTextInput extends React.Component {
 
   focusTextInput() {
     // Explicitly focus the text input using the raw DOM API
+    // Note: we're accessing "value" to get the DOM node
     this.textInput.value.focus();
   }
 
@@ -257,20 +248,18 @@ All things considered, we advise against exposing DOM nodes whenever possible, b
 
 ### Callback Refs
 
-The above examples use `React.createRef()`, which can be thought of as "object refs". React also supports an alternative approach called "callback refs", which offer the same functionality as object refs, but also also give more fine-grain control over when refs are set and unset. Callback refs work in much the same way as object refs.
+React also supports another way to set refs called "callback refs", which give more fine-grain control over when refs are set and unset.
 
-With callback refs, instead of creating a ref with `React.createRef()`, you pass normal JavaScript functions to `ref` attributes. When the `ref` attribute is used on an HTML element, the `ref` callback receives the underlying DOM element as its argument. For example, this code uses the `ref` callback to store a reference to a DOM node:
+Instead of passing a `ref` attribute created by `createRef()`, you pass a function. The function receives the React component instance or HTML DOM element as its argument, which can be stored and accessed elsewhere. This example uses the `ref` callback to store a reference to a DOM node:
 
-```javascript{8,9,19}
+```javascript{6,17}
 class CustomTextInput extends React.Component {
   constructor(props) {
     super(props);
-    this.focusTextInput = this.focusTextInput.bind(this);
-  }
-
-  focusTextInput() {
-    // Explicitly focus the text input using the raw DOM API
-    this.textInput.focus();
+    this.focusTextInput = () => {
+      // Explicitly focus the text input using the raw DOM API
+      this.textInput.focus();
+    };
   }
 
   render() {
@@ -294,7 +283,7 @@ class CustomTextInput extends React.Component {
 
 React will call the `ref` callback with the DOM element when the component mounts, and call it with `null` when it unmounts. `ref` callbacks are invoked before `componentDidMount` or `componentDidUpdate` lifecycle hooks.
 
-Using the `ref` callback just to set a property on the class is a common pattern for accessing DOM elements. The preferred way is to set the property in the `ref` callback like in the above example. There is even a shorter way to write it: `ref={input => this.textInput = input}`. 
+Using the `ref` callback to set a property on the class is a common pattern for accessing DOM elements. The preferred way is to set the property in the `ref` callback like in the above example. There is even a shorter way to write it: `ref={input => this.textInput = input}`. 
 
 You can pass callback refs between components like you can with object refs that were created with `React.createRef()`.
 
--- a/content/docs/refs-and-the-dom.md
+++ b/content/docs/refs-and-the-dom.md
@@ -111,7 +111,7 @@ React will assign the `value` property with the DOM element when the component m
 
 If we wanted to wrap the `CustomTextInput` above to simulate it being clicked immediately after mounting, we could use a ref to get access to the custom input and call its `focusTextInput` method manually:
 
-```javascript{4,7,12}
+```javascript{4,8,13}
 class AutoFocusTextInput extends React.Component {
   constructor(props) {
     super(props);
--- a/content/docs/refs-and-the-dom.md
+++ b/content/docs/refs-and-the-dom.md
@@ -31,9 +31,211 @@ Your first inclination may be to use refs to "make things happen" in your app. I
 
 ### Adding a Ref to a DOM Element
 
+>**Note:**
+>
+>The examples below have updated to use the new `React.createRef()` API, introduced in React 16.3.
+
+
 React supports a special attribute that you can attach to any component. The `ref` attribute takes a callback function, and the callback will be executed immediately after the component is mounted or unmounted.
 
-When the `ref` attribute is used on an HTML element, the `ref` callback receives the underlying DOM element as its argument. For example, this code uses the `ref` callback to store a reference to a DOM node:
+When the `ref` attribute is used on an HTML element, the `ref` created in the constructor with `React.createRef()` receives the underlying DOM element as its `value` property. For example, this code uses a `ref` to store a reference to a DOM node:
+
+```javascript{8,9,19}
+class CustomTextInput extends React.Component {
+  constructor(props) {
+    super(props);
+    // create a ref to store the textInput DOM element
+    this.textInput = React.createRef();
+    this.focusTextInput = this.focusTextInput.bind(this);
+  }
+
+  focusTextInput() {
+    // Explicitly focus the text input using the raw DOM API
+    this.textInput.value.focus();
+  }
+
+  render() {
+    // tell React that we want the associate the <input> ref
+    // to the `textInput` that we created in the constructor
+    return (
+      <div>
+        <input
+          type="text"
+          ref={this.textInput} />
+        <input
+          type="button"
+          value="Focus the text input"
+          onClick={this.focusTextInput}
+        />
+      </div>
+    );
+  }
+}
+```
+
+React will assign the `value` property with the DOM element when the component mounts, and assign it back to `null` when it unmounts. `ref` updates happen before `componentDidMount` or `componentDidUpdate` lifecycle hooks.
+
+### Adding a Ref to a Class Component
+
+When the `ref` attribute is used on a custom component declared as a class, the `ref` object receives the mounted instance of the component as its `value`. For example, if we wanted to wrap the `CustomTextInput` above to simulate it being clicked immediately after mounting:
+
+```javascript{3,9}
+class AutoFocusTextInput extends React.Component {
+  constructor(props) {
+    super(props);
+    this.textInput = React.createRef();
+  }
+  componentDidMount() {
+    this.textInput.value.focusTextInput();
+  }
+
+  render() {
+    return (
+      <CustomTextInput ref={this.textInput} />
+    );
+  }
+}
+```
+
+Note that this only works if `CustomTextInput` is declared as a class:
+
+```js{1}
+class CustomTextInput extends React.Component {
+  // ...
+}
+```
+
+### Refs and Functional Components
+
+**You may not use the `ref` attribute on functional components** because they don't have instances:
+
+```javascript{1,7}
+function MyFunctionalComponent() {
+  return <input />;
+}
+
+class Parent extends React.Component {
+  constructor(props) {
+    super(props);
+    this.textInput = React.createRef();
+  }
+  render() {
+    // This will *not* work!
+    return (
+      <MyFunctionalComponent ref={this.textInput} />
+    );
+  }
+}
+```
+
+You should convert the component to a class if you need a ref to it, just like you do when you need lifecycle methods or state.
+
+You can, however, **use the `ref` attribute inside a functional component** as long as you refer to a DOM element or a class component:
+
+```javascript{2,3,6,13}
+function CustomTextInput(props) {
+  // textInput must be declared here so the ref can refer to it
+  let textInput = React.createRef();
+
+  function handleClick() {
+    textInput.value.focus();
+  }
+
+  return (
+    <div>
+      <input
+        type="text"
+        ref={textInput} />
+      <input
+        type="button"
+        value="Focus the text input"
+        onClick={handleClick}
+      />
+    </div>
+  );  
+}
+```
+
+### Exposing DOM Refs to Parent Components
+
+In rare cases, you might want to have access to a child's DOM node from a parent component. This is generally not recommended because it breaks component encapsulation, but it can occasionally be useful for triggering focus or measuring the size or position of a child DOM node.
+
+While you could [add a ref to the child component](#adding-a-ref-to-a-class-component), this is not an ideal solution, as you would only get a component instance rather than a DOM node. Additionally, this wouldn't work with functional components.
+
+Instead, in such cases we recommend exposing a special prop on the child. The child would take a prop with an arbitrary name (e.g. `inputRef`) and attach it to the DOM node as a `ref` attribute. This lets the parent pass its ref to the child's DOM node through the component in the middle.
+
+This works both for classes and for functional components.
+
+```javascript{4,13}
+function CustomTextInput(props) {
+  return (
+    <div>
+      <input ref={props.inputRef} />
+    </div>
+  );
+}
+
+class Parent extends React.Component {
+  constructor(props) {
+    super(props);
+    this.inputElement = React.createRef();
+  }
+  render() {
+    return (
+      <CustomTextInput inputRef={this.inputElement} />
+    );
+  }
+}
+```
+
+In the example above, `Parent` passes its class property `this.inputElement` as an `inputRef` prop to the `CustomTextInput`, and the `CustomTextInput` passes the same ref as a special `ref` attribute to the `<input>`. As a result, `this.inputElement.value` in `Parent` will be set to the DOM node corresponding to the `<input>` element in the `CustomTextInput`.
+
+Note that the name of the `inputRef` prop in the above example has no special meaning, as it is a regular component prop. However, using the `ref` attribute on the `<input>` itself is important, as it tells React to attach a ref to its DOM node.
+
+This works even though `CustomTextInput` is a functional component. Unlike the special `ref` attribute which can [only be specified for DOM elements and for class components](#refs-and-functional-components), there are no restrictions on regular component props like `inputRef`.
+
+Another benefit of this pattern is that it works several components deep. For example, imagine `Parent` didn't need that DOM node, but a component that rendered `Parent` (let's call it `Grandparent`) needed access to it. Then we could let the `Grandparent` specify the `inputRef` prop to the `Parent`, and let `Parent` "forward" it to the `CustomTextInput`:
+
+```javascript{4,12,22}
+function CustomTextInput(props) {
+  return (
+    <div>
+      <input ref={props.inputRef} />
+    </div>
+  );
+}
+
+function Parent(props) {
+  return (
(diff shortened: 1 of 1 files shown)


bvaughn:
First deploy attempt failed because of the same transient Sharp.node issue:
> Error: Cannot find module '../build/Release/sharp.node'

This module is such a pain. Restarting the build without cache.

content/docs/refs-and-the-dom.md
```diff
@@ -31,30 +31,37 @@ Your first inclination may be to use refs to "make things happen" in your app. I
...
 
+>**Note:**
+>
+>The examples below have updated to use the new `React.createRef()` API, introduced in React 16.3.
+
+
 React supports a special attribute that you can attach to any component. The `ref` attribute takes a callback function, and the callback will be executed immediately after the component is mounted or unmounted.
 
-When the `ref` attribute is used on an HTML element, the `ref` callback receives the underlying DOM element as its argument. For example, this code uses the `ref` callback to store a reference to a DOM node:
+When the `ref` attribute is used on an HTML element, the `ref` created in the constructor with `React.createRef()` receives the underlying DOM element as its `value` property. For example, this code uses a `ref` to store a reference to a DOM node:
```
bvaughn: I think this introduction to `createRef` could be clearer.

Maybe split this into two sentences? First, introduce `React.createRef()` and explain that you use it either in the constructor or via a property initializer. Then explain that you pass it to the `ref` attribute etc.

content/docs/refs-and-the-dom.md
```diff
@@ -31,30 +31,37 @@ Your first inclination may be to use refs to "make things happen" in your app. I
...
 
 ```javascript{8,9,19}
 class CustomTextInput extends React.Component {
   constructor(props) {
     super(props);
+    // create a ref to store the textInput DOM element
+    this.textInput = React.createRef();
```
bvaughn: These changes break the line highlighting above
> javascript{8,9,19}

You'll want to update the line number offsets.

Might also be a nice time to move these examples into the ["examples" folder](https://github.com/reactjs/reactjs.org/tree/master/examples) so Prettier can manage the formatting for us. (Totally optional though.) This would allow you to use the nicer "highlight-line" and "highlight-range" directives.

content/docs/refs-and-the-dom.md
```diff
@@ -31,30 +31,37 @@ Your first inclination may be to use refs to "make things happen" in your app. I
...
     // Explicitly focus the text input using the raw DOM API
-    this.textInput.focus();
+    this.textInput.value.focus();
   }
 
   render() {
-    // Use the `ref` callback to store a reference to the text input DOM
-    // element in an instance field (for example, this.textInput).
+    // tell React that we want the associate the <input> ref
+    // to the `textInput` that we created in the constructor
```
bvaughn: "associate _with_" rather than "associate _to_"?

Or maybe better yet:
> // Tell React to store a reference to the text input DOM
> // element in the `textInput` ref we created in the constructor.

content/docs/refs-and-the-dom.md
```diff
@@ -66,24 +73,25 @@ class CustomTextInput extends React.Component {
...
+React will assign the `value` property with the DOM element when the component mounts, and assign it back to `null` when it unmounts. `ref` updates happen before `componentDidMount` or `componentDidUpdate` lifecycle hooks.
 
 ### Adding a Ref to a Class Component
 
-When the `ref` attribute is used on a custom component declared as a class, the `ref` callback receives the mounted instance of the component as its argument. For example, if we wanted to wrap the `CustomTextInput` above to simulate it being clicked immediately after mounting:
+When the `ref` attribute is used on a custom component declared as a class, the `ref` object receives the mounted instance of the component as its `value`. For example, if we wanted to wrap the `CustomTextInput` above to simulate it being clicked immediately after mounting:
 
 ```javascript{3,9}
 class AutoFocusTextInput extends React.Component {
+  constructor(props) {
```
bvaughn: Same comment about the line highlights being off here.

content/docs/refs-and-the-dom.md
```diff
@@ -107,11 +115,14 @@ function MyFunctionalComponent() {
 }
 
 class Parent extends React.Component {
+  constructor(props) {
```
bvaughn: Line highlights are off.

content/docs/refs-and-the-dom.md
```diff
@@ -165,17 +176,19 @@ function CustomTextInput(props) {
...
+    super(props);
+    this.inputElement = React.createRef();
+  }
```
bvaughn: Line highlights are off.

content/docs/refs-and-the-dom.md
```diff
@@ -202,24 +215,91 @@ function Parent(props) {
...
   render() {
     return (
-      <Parent
-        inputRef={el => this.inputElement = el}
-      />
+      <Parent inputRef={this.inputElement} />
```
bvaughn: Line highlights are off

content/docs/refs-and-the-dom.md
```diff
@@ -202,24 +215,91 @@ function Parent(props) {
...
 
 All things considered, we advise against exposing DOM nodes whenever possible, but this can be a useful escape hatch. Note that this approach requires you to add some code to the child component. If you have absolutely no control over the child component implementation, your last option is to use [`findDOMNode()`](/docs/react-dom.html#finddomnode), but it is discouraged.
 
+### Callback Refs
+
+The above examples use `React.createRef()`, which can be thought of as "object refs". React also supports an alternative approach called "callback refs", which offer the same functionality as object refs, but also also give more fine-grain control over when refs are set and unset. In much the same way object refs work, callback refs can be used in a similar approach.
```
bvaughn: > In much the same way object refs work, callback refs can be used in a similar approach.

This sentence seems redundant. Maybe:

> Callback refs work in much the same way as object refs.

content/docs/refs-and-the-dom.md
```diff
@@ -202,24 +215,91 @@ function Parent(props) {
...
+
+With callback refs, instead of creating a ref with `React.createRef()`, you instead pass normal JavaScript functions to `ref` attributes. When the `ref` attribute is used on an HTML element, the `ref` callback receives the underlying DOM element as its argument. For example, this code uses the `ref` callback to store a reference to a DOM node:
```
bvaughn: Nit: Too many "instead"s in the first sentence. Remove the second one?

content/docs/refs-and-the-dom.md
```diff
@@ -202,24 +215,91 @@ function Parent(props) {
...
+  }
+
+  render() {
+    // Use the `ref` callback to store a reference to the text input DOM
+    // element in an instance field (for example, this.textInput).
+    return (
+      <div>
+        <input
+          type="text"
+          ref={(input) => { this.textInput = input; }} />
```
bvaughn: Curious if theres' any reason we bind `focusTextInput` but use an inline arrow function for the ref-setter? Maybe consistency one way or the other might be nice?

In this case, maybe bind the ref setter so the next paragraph is more accurate? (The inline setter will get called for each update too.)

Also the third paragraph, showing the "shorter way" would maybe be more meaningful.

trueadm: The examples were the same as before, just with the new API.

bvaughn: I still think we could improve them 😄

content/docs/refs-and-the-dom.md
```diff
@@ -31,30 +31,61 @@ Your first inclination may be to use refs to "make things happen" in your app. I
 
 ### Adding a Ref to a DOM Element
 
-React supports a special attribute that you can attach to any component. The `ref` attribute takes a callback function, and the callback will be executed immediately after the component is mounted or unmounted.
+>**Note:**
+>
+>The examples below have updated to use the new `React.createRef()` API, introduced in React 16.3.
```
alexkrolick: I think this will read a little weird in a few months when `createRef` is no longer new.

> The examples below use the `React.createRef()` API introduced in React 16.3.

content/docs/refs-and-the-dom.md
```diff
@@ -31,30 +31,61 @@ Your first inclination may be to use refs to "make things happen" in your app. I
...
 
-When the `ref` attribute is used on an HTML element, the `ref` callback receives the underlying DOM element as its argument. For example, this code uses the `ref` callback to store a reference to a DOM node:
 
-```javascript{8,9,19}
+Refs can be created using `React.createRef()` and are commonly assigned to a component class instance in the constructor or via a property initializer.
```
alexkrolick: It's a bit hard to grok this initial example without context. How about leading with the second one that includes the render method?

Even better may be the canonical "autofocus an input" example that shows how to use a ref end-to-end.

content/docs/refs-and-the-dom.md
```diff
@@ -202,24 +239,91 @@ function Parent(props) {
...
 
 All things considered, we advise against exposing DOM nodes whenever possible, but this can be a useful escape hatch. Note that this approach requires you to add some code to the child component. If you have absolutely no control over the child component implementation, your last option is to use [`findDOMNode()`](/docs/react-dom.html#finddomnode), but it is discouraged.
 
+### Callback Refs
+
+The above examples use `React.createRef()`, which can be thought of as "object refs". React also supports an alternative approach called "callback refs", which offer the same functionality as object refs, but also also give more fine-grain control over when refs are set and unset. Callback refs work in much the same way as object refs.
```
alexkrolick: IMO this part could be more direct:

> React also supports another way to set refs called "callback refs", which give more fine-grain control over when refs are set and unset.

> Instead of passing a `ref` attribute created by `createRef`, you pass a function. The function receives the React component instance or HTML DOM element as its argument, which can be stored and accessed elsewhere. This example uses the `ref` callback to store a reference to a DOM node:

> `<example>`

content/docs/refs-and-the-dom.md
```diff
@@ -202,24 +239,91 @@ function Parent(props) {
...
+        />
+      </div>
+    );
+  }
+}
+```
+
+React will call the `ref` callback with the DOM element when the component mounts, and call it with `null` when it unmounts. `ref` callbacks are invoked before `componentDidMount` or `componentDidUpdate` lifecycle hooks.
+
+Using the `ref` callback just to set a property on the class is a common pattern for accessing DOM elements. The preferred way is to set the property in the `ref` callback like in the above example. There is even a shorter way to write it: `ref={input => this.textInput = input}`.
```
alexkrolick: Why "just"?

content/docs/refs-and-the-dom.md
```diff
@@ -31,30 +31,61 @@ Your first inclination may be to use refs to "make things happen" in your app. I
...
     super(props);
+    // create a ref to store the textInput DOM element
+    this.textInput = React.createRef();
     this.focusTextInput = this.focusTextInput.bind(this);
   }
 
   focusTextInput() {
     // Explicitly focus the text input using the raw DOM API
-    this.textInput.focus();
+    this.textInput.value.focus();
```
alexkrolick: ⚠️ This is big change from the previous callback API and it's easy to miss - should probably call this out more for people already familiar with the old way and just looking to see what's new

content/docs/refs-and-the-dom.md
bvaughn: Yeah, we could keep the sentences separate (I like them) but maybe just have one code example (that highlights both lines)

alexkrolick: > commonly assigned to a **component class instance** in the constructor or via a property initializer.

Shouldn't it be instance _property_?

> Refs can be created using `React.createRef()` and are commonly assigned to an instance property when a component is constructed. _(Avoids going into class property vs constructor details)._

bvaughn: I dig that.

trueadm:
Thanks for the feedback @alexkrolick @bvaughn. I've addressed (hopefully) all the feedback. Let me know if I missed something.

alexkrolick:
Nice, I'll look through it again 👍

alexkrolick:
@trueadm making a few more suggestions in PRs

trueadm:
@alexkrolick Sorry, I didn't realize it was a work in progress :P

alexkrolick:
@trueadm all done 😄

trueadm:
@alexkrolick Thanks for the help! @bvaughn want to give it once last proof check? If I merge, will this go to the versioned 16.3 site or do we have to wait?

bvaughn:
Will try to look tomorrow on my flight back.

Don't merge until we release 16.3 though, or it will go live on the site (before the actual API is live) 😄

content/docs/refs-and-the-dom.md
```diff
@@ -11,6 +11,8 @@ redirect_from:
 permalink: docs/refs-and-the-dom.html
 ---
 
+Refs provide a way to access DOM nodes and React elements created by React by getting a _reference_ to the element in the render method.
```
bvaughn: I think there's something awkward about this wording.

"_React elements created by React_" sounds a bit redundant and "_getting a reference to the element in the render method_" sounds potentially misleading (since refs are rarely used _in_ `render` in my experience).

Maybe something simpler? Like:

> Refs provide a way to access DOM nodes or React elements created in the `render` method.

content/docs/refs-and-the-dom.md
```diff
@@ -29,32 +31,69 @@ For example, instead of exposing `open()` and `close()` methods on a `Dialog` co
 
 Your first inclination may be to use refs to "make things happen" in your app. If this is the case, take a moment and think more critically about where state should be owned in the component hierarchy. Often, it becomes clear that the proper place to "own" that state is at a higher level in the hierarchy. See the [Lifting State Up](/docs/lifting-state-up.html) guide for examples of this.
 
-### Adding a Ref to a DOM Element
+>**Note:**
+>
+>The examples below have updated to use the `React.createRef()` API introduced in React 16.3.
```
bvaughn: > The examples below have updated to use...

Missing word? Should be "have _been_ updated"?

Also I think it might be worth adding a second sentence:

> The examples below have been updated to use the `React.createRef()` API introduced in React 16.3. If you are using an earlier release of React, we recommend using [#callback-refs](callback refs) instead.

content/docs/refs-and-the-dom.md
```diff
@@ -29,32 +31,69 @@ For example, instead of exposing `open()` and `close()` methods on a `Dialog` co
...
+### Accessing Refs
+
+When a ref is passed to an element in `render`, a reference to the node becomes accessible at the `value` attribute of the ref.
+
+```javascript
+const node = this.myRef.value
+```
 
-```javascript{8,9,19}
+The value of the ref differs depending on whether the node is an HTML DOM node, a React class component instance, or a stateless functional component:
```
bvaughn: I think we should maybe drop the "_or a stateless functional component_" bit in this sentence. It seems a little confusing, since you can't actually use a ref in this case. (I think we should leave the bullet point below that explicitly states this.)

content/docs/refs-and-the-dom.md
```diff
@@ -29,32 +31,69 @@ For example, instead of exposing `open()` and `close()` methods on a `Dialog` co
...
+
+- When the `ref` attribute is used on an HTML element, the `ref` created in the constructor with `React.createRef()` receives the underlying DOM element as its `value` property.
+- When the `ref` attribute is used on a custom component declared as a class, the `ref` object receives the mounted instance of the component as its `value`.
```
bvaughn: Wording nit. Maybe:

> When the `ref` attribute is used on a custom class component, the `ref` object receives the mounted instance of the component as its `value`.

content/docs/refs-and-the-dom.md
```diff
@@ -29,32 +31,69 @@ For example, instead of exposing `open()` and `close()` methods on a `Dialog` co
...
+  }
+  render() {
+    return <div ref={this.myRef} />;
+  }
+}
+```
+
+### Accessing Refs
+
+When a ref is passed to an element in `render`, a reference to the node becomes accessible at the `value` attribute of the ref.
```
bvaughn: Wording nit: We have a lot of sentences with both "ref" and "reference". It's fine, but I wonder if it might read better if we swapped "reference" with "pointer" in some of these places?

> When a ref is passed to an element in `render`, a pointer to the node becomes accessible at the `value` attribute of the ref.

content/docs/refs-and-the-dom.md
```diff
@@ -29,32 +31,69 @@ For example, instead of exposing `open()` and `close()` methods on a `Dialog` co
...
+
+- When the `ref` attribute is used on an HTML element, the `ref` created in the constructor with `React.createRef()` receives the underlying DOM element as its `value` property.
+- When the `ref` attribute is used on a custom component declared as a class, the `ref` object receives the mounted instance of the component as its `value`.
+- **You may not use the `ref` attribute on functional components** because they don't have instances.
+
+The examples below demonstrate the differences.
+
+#### Adding a Ref to a DOM Element
+
+This code uses a `ref` to store a reference to a DOM node:
```
bvaughn: Wording nit: Same was what I mentioned above. I wonder if this would read nicer:

> This code uses a `ref` to store a pointer to a DOM node:

content/docs/refs-and-the-dom.md
```diff
@@ -66,24 +105,26 @@ class CustomTextInput extends React.Component {
...
+React will assign the `value` property with the DOM element when the component mounts, and assign it back to `null` when it unmounts. `ref` updates happen before `componentDidMount` or `componentDidUpdate` lifecycle hooks.
 
-### Adding a Ref to a Class Component
+#### Adding a Ref to a Class Component
 
-When the `ref` attribute is used on a custom component declared as a class, the `ref` callback receives the mounted instance of the component as its argument. For example, if we wanted to wrap the `CustomTextInput` above to simulate it being clicked immediately after mounting:
+If we wanted to wrap the `CustomTextInput` above to simulate it being clicked immediately after mounting, we could use a ref to get access to the custom input and call its `focusTextInput` method manually:
 
-```javascript{3,9}
+```javascript{4,7,12}
```
bvaughn: The line-highlights are still off for this example. You can see it here:
https://deploy-preview-637--reactjs.netlify.com/docs/refs-and-the-dom.html#adding-a-ref-to-a-class-component

Should be {4,8,13} rather than {4,7,12}.

This (along with Prettier formatting) is why [I suggested moving these examples](https://github.com/reactjs/reactjs.org/pull/637/files#r170019692) to the examples folder. It's easier to manage highlights inline rather than at the top of the code snippet. 😄

content/docs/refs-and-the-dom.md
```diff
@@ -97,21 +138,24 @@ class CustomTextInput extends React.Component {
...
   return <input />;
 }
 
 class Parent extends React.Component {
+  constructor(props) {
+    super(props);
+    this.textInput = React.createRef();
+  }
   render() {
     // This will *not* work!
```
bvaughn: Unrelated to this diff: I wonder if we should highlight this (comment) line too?

content/docs/refs-and-the-dom.md
bvaughn: Fixed. 👍

content/docs/refs-and-the-dom.md
bvaughn: Fixed. 👍

content/docs/refs-and-the-dom.md
```diff
@@ -151,11 +195,11 @@ In rare cases, you might want to have access to a child's DOM node from a parent
 
 While you could [add a ref to the child component](#adding-a-ref-to-a-class-component), this is not an ideal solution, as you would only get a component instance rather than a DOM node. Additionally, this wouldn't work with functional components.
 
-Instead, in such cases we recommend exposing a special prop on the child. The child would take a function prop with an arbitrary name (e.g. `inputRef`) and attach it to the DOM node as a `ref` attribute. This lets the parent pass its ref callback to the child's DOM node through the component in the middle.
+Instead, in such cases we recommend exposing a special prop on the child. The child would take a prop with an arbitrary name (e.g. `inputRef`) and attach it to the DOM node as a `ref` attribute. This lets the parent pass its ref to the child's DOM node through the component in the middle.
```
bvaughn: Unrelated to this diff, but...

> The child would take a prop with an arbitrary name (e.g. `inputRef`)

I don't know about the use of the word "arbitrary" here. I wonder if this wouldn't read better:

> Instead, in such cases we recommend exposing a special prop on the child. This prop can be named anything other than `ref` (e.g. `inputRef`). The child component can then forward the prop to the DOM node as a `ref` attribute. This lets the parent pass its ref to the child's DOM node through the component in the middle.

content/docs/refs-and-the-dom.md
bvaughn: Fixed 👍

content/docs/refs-and-the-dom.md
```diff
@@ -165,25 +209,27 @@ function CustomTextInput(props) {
...
-        inputRef={el => this.inputElement = el}
-      />
+      <CustomTextInput inputRef={this.inputElement} />
     );
   }
 }
 ```
 
-In the example above, `Parent` passes its ref callback as an `inputRef` prop to the `CustomTextInput`, and the `CustomTextInput` passes the same function as a special `ref` attribute to the `<input>`. As a result, `this.inputElement` in `Parent` will be set to the DOM node corresponding to the `<input>` element in the `CustomTextInput`.
+In the example above, `Parent` passes its class property `this.inputElement` as an `inputRef` prop to the `CustomTextInput`, and the `CustomTextInput` passes the same ref as a special `ref` attribute to the `<input>`. As a result, `this.inputElement.value` in `Parent` will be set to the DOM node corresponding to the `<input>` element in the `CustomTextInput`.
```
bvaughn: nit: I think this example might read clearer if the parent class property was `inputRef` as well.

content/docs/refs-and-the-dom.md
bvaughn: Fixed 👍

content/docs/refs-and-the-dom.md
```diff
@@ -200,26 +246,90 @@ function Parent(props) {
...
-        inputRef={el => this.inputElement = el}
-      />
+      <Parent inputRef={this.inputElement} />
     );
   }
 }
 ```
 
-Here, the ref callback is first specified by `Grandparent`. It is passed to the `Parent` as a regular prop called `inputRef`, and the `Parent` passes it to the `CustomTextInput` as a prop too. Finally, the `CustomTextInput` reads the `inputRef` prop and attaches the passed function as a `ref` attribute to the `<input>`. As a result, `this.inputElement` in `Grandparent` will be set to the DOM node corresponding to the `<input>` element in the `CustomTextInput`.
+Here, the ref `this.inputElement` is first specified by `Grandparent`. It is passed to the `Parent` as a regular prop called `inputRef`, and the `Parent` passes it to the `CustomTextInput` as a prop too. Finally, the `CustomTextInput` reads the `inputRef` prop and attaches the passed ref as a `ref` attribute to the `<input>`. As a result, `this.inputElement.value` in `Grandparent` will be set to the DOM node corresponding to the `<input>` element in the `CustomTextInput`.
```
bvaughn: nit: Same thought here as above w.r.t. `this.inputElement` reading simpler as `this.inputRef`.

content/docs/refs-and-the-dom.md
```diff
@@ -200,26 +246,90 @@ function Parent(props) {
...
 
 All things considered, we advise against exposing DOM nodes whenever possible, but this can be a useful escape hatch. Note that this approach requires you to add some code to the child component. If you have absolutely no control over the child component implementation, your last option is to use [`findDOMNode()`](/docs/react-dom.html#finddomnode), but it is discouraged.
```
bvaughn: > All things considered, we advise against exposing DOM nodes whenever possible, but this can be a useful escape hatch.

Unrelated to this diff wording nit:

> When possible, we advise against exposing DOM nodes, but it can be a useful escape hatch.

content/docs/refs-and-the-dom.md
```diff
@@ -200,26 +246,90 @@ function Parent(props) {
...
 
+### Callback Refs
+
+React also supports another way to set refs called "callback refs", which give more fine-grain control over when refs are set and unset.
```
bvaughn: Grammar nit: "which give" should be "which gives" (b'c of "another way")

content/docs/refs-and-the-dom.md
bvaughn: I still think this, by the way. I don't think it's really a performance concern, but I think there's a tiny chance of confusion here. React will unset the old ref and re-set the new ref each render cycle, which I have known to confuse people in the past.

We even mention this as a caveat below. 😄

content/docs/refs-and-the-dom.md
```diff
@@ -200,26 +246,90 @@ function Parent(props) {
...
+    );
+  }
+}
+```
+
+In the example above, `Parent` passes its ref callback as an `inputRef` prop to the `CustomTextInput`, and the `CustomTextInput` passes the same function as a special `ref` attribute to the `<input>`. As a result, `this.inputElement` in `Parent` will be set to the DOM node corresponding to the `<input>` element in the `CustomTextInput`.
+
 ### Legacy API: String Refs
 
 If you worked with React before, you might be familiar with an older API where the `ref` attribute is a string, like `"textInput"`, and the DOM node is accessed as `this.refs.textInput`. We advise against it because string refs have [some issues](https://github.com/facebook/react/pull/8333#issuecomment-271648615), are considered legacy, and **are likely to be removed in one of the future releases**. If you're currently using `this.refs.textInput` to access refs, we recommend the callback pattern instead.
```
bvaughn: Unrelated to this diff: I wonder if we shouldn't make that last sentence stand out more by making it a bolded **note**:

> **Note**
>
> If you're currently using `this.refs.textInput` to access refs, we recommend the callback pattern instead.

content/docs/refs-and-the-dom.md
alexkrolick: > The value of the ref differs depending on the type of the node: ... bullets...

content/docs/refs-and-the-dom.md
alexkrolick: Refs provide a way to _reference_ (get a reference to?) DOM nodes and React elements created by React.

content/docs/refs-and-the-dom.md
alexkrolick: Maybe... but pointer isn't particularly common terminology in JS-land so I'd worry it might confuse some people.

content/docs/refs-and-the-dom.md
alexkrolick: Changed in https://github.com/trueadm/reactjs.org/pull/3/files

content/docs/refs-and-the-dom.md
alexkrolick: Changed in https://github.com/trueadm/reactjs.org/pull/3/files

content/docs/refs-and-the-dom.md
alexkrolick: https://github.com/trueadm/reactjs.org/pull/4

content/docs/refs-and-the-dom.md
alexkrolick: It might be obvious, but at some point in the page it should probably say that "ref" is short for "reference"... 💭

content/docs/refs-and-the-dom.md
bvaughn: Terms like "pointer" and "reference" aren't just used in programming. I think they have broader meaning outside of programming that's relevant.

Maybe someone like @gaearon could weigh-in with an opinion about non-native English speakers and if there are concerns there though. I have no idea.

content/docs/refs-and-the-dom.md
bvaughn: The line highlights are still off for this example...

screen shot 2018-03-07 at 10 35 05 am

bvaughn:
I merged this PR into the 16.3 docs PR since they need to go out at the same time anyway (see bvaughn/reactjs.org/pull/1).

PR #638, opened 2018-02-22
Edit compiled JS

mrassili:
This code : ```<div className="sidebar" />``` compiles into JS, giving only 2 arguments like that ```React.createElement('div', {className: 'sidebar'})``` **There's no 'null' as third argument.**


threepointone:
This was fixed elsewhere, but thank you very much for your PR!

PR #640, opened 2018-02-23
Added rel=canonical <link/> tag to the <head/>

### LINKED ISSUE #605: Test Utils page is available in Google search results, but the front-end redirects the visitor
## Note from maintainers: If you'd like to work on this issue, please comment to claim it so we don't have duplication of work – thanks!

It seems like there's a problem with routing on https://reactjs.org/docs/test-utils

I'm getting redirected to Page Not Found by the router on the front-end.

`curl https://reactjs.org/docs/test-utils` returns the expected document.

It's accessible via https://reactjs.org/docs/test-utils.html as well.

nilfalse:
This should prevent wrong pages to appear in SERP.
Fixes #605

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ This repo contains the source code and documentation powering [reactjs.org](http
 
 ## Translation
 
-If you are interesting in translating `reactjs.org`, please join the Crowdin.
+If you are interested in translating `reactjs.org`, please join the Crowdin.
 
 * [Crowdin - React](https://crowdin.com/project/react)
 
--- a/src/components/TitleAndMetaTags/TitleAndMetaTags.js
+++ b/src/components/TitleAndMetaTags/TitleAndMetaTags.js
@@ -19,6 +19,7 @@ type Props = {
 const TitleAndMetaTags = ({title, ogDescription, ogUrl}: Props) => {
   return (
     <Helmet title={title}>
+      {ogUrl && <link rel="canonical" href={ogUrl} />}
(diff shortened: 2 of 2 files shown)


sophiebits:
The URL https://deploy-preview-640--reactjs.netlify.com/docs/typechecking-with-proptypes in the deploy preview still shows "Page Not Found". We would want to fix that too if we are going to go with the canonical solution.

nilfalse:
hi @sophiebits,
that was on purpose: the point is to just provide a hint to search engines while maintaining the site structure you have had so far.

**tl;dr.** IMO, a user should not face any changes on the website after we fix the issue with search engines and this is exactly what this PR does in its current state.

If you take a closer look at the page at the URL you referred to, you may notice that with this PR in place there is a `<link rel="canonical" href=... />` tag being added to the source code of the page, which is completely invisible for the end user, but plays very valuable role for most search engines.

Furthermore, I think either of the following solutions would potentially harm SEO ranks of the website:
a] introducing valid mirroring pages without `.html` postfix in their URL while also keeping those with `.html`;
b] redirecting (with code 301) users from `.html` pages to their non-`.html` counterparts.

Because:
a] would introduce repeating content (which search engines may penalize)
and b] would lower the SEO score for a page doing 301 in favour of the target of the redirect, but the target may still not be in the search engine index by the time and so not in the search results at all.

nilfalse:
@sophiebits any ideas regarding this?

nilfalse:
@sophiebits any news here?

PR #645, opened 2018-02-26
Added new sections to Optimizing Performance article

### LINKED ISSUE #639: Possible issue in "Components and Props" section of docs - functional vs. class components
The current documentation says:

> The simplest way to define a component is to write a JavaScript function:
> [...]
> You can also use an ES6 class to define a component:
> [...]
> The above two components are equivalent from React’s point of view.

I just encountered a contradictory comment from @trueadm, an engineer on the React core team:
> In React 16, functional components are quite a bit faster than class components. On a side note, this optimization is something we’re aware of and have been talking about a bit recently. :)

https://medium.com/@trueadm/in-react-16-functional-components-are-quite-a-bit-faster-than-class-components-1afca4931d7c

Does this mean that the possible future optimizations for functional components that were mentioned in previous versions of the docs and blog posts have now been implemented? If so, this should be mentioned in the docs (and probably should have been in the release notes for React 16 too, but I don't see it mentioned there).

If this has indeed been implemented, then it should also be noted that this is a micro-optimization that still pales in comparison with using PureComponent (or your own efficient `shouldComponentUpdate()` logic), as explained in https://github.com/facebook/react/issues/5677#issuecomment-165125151:

> For complex components, defining `shouldComponentUpdate` (eg. pure render) will generally exceed the performance benefits of stateless components. The sentences in the docs are hinting at some future optimizations that we have planned, whereby we won't allocate an internal instance for stateless functional components (we will just call the function). We also might not keep holding the props, etc. Tiny optimizations. We don't talk about the details in the docs because the optimizations aren't actually implemented yet (stateless components open the doors to these optimizations).

mbrowne:
Added new sections to Optimizing Performance article: "Functional vs. Class Components" and "When to Use the Different Component Types".

Fixes #639

### DIFF
--- a/content/docs/components-and-props.md
+++ b/content/docs/components-and-props.md
@@ -42,9 +42,11 @@ class Welcome extends React.Component {
 }
 ```
 
-The above two components are equivalent from React's point of view.
+The above two components are equivalent from React's point of view (aside from a small performance difference). *
 
 Classes have some additional features that we will discuss in the [next sections](/docs/state-and-lifecycle.html). Until then, we will use functional components for their conciseness.
+
+&ast; *By default, functional components are a little faster than class components; see [Optimizing Performance](/docs/optimizing-performance.html).*
 
 ## Rendering a Component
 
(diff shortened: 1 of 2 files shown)


rickhanlonii:
Sorry for the long delay here! Since this PR was created against the old site I'm going to close this, but I also don't think it's super relevant any more (see my comment on the issue).

PR #650, opened 2018-02-28

ravshansbox:

#<Issue>


alexkrolick:
It's React that cares about the component reference, ultimately.

`React.createElement(Welcome)` vs `React.createElement('div')`

The transpiler has a say in whether the name becomes a string or variable... but that's probably a bit confusing without linking to the React Without JSX docs.

ravshansbox:
Hmm, yeap, completely agree.

PR #651, opened 2018-02-28
React.cloneElement => props are objects not arrays

iAmCalvinL:
Changing brackets to squiggly brackets because props being passed into cloned elements should be objects and not arrays. This can be confusing.

### DIFF
--- a/content/docs/reference-react.md
+++ b/content/docs/reference-react.md
@@ -104,7 +104,7 @@ Code written with [JSX](/docs/introducing-jsx.html) will be converted to use `Re
 ```
 React.cloneElement(
   element,
-  [props],
+  {props},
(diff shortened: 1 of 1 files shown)


alexkrolick:
In this case, square brackets mean the argument is optional. Putting {} around props would create an object with a key called "props", which we don't want.

Marcholio:
Could this be closed then if the change is rejected?

BartoszKlonowski:
Just like @alexkrolick said: passing `props` to `createElement` should be done as an object, but putting it inside `{}` brackets creates a new object with a given key.
Let me close this PR, moreover as `React.cloneElement` is already marked as legacy API and the page edited in this PR was moved.

PR #655, opened 2018-03-01
Small code optimization in the document.

shafiulbashar:
Small optimization proposed in the document to reduce cyclic complexity in if-else condition.

#<Issue>


alexkrolick:
It's not necessarily an optimization if you have to fire React.createElement twice in the logged-in scenario. That will most likely be more expensive than an `if` check.

sustbird:
Thank you for your explanation.

PR #663, opened 2018-03-04
Improve a11y in homepage examples with some labels

jdan:
Hiya!

After some [chatter on twitter](https://twitter.com/pauljadam/status/969969247714766848) about the a11y of the react homepage examples, I took the liberty of adding some labels. 

This changeset adds labels to the Todo and markdown code examples on the homepage - the aim being that if folks have more exposure to something like a semantic <label> tag while learning react, maybe the habit will stick.

a screenshot from the reactjs homepage code example for rendering a todo list - showing source code on the left with label tags added, and rendered output on the right

a screenshot from the reactjs homepage code example for rendering markdown - showing source code on the left with label tags added, and rendered output on the right

I took a bit of creative license on the markdown textarea's label - and changed the default state of the component to render `Hello, **world**!` instead of `Type some *markdown* here` since the label conveys that. Happy to bikeshed this.

Thanks!

### DIFF
--- a/content/home/examples/a-component-using-external-plugins.js
+++ b/content/home/examples/a-component-using-external-plugins.js
@@ -2,7 +2,7 @@ class MarkdownEditor extends React.Component {
   constructor(props) {
     super(props);
     this.handleChange = this.handleChange.bind(this);
-    this.state = { value: 'Type some *markdown* here!' };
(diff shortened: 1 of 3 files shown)


alexkrolick:
👍 Thank you!

PR #665, opened 2018-03-06
Add docs for new Context API

alexkrolick:
[Rendered](https://deploy-preview-665--reactjs.netlify.com/docs/context.html)


content/docs/context.md
```diff
@@ -4,35 +4,71 @@ title: Context
...
-## Why Not To Use Context
+`embed:context/theme-detailed-themed-button.js`
 
-The vast majority of applications do not need to use context.
+`embed:context/theme-detailed-app.js`
 
-If you want your application to be stable, don't use context. It is an experimental API and it is likely to break in future releases of React.
 
-If you aren't familiar with state management libraries like [Redux](https://github.com/reactjs/redux) or [MobX](https://github.com/mobxjs/mobx), don't use context. For many practical applications, these libraries and their React bindings are a good choice for managing state that is relevant to many components. It is far more likely that Redux is the right solution to your problem than that context is the right solution.
+## Experimental API (Deprecated in React 16.3)
```
bvaughn: Maybe this could be broken out into its own "Legacy context" page that is linked to from this page- but isn't on the side bar anywhere?

content/docs/context.md
```diff
@@ -4,35 +4,71 @@ title: Context
...
-> Note:
->
-> `React.PropTypes` has moved into a different package since React v15.5. Please use [the `prop-types` library instead](https://www.npmjs.com/package/prop-types) to define `contextTypes`.
->
->We provide [a codemod script](/blog/2017/04/07/react-v15.5.0.html#migrating-from-react.proptypes) to automate the conversion.
+Context provides a way to pass data through the component tree without having to pass props down manually at every level.
+
+## Using Context
+
+Here is an example illustrating how you might inject a "theme" using context:
```
bvaughn: I prefer the introductory sequence from the [old "How to Use Context"](https://reactjs.org/docs/context.html#how-to-use-context) section more because of how it setup the problem.

I would suggest following that pattern, but putting it down under the "Typical Usage" header. I think the API should be first on this page.

alexkrolick: I think it's confusing how the old leads with a large codeblock illustrating non-context APIs. Maybe we can invert the order to highlight the context way before the prop-drilling way?

Moving the API section to the top makes sense.

bvaughn: I think we could maybe make the initial code block _shorter_, but I think it's useful to present APIs in the form of "Here is a scenario. How would you solve it? You can solve it using this API."

src/theme.js
```diff
@@ -336,7 +336,7 @@ const sharedStyles = {
       },
 
       '& li': {
-        marginTop: 20,
+        marginTop: 10,
```
alexkrolick: Reduces spacing between list elements:

After:
image

Before:
image

alexkrolick:
@bvaughn broke out `legacy-context.md` and reintroduced the props drilling example under "Motivation"

content/docs/context.md
```diff
@@ -4,39 +4,83 @@ title: Context
...
+  - [Dynamic Context](#dynamic-context)
+- [Motivation](#motivation)
+- [Legacy API](#legacy-api)
+
+## API
+
+### `React.createContext`
+
+```js
+const {Provider, Consumer} = React.createContext([default]);
```
bvaughn: Why `[default]` ? (Why the array syntax?) Suggestion:

```js
const {Provider, Consumer} = React.createContext(defaultValue);
```

content/docs/context.md
```diff
@@ -4,39 +4,83 @@ title: Context
...
+```
+
+Creates a `{ Provider, Consumer }` pair.
+
+Takes one argument, the default context that Consumers will receive when they don't have a matching Provider.
```
bvaughn: Suggestion:

> Accepts an optional `default` value to be passed to `Consumers` without a `Provider` ancestor.

content/docs/context.md
```diff
@@ -4,39 +4,83 @@ title: Context
...
+### `Provider`
 
-With React, it's easy to track the flow of data through your React components. When you look at a component, you can see which props are being passed, which makes your apps easy to reason about.
+```js
+<Provider value={/* some value */}>
+```
+
+A React component that allows Consumers to subscribe to context changes.
+
+Takes one prop, `value`, which will be passed to the [render prop](/docs/render-props.html) of child Consumers for the matching context anywhere in the component tree. One Provider can be connected to many Consumers.
```
bvaughn: Suggestion:

> Accepts a `value` prop to be passed to Consumers that are descendants of this Provider. One Provider can be connected to many Consumers. Providers can be nested to override values deeper within the tree.

content/docs/context.md
```diff
@@ -4,39 +4,83 @@ title: Context
...
+<Consumer>
+ { value => { /* render something based on the context value */ } }
+</Consumer>
+```
 
-In some cases, you want to pass data through the component tree without having to pass the props down manually at every level.
-You can do this directly in React with the powerful "context" API.
+A React component that subscribes to context changes. 
+
+Takes a function as the `children` prop that receives the `value` prop of the matching Provider. This function will be called whenever the Provider's value is updated.
```
bvaughn: Suggestion:

> Requires a function as a child. This function receives the current context `value` and returns a React node. This function will be called whenever the Provider's value is updated.

examples/context/theme-example.js
```diff
@@ -0,0 +1,42 @@
+const defaultTheme = 'light';
+// highlight-next-line
+const ThemeContext = React.createContext(defaultTheme);
```
bvaughn: Nit: Maybe get rid of `defaultTheme` variable in favor of a comment?

```js
// Initial context with a default value of "light" (theme)
const ThemeContext = React.createContext('light');
```

examples/context/theme-example.js
```diff
@@ -0,0 +1,42 @@
...
+
+class ThemeProvider extends React.Component {
+  state = {theme: 'light'};
```
bvaughn: Maybe show a different value than the default?
Maybe don't show a class component with `state` at all? Because it isn't obvious why it's necessary in this case.

examples/context/theme-example.js
```diff
@@ -0,0 +1,42 @@
...
+        {theme => <Button theme={theme} />}
+      </ThemeContext.Consumer>
+    );
+  }
+}
+
+class App extends React.Component {
+  // highlight-range{3}
+  // highlight-range{5}
+  // highlight-range{7}
```
bvaughn:
```js
// highlight-range{3,5,7}
```

examples/context/theme-example.js
```diff
@@ -0,0 +1,42 @@
...
+  render() {
+    return (
+      <ThemeProvider>
+        <SomeComponent>
+          <ThemedButton />
```
bvaughn: I wonder if we should omit `ThemeProvider` entirely from this example? Or at least show that `ThemedButton` can live outside of the provider (using the default value)?

examples/context/theme-detailed-app.js
```diff
@@ -0,0 +1,37 @@
...
+  };
+
+  changeHighlightColor = () => {
+    const colors = ['red', 'blue', 'green'];
+    const randomColor =
+      colors[Math.floor(Math.random() * 3)];
+    this.setState({
+      theme: {
+        ...this.state.theme,
+        highlight: randomColor,
```
bvaughn: I'm not a big fan of this randomization.

Maybe we could simplify this a bit with a `toggle` function that flips between "light" and "dark" theme or something?

examples/context/theme-detailed-app.js
```diff
@@ -0,0 +1,37 @@
+import ThemeContext from './theme-context';
+import ThemedButton from './button';
+
+class App extends React.Component {
+  state = {
+    theme: {
+      highlight: 'blue',
+      accent: 'purple',
```
bvaughn: nit: It's also a little weird that we have a default value and `state` with the same value. People might not understand why default value exists if we always do this.

content/docs/context.md
```diff
@@ -51,6 +95,11 @@ class Message extends React.Component {
   render() {
     return (
       <div>
+        {/*
+          The Message component must take `color` as as prop to pass it to the
+          Button. Using context, the Button could connect to the color context
+          on its own.
+        */}
```
bvaughn: This block comment within JSX does not read very well. Move it into the `render` function as just inline comments?

content/docs/context.md
```diff
@@ -4,39 +4,83 @@ title: Context
...
 
-If you insist on using context despite these warnings, try to isolate your use of context to a small area and avoid using the context API directly when possible so that it's easier to upgrade when the API changes.
+**app.js**
+`embed:context/theme-detailed-app.js`
 
-## How To Use Context
+## Motivation
 
-Suppose you have a structure like:
+Context is designed to relieve the pain of passing props down through a deeply nested component tree. For example, in the code below we manually thread through a color prop in order to style the Button and Message components. Using context, we can avoid passing props through intermediate elements.
```
bvaughn: I'm still not a fan of this sequencing. We show motivation _after_ we show usage examples. That feels backwards to me. I think it would be better to present a problem, and then how to solve it using Context. Step by step.

content/docs/legacy-context.md
```diff
@@ -0,0 +1,215 @@
+---
+id: legacy-context
+title: Legacy Context
+permalink: docs/legacy-context.html
+---
+
+> This API is deprecated as of React 16.3.
```
bvaughn: This should clarify that the feature will be removed in 17.

Suggestion>

> The legacy API has been deprecated and will be removed in version 17. Instead, use [the new context](/docs/context.html) API introduced with version 16.3.
>
> The legacy API will continue working for all 16.x releases.

content/docs/legacy-context.md
```diff
@@ -0,0 +1,215 @@
...
+> 
+> The API will be supported in all 16.x releases, but applications using it should migrate to the [new API](/docs/context.html). The experimental API lacked a safe mechanism to update context. Version 16.3 introduced a new context API that is more efficient and supports both static type checking and deep updates.
+
+## How To Use Context
+
+> This section documents a deprecated API. See the [new API](/docs/context.html).
```
bvaughn: I think these notes are pretty redundant with the one at the head of the page. Probably can remove them?

examples/context/theme-example.js
alexkrolick: Ah, I was looking for docs on that

content/docs/legacy-context.md
alexkrolick: Probably so. My concern is that it's a long page and someone might get a direct link to a section without noticing the deprecation warning.

content/docs/context.md
alexkrolick: I was thinking [] indicated optional arguments. Replacing it with a text note is fine; it's not a real type signature or anything.

content/docs/context.md
alexkrolick: Good idea, ancestor/descendant gets around the awkward "matching context" phrasing.

examples/context/theme-example.js
alexkrolick: Actually copied this from the blog post: https://github.com/reactjs/reactjs.org/pull/587/files#diff-551ec185a5d5f7818d2b02fbbf901c28

I'll simplify this one since the dynamic example covers stateful providers.

examples/context/theme-detailed-app.js
alexkrolick: I was hoping to show something with a `value` more complex than a string... maybe an Avatar component with a default pic?

examples/context/theme-detailed-app.js
alexkrolick: What's a good application use case for the default value? The RFC example just says ["to ensure type correctness"](https://github.com/reactjs/rfcs/blob/master/text/0002-new-version-of-context.md#detailed-design).

bvaughn: That's a good observation. I wasn't considering the type-correctness angle. 👍

I was thinking it would be nice for the docs to highlight, in one of the examples, that the default value gets passed to a Consumer if there is no Provider ancestor.

content/docs/context.md
bvaughn: Gotcha. I don't think thats' what "[]" implies to me. 👍

examples/context/theme-example.js
bvaughn: 😆 That's fair!

examples/context/theme-detailed-app.js
bvaughn: Ah, I see. An object value or something seems fine. I just think the randomness maybe detracts from the example some.

content/docs/legacy-context.md
bvaughn: It's possible. Definitely seems redundant for the first section though, since the two notes are only like 50 pixels apart. I don't feel as strongly about the following sections.

bvaughn:
Looks like CI is saying Prettier needs to be run.

content/docs/context.md
```diff
@@ -4,220 +4,88 @@ title: Context
...
-You can do this directly in React with the powerful "context" API.
-
-> Note:
->
-> A [new, safe version of context](https://github.com/reactjs/rfcs/blob/master/text/0002-new-version-of-context.md) is under development for the upcoming 16.3 release.
-
+Context provides a way to pass data through the component tree without having to pass props down manually at every level.
 
-## Why Not To Use Context
+Typically, data in a React application is passed top-down (parent to child) via props. But sometimes it's useful to pass values through multiple levels without adding props to every intermediate component. Examples include a language preference, or a UI theme. Many components may rely on those but you don't want to have to pass a locale prop and a theme prop through every level of the tree.
```
bvaughn: This paragraph feels like it ends a bit prematurely. Maybe we can close by reiterating the opening line?

Wording suggestions:

> In a typical React application, data is passed top-down (parent to child) via props, but this can be cumbersome for certain types of props (e.g. locale preference, UI theme) that are required by many components within an application. Context provides a way to share values like this between components without having to explicitly pass a prop through every level of the tree.

content/docs/context.md
```diff
@@ -4,220 +4,88 @@ title: Context
...
-The vast majority of applications do not need to use context.
+- [Motivation](#motivation)
+- [API](#api)
+  - [React.createContext](#reactcreatecontext)
+  - [Provider](#provider)
+  - [Consumer](#consumer)
+- [Examples](#examples)
+  - [Static Context](#static-context)
+  - [Dynamic Context](#dynamic-context)
+- [Legacy API](#legacy-api)
```
bvaughn: I like this table of contents 👍

content/docs/context.md
```diff
@@ -4,220 +4,88 @@ title: Context
...
 
-If you want your application to be stable, don't use context. It is an experimental API and it is likely to break in future releases of React.
 
-If you aren't familiar with state management libraries like [Redux](https://github.com/reactjs/redux) or [MobX](https://github.com/mobxjs/mobx), don't use context. For many practical applications, these libraries and their React bindings are a good choice for managing state that is relevant to many components. It is far more likely that Redux is the right solution to your problem than that context is the right solution.
+## Motivation
 
-If you're still learning React, don't use context. There is usually a better way to implement functionality just using props and state.
+Context is designed to relieve the pain of passing props down through a deeply nested component tree. For example, in the code below we manually thread through a color prop in order to style the Button and Message components. Using context, we can avoid passing props through intermediate elements.
```
bvaughn: nit: I would move the "_Using context, we can avoid passing props through intermediate elements._" sentence to below the coding example, so it's clear that the example isn't showing what context _can_ do.

content/docs/context.md
```diff
@@ -4,220 +4,88 @@ title: Context
...
 
-If you insist on using context despite these warnings, try to isolate your use of context to a small area and avoid using the context API directly when possible so that it's easier to upgrade when the API changes.
+`embed:context/motivation.js`
```
bvaughn: It still feels wrong to me to have a "motivation" section that does not show how to "fix" the problem we introduce.

content/docs/context.md
```diff
@@ -4,220 +4,88 @@ title: Context
...
-
-## Referencing Context in Lifecycle Methods
-
-If `contextTypes` is defined within a component, the following [lifecycle methods](/docs/react-component.html#the-component-lifecycle) will receive an additional parameter, the `context` object:
-
-- [`constructor(props, context)`](/docs/react-component.html#constructor)
-- [`componentWillReceiveProps(nextProps, nextContext)`](/docs/react-component.html#componentwillreceiveprops)
-- [`shouldComponentUpdate(nextProps, nextState, nextContext)`](/docs/react-component.html#shouldcomponentupdate)
-- [`componentWillUpdate(nextProps, nextState, nextContext)`](/docs/react-component.html#componentwillupdate)
+Requires a [function as a child](/docs/render-props.html#using-props-other-than-render). This function receives the current context value and returns a React node, and will be called whenever the Provider's value is updated.
```
bvaughn: I think the last sentence is a bit too long. Suggestion:

> Requires a [function as a child](/docs/render-props.html#using-props-other-than-render). This function receives the current context value and returns a React node. It will be called whenever the Provider's value is updated.

examples/context/theme-example.js
```diff
@@ -0,0 +1,55 @@
...
+      <ThemedButton />
+    </div>
+  );
+};
+
+class App extends React.Component {
+  render() {
+    // The ThemedButton button inside the ThemeProvider
+    // uses the dark theme while the one outside uses the
+    // default light theme
```
bvaughn: I like the changes to this example 👍

examples/context/theme-detailed-app.js
```diff
@@ -0,0 +1,30 @@
+import ThemeContext, {themes} from './theme-context';
```
bvaughn: I don't think we should show a mixed import like this.

Maybe we can import `themes` from a separate `themes.js` file? Or change `theme-context` to remove the default export.

examples/context/theme-detailed-app.js
```diff
@@ -0,0 +1,30 @@
...
+    //highlight-range{2,6}
+    return (
+      <ThemeContext.Provider value={this.state.theme}>
+        <ThemedButton onClick={this.toggleTheme}>
+          Change Theme
+        </ThemedButton>
+      </ThemeContext.Provider>
+    );
+  }
+}
```
bvaughn: I like the changes to the theme example as well.

src/theme.js
bvaughn: I think it looks nice.

alexkrolick:
Thanks for the review @bvaughn. Added the solution to the motivation problem. I think the highlights emphasize the benefits pretty well.

bvaughn:
You're welcome, @alexkrolick. Thanks for owning this change! It's a big help for the upcoming release.

bvaughn:
I'd like to merge this branch into the `bvaughn:16.3-release-blog-post` branch (PR #587) so these docs can be released with the other 16.3 stuff. That cool with you?

alexkrolick:
Yep: https://github.com/bvaughn/reactjs.org/pull/3

bvaughn:
Thanks!

Issue #505, opened 2018-01-10
[Documentation] - Minor change to description of React.Children.map


AnderssonChristian:
[This section](https://reactjs.org/docs/react-api.html#reactchildrenmap) of the documentation states:

> If children is null or undefined, returns null or undefined rather than an array.

This seems to imply that if you do `this.props.children.map()` you will get an (empty) array back in case children is null or undefined. This isn't the case because you'll get a TypeError.

Had `this.props.children.map()` returned an empty array it would effectively have the same effect in `render()` as returning null or undefined. So perhaps clarify this a bit:

> If children is null or undefined, returns null or undefined rather than throwing an error.

EarthSchlange:
If I am understanding the issue correctly, the function `React.Children.map`  will return as null or undefined vs an error not array from using `this.props.children.map()` directly seen here: [Looping Over Children - mxstbr.blog](https://mxstbr.blog/2017/02/react-children-deepdive/#looping-over-children). This change would better describe what is happening because using `this.props.children.map()` does not return an array but a TypeError.  

![react children map error screenshot mxstbr blog 9 mar 2018](https://user-images.githubusercontent.com/5054653/37238270-8056e41e-23d7-11e8-9705-ce450fc0dd6a.png)

eps1lon:
> This seems to imply that if you do this.props.children.map() you will get an (empty) array back in case children is null or undefined. This isn't the case because you'll get a TypeError.

The section is not talking about `this.props.children` but `React.Children`. `this.props.children.map` will throw a TypeError because `this.props.children` and `React.Children` are  completely different things. Even if `this.props.children` is an array then `this.props.children.map` and `React.Children.map` have different semantics.

PR #671, opened 2018-03-10
Minor Change with React.Children.map Docs to Reflect TypeError

EarthSchlange:
Minor documentation change to reflect the TypeError from `this.props.children.map()`
Also removes extra space after comparision. 
Fixes Issue #505


eps1lon:
Thanks for your work but the underlying issue had some false assumptions. See https://github.com/reactjs/reactjs.org/issues/505#issuecomment-678807542 for more context.

PR #672, opened 2018-03-11
improve code style in state with default value

uxitten:


alexkrolick:
Thanks for the PR. Class property syntax is not standardized yet, so the docs try to avoid using them: https://github.com/reactjs/reactjs.org/blob/master/CONTRIBUTING.md#dont-use-features-that-arent-standardized-yet

PR #682, opened 2018-03-13
Remove indirection

joeyhub:


billyjanitsch:
This is incorrect: `() => this.tick()` is necessary to ensure the `tick` method is called with the correct context so that `this` refers to the component instance when it calls `this.setState`.

yangshun:
@billyjanitsch is right. @joeyhub you can test the change out on the reactjs.org homepage itself and see that the timer does not increase anymore.

*I left this page for 5 seconds after making the change before taking the screenshot*

screen shot 2018-03-24 at 9 12 40 am

PR #684, opened 2018-03-14
Explicitly call out a potential caveat with Render Prop components

mixu:
The documentation recommends using a bound function as the input to a render prop component, but this may be problematic if the render prop component itself is pure.

I think it is worth noting explicitly that render prop components that are pure may be problematic in a fairly common case: if the wrapping component is stateful and the render prop component (or a HOC it makes use of) is pure.

I ran into this and while it may be obvious to someone familiar with all the internals, it wasn't something I considered initially when converting an existing HOC (which was pure via another HOC) to the render props style.

### DIFF
--- a/content/docs/render-props.md
+++ b/content/docs/render-props.md
@@ -329,3 +329,59 @@ class MouseTracker extends React.Component {
 ```
 
 In cases where you cannot bind the instance method ahead of time in the constructor (e.g. because you need to close over the component's props and/or state) `<Mouse>` should extend `React.Component` instead.
+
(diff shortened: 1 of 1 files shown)


alexkrolick:
This is a good caveat to call out, but I think the full example could probably be replaced by a smaller note. For example:

```markdown
> Note:
>
> Components accepting render props should not extend React.PureComponent
> if the render function is not a pure function of props.
```

What do you think?

mixu:
sgtm!

PR #657, opened 2018-03-02
make code sample a little more concise

kaycebasques:
Just a little, opinionated suggestion. You can close this if it doesn't match your general code sample style.

PR #686, opened 2018-03-14
Adds Ternary Operator to Make Example Concise Issue #657

EarthSchlange:
Created fix for #657 to make the code more concise based using @alexkrolick suggestion. I am still getting used to Github, so I hope providing this pull request does not overstep @kaycebasques oppotunity to resolve this issue.

### DIFF
--- a/content/docs/conditional-rendering.md
+++ b/content/docs/conditional-rendering.md
@@ -92,10 +92,12 @@ class LoginControl extends React.Component {
 
   render() {
     const isLoggedIn = this.state.isLoggedIn;
-
-    let button = isLoggedIn
-                 ? <LogoutButton onClick={this.handleLogoutClick} />
-                 : <LoginButton onClick={this.handleLoginClick} />;
+    
+    let button = isLoggedIn ? (
+      <LogoutButton onClick={this.handleLogoutClick} />
+    ) : (
+      <LoginButton onClick={this.handleLoginClick} />
+    );
 
     return (
       <div>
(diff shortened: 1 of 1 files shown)


alexkrolick:
I think it's a little ugly too but it's the way Prettier formats it.

https://github.com/reactjs/reactjs.org/blob/master/CONTRIBUTING.md#style

alexkrolick:
Darn, should have made that a `const` too

gaearon:
I reverted this. The use of `if` was intentional.

This is a guide for beginners. Beginners are confused by ternaries. Especially if they're introduced before proper `if` statements.

If you look at the context, right after the example there was a sentence:

>While declaring a variable and using an `if` statement is a fine way to conditionally render a component, sometimes you might want to use a shorter syntax. There are a few ways to inline conditions in JSX, explained below.

It lost its meaning after this change. Please always consider examples together with surrounding code.

There was also a CodePen for this example. The update wasn't synced to it so it was confusing to see them diverge.

Finally, we explain ternaries later in this document: https://reactjs.org/docs/conditional-rendering.html#inline-if-else-with-conditional-operator. So it's confusing we used one before we explained it, and also that we repeat it two times.

alexkrolick:
:( my bad, should have looked beyond the diff.

gaearon:
No problem, it can be hard to track all those things! Just explaining the reasoning so you have more context in the future :-) Don't hold off merging stuff in the future because of this.

PR #688, opened 2018-03-14
fix: NavigationFooter links pull paginated title from nav metadata

### LINKED ISSUE #451: "next" footer links don't pull title from source

raunofreiberg:
Fixes #451.

**Successfully ran:**
- yarn check-all

### DIFF
--- a/src/components/MarkdownPage/MarkdownPage.js
+++ b/src/components/MarkdownPage/MarkdownPage.js
@@ -31,6 +31,21 @@ type Props = {
   titlePostfix: string,
 };
 
+const getPaginatedTitle = (
+  sectionList: Array<Object>,
+  templateFile: ?string,
+) => {
+  if (!templateFile) {
+    return null;
+  }
+
+  const sectionItems = sectionList.map(section => section.items);
+  const flattenedSectionItems = [].concat.apply([], sectionItems);
+  const linkId = templateFile.replace('.html', '');
+
+  return flattenedSectionItems.find(item => item.id === linkId);
+};
+
 const MarkdownPage = ({
   authors = [],
   createLink,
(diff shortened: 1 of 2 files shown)


src/components/MarkdownPage/MarkdownPage.js
```diff
@@ -31,6 +31,21 @@ type Props = {
   titlePostfix: string,
 };
 
+const getPaginatedTitle = (
```
alexkrolick: Looks great! My only nit is that this method name returns an  object, not just the title string, so a name like `getPageById` might be better than `getPaginatedTitle`.

raunofreiberg: I like this method name much better. Renamed in last commit. Thanks!

raunofreiberg:
Found a small bug with the template file, more specifically I forgot to append the `.html` extension back to the remapped URL of the page. Fixed in last commit.

raunofreiberg:
@alexkrolick anything else you'd like me to do with this PR? :-)

gaearon:
Yay, thank you for fixing! This was super annoying.

raunofreiberg:
My pleasure :-)

PR #690, opened 2018-03-15
add note to test utils renderIntoDocument

kentcdodds:
Adding clarification because (as noted in [the source](https://github.com/facebook/react/blob/8a09a2fc538e98523f67022095de9120e5dc2c32/packages/react-dom/src/test-utils/ReactTestUtils.js#L90-L98)) it's a slight misnomer and can be confusing.


kentcdodds:
Yikes, I just realized that `ReactDOM.render` returns `null` if you're rendering a function component 😱 That makes this method pretty unreliable for my use cases :-(

threepointone:
I'm going to close this, but we're likely going to revisit this in some fashion anyway. Besides, I suppose your own lib does a smashing job of this now :P

PR #694, opened 2018-03-20
Error in the article

es-repo:
The article contains error when says: 

"If you forget to bind `this.handleClick` and pass it to `onClick`, `this` will be `undefined` when the function is actually called."

 `this` will point to "window" object (in browser), not be undefined. Probably author meant `this.setState` will be undefined.


es-repo:
Closing it because in "strict" mode `this` will be actually undefined.

PR #695, opened 2018-03-20
Change "Wrapped" to "Original" in HOC docs

callumlocke:
When dealing with a higher order component functions, what does `WrappedComponent` mean? Is it the outer component that *got* wrapped, or is it the inner component that *gets* wrapped?

The React docs opt for the latter interpretation: `WrappedComponent` refers to the the original, uh, unwrapped component.

Arguments about intuitiveness aside, the name is at best ambiguous, so I propose changing it to `OriginalComponent`, as there is no reasonable misinterpretation of this name.

### Use in other libraries

It's true that other libraries also use `WrappedComponent` to refer to the unwrapped component. For example, a react-intl component provides access to the original, unwrapped component through `IntlComponent.WrappedComponent` for testing purposes. But the change I'm proposing should not introduce any conflict, because this proposal is not to repurpose the name `WrappedComponent` but to just stop using it.

### DIFF
--- a/content/docs/higher-order-components.md
+++ b/content/docs/higher-order-components.md
@@ -9,7 +9,7 @@ A higher-order component (HOC) is an advanced technique in React for reusing com
 Concretely, **a higher-order component is a function that takes a component and returns a new component.**
 
 ```js
-const EnhancedComponent = higherOrderComponent(WrappedComponent);
+const EnhancedComponent = higherOrderComponent(OriginalComponent);
 ```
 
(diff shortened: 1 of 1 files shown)


BartoszKlonowski:
Hello @callumlocke!
Unfortunately, the content you are modifying in this PR is no longer present in the latest docs, so let me close this PR as outdated.
However, please feel welcomed to create new PRs for any other open issues.
Thank you for taking the time to contribute 👍

PR #697, opened 2018-03-21
fix word in reconciliation doc: "controlled" -> "uncontrolled"

metametadata:
Looking at linked CodePen examples and based on  https://medium.com/@jedwards8/uncontrolled-components-are-an-anti-pattern-abbdd86fd39e I think it's a more correct wording.

### DIFF
--- a/content/docs/reconciliation.md
+++ b/content/docs/reconciliation.md
@@ -140,7 +140,7 @@ When that's not the case, you can add a new ID property to your model or hash so
 
 As a last resort, you can pass an item's index in the array as a key. This can work well if the items are never reordered, but reorders will be slow.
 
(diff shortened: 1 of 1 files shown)


alexkrolick:
`Controlled` is correct - the issue is event handlers that are bound to array indexes.

metametadata:
Thank you for a quick response. Is there any more info about the issue with "event handlers that are bound to array indexes" somewhere? I'm confused because CodePen examples from the article demonstrate uncontrolled inputs only (`<input/>`).

alexkrolick:
Did you try the Codepen examples from the docs as well? https://reactjs.org/docs/reconciliation.html#keys (last paragraph).

https://codepen.io/alexkrolick/pen/WzjRRB?editors=0010

metametadata:
Yes, these are the examples I meant. If I understand everything correctly, according to [glossary](https://reactjs.org/docs/glossary.html#controlled-vs-uncontrolled-components), the problematic input from the GIF with `1` entered is *uncontrolled*:

```html
<td>
    <input />
</td>
```

That's why I wanted to reword the doc.

PR #700, opened 2018-03-23
Clarify PureComponent render prop caveat

billyjanitsch:
The example given to illustrate the caveat with PureComponents that use render props doesn't quite make sense. The _binding_ isn't what ensures that `this.renderTheCat` refers to the same function when used in `render`, and the method will work just fine when passed to the render prop unbound because it doesn't reference `this`.

If anything, I think including a bound method in the example is dangerous, because if the method _did_ reference `this.props` or `this.state`, it would be incorrect to pass a bound version to the render prop because `Mouse` would fail to re-render when props/state change, as noted in the paragraph following the example.

This removes the binding and clarifies the reasoning.

/cc @mjackson

### DIFF
--- a/content/docs/render-props.md
+++ b/content/docs/render-props.md
@@ -305,14 +305,8 @@ To get around this problem, you can sometimes define the prop as an instance met
 
 ```js
 class MouseTracker extends React.Component {
-  constructor(props) {
-    super(props);
-
-    // This binding ensures that `this.renderTheCat` always refers
-    // to the *same* function when we use it in render.
-    this.renderTheCat = this.renderTheCat.bind(this);
-  }
-
+  // Defined as an instance method, `this.renderTheCat` always
+  // refers to *same* function when we use it in render
   renderTheCat(mouse) {
     return <Cat mouse={mouse} />;
   }
@@ -328,4 +322,4 @@ class MouseTracker extends React.Component {
 }
 ```
 
-In cases where you cannot bind the instance method ahead of time in the constructor (e.g. because you need to close over the component's props and/or state) `<Mouse>` should extend `React.Component` instead.
+In cases where you cannot define the prop statically (e.g. because you need to close over the component's props and/or state) `<Mouse>` should extend `React.Component` instead.


content/docs/render-props.md
```diff
@@ -305,14 +305,8 @@ To get around this problem, you can sometimes define the prop as an instance met
...
-    // This binding ensures that `this.renderTheCat` always refers
-    // to the *same* function when we use it in render.
-    this.renderTheCat = this.renderTheCat.bind(this);
-  }
-
+  // Defined as an instance method, `this.renderTheCat` always
+  // refers to *same* function when we use it in render
   renderTheCat(mouse) {
     return <Cat mouse={mouse} />;
   }
```
Kovensky: In cases like this, I'd actually prefer to put it as a function completely outside the class to begin with. It's better for type checking (no `this` allowed), minified code size (local identifier, no associated object prop) and runtime memory size (don't need to keep a reference to the function in the prototype).

Kovensky:
@billyjanitsch 
> If anything, I think including a bound method in the example is dangerous, because if the method did reference this.props or this.state, it would be incorrect to pass a bound version to the render prop because Mouse would fail to re-render when props/state change

It would be _more incorrect_ to pass a non-bound version to the render prop because, depending on how the prop is called, it may receive `undefined`, `window`, or even another component's `props` object as its `this` value. In that very example, the `this` value in an unbound `renderTheCat` will be the `this.props` object of `Mouse`.

If `Mouse` causes an update on itself by calling `this.setState()` on itself, that update _should_ bypass the implicit `shouldComponentUpdate` check of its parent, so the problem you describe should not happen. Alternatively, if it does, it would mean it's not possible to use other components with state inside a `PureComponent`.

billyjanitsch:
@Kovensky 

>It would be more incorrect to pass a non-bound version to the render prop because, depending on how the prop is called, it may receive undefined, window, or even another component's props object as its this value. In that very example, the this value in an unbound renderTheCat will be the this.props object of Mouse.

Of course. I'm not saying that it would be correct to pass the unbound method if it referenced `this`. I'm saying that it would be incorrect to pass the method bound _or_ unbound (see below).

>If Mouse causes an update on itself by calling this.setState() on itself, that update should bypass the implicit shouldComponentUpdate check of its parent, so the problem you describe should not happen.

The problem isn't when `Mouse` causes an update on itself. It's when you might expect `Mouse` to re-render because props/state above that your renderer closes over have changed, but it doesn't, because the renderer itself hasn't changed by reference. Consider this snippet:

```js
class Foo extends PureComponent {
  render() {
    return this.props.render('abc')
  }
}

class Bar extends Component {
  constructor(props, context) {
    super(props, context)
    this.renderBaz = this.renderBaz.bind(this)
  }

  renderBaz(value) {
    return value + this.props.value
  }

  render() {
    return <Foo render={this.renderBaz} />
  }
}

<Bar value="def" /> // renders "abcdef"
<Bar value="ghi" /> // still renders "abcdef"
// Foo doesn't re-render, because none of its props have changed by reference
```

It's a pretty subtle mistake, which is why I think it's dangerous to suggest binding as a solution, especially when the binding isn't even necessary in the particular example.

Kovensky:
Your example is another pattern entirely, compared to the example. The example is a PureComponent using a Component; your example is the other way around.

Arguably, components that call render props cannot ever be PureComponent.

billyjanitsch:
>Your example is another pattern entirely, compared to the example. The example is a PureComponent using a Component; your example is the other way around.

I don't think so. In the example, `Mouse` extends PureComponent and `MouseTracker` extends Component. `MouseTracker` uses `Mouse`.

Kovensky:
Yeah, the `Mouse` would either be incorrect to be a PureComponent, or the component using `Mouse` would be incorrect for not actually giving it a real pure function as callback.

Unless you want to require actual pure functions as callbacks, the component calling the callback cannot be PureComponent.

content/docs/render-props.md
```diff
@@ -328,4 +322,4 @@ class MouseTracker extends React.Component {
 }
 ```
 
-In cases where you cannot bind the instance method ahead of time in the constructor (e.g. because you need to close over the component's props and/or state) `<Mouse>` should extend `React.Component` instead.
+In cases where you cannot define the prop statically (e.g. because you need to close over the component's props and/or state) `<Mouse>` should extend `React.Component` instead.
```
HunderlineK: Is it not a breach of Dependency Inversion principle? The `Mouse` component cannot possibly know whether the components that will use it will pass over static props or not

PR #704, opened 2018-03-23
fixed typos

dheerajteta:
fixed typos


yangshun:
@damaera This is not a typo. The example is not JSX, it's the `onclick` attribute in plain HTML. Refer to [onclick docs](https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XUL/Attribute/onclick)

gaearon:
This is intentional (the example is showing HTML which is case insensitive)

PR #707, opened 2018-03-25
Update forms.md

m-combinator:
Edited the code to decide on the state value based on event.target.name rather than event.target.type to reflect better the explanation given in the preceding paragraph:

When you need to handle multiple controlled `input` elements, you can add a `name` attribute to each element and let the handler function choose what to do based on the value of `event.target.name`.


threepointone:
This was fixed separately, but thank you for the PR!

PR #717, opened 2018-03-27
Add a warning about the synthetic event system's propagation logic

aweary:
Stemming from https://github.com/facebook/react/issues/12328 and other issues we've seen in the past, people sometimes aren't aware that the synthetic event system's propagation will sometimes be out of sync with the native event system.

Here's a quick attempt at adding a warning about it. Feedback welcome!


IPWright83:
@aweary Thanks! :)

PR #719, opened 2018-03-27
Use examples without types in the post

gaearon:
In response to feedback in https://news.ycombinator.com/item?id=16693633.

Seeing type signatures upfront is a little intimidating, especially because it gives the feeling you need to know what `$Shape` means etc. I hid it in a gist.


gaearon:
screen shot 2018-03-28 at 00 35 24
screen shot 2018-03-28 at 00 35 29

examples/update-on-async-rendering/definition-getsnapshotbeforeupdate.js
```diff
@@ -0,0 +1,9 @@
+class Example extends React.Component {
+  getSnapshotBeforeUpdate(prevProps, prevState) {
+    // ...
```
acdlite: Maybe return snapshot here?

gaearon: Then it'll be a fully fledged example. Maybe I should just remove `componentDidUpdate`? There's a full example below anyway.

gaearon: I'll remove cDU to focus this on the new method signature.

bvaughn:
Thanks dan!

PR #732, opened 2018-03-29
Update tutorial.md

vuhuucuong:
Update `jumpTo` method, so the **Showing the Moves** list will be update immediately after click `Go to move #...`, not after click some squares.


rickhanlonii:
Sorry for the long delay here. Since this PR was created against the old site I'm going to close this, but if you think this is still needed on the new doc site please either file an issue or submit a new PR.

PR #737, opened 2018-03-30
Add JS Party to the podcasts page

adamstac:
Add us to the list 🎉

### DIFF
--- a/content/community/podcasts.md
+++ b/content/community/podcasts.md
@@ -12,6 +12,8 @@ Podcasts dedicated to React and individual podcast episodes with React discussio
 
 - [The React Podcast](http://reactpodcast.com) - The podcast about everything React.js, hosted by [React Training](https://reacttraining.com)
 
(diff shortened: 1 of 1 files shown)


gaearon:
There's not as much React specific content there, is it? I guess we could say the same for JS Air so maybe we shouldn't have included it.

adamstac:
It comes up often enough to be included as a resource IMO.

adamstac:
Heads up!

We may not feature React as the primary conversation, but we do talk about it often. For example — we talked about [Reactivesearch](https://opensource.appbase.io/reactivesearch/) in the most recent episode ~> https://changelog.com/jsparty/20

It would be a shame to not include it. Folks looking for good JavaScript and React content and conversations would certainly enjoy our community focused podcast.

BartoszKlonowski:
Hello @adamstac! Unfortunately the podcasts list page is no longer maintained.
Let me close this PR as outdated, but please feel invited to create new PRs to open issues.
Thank you!

PR #738, opened 2018-03-30
React.forwardRef accepts a render function, not Functional Component

koba04:
`React.forwardRef` accepts a function like `render(props: Props, ref: Refs)`, which is similar with Functional Component but Functional Component accepts a legacy context as the 2nd argument.

I know `React.forwardRef` works fine with Functional Component in most case but  I think it should distinguish between them.

### DIFF
--- a/examples/16-3-release-blog-post/forward-ref-example.js
+++ b/examples/16-3-release-blog-post/forward-ref-example.js
@@ -6,7 +6,11 @@ function withTheme(Component) {
         {theme => (
           // Assign the custom prop "forwardedRef" as a ref
           // highlight-next-line
-          <Component {...rest} ref={forwardedRef} theme={theme} />
+          <Component
+            {...rest}
+            ref={forwardedRef}
+            theme={theme}
+          />
         )}
       </ThemeContext.Consumer>
     );
--- a/examples/16-3-release-blog-post/forward-ref-example.js
+++ b/examples/16-3-release-blog-post/forward-ref-example.js
@@ -20,9 +20,21 @@ function withTheme(Component) {
   // We can pass it along to ThemedComponent as a regular prop, e.g. "forwardedRef"
   // And it can then be attached to the Component.
   // highlight-range{1-3}
-  return React.forwardRef((props, ref) => (
-    <ThemedComponent {...props} forwardedRef={ref} />
-  ));
+  function refForwarder(props, ref) {
+    return (
+      <ThemedComponent {...props} forwardedRef={ref} />
+    );
+  }
+
+  // These next lines are not necessary,
+  // But they do give the component a better display name in DevTools,
+  // e.g. "ForwardRef(withTheme(MyComponent))"
+  // highlight-range{1-2}
+  const name = Component.displayName || Component.name;
+  refForwarder.displayName = `withTheme(${name})`;
+
+  // highlight-next-line
+  return React.forwardRef(refForwarder);
 }
 
 // highlight-next-line
--- a/examples/16-3-release-blog-post/forward-ref-example.js
+++ b/examples/16-3-release-blog-post/forward-ref-example.js
@@ -1,27 +1,24 @@
 function withTheme(Component) {
-  // Note the second param "ref" provided by React.forwardRef.
-  // We can attach this to Component directly.
-  // highlight-range{1,5}
-  function ThemedComponent(props, ref) {
+  // highlight-next-line
+  function ThemedComponent({forwardedRef, ...rest}) {
     return (
       <ThemeContext.Consumer>
         {theme => (
-          <Component {...props} ref={ref} theme={theme} />
+          // Assign the custom prop "forwardedRef" as a ref
+          // highlight-next-line
+          <Component {...rest} ref={forwardedRef} theme={theme} />
         )}
       </ThemeContext.Consumer>
     );
   }
 
-  // These next lines are not necessary,
-  // But they do give the component a better display name in DevTools,
-  // e.g. "ForwardRef(withTheme(MyComponent))"
-  // highlight-range{1-2}
-  const name = Component.displayName || Component.name;
-  ThemedComponent.displayName = `withTheme(${name})`;
-
-  // Tell React to pass the "ref" to ThemedComponent.
-  // highlight-next-line
-  return React.forwardRef(ThemedComponent);
+  // Note the second param "ref" provided by React.forwardRef.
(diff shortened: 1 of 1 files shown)


gaearon:
I agree this is clearer and would also prefer this. @bvaughn?

examples/16-3-release-blog-post/forward-ref-example.js
```diff
@@ -1,27 +1,28 @@
...
-  ThemedComponent.displayName = `withTheme(${name})`;
-
-  // Tell React to pass the "ref" to ThemedComponent.
-  // highlight-next-line
-  return React.forwardRef(ThemedComponent);
+  // Note the second param "ref" provided by React.forwardRef.
+  // We can pass it along to ThemedComponent as a regular prop, e.g. "forwardedRef"
+  // And it can then be attached to the Component.
+  // highlight-range{1-3}
+  return React.forwardRef((props, ref) => (
```
bvaughn: I don't think we should encourage passing anonymous functions to `React.forwardRef` because it means DevTools won't be able to show a meaningful display name. (This is why the previous example forwarded `displayName`.)

bvaughn: DevTools shows `ForwardRef` only if the function is anonymous. If the function has a name, DevTools will show `ForwardRef(ComponentName)` which is much more useful. This is why the original code example I showed had the name copying.

bvaughn: > I agree with you. But I think distinguishing between the argument of `React.forwardRef` and Functional Component is more important than that.

I don't see them as mutually exclusive. Why not something like this?

```js
function withTheme(Component) {
  // highlight-next-line
  function ThemedComponent({ forwardedRef, ...rest }) {
    return (
      <ThemeContext.Consumer>
        {theme => (
          // Assign the custom prop "forwardedRef" as a ref
          // highlight-next-line
          <Component {...rest} ref={forwardedRef} theme={theme} />
        )}
      </ThemeContext.Consumer>
    );
  }

  // Note the second param "ref" provided by React.forwardRef.
  // We can pass it along to ThemedComponent as a regular prop, e.g. "forwardedRef"
  // And it can then be attached to the Component.
  // highlight-range{1-3}
  function refForwarder(props, ref) {
    return <ThemedComponent {...props} forwardedRef={ref} />;
  }

  // These next lines are not necessary,
  // But they do give the component a better display name in DevTools,
  // e.g. "ForwardRef(withTheme(MyComponent))"
  // highlight-range{1-2}
  const name = Component.displayName || Component.name;
  refForwarder.displayName = `withTheme(${name})`;

  // highlight-next-line
  return React.forwardRef(refForwarder);
}

// highlight-next-line
const fancyButtonRef = React.createRef();

// fancyButtonRef will now point to FancyButton
// highlight-range{4}
<FancyThemedButton
  label="Click me!"
  onClick={handleClick}
  ref={fancyButtonRef}
/>;
```

> In addition to that, it might cause a problem with static type checking.

Please give an example of this.

koba04: > I don't see them as mutually exclusive. Why not something like this?

I don't think they are exclusive, but I was not sure what a value `React.forwardRef` accepts?

I thought it accepts a callback function, not Functional Component.
But if the argument is a Functional Component, I think your example makes sense.
Feel free to close this.

> Please give an example of this.

`React.forwardRef` accepts `(props: Props, ref: Ref) => ReactElement`, which seems to be imcompatible with the type definition.

https://github.com/facebook/flow/blob/master/lib/react.js#L132

Actually, the `context` is specified as `any` so the incompatibility does not cause an error though.

bvaughn: > I thought it accepts a callback function, not Functional Component.

`forwardRef` _does_ accept a function. To be clear, the only difference in signature between the two is that a "functional component" accepts `(props, context)` as its parameters and the `forwardRef` callback accepts `(props, ref)`.

I can see how this similarity could cause confusion.

In my example above, I used a lowercase name for `refForwarder` to try to make it look a little less like a typical functional component name.

> > In addition to that, it might cause a problem with static type checking.

> Please give an example of this.

I think your original comment about this was based on a misunderstanding. Hopefully my comment above clears things up but if not let me know and I'll elaborate.

koba04: @bvaughn Thanks! That's good for me. I've updated this PR.

gaearon:
I still feel like explaining intricacies of `displayName` assignment gets in the way of communicating what this feature is for, and doesn't make the blog post very helpful. Instead, the indirection makes me skip over the whole example because I find it too difficult to comprehend. I'd still prefer the original version suggested by @koba04 (and I'd remove the `displayName` logic here altogether). We could keep this logic as a "tip" in ForwardRef doc instead.

I don't have a strong opinion though.

bvaughn:
Hm 🙁

People will likely copy-paste our example though. If we don't show the naming bit, everyone's DevTools will likely end up with the anonymous forward ref label.

I don't feel strongly about this. If you both think the previous, anonymous function version was more readable- I'll defer to your judgement. 😄

(Sorry for the back and forth on this, @koba04)

gaearon:
Blog post is introductory. With time people will start copy-pasting from the main doc. The main doc can have a section that shows an example with a screenshot. People who write HOCs probably will read this doc in depth because it's not a beginner-friendly feature anyway, and by the point they need it they're probably comfortable reading docs.

bvaughn:
I think I'm probably too close to this to trust my opinion. I trust yours, Dan.

gaearon:
Let's add a link to main doc ("Forwarding Refs") after the code example. In the main doc, let's add a separate section about "Customizing Name in DevTools".

bvaughn:
I like that suggestion.

bvaughn:
This looks nice. Thanks for the change, @koba04. Dan or I can make the follow up change he proposed. (Or you could in another diff, if you're interested.)

gaearon:
Related: I found it a bit overwhelming that we have three very similar examples but with slightly different naming in the blog post, in Forwarding Refs doc, and in `forwardRef()` reference.

bvaughn:
Yeah, maybe.

I thought sharing a use case (e.g. theming) might make things easier for people, since they wouldn't need to stop and think about a new use case and why you'd want to use it, etc.

koba04:
Thanks! I'll try to do the follow-up changes!

PR #740, opened 2018-03-30
Use correct component name in context doc example

CMTegner:
I'm assuming the intention was to reference the `FancyButton` component defined directly above, unless I'm misunderstanding how to read this example.


gaearon:
Please run `yarn prettier`?

PR #741, opened 2018-03-30
Update forwarding-refs-fancy-button.js

Saifadin:
This should be `FancyButton` or else it wouldn't make sense


Saifadin:
Great 👍🏽

PR #742, opened 2018-03-30
a11y-doc: Adjust Focus Control to new 16.3 refs + link fixes.

AlmeroSteyn:
With release of React 16.3 and the changes to the **ref** API I have adjusted the accessibility doc to reflect this.

I also fixed a few dead/broken links I found on this page.

I have chosen not to refer to `forwardRef` in the a11y doc as it is more for creators of HOC's and trying to explain it again on this page would detract from achieving accessibility IMHO.

Content changes are limited to the **Programmatically managing focus** section. Changes are summarised in the following captures:

image

image

### DIFF
--- a/content/docs/accessibility.md
+++ b/content/docs/accessibility.md
@@ -142,18 +142,25 @@ we need to programmatically nudge the keyboard focus in the right direction. For
 
 MDN Web Docs takes a look at this and describes how we can build [keyboard-navigable JavaScript widgets](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Keyboard-navigable_JavaScript_widgets).
 
-To set focus in React, we can use [Refs to DOM elements](refs-and-the-dom.html).
+To set focus in React, we can use [Refs to DOM elements](/docs/refs-and-the-dom.html).
 
 Using this, we first create a ref to an element in the JSX of a component class:
 
-```javascript{2-3,7}
+```javascript{4-5,8-9,13}
 render() {
+  constructor(props) {
+    super(props);
+    // create a ref to store the textInput DOM element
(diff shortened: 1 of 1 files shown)


AlmeroSteyn:
NOTE: I accidentally duplicated #708 in this fix.

content/docs/accessibility.md
```diff
@@ -163,33 +170,43 @@ Then we can focus it elsewhere in our component when needed:
...
+}
 />
 
 // Now you can set focus when required.
-this.inputEl.focus();
+this.inputElement.current.focus();
 ```
 
-A great focus management example is the [react-aria-modal](https://github.com/davidtheclark/react-aria-modal). This is a relatively rare example of a fully accessible modal window. Not only does it set initial focus on 
+When using a HOC to extend components, the ref can also be passed through the HOC to the wrapped component using the pattern above.
```
gaearon: Maybe we should promote `React.forwardRef()` here instead?

AlmeroSteyn:
@gaearon I have adjusted the part about HOC's to mention ref forwarding. 

To avoid too much code example duplication and keep focus on the a11y issue at hand I have still added a code example as this can be found through the link.

The sentence has now become:

>When using a HOC to extend components, it is recommended to [forward the ref](/docs/forwarding-refs.html) to the wrapped component using the `forwardRef` function of React. If a third party HOC
does not implement ref forwarding, the above pattern can still be used as a fallback.

 What do you think?

gaearon:
Sounds good. Mind rebasing?

AlmeroSteyn:
Done!

AlmeroSteyn:
Wow that was super quick! Thank you!

PR #745, opened 2018-03-30
Fix cDU to correctly make async request

nateq314:
When a new `id` prop is passed in, `getDerivedStateFromProps` sets `this.state.externalData` to `null`. If I understand the new API correctly the return value of gDSFP() either becomes the new state or is merged into the prev state? If so, then `componentDidUpdate` should examine the new state (`this.state.externalData`), not `prevState.externalData`. Otherwise the async request will never happen.

### DIFF
--- a/examples/update-on-async-rendering/updating-external-data-when-props-change-after.js
+++ b/examples/update-on-async-rendering/updating-external-data-when-props-change-after.js
@@ -25,7 +25,7 @@ class ExampleComponent extends React.Component {
 
   // highlight-range{1-5}
   componentDidUpdate(prevProps, prevState) {
-    if (prevState.externalData === null) {
+    if (this.state.externalData === null) {
       this._loadAsyncData(this.props.id);
     }
   }


nateq314:
Glad to help! Unrelated question: Any suggestions for familiarizing oneself with or attempting to wrap ones head around the React codebase? I'm really interested in understanding the internals and contributing somehow in the future but it's so intimidating in terms of sheer lines of code. Not even sure where to start. Other than taking a core package like react-reconciler and proceeding directly to a line by line deep dive, is there some other way to get a high-level understanding of the major components/structures and their relationships? For example stuff like, what exactly is a fiber, what is their structure and behavior and how do they interact. How does the upcoming async-mode work in terms of prioritizing/scheduling updates from a high level whiteboard point of view. How does reconciliation work. Etc etc. But with a focus on code and internals, not the API doc stuff on reactjs.org. Any tips you might have? Just asking. No offense if you don't have time to answer. Know you guys are all super busy between work and family.

gaearon:
There’s an outdated document here but it’s pretty old and some things changed. https://github.com/acdlite/react-fiber-architecture

There’s some background that’s still relevant here. https://github.com/facebook/react/issues/7942

This might be helpful. https://engineering.hexacta.com/didact-fiber-incremental-reconciliation-b2fe028dcaec

Other than that we don’t currently have a lot of info but I’d encourage you to look at five  files:

* ReactFiberBeginWork (“entering” a component)
* ReactFiberCompleteWork (“leaving” a component)
* ReactFiberCommitWork (flushing changes to DOM)
* ReactFiberScheduler (choosing next component to work on)
* ReactChildFiber (diffing insertions/deletions of children)

I’m happy to answer questions (perhaps in a separate issue) and it would be great if you could write a doc with your learnings 🙂

nateq314:
Thank you! Will do.

PR #746, opened 2018-03-30
Minor: context/theme-detailed-app.js broken

onpaws:
Today I tested out the new Context API using using [CRA](https://github.com/facebook/create-react-app). 
I copy/pasted the example in, and was surprised to encounter a runtime exception. 
So this tiny PR is intended to make things easier for future Context API learners.

image

`ReferenceError: Must call super constructor in derived class before accessing 'this' or returning from derived constructor`


onpaws:
Looks like it worked. Courtesy of Netlify:
image

gaearon:
Thanks!

PR #747, opened 2018-03-31
Recommend getDerivedStateFromProps

chucksellick:
Fix a couple of paragraphs that recommended UNSAFE_componentWillReceiveProps
where it would now be more correct to recommend getDerivedStateFromProps.


gaearon:
Thanks

PR #758, opened 2018-04-03
Remove overwriting a default rule from .prettierrc

HarrisJT:
According to http://json.schemastore.org/prettierrc, printWidth default is already 80.


alexkrolick:
Let's keep it to be explicit. The examples folder rc file use a different width.

PR #767, opened 2018-04-04
Add a section "Customizing Name in DevTools" in Forwarding Refs

koba04:
This is a follow-up PR of https://github.com/reactjs/reactjs.org/pull/738#issuecomment-377989360
I've added a "Customizing Name in DevTools" section and added a link to it into a blog post about v16.3.

> Related: I found it a bit overwhelming that we have three very similar examples but with slightly different naming in the blog post, in Forwarding Refs doc, and in forwardRef() reference.

https://github.com/reactjs/reactjs.org/pull/738#issuecomment-377990615

Should I use the same example through the pages?

### DIFF
--- a/content/blog/2018-03-29-react-v-16-3.md
+++ b/content/blog/2018-03-29-react-v-16-3.md
@@ -52,6 +52,8 @@ HOCs typically [pass props through](/docs/higher-order-components.html#conventio
 The new `forwardRef` API solves this problem by providing a way for us to intercept a `ref` and forward it as a normal prop:
 `embed:16-3-release-blog-post/forward-ref-example.js`
 
+[Learn more about the forwardRef API here.](/docs/forwarding-refs.html)
+
 ## Component Lifecycle Changes
 
 React's class component API has been around for years with little change. However, as we add support for more advanced features (such as [error boundaries](/docs/react-component.html#componentdidcatch) and the upcoming [async rendering mode](/blog/2018/03/01/sneak-peek-beyond-react-16.html)) we stretch this model in ways that it was not originally intended.
--- a/content/docs/forwarding-refs.md
+++ b/content/docs/forwarding-refs.md
@@ -28,7 +28,7 @@ For example, the following component will appear as "*ForwardRef*" in the DevToo
 
 `embed:forwarding-refs/wrapped-component.js`
 
-If you name the render function, DevTools will include its name (e.g. "*ForwardRef(myFunction)*"):
+If you name the render function, DevTools will also include its name (e.g. "*ForwardRef(myFunction)*"):
 
 `embed:forwarding-refs/wrapped-component-with-function-name.js`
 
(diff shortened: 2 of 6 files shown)


content/blog/2018-03-29-react-v-16-3.md
```diff
@@ -52,6 +52,8 @@ HOCs typically [pass props through](/docs/higher-order-components.html#conventio
 The new `forwardRef` API solves this problem by providing a way for us to intercept a `ref` and forward it as a normal prop:
 `embed:16-3-release-blog-post/forward-ref-example.js`
 
+[Learn more about customizing name in DevTools here.](/docs/forwarding-refs.html#customizing-name-in-devtools)
```
bvaughn: Let's make this link more generic, like the ones above. (aka "Learn more about the `forwardRef` API here" and then link to /docs/forwarding-refs.html)

content/docs/forwarding-refs.md
```diff
@@ -20,3 +20,8 @@ This means that refs intended for our `FancyButton` component will actually be a
 Fortunately, we can explicitly forward refs to the inner `FancyButton` component using the `React.forwardRef` API. `React.forwardRef` accepts a render function that receives `props` and `ref` parameters and returns a React node. For example:
 `embed:forwarding-refs/log-props-after.js`
 
+## Customizing Name in DevTools
```
bvaughn: Wording suggestion:
> Displaying a custom name in DevTools

content/docs/forwarding-refs.md
```diff
@@ -20,3 +20,8 @@ This means that refs intended for our `FancyButton` component will actually be a
...
+
+`React.forwardRef` accepts a render function. You can give the render function a better display name in DevTools.
```
bvaughn: Maybe we could make this section a little more detailed?

---

`React.forwardRef` accepts a render function. React DevTools uses this function to determine what to display for the ref forwarding component.

For example, the following component will appear as "_ForwardRef_" in the DevTools:
```js
const WrappedComponent = React.forwardRef((props, ref) => {
  return <LogProps {...props} forwardedRef={ref} />;
});
```

If you name the render function, DevTools will include its name (e.g. "_ForwardRef(myFunction)_"):
```js
const WrappedComponent = React.forwardRef(
  function myFunction(props, ref) {
    return <LogProps {...props} forwardedRef={ref} />;
  }
);
```

You can even set the function's `displayName` property to include the component you're wrapping:
```js
function logProps(Component) {
  class LogProps extends React.Component {
    // ...
  }

  function forward(props, ref) {
    return <LogProps {...props} forwardedRef={ref} />;
  }

  // Give this component a more helpful display name in DevTools.
  // e.g. "ForwardRef(logProps(MyComponent))"
  const name = Component.displayName || Component.name;
  forward.displayName = `logProps(${name})`;

  return React.forwardRef(forward);
}
```

examples/forwarding-refs/customizing-name.js
```diff
@@ -0,0 +1,17 @@
+function logProps(Component) {
+  class LogProps extends React.Component {
+    // ...
+  }
+
+  function forwardRef(props, ref) {
+    return <LogProps {...props} forwardedRef={ref} />;
+  }
+
+  // They do give the component a better display name in DevTools,
```
bvaughn: Wording suggestion:

> Give this component a more helpful display name in DevTools.

koba04:
@bvaughn Thank you for your reviews! I've fixed them.

koba04:
Thanks!

PR #769, opened 2018-04-05
add reference to react-testing-library

kentcdodds:
because it will help engineers write better tests with react. :)

### DIFF
--- a/content/docs/addons-test-utils.md
+++ b/content/docs/addons-test-utils.md
@@ -20,6 +20,8 @@ var ReactTestUtils = require('react-dom/test-utils'); // ES5 with npm
 > Note:
 >
 > Airbnb has released a testing utility called Enzyme, which makes it easy to assert, manipulate, and traverse your React Components' output. If you're deciding on a unit testing utility to use together with Jest, or any other test runner, it's worth checking out: [http://airbnb.io/enzyme/](http://airbnb.io/enzyme/)
+>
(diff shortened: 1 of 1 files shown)


content/docs/addons-test-utils.md
```diff
@@ -20,6 +20,8 @@ var ReactTestUtils = require('react-dom/test-utils'); // ES5 with npm
 > Note:
 >
 > Airbnb has released a testing utility called Enzyme, which makes it easy to assert, manipulate, and traverse your React Components' output. If you're deciding on a unit testing utility to use together with Jest, or any other test runner, it's worth checking out: [http://airbnb.io/enzyme/](http://airbnb.io/enzyme/)
+>
+> Alternatively, there is another testing utility called react-testing-library. This is a simple and small utility that helps your tests give you confidence your components are working by encouraging you to write tests that use your components as the end users use them. It also works with any test runner: [https://git.io/react-testing-library](https://git.io/react-testing-library)
```
alexkrolick: ~~simple~~

~~small~~

kentcdodds: Happy to make the change, but I'm curious what you have against the language. When compared to the previous recommendation of enzyme it is smaller and simpler so I thought it was a useful distinction to make.

alexkrolick: IMO of course

- "Simple" is subjevtive and there's kind of a campaign to keep it out the docs (cc @acdlite) 
- Small could be more precise

> Alternatively, there is another testing utility called react-testing-library designed to encouraging writing tests that use your components as the end users use them.

kentcdodds: I like your rewording. Mine was too wordy. I'll update. Thanks.

PR #770, opened 2018-04-05
Clarify when Consumers get the default value

alexkrolick:
https://twitter.com/timarney/status/981912479730159617

### DIFF
--- a/content/docs/context.md
+++ b/content/docs/context.md
@@ -45,9 +45,9 @@ Using context, we can avoid passing props through intermediate elements:
 const {Provider, Consumer} = React.createContext(defaultValue);
 ```
 
-Creates a `{ Provider, Consumer }` pair.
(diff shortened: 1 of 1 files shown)


content/docs/context.md
```diff
@@ -59,6 +59,10 @@ A React component that allows Consumers to subscribe to context changes.
 
 Accepts a `value` prop to be passed to Consumers that are descendants of this Provider. One Provider can be connected to many Consumers. Providers can be nested to override values deeper within the tree.
 
+> Note:
+>
+> If `value` is not provided, Consumers receive `undefined`. To use the default value in a Consumer instead, remove the Provider altogether.
+
```
gaearon: I feel like this note is more confusing than what it explains. It seems to *suggest* removing a Provider as a solution to some problem but usually you don't want to do that. Maybe there's a way to reword where it doesn't sound like we *recommend* removing Providers.

gaearon:
I rewrote it, IMO this is a bit clearer

alexkrolick:
Thanks @gaearon that should do it

PR #771, opened 2018-04-05
Move constructor out of render

codecrutch:
Fix React.createRef example.

Before:
screen shot 2018-04-05 at 2 18 08 pm

After:
screen shot 2018-04-05 at 2 17 54 pm


alexkrolick:
😅 woops thanks

PR #780, opened 2018-04-08
Added updating nested context default value example

tjallingt:
Small addition to https://github.com/reactjs/reactjs.org/pull/776
It was missing an example showing how the default value for the context needs to change to match the value passed to the provider.

### DIFF
--- a/examples/context/updating-nested-context-context.js
+++ b/examples/context/updating-nested-context-context.js
@@ -4,7 +4,7 @@
 // highlight-range{2-4}
 export const ThemeContext = React.createContext({
   theme: themes.dark,
-  // optionally you can pass functions that log an error
+  // implement default behavior...
   toggleTheme: () =>
     console.error('default toggleTheme called'),
 });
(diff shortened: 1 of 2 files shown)


tjallingt:
Additionally it might be worthwhile to update the entire set of examples to instead show how to wrap the provider with a component holding its state and updaters instead of putting that in the App component...

Dunno if the added complexity makes that hard to understand to people 😕

tjallingt:
ping @alexkrolick since you helped out with this the last time, have another look?

examples/context/updating-nested-context-context.js
```diff
@@ -0,0 +1,10 @@
+// highlight-range{1-2}
+// Make sure the shape of the default value passed to
+// createContext matches the shape that the consumers expect!
+// highlight-range{2-4}
+export const ThemeContext = React.createContext({
+  theme: themes.dark,
+  // optionally you can pass functions that log an error
+  toggleTheme: () =>
+    console.error('default toggleTheme called'),
```
alexkrolick: how about just a comment here like `// implement default behavior...`

tjallingt:
On second thought maybe it should just be `toggleTheme: () => {}`

PR #783, opened 2018-04-09
Dynamically append path alias (aaa/bbb.html = aaa/bbb = aaa/bbb/) int…

### LINKED ISSUE #605: Test Utils page is available in Google search results, but the front-end redirects the visitor

whitedogg13:
This PR tries to fix #605 

The issue is following 404 pages, such as:

* [`https://reactjs.org/docs/hello-world`](https://reactjs.org/docs/hello-world)
* [`https://reactjs.org/docs/hello-world/`](https://reactjs.org/docs/hello-world/)

As we can see, only https://reactjs.org/docs/hello-world.html can work correctly. The common alias we consider the same as it won't work.

The brute force approach for this issue might be simply modifying all *.md files for the `redirect_from` to include the **alias** paths. However, I think a better approach will be to append those **alias** path dynamically during gatsby's `onCreateNode` API.

### DIFF
--- a/content/docs/reference-pure-render-mixin.md
+++ b/content/docs/reference-pure-render-mixin.md
@@ -3,7 +3,7 @@ id: pure-render-mixin
 title: PureRenderMixin
 layout: docs
 category: Reference
-permalink: docs/pure-render-mixin.html
+permalink: docs/pure-render-mixin-old.html
 ---
 
 > Note
--- a/gatsby/onCreateNode.js
+++ b/gatsby/onCreateNode.js
@@ -10,16 +10,14 @@
 const BLOG_POST_FILENAME_REGEX = /([0-9]+)\-([0-9]+)\-([0-9]+)\-(.+)\.md$/;
 
 function buildRedirectString(permalink, redirect_from) {
-  if (!permalink || permalink.indexOf('.html') === -1) {
-    return '';
+  if (!permalink || !permalink.endsWith('.html')) {
+    return redirect_from ? JSON.stringify(redirect_from) : '';
   }
 
-  let basePath = permalink.slice(0, permalink.indexOf('.html'));
+  let basePath = permalink.slice(0, -'.html'.length);
   let redirects = [basePath, basePath + '/'];
   if (Array.isArray(redirect_from)) {
     redirects = redirects.concat(redirect_from);
-  //} else if (redirect_from) {
-  //  redirects.push(redirect_from);
   }
 
   return JSON.stringify(redirects);
--- a/gatsby/onCreateNode.js
+++ b/gatsby/onCreateNode.js
@@ -9,20 +9,20 @@
 // Parse date information out of blog post filename.
 const BLOG_POST_FILENAME_REGEX = /([0-9]+)\-([0-9]+)\-([0-9]+)\-(.+)\.md$/;
 
-function addPathAliasToRedirects(permalink, redirect_from) {
-  if (!permalink || !redirect_from || permalink.indexOf('.html') === -1) {
-    return redirect_from;
+function buildRedirectString(permalink, redirect_from) {
+  if (!permalink || permalink.indexOf('.html') === -1) {
+    return '';
   }
 
   let basePath = permalink.slice(0, permalink.indexOf('.html'));
   let redirects = [basePath, basePath + '/'];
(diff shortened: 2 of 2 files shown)


gatsby/onCreateNode.js
```diff
@@ -9,6 +9,22 @@
 // Parse date information out of blog post filename.
 const BLOG_POST_FILENAME_REGEX = /([0-9]+)\-([0-9]+)\-([0-9]+)\-(.+)\.md$/;
 
+function buildRedirectString(permalink, redirect_from) {
+  if (!permalink || permalink.indexOf('.html') === -1) {
```
sophiebits: Instead of indexOf, can we do `!permalink.endsWith('.html')` so it won't catch if there is .html elsewhere in the name?

gatsby/onCreateNode.js
```diff
@@ -9,6 +9,22 @@
...
+    return '';
+  }
+
+  let basePath = permalink.slice(0, permalink.indexOf('.html'));
```
sophiebits: `.slice(0, -'.html'.length)`

gatsby/onCreateNode.js
```diff
@@ -9,6 +9,22 @@
 // Parse date information out of blog post filename.
 const BLOG_POST_FILENAME_REGEX = /([0-9]+)\-([0-9]+)\-([0-9]+)\-(.+)\.md$/;
 
+function buildRedirectString(permalink, redirect_from) {
+  if (!permalink || permalink.indexOf('.html') === -1) {
+    return '';
```
sophiebits: Should return redirect_from.

gatsby/onCreateNode.js
```diff
@@ -9,6 +9,22 @@
...
+  }
+
+  let basePath = permalink.slice(0, permalink.indexOf('.html'));
+  let redirects = [basePath, basePath + '/'];
+  if (Array.isArray(redirect_from)) {
+    redirects = redirects.concat(redirect_from);
+  //} else if (redirect_from) {
```
sophiebits: remove these commented lines

content/docs/reference-pure-render-mixin.md
```diff
@@ -3,7 +3,7 @@ id: pure-render-mixin
 title: PureRenderMixin
 layout: docs
 category: Reference
-permalink: docs/pure-render-mixin.html
+permalink: docs/pure-render-mixin-old.html
```
sophiebits: Is this file ever accessible? If not let's just delete it.

gatsby/onCreateNode.js
gaearon: Note: `endsWith()` isn't supported by IE unless we polyfill it

sophiebits: I sorta assumed this runs at build time. If not, `!/\.html$/.test(permalink)` would work.

gaearon: Oops. I think you're right

whitedogg13: Thanks, will do that!

whitedogg13:
@sophiebits Thanks for the solid review! I have learned something from them.

I follow most of your suggestions, with some minor differences listed below:

* I'm not sure if the old `content/docs/reference-pure-render-mixin.md` is indeed not used at all (since the `permalink` is duplicated, it's hard to grep), so I think it's safer to just rename it and keep it that way.
* for the case `(!permalink || !permalink.endsWith('.html'))`, directly return `redirect_from` might cause error since the `redirect_from` might be an array, but caller expect this function to return a string (so this function is named `buildRedirectString`, as a hint). I also consider the case agains stringifying `undefined` or `null` in the latest mod.  

Thanks!

whitedogg13:
Hi, @sophiebits just want to know if I should further update this PR? Looks like this issue is still there. (sorry for poking, I know you guys are definitely busy)

Thanks!

PR #785, opened 2018-04-10
Add SRI hashes for production usage of cdn react

magicmark:
In production, seems like best practice to provide the SRI hash to verify the integrity of a 3rd party script.

More info:
- https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity
- https://hacks.mozilla.org/2015/09/subresource-integrity-in-firefox-43/
- https://www.srihash.org/

Worth mentioning that this shouldn't break older browsers, since this will just be an extra attribute that will be ignored.

(It might seem overly verbose, but it seems like a responsible thing to do in the official React documentation, since it's probably the source of a lot of copy+pasted code.)

Thanks!

### DIFF
--- a/content/docs/cdn-links.md
+++ b/content/docs/cdn-links.md
@@ -36,7 +36,7 @@ We also recommend to verify that the CDN you are using sets the `Access-Control-
 
 This enables a better [error handling experience](/blog/2017/07/26/error-handling-in-react-16.html) in React 16 and later.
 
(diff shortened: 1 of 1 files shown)


Marcholio:
This has been open for a while, could this be merged soon?

rickhanlonii:
Sorry for the long delay here. Going to close since we don't publish UDM anymore, so there are no docs for this on the new site.

PR #790, opened 2018-04-10
Use the official talk video from YouTube

gaearon:
This version fixes the glitch in the original video: https://cl.ly/3j0b2h1X2S3R

We'll lose the view count but meh.


gaearon:
Yep.

PR #792, opened 2018-04-11
Add section about Pointer Events

philipp-spiess:
This PR adds a section about the state of Pointer Events in React.

This should be merged only if facebook/react#12507 is accepted as well.


philipp-spiess:
I guess this can be merged now :)

PR #797, opened 2018-04-12
add chromevox to a11y documentation

jlr7245:
Adding the truly excellent [ChromeVox extension](https://chrome.google.com/webstore/detail/chromevox/kgejglhpjiefppelpmljglcjbhoiplfn?hl=en) to the list of accessibility tools available.

### DIFF
--- a/content/docs/accessibility.md
+++ b/content/docs/accessibility.md
@@ -330,6 +330,8 @@ Testing with a screen reader should form part of your accessibility tests.
 
 Please note that browser / screen reader combinations matter. It is recommended that you test your application in the browser best suited to your screen reader of choice.
 
+### Commonly Used Screen Readers
+
 #### NVDA in Firefox
 
 [NonVisual Desktop Access](https://www.nvaccess.org/) or NVDA is an open source Windows screen reader that is widely used.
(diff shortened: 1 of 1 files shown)


gaearon:
cc @AlmeroSteyn @jessebeach does this look good to you?

AlmeroSteyn:
@gaearon Thank you for that! Would love to help, however the invite seems to be revoked?

@jlr7245 I love the idea of people adding to this doc, especially other good tools out there.

As Chrome is not yet represented as a supported browser in the list of screen readers I think this is a good idea, however I would suggest keeping the frequency of use in mind. Check out the [WebAIM Screen Reader Survey](https://webaim.org/projects/screenreadersurvey7/#primary).

The screen readers already listed are by far the most used and as testing with screen readers usually scare developers perhaps we should note that when adding something like ChromeVox. Also issues with these three have the most impact on users. So perhaps we should tell people that should they only pick one or two to start with, that it should rather be JAWS, NVDA or VoiceOver. 

Then there could possibly be a section for "Other screen readers" or something like that.

What do you think?

AlmeroSteyn:
Yeah good idea to get dev awareness this way.

As for talking about the survey, my feeling is that the doc should rather just use this info to nudge devs in the right direction. IMO just adding the headers as you mention should be enough.

jlr7245:
Should be all set!

content/docs/accessibility.md
```diff
@@ -330,6 +330,10 @@ Testing with a screen reader should form part of your accessibility tests.
 
 Please note that browser / screen reader combinations matter. It is recommended that you test your application in the browser best suited to your screen reader of choice.
 
+### Commonly Used Screen Readers
+
+Refer to the [WebAIM Screen Reader Survey](https://webaim.org/projects/screenreadersurvey7/#primary) for a listing of screen readers and their usages.
```
AlmeroSteyn: I still have the feeling that we should drop this line from the doc. 

By just having the headings of **Commonly used** and **Other** I think we are guiding the devs with what is in the survey. 

The survey has a lot of info and captured at a specific point in time so even today it could already be different. I don't feel that the React Docs is the place for that, rather that we should be using the a11y doc page to guide devs with the information we have. 

Can you remove this line?

AlmeroSteyn:
Looking good thanks, just the one point to still address :-)

gaearon:
I sent the invites again. If they don't follow through let me know.

gaearon:
Oh, maybe it's our open source permission system messing it up. I'll ask internally.

AlmeroSteyn:
@gaearon Brilliant thanks! This time it worked and I accepted.

jlr7245:
@AlmeroSteyn Changes made!

AlmeroSteyn:
Thank you for taking the time to add to this part of the docs!

PR #798, opened 2018-04-13
Add 'visual components' use case for forwarding refs

duailibe:
After a [discussion](https://twitter.com/duailibe/status/984587505801588736) on Twitter about the forwarding refs docs being an advanced use case, I was suggested to PR another use case which may happen quite frequently when using "visual components" (like a `FancyButton` that uses a `button` element) but you may still need a `ref` to the native element.

This is my first contribution to the docs and English is not my first language so please all suggestions are welcome!

### DIFF
--- a/content/docs/reference-react.md
+++ b/content/docs/reference-react.md
@@ -240,8 +240,8 @@ You can also use it with the shorthand `<></>` syntax. For more information, see
 
 `embed:reference-react-forward-ref.js`
 
-In the above example, React passes a `ref` given to `<FancyButton ref={ref}>` element as a second argument to the rendering function inside the `React.forwardRef` call. This rendering function further passes the `ref` to the `<button ref={ref}>` element.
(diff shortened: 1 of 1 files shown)


gaearon:
This is great. I'm going to edit it a bit and take it in. Thanks.

bvaughn:
Thanks for putting up a PR @duailibe! This looks like a nice improvement!

gaearon:
OK, I pushed a more comprehensive rewrite. I used this example as the base, expanded the explanation around it to really hammer how this feature works, and changed the blog post and the API reference to focus on the simple example. We still have the complex one on the page later, and we have a section both in HOC and in context docs with the complex one.

gaearon:
Blog post section: https://deploy-preview-798--reactjs.netlify.com/blog/2018/03/29/react-v-16-3.html#forwardref-api

Reference: https://deploy-preview-798--reactjs.netlify.com/docs/react-api.html#reactforwardref

Full doc: https://deploy-preview-798--reactjs.netlify.com/docs/forwarding-refs.html

PR #799, opened 2018-04-13
Fixed invalid getSnapshotBeforeUpdate() example code

bvaughn:
Realized the code example for `getSnapshotBeforeUpdate()` didn't make sense. It's been fixed (and can be played with [here](https://codesandbox.io/s/2w2zrnz89r)).

The two sections that have been updated are [the component reference](https://deploy-preview-799--reactjs.netlify.com/docs/react-component.html#getsnapshotbeforeupdate) and [this update recipe](https://deploy-preview-799--reactjs.netlify.com/blog/2018/03/27/update-on-async-rendering.html#reading-dom-properties-before-an-update).

Related Twitter thread: https://twitter.com/brian_d_vaughn/status/984604747511382016

### DIFF
--- a/examples/react-component-reference/get-snapshot-before-update.js
+++ b/examples/react-component-reference/get-snapshot-before-update.js
@@ -1,5 +1,8 @@
 class ScrollingList extends React.Component {
-  listRef = React.createRef();
+  constructor(props) {
+    super(props);
+    this.listRef = React.createRef();
+  }
 
   getSnapshotBeforeUpdate(prevProps, prevState) {
     // Are we adding new items to the list?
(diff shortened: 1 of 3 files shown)


examples/react-component-reference/get-snapshot-before-update.js
```diff
@@ -3,19 +3,21 @@ class ScrollingList extends React.Component {
```
gaearon: Just realized we're using class property syntax above. These are stable docs so we shouldn't be.

bvaughn:
Going to just merge this, since the current (live) example is bad. I'll follow up with another PR if there's feedback after the merge! 😄

bvaughn:
Weird. This didn't show as merged until after I submitted my comment. Sorry!

examples/react-component-reference/get-snapshot-before-update.js
bvaughn: Thanks for changing this!

Issue #85, opened 2017-10-08
Documentation versions


bvaughn:
This issue was originally reported by @shripadk via facebook/react/issues/2424

It would be great to have documentation for various versions of React similar to http://nodejs.org/docs/

Netlify already permanently caches deployments so we should be able to use it to link to older versions of the docs at specific times. The new website header also has a UI label for the current version that could easily be turned into a drop-down selector for toggling between past versions.

### How would the drop-down menu work?

We could store a list of snapshot deployments (eg 16.0, 16.1, etc) in a JSON or YML config file in GitHub and the latest deployment of the docs site could use it to build the list of links to older versions of the docs.

Older versions of the site could make an HTTP request to load the list dynamically from eg reactjs.org/some/static/path/versions.json. This would enable older versions to be updated with links back to newer versions of the docs.

### Why not build and deploy older versions of the docs?

One or two people have asked why not just build old versions of the docs now, with the newer site. My concerns with this would be:
* Gatsby's config and our view templates tie the markdown content and layout together in such a way that changes to either would make building older versions of the docs difficult (eg reorganizing our directory structure would require forking our Gatsby build process).
* It would complicate translation efforts via Crowdin.
* It would probably slow the build process down.

bjrmatos:
it will be also cool to have a reference to the initial version that an API was introduced.

example:

<img width="786" alt="captura de pantalla 2017-10-08 a las 3 21 32 p m" src="https://user-images.githubusercontent.com/4262050/31320593-a50ef340-ac3c-11e7-8e0d-7d2f6e1ebb14.png">

bvaughn:
Agreed that could be a nice addition as well! 😄

KyleAMathews:
An interesting option too would be to use [git-worktree](https://git-scm.com/docs/git-worktree) to checkout past branches to build older versions of the docs within Gatsby.

KyleAMathews:
It'd slow things down builds somewhat but since the older stuff will be cached after the first build, it might be quite workable still.

bvaughn:
What would be the benefit of building past branches like this vs just linking to snapshots in time?

It would give us the chance to improve old docs or fix bugs, but I'm not sure that's super important.

KyleAMathews:
It just feels like an odd navigation scheme to link to an old version of the website instead of just linking to older docs. But agree it's far simpler and probably perfectly adequate.

bvaughn:
[...]

PR #801, opened 2018-04-13
Added redirects for old React versions

bvaughn:
Use [Netlify's redirects API](https://www.netlify.com/docs/redirects/)- and the fact that Netlify caches immutable builds of older versions of the site- to expose previous versions of the React documentation via a [**new "React Versions" page**](https://deploy-preview-801--reactjs.netlify.com/versions).

I did the following to identify the correct revisions to use:
1. Lookd at the [`CHANGELOG`](https://github.com/facebook/react/blob/master/CHANGELOG.md) for release dates.
2. Cross-references [Netlify build history for the `master` branch](https://app.netlify.com/sites/reactjs/deploys?filter=master)
3. Picked the most recent build _before_ we updated the site (e.g. merged a PR that adds new API docs)

Unfortunately, our Netlify history begins with version 16 RC3 (so we can't link back to an old 15.6 release).

Since Netlify has built this PR preview as [5ad0def873f2cf6608e0d48a--reactjs.netlify.com](https://5ad0def873f2cf6608e0d48a--reactjs.netlify.com/), we should be able to test these new redirects as:
* 16.0: https://5ad0def873f2cf6608e0d48a--reactjs.netlify.com/version/16.0
* 16.1: https://5ad0def873f2cf6608e0d48a--reactjs.netlify.com/version/16.1
* 16.2: https://5ad0def873f2cf6608e0d48a--reactjs.netlify.com/version/16.2

A `robots.txt` file has also been added- along with `rel="nofollow"` attributes- to prevent search crawlers from indexing older versions of the site. (We want the latest release to be what shows up in search engines at all times.)

This versioning strategy should also work well by default with any [localization strategy](https://github.com/reactjs/reactjs.org/issues/82) we use.

**TODO**:
- [x] Add a UI mechanism for viewing old revisions
- [x] Don't let search engines crawl older versions
- [x] Make versions configurable via a single YAML file
- [ ] Determine if there's a way to use a subdomain of reactjs.org for versions rather than the netlify URL

Resolves issue #85

### Screenshot
screen shot 2018-04-14 at 9 53 34 am

### DIFF
--- a/src/pages/versions.js
+++ b/src/pages/versions.js
@@ -39,11 +39,13 @@ const Versions = () => (
                     Changelog
                   </a>
                 </li>
-                <li>
-                  <a href={version.path} rel="nofollow">
-                    Documentation
-                  </a>
-                </li>
+                {version.path && (
+                  <li>
+                    <a href={version.path} rel="nofollow">
+                      Documentation
+                    </a>
+                  </li>
+                )}
               </ul>
             </div>
           ))}
--- a/src/pages/versions.js
+++ b/src/pages/versions.js
@@ -1,0 +1,54 @@
+/**
+ * Copyright (c) 2013-present, Facebook, Inc.
+ *
+ * @emails react-core
+ * @flow
+ */
+
+import Container from 'components/Container';
+import Header from 'components/Header';
+import TitleAndMetaTags from 'components/TitleAndMetaTags';
+import React from 'react';
+import {sharedStyles} from 'theme';
+
+const Versions = () => (
+  <Container>
+    <div css={sharedStyles.articleLayout.container}>
+      <div css={sharedStyles.articleLayout.content}>
+        <Header>React Versions</Header>
+        <TitleAndMetaTags title="React - Versions" />
+        <div css={sharedStyles.markdown}>
+          <p>
+            A complete release history for React is available{' '}
+            <a
+              href="https://github.com/facebook/react/releases"
+              target="_blank"
+              rel="noopener">
+              in GitHub
+            </a>.
+          </p>
+          <p>Documentation for recent releases can also be accessed below:</p>
+          <ul>
+            <li>
+              <a href="/version/16.2" rel="nofollow">
+                16.2.0
+              </a>
+            </li>
+            <li>
+              <a href="/version/16.1" rel="nofollow">
+                16.1.1
+              </a>
+            </li>
+            <li>
+              <a href="/version/16.0" rel="nofollow">
+                16.0.0
+              </a>
+            </li>
+          </ul>
+        </div>
+      </div>
+    </div>
+  </Container>
+);
+
+export default Versions;
--- a/src/pages/versions.js
+++ b/src/pages/versions.js
@@ -11,6 +11,9 @@ import TitleAndMetaTags from 'components/TitleAndMetaTags';
 import React from 'react';
 import {sharedStyles} from 'theme';
 
+// $FlowFixMe This is a valid path
+import versions from '../../content/versions.yml';
+
 const Versions = () => (
   <Container>
     <div css={sharedStyles.articleLayout.container}>
@@ -29,21 +32,13 @@ const Versions = () => (
           </p>
           <p>Documentation for recent releases can also be accessed below:</p>
           <ul>
-            <li>
-              <a href="/version/16.2" rel="nofollow">
-                16.2.0
-              </a>
-            </li>
-            <li>
-              <a href="/version/16.1" rel="nofollow">
-                16.1.1
-              </a>
-            </li>
-            <li>
-              <a href="/version/16.0" rel="nofollow">
-                16.0.0
-              </a>
-            </li>
+            {versions.map(version => (
+              <li key={version.title}>
+                <a href={version.path} rel="nofollow">
+                  {version.title}
+                </a>
+              </li>
+            ))}
           </ul>
         </div>
       </div>
--- a/src/pages/versions.js
+++ b/src/pages/versions.js
@@ -28,18 +28,25 @@ const Versions = () => (
               target="_blank"
               rel="noopener">
               in GitHub
-            </a>.
+            </a>. Documentation for recent releases can also be found below:
           </p>
-          <p>Documentation for recent releases can also be accessed below:</p>
-          <ul>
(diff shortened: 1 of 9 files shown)


gaearon:
Will Netlify ever remove them?

consultwithmike:
This is awesome!

Looks good at first with this screenshot:

screen shot 2018-04-13 at 11 15 12 am

and remains good if I use the menus to navigate:

image

but if I use Search it will use the latest version:

image

bvaughn:
> Will Netlify ever remove them?

I don't think so. On their [features page](https://www.netlify.com/features/) they say each URL is permanent:
screen shot 2018-04-13 at 10 18 09 am

gaearon:
Well, as long as Netlify is in business 😛

alexkrolick:
Search always uses the newest version, even for deployed previews and localhost. Probably should be tackled separately.

consultwithmike:
@alexkrolick we can tackle it separately with a different issue if we want, and I fully suspected that there was a code change coming down the pipeline, I just wanted to drop that information on the issue so we can get it handled.

alexkrolick:
Issue for versioned search: https://github.com/reactjs/reactjs.org/issues/803

bvaughn:
I added a drop-down toggle button for version switching. I'd love to get feedback, but I think I'm going to immediately replace it with a "older versions" index page because:
* It would not work for mobile (since we don't even show the version number)
* This would not work well once we have more than a few versions

I just wanted to add it so it would be available in the history of this PR.

bvaughn:
Okay, I changed the version link in the header to link to a "Versions" page, which points to both the CHANGELOG _and_ docs for recent releases of the site.

alexkrolick:
Is there some way to source both the redirects and the page content from the same source? Like `versions.yml`:
```
versions:
	- title: 16.0
	  version: 16.0.0
	  url: 5ad0def873f2cf6608e0d48a--reactjs.netlify.com/version/16.0
	- title: 16.1
	  version: 16.1.1
	  url: 5ad0def873f2cf6608e0d48a--reactjs.netlify.com/version/16.1
```

bvaughn:
Yeah! Let me see if I can whip up another Gatsby startup plugin thingy for that.

bvaughn:
Ok. I added a Gatsby plugin that reads version values from a YAML file and adds them as Netlify redirect rules. Also updated the Versions page to read from this same YAML file. It's a good bit of extra code compared to the previous version- but it simplifies the process of adding new versions.

src/pages/versions.js
```diff
@@ -0,0 +1,49 @@
...
+              target="_blank"
+              rel="noopener">
+              in GitHub
+            </a>.
+          </p>
+          <p>Documentation for recent releases can also be accessed below:</p>
+          <ul>
+            {versions.map(version => (
+              <li key={version.title}>
+                <a href={version.path} rel="nofollow">
```
gaearon: Since the top link used to always lead to a release, what do you think about making releases first-class entities on this page?

Something like

```
# React 16.2.0

* Changelog
* Documentation

# React 16.1.0

* Changelog
* Documentation

```

bvaughn: I dig it.

bvaughn:
I think we should move forward with this and follow up with the vanity subdomain URLs (assuming Netlify supports this- I've reached out to them about it on Twitter). This is a useful feature for the community, even as-is (with ugly URLs). 😄

bvaughn:
cc @danoc in case you know of a way to do this already ^

bvaughn:
I wonder if a modified version of this plug-in could be useful for other Gatsby sites? Or maybe as an addition to the existing [`gatsby-plugin-netlify` plugin](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-plugin-netlify/)... cc @KyleAMathews

KyleAMathews:
Cool idea! The thought of duplicating documentation every time you do a new release has always filled me loathing (more-or-less) so this seems tons better. Relatively few people care about the older docs so pointing them to a snapshot seems like the least work for the most benefit.

It seems like this could be a lot simpler by fixing the redirect bug? Then you'd just need to read in the yaml file and create the redirects and the existing gatsby-plugin-netlify would setup the server redirects for you. I think all we'd need to do is to add a check [here](https://github.com/gatsbyjs/gatsby/blob/462e3b5b57c444e8d6acd3321422e29c06d69674/packages/gatsby/src/redux/actions.js#L967) if the to path is absolute and if so, don't add the pathPrefix.

danoc:
> cc @danoc in case you know of a way to do this already ^

AFAIK, the only way to do this right now is with the "branch subdomains" feature in Netlify. Would be great if there was a way to "tag" certain Netlify deploys and provide clean-lookin' URLs. 😄

bvaughn:
FYI there is a somewhat hacky option we could use to generate vanity subdomains (e.g. 16-0.reactjs.org) as per Netlify's tweet: https://twitter.com/NetlifySupport/status/985338587955810304

I'm not convinced this is worth it, TBH.

bvaughn:
> It seems like this could be a lot simpler by fixing the redirect bug? Then you'd just need to read in the yaml file and create the redirects and the existing gatsby-plugin-netlify would setup the server redirects for you.

Yes~ agreed! I assumed this bug was somewhat tricky to fix since it's been around for a while, but that would be an easier solution for sure. 😄

KyleAMathews:
Nah, more no one has cared yet :-)

bvaughn:
**Update on the vanity subdomains**: [According to Netlify support](https://twitter.com/NetlifySupport/status/985338587955810304), the only way to have a subdomain like 16-0.reactjs.org point to a cached snapshot of the site would be to create a Git branch from the commit that generated that build, and then setup a [branch subdomain](https://www.netlify.com/docs/custom-domains/#branch-subdomains). This approach would, I suppose, have the small added benefit of allowing us to update old branches if we ever needed to- but it also adds a bit of setup work each time we want to create a version.

I also tried another approach- using a [200 proxy redirect](https://www.netlify.com/docs/redirects/#proxying) to gives us URLs like reactjs.org/16.0. Unfortunately it didn't work. (The latest version of the site loaded, rather than the older one I was trying to proxy.)

gaearon:
Now I'm hungry 😢

PR #809, opened 2018-04-15
bind the tick method in counstructor - remove arrow function in setInterval

alirezavalizade:


alexkrolick:
This creates a new function, same as the arrow syntax, so it doesn't seem helpful to add the extra indirection with `.bind`. There's also some unrelated changes to Babel.

PR #810, opened 2018-04-16
Add missing single quotes to argument

Jemt:
First argument passed to React.createElement should be of type string but single quotes were omitted


gaearon:
It’s only a string for platform provided element types like `'div'`. For custom components it’s a function or a class that’s available in the scope.

PR #811, opened 2018-04-16
React 16.3.2

gaearon:


bvaughn:
I've updated the `versions.yml` file in #801 to point to this latest changelog as well.

PR #812, opened 2018-04-16
"Introducing JSX" is missing a step

stmoreau:
See --> https://github.com/reactjs/reactjs.org/issues/794
Hope that helps :bowtie:


gaearon:
I tweaked a bit. Thanks.

PR #813, opened 2018-04-16
docs(refs-and-the-dom): typo in CustomTextInput

MarcBouchard:
Correct a typo in the render method comments for the "CustomTextInput" component of the "Adding a Ref to a DOM Element" example.  The first line of the comments reads "tell React that we want the associate the <input> ref".  The word "the" should be "to" resulting in the comment "tell React that we want to associate the <input> ref".


gaearon:
Thanks

PR #815, opened 2018-04-17
docs(error-boundaries):remove react beta reference

johneas10:
Noticed a reference to React Beta while going through the Error Boundaries documentation.
I believe it should point to the React 16 release notes instead.


gaearon:
Thanks

PR #837, opened 2018-04-24
Linked to an example "from scratch" implementation

paradoxinversion:
@gaearon contacted me about adding a link my article on implementing react from scratch.  Let me know if anything needs to be changed.


gaearon:
@paradoxinversion Can you please amend the article to include major versions explicitly when installing? This way your guide won't break when there's a new version of Babel etc.

paradoxinversion:
@gaearon totally, I'll be able to make those edits within couple of hours. I also don't have any problems with keeping the article up-to-date in the future.

alexkrolick:
ok great!

paradoxinversion:
The article now targets specific versions when instructing folks to download packages: ie.` npm install --save-dev webpack@4.6.0 webpack-cli@2.0.14 ... `, and I've noted my node version in the first paragraph.

Issue #840, opened 2018-04-25
Clarify portals docs 


gaearon:
https://reactjs.org/docs/portals.html

We should make it clearer in the first paragraph it’s not imperative and doesn’t “work” outside of render. There is a common confusion that this should be called in lifecycles (where it has no effect).

danburzo:
_(Moved from #911)_ — I think we should also include a mention of whether the Portal will take control of the DOM element it receives as `container` (and, consequently, it needs its own DOM element to render into), similarly to how it's noted  for `ReactDOM.render()`. Clarifying this this can help people avoid creating useless DOM nodes "just in case" — that is, if there aren't any caveats to using, for example, `document.body` as a target.

Another thing I was momentarily surprised by — although it makes perfect sense — is that if you rely on CSS classes to style an element that you then transport to a different part of the the DOM, it will not inherit styles from its parent React element. Maybe warrants a tiny note?

rickhanlonii:
@gaearon can we close this? I think you explained this well in the new docs https://react.dev/reference/react-dom/createPortal

PR #845, opened 2018-04-26
Clarify createPortal usage

raunofreiberg:
#840. I think this clarification makes more sense as a note.

### DIFF
--- a/content/docs/portals.md
+++ b/content/docs/portals.md
@@ -14,7 +14,7 @@ The first argument (`child`) is any [renderable React child](/docs/react-compone
 
 > Note:
 >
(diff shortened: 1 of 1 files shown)


gaearon:
I don't think this is technically accurate. You could call it in a lifecycle, then put the result into state and then use that in render. Not necessarily useful but possible.

The important part is that it doesn't do anything *imperatively* and it just produces an object similar to a React element. The portal only "works" when this object is included in the rendered tree.

raunofreiberg:
What about something like this? I don't really like the word "work/working" and tried to avoid it.

rickhanlonii:
Sorry for the long delay here. Since this PR was created against the old site I'm going to close this, but I think the new docs explain this pretty well: https://react.dev/reference/react-dom/createPortal

PR #846, opened 2018-04-26
Add accessibility guidelines link to portals doc

mvargeson:


gaearon:
We probably want to be more specific there. cc @AlmeroSteyn

AlmeroSteyn:
@mvargeson Great that you thought about linking through to the a11y doc!

As managing keyboard focus becomes very important with the functionalities created with portals, we should probably mention that shortly and change the link to drill down into that section of the a11y doc https://reactjs.org/docs/accessibility.html#programmatically-managing-focus

This sections contains reference to the `react-aria-modal` library which illustrates perfectly how focus should be managed in this case, and it uses the `react-displace` library (https://github.com/davidtheclark/react-displace) which makes use of portals.

Perhaps also a good idea to add a link to the WAI-ARIA Authoring Practices for a dialog modal (https://www.w3.org/TR/wai-aria-practices-1.1/#dialog_modal).

What do you think?

AlmeroSteyn:
@mvargeson could you still make the changes to this PR? This is really great to add as modals are causing a lot of accessibility issues so would be awesome to merge.

mvargeson:
@AlmeroSteyn This completely fell off my mind. Thanks for the bump. I'll be able to make these changes this weekend.

AlmeroSteyn:
Ah no worries, I know the feeling. Thank you for picking this up again!

content/docs/portals.md
```diff
@@ -44,7 +44,9 @@ A typical use case for portals is when a parent component has an `overflow: hidd
 
 > Note:
 >
-> It is important to remember, when working with portals, you'll need to make sure to follow the proper [accessibility guidelines](/docs/accessibility.html).
+> It is important to remember, when working with portals, that managing keyboard focus becomes very important. You'll need to make sure to follow the proper [accessibility guidelines](/docs/accessibility.html#programmatically-managing-focus).
```
AlmeroSteyn: First sentence has "important" twice. Second sentence feels a little bulky. Can you combine it? 

Something like 
```
When working with portals, remember that [managing keyboard focus](/docs/accessibility.html#programmatically-managing-focus) becomes very important.
```

content/docs/portals.md
```diff
@@ -44,7 +44,9 @@ A typical use case for portals is when a parent component has an `overflow: hidd
...
+>
+> You'll also want to read through the recommendations described in the [WAI-ARIA Authoring Practices](https://www.w3.org/TR/wai-aria-practices-1.1/#dialog_modal) for a dialog model.
```
AlmeroSteyn: The sentence creates the idea that this pattern needs to be applied to all uses of portals. Indicate that this resource is specifically useful for modals.

Something like:

```
For modal dialogs, ensure that everyone can interact with them by following the [WAI-ARIA Modal Authoring Practices](https://www.w3.org/TR/wai-aria-practices-1.1/#dialog_modal).
```

mvargeson:
Thank you for the suggestions @AlmeroSteyn! This is ready for re-review at your convenience.

alexkrolick:
Squash and merge should be fine

AlmeroSteyn:
Learnt something new :-)

AlmeroSteyn:
Thanks again @mvargeson for coming with the idea and your patience :-)

PR #847, opened 2018-04-26
Mentioning forceUpdate in the Updating part of the component lifecycle

manuelbieh:
I'm not sure if it was left out intentionally because it's considered bad practice but or the sake of completeness I added a third option that causes a component to rerender: explicitly calling `forceUpdate()` on that component.


threepointone:
Yeah, I think it wasn't explicitly called out. The page includes docs for forceUpdate later, so I think it's fine. Thank you for your PR!

PR #863, opened 2018-05-06
Rename "functional component" to "function component"

gaearon:
Related to discussion in https://mobile.twitter.com/getify/status/993127777057550336

Yay or nay?

### DIFF
--- a/content/blog/2015-10-07-react-v0.14.md
+++ b/content/blog/2015-10-07-react-v0.14.md
@@ -104,12 +104,12 @@ If you can’t use `npm` yet, we provide pre-built browser builds for your conve
 
     With this change, we’re deprecating `.getDOMNode()` and replacing it with `ReactDOM.findDOMNode` (see below). If your components are currently using `.getDOMNode()`, they will continue to work with a warning until 0.15.
 
-- #### Stateless functional components
+- #### Stateless function components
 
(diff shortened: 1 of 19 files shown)


alexkrolick:
"Function component" sounds wrong grammatically. What about "component function"?

gaearon:
We do say “class component” so I don’t think it’s too different. Just a bit unusual.

gaearon:
Like “strawberry ice-cream”.

alexkrolick:
Ok, as long as you stick to the ice cream analogy 😉

sebmarkbage:
How are they not functional? They have to be pure. Are supposed to be used with lazy evaluation in a monadic way.

Even with state they are pure because the state will be based on algebraic effects.

If anything not calling them functional is misleading since they follow a strict set of rules that all come from functional programming. Everything is functions, including classes.

gaearon:
>Everything is functions, including classes.

That's the point, isn't it?

It's not that function components aren't functional-by-nature. It's that they're indeed *all* functional, including classes. Therefore saying "functional" to mean "a component that happens to be a function, as opposed to a class" doesn't mean as much sense. But that's exactly how we use the term right now, and what this PR could change.

This doesn't mean either of them stop being functional. Just that we can might prefer to use the word "functional" to mean the paradigm and overall approach, rather than the component definition syntax.

Tbh I can argue both ways for this.

gaearon:
(Chatted with @sebmarkbage and he's not opposed)

PR #865, opened 2018-05-07
add opening brackets

zuhrinurhuda:

### DIFF
--- a/content/docs/lists-and-keys.md
+++ b/content/docs/lists-and-keys.md
@@ -268,7 +268,7 @@ In the examples above we declared a separate `listItems` variable and included i
 ```js{3-6}
 function NumberList(props) {
   const numbers = props.numbers;
-  const listItems = numbers.map((number) =>
+  const listItems = numbers.map((number) => (
     <ListItem key={number.toString()}
               value={number} />
   );


content/docs/lists-and-keys.md
```diff
@@ -268,7 +268,7 @@ In the examples above we declared a separate `listItems` variable and included i
 ```js{3-6}
 function NumberList(props) {
   const numbers = props.numbers;
-  const listItems = numbers.map((number) =>
+  const listItems = numbers.map((number) => (
```
bvaughn: This is valid:
```js
const numbers = [1, 2, 3, 4, 5];
const listItems = numbers.map((number) =>
  <li>{number}</li>
);
```
The parens are balanced.

The proposed change would be invalid though.

Maybe it helps if you add newlines:
```js
const numbers = [1, 2, 3, 4, 5];
const listItems = numbers.map(
  (number) => ( // New opening paren...
    <li>{number}</li>
  // But no closing paren
);
```

bvaughn:
Thanks for the PR! Looks like it may be a mistake though. 😄

content/docs/lists-and-keys.md
zuhrinurhuda: Thanks for the explanation @bvaughn , sorry for my mistake..

bvaughn: No worries!

PR #866, opened 2018-05-07
Clarify that synthetic events aren’t exactly the same as native events

Emuentes:
…s, and do not simply map to, native browser events.

This should address issue #68:
https://github.com/reactjs/reactjs.org/issues/68


content/docs/reference-events.md
```diff
@@ -10,9 +10,9 @@ This reference guide documents the `SyntheticEvent` wrapper that forms part of R
 
 ## Overview
 
-Your event handlers will be passed instances of `SyntheticEvent`, a cross-browser wrapper around the browser's native event. It has the same interface as the browser's native event, including `stopPropagation()` and `preventDefault()`, except the events work identically across all browsers.
+Your event handlers will be passed instances of `SyntheticEvent`, a cross-browser wrapper around the browser's native event. It has the same interface as the browser's native event, including `stopPropagation()` and `preventDefault()`, except the events work identically across all browsers. 
 
-If you find that you need the underlying browser event for some reason, simply use the `nativeEvent` attribute to get it. Every `SyntheticEvent` object has the following attributes:
+The synthetic events are different from, and do not map directly to, the browser's native events. If you find that you need the underlying browser event for some reason, simply use the `nativeEvent` attribute to get it. Every `SyntheticEvent` object has the following attributes:
```
eps1lon, line 15: An example would help clarify what this means I guess. I think `mouseenter` is a good example where `event.nativeEvent` in `onMouseEnter` points to a `mouseout` event. 

```suggestion
If you find that you need the underlying browser event for some reason, simply use the `nativeEvent` attribute to get it. The synthetic events are different from, and do not map directly to, the browser's native events. For example in `onMouseLeave` `event.nativeEvent` will point to a `mouseout` event. The specific mapping is not part of the public API and may change at any time. Every `SyntheticEvent` object has the following attributes:
```

But we should make it clear that this mapping isn't part of react's public API. The specific mapping is an implementation detail. 

/cc @trueadm

eps1lon:
@Emuentes Thanks!

PR #870, opened 2018-05-07
Remove deprecated lifecycle method from docs

bryzaguy:


eps1lon:
Thanks for your work. This section was already updated with a similar patch in #2435.

PR #876, opened 2018-05-12
Fix grammar mistake in rendering-elements.md

chaficnajjar:
"contents" is plural so it should be followed by "have" instead of "has".


threepointone:
This was fixed elsewhere. Thanks!

PR #878, opened 2018-05-14
Update tools-starter-kits.md

jonidelv:
Add [generator-create-redux-app](https://github.com/jonidelv/generator-create-redux-app) into the starter kits list.


jonidelv:
Is this ready to be merged?

PR #879, opened 2018-05-14
Update faq-build.md

xliank:
avoid using comment in front of component

### DIFF
--- a/content/docs/faq-build.md
+++ b/content/docs/faq-build.md
@@ -33,13 +33,15 @@ No! Check out ["React Without ES6"](/docs/react-without-es6.html) to learn more.
 
 
 ```jsx
+{/* Avoid this. */} 
 <div>
-  {/* Avoid this. */} <span>It won't works</span>
+  <span>It won't works</span>
 </div>
 ```
 
 ```jsx
+{/* Avoid this in React Native. */} 
 <View>
-  {/* Avoid this in React Native. */} <Text>It won't works</Text>
+ <Text>It won't works</Text>
 </View>
 ```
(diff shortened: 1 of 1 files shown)


content/docs/faq-build.md
```diff
@@ -30,3 +30,16 @@ No! Check out ["React Without ES6"](/docs/react-without-es6.html) to learn more.
   Hello, {name}! 
 </div>
 ```
+
+
+```jsx
+<div>
+  {/* Avoid this. */} <span>It won't works</span>
```
alexkrolick: It does work - this is the same as the previous example. https://codepen.io/alexkrolick/pen/YLOxrX

Perhaps you meant:

```jsx
return `{/* Avoid this. */} <span>It won't works</span>`
```

which fails because the return value is not wrapped in a single tag. The same problem would occur if the comment node was instead a tag (div, span, etc.). The solution is to wrap in a React.Fragment.

xliank: Yeah, i'm sure you're right.
i'm sure somebody will use comment like this : 

````
return (
    {/* <div class='oldClass'> */ } 
    <div class='newClass'>
      <span>It won't work</span>
    </div>
)
````
or 


````
return (
    {/* <View className='oldClass'> */ } 
    <View className='newClass'>
      <Text>It won't work</Text>
    </View >
)
````
I'm not sure why eslint & metro builder in React Native didn't notice it.
It's just show errror without any hint which file : 
````
Cannot add a child that doesn't have a 
YogaNode to a parent without a measure function! 
````


nb: sorry for my very bad English, it's my fourth language

PR #886, opened 2018-05-16
React.Children.map typo in function signature reference

jonathanazulay:
I was learning how to use React.Children.map and the point about `thisArg` confused me.

`React.Children.map(children, function[(thisArg)])`

By testing I figured out that thisArg is a third optional argument that you can pass in to set this. So shouldn't that line look more like this:

`React.Children.map(children, fn[, thisArg])`

### DIFF
--- a/content/docs/reference-react.md
+++ b/content/docs/reference-react.md
@@ -159,7 +159,7 @@ Verifies the object is a React element. Returns `true` or `false`.
 #### `React.Children.map`
 
 ```javascript
-React.Children.map(children, function[(thisArg)])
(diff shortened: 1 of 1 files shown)


BartoszKlonowski:
Hello @jonathanazulay!
Yes, you are right about the `thisArg` being an optional parameter 👍
However, the page you refer to in this PR has already been moved and improved with new syntax of [`Children.map(children, fn, thisArg?)`](https://react.dev/reference/react/Children#children-map) method.
Let me close this PR as it's pretty much outdated now, but please feel invited to create more PRs for other open issues.
Thank you!

PR #888, opened 2018-05-17

robphoenix:
Use more accessible language in describing how `setState()` places component updates into a queue.


robphoenix:
Thanks for the feedback @alexkrolick 
Yeah it would be good not to (wrongly) imply what is actually going on under the hood.

Perhaps...

`setState()` queues up changes to the component state and tells React that this component...

`setState()` sets up changes to the component state and tells React that this component...

`setState()` lines up changes to the component state and tells React that this component...

I did think of *prepares* as well but that implies some sense of doing something to the changes, getting them ready, rather than just lining them up for rendering. I could easily overthink this 😆 What do you think?

PR #892, opened 2018-05-18
difference in document vs code in code pen

vijayj:
Changing this from    const history = this.state.history.slice(0, this.state.stepNumber + 1); to     const history = this.state.history; 
As slicing the history leads to losing the list elements and inability to go back to current state


BartoszKlonowski:
Hello @vijayj!
I don't see this mismatch anymore in the latest version of tutorial in the docs, so let me close this PR as outdated.
If you would still like to contribute, please create new PR for other open issues.

PR #893, opened 2018-05-18
Switch key to use index to ensure uniqueness

william-reed:
Not necessary in this case but if the list were not unique, using the value in the list as the key would present some problems

### DIFF
--- a/content/docs/lists-and-keys.md
+++ b/content/docs/lists-and-keys.md
@@ -102,8 +102,8 @@ Keys help React identify which items have changed, are added, or are removed. Ke
 
 ```js{3}
 const numbers = [1, 2, 3, 4, 5];
-const listItems = numbers.map((number) =>
-  <li key={number.toString()}>
+const listItems = numbers.map((number, index) =>
+  <li key={index}>
     {number}
   </li>
 );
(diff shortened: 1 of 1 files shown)


EarthSchlange:
I don't think this change is really necessary. As the use of an index is not generally recommended. Further the use of 'index' is explained between lines 112-133 on content/docs/lists-and-keys.md:

> The best way to pick a key is to use a string that uniquely identifies a list item among its siblings. Most often you would use IDs from your data as keys:
>  
>  ```js{2}
>  const todoItems = todos.map((todo) =>
>    <li key={todo.id}>
>      {todo.text}
>    </li>
>  );
>  ```
>  
>  When you don't have stable IDs for rendered items, you may use the item index as a key as a last resort:
>  
>  ```js{2,3}
>  const todoItems = todos.map((todo, index) =>
>    // Only do this if items have no stable IDs
>    <li key={index}>
>      {todo.text}
>    </li>
>  );
>  ```
>  
>  We don't recommend using indexes for keys if the order of items may change. This can negatively impact performance and may cause issues with component state. Check out Robin Pokorny's article for an [in-depth explanation on the negative impacts of using an index as a key](https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318). If you choose not to assign an explicit key to list items then React will default to using indexes as keys.
>  
>  Here is an [in-depth explanation about why keys are necessary](/docs/reconciliation.html#recursing-on-children) if you're interested in learning more.
>

threepointone:
Yeah, this example specifically uses those keys with the assumption that they're unique. It's way more dangerous to use indices, since they don't handle reorodering. Thanks for the PR though!

PR #894, opened 2018-05-19
Add a sentence to the end of componentWillUnmount()

barnonahill:
Adds a sentence explaining not to use setState in componentWillUnmount, see https://github.com/facebook/react/issues/12111


gaearon:
Thx

Issue #895, opened 2018-05-21
Clarify Context API


danburzo:
Hi there,

I'm trying to get up to speed with the Context API and something doesn't quite click with me. This issue outlines my experience with the [docs page](https://reactjs.org/docs/context.html). 

Mainly, I don't understand to what extent this API introduces new rendering paths / optimizations to how components usually work.

What kind of React components are _Provider_ and _Consumer_? 

Is _Provider_ a normal component, in that, when included in a larger component's render() method, it will re-render its whole subtree whenever the parent's render() method is invoked? In that sense, is it clearer to say that the _Consumer_ "subscribes" to changes in the Provider's value just by virtue of it re-rendering through the normal React mechanism? Or is it more like a PureComponent that only re-renders when its props (in this case, the `value` prop) change?

I, for one, would welcome more details into how the communication between a Provider and a Consumer works. The [Caveats](https://reactjs.org/docs/context.html#caveats) section does not explain what kind of optimization is enabled when lifting the state up.

Is it that the context propagation breaks through any `shouldComponentUpdate` "shields" set up in parent components, so that Consumers can update?

Sorry if this is a bit rambling, I'm just trying to get a firm grasp of how it works, so I can use it at its full potential... 😬

P.S. Performance / re-rendering concerns aside, I found it intuitive enough to write components with the Context API; what I'd be interested in is _reasoning about how the data flows_.

gaearon:
Both Provider and Consumer are special kinds of components. They’re implemented by React itself and can’t be written by a user. So conceptually they’re on the same level as “functional component” or “class component”, another separate code path that behaves differently.

If you re-render the Provider, it will indeed follow the normal React mechanism. But that is not sufficient for context propagation because Consumers might be inside of trees that bailed out due to shouldComponenUpdate. So context should propagate down even through the trees that “refused” to render.

>Is it that the context propagation breaks through any shouldComponentUpdate "shields" set up in parent components, so that Consumers can update?

Yes. And it’s not using subscriptions (they take time and memory to set up, but we want context to have no extra cost for the initial render). Instead, whenever we see a context Provider whose value has changed, we quickly scan the tree below, and mark matching consumers as needing an update (just like as if they received setState).

Hope that helps. Feel free to send a PR with this description if you found it useful.

danburzo:
Thank you, @gaearon, that clarifies things for me. I'm going to spend some more time with the API to grasp its nuances, and will come back with a PR once I'm sure I got everything correctly.

PR #896, opened 2018-05-21
Clarify change propagation in Context API

danburzo:
In regards to #895, I think a cursory mention of how Providers communicate with Consumers in the context of `shouldComponentUpdate` returning `false` goes a long way in creating the appropriate mental model for this feature. May not be the best phrasing, but I'd appreciate something along these lines for 3hrs-ago me :)


gaearon:
I like this. Thanks.

danburzo:
Great, my pleasure!

PR #901, opened 2018-05-23
update to add-react-to-a-new-app

dbouvie:
the sequence of commands for install using npm is a bit confusing because it looks like (because of the blank line) the npm install and the create-react-app is one command with line wrapping. Adding a blank line makes it clear these are two commands.


saranshkataria:
closing this out since this has been moved to a different section now. Thanks

PR #904, opened 2018-05-23
16.4 blog post

acdlite:
TODO:
- [ ] ~Section about `React.unstable_Profiler` (@bvaughn)~ Never mind.
- [ ] ~Explain why async rendering isn't in this release (?)~ Nah, don't need this either.
- [x] Update `getDerivedStateFromProps` docs to reflect new behavior


content/blog/2018-03-29-react-v-16-4.md
```diff
@@ -0,0 +1,128 @@
+---
```
gaearon: Don't forget to change the date in the filename

content/blog/2018-05-23-react-v-16-4.md
```diff
@@ -0,0 +1,128 @@
...
+```
+
+You can change it to do this instead:
+
+```js
+static getDerivedStateFromProps(props, state) {
+  const prevProps = state.prevProps;
+  // Compare incoming props to previous props
+  const controlledValue =
+    prevProps.value !== state.controlledValue
```
bvaughn: This doesn't look quite right. Wouldn't this block every `setState` update?

content/blog/2018-05-23-react-v-16-4.md
```diff
@@ -0,0 +1,128 @@
...
+      ? prevProps.value
+      : state.controlledValue;
+  return {
+    // Store the previous props in state
+    prevProps: props,
+    controlledValue,
+  };
```
bvaughn: I think we should avoid this example. Even aside from gDSFP, this approach of half-controlled/half-uncontrolled leads to an awkward/ambiguous API.

For example, the only way to reset a half-controlled component's state to a value safely is to render twice, temporarily setting the prop to something else (e.g. null) and then setting it back.

I think instead, we should encourage fully controlled (no `state`) or fully uncontrolled (`props.defaultValue` sets initial `state` and never gets synced).

Maybe put another way, I think we may want to avoid a component ever updating a `state` attribute via `setState` that `props` also updates.

bvaughn: Maybe it's okay we show this, since we want to cover a before and after case. I don't know. In general, I'm just concerned about any perceived encouragement of this API/design. 😄

content/blog/2018-05-23-react-v-16-4.md
```diff
@@ -0,0 +1,128 @@
...
+
+Side effects in `getDerivedStateFromProps` were already disallowed, but since it now fires more often than it used to, the recent change may expose previously undiscovered bugs.
+
+### Compare Incoming Props to Previous Props When Computing Controlled Values
+
+The following code will not work. It assumes `getDerivedStateFromProps` only fires on prop changes. But since it now fires for state changes, too, local updates to the controlled value will be ignored, since the props version always overrides it.
+
+```js
+static getDerivedStateFromProps(props, state) {
+  if (props.value !== state.controlledValue) {
```
bvaughn: It might be worth mentioning that this pattern was flawed to begin with- even setting the recent gDSFP change aside- because it would override `state` even when unrelated `props` changed.

bvaughn: I don't know if that would be more or less confusing. It might distract from the thing you're trying to emphasize. I don't know.

content/blog/2018-05-23-react-v-16-4.md
```diff
@@ -0,0 +1,120 @@
...
+    prevProps.value !== props.value
+      ? props.value
+      : state.controlledValue;
+  return {
+    // Store the previous props in state
+    prevProps: props,
+    controlledValue,
+  };
+}
+```
```
gaearon: Let's just at least mention that most components **don't** need this pattern. We're only documenting for the cases you're already using it for. But if you just need some value from props, you should just read it from props, and not attempt to put it into state.

gaearon: For example:

>**Note that typically the vast majority of the components don't need `getDerivedStateFromProps` at all.** It shouldn't be used often, as making state dependent on props often leads to logic that is hard to understand. If possible, prefer to have one source of truth for any value (either in props or in state), and [lift state up](https://reactjs.org/docs/lifting-state-up.html) if it's necessary to share it between multiple components.

acdlite: Ok, but I have mixed feelings of telling people not to use a feature, especially one we just added, unless we give them a clear alternatives. We need to follow up with a more comprehensive post soon (like this week or next).

content/blog/2018-05-23-react-v-16-4.md
```diff
@@ -0,0 +1,147 @@
...
+
+### 2. Compare Incoming Props to Previous Props When Computing Controlled Values
+
+The following code assumes `getDerivedStateFromProps` only fires on prop changes:
+
+```js
+static getDerivedStateFromProps(props, state) {
+  if (props.value !== state.controlledValue) {
+    return {
+      // Since this method fires on both props and state changes, local updates
```
bvaughn: Super tiny nit. Please feel free to ignore.

Suggestion: Change "_Since this method fires_" to "_Because this method fires_"

PR #912, opened 2018-05-25
Clarify class lifetime in class conversion example

sophiebits:


gaearon:
Edited a bit so it doesn't seem like it's a singleton. If you disagree let's find a better wording next week :-)

PR #913, opened 2018-05-26
move code to example: composition vs inheritance

cyan33:
As is mentioned in https://github.com/reactjs/reactjs.org/issues/246, I'm trying to move the code from `docs/composition-vs-inheritance` to `/examples`. Just want to send a small PR to see if I'm doing it the right way so that I could move on for the rest work.

There are several things that I noticed though:
1. I tried to convert codepen link to a codesandbox one, but the thing is: there is no global variable name `React` and `ReactDOM` in codesandbox. So I had to add it manually through `import` but it clearly adds noise to the previous compact example. Is there a quick fix about this? (I'm not quite familiar with codesandbox)
2. What if the codepen example contains css styles? How can I combine all of those to a js file? Or should I come up a new code example maybe?

cc @washingtonsoares @bvaughn

### DIFF
--- a/content/docs/composition-vs-inheritance.md
+++ b/content/docs/composition-vs-inheritance.md
@@ -17,67 +17,21 @@ Some components don't know their children ahead of time. This is especially comm
 
 We recommend that such components use the special `children` prop to pass children elements directly into their output:
 
-```js{4}
-function FancyBorder(props) {
-  return (
-    <div className={'FancyBorder FancyBorder-' + props.color}>
-      {props.children}
-    </div>
-  );
-}
-```
+`embed:composition-vs-inheritance/fancy-border-example.js`
 
 This lets other components pass arbitrary children to them by nesting the JSX:
 
-```js{4-9}
-function WelcomeDialog() {
-  return (
-    <FancyBorder color="blue">
-      <h1 className="Dialog-title">
-        Welcome
-      </h1>
-      <p className="Dialog-message">
-        Thank you for visiting our spacecraft!
-      </p>
-    </FancyBorder>
-  );
-}
-```
+`embed:composition-vs-inheritance/welcome-dialog-example.js`
 
-[Try it on CodePen.](https://codepen.io/gaearon/pen/ozqNOV?editors=0010)
+[Try it on CodeSandbox.](codesandbox://composition-vs-inheritance/fancyborder-welcome-dialog-example)
 
 Anything inside the `<FancyBorder>` JSX tag gets passed into the `FancyBorder` component as a `children` prop. Since `FancyBorder` renders `{props.children}` inside a `<div>`, the passed elements appear in the final output.
 
 While this is less common, sometimes you might need multiple "holes" in a component. In such cases you may come up with your own convention instead of using `children`:
 
-```js{5,8,18,21}
-function SplitPane(props) {
-  return (
-    <div className="SplitPane">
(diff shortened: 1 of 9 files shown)


content/docs/composition-vs-inheritance.md
```diff
@@ -17,65 +17,19 @@ Some components don't know their children ahead of time. This is especially comm
...
-      }
-      right={
-        <Chat />
-      } />
-  );
-}
-```
+`embed:composition-vs-inheritance/split-pane-example.js`
 
 [Try it on CodePen.](https://codepen.io/gaearon/pen/gwZOJp?editors=0010)
```
washingtonsoares: Hello!

It's the first time I'm reading the code and I'm wondering.
Why don't we move this example to CodeSandbox?

cyan33: Yeah my intention was to move it to CodeSandbox. But in this codepen example, we have both css and javascript and I'm not sure about how to do that within a single javascript file in the `example` folder. Any thoughts?

washingtonsoares: One question: Is it possible to reference multiple files through a  `codesandbox: //` link? If it is possible, I see that it is possible to solve the problem if we create a  `style.css` file and reference it in ` index.js`.

washingtonsoares: Another idea would be to create `css-in-js`, but I'm afraid to cause some unnecessary 'complexity' in the documentation.

What you think about ?

cyan33: > Is it possible to reference multiple files through a codesandbox: // link?

I've tried that but it didn't work. Is it a gatsby plugin that we are using? Maybe should see how that works.

> Another idea would be to create css-in-js...

For the second approach, I think it could add much extraneous noise to the example snippets, which is meant to be clear and compact.

cyan33:
@bvaughn That sounds interesting. Could you show me the link to the gatsby plugin you wrote?

bvaughn:
I'm thinking you could pass it a comma-separated list of files. This would be a backwards compatible change too.

So a single file would continue to be:
```js
[codepen://components-and-props/rendering-a-component.js]
```

And multiple files would become:
```js
[codepen://components-and-props/rendering-a-component.js,components-and-props/rendering-a-component.css]
```

washingtonsoares:
Hello guys!

All the pull requests that refer to this pull request have already been merged.
I would be very happy if we could finalize this too! :smiley: 

I can help by resolving conflicts and reviewing the code.

What do you think ? @cyan33 @bvaughn

washingtonsoares:
@bvaughn @cyan33  Do you think it would be a good idea to continue this work? I think it will not be so difficult, since the PRs related in Gatsby have already been merged.

cyan33:
@washingtonsoares Hey Washington, hope you've doing well and thanks for following up on this pull request. I'll take a look after work!

BartoszKlonowski:
Hello @cyan33!
I was walking through that part of docs and saw that the `composition-vs-inheritance` does no longer exists in the docs and I was unable to find any exact replacement. We already have a bunch of good examples in [Passing Props to a Component](https://react.dev/learn/passing-props-to-a-component#passing-jsx-as-children) page, so I don't think this PR you created is still required.
Let me close this as outdated, but thank you very much for taking the time to contribute to the docs! 👍

PR #923, opened 2018-06-01
[A11Y-docs]: Add section on pointer events with outside click example

### LINKED ISSUE #849: Document "outer click" alternatives in Accessibility section
This comes up pretty often: https://mobile.twitter.com/ryanflorence/status/989940719451127808

Not to steal @ryanflorence’s thunder but I think we should have something on the Accessibility page that helps explain how to deal with this scenario.

cc @AlmeroSteyn

AlmeroSteyn:
Closes #849 

I have added a section on **Pointer and mouse events** to the accessibility doc. 

I though this a good place to highlight the issue with outside click and to use that as a warning not to depend on mouse events alone.

Here is a summary screenshot of the full section:

image

I used animated gifs to illustrate the problem as I usually find it harder to convince developers with these kind of matters without a visual representation.

The code borrows heavily from @ryanflorence 's example linked in the related issue.

Can someone please review the changes?

cc @gaearon


content/docs/accessibility.md
```diff
@@ -217,6 +217,120 @@ initially triggered the modal.
...
+keyboard users cannot use your application.
+
+To illustrate this, let's look at a prolific example of broken accessibility caused by click events. This is the outside click pattern, where a user can disable an opened popover by clicking outside the element. 
+
+<img src="../images/docs/outerclick-with-mouse.gif" alt="A toggle button opening a popover list implemented with the click outside pattern and operated with a mouse showing that the close action works." />
+
+This is typically implemented by attaching a `click` event to the `window` object that closes the popover:
+
+```javascript{12-16,27}
+class OuterClickExample extends Component {
```
danburzo: I think most examples use `React.Component` instead of `Component`, for clarity.

content/docs/accessibility.md
```diff
@@ -217,6 +217,120 @@ initially triggered the modal.
...
+  constructor(props) {
+    super(props);
+
+    this.state = { isOpen: false };
+    this.toggleContainer = createRef();
```
danburzo: Same as with `React.Component`, this should be `React.createRef()`.

content/docs/accessibility.md
```diff
@@ -217,6 +217,120 @@ initially triggered the modal.
...
+
+    this.onClickHandler = this.onClickHandler.bind(this);
+  }
+
+  componentDidMount() {
+    window.addEventListener('click', event => {
```
danburzo: Adding an anonymous function as an event listener, and not removing it on `componentWillUnmount`, will cause it to be added over and over again to `window`. It should be a method, much like `onClickHandler`. As a bonus, it's clearer that it's an `onClickOutsideHandler`.

content/docs/accessibility.md
```diff
@@ -217,6 +217,120 @@ initially triggered the modal.
...
+      if (this.state.isOpen && !this.toggleContainer.current.contains(event.target)) {
+        this.setState({ isOpen: false });
+      }
+    });
+  }
+
+  onClickHandler() {
+    this.setState({
```
danburzo: Since it depends on the current state, a better idea is to use the:
```js
this.setState(current_state => {
  isOpen: !current_state.isOpen
})
```
style of `setState` invocation.

content/docs/accessibility.md
AlmeroSteyn: Changed.

content/docs/accessibility.md
AlmeroSteyn: Changed.

content/docs/accessibility.md
AlmeroSteyn: Agreed and changed.

content/docs/accessibility.md
AlmeroSteyn: Changed.

content/docs/accessibility.md
```diff
@@ -217,6 +217,120 @@ initially triggered the modal.
...
+      isOpen: false
+    });
+  }
+
+  render() {
+    return (
+      <div className="menu-wrapper">
+        <button
+          onClick={this.onClickHandler}
+          onBlur={this.onBlurHandler}
```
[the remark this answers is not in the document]
AlmeroSteyn: Yes I was facepalming myself so hard. Was focussing on the text more than bulding out the example. It should indeed be on the parent.

AlmeroSteyn:
@danburzo: Thank you very much for taking the time to review. 

I have built out the example and checked the keyboard interaction. However, this requires a small hack with setTimeout as discussed in [this article on handling focus and blur](https://medium.com/@jessebeach/dealing-with-focus-and-blur-in-a-composite-widget-in-react-90d3c3b49a9b) to track whether focus leaves the parent container.

It should be working now, but I am unable to test the Firefox/Mac combination, would you be able to double check that? It works in Firefox on Windows for me.

danburzo:
@AlmeroSteyn It still does not work in FF/Mac, unfortunately. I'll be back with a more useful diagnosis once I can carve some time to play with your example.

AlmeroSteyn:
@danburzo Thank you so much. Would do it myself but I just don't have access to that setup at the present time to test. I really appreciate it.

Just to be sure I copied the example from the rendered docs and used that code directly and tested in IE, Edge, Chrome and Firefox on Windows and they all work.

I am very curious why it wont work on Mac. I see there has been issues with `blur` events on Mac/Firefox in the past but that was a long time ago.

AlmeroSteyn:
I was just typing about React bubbling `blur` and `focus` when I saw your upadate :-) 

As far as I know this is currently the only way to handle this kind of behaviour as `focusout` and `focusin`  has not been implementned natively in React. There is a [longstanding issue for it](https://github.com/facebook/react/issues/6410) which I see is mentioned in the issue you mention above.

I agree that it could more than likely be a React bug, would you be able to confirm that in your Firefox?

I suppose the other option would be to attach `focusin` and `focusout` eventhandlers using the `ref` but I'd rather use more standard React code in the accessible example.

danburzo:
Logged an issue here: https://github.com/facebook/react/issues/12993, could you confirm on that thread that the simplified test works on FF/Win?

AlmeroSteyn:
Awesome, thank you! I have responded on the other issue.

Nice to have found this!

AlmeroSteyn:
@danburzo Aside from the issue on Firefox, are you happy with the changes after your review?

AlmeroSteyn:
I would like to merge this as it is open for a while now. Does anyone have any additional comments on this PR? If not I will merge soon.

Issue #925, opened 2018-06-03
Update getDerivedStateFromProps parameter documentation


eluchsinger:
The [documentation ](https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops) of `static getDerivedStateFromProps` does not explain what parameters you are getting. 

I prefer the way other methods, like `componentDidUpdate` are [documented](https://reactjs.org/docs/react-component.html#componentdidupdate), where the parameters names explain what they are.

For example, `getDerivedStateFromProps` should be defined as follows:
```es6
static getDerivedStateFromProps(nextProps, prevState)
```
instead of:

```es6
static getDerivedStateFromProps(props, state)
```

This simple change would make the life of the developers easier, who are new to React or changing from the deprecated lifecycle methods.

pinkbunny1:
@eluchsinger 
I agree because I was looking at the source to see if the props and state is from prev or next. Also looking at the source, it was not so clear. https://github.com/facebook/react/blob/a7d07ff24d8d3d925ab45f5eb9be42a62a345dad/packages/react-dom/src/server/ReactPartialRenderer.js#L473

So state is from `inst` that is instance of component that takes `element.prop`. Not sure how taking the current `element.prop` gives prev `state` .

PR #926, opened 2018-06-03
Updated documentation of getDerivedStateFromProps

eluchsinger:
Renamed parameters to be more informative

I took myself the time to make the change proposed in #925, in case it gets accepted.


alexkrolick:
This was just changed: https://github.com/reactjs/reactjs.org/pull/910

PR #930, opened 2018-06-05
Mention getSnapshotBeforeUpdate should return.

toshi38:
If getSnapshotBeforeUpdate does not return a snapshot value or null
it will cause a warning.  Document that it should return this value.


toshi38:
Seems to break on something unrelated to my changes, can someone retrigger this? ... or do you prefer I force push to retrigger?

gaearon:
Looks good, thanks.

PR #931, opened 2018-06-05
Blog: When to Use Derived State

bvaughn:
Based on some initial feedback, I should probably make the following tweaks:
- [x] Clarify that gDSFP is only meant for rare use cases (like the examples I link out to from the previous blog post). I can probably do this by changing the title back to something like, "You Probably Don't Need Derived State".
- [ ] Make it clearer (maybe by showing it in examples, or maybe just calling it out more explicitly) that the pitfalls I'm showing for gDSFP also apply to cWRP. I might be able to achieve this by showing the before/bad examples using cWRP.
- [ ] More explicitly mention that gDSFP is an _advanced_ lifecycle, and it isn't meant to be used most of the time. (The title change might also convey this.)
- [ ] Add a note to the intro that the post goes in depth, but they key takeaway is that prop mirroring has always been a bad idea, and the post explains why (and some alternatives if your code already relies on it).


content/blog/2018-06-07-when-to-use-derived-state.md
```diff
@@ -0,0 +1,327 @@
+---
+title: "When should I use derived state?"
```
gaearon: Nit: All Capitalized Please

gaearon: I'd also prefer “When to Use Derived State?” because I think a personal-sounding title looks a bit odd in the list. Like someone posted their question on our blog. :-)

screen shot 2018-06-05 at 6 40 34 pm

bvaughn: Oh, whoops. I mean to rename the title to match the file (which was already "when to use derived state"). Good call.

gaearon:
The fonts end up really jumpy.

screen shot 2018-06-05 at 6 45 03 pm

Maybe let's make the intro shorter? Or not use the code highlighting inside a link. Or fix it to look better.

content/blog/2018-06-07-when-to-use-derived-state.md
```diff
@@ -0,0 +1,327 @@
...
+[`getDerivedStateFromProps`](/docs/react-component.html#static-getderivedstatefromprops) was introduced in 16.3. At the time, we provided [some examples](/blog/2018/03/27/update-on-async-rendering.html#examples) of how to use the new lifecycle to derive state from props. Our primary goal was to help people migrate from legacy lifecycles to newer ones that are safer to use with the [upcoming async rendering mode](blog/2018/03/01/sneak-peek-beyond-react-16.html).
+
+Since that blog post was written, we've observed some common usage patterns for both the legacy `componentWillReceiveProps` and the newer `getDerivedStateFromProps` that cause bugs or otherwise confusing behavior. The 16.4 bugfix [makes this behavior more predictable](https://github.com/facebook/react/issues/12898), but the fixes aren't always obvious. Hopefully this blog post will help!
+
+Let's take a look at the following topics:
+* [When should I use derived state?](#when-should-i-use-derived-state)
+* [When should I avoid derived state?](#when-should-i-avoid-derived-state)
+* [What about memoization?](#what-about-memoization)
+
+# When should I use derived state?
```
gaearon: These should all be level-two headings or lower, I think.

content/blog/2018-06-07-when-to-use-derived-state.md
```diff
@@ -0,0 +1,327 @@
...
+
+The most common mistake with derived state is mixing these two. In other words, when props control a value in state that is also updated by `setState` calls. This may sound similar to the [external data loading example](/blog/2018/03/27/update-on-async-rendering.html#fetching-external-data-when-props-change) mentioned above, but it's different in a few important ways.
+
+In the loading example, there is a clear source of truth for both the "source" prop and the "loading" state. When the source prop changes, the loading state should **always** be overridden. Conversely the state should only be overridden when the prop **changes** and should otherwise be managed by the component.
+
+Problems arise when any of these constraints are changed. This typically comes in two forms. Let's take a look at both.
+
+## Anti-pattern: Props always override state
```
gaearon: So this should probably become level-three etc.

bvaughn:
Okay. I've added a link to a Code Sandbox timer-based demo. I am unable to test it currently b'c of the terrible Internet connection I'm on.

content/blog/2018-06-07-when-to-use-derived-state.md
```diff
@@ -0,0 +1,327 @@
...
+
+For the simple example above, we could "fix" the problem of unexpected rerenders using `shouldComponentUpdate`. However in practice, components usually accept multiple props, and our component would rerender if any one of them changed. Even if none of them "changed", props that are functions or objects are often created inline and so will always bypass `shouldComponentUpdate`. For example, what if our component accepted a function to validate the current email address?
+```js
+<EmailInput
+  email={this.props.user.email}
+  validate={this.validateEmail.bind(this)}
+/>
+```
+
+The above example binds the validation callback inline and so it will pass a new function prop every time it renders- effectively bypassing `shouldComponentUpdate` entirely. [Here is a demo](codesandbox://when-to-use-derived-state/derived-state-anti-pattern) that uses a timer to illustrate this problem. Since you can't prevent others from using your component this way, it's inherently fragile.
```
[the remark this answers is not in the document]
bvaughn: I don't want to get into the "other data" explanation quite yet, because it bleeds into the technique of checking for the specific changed prop, which isn't introduced quite yet.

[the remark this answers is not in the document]
bvaughn: I'll post a re-wording of this section. Does it help?

content/blog/2018-06-07-when-to-use-derived-state.md
```diff
@@ -0,0 +1,331 @@
...
+  // ...
+}
+```
+
+We've just made a big improvement. Now our component will no longer erase what we've typed every time it renders. At this point, what we have is fairly similar to the [scroll direction example](blog/2018/03/27/update-on-async-rendering.html#updating-state-based-on-props) we mentioned before. (It will also more closely mirror how `componentWillReceiveProps` would have been used in the past.)
+
+There is still one subtle problem though, and it's probably easiest to illustrate with an example. Let's say our `EmailInput` component is used inside of an "edit profile" form. The first time the form is rendered, the component will display our email address. Let's say we edited the form (including our email address) but then changed our mind and clicked a "reset" button. At this point, we would expect all form fields to return to their initial values– but the `EmailInput` component **will not be reset**. Do you know why?
+
+The reason for this is that `props.email` never actually changed in the above scenario. Both times the "edit profile" form rendered, it would pass our saved email address via props.
+
```
gaearon: WDYT about adding:

>This problem manifests itself even if you don't have a "Reset" button. For example, imagine a password manager app using the above input implementation. When we navigate between the detail forms for two accounts with the same initial email, the input would fail to reset because the prop value from above is the same! That would be a surprise to the user, as a draft change to one account shouldn't affect another account after navigation.

content/blog/2018-06-07-when-to-use-derived-state.md
```diff
@@ -0,0 +1,331 @@
...
+As a general rule, if you're not sure about whether you should use `getDerivedStateFromProps`, here are some questions to ask yourself:
+* Is the state **derived** from props (as opposed to just mirroring it)?
+* Is the state update specifically triggered by a **props change** (and not just the current props value)?
+
+If your answer to either of the above questions is "no" then there are better patterns to use.
+
+## When should I avoid derived state?
+
+The terms ["controlled"](/docs/forms.html#controlled-components) and ["uncontrolled"](/docs/uncontrolled-components.html) are often used to refer to form components, but they can also be used to describe where a component's state lives. A piece of behavior that is passed in as props can be thought of as **controlled** and one that exists only in internal state can be thought of as **uncontrolled**.
```
gaearon: "piece of behavior" sounds like you're talking about passing methods. Maybe just "data"? I would also explain why we use these terms.

>Data that is passed in as props can be thought of as **controlled** because any parent component can "control" this data and change or reset it at any time. A controlled component can't reset its own state, but it can ask a parent to do so—typically using a callback prop like `onChange`.
>
>On the other hand, data that only exists in the component's own local state can be thought of as **uncontrolled**. This is because the parent has no control over it, and can't directly change or reset it.

content/blog/2018-06-07-when-to-use-derived-state.md
```diff
@@ -0,0 +1,331 @@
+---
+title: "When to Use Derived State?"
+author: [bvaughn]
+---
+
+React 16.4 included a [bugfix for getDerivedStateFromProps](/blog/2018/05/23/react-v-16-4.html#bugfix-for-getderivedstatefromprops) that caused some pain in the React community. We apologize for this. We take such changes very seriously. In this case we believe it was the right decision, but we could have done a better job with communication.
+
+[`getDerivedStateFromProps`](/docs/react-component.html#static-getderivedstatefromprops) was introduced in 16.3. At the time, we provided [some examples](/blog/2018/03/27/update-on-async-rendering.html#examples) of how to use the new lifecycle to derive state from props. Our primary goal was to help people migrate from legacy lifecycles to newer ones that are safer to use with the [upcoming async rendering mode](blog/2018/03/01/sneak-peek-beyond-react-16.html).
```
gaearon: Do we need the `getDerivedStateFromProps` link here? There's a lot of links in this paragraph and it makes it a bit harder to scan. I don't think the link is particularly useful because it just gives you a one-sentence description.

If we remove the link I'd suggest starting with English words, e.g.

>The `getDerivedStateFromProps` lifecycle was [...]

content/blog/2018-06-07-when-to-use-derived-state.md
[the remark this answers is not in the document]
gaearon: Yeah this is great

content/blog/2018-06-07-when-to-use-derived-state.md
```diff
@@ -0,0 +1,331 @@
...
+
+There is still one subtle problem though, and it's probably easiest to illustrate with an example. Let's say our `EmailInput` component is used inside of an "edit profile" form. The first time the form is rendered, the component will display our email address. Let's say we edited the form (including our email address) but then changed our mind and clicked a "reset" button. At this point, we would expect all form fields to return to their initial values– but the `EmailInput` component **will not be reset**. Do you know why?
+
+The reason for this is that `props.email` never actually changed in the above scenario. Both times the "edit profile" form rendered, it would pass our saved email address via props.
+
+This design is fundamentally flawed, but it's an easy mistake to make. [I've made it myself.](https://twitter.com/brian_d_vaughn/status/959600888242307072) Fortunately there are two alternatives that work better. The key to both is for **state to only be owned by a single component**. Let's take a look at each.
```
gaearon: I think "state to only be owned by a single component" can be misread as "move all your application state to one component".

Maybe:

>The key to both is that **for any piece of state, you need to pick a single component that owns it as the source of truth, and avoid duplicating it in other components**. Let's take a look at each.

content/blog/2018-06-07-when-to-use-derived-state.md
```diff
@@ -0,0 +1,331 @@
...
+
+  handleChange = event => this.setState({ email: event.target.value });
+}
+```
+
+This design makes the component a little easier to use, but it also has a downside. Let's say that our "edit profile" form could be used to edit multiple users. Whenever a new user is selected, the form should reset the email field. If our `EmailInput` is uncontrolled, it will ignore any changes to `props.defaultEmail`. So how could we support this use case?
+
+#### Option 1: Reset uncontrolled component with a `key`
+
+One approach would be to use a special React prop called `key`. React uses this property to decide whether to [_update_ a component instance or _create_ a new one](/docs/reconciliation.html#keys). In our case, we could use an attribute like a user id to recreate the email input any time a new user is selected. Each time the `EmailInput` is recreated, it will "reset" its state to the value specified by the `defaultEmail` prop.
```
gaearon: We tend to say "attribute" rather than "prop" for key because it doesn't end up in your props.

content/blog/2018-06-07-when-to-use-derived-state.md
```diff
@@ -0,0 +1,331 @@
...
+
+```js
+<EmailInput
+  defaultEmail={this.props.user.email}
+  key={this.props.user.id}
+/>
+```
```
gaearon: I'd add here:

>Note that with this approach, you don't have to add `key` to every input. Instead, it might make sense to put a `key` on the whole form. Every time the key changes, the DOM and components will be re-created with a freshly initialized state. While this may sound slow, in practice the difference is often insignificant, and it can even be faster if the components have heavy logic that runs on updates.

content/blog/2018-06-07-when-to-use-derived-state.md
```diff
@@ -0,0 +1,331 @@
...
+  // ...
+}
+```
+
+The parent form component could then [use a `ref` to call this method](/docs/glossary.html#refs).
+
+> Note
+>
+> Refs can be useful when interfacing with imperative third party APIs or in certain, limited use cases like the one mentioned above, but generally we recommend you use them sparingly.
+
```
gaearon: I think it would help if there was a small recap here that gave me a quick path to choosing which option/alternative to use. For example by now I've already forgotten about controlled components but that may be the best solution for my use case.

content/blog/2018-06-07-when-to-use-derived-state.md
```diff
@@ -0,0 +1,331 @@
...
+    }
+    // Other derived fields would go here...
+    return updatedState;
+  }
+
+  // ...
+}
+```
+
+Although the above pattern _works_, it isn't what `getDerivedStateFromProps` is meant to be used for. This component isn't reacting to **changes in props**. It's just computing values based on the **current props**, and there is a simpler way to do this that doesn't event require state.
```
gaearon: "even"

content/blog/2018-06-07-when-to-use-derived-state.md
```diff
@@ -0,0 +1,331 @@
...
+        <input onChange={this.handleChange} value={this.state.filterText} />
+        <ul>{this.state.filteredList.map(item => <li key={item.id}>{item.text}</li>)}</ul>
+      </Fragment>
+    );
+  }
+
+  handleChange = event => this.setState({ filterText: event.target.value });
+}
+```
+
```
gaearon: I think this example would benefit from one intermediate simpler "after" version that just makes it a pure component, and does the calculation in render. I think we need to push harder against the preconception that you have to move out all the calculation from the render method. A lot of this comes from experience with Ember/Vue and not from any real performance concern.

Then we can say "but what if the component takes some other often-changing prop? Then memoization might be beneficial" and go into the example below.

content/blog/2018-06-07-when-to-use-derived-state.md
```diff
@@ -0,0 +1,331 @@
...
+This is much simpler and should perform just as well as the previous version!
+
+There are a couple of constraints to consider when using memoization though:
+
+1. In most cases, you'll want to **attach the memoized function to a component instance**. This prevents multiple instances of a component from resetting each other's memoized keys.
+1. Typically you'll want to use a memoization helper with a **limited cache size** in order to prevent memory from "leaking" over time. (In the example above, we used `memoize-one` because it only caches the most recent argument and result.)
+1. Memoization will not work properly if `props.items` is recreated each time the parent component renders. Fortunately, this should not be the case for large arrays. If it were, you'd want to address that problem first! (This limitation also applies to the `getDerivedStateFromProps` version above.)
```
gaearon: "Fortunately" sounds like we guarantee it somehow. "Hopefully"?

content/blog/2018-06-07-when-to-use-derived-state.md
bvaughn: I dig it.

content/blog/2018-06-07-when-to-use-derived-state.md
bvaughn: Yeah, "data" probably works better than "piece of behavior 😄

content/blog/2018-06-07-when-to-use-derived-state.md
bvaughn: 👍

content/blog/2018-06-07-when-to-use-derived-state.md
bvaughn: Ooh, that's a nice suggestion. Thanks.

content/blog/2018-06-07-when-to-use-derived-state.md
bvaughn: How about something like...
> To recap, there are a couple of options for resetting an uncontrolled component's state. Which one is best depends on the component you are writing. Here is a rough guide:
1. To reset _all internal state_, use the `key` attribute.
2. To reset _some internal state_, watch for changes in a special property (e.g. `props.id`).
3. Otherwise, consider falling back to an imperative instance method called with a `ref`.

content/blog/2018-06-07-when-to-use-derived-state.md
bvaughn: Oh nice. I was uncertain about this wording choice to begin with. I like that.

content/blog/2018-06-07-when-to-use-derived-state.md
gaearon: This leaves out the (often best) option: lift the state up and deal with it explicitly in the parent. For example, a form might want to have both `draftValues` and `committedValues` as state in a parent component.

gaearon: (I'm proposing to summarize the whole "how to avoid anti-patterns" section, not just the uncontrolled one)

bvaughn: Do you have proposed wording for this summary?

It can be a little hard to just insert changes like this after I've read and re-read the post so may times. I get kind of stuck in my thinking patterns.

Let me push what I have here, and then maybe you can make one last round fo suggestions.

bvaughn: I will make an attempt at doing a more comprehensive summary though.

content/blog/2018-06-07-when-to-use-derived-state.md
bvaughn: Oh, snap. I think I missed this one comment. I'm going to step away for a few minutes, and then I'll re-read this section with your comment in mind.

content/blog/2018-06-07-when-to-use-derived-state.md
gaearon: 1. **If you're trying to "mirror" a prop value in the state, you can make the component controlled, and consolidate the two diverging values in the state of some parent component.** For example, an input tracking a "draft" `state.value` that's derived from a "committed" `props.value` could instead always use `props.value` and call `props.onChange`. A form component above could then explicitly own two diverging values in its own state, such as `state.draftValue` and `state.committedValue`, pass one of them down, and reset either of them when appropriate. (Don't forget you can use a nested object if the form contains multiple fields!)

2. **If you're trying to "reset" one or more state fields when a particular unrelated prop (usually an ID) changes,** you have a few options:

(your list)

bvaughn: It's not clear if you're suggesting replacing my intro paragraph (below) or inserting between it and the uncontrolled bullets.

I think my recap had a bit of context that this suggestion loses.

I'll try to combine the two.

gaearon: I meant that your list comes under the second point :-)

screen shot 2018-06-05 at 11 02 45 pm

content/blog/2018-06-07-when-to-use-derived-state.md
bvaughn: I like this suggestion. I've pushed a change.

content/blog/2018-06-07-when-to-use-derived-state.md
```diff
@@ -0,0 +1,384 @@
...
+    return <input onChange={this.handleChange} value={this.state.email} />;
+  }
+
+  handleChange = event => {
+    this.setState({ email: event.target.value });
+  };
+}
+```
+
+At first, this component might look okay. State is initialized to the value specified by props and updated when we type into the `<input>`. But once our component re-renders– either because it called `setState` or because its parent re-rendered– anything we've typed into the `<input>` will be lost! (See [this demo for an example](codesandbox://when-to-use-derived-state/derived-state-anti-pattern).)
```
gaearon: The demo code looks different than the snippet above.

content/blog/2018-06-07-when-to-use-derived-state.md
```diff
@@ -0,0 +1,384 @@
...
+
+For the simple example above, we could "fix" the problem of unexpected re-renders using `shouldComponentUpdate`. However in practice, components usually accept multiple props, and our component would re-render if any one of them changed. Even if none of them "changed", props that are functions or objects are often created inline and so will always bypass `shouldComponentUpdate`. For example, what if our component accepted a function to validate the current email address?
+```js
+<EmailInput
+  email={this.props.user.email}
+  validate={this.validateEmail.bind(this)}
+/>
+```
+
+The above example binds the validation callback inline and so it will pass a new function prop every time it renders– effectively bypassing `shouldComponentUpdate` entirely. [Here is a demo](codesandbox://when-to-use-derived-state/derived-state-anti-pattern) that uses a timer to illustrate this problem. Because it might break at any time your component needs to accept new props, this design pattern is inherently fragile.
```
gaearon: I understand why you did the `prevProps` comparison in that demo link but I think it would be easier to understand if instead you used the same exact code as in the inline example above, but pinned the React version at 16.3.

*Then* it would be a convincing demo of it not working in 16.3 (as advertised above).

content/blog/2018-06-07-when-to-use-derived-state.md
```diff
@@ -0,0 +1,384 @@
...
+
```
gaearon: I would add 

>Even before `getDerivedStateFromProps` was introduced, the same exact pattern led to bugs in `componentWillReceiveProps`. Here is another demo that shows it.

(and link to a cWRP sandbox port of this snippet)

content/blog/2018-06-07-when-to-use-derived-state.md
```diff
@@ -0,0 +1,384 @@
...
+There is still one subtle problem though, and it's probably easiest to illustrate with an example. Let's say our `EmailInput` component is used inside of an "edit profile" form. The first time the form is rendered, the component will display our email address. Let's say we edited the form (including our email address) but then changed our mind and clicked a "reset" button. At this point, we would expect all form fields to return to their initial values– but the `EmailInput` component **will not be reset**. Do you know why?
+
+The reason for this is that `props.email` never actually changed in the above scenario. Both times the "edit profile" form rendered, it would pass our saved email address via props.
+
+This problem could manifest itself even without a "Reset" button. For example, imagine a password manager app using the above input component. When navigating between details for two accounts with the same initial email, the input would fail to reset because the prop value passed to the component would be the same! This would be a surprise to the user, as a draft change to one account would appear to affect other accounts that happened to share the same email.
```
gaearon: Let's change "Reset" to "reset" so casing is consistent with the paragraph where we introduced it

content/blog/2018-06-07-when-to-use-derived-state.md
```diff
@@ -0,0 +1,384 @@
...
+
+For **uncontrolled** components, if you're trying to **"reset" one or more state fields** when a particular unrelated prop (usually an ID) changes, you have a few options:
+1. To reset _all internal state_, use the `key` attribute.
+2. To reset _some internal state_, watch for changes in a special property (e.g. `props.id`).
+3. Otherwise, consider falling back to an imperative instance method called with a `ref`.
+
+## What about memoization?
+
+Another thing we've seen derived state used for is memoization. This isn't necessarily bad, but there are reasons you may want to avoid it. Hopefully the above examples illustrate that there is a certain amount of complexity inherent in using `getDerivedStateFromProps`. This complexity increases with each derived property. For example, if we add a second derived field to our component state then our implementation would need to separately track changes to both:
```
gaearon: Maybe:

>We've also seen derived state used to ensure an expensive value is recomputed only when the inputs change, also known as [memoization](https://en.wikipedia.org/wiki/Memoization).

I'm sure many people who use this pattern don't know its name

content/blog/2018-06-07-when-to-use-derived-state.md
```diff
@@ -0,0 +1,384 @@
...
+```js
+// PureComponents only re-render if at least one state or prop value changes.
+// Change is determined by doing a shallow comparison of state and prop keys.
+class Example extends PureComponent {
+  // State only needs to hold the current filter text value:
+  state = {
+    filterText: ""
+  };
+
+  render() {
```
gaearon: Maybe add here:

```js
    // The render method on PureComponent is not called unless either
    // props.list or state.filterText have referentially different values
```

content/blog/2018-06-07-when-to-use-derived-state.md
```diff
@@ -0,0 +1,384 @@
...
+
+    return (
+      <Fragment>
+        <input onChange={this.handleChange} value={this.state.filterText} />
+        <ul>{filteredList.map(item => <li key={item.id}>{item.text}</li>)}</ul>
+      </Fragment>
+    );
+  }
+
+  handleChange = event => {
```
gaearon: Nit (in this and other examples): I think most or maybe all doc examples put event handlers before render so it jumps out a little bit.

content/blog/2018-06-07-when-to-use-derived-state.md
```diff
@@ -0,0 +1,384 @@
...
+    this.setState({ filterText: event.target.value });
+  };
+}
+```
+
+This is much simpler and should perform just as well as the previous version!
```
gaearon: Are we certain? If yes, let's drop "should" and just say "performs"?

content/blog/2018-06-07-when-to-use-derived-state.md
```diff
@@ -0,0 +1,384 @@
...
+1. Typically you'll want to use a memoization helper with a **limited cache size** in order to prevent memory from "leaking" over time. (In the example above, we used `memoize-one` because it only caches the most recent argument and result.)
+1. Memoization will not work properly if `props.items` is recreated each time the parent component renders. Hopefully, this should not be the case for large arrays. If it were, you'd want to address that problem first! (This limitation also applies to the `getDerivedStateFromProps` version above.)
+
+## In closing
+
+The examples above are intentionally simplified in order to highlight specific coding patterns. In real world applications, components often contain a mix of controlled and uncontrolled behaviors. This is okay! Just be careful to ensure that each behavior has a clear source of truth in order to avoid the anti-patterns mentioned above.
+
+If you have a use case that you think falls outside of these patterns, please share it with us on [GitHub](https://github.com/reactjs/reactjs.org/issues/new) or Twitter!
```
gaearon: One thing we haven't covered at all is Redux. In that case it's not clear what the recommended solution is (you can't lift draft state up without making it slow).

content/blog/2018-06-07-when-to-use-derived-state.md
bvaughn: Yeah. Unfortunately I can't do that with the code sandbox repl plug-in, and I've used all my free sandboxes up 😅 maybe we should temporarily use code pen for this?

content/blog/2018-06-07-when-to-use-derived-state.md
bvaughn: Not sure I want to cover Redux in this post. I haven't used Redux in a long time, and I haven't thought about that use case much as a result. Do you think it's necessary?

content/blog/2018-06-07-when-to-use-derived-state.md
gaearon: Yea let's just use CodePen for this post.

content/blog/2018-06-07-when-to-use-derived-state.md
gaearon: I think it's necessary (it seems like one of the most popular reasons why people even *want* to "mirror" state—the committed one lives in Redux, and the draft is in local state). Judging by the thread, Redux folks also seem to often be unaware of this problem because PureComponent obscures it.

I don't think we need to mention it by name or dedicate a section, but just casually mentioning in an appropriate section that it *also* serves as a solution when the source of truth is outside React would help.

gaearon: I'm not sure which section that would be by the way.

content/blog/2018-06-07-when-to-use-derived-state.md
bvaughn: Yeah. I intended the demo to show the general problem, not really to exactly mirror the code so far. (I started mirroring the code, but then I felt that writing a validate function was weird, and using it kind of distracted from the point I was trying to make...)

content/blog/2018-06-07-when-to-use-derived-state.md
bvaughn: Okay. Maybe something like this...

screen shot 2018-06-05 at 8 54 51 pm

content/blog/2018-06-07-when-to-use-derived-state.md
bvaughn: I've made a new sandbox pinned to 16.3 https://codesandbox.io/s/m3w9zn1z8x

bvaughn:
Okay. I think I've responded to the things you haven't yet responded to. I mentioned Redux (briefly) and updated the Code Sandbox demos to use 16.3 and better match the examples they follow in the post.

content/blog/2018-06-07-when-to-use-derived-state.md
gaearon: It can just be an empty function, no? That would still show the point.

My concern is that it's confusing to have the gDSFP implementation itself be different when the post is highlighting antipatterns in gDSFP. This makes pattern-matching style of learning harder.

content/blog/2018-06-07-when-to-use-derived-state.md
gaearon: Redux users tend to want to reset it though.. :-(

content/blog/2018-06-07-when-to-use-derived-state.md
gaearon: Note I haven't looked at latest edit yet

content/blog/2018-06-07-when-to-use-derived-state.md
bvaughn: The rest of the blog post covers that. I don't think we need to go into great depth about Redux, specifically, in this section.

content/blog/2018-06-07-when-to-use-derived-state.md
bvaughn: The confusion I was concerned about was _how_ the validation function is used. Does it sit between the change handler and the `setState` call? Yeah, probably- but now we're introducing a new kind of semi-controlled behavior into the example which I was worried might distract.

I _think_ I kind of hand waved over this in the latest Code Sandbox, but feedback is welcome.

content/blog/2018-06-07-when-to-use-derived-state.md
```diff
@@ -0,0 +1,388 @@
...
+
+As a general rule, if you're not sure about whether you should use `getDerivedStateFromProps`, here are some questions to ask yourself:
+* Is the state **derived** from props (as opposed to just mirroring it)?
+* Is the state update specifically triggered by a **props change** (and not just the current props value)?
+
+If your answer to either of the above questions is "no" then there are better patterns to use.
+
+## When should I avoid derived state?
+
+The terms ["controlled"](/docs/forms.html#controlled-components) and ["uncontrolled"](/docs/uncontrolled-components.html) are often used to refer to form components, but they can also be used to describe where a component's state lives. Data that is passed in as props can be thought of as **controlled** and one that exists only in internal state can be thought of as **uncontrolled**.
```
gaearon: As I suggested in https://github.com/reactjs/reactjs.org/pull/931#discussion_r193184074, I think even a small elaboration on *why* one is called "controlled" and another is not would help. (Mine was probably too lengthy.) I don't think it's obvious props = controlled because parent "controls" them. Even saying this once would be helpful IMO.

bvaughn: Oh, whoops. I think I missed that point somehow. I just noticed the "data" part. I'll squeeze a small explanation in.

content/blog/2018-06-07-when-to-use-derived-state.md
```diff
@@ -0,0 +1,388 @@
...
+
+There is still one subtle problem though, and it's probably easiest to illustrate with an example. Let's say our `EmailInput` component is used inside of an "edit profile" form. The first time the form is rendered, the component will display our email address. Let's say we edited the form (including our email address) but then changed our mind and clicked a "reset" button. At this point, we would expect all form fields to return to their initial values– but the `EmailInput` component **will not be reset**. Do you know why?
+
+The reason for this is that `props.email` never actually changed in the above scenario. Both times the "edit profile" form rendered, it would pass our saved email address via props.
+
+This problem could manifest itself even without a "reset" button. For example, imagine a password manager app using the above input component. When navigating between details for two accounts with the same initial email, the input would fail to reset because the prop value passed to the component would be the same! This would be a surprise to the user, as a draft change to one account would appear to affect other accounts that happened to share the same email.
```
gaearon: I think it would be nice to split "When navigating [...]" sentence in two. I know I wrote it but it's too long.

content/blog/2018-06-07-when-to-use-derived-state.md
```diff
@@ -0,0 +1,388 @@
...
+
+This design is fundamentally flawed, but it's also an easy mistake to make. [I've made it myself.](https://twitter.com/brian_d_vaughn/status/959600888242307072) Fortunately there are two alternatives that work better. The key to both is that **for any piece of state, you need to pick a single component that owns it as the source of truth, and avoid duplicating it in other components.** Let's take a look at each.
```
gaearon: "Let's take a look at each of the alternatives"? After the expansion of the bold text, I'm losing what "each" referred to.

content/blog/2018-06-07-when-to-use-derived-state.md
```diff
@@ -0,0 +1,388 @@
...
+  state = {
+    email: this.props.defaultEmail,
+    prevPropsID: this.props.id
+  };
+
+  static getDerivedStateFromProps(props, state) {
+    // Any time the current user changes,
+    // Reset any parts of state that are tied to that user.
+    // In this simple example, that's just the email.
+    if (props.id !== state.prevPropsID) {
```
gaearon: I can imagine somebody thinking `id` is a special thing like `key`. Maybe let's make it `userID` as a more obviously custom one?

content/blog/2018-06-07-when-to-use-derived-state.md
```diff
@@ -0,0 +1,388 @@
...
+
+> Note
+>
+> Refs can be useful when interfacing with imperative third party APIs or in certain, limited use cases like the one mentioned above, but generally we recommend you use them sparingly.
+
+-----
+
+To recap, when designing a component, it is important to decide whether its data will be controlled or uncontrolled.
+
+If you're trying to **"mirror" a prop value in the state**, you can make the component **controlled**, and consolidate the two diverging values in the state of some parent component. For example, rather than accepting a "committed" `props.value` and tracking a "draft" `state.value`– a component could only display `props.value` and call `props.onChange` to notify the parent of a "draft" update. A form component above could then explicitly manage both values in its own state, (e.g. `state.draftValue` and `state.committedValue`), pass one of them down, and reset either of them when appropriate. (Don't forget you can use a nested object if the form contains multiple fields!)
```
gaearon: Finding the dash before "a component" a bit odd. Don't know why.

content/blog/2018-06-07-when-to-use-derived-state.md
gaearon: Yeah upon re-reading I think this is sufficient

bvaughn: Nice! 😄

content/blog/2018-06-07-when-to-use-derived-state.md
```diff
@@ -0,0 +1,388 @@
+---
+title: "When to Use Derived State?"
+author: [bvaughn]
+---
+
+React 16.4 included a [bugfix for getDerivedStateFromProps](/blog/2018/05/23/react-v-16-4.html#bugfix-for-getderivedstatefromprops) which made some existing bugs in React components reproduce more consistently. We carefully considered this change, an although we believe it was the right decision, we apologize for any inconvenience it caused. In this post, we will explain some common problems with deriving state, and the preferred solutions to them.
```
gaearon: "and although"

content/blog/2018-06-07-when-to-use-derived-state.md
```diff
@@ -0,0 +1,388 @@
...
+
+There are a couple of constraints to consider when using memoization though:
+
+1. In most cases, you'll want to **attach the memoized function to a component instance**. This prevents multiple instances of a component from resetting each other's memoized keys.
+1. Typically you'll want to use a memoization helper with a **limited cache size** in order to prevent memory from "leaking" over time. (In the example above, we used `memoize-one` because it only caches the most recent argument and result.)
+1. Memoization will not work properly if `props.items` is recreated each time the parent component renders. Hopefully, this should not be the case for large arrays. If it were, you'd want to address that problem first! (This limitation also applies to the `getDerivedStateFromProps` version above.)
+
+## In closing
+
+The examples above are intentionally simplified in order to highlight specific coding patterns. In real world applications, components often contain a mix of controlled and uncontrolled behaviors. This is okay! Just be careful to ensure that each behavior has a clear source of truth in order to avoid the anti-patterns mentioned above.
```
gaearon: Maybe we could add a small blurb here saying that in general `getDerivedStateFromProps` should be used sparely, and that its verbosity exposes the complexity of what you're trying to do.

content/blog/2018-06-07-when-to-use-derived-state.md
```diff
@@ -0,0 +1,388 @@
...
+
+The `getDerivedStateFromProps` lifecycle was introduced in 16.3. At the time, we provided [some examples](/blog/2018/03/27/update-on-async-rendering.html#examples) of how to use the new lifecycle to derive state from props. Our primary goal was to help people migrate from legacy lifecycles to newer ones that are safer to use with the [upcoming async rendering mode](blog/2018/03/01/sneak-peek-beyond-react-16.html).
+
+Since that blog post was written, we've observed some common usage patterns for both the legacy `componentWillReceiveProps` and the newer `getDerivedStateFromProps` that cause bugs or otherwise confusing behavior. The 16.4 bugfix [makes this behavior more predictable](https://github.com/facebook/react/issues/12898), but the fixes aren't always obvious. Hopefully this blog post will help!
```
flarnie: > Hopefully this blog post will help!

This is a nit, but I think we could phrase this more confidently. Something like "We've written this blog post to help fix those cases." or "This blog post shares the solutions we've found to work internally at Facebook."

content/blog/2018-06-07-when-to-use-derived-state.md
```diff
@@ -0,0 +1,388 @@
...
+
+So what does it mean to update state as the result of changes in props? Our previous blog post provided a few examples, like [managing the current scroll direction based on an offset prop](/blog/2018/03/27/update-on-async-rendering.html#updating-state-based-on-props) or [loading external data specified by a source prop](/blog/2018/03/27/update-on-async-rendering.html#fetching-external-data-when-props-change).
+
+As a general rule, if you're not sure about whether you should use `getDerivedStateFromProps`, here are some questions to ask yourself:
+* Is the state **derived** from props (as opposed to just mirroring it)?
+* Is the state update specifically triggered by a **props change** (and not just the current props value)?
+
+If your answer to either of the above questions is "no" then there are better patterns to use.
```
flarnie: I like this phrasing - "there are better patterns to use".

bvaughn:
I'm closing this PR because it has gotten a little noisy. I'll apply feedback and create a fresh PR with it (and link it here).

bvaughn:
Continued with PR #934

PR #932, opened 2018-06-06
Update TIC TAC TOE example to use ES6 string templates 

lucas-aragno:
We noticed that most of the code is written using ES6 features so it might be good to keep it consistent


gaearon:
Thanks but we’d rather keep the syntax familiar for people who don’t necessarily know these new features. Since they’re not essential for the actual problem being solved.

PR #933, opened 2018-06-06
Add React Boston 2018 to upcoming conferences

andrewrota:
We've announced React Boston 2018 and would like to add it to the upcoming conferences page.

Thanks!
Andrew


andrewrota:
Added with #1081

PR #947, opened 2018-06-10
Reorder and rewrite some parts of lifecycle reference

gaearon:
A few changes based on common points of confusion I'm seeing.

* I grouped legacy lifecycles in the TOC of each section together. I know this removes "ordering" information about them but they're legacy so IMO it's not too bad, and it makes it clearer for new users which are the good ones.

* I split the actual content section into Commonly Used Lifecycle Methods, Rarely Used Lifecycle Methods, and Legacy Lifecycle Methods. This both helps with learning (you know which ones to learn first), and helps stress you shouldn't use the others one often.

* I removed some details that I think are unnecessary here (like fragment example) and which already have dedicated pages.

* Changed the `constructor` section to be more explicit that copying props into state is rarely a good idea, linked it to our new blog post.

* Changed cWRP and gDSFP docs to link to the new blog post.


gaearon:
I was in the mood to ship something so I merged and happy to address follow ups.

PR #948, opened 2018-06-10
Mark common lifecycles as bold

gaearon:
Haven’t looked at how it renders yet but I think this may help clarify.


gaearon:
I think I like this

PR #955, opened 2018-06-11
Add propTypes to class properties list

gnapse:
The list of class properties valid for a `React.Component` class should include `propTypes`.


gaearon:
Can we make it a local section on the page that contains a short paragraph that links to the propTypes doc?

gnapse:
Sure, will do.

PR #956, opened 2018-06-11
Quick Start -> Main Concepts

gaearon:
I was never a fan of **Quick Start** as a name for that section. There's nothing really "quick" about going through all these pages. It wasn't clear how "Quick Start" is different from "Tutorial", or that it was ordered in a step-by-step way.

What this section really is, is a walkthrough of React concepts. That's too long so I settled for "Main Concepts" as a name. I'm very open to other descriptive names if you have better ideas but I can't look at Quick Start anymore.

"Hello World" looks a bit odd in "Main Concepts". But I can live with it. We can rename it to "Before We Begin" or something similar.

For now I just want to get rid of Quick Start. With "concepts" in the name it's much clearer that when you're confused in any other section, this is where you should be looking for an intro.


content/docs/nav.yml
```diff
@@ -78,7 +78,7 @@
       title: Uncontrolled Components
     - id: web-components
       title: Web Components
-- title: Reference
+- title: API Reference
```
gaearon: Also this.

PR #958, opened 2018-06-11
Update 2018-06-07-you-probably-dont-need-derived-state.md

balavishnu-vj:
Typo: I think it should be props.list instead of props.items.


gaearon:
Thanks

PR #962, opened 2018-06-11
Reorganize Hello World Page

alexkrolick:
Emphasizes the text that explains the organization of the Main Concepts section and the introductory code by moving the codeblock to the very top and adding a "following along" heading for the install steps

### DIFF
--- a/content/docs/hello-world.md
+++ b/content/docs/hello-world.md
@@ -12,7 +12,7 @@ redirect_from:
   - "docs/getting-started-zh-CN.html"
 ---
 
-Get Started With React
+Getting Started With React
 
 The smallest React example looks like this:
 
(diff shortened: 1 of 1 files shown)


content/docs/hello-world.md
```diff
@@ -12,7 +12,7 @@ redirect_from:
   - "docs/getting-started-zh-CN.html"
 ---
 
-The easiest way to get started with React is to use [this Hello World example code on CodePen](codepen://hello-world). You don't need to install anything; you can open it in another tab and follow along as we go through examples. If you'd rather use a local development environment, check out the [Installation](/docs/try-react.html) section.
+Getting Started With React
```
gaearon: Do we ever need this heading?

alexkrolick: Looks a bit weird if the next like becomes the subheading

gaearon: I don't think I agree.

It's not a subheading, it's a leading paragraph. We use it this way on all the other pages. I don't think we use this style for a subheading.

alexkrolick: Ok

PR #963, opened 2018-06-11
Reorder data-fetching tools

VinayakBagaria:
Keep home-built Relay Modern at top rather than Apollo client.


threepointone:
I think this was meant to be alphabetical order. Thanks for the PR!

PR #968, opened 2018-06-11
docs(forwarding-refs.md): add warning to third-party library maintainers

jedwards1211:
@gaearon what do you think of this message?

### DIFF
--- a/content/docs/forwarding-refs.md
+++ b/content/docs/forwarding-refs.md
@@ -37,7 +37,7 @@ Here is a step-by-step explanation of what happens in the above example:
 >
 >Ref forwarding is not limited to DOM components. You can forward refs to class component instances, too.
 
-## Warning for third-party library maintainers
+## Warning for component library maintainers
 
 **If you start using `forwardRef` in a third-party library, you should release it as a breaking change.**.  This is because your library has an observably different behavior (including the component type and what the ref gets assigned to), and can break apps and other libraries that depend on the old behavior.
 
--- a/content/docs/forwarding-refs.md
+++ b/content/docs/forwarding-refs.md
@@ -39,9 +39,9 @@ Here is a step-by-step explanation of what happens in the above example:
 
 ## Warning for third-party library maintainers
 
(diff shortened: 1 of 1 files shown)


content/docs/forwarding-refs.md
```diff
@@ -37,6 +37,12 @@ Here is a step-by-step explanation of what happens in the above example:
 >
 >Ref forwarding is not limited to DOM components. You can forward refs to class component instances, too.
 
+## Warning for third-party library maintainers
+
+**If you start using `forwardRef` in a third-party library, you should release it as a breaking change.**.  This is because the value returned by `forwardRef` is of a different type than React components have historically been (a `class` extending `React.Component` or a `function`) and many HoC libraries (e.g. `react-redux`' `connect()`) aren't currently compatible with such components.
```
gaearon: I would just leave it at "has an observably different behavior (including the component type and what the ref gets assigned to), and can break apps and other libraries that depend on the old behavior." I don't think there's a need for jargon like "HoC" or mentioning specific libraries.

content/docs/forwarding-refs.md
```diff
@@ -37,6 +37,12 @@ Here is a step-by-step explanation of what happens in the above example:
...
+
+Even if you check whether `React.forwardRef` is defined and only use it if it is present, you should release a breaking change, because it may still cause problems for users of your library.
```
gaearon: We can just say "Conditionally applying `forwardRef` when it exists is also not recommended for the same reasons: it changes how your library behaves and can break your users' apps as they upgrade React versions."

content/docs/forwarding-refs.md
jedwards1211: okay, sounds good

jedwards1211:
@gaearon okay, I made your suggested changes

content/docs/forwarding-refs.md
```diff
@@ -37,6 +37,12 @@ Here is a step-by-step explanation of what happens in the above example:
 >
 >Ref forwarding is not limited to DOM components. You can forward refs to class component instances, too.
 
+## Warning for third-party library maintainers
```
gaearon: Maybe "for component library maintainers" to be clearer

gaearon:
thanks

PR #972, opened 2018-06-13
feat: navigating pages via right and left arrow keys

raunofreiberg:
Adds the ability to navigate through pages by using the left and right arrow keys. Opt-in via the `enableKeyboardNavigation` prop on specific documentation pages (currently enabled for [docs](https://reactjs.org/docs/hello-world.html)). Any pages that have the `prev` or `next` prop defined would be able to use this logic.

This is just an idea that I had and it is completely WIP and implemented on the go, feedback is welcome and I would not take it personally if this feature does not make it in after all :-)

### DIFF
--- a/src/components/MarkdownPage/MarkdownPage.js
+++ b/src/components/MarkdownPage/MarkdownPage.js
@@ -9,13 +9,15 @@ import Container from 'components/Container';
 import Flex from 'components/Flex';
 import MarkdownHeader from 'components/MarkdownHeader';
 import NavigationFooter from 'templates/components/NavigationFooter';
-import React from 'react';
+import React, {Component} from 'react';
 import StickyResponsiveSidebar from 'components/StickyResponsiveSidebar';
 import TitleAndMetaTags from 'components/TitleAndMetaTags';
 import findSectionForPath from 'utils/findSectionForPath';
 import toCommaSeparatedList from 'utils/toCommaSeparatedList';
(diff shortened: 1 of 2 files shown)


src/components/MarkdownPage/MarkdownPage.js
```diff
@@ -43,106 +54,160 @@ const getPageById = (sectionList: Array<Object>, templateFile: ?string) => {
...
-                {markdownRemark.fields.path && (
-                  <div css={{marginTop: 80}}>
-                    <a
-                      css={sharedStyles.articleLayout.editLink}
-                      href={`https://github.com/reactjs/reactjs.org/tree/master/content/${
-                        markdownRemark.fields.path
-                      }`}>
-                      Edit this page
-                    </a>
+  render() {
```
raunofreiberg: No changes in the render method besides the `next` and `prev` variables moved to a class property. The diff is purely cosmetic due to the conversion from a functional component to a class.

gaearon:
Seems like a bit too aggressive? I think it's easy to press this by mistake.

raunofreiberg:
Fair point. Maybe show a tooltip to either inform about this functionality or allow it to be opt-in via a flag by the user?

threepointone:
Thanks for the effort, but I think we won't include this functionality. I tried it out, and I think it might confuse people. Thank you very much for the effort!

PR #973, opened 2018-06-13
updated docs on context rerender.

sach2211:
While playing with context I realized that first level Consumers (components which have consumers inside their render method and none of their ascendents nodes are a consumer) do not actually rerender, but do react to the updated context value.

Would be good idea to mention this in the documentation.

eg - https://codesandbox.io/s/l2kx18n35z

### DIFF
--- a/content/docs/context.md
+++ b/content/docs/context.md
@@ -76,7 +76,9 @@ Requires a [function as a child](/docs/render-props.html#using-props-other-than-
 > 
 > For more information about the 'function as a child' pattern, see [render props](/docs/render-props.html).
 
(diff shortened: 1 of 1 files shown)


yvesgurcan:
Good catch! This is a subtle thing but I'm glad you emphasized it in the docs since I would have assumed the parent components rerenders when the value of the Provider changes.

BartoszKlonowski:
Hello @sach2211, thank you for taking the time to contribute!
I'm afraid the page and section you update is no longer present in the docs, so let me close this PR as outdated.
However, as there is a comment finding this PR useful, if you find it worth rebasing and placing in proper page please reopen.

PR #974, opened 2018-06-13
Add learnreact.design course link and description

lintonye:
This PR adds the link and description of learnreact.design, a series of React courses. It's unique because the courses are tailored for designers, not developers.


gaearon:
Thanks!

PR #976, opened 2018-06-14
Update lifting-state-up.md

rajeshkumarsampath:
Conditionally displaying the Boiling Verdict  only if the user enters the temperature value in the input field


threepointone:
This doesn't seem necessary for the usecase that the example focuses on. But thank you for the effort!

PR #990, opened 2018-06-22

adamsaghy:
`componentWillUpdate` method will be called before the render method so the returning value of the `render` method does not affect.


gaearon:
Fair.

PR #991, opened 2018-06-23
Link to Parcel in Starter Kits

alexkrolick:

### DIFF
--- a/content/community/tools-starter-kits.md
+++ b/content/community/tools-starter-kits.md
@@ -13,6 +13,7 @@ permalink: community/starter-kits.html
 * **[nwb](https://github.com/insin/nwb)** - A toolkit for React apps, libraries and other npm modules for the web
 * **[razzle](https://github.com/jaredpalmer/razzle)** - Create server-rendered universal JavaScript applications with no configuration
 * **[Neutrino](https://neutrino.js.org/)** - Create and build modern JavaScript applications with zero initial configuration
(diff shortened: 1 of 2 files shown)


gaearon:
Actually sorry, I didn't read the link content and assumed there would be a larger section. This is a bit confusing without context. Let's just mention the React preset as a link, like we did with nwb and Neutrino?

content/community/tools-starter-kits.md
```diff
@@ -13,6 +13,7 @@ permalink: community/starter-kits.html
 * **[nwb](https://github.com/insin/nwb)** - A toolkit for React apps, libraries and other npm modules for the web
 * **[razzle](https://github.com/jaredpalmer/razzle)** - Create server-rendered universal JavaScript applications with no configuration
 * **[Neutrino](https://neutrino.js.org/)** - Create and build modern JavaScript applications with zero initial configuration
+* **[Parcel](https://parceljs.org)** - Blazing fast, zero configuration web application bundler that [works with React](https://parceljs.org/recipes.html#react)
```
gaearon: Minor nit: let’s avoid marketing words like “blazing” in our docs, they don’t mean anything 🙂

alexkrolick: Yeah, I let it slide because Gatsby was already using it. Trying to strike a balance between describing the projects as they describe themselves (including marketing) and editorializing.

Maybe we can clean up the recommendations with more specific reasons they are recommended.

gaearon: For now I just removed this from both

PR #992, opened 2018-06-23
Rewrite "adding React to existing app"

### LINKED ISSUE #988: Rewrite "adding React to existing app"
We should throw this way and start over.

https://reactjs.org/docs/add-react-to-an-existing-app.html

Instead it should be two steps:

* Add a React CDN script tag to your page
* Run Babel CLI on a directory to transform JSX/ES6 files into ES5

That's it. No package managers, no minifiers, no bundlers, etc.
We should move all that jazz somewhere else.

If you're adding React to an existing app, it's not our concern how you minify/bundle/etc. If you don't want to, it's really your business.

Otherwise we're looking clowny with all this setup.

alexkrolick:
Fixes #988

[Rendered](https://deploy-preview-992--reactjs.netlify.com/docs/add-react-to-an-existing-app.html)


gaearon:
I have some ideas for a larger set of changes. Will do on top of this. Thanks!

alexkrolick:
Ok cool 😄

content/docs/add-react-to-an-existing-app.md
```diff
@@ -6,91 +6,92 @@ prev: add-react-to-a-new-app.html
...
-To install React with npm, run:
+### 2. Write your components
 
-```bash
-npm init
-npm install --save react react-dom
+```js
+// src/button.js
+
+class Button extends React.Component {
```
sw-yx: prefer to use functional components unless you have a need for classes

content/docs/add-react-to-an-existing-app.md
```diff
@@ -6,91 +6,92 @@ prev: add-react-to-a-new-app.html
...
 ReactDOM.render(
-  <h1>Hello, world!</h1>,
-  document.getElementById('root')
+  <Button label="Edit" />,
+  document.querySelector('.edit')
 );
 ```
 
-This code renders into a DOM element with the id of `root`, so you need `<div id="root"></div>` somewhere in your HTML file.
+### 3. Compile with Babel
```
sw-yx: we don't need babel to compile, we can require inline, the whole point of this piece is that we can use babel in a script:

```
// babel in a script

<script src="https://unpkg.com/babel-standalone@6.26.0/babel.js"></script>

// lower down

<script type="text/babel">
```

content/docs/add-react-to-an-existing-app.md
```diff
@@ -6,91 +6,92 @@ prev: add-react-to-a-new-app.html
...
-Learn [how to tell if your website is serving the right version of React](/docs/optimizing-performance.html#use-the-production-build), and how to configure the production build process most efficiently:
+### 5. Production
 
-* [Creating a Production Build with Create React App](/docs/optimizing-performance.html#create-react-app)
-* [Creating a Production Build with Single-File Builds](/docs/optimizing-performance.html#single-file-builds)
-* [Creating a Production Build with Brunch](/docs/optimizing-performance.html#brunch)
-* [Creating a Production Build with Browserify](/docs/optimizing-performance.html#browserify)
-* [Creating a Production Build with Rollup](/docs/optimizing-performance.html#rollup)
-* [Creating a Production Build with webpack](/docs/optimizing-performance.html#webpack)
+You probably already have a minification step for your JavaScript - don't forget to minify your component files too!
```
sw-yx: don't need this, if you are using the prod version of react ( I think)

content/docs/add-react-to-an-existing-app.md
alexkrolick: That's extremely slow - this is meant for people adding React to an actual app. If they are just trying React they should be referring to the Try React page.

content/docs/add-react-to-an-existing-app.md
alexkrolick: This is familiar syntax that always works. Shouldn't have to explain function components on the install page.

content/docs/add-react-to-an-existing-app.md
alexkrolick: The  component scripts won't minify themselves 😄 
The production CDN links are minified copies of React & ReactDOM though

alexkrolick:
@gaearon I like using the existing HTML page as a starting point, good idea

What do you think about using a `details` tag to hide the code samples inline, but make them expandable? I worry that relying on external links makes it hard to maintain context.

<details>
<summary>Like this</summary>

<br />

**[Download examples](https://gist.github.com/gaearon/6668a1f6986742109c00a581ce704605)**

_index.html_

```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="UTF-8" />
    <title>Add React in One Minute</title>
  </head>
  <body>

    <h2>Add React in One Minute</h2>
    <p>This page demonstrates using React with no build tooling.</p>
    <p>React is loaded as a script tag.</p>

    <!-- We will put our React component inside this div. -->
    <div class="like_button_container"></div>

    <!-- Load React. -->
    <!-- Change .development.js to .production.min.js in both tags before deployment! -->
    <script src="https://unpkg.com/react@16/umd/react.development.js" crossorigin></script>
    <script src="https://unpkg.com/react-dom@16/umd/react-dom.development.js" crossorigin></script>

    <!-- Load our React component. -->
    <script src="like_button.js"></script>

  </body>
</html>
```

_like_button.js_

```js
'use strict';

class LikeButton extends React.Component {
  constructor(props) {
    super(props);
    this.state = { liked: false };
  }

  render() {
    if (this.state.liked) {
      return 'You liked this.';
    }

    return React.createElement(
      'button',
      { onClick: () => this.setState({ liked: true }) },
      'Like'
    );
  }
}

// Find the DOM container we defined in HTML.
let domContainer = document.querySelector('.like_button_container');

// Show the LikeButton component inside our DOM container.
ReactDOM.render(React.createElement(LikeButton), domContainer);
```

</details>

content/docs/add-react-to-a-website.md
```diff
@@ -0,0 +1,173 @@
...
+3. Run `npm install babel-cli@6 babel-preset-react-app@3`
+
+You can now use JSX!
+
+### Compile JSX with One Command
+
+**Make sure you've followed the previous steps.** Then create a folder called `src` and run this terminal command:
+
+```
+npx babel --watch src --out-dir . --presets react-app/prod
```
alexkrolick: screen shot 2018-06-25 at 11 19 23 am

screen shot 2018-06-25 at 11 20 22 am

gaearon: I suspect you missed the previous steps, as it mentions? :-)

gaearon:
Superseded by https://github.com/reactjs/reactjs.org/pull/996.
Thanks for starting this

PR #996, opened 2018-06-25
Rewrite Installation and related docs

gaearon:
Based off https://github.com/reactjs/reactjs.org/pull/992, I ended up rewriting much of it so I'll send as a separate PR. Something's wrong with Netlify build of that one too.


content/docs/add-react-to-a-website.md
```diff
@@ -0,0 +1,179 @@
...
+>Tip
+>
+>This code defines a React component called `LikeButton`. Don't worry if you don't understand it yet -- we'll cover the building blocks of React later in our [main concepts guide](/docs/hello-world.html) and a [hands-on tutorial](/tutorial/tutorial.html). For now, let's just get it showing on the screen!
+
+After the starter code, add two lines to the bottom of `like_button.js`:
+
+```js{3,4}
+// ... the starter code you pasted ...
+
+const domContainer = document.querySelector('.like_button_container');
```
sophiebits: kinda weird IMO to use a class name then only support one at this step

content/docs/add-react-to-a-website.md
```diff
@@ -0,0 +1,179 @@
...
+If you don't have a minification step for your scripts, [here's one way to set it up](https://gist.github.com/gaearon/42a2ffa41b8319948f9be4076286e1f3).
+
+## Optional: Try React with JSX
+
+In the examples above, we only relied on features that are natively supported by the browsers. This is why we used a JavaScript function call to tell React what to display:
+
+```js
+const e = React.createElement;
+
+// Display a "Like" button
```
sophiebits: can we say

> Display a "Like" \<button>

content/docs/add-react-to-a-website.md
```diff
@@ -0,0 +1,179 @@
...
+3. Run `npm install babel-cli@6 babel-preset-react-app@3`
+
+You can now use JSX!
+
+### Compile JSX with One Command
+
+Create create a folder called `src` and run this terminal command:
+
+```
+npx babel --watch src --out-dir . --presets react-app/prod
```
sophiebits: why `.`? feels kinda weird to me for the src and out trees to not be disjoint

content/docs/create-a-new-react-app.md
```diff
@@ -0,0 +1,83 @@
...
+
+Use an integrated toolchain for the best user and developer experience.
+
+This page describes a few popular React toolchains which help with tasks like:
+
+* Using third-party modules from npm.
+* Surfacing common mistakes during development.
+* Fast development with live-editing CSS and JS.
+* Optimizing the bundles for production.
+* Running unit tests.
```
sophiebits: maybe something about scaling to many files and many components easier?

content/docs/create-a-new-react-app.md
```diff
@@ -0,0 +1,83 @@
...
+
+### Gatsby
+
+**[Gatsby](https://www.gatsbyjs.org/)** is the best way to create **static websites** with React. It lets you use React components, but outputs pre-rendered HTML and CSS to guarantee the fastest load time.
+
+Learn Gatsby from [its official guide](https://www.gatsbyjs.org/docs/) and a [gallery of starter kits](https://www.gatsbyjs.org/docs/gatsby-starters/).
+
+### More Flexible Toolchains
+
+The following toolchains offer more flexiblity and choice. We recommend them to more experienced users:
```
sophiebits: do we really "recommend" all of these? I'm not familiar with all of them but would probably recommend most people turn to webpack+babel first, not sure if I'm skewed

content/docs/getting-started.md
```diff
@@ -0,0 +1,101 @@
...
+  - "download.html"
+  - "downloads.html"
+  - "docs/try-react.html"
+  - "docs/tooling-integration.html"
+  - "docs/package-management.html"
+  - "docs/language-tooling.html"
+  - "docs/environments.html"
+---
+
+This page is an overview of the React documentation and related resources.
```
sophiebits: maybe a note about what React is, linking back to the homepage? in case someone misses it

content/docs/getting-started.md
```diff
@@ -0,0 +1,101 @@
...
+- [Versioned Documentation](#versioned-documentation)
+- [Something Missing?](#something-missing)
+
+## Try React
+
+React is designed for gradual adoption, and you can use as little or as much React as you need. Whether you want to get a taste of React, add some interactivity to a simple HTML page, or start a complex React-powered app, the links in this section will help you get started.
+
+### Online Playgrounds
+
+If you're interested in playing around with React, you can use an online code playground. Try a Hello World template on [CodePen](codepen://hello-world) or [CodeSandbox](https://codesandbox.io/s/new). If you prefer to use your own text editor, you can also [download this HTML file](https://raw.githubusercontent.com/reactjs/reactjs.org/master/static/html/single-file-example.html), edit it, and open it from the local filesystem in your browser. It does a slow runtime code transformation, so don't use it in production.
```
sophiebits: I might walk back "don't use" slightly since there are a lot of toys that can be used like this – maybe like "so it's not the best choice when performance matters"? or "so we'd only recommend using this for simple demos"?

content/docs/getting-started.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Learn React
+
+People come to React from different backgrounds and with different learning styles. Whether you prefer a more theoretical or a more practical approach, we hope you'll find something useful in this section.
+
+Like any unfamiliar approach, React does have a learning curve. With practice and some patience, you will get the hang of it.
+
+### JavaScript Resources
+
+The React documentation assumes some familiarity with the JavaScript language. You don't have to be an expert, but it's harder to learn both React and JavaScript at the same time. We recommend going through [this JavaScript tutorial](https://developer.mozilla.org/en-US/docs/Web/JavaScript/A_re-introduction_to_JavaScript) to check your knowledge level. It will likely take you between 30 minutes and an hour but will help you feel more confident learning React.
```
sophiebits: should we say something about needing to know how to program? should be clear but…

content/docs/getting-started.md
```diff
@@ -0,0 +1,101 @@
...
+
+Whenever you get confused by something in JavaScript, [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript) and [javascript.info](http://javascript.info/) are great websites to check. There are also [community support forums](/community/support.html) where you can ask for help.
+
+### Main Concepts
```
sophiebits: You recommend people start here first? I always imagined most people should use the tutorial as their first code intro. My instinct would be to swap this with the tutorial. Also as someone familiar with the docs I had no idea where this would link before clicking it, not sure if that is a bad thing.

content/tutorial/tutorial.md
```diff
@@ -59,12 +59,18 @@ You can now skip the second setup option, and go to the [Overview](#overview) se
...
 ### Setup Option 2: Local Development Environment
 
-**This is completely optional and not required for this tutorial!**
+This is completely optional and not required for this tutorial!
+
+<br>
+
+<details>
+
+<summary><b>Optional: Click to expand the local development instructions</b></summary>
```
sophiebits: "local development instructions" is a bit vague. Something like this maybe:

"Optional: Instructions for following along locally using your preferred text editor"

content/docs/create-a-new-react-app.md
gaearon: Yeah they're all pretty good and do defaults right. Babel/webpack unfortunately don't—people always end up shipping development builds. Webpack 4 might be better on this front but I'd think twice before recommending it directly.

content/docs/add-react-to-a-website.md
gaearon: I didn't want to explain that we use an ID, and then in another example we use a class. But I now realized I already collapsed by other example from a section to a downloadable snippet. So I guess I can make this an ID.

content/docs/add-react-to-a-website.md
gaearon: niiice

content/docs/add-react-to-a-website.md
gaearon: Because then we don't have to mention that they need to update path in HTML and in Uglify command they learned in previous section. I figured people who would feel it's weird are the ones who'd guess to change the output directory, and the rest would break their setups.

content/docs/add-react-to-a-website.md
```diff
@@ -0,0 +1,179 @@
...
+
+>Note
+>
+>This strategy is mostly useful while React-powered parts of the page are isolated from each other. Inside React code, it's easier to use [component composition](/docs/components-and-props.html#composing-components) instead.
+
+### Tip: Minify JavaScript for Production
+
+Before deploying your website to production, be mindful that unminifed JavaScript can significantly slow down the page for your users.
+
+If you already minify the application scripts, all you need to do is to ensure that the deployed HTML loads the versions of React ending in `production.min.js`:
```
sophiebits: maybe ", your site will be production-ready if you ensure…"

content/docs/getting-started.md
gaearon: I'm kind of torn on this. Could do both ways. Let's try Tutorial first.

I'm a bit worried that Tutorial being a single page makes it seem like a bigger commitment.

PR #1000, opened 2018-06-26
Make componentDidUpdate work better. (No need compare props all the time)

TronNatthakorn:
This is my opinion that I think if not have this case it force everyone always compare props in componentDidUpdate.
That look strange. And I think this is better.


gaearon:
This will always fire if the prop is true, even if this specific prop hasn’t changed. This may not be the behavior you want—for example if you want to send a request when something *becomes* true.

TronNatthakorn:
@gaearon Thank for share this knowledge for me. I already understand you.

This my fault understand concept. It should not fire if props not change. :)

PR #1009, opened 2018-06-27
add note about eslint:no-label

swyxio:
closes https://github.com/reactjs/reactjs.org/issues/1004

### DIFF
--- a/content/docs/faq-state.md
+++ b/content/docs/faq-state.md
@@ -76,6 +76,18 @@ handleSomething() {
 }
 ```
 
+One common pitfall when writing arrow functions is accidentally relying on labels and implicit returns:
+
+```js
+// good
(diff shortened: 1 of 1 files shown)


content/docs/faq-state.md
```diff
@@ -76,6 +76,18 @@ handleSomething() {
 }
 ```
 
+One common pitfall when writing arrow functions is accidentally relying on labels and implicit returns:
```
alexkrolick: I would call this out as a 'rarely-used language feature called "labels"'

rickhanlonii:
Hey @swyxio, good to see you! Sorry we never reviewed this. Going to close since it's against the old docs site, but if you think this is still needed I think there are a few good places to mention this in the new docs. Will comment on the issue.

PR #1010, opened 2018-06-27
Add link to /render-props.html in Functions as Children section

saltycrane:


threepointone:
The text is talking about passing *any* kind of data structure to children, not just functions. We don't really recommend anything other react elements (and functions in some circumstances) anyway. I'm going to close this, but we'll probably revisit this in the future. Thank you very much for the PR!

PR #1011, opened 2018-06-27
Code of conduct

alocke12992:

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,64 @@
+# Facebook Open Source Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as
+contributors and maintainers pledge to making participation in our project and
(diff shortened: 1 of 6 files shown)


alocke12992:
Note that this issue exists on all of the fbOpensource projects. The "Code of Conduct" link is broken.

gaearon:
Thanks for reporting. We should fix the link.

zpao:
Just a heads up: code.facebook.com migrated today to code.fb.com. We lost a couple pages along the way. I'll be recreating the code of conduct on opensource.fb.com as a replacement shortly, will update here when that's up.

zpao:
Eh, I just made https://code.fb.com/codeofconduct/ instead. Takes the pressure off moving this into your own docs. I'll probably make that old link redirect to a different location at some point, it's just a bit more work right now.

gaearon:
Thanks for fixing the link @zpao!

PR #1016, opened 2018-06-28
Output Babel to directory

jamiebuilds:
It seems like a mistake to output all the files in `src` to the cwd, once they get setup they might realize they want to `.gitignore` the built files, it also could conflict with other files in the directory.

Also: New users might not realize what the `.` means, where if they see `dist` in both spots they could make the connection and customize it to their needs.


alexkrolick:
I _think_ we'll probably move to this if we can get it in good working order: https://github.com/reactjs/reactjs.org/pull/999

Issue #1017, opened 2018-06-28
Explanation for "mountNode" on home page code examples would help


craigdanj:
So I hit the Reactjs.org home page and while reading the code examples I ran into this variable called "mountNode".

Eg:

```js
ReactDOM.render(
  <HelloMessage name="Taylor" />,
  mountNode
);
```

Unfortunately the page offers no explanation for what this is or how to create that variable.

Since these are examples that beginners will usually read I think it might help to provide at least a one line explanation somewhere on the page. Or maybe at least show the variable being assigned a reference to the root element.

FYI: I later learnt what mountNode is but this is for the benefit of others trying to learn React.

**Example of someone else getting confused:**
https://stackoverflow.com/questions/23600443/uncaught-referenceerror-mountnode-is-not-defined

SeasonLeee:
Thank god, I am not alone. I have been using the search feature of the browser to look for the definition of this mysterious 'mountNode'. I couldn't get the intention of its usage by simply reading its name nor I couldn't even figure where on earth does it pop up. And the craziest thing is React have been in the market for at least 5 years and its popularity is so wide that you would be shamed if you never used it as a frontend developer. So, is the homepage of React have been like this all the time or just recently update and accidentally forgot to mention what this little thing for? Anyway, thanks for the issue fixing. Although as a great dummy, I still confused even there's an update of the definition of the variable. Yet, the homepage still remains the same...

craigdanj:
> So, is the homepage of React have been like this all the time or just recently update and accidentally forgot to mention what this little thing for?

Not sure really... just noticed it when I read it and figured it could be fixed.

Glen-Dango:
i had same question... iam now more confused by github...where is the answer to this question?

RadGH:
I am a react newbie trying to get the examples up and running on my own Laragon local server. I got it all set up like the examples but I get that mountNode error.

I've since found the answer but not without frustration. The top result on Google reads: "Make sure your mountNode variable exists.". Yes that is exactly the problem from the front page.

I eventually went to the "Get Started" page on the website and the online codepen playground also explains the HTML portion, which is not mentioned anywhere on the front page (what?!). 

I don't understand why this information is not explained in the very first piece of code you read when you decide to try React. It adds a very unnecessary and frustrating barrier to entry. And like in most programming, you spend a whole 10 minutes trying to find the answer and it's just a really simple one.

[...]

PR #1019, opened 2018-06-28
Comment to help user understand 'mountNode' on home page example.

craigdanj:
#1017 
This is to help users understand the mountNode variable on the home page of reactjs.org sicne currently it offers no explanation. This lack of an explanation might confuse users. 

eg: https://stackoverflow.com/questions/23600443/uncaught-referenceerror-mountnode-is-not-defined

### DIFF
--- a/content/home/examples/a-component-using-external-plugins.js
+++ b/content/home/examples/a-component-using-external-plugins.js
@@ -36,4 +36,7 @@ class MarkdownEditor extends React.Component {
   }
 }
 
-ReactDOM.render(<MarkdownEditor />, mountNode);
+ReactDOM.render(
+  <MarkdownEditor />,
+  document.getElementById('external-plugin-example')
+);
--- a/content/home/examples/a-simple-component.js
+++ b/content/home/examples/a-simple-component.js
@@ -8,7 +8,8 @@ class HelloMessage extends React.Component {
   }
 }
 
+
 ReactDOM.render(
   <HelloMessage name="Taylor" />,
-  mountNode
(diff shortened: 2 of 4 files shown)


alexkrolick:
This is a good change, but the formatting in the front page code blocks is very narrow. Can you look at the deploy preview and figure out how to wrap the comment so it fits?

Thanks!

alexkrolick:
Maybe we could pull the react-dom part into the home page examples plugin? Mountnode is defined there and needed to make the examples work with live edits.

craigdanj:
@gaearon I actually like your idea better than mine. I agree mine isn't the best solution.

Can we replace 'mountNode' with

document.getElementById('root')

In all the examples? Would that make sense?

gaearon:
I think it will be confusing to have `root` in several examples since IDs are supposed to be unique. But if you give each container a different ID (e.g. `simple-component`, `todo-app`, etc) then it will make sense.

craigdanj:
@gaearon I've updated the first example to use ``first-component``. I figured if I add ``simple-component`` it'll look like I'm rendering the helloWorld component to a simple component when it actually is itself the simple component. Let me know what you think.

craigdanj:
@gaearon Ok so I've updated all the examples to use getElementById() and have used ids that correspond to each example's title (like you suggested) since this makes more sense for the documentation.

Let me know if the naming is appropriate and if there is anything else I need to change since the main website has live code editing.

gaearon:
I did something similar in https://github.com/reactjs/reactjs.org/pull/1526. I couldn't merge this PR because it didn't actually work, but thanks for starting it!

Issue #1014, opened 2018-06-27
JSX syntax highlighting is messed up


gaearon:
Noticed this on https://reactjs.org/docs/add-react-to-a-website.html:

<img width="580" alt="screen shot 2018-06-28 at 00 32 51" src="https://user-images.githubusercontent.com/810438/42005126-db8e9cd4-7a6a-11e8-8b2d-c1d084e92289.png">

I don't know why. Help would be appreciated!

KhodorAmmar:
Looks like an issue in `gatsby-remark-prismjs`. Updating it from `1.2.1` to `1.2.24` does the trick

gaearon:
Thank you for fixing this, it was super annoying

PR #1026, opened 2018-07-01
Fix issue with JSX highlighting

KhodorAmmar:
This commit will fix issue #1014

I've updated `gatsby-remark-prismjs` to `v1.2.24`. This caused some conflicts with `gatsby-remark-embed-snippet`, so I've had to update that too `v1.0.22`

JSX syntax looks OK now: 

Previous: 

screen shot 2018-07-01 at 11 07 50

Current:

screen shot 2018-07-01 at 11 07 37


gaearon:
Thank you!

PR #1029, opened 2018-07-02
Add comments for calculateWinner()'s array

retrazil:
It's a little difficult to figure what exactly the numbers inside the array mean.


BartoszKlonowski:
Hello @retrazil!
In the updated version of that example the `calculateWinner` helper function is already briefly explained regarding what it does, so I have a feeling adding a comment in each line will only make it look chaotic at first glance. Especially that the updated docs already says that `Don't worry too much about the `calculateWinner` function; it's not specific to React`, so adding comments explaining every line would only drove an unwanted attention to that function.
Let me close this PR, but please feel free to create new PRs to other open issues.

PR #1030, opened 2018-07-02

silvaw:
Using WarningBanner with a capital W, causes IDE warns me that "Primitive value returned from constructor will be lost when called with 'new'".
IDE expect function names to start with a lowercase letter. Since WarningBanner function is called with a capital W, it thinks it's a class declaration.


alexkrolick:
JSX element names must be capitalized:

https://github.com/reactjs/reactjs.org/blob/master/content/docs/jsx-in-depth.md#user-defined-components-must-be-capitalized

Sounds like your linter needs to be changed to reflect this

PR #1032, opened 2018-07-03
Add signature to getSnapshotBeforeUpdate section

andrewiggins:
Add the signature for `getSnapshotBeforeUpdate` to its section, just like all other lifecycle method sections have.


gaearon:
Thanks!

PR #1038, opened 2018-07-05
docs: fix minor typo

ivandelabeldad:


gaearon:
I don't think it's a typo (both variants are acceptable, you could pronounce it as "el-i"). But this does seem more common!

PR #1041, opened 2018-07-06
Halogen link is missing

terrierscript:
I found http://madscript.com/halogen is missing.


saranshkataria:
This has already been fixed. So closing this out. Thanks

PR #1042, opened 2018-07-06
Small code comment to help identify React vs HTML

MendelBak:


threepointone:
I'm going to push back on this and reject simply because the lines before those blocks already say the same. Thank you for the PR!

PR #1046, opened 2018-07-08
Update conferences.md

vincentmayers:
added Connect tech to react conferences


Marcholio:
The conflicts should be resolved and then either merged or closed.

PR #1049, opened 2018-07-09
Clarify that object spread syntax is standard in ES2018

lazarljubenovic:


harish-sethuraman:
Seems like this is fixed now.
link to the section: https://reactjs.org/docs/optimizing-performance.html#the-power-of-not-mutating-data

Thanks for sending a PR. Sorry for the late turn around time.

PR #1053, opened 2018-07-11
dont write state updation methods in constructor

itaditya:
Advantages of this approach-

1. No need to keep state updation methods in state.
1. In modern js class syntax, we don't need constructor generally, but because of the current implementation, we have to define constructor method.

With the approach I propose, we can just do -
```jsx
class App extends React.Component {
  state = {
    theme: themes.light
  }
}
```
1. More in line to how most of us commonly process a component code, (define initial state in constructor, then all class methods and at last render method)
1, Only the state needed by other nested components has to be passed to provider. The App component can still keep it's internal state hidden.

### DIFF
--- a/examples/context/updating-nested-context-app.js
+++ b/examples/context/updating-nested-context-app.js
@@ -5,29 +5,31 @@ class App extends React.Component {
   constructor(props) {
     super(props);
 
-    this.toggleTheme = () => {
-      this.setState(state => ({
-        theme:
-          state.theme === themes.dark
-            ? themes.light
-            : themes.dark,
-      }));
-    };
-
-    // highlight-range{1-2,5}
+    // highlight-range{1-2}
     // State also contains the updater function so it will
     // be passed down into the context provider
     this.state = {
-      theme: themes.light,
-      toggleTheme: this.toggleTheme,
+      theme: themes.light
     };
   }
 
+  toggleTheme = () => {
+    this.setState(state => ({
+      theme:
+        state.theme === themes.dark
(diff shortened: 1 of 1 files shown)


itaditya:
I am not able to understand why Circle CI is failing. Please help

alexkrolick:
- CI is failing because the code is not formatted - there is a lint command in package.json to fix it
- The approach you are proposing is actually mentioned in [Caveats](https://reactjs.org/docs/context.html#caveats). The code will cause unnecessary re-renders because a new object is constructed each time the provider renders, and the consumers use object identity to check if they should re-render.

Would you like to replace this PR with a link to that section in the first example, so it's more discoverable?

itaditya:
hi @alexkrolick, thanks !! now I understand why ci was failing. 

Regarding your 2nd point, I'm not sure if I understand it correctly. Are you saying, that whenever the state of Parent component (containing Provider) will change, a new object will be created and passed as value to Provider? I am a beginner in react, so I don't how much will it hurt performance. 

> Would you like to replace this PR with a link to that section in the first example, so it's more discoverable?
Please guide me on this..

alexkrolick:
Yes, the Provider will cause all of the Consumers to rerender if it creates a new object each time. The consumers check the object identity to see if they should update.

I was suggesting you add a link to the Caveats section higher in the page, near the example you were going to update.

itaditya:
Okay, so should I close this PR?

threepointone:
I'm going to close this PR. Thanks for the effort!

PR #1059, opened 2018-07-14
Update 2018-06-07-you-probably-dont-need-derived-state.md

everdimension:
A single form of *argument* confused me here. The memoization func in question accepts multiple arguments.


gaearon:
Thanks

PR #1069, opened 2018-07-18
Refs and the DOM: clarify when ref callbacks are invoked

dminkovsky:
Re: https://github.com/facebook/react/issues/12183#issuecomment-364108874


dminkovsky:
Hi @alexkrolick, thanks for your review. Do you know why the build failed? Is it a false alarm?

alexkrolick:
It failed because of a build configuration issue with Netlify which should be fixed now

PR #1085, opened 2018-07-25
Add "Before You Use Context"

gaearon:
Seeing how often this comes up, let's just add it here


gaearon:
I'll merge but if anyone feels strongly about something, feel free to edit directly on master

Issue #1084, opened 2018-07-24
Indentation is messed up in all inline examples


gaearon:
Don't know when this happened but it looks pretty bad.

https://reactjs.org/docs/introducing-jsx.html

<img width="539" alt="screen shot 2018-07-25 at 1 46 02 am" src="https://user-images.githubusercontent.com/810438/43171854-81f65520-8fac-11e8-90e5-9ae00ac5bddf.png">

I see this everywhere. Did we update to a broken version of something?

sophiebits:
To anyone interested: reproing locally and bisecting might be a good way to start debugging this.

Widdershin:
I'll have a look 😄

gaearon:
(See the latest in https://github.com/reactjs/reactjs.org/pull/1086#issuecomment-407597030)

killynathan:
Hi, this is my first attempt at trying to contribute. I am a little confused as to whether this issue is resolved . If it isn't, I would love to take a stab at it.

KhodorAmmar:
Might this be related to the update I did in #1014 ? I don’t remember it caused any issues back then though

maddhruv:
@gaearon seems the issue is resolved now 
can be closed?

PR #1086, opened 2018-07-25
Fix example indentation

lauriejones:
Addresses #1084

Not sure why having margin set on the `<pre>` causes the weird indentation, but moving the margin to a padding in the wrapping element fixes it, and is my preferred spacing solution.

Alternatively adding `display: block;` to the `<pre>` tag styling also fixes the issue, but I am not sure if we always want the pre/code elements to be block. Are they ever displayed inline?

**Update**
Adding `display: block;` to the `<pre>` tag styling may be the better solution as the examples appear to not always be rendered inside the `.gatsby-highlight` class

### DIFF
--- a/src/prism-styles.js
+++ b/src/prism-styles.js
@@ -31,7 +31,6 @@ css.global('.gatsby-highlight', {
   borderRadius: 10,
   overflow: 'auto',
   tabSize: '1.5em',
-  padding: '1rem',
 });
 
 css.global(
@@ -40,7 +39,9 @@ css.global(
 .gatsby-highlight pre[class*="gatsby-code-"],
 .gatsby-highlight pre.prism-code`,
   {
+    display: 'block',
     height: 'auto !important',
+    margin: '1rem',
     fontSize: 14,
     lineHeight: '20px',
     whiteSpace: 'pre-wrap',
--- a/src/prism-styles.js
+++ b/src/prism-styles.js
@@ -31,6 +31,7 @@ css.global('.gatsby-highlight', {
   borderRadius: 10,
   overflow: 'auto',
   tabSize: '1.5em',
+  padding: '1rem',
 });
 
 css.global(
@@ -40,7 +41,6 @@ css.global(
 .gatsby-highlight pre.prism-code`,
   {
     height: 'auto !important',
-    margin: '1rem',
     fontSize: 14,
     lineHeight: '20px',
     whiteSpace: 'pre-wrap',


iagomelanias:
> Not sure why having margin set on the `<pre>` causes the weird indentation, but moving the margin to a padding in the wrapping element fixes it, and is my preferred spacing solution.

Besides that, the class `gatsby-code-jsx` is being used both in `pre` and `code` elements. Removing the duplicate class in `code` element also resolves this issue. 

You can check this by comparing the [current html version](https://reactjs.org/docs/introducing-jsx.html) with a [previous working version](https://web.archive.org/web/20180704215846/https://reactjs.org/docs/introducing-jsx.html).

**Current**:
screen shot 2018-07-24 at 21 31 06

**Previous version**:
screen shot 2018-07-24 at 21 29 35

I'm not sure how this bug was introduced tho.

gaearon:
Seems like this causes the highlighting to have margins. Can we avoid that?

screen shot 2018-07-25 at 2 32 29 am

Also, do we know what broke it?

gaearon:
>I'm not sure how this bug was introduced tho.

Probably a package upgrade but I don't know which one.
Could be: https://github.com/reactjs/reactjs.org/commit/d9cd3e311f1bae4639db18bc42ec219f3bf1b6a0, https://github.com/reactjs/reactjs.org/commit/45a72f9e71ed5b71100ede3067d633b571205775, https://github.com/reactjs/reactjs.org/commit/242dbdd7cc6b6cf7fdc24aa3a7f6215cd737406e.

Want to investigate?

iagomelanias:
Yes, i'm gonna take a look now. 🙂

gaearon:
I merged a quick fix in https://github.com/reactjs/reactjs.org/pull/1087 but would still appreciate more info if there was a bug in some update or if we just relied on implementation detail. Thanks for finding the cause!

DSchau:
@gaearon I think it's 26892808f4aaba7e1b0e58b63c87b4e3cdebf670 that caused the issue, specifically _probably_ updating gatsby-remark-prismjs.

Swapping to that exact commit demonstrates the issue, and is I believe the first commit to show the indentation regression.

gaearon:
Thanks for checking. Do you think the new behavior was intentional?

DSchau:
Can't answer that one! Pinning `gatsby-remark-prismjs` to `1.2.11` seems to be the root "fix," but I honestly think what you've already done in #1087 is probably just as good of a fix :)

gaearon:
Can you find the `gatsby-remark-prismjs` commit that broke this?

iagomelanias:
I believe it was this one https://github.com/gatsbyjs/gatsby/commit/6f565042b9c1f337356ef73eac0654411c8d19d9#diff-912e9bd2ec5f0fdf7da012659ea44444. By the commit name, it seems it's intentional.

PR #1093, opened 2018-07-26
Improve syntax colors

atomiks:
Hey, so I think the syntax colors of the code on the docs could be improved a bit. There are some problems with visual distinctions currently that makes the code a bit harder to read than it could be.

### Numbered issues

1. Operators and variables/plain text are not distinct enough
2. Numbers and functions are too visually similar
3. Keywords and props should be distinct
4. The line highlight should be lighter than the background in a dark theme (inverse of white, which would have a darker color as the highlight). The current dark highlight looks more like a well than a highlight. 
5. Expressions in props are colored red when they should be white (consistent with the child node below: `{props.value}`)

### Other

+ Increase line-height slightly to `1.6` instead of `20px`
+ General aesthetic changes that makes the code more vibrant and attractive (imo)
+ I also feel like there could be more granularity in the actual parser to allow for more color distinctions, but it's mostly okay I think.

### DIFF
--- a/src/prism-styles.js
+++ b/src/prism-styles.js
@@ -10,17 +10,18 @@ import {colors} from 'theme';
 
 const prismColors = {
   char: '#D8DEE9',
-  comment: '#999999',
-  keyword: '#c5a5c5',
-  lineHighlight: '#14161a',
-  primitive: '#5a9bcf',
-  string: '#8dc891',
+  comment: '#889dd0',
+  keyword: '#c8afff',
+  lineHighlight: '#2f354c',
+  primitive: '#f99e65',
+  string: '#84eabd',
   variable: '#d7deea',
-  boolean: '#ff8b50',
-  punctuation: '#5FB3B3',
-  tag: '#fc929e',
-  function: '#79b6f2',
-  className: '#FAC863',
+  boolean: '#ffa3d4',
+  punctuation: '#8bd7f0',
+  tag: '#ff94be',
+  function: '#82bdff',
+  className: '#ffe198',
+  attrName: '#ffdbaa',
   method: '#6699CC',
   operator: '#fc929e',
 };
@@ -33,6 +34,10 @@ css.global('.gatsby-highlight', {
   tabSize: '1.5em',
 });
 
+css.global('.token.script.language-javascript', {
+  color: colors.white,
+});
+
 css.global(
   `
 .gatsby-highlight > code[class*="gatsby-code-"],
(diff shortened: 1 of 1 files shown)


gaearon:
What do you think about trying Night Owl by @sdras?
https://marketplace.visualstudio.com/items?itemName=sdras.night-owl

Maybe we could use it.

AlmeroSteyn:
A great PR to score even better on accessibility for the docs as a whole. So my two cents:

1. Colour contrast of the text to background should be sufficient. There is a tool for that (https://developer.paciellogroup.com/resources/contrastanalyser/). If it passes on level AA then most user should be able to easily read it.
2. The font itself. A good test is that it shows clear difference between capital i and smaller case L, between zero and O and between rn and m.

Night Owl scores pretty good in this from what I can see. Only the comments could be missed by many due to the very low contrast so I'd bump that up and the ligatures could cause trouble for readers with dyslexia.

atomiks:
Night Owl looks beautiful, the problem is that the highlighter's grammar is not granular enough (e.g. can't differentiate between tags and components):

atomiks:
Alright, would you like me to change the PR to use Night Owl colors? (And also change the code background color -- which accidentally wasn't committed here)

Where is the parser located? The granularity is mostly fine except for that single problem which makes a big difference to the overall vibe imo. Is it possible to make a quick change to the parser to add a component class name (detecting if the first letter is uppercase?). Then we could rollout the change including the juicy tomato color 🍅

gaearon:
>Alright, would you like me to change the PR to use Night Owl colors? (And also change the code background color -- which accidentally wasn't committed here)

Let's do it in a separate PR so we can decide?

>Where is the parser located? The granularity is mostly fine except for that single problem which makes a big difference to the overall vibe imo. Is it possible to make a quick change to the parser to add a component class name (detecting if the first letter is uppercase?). Then we could rollout the change including the juicy tomato color 🍅

I'm not certain. We seem to use Prism for highlighting. So it probably comes from `gatsby-remark-prismjs` plugin. I see `prismjs` [here](https://github.com/reactjs/reactjs.org/blob/461394c3d80c8dc4a691091612cb2821fcc2453c/yarn.lock#L8301-L8305). Maybe we can contribute to this package. Or fork it and point lockfile to our fork using Yarn resolutions. Either would be fine to me.

atomiks:
Looks like a PR was made very recently: https://github.com/PrismJS/prism/pull/1519
```

rickhanlonii:
Sorry for the delay! Going to close this since the new docs code and sandboxes are completely different.

PR #1098, opened 2018-07-27
Update lifecycles.js - minor fix of example code

nateq314:
`theme ? 'dark' : 'light'` will always use the dark theme no matter what `theme` is set to in context. Guessing this is not what was intended? This PR fixes it.


alexkrolick:
Thanks!

PR #1100, opened 2018-07-28
[accessibility] Minor clarifications

st-sloth:
Removed trailing whitespaces in the file (well, IDE took care of that; though I would think that would be automated in the pre-commit or something).
Changed description of `aria-*` attributes form "lowercased" to "hypen-cased".
Clarified the use of React.Fragment - before it seemed like one MUST always use `<>...</>` if they don't need the `key` prop. Also added the links to the corresponding doc.

### DIFF
--- a/content/docs/accessibility.md
+++ b/content/docs/accessibility.md
@@ -30,7 +30,7 @@ Note that all `aria-*` HTML attributes are fully supported in JSX. Whereas most 
 
 ```javascript{3,4}
 <input
-  type="text" 
+  type="text"
   aria-label={labelText}
   aria-required="true"
   onChange={onchangeHandler}
@@ -49,7 +49,7 @@ Sometimes we break HTML semantics when we add `<div>` elements to our JSX to mak
 In these cases we should rather use React Fragments to group together multiple elements.
 
 Use `<Fragment>` when a `key` prop is required:
- 
+
 ```javascript{1,8,11}
 import React, { Fragment } from 'react';
 
@@ -72,14 +72,14 @@ Use `<></>` syntax everywhere else:
 
 ```javascript
 function ListItem({ item }) {
-  return ( 
+  return (
     <>
       <dt>{item.term}</dt>
       <dd>{item.description}</dd>
     </>
-  );    
+  );
 }
-``` 
+```
 
 ## Accessible Forms
 
@@ -108,7 +108,7 @@ Error situations need to be understood by all users. The following link shows us
 
 ## Focus Control
 
-Ensure that your web application can be fully operated with the keyboard only: 
+Ensure that your web application can be fully operated with the keyboard only:
 
 - [WebAIM talks about keyboard accessibility](http://webaim.org/techniques/keyboard/)
 
@@ -124,7 +124,7 @@ Only ever use CSS that removes this outline, for example by setting `outline: 0`
 
 Provide a mechanism to allow users to skip past navigation sections in your application as this assists and speeds up keyboard navigation.
 
-Skiplinks or Skip Navigation Links are hidden navigation links that only become visible when keyboard users interact with the page. They are very easy to implement with 
+Skiplinks or Skip Navigation Links are hidden navigation links that only become visible when keyboard users interact with the page. They are very easy to implement with
 internal page anchors and some styling:
 
 - [WebAIM - Skip Navigation Links](http://webaim.org/techniques/skipnav/)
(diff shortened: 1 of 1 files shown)


content/docs/accessibility.md
```diff
@@ -46,18 +46,28 @@ in our websites will often give us accessibility for free.
...
+  );
+}
+
 function Glossary(props) {
   return (
     <dl>
       {props.items.map(item => (
-        // Without the `key`, React will fire a key warning
+        // When mapping a collection, the `key` prop is required.
+        // Without it, React will fire a key warning
```
AlmeroSteyn: This duplicates information available elsewhere. I think it could be more terse:

Something like: "Fragments also require a `key` prop when mapping collections.

content/docs/accessibility.md
```diff
@@ -68,18 +78,20 @@ function Glossary(props) {
 }
 ```
 
-Use `<></>` syntax everywhere else:
+You can also use the [short syntax](/docs/fragments.html#short-syntax), if your tooling supports it, but only if you don't need any props, like the `key`:
```
AlmeroSteyn: The point is great. Just some bulky language ([comma] if [comma] but only if).

Would suggest something like:
"When you don't need any props on the Fragment tag you can also use the [short syntax](/docs/fragments.html#short-syntax), if your tooling supports it."

It should already be clear from the example above that `key` is one of the props you could use on `Fragment`

content/docs/accessibility.md
st-sloth: Now that you mentioned it, I think it's not specifically Fragments that require a `key` prop but any component or element that is produced from mapping a collection.

Do I understand this right?

If so, what do you think of "Just like other components Fragments should have a `key` prop when mapping a collection"?

content/docs/accessibility.md
st-sloth: That's much better, thanks!

content/docs/accessibility.md
AlmeroSteyn: Yes, there is indeed a `key` prop required on anything that is mapped for the React Reconciler to know when the collection change and not accidentally mess up the displayed sequence or omit elements.

You could say something like that, although the key is required on more than components. Also JSX elements, should they be mapped.

st-sloth: I abstained from "anything" after I checked that mapping plain strings is okay without a `key` (well, there is nowhere one can put it either :) ). So it's anything that ends up being a `React.createElement()` result, right?

While we're polishing this sentence before I make the commit, which is acceptable?

1. "Just like other components and elements, Fragments should have a key prop when mapping a collection"
2. "Just like other JSX elements, Fragments should have a key prop when mapping a collection" (which doesn't cover render functions without JSX)
3. "Just like other React-elements, Fragments should have a key prop when mapping a collection"

?

As for the first part of these options, I feel that it would be better to make a connection in a reader's mind with the more common requirement of `key`, that they may or may not know of.

AlmeroSteyn: I hate to be opinionated, LOL. but the best I can come up with is "Fragments also require a `key` prop when mapping collections" as that then does not need any clarification of what other elements required the key prop. Does not mean you have to use this verbatim, just something that does not get us in the mess of needing to explain something we explained ourselves into in the first place.

I also think that re-explaining keys is not required here as this is in the Advanced Guide section so we can assume that a the reader has already covered this basic concept.

st-sloth:
@AlmeroSteyn updated as requested.

AlmeroSteyn:
LGTM. Thanks for your patience and for contributing to the a11y docs!

jsfeldman:
I'm completely new to react so it's definitely possible I don't know what I'm talking about, but as I was reading through the documentation I found this part a little confusing.

Since `ListItem` is part of the same code snippet as `Glossary` I expected the example to read something like this:
```
import React, { Fragment } from 'react';

function ListItem({ item }) {
  return (
    <Fragment>
      <dt>{item.term}</dt>
      <dd>{item.description}</dd>
    </Fragment>
  );
}

function Glossary(props) {
  return (
    <dl>
      {props.items.map(item => (
        // Fragments should also have a `key` prop when mapping collections
        <ListItem item={item} key={item.id} />
      ))}
    </dl>
  );
}
```
Right now, the one code snippet is showing two different ways to use React fragments that are unrelated to each other, I think. Although, maybe I'm still misunderstanding it.

I think it might be useful to add `ListItem` from the first example to the second example and have a comment that says they're equivalent. I think that might be more in line with how the rest of the documentation is written.

I realize this section of the documentation is more about accessibility than how to write Fragments, but I thought I'd just throw this out there in case other people are having the same confusion. Not a big deal though. It could just be me.

AlmeroSteyn:
@jsfeldman These docs are there for people who are new to React, so don't sell yourself short. Great to get feedback from someone who reads it completely cold.

The purpose of the second example is to illustrate that the `<Fragment>` tag can also takes the `key`  prop when mapping collections, so to put the `<ListItem>` in there would lose that bit of information. I think it is good to keep that intact as it may not be immediately obvious that you can directly map with `<Fragment>` as container.

However, I think you are right that the example could be confusing due to the collocation of the two components and then they have no relation to each other. So perhaps the solution would be to split up the two into separate code blocks with that comment now describing the second code block?

You place this comment on a closed issue so I saw it by luck as I was still subscribed to it. Better to make an actual issue or even far better how about making a PR with this change? Jump right in and improve the docs! What do you say?

jsfeldman:
@AlmeroSteyn I see what you mean about the key prop. I'll take a stab at it when I get home and submit a PR! And I'll submit an issue or PR next time I run into something.
Thanks!

PR #1103, opened 2018-07-30
Add false to conditional-rendering

donavon:
Conditional rendering can return `null` or `false`. Add the `false` condition and show an example that benefits from using `false` over `null`.

Here's the current example:
```js
function WarningBanner(props) {
  if (!props.warn) {
    return null;
  }

  return (
    <div className="warning">
      Warning!
    </div>
  );
}
```

But by returning `false`, you can do this:
```javascript
const WarningBanner = ({ warn }) => (
  warn && (
    <div className="warning">
      Warning!
    </div>
  )
);
```

### DIFF
--- a/content/docs/conditional-rendering.md
+++ b/content/docs/conditional-rendering.md
@@ -190,7 +190,7 @@ Just like in JavaScript, it is up to you to choose an appropriate style based on
 
 ### Preventing Component from Rendering
 
-In rare cases you might want a component to hide itself even though it was rendered by another component. To do this return `null` instead of its render output.
+In rare cases you might want a component to hide itself even though it was rendered by another component. To do this return `null` or `false` instead of its render output.
 
 In the example below, the `<WarningBanner />` is rendered depending on the value of the prop called `warn`. If the value of the prop is `false`, then the component does not render:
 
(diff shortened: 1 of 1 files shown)


donavon:
FYI... Direct link to changed docs:

https://deploy-preview-1103--reactjs.netlify.com/docs/conditional-rendering.html#preventing-component-from-rendering

content/docs/conditional-rendering.md
```diff
@@ -238,6 +238,18 @@ ReactDOM.render(
 );
 ```
 
+In this refactored version of `<WarningBanner />` (below), you can see the benefits of returning `false` over `null`. Here we use a logical AND to return either `false` or the rendered warning.
```
alexkrolick: Would rather explain the benefit than leave it as an exercise to the reader. I assume you are saying the `&&` is shorter and inline-able?

donavon: Any better?
> In this refactored version of `<WarningBanner />` (below), the benefits of returning `false` over `null` is that `false` is the natural result of `props.warn` – i.e. no additional values need to be returned after the condition is determined. Here we are using a logical AND to return either `false` or the rendered warning, or to put it in plain English: "return `false` if the `warn` prop is `false`, otherwise render the warning".

I'd also like to link to [this article](https://medium.freecodecamp.org/conditional-rendering-in-react-using-ternaries-and-logical-and-7807f53b6935) to explain it more, but not sure about Facebook's policy on this.

BartoszKlonowski:
Hello @donavon, thank you for the contribution!
From what I can see the example with `WarningBanner` is already removed from the docs, also the conditional rendering now is described with it's own separate section, in my opinion efficiently enough:
https://github.com/reactjs/react.dev/blob/e33ec30515561d715e35b2c437402c5c58d4ecd8/src/content/learn/conditional-rendering.md?plain=1#L263
Please, let me close this PR as already outdated, but we welcome more contributions on other still open issues.

PR #1104, opened 2018-08-01
Upgrade to Gatsby v2

alexandernanberg:
Went through the site and followed the current [migration guide](https://github.com/gatsbyjs/gatsby/blob/master/docs/docs/migrating-from-v1-to-v2.md). 

In my local testing everything seems to be working, only problem I encountered was that the `/blog/` redirect stopped working for some reason, looks like it's an issue inside Gatsby but not 100% sure on that one.

We might want to wait until Gatsby releases a RC or a stable release before merging this.


gaearon:
There seems to be a difference in font sizes with https://reactjs.org/, at least for the homepage

alexandernanberg:
Good catch, `gatsby develop` produces the right font sizes but `gatsby build` does not. I'll continue investigating!

sw-yx:
cc @kkemple and @jlengstorf, good candidate for pairing here

jlengstorf:
Hey, @alexandernanberg! If you'd like to set up a pair programming session to work through any of the migration, please let me know. I'm happy to jump on a screen share and dig through any tricky bits. (This also helps us improve the migration guide.)

Thanks!

alexandernanberg:
@jlengstorf I'd be down for that! Although I don't think there is really that much left, but as you said there were/are some tricky bits that I might need some guidance in

alexandernanberg:
@jlengstorf Yes, booked one that works for me!

octalmage:
I believe the font-size issue is related to https://github.com/gatsbyjs/gatsby/pull/6760, so I'm curious what you find.

alexandernanberg:
@octalmage Yep that's correct, although in this case it was possible to solve in another way. We were using `glamor/reset` (basically just a mirror of normalize.css) and by switching to use `normalize.css` directly the problem was resolved. Apparently you can easily run into problems when mixing different ways to do styling (css-in-js, css-modules etc)

octalmage:
@alexandernanberg thanks for sharing! That makes sense, css is built in to Gatsby so it gets loaded first. Switching to normalize.css forces it to the top.

gaearon:
Let me know when you think it’s ready and we can cut over!

alexandernanberg:
Will do! The redirect and style issues have been solved, but there is a IE regression/bug that is currently blocking (https://github.com/gatsbyjs/gatsby/issues/7003). When that gets fixed we should be good to go!

alexandernanberg:
I don't fully understand what this is for or how it should work, on the current site only this `/docs/error-decoder.html` path works, not `/docs/error-decoder` etc. I'm missing something?

https://github.com/reactjs/reactjs.org/blob/156e6f0e675aad7bc9016ee2b004603caadb4d61/gatsby/onCreatePage.js#L13-L20

gaearon:
No idea, sorry.

jlengstorf:
@alexandernanberg @gaearon I think this is a custom thing to allow for truncated errors in production, right?

E.g. https://reactjs.org/docs/error-decoder.html?invariant=42&args

I don't know how it works, but I think that's why it's set up the way it's set up.

alexandernanberg:
@jlengstorf Sorry I should have been more specific, I knew about the truncated errors bit. But since the error info is passed as a query param the the code doesn't really affect anything right? Seems to be working if I comment it out. Anyway, probably best to keep it just to be safe. 

To migrate this to v2 I would replace  `page.matchPath = 'docs/error-decoder:path?';` with `page.matchPath = 'docs/error-decoder/*';`, correct?

The [example](https://github.com/gatsbyjs/gatsby/blob/master/docs/docs/migrating-from-v1-to-v2.md#change-client-paths-to-use-a-splat) in the migration guide has a `/` before the `:path`, but the code here does not. Might not make any difference but I just wanted to be sure!

jlengstorf:
@alexandernanberg I don't think you'd want to add a slash in this case. We swapped out Express-style path wildcards (e.g. `:path`) for splats (`*`) due to the Reach Router migration — the slash would break those URLs in this case, I think.

alexandernanberg:
@jlengstorf Okay, so how would we keep the current behavior? Just leave it as it is?

jlengstorf:
@alexandernanberg I'm not 100% because I don't know how to test it, but I would try:

```diff
  // page.matchPath is a special key that's used for matching pages only on the client. 
  // Explicitly wire up all error code wildcard matches to redirect to the error code page. 
  if (page.path.includes('docs/error-decoder.html')) { 
-   page.matchPath = 'docs/error-decoder:path?';
+   page.matchPath = 'docs/error-decoder*?'; 
    page.context.slug = 'docs/error-decoder.html'; 
  
    createPage(page); 
  } 
```

jlengstorf:
As far as I can tell with a pretty quick look, the named path is _not_ used in the `ErrorDecoder` component or at any point in the component chain where it's used, so there shouldn't be any issues with swapping out for the splat as shown in the diff above.

alexandernanberg:
Looks like it's been there since the migration to Gatsby v1. @bvaughn looks like you were the one to add that, can you provide some context?

bvaughn:
> @bvaughn looks like you were the one to add that, can you provide some context?

Nope, sorry. 😄 If it isn't still necessary, maybe it was at some earlier time? Gatsby changed a lot since we began using it. I don't remember anything specific about these few lines from 10+ months ago though.

alexandernanberg:
@bvaughn Okay, thanks anyway 😄 

Removed that logic and looks like everything is working now. Would love for other people to try the site to verify this. The Netlify link can be found in the top of the comments!

gaearon:
I noticed two small things:

* The all-caps section headers in right side nav seem to look a bit different (font size? spacing?). I kinda like the new ones better.
* When you go to Getting Started and click on any item in TOC, you'll notice we have something that adjusts the anchor scroll top after the navigation so that the top nav wouldn't obscure the header. The new website doesn't seem to have that logic (or it doesn't work). I think we'll want to have this.

Other than that, looks good to me

alexandernanberg:
Yeah, looks like there were a problem with specificity on the old site causing the `font-family` to be set to `sans-serif` instead of the system font stack. But that's fixed now since we are using `normalize.css` directly, hence why it looks different.

I've created a [PR](https://github.com/gatsbyjs/gatsby/pull/7299) for the other problem you described. Was an issue with links that wasn't using the `Link` component from `gatsby`.

gaearon:
Is this ready to go in?

alexandernanberg:
Nope, not yet unfortunately. The bug with the links in the TOC is not solved (waiting on an upstream fix)

alexandernanberg:
@gaearon I think this should be ready now!

gaearon:
Looks legit. Let's give it a go! Thank you ❤️

bvaughn:
😮 ❤️

KyleAMathews:
Yahoo! Thanks @alexandernanberg for your help here and fixes to Gatsby!

gaearon:
Can you please look at this? https://github.com/reactjs/reactjs.org/issues/1201

PR #1109, opened 2018-08-06
Remove usage of class fields in blogpost

alexandernanberg:
Saw that the blog post was using class fields which are discouraged by the [guidelines](https://github.com/reactjs/reactjs.org/blob/master/CONTRIBUTING.md#dont-use-features-that-arent-standardized-yet).

If this was intended please close this PR 😄


gaearon:
Event handlers are also using them. To be honest in this case I'm okay with letting it slide. It's getting pretty noisy otherwise and distracts from the point of the post.

PR #1112, opened 2018-08-09
Add caveat to React.createContext docs

rtsao:
Addresses https://github.com/facebook/react/issues/13346

### DIFF
--- a/content/docs/context.md
+++ b/content/docs/context.md
@@ -227,6 +227,8 @@ One issue with the render prop API is that refs don't automatically get passed t
 
 ## Caveats
 
+### Value reference equality
+
 Because context uses reference identity to determine when to re-render, there are some gotchas that could trigger unintentional renders in consumers when a provider's parent re-renders. For example, the code below will re-render all consumers every time the Provider re-renders because a new object is always created for `value`:
 
 `embed:context/reference-caveats-problem.js`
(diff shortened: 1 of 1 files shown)


jeffhandley:
I have found the `peerDependencies` isn't enough of a solution. Even with peer dependencies, my packages tend to have copies of those peer dependencies within the `node_modules` folder to satisfy `devDependencies` of the same packages so that unit tests can succeed. Inevitably, the singleton issue still surfaces.

Following an approach similar to what [this post describes](https://derickbailey.com/2016/03/09/creating-a-true-singleton-in-node-js-with-es6-symbols/), I've now set up my context module to use a `Symbol` and store the singleton instance on `global` using that `Symbol`, and ensuring that no second instance is ever created.

Here's an example of a module that I am using that provides a context object:

``` jsx
import React from 'react';

const CONTEXT_KEY = Symbol.for('@jeffhandley/ui-context/Context');
const globalSymbols = Object.getOwnPropertySymbols(global);
const hasContext = (globalSymbols.indexOf(CONTEXT_KEY) > -1);

if (!hasContext) {
    global[CONTEXT_KEY] = React.createContext();
}

export default global[CONTEXT_KEY];
```

This approach has allowed me to end up with multiple copies of my `ui-context` package across my project's monorepo (and multiple repos), without encountering any issues of the context `Provider` and `Consumer` not being linked.

I'm curious if others have found they've also needed to use `global` with a `Symbol` to get a true singleton of their React Context module.

jeffhandley:
I noticed this PR is still sitting open. It's surely quite stale at this point. Should it get closed out?

rickhanlonii:
Closing since this PR is for the old docs, but if you think this is still needed feel free to create a PR to add a caveat to the [createContext](https://react.dev/reference/react/createContext) docs.

PR #1127, opened 2018-08-16
Add info to Children.toArray docs about invalid elements

AEgan:
This adds to the `React.Children.toArray` documentation to mention that it will filter out invalid elements.

I needed to filter out invalid child elements in an application I am working on. When searching for the best approach, I came across an old github issue https://github.com/facebook/react/issues/2956, and found that the discussion led to the creation of `React.Children.toArray`. I didn't realize that `toArray` would do this, even after reading the docs for it. I thought it would be helpful to add, so I decided to open a PR. Feel free to suggest any changes to the wording, or close if you think it's not necessary.

### DIFF
--- a/content/docs/reference-react.md
+++ b/content/docs/reference-react.md
@@ -198,7 +198,7 @@ Verifies that `children` has only one child (a React element) and returns it. Ot
 React.Children.toArray(children)
 ```
 
-Returns the `children` opaque data structure as a flat array with keys assigned to each child. Useful if you want to manipulate collections of children in your render methods, especially if you want to reorder or slice `this.props.children` before passing it down.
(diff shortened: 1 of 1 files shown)


kumarharsh:
+1 - Please merge this in. It takes an inordinate amount of time in parsing github issues and test cases to figure out that `toArray()` does indeed filter out `undefined`, `null`, etc (https://github.com/facebook/react/blob/4221565e1552425e9416035b5a4bf0ff567d18be/packages/react/src/__tests__/ReactChildren-test.js#L814)

BartoszKlonowski:
Hello @AEgan, thank you for taking the time to contribute!
Would you still like to proceed with this PR? It would require to be rewritten as the sentence modified in this PR no longer exists in the latest docs. Otherwise I won't be able to review it.

AEgan:
@BartoszKlonowski I completely forgot about this! 

looks like the caveat is included in the updated docs
- https://github.com/reactjs/react.dev/blob/main/src/content/reference/react/Children.md#caveats-children-toarray-caveats
- https://react.dev/reference/react/Children#children-toarray-caveats

Thanks for the reminder, I'll close out this PR

PR #1129, opened 2018-08-18
Use const for ref in functional component docs

andys8:
This PR changes the example to use `const` over `let` for a reference.

```js
let textInput = React.createRef();
```

### Cause

It's good practice to go with `const` over `let` by default if there is no intention to modify the value. In this specific example it is important to clarify that the created reference will change its value (accessed with `.current`), but the javascript reference of `textInput` won't be changed. The usage of `const` will help the reader understanding the expected behavior.

### Demo

Example still working properly.
https://codesandbox.io/s/0qozvzx3zp?module=%2Fsrc%2FComp.js


threepointone:
Thanks! will merge once CI passes.

PR #1142, opened 2018-08-27
added abort controllers in faq-ajax.md documentation

adeelibr:
This PR is related to https://github.com/reactjs/reactjs.org/issues/1068

### DIFF
--- a/content/docs/faq-ajax.md
+++ b/content/docs/faq-ajax.md
@@ -38,10 +38,18 @@ class MyComponent extends React.Component {
       isLoaded: false,
       items: []
     };
+    // Note: If the component unmounts before the AJAX call is complete
+    // you may see a error like 'Can’t call setState (or forceUpdate) 
+    // on an unmounted component.' To avoid this, we can abort any XHR request
+    // in componentWillUnmount lifecycle
+    // Read more on abort controller in AJAX calls
+    // https://developer.mozilla.org/en-US/docs/Web/API/AbortController/abort 
+    this.abortController = new window.AbortController();
+    this.mySignal = this.abortController.signal;
   }
 
   componentDidMount() {
-    fetch("https://api.example.com/items")
+    fetch("https://api.example.com/items", { signal: this.mySignal })
       .then(res => res.json())
       .then(
         (result) => {
@@ -54,12 +62,22 @@ class MyComponent extends React.Component {
         // instead of a catch() block so that we don't swallow
         // exceptions from actual bugs in components.
         (error) => {
+          if (error.name === 'AbortError') {
+            // If the AJAX request was aborted you can check that here
+            console.log('Fetch aborted:', error.message);
+          }
           this.setState({
             isLoaded: true,
             error
           });
         }
       )
+  }
+
+  componentWillUnmount() {
+    // When the component unmounts abort the AJAX request, to avoid
+    // unnecessary memory leaks in your application.
+    this.abortController.abort();
   }
 
   render() {


adeelibr:
@alexkrolick what is the process for the PR review in this project? :)

content/docs/faq-ajax.md
```diff
@@ -38,10 +38,18 @@ class MyComponent extends React.Component {
...
+    // in componentWillUnmount lifecycle
+    // Read more on abort controller in AJAX calls
+    // https://developer.mozilla.org/en-US/docs/Web/API/AbortController/abort 
+    this.abortController = new window.AbortController();
+    this.mySignal = this.abortController.signal;
   }
 
   componentDidMount() {
-    fetch("https://api.example.com/items")
+    fetch("https://api.example.com/items", { signal: this.mySignal })
```
TryingToImprove: What is the reason for not doing

```js
fetch("https://api.example.com/items", { signal: this.abortController.signal })
```

TryingToImprove:
I like the idea of explaining how to abort a AJAX-call with React, but I kind of feel like it is more "expert"-knowledge than basic.

For many will the `fetch` API still be new to them, and learning how to abort the request have only just recent been deployed to the browsers, whereas `fetch` have been for much longer.

https://caniuse.com/#search=fetch and https://caniuse.com/#search=abortcontroller

adeelibr:
> I like the idea of explaining how to abort a AJAX-call with React, but I kind of feel like it is more "expert"-knowledge than basic.

@TryingToImprove This is the only way to cancel XHR subscriptions and avoid setState memory leaks. This should be enforced when making an XHR request to safely cancel your network requests. Otherwise you will keep on getting the `setState warning error`.

TryingToImprove:
@adeelibr As far as I understand the code, the warning will still occur, since in the `catch` there are still a call to `setState` after the logging.

According to https://reactjs.org/blog/2015/12/16/ismounted-antipattern.html a developer could handle the mounted state by having a property `isMounted` and checking it before calling `setState`.

adeelibr:
> @adeelibr As far as I understand the code, the warning will still occur, since in the catch there are still a call to setState after the logging.

Yes you are right I have updated the example code, for this as well.

> According to https://reactjs.org/blog/2015/12/16/ismounted-antipattern.html a developer could handle the mounted state by having a property isMounted and checking it before calling setState.

yes, you can do it with isMounted, but there is still a network call happening that doesn't know that it is needed anymore. you are not properly handling the error that way just hiding it. Kind of.

I wrote an article about isMounted approach as well https://medium.freecodecamp.org/how-to-work-with-react-the-right-way-to-avoid-some-common-pitfalls-fc9eb5e34d9e but AbortControllers are the proper way to go and I strongly believe that people should be aware about this.

alexkrolick:
It's probably good to show both approaches, since AbortController isn't available in all browsers https://caniuse.com/#feat=abortcontroller and AJAX libraries have different API - some even include cancellation.

adeelibr:
> It's probably good to show both approaches, since AbortController isn't available in all browsers https://caniuse.com/#feat=abortcontroller and AJAX libraries have different APIs - some even include cancellation.

Would that be okay, If i can link my article in the React documentation. As a reference how you can use `axios` AbortController in react.

Regarding the part where it isn't available on all browsers, we can use pollyfills, I like this one https://www.npmjs.com/package/abortcontroller-polyfill We can mention in the docs, about using a pollyfill to handle it in browsers that don't yet support it.

Your thoughts? @alexkrolick

BartoszKlonowski:
Closing as per last comment in the issue.
@adeelibr Thank you for the contribution! Still, please feel free to open new PRs to other open issues.

PR #1148, opened 2018-08-31
Patch the DOM to fix Google Translate

gaearon:
I think it's unacceptable to have Google Translate broken on our website, especially considering we don't yet offer any official translations. See https://github.com/facebook/react/issues/12460.

So I'm adding a hack described in https://github.com/facebook/react/issues/11538#issuecomment-417504600. I'm not too worried about performance since this is mostly a static website anyway.

Before the hack:

After the hack:

### DIFF
--- a/src/layouts/index.js
+++ b/src/layouts/index.js
@@ -10,6 +10,7 @@ import 'array-from';
 import 'string.prototype.includes';
 import 'string.prototype.repeat';
 
+import patchDOMForGoogleTranslate from 'utils/patchDOMForGoogleTranslate';
 import React, {Component} from 'react';
 import Flex from 'components/Flex';
 import Footer from 'components/LayoutFooter';
@@ -21,6 +22,8 @@ import '../prism-styles';
 import 'glamor/reset';
 import 'css/reset.css';
 import 'css/algolia.css';
+
+patchDOMForGoogleTranslate();
 
 type Props = {
   children: Function,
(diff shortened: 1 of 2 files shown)


gaearon:
cc @kylemathews dunno if you get questions about this but you might find it useful.. just don't make it a default.

gaearon:
cc @kyleamathews dunno if you get questions about this but you might find it useful.. just don't make it a default.

KyleAMathews:
Yeah, there's been one issue about it https://github.com/gatsbyjs/gatsby/issues/6300

Pointed people there here.

gaearon:
I think https://github.com/gatsbyjs/gatsby/issues/6300 is different and is about https://github.com/reactjs/reactjs.org/issues/467. We need to fix it too. Can you help? I’m not sure where to look.

KyleAMathews:
@gaearon yeah — I'm pretty sure we could support it when some other tool mutates the URL assuming the patterns are consistent. I'm tied up currently getting Gatsby v2 out but I'd be happy to support someone who wants to work on it.

ivanm696:
> Не удалось развернуть предварительный просмотр для _ReactJS_ .
> 
> Создано с коммитом [d72a2c7](https://github.com/reactjs/react.dev/commit/d72a2c7df862cc28b32c7de638d63d69c2809b68)
> 
> https://app.netlify.com/sites/reactjs/deploys/5b888aa682d3f124a5d5bd7e

🤖

PR #1154, opened 2018-09-04
Update conditional-rendering.md to utilize function within setState()

tom-wagner:
This code seems much cleaner to me and is a great way to reinforce the under-utilized approach of passing a function to setState(), which was introduced in the previous section: https://reactjs.org/docs/handling-events.html


tom-wagner:
Here is the link to the specific portion of the updated documentation for reference:

https://deploy-preview-1154--reactjs.netlify.com/docs/conditional-rendering.html#element-variables

tom-wagner:
What's the hold up?

threepointone:
I'm going to push back on this since it's arguable that this approach is better. But thank you for the PR!

PR #1155, opened 2018-09-04
Change name of setState updater first arg to 'state'

mmcgahan:
Followup from tweet about `setState` updater docs, with support from @gaearon:

> Possible confusion in React docs for `setState` updater pattern - calling first arg `prevState` seems odd because it could actually be a future value of state relative to `this.state`. Should it just be called `state` and align with `getDerivedStateFromProps` args?

https://twitter.com/mikemcgahan/status/1037074327269568512

The reference to 'previous state' is ambiguous in the 'updater' context, so this PR changes the suggested `prevState` arg name to `state`. This change helps conceptually align the 'updater' function with the behavior of `getDerivedStateFromProps` - both are pure functions that take a 'current' `state` arg and return `partialState` updates.

### DIFF
--- a/content/docs/conditional-rendering.md
+++ b/content/docs/conditional-rendering.md
@@ -215,8 +215,8 @@ class Page extends React.Component {
   }
 
   handleToggleClick() {
-    this.setState(prevState => ({
-      showWarning: !prevState.showWarning
+    this.setState(state => ({
+      showWarning: !state.showWarning
     }));
   }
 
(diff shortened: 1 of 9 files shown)


karlli1976:
It's glad to know I'm not alone :)

Actually I found the code on [Try It On CodePen](https://codepen.io/gaearon/pen/Xjoqwm?editors=0010) still uses `prevState`, which causes me a bit confused: according to the context, shouldn't it be `state`?

And I found if I change `prevState` to `prevMyState`, the code still works.

```
this.setState(prevMyState => ({
      showWarning: !prevMyState.showWarning
    }));
```

mmcgahan:
@karlli1976 yeah, you can technically name it whatever you want - it won't break the code, but it may mislead people who read `prevState` as referring to some sort of cached or 'past' state. I can't update the CodePen code, but it's probably not critical

PR #1156, opened 2018-09-05
Update to the semantic HTML section of the a11y page

jsfeldman:
I moved things around slightly in the semantic HTML section to make it flow a bit better as discussed [here](https://github.com/reactjs/reactjs.org/pull/1100#issuecomment-418185340), while still trying to keep it concise.
Feel free to change the wording as you please.

### DIFF
--- a/content/docs/accessibility.md
+++ b/content/docs/accessibility.md
@@ -65,27 +65,15 @@ function ListItem({ item }) {
 function Glossary(props) {
   return (
     <dl>
-      <ListItem item={term: 'Fragment', description: 'A piece of something.' }>
-      <ListItem item={term: 'React', description: 'Respond.' }>
+      {props.items.map(item => (
+        <ListItem item={item} key={item.id} />
+      ))}
     </dl>
   );
 }
 ```
 
-When you don't need any props on the Fragment tag you can also use the [short syntax](/docs/fragments.html#short-syntax), if your tooling supports it:
-
-```javascript{3,6}
-function ListItem({ item }) {
-  return (
-    <>
-      <dt>{item.term}</dt>
-      <dd>{item.description}</dd>
-    </>
-  );
-}
-```
-
-You may also map a collection of items to an array of fragments as you would any other type of element:
+You can map a collection of items to an array of fragments as you would any other type of element as well:
 
 ```javascript{6,9}
 function Glossary(props) {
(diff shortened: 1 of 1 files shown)


content/docs/accessibility.md
```diff
@@ -65,13 +65,8 @@ function ListItem({ item }) {
...
     <dl>
-      {props.items.map(item => (
-        // Fragments should also have a `key` prop when mapping collections
-        <Fragment key={item.id}>
-          <dt>{item.term}</dt>
-          <dd>{item.description}</dd>
-        </Fragment>
-      ))}
+      <ListItem item={term: 'Fragment', description: 'A piece of something.' }>
+      <ListItem item={term: 'React', description: 'Respond.' }>
```
AlmeroSteyn: I think it improves the doc to show the `ListItem` used inside `Glossary`, good idea! 

However it would not be used like this in applications as the reason to componentise the `<dt>`/`<dd>` combination would be to re-use it. Otherwise I personally would just write them inline. Saves code.

Can you change this to use a `map` that then renders the `ListItem` component? THAT is actually he great benefit of `<Fragment>` in these cases.  Then it also flows nicely into the example where you show that the `<Fragment>` also needs a key when being mapped in place.

content/docs/accessibility.md
```diff
@@ -90,6 +85,24 @@ function ListItem({ item }) {
...
+        <Fragment key={item.id}>
+          <dt>{item.term}</dt>
+          <dd>{item.description}</dd>
+        </Fragment>
+      ))}
+    </dl>
+  );
+}
+```
+
```
AlmeroSteyn: Nice. But can this be swapped with the example above showing the empty `<></>` way of writing Fragments? I think this would then more logically flow. First showing that you have to put a key on `Fragment` when mapping and THEN if there are no props and your tooling supports it you can use the clean tags. What do you think?

content/docs/accessibility.md
jsfeldman: I was thinking of just showing it being used like what was done [here](https://reactjs.org/docs/components-and-props.html#composing-components), but that makes sense. Should be updated now to match your suggestions.

content/docs/accessibility.md
jsfeldman: Sure! Makes sense.

jsfeldman:
Okay. All the suggested changes have been made. Let me know what you think 😄.

AlmeroSteyn:
Looks good. Thank you!

AlmeroSteyn:
Awesome! Thank you!

PR #1158, opened 2018-09-05
Fix bug in getDerivedStateFromProps example

redbar0n:
prevPropsUserID needs to be null to be set properly when the component is mounted, due to the conditional check in getDerivedStateFromProps. The conditional would otherwise be 'false' on mount, and thus no state would be given.

email should probably also be null, for consistency, even though it won't make a practical difference here.


BartoszKlonowski:
Hello @redbar0n, thank you for the contribution!
Unfortunately the legacy content and blogs are no longer maintained, so I will close this PR as outdated.
However, please feel encouraged to create new PRs for other open issues!

PR #1161, opened 2018-09-07
Add Install React App Guide div id

moumen-soliman:
Add installation Guide for React App div id URL to `create-a-new-react-app.html` in the top of page


threepointone:
This doesn't seem desirable, but thank you for the PR!

PR #1164, opened 2018-09-09
clarify where warning appears

### LINKED ISSUE #1163: Tutorial: make clear warnings are to be found in dev console
I'm completely new to react and when I tried following https://reactjs.org/tutorial/tutorial.html , everything went smoothly except at the part before "Picking a Key" where it says

> For now, we should see a list of the moves that have occurred in the game and a warning that says:
> > Warning: Each child in an array or iterator should have a unique “key” prop. Check the render method of “Game”.

I didn't realize codepen example had secretly overridden console.log to display the error msg in the page itself, so I thought I should be seeing the warning on my test page as well. I spent a long time looking for it, reexamining my code and setup, etc etc, and finally ended up asking and getting an answer on stackoverflow.

It's silly how the simplest things can become impossible hurdle for beginners who are learning on their own. I think a clarification that 1. for people following along with their own dev environment, this is to be found in the browser's developer console, and 2. the codepen example had changed console.log behavior but that's not the default to be expected.

Thank you very much.

alexkrolick:
fixes #1163

### DIFF
--- a/content/tutorial/tutorial.md
+++ b/content/tutorial/tutorial.md
@@ -1052,7 +1052,7 @@ Let's `map` over the `history` in the Game's `render` method:
 
 **[View the full code at this point](https://codepen.io/gaearon/pen/EmmGEa?editors=0010)**
 
-For each move in the tic-tac-toes's game's history, we create a list item `<li>` which contains a button `<button>`. The button has a `onClick` handler which calls a method called `this.jumpTo()`. We haven't implemented the `jumpTo()` method yet. For now, we should see a list of the moves that have occurred in the game and a warning that says:
+For each move in the tic-tac-toes's game's history, we create a list item `<li>` which contains a button `<button>`. The button has a `onClick` handler which calls a method called `this.jumpTo()`. We haven't implemented the `jumpTo()` method yet. For now, we should see a list of the moves that have occurred in the game and a warning in the browser dev console that says:
 
 >  Warning:
 >  Each child in an array or iterator should have a unique "key" prop. Check the render method of "Game".


content/tutorial/tutorial.md
```diff
@@ -1052,7 +1052,7 @@ Let's `map` over the `history` in the Game's `render` method:
 
 **[View the full code at this point](https://codepen.io/gaearon/pen/EmmGEa?editors=0010)**
 
-For each move in the tic-tac-toes's game's history, we create a list item `<li>` which contains a button `<button>`. The button has a `onClick` handler which calls a method called `this.jumpTo()`. We haven't implemented the `jumpTo()` method yet. For now, we should see a list of the moves that have occurred in the game and a warning that says:
+For each move in the tic-tac-toes's game's history, we create a list item `<li>` which contains a button `<button>`. The button has a `onClick` handler which calls a method called `this.jumpTo()`. We haven't implemented the `jumpTo()` method yet. For now, we should see a list of the moves that have occurred in the game and a warning in the browser dev console that says:
```
gaearon: I don't think "dev console" is a common way to say it. IMO either just "browser console" or "developer tools console" would be better.

gaearon: Also if you're following along using the CodePen, you will actually see the warning in the rendered output itself.

alexkrolick: Yes, see #1163

welkin25: That was precisely the problem as I described in #1163 ... I thought my page should look like the CodePen example and spent a really long time trying to find the “missing” warning. Perhaps it would be better to change CodePen ex to show console so it matches what people would see in their own dev environment?

alexkrolick: changed the wording to "developer tools console"

PR #1170, opened 2018-09-11
Fix code explanation of setState({…}) behavior

dtinth:
- The example mutates `previousState`. Fixed by prepending `{}` to the argument list.
- Two different variables (previousState, state) are used, which may cause confusion, fixed by always using the name `previousState`.


BartoszKlonowski:
Hello @dtinth!
The part of the docs you update in this PR is no longer present in the latest version of it, so this change is no longer needed.
Thank you!

PR #1171, opened 2018-09-11
Update render-props.md

moismat:
to make it a little bit clearer and to stay consistent with the sentence before it. We're talking here in this paragraph about the "behaviour".

"How can we reuse **this behavior** in another component? In other words, if **another component** needs to know about the cursor position, can we encapsulate ~~that~~ **this behavior** so that we can easily share it with **that component**?

**this** => behaviour 
**that** => another component


threepointone:
Ehhhh I'm going ot push back since this sounds liek a nit :) But thank you for the PR!

PR #1175, opened 2018-09-12

drcmda:
The ternary syntax using parenthesis was an unconventional prettier invention, going out of its way to format react ternaries differently from normal ternaries (if i remember correctly, because vim users had trouble re-formating without a mouse). If they made the right decision or not is one thing, but the additional noise could mislead beginners here, thinking they have to add more characters to make a simple 2-slot ternary, or that it is a "react thing".


threepointone:
Sure, why not. Thanks!

PR #1176, opened 2018-09-12
Introducing the new profiler

bvaughn:
# [View the formatted blog post](https://deploy-preview-1176--reactjs.netlify.com/blog/2018/09/10/introducing-the-react-profiler.html)


sophiebits:
Idea: can we embed a live profiler demo in the post? (Just a iframe to now.sh or w/e so it isn’t a maintenance burden.)

bvaughn:
That...is a pretty cool idea. Maybe! I'll play around with it.

bvaughn:
I think I'll push a profiling-capable demo to now and share as a link with folks, but separately from the Blog post. That seems like a reasonable compromise.

bvaughn:
http://react-devtools-profiler-demo.now.sh/

Issue #516, opened 2018-01-13
[Documentation] - Minor change to code example of 'Tutorial: Taking Turns'


bpasieka:
The [tutorial.md](https://github.com/reactjs/reactjs.org/blob/master/content/tutorial/tutorial.md#taking-turns) ~line 560:

based on [State Updates May Be Asynchronous](https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous) should it be:
```
  this.setState((state) => ({
    squares: squares,
    xIsNext: !state.xIsNext
  }));
```
instead of 
```
this.setState({
  squares: squares,
  xIsNext: !this.state.xIsNext,
});
```
The same for all other places this piece of code is used.

PS. let me know if it's okay for me to create a PR to fix it.
Thanks!

andrewvarga:
+1, I wanted to propose the exact same thing.

alexkrolick:
Both should work in this example; the functional form is primarily used when several `setState` calls are made and they depend on the updated value. For example:

```jsx
// this.state.counter === 0
this.setState({ counter: this.state.counter + 1 })
this.setState({ counter: this.state.counter + 1 })
this.setState({ counter: this.state.counter + 1 })
// counter is still 1 instead of 3 because the 2nd and 3rd calls don't use the new value
```

andrewvarga:
Sure, it will work in this simple case, but generally it may not, and in my opinion a tutorial should encourage best practices.

To paraphrase the documentation:
"Because this.props and this.state may be updated asynchronously, you should not rely on their values for calculating the next state."

alexkrolick:
You're welcome to make a PR to propose the change 😄

BartoszKlonowski:
Hello everyone!
As I can see the latest implementation of the tic-tac-toe tutorial does no longer use the class components, and the `xIsNext` is based on the hooks, also the current implementation looks fine and works well.
I don't find this issue any longer valid, so let me close it as well as the PR fixing it (CC: @meltuhamy).
If you find this issue still existing please post a comment.
Thank you!

PR #1180, opened 2018-09-13
Update tutorial to use setState with callback when depending on previous state

meltuhamy:
The tutorial currently uses `setState` while referencing previous state using `this.state`. 

According to the docs, "you should not rely on their values for calculating the next state". 

This PR makes the tutorial consistent with this and explains this best practice. 

Relates to issue #516


BartoszKlonowski:
Closing as per discussion in the linked issue.
Thank you for taking the time to contribute, @meltuhamy 👍

PR #1181, opened 2018-09-13
Update tutorial: Just swap Ben and Alexa ordering

Hocdoc:
The text below of the example state that Alexa and Ben's ordering is just swapped. Why is the number of tasks than changed?

### DIFF
--- a/content/tutorial/tutorial.md
+++ b/content/tutorial/tutorial.md
@@ -1073,9 +1073,9 @@ Imagine transitioning from
 to
 
 ```html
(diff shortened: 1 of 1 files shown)


sophiebits:
The fact that the numbers change is intentional. If they didn't, it would lead you to believe that comparing the text would be a more viable way to automatically key the components.

However, I agree that this can be a little confusing without any acknowledgement in the prose. I pushed an update in 95f0232d to hopefully make this clearer, along with a few other wording tweaks.

PR #1189, opened 2018-09-18
Renamed schedule/tracking API references to schedule/tracing

bvaughn:
I did not leave notes about the earlier names because I didn't want the blog post to be confusing, although I did update both gists to link to previous revisions (when it was "track").


bvaughn:
This is vaguely time sensitive so I'm merging it 😄 I'll change things retroactively if there are any concerns!

PR #1209, opened 2018-09-29
fix: redirect slugs with and without trailing slashes

### LINKED ISSUE #1208: Links that don't end with / work on fresh load but not on client navigation
For example https://reactjs.org/blog (no slash at the end) loads, but it's broken when clicked as a link.
You can revert https://github.com/reactjs/reactjs.org/pull/1207 locally to see the problem.

jlengstorf:
fixes #1208


jlengstorf:
Should the bot auto-update after I sign the CLA? Or do I need to close/reopen?

gaearon:
Nice, thanks! Is blog the only special case here? What about other `createRedirect`s? How do you know this is the right place to fix it architecturally? How does it rely to existing comment about `/` higher in the file?

gaearon:
Thank you!

PR #1211, opened 2018-09-29
Change highlight color

gaearon:
That's something that looked odd to me since the redesign, and after talking to a designer they found it confusing too that our highlighting is inverted. Text editors don't do that. Why do we?

screen shot 2018-09-29 at 7 21 47 pm

So I'm changing it to be lighter instead:

screen shot 2018-09-29 at 7 19 56 pm

This is similar to what I see editors do:

image 2018-09-28 at 7 12 40 pm 1
screen shot 2018-09-29 at 7 22 50 pm

I tried to not make it *too* light since it's meant to emphasize code. I think this is acceptable. I'm happy to darken it a few notches if somebody wants to.

### DIFF
--- a/src/prism-styles.js
+++ b/src/prism-styles.js
@@ -12,7 +12,7 @@ const prismColors = {
   char: '#D8DEE9',
   comment: '#999999',
   keyword: '#c5a5c5',
-  lineHighlight: '#14161a',
+  lineHighlight: '#353b45', // colors.dark + extra lightness
   primitive: '#5a9bcf',
   string: '#8dc891',
   variable: '#d7deea',


sophiebits:
❤️

gaearon:
Yeah I was a bit concerned about this, thanks for jumping on it.
Can you send a PR for proposed changes so I can look?

AlmeroSteyn:
@gaearon Will do. 

However my dev env is bust. Get the same issues as here https://github.com/reactjs/reactjs.org/issues/1088

As soon as I can get it up and running again I will get the PR going when I have a free moment.

bvaughn:
#1218

PR #1212, opened 2018-09-29
Adding a couple of redirects for pages returning 404

enmanuelduran:
**This Pull Request does a couple of simple things:**

- [ ] Adds a couple of redirects to improve accesibility in some sections, these URL's were returning a 404 before so it shouldn't be a problem:
- `/community` for `/community/support.html`
- `/tutorial` for `/tutorial/tutorial.html`
- [ ] Improves and Deletes some blocks of unused code
- [ ] Fixes liting errors

Another reason behind the added redirects is that that `/blog` and `/docs` are behaving in this same way so maybe it's a good idea to have this as a consistent behavior between sections.

### DIFF
--- a/content/community/support.md
+++ b/content/community/support.md
@@ -5,6 +5,7 @@ layout: community
 sectionid: community
 permalink: community/support.html
 redirect_from:
+  - "community/"
   - "support.html"
 ---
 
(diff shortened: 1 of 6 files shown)


Marcholio:
There are some conflicts that need to be resolved

PR #1216, opened 2018-10-01
Add integrity hashes to project yarn.lock

alejandronanez:
Running `yarn` using `yarn@1.10.1` will add `integrity hashes` to every entry in the `yarn.lock` file.

This PR only updates the `yarn.lock` with those new fields.


rickhanlonii:
Sorry for the delay, closing since we have these now 👍

PR #1219, opened 2018-10-01
Revert "feat: tweak plugin to use onCreateNode API, and get hot reloading working"

gaearon:
Reverts reactjs/reactjs.org#1205

Really sad about this but it seems to be breaking *some* PR builds (for example https://app.netlify.com/sites/reactjs/deploys/5bb2a8401f12b75242e69e98), and I don't know how to fix it. It also seems to have caused reoccurrence of https://github.com/reactjs/reactjs.org/issues/1088#issuecomment-425732762 on Windows.


bvaughn:
Really? 😦It seemed okay with my (limited) local testing.

The failed build you linked to has a GraphQL error in `src/pages/docs/error-decoder.html.js`. Seems unlikely that was related to this change?

gaearon:
Seems like unrelated, yep. I only thought it's related because https://github.com/reactjs/reactjs.org/issues/1088 started getting new reports around the same time.

DSchau:
@gaearon glad my PR didn't get reverted 😱 🎉 

I'll see if I can find some time to fix this. It seems like an intermittent bug in Gatsby's `File` type detection (the query itself should be fine, and _seems_ to be fine most of the time).

PR #1223, opened 2018-10-03
Add docs for getDerivedStateFromError [16.6]

bvaughn:
# Do not merge until React 16.6 has been released

Update `React.Component` reference and "Error Boundary" docs pages to include the newly added `static getDerivedStateFromError()` lifecycle.

### DIFF
--- a/content/docs/error-boundaries.md
+++ b/content/docs/error-boundaries.md
@@ -22,18 +22,21 @@ Error boundaries are React components that **catch JavaScript errors anywhere in
 > * Server side rendering
 > * Errors thrown in the error boundary itself (rather than its children)
 
-A class component becomes an error boundary if it defines a new lifecycle method called `componentDidCatch(error, info)`:
+A class component becomes an error boundary if it defines either (or both) of the lifecycle methods [`static getDerivedStateFromError()`](/docs/react-component.html#static-getderivedstatefromerror) or [`componentDidCatch()`](/docs/react-component.html#componentdidcatch).
 
-```js{7-12,15-18}
+```js{7-10,12-15,18-21}
 class ErrorBoundary extends React.Component {
   constructor(props) {
     super(props);
     this.state = { hasError: false };
   }
 
+  static getDerivedStateFromError(error) {
+    // Update state so the next render will show the fallback UI.
+    return { hasError: true };
+  }
+
   componentDidCatch(error, info) {
-    // Display fallback UI
-    this.setState({ hasError: true });
     // You can also log the error to an error reporting service
     logErrorToMyService(error, info);
   }
@@ -43,7 +46,8 @@ class ErrorBoundary extends React.Component {
       // You can render any custom fallback UI
       return <h1>Something went wrong.</h1>;
     }
-    return this.props.children;
+
+    return this.props.children; 
   }
 }
 ```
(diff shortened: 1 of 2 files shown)


content/docs/reference-react-component.md
```diff
@@ -311,21 +313,110 @@ In the above examples, it is important to read the `scrollHeight` property in `g
...
+It receives two parameters:
 
-A class component becomes an error boundary if it defines this lifecycle method. Calling `setState()` in it lets you capture an unhandled JavaScript error in the below tree and display a fallback UI. Only use error boundaries for recovering from unexpected exceptions; don't try to use them for control flow.
+1. `error` - The error that was thrown.
+2. `info` - An object with a `componentStack` key containing [information about which component threw the error](/docs/error-boundaries.html#component-stack-traces).
 
-For more details, see [*Error Handling in React 16*](/blog/2017/07/26/error-handling-in-react-16.html).
+
+`componentDidCatch()` is called during the "commit" phase, so side-effects are permitted.
+It can be used for things like logging errors:
```
acdlite: Edit: It *should* be used

content/docs/error-boundaries.md
```diff
@@ -22,18 +22,21 @@ Error boundaries are React components that **catch JavaScript errors anywhere in
 > * Server side rendering
 > * Errors thrown in the error boundary itself (rather than its children)
 
-A class component becomes an error boundary if it defines a new lifecycle method called `componentDidCatch(error, info)`:
+A class component becomes an error boundary if it defines either (or both) of the lifecycle methods [`static getDerivedStateFromError()`](/docs/react-component.html#static-getderivedstatefromerror) or [`componentDidCatch()`](/docs/react-component.html#componentdidcatch).
```
acdlite: If I'm new to React, I read this and think "Why are there two methods that do the same thing?"

content/docs/reference-react-component.md
```diff
@@ -311,21 +313,110 @@ In the above examples, it is important to read the `scrollHeight` property in `g
...
+
+    return this.props.children; 
+  }
+}
+```
 
 > Note
 > 
-> Error boundaries only catch errors in the components **below** them in the tree. An error boundary can’t catch an error within itself.
+> In the event of an error, you can render a fallback UI with `componentDidCatch()` by calling `setState`, but this will be deprecated in a future release.
```
acdlite: Can we move this explanation earlier, and explain that defining `componentDidCatch` without also defining `getDerivedStateFromError` is also deprecated, and in the future you *must* define `getDerivedStateFromError` in order for the component to act as an error boundary?

content/docs/error-boundaries.md
bvaughn: That's good input. I'll add a little more context there.

bvaughn:
I've tagged this with a new "v16.6" label to make it easy to merge when the release goes out (since I won't be around to remember to do it).

content/docs/reference-react-component.md
```diff
@@ -311,21 +313,110 @@ In the above examples, it is important to read the `scrollHeight` property in `g
...
```
Andarist: what are advantages of separating those 2 things? i understand the part about side effects being allowed in one and not in the other one, but `componentDidCatch` is working for both cases at the moment, so what's the reason behind separating them? non of them should really be called often (they are last resort kind of thing) so it raises a question if they are both needed as they will for sure be confusing for newcomers

bvaughn: My understanding of this is that there are a couple of motivating factors:

**It works with server-side rendering**. `componentDidCatch` is a _commit phase_ lifecycle, but there is no commit phase on the server. `getDerivedStateFromError` is a _render phase_ lifecycle, and so it can be used to enable error handling on the server.

**Render phase recovery is safer**. The story for error recovery via `componentDidCatch` is a little janky, since it relies on an intermediate commit of "null" for everything below the component that errored. This might result in subsequent errors inside of any components higher up in the tree that implement `componentDidMount` or `componentDidUpdate` and _just assume_ that their refs will be non-null (because they always are in the non-error case).

**It's more optimal for async rendering**. Because state-updates from commit phase lifecycles are _always synchronous_, and because `componentDidCatch` is called during the commit phase– using `componentDidCatch` for error recovery is not optimal because it forces the fallback UI to always render synchronously.

Andarist: I ofc dont judge the chosen APIs too harshly as Im not competent enough to do that, not knowing all intricacies of the implementation. Just commenting from the user's perspective. And thanks for the explanation!

Just want to comment a little bit on the given arguments

- **It works with server-side rendering** - this is appealing
- **Render phase recovery is safer.** - seems like implementation detail that could be handled differently, but i understand it would probably make the implementation more clunky as it would have to be special cased commit hook
- **It's more optimal for async rendering.** - not the strongest argument (as you have mentioned - those should be edge cases and thus being optimal for rendering shouldnt matter as much)

PR #1225, opened 2018-10-03
Location updated Delhi NCR

anoop-gupt:


anoop-gupt:
I have updated the location of React meetup in Delhi NCR. Please merge it.

PR #1231, opened 2018-10-05
Update Render Props.md to include link to Robin Wieruch's article

SeanGroff:
Robin's article on render props uses a real-life example to explain the render prop pattern. I've felt for a while it's so good it belongs somewhere in the official docs. 

Let me know if this isn't the best place to include his article. I'll gladly create another PR to include it somewhere else in the React docs.


rwieruch:
@alexkrolick even though render props are not heavily used anymore, many people tell me how much this article helped them to understand the concept. Any chance to get it merged? 

Thanks so much @SeanGroff for the shoutout 👍

PR #1232, opened 2018-10-06
Update examples.md

draftstudios:
adding side-scroller demo :)


gaearon:
Thanks, but the code formatting is very inconsistent and I'm not sure this example teaches much.

PR #1233, opened 2018-10-06
Update cross-origin-errors.md

harisbaig100:
I have added the methods to bypass CORB through browsers in any OS.


rickhanlonii:
Sorry for the delay, closing since these docs no longer exist on the new site - but also we probably shouldn't document how to disable cross-orgin checks because usually the issue is real and should be fixed instead.

PR #1235, opened 2018-10-08
Improvement in the documentation for usage without ES6

enmanuelduran:
I was reading the documentation of [mixins for react without ES6](https://reactjs.org/docs/react-without-es6.html#mixins) and I found that in the mixins example we're changing the seconds state without using a function, I understand that maybe this is done in this way because we have an interval of one second between each call of the `tick` method that sets the state, but maybe since [we recommend to use  functions for such cases](https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous) it would be a good idea to use the function to avoid confusion? or we could add a text explaining why we didn't use the function there?

_just a thought_

### DIFF
--- a/content/docs/react-without-es6.md
+++ b/content/docs/react-without-es6.md
@@ -205,7 +205,11 @@ var TickTock = createReactClass({
     this.setInterval(this.tick, 1000); // Call a method on the mixin
   },
   tick: function() {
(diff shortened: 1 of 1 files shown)


rickhanlonii:
Sorry for the delay, closing since these docs are gone.

PR #1236, opened 2018-10-08
correction: returning `null` does fire components lifecycle methods

cb504:
removed the word __not__

See class `WarningBanner2` in https://codepen.io/icbm504/pen/OBbEBB?editors=1111
... and check console when clicking `show/hide` button


gaearon:
It says "does not affect". Not "does not fire".

Maybe let's change the whole thing to "still fires"?

threepointone:
Closing for lack of updates.

PR #1237, opened 2018-10-08
Terminology: Lifecycle hooks -> Lifecycle methods

gaearon:
We sometimes say "lifecycle hooks" and sometimes "lifecycle methods". This makes it consistent. "Methods" is more descriptive because they *are* methods.


sophiebits:
dan is very sneaky

PR #1239, opened 2018-10-09
Added arrow function to props.onClick

esadler44:
Earlier in the example it even warns about setting onClick={<function>} as it will call that function on every render call.
This simply wraps it to prevent that inefficiency and only call the function on click.


esadler44:
This change probably needs to be propagated through to other parts of the example, but I just want to make sure that this is correct first.

esadler44:
Nevermind, I've started learning React and I now understand what's going on here, and how the injected JSX gets called when a render happens.

PR #1244, opened 2018-10-11
Upgrade to latest Gatsby packages

KyleAMathews:
This includes a number of fixing including this PR which prevents errors
during development when editing markdown files gatsbyjs/gatsby#8859

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "unist-util-visit": "^1.1.3"
   },
   "engines": {
-    "node": "^8.4.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -56,7 +56,7 @@
     "unist-util-visit": "^1.1.3"
   },
   "engines": {
-    "node": "^8.4.0",
+    "node": ">8.4.0",
```
KyleAMathews: So I can use node 10.x

gaearon:
Thanks

PR #1246, opened 2018-10-11
Update motivation-solution.js

cesar:
Changing the name of the variable `theme` to `value` might make it a bit more intuitive at first glance.


threepointone:
I'm going to push back, this isn't as significant as you'd think, so I don't want to pull it in. But thank you for the PR!

PR #1251, opened 2018-10-12
Type Error

zen-coder:


threepointone:
This isn't right, since the ref points to an instance of CustomTextInput. Thanks for the effort!

PR #1256, opened 2018-10-14
Update tutorial create-react-app to use npx

nywillb:
I changed the instructions from globally installing create-react-app to simply using npx.


saranshkataria:
Since this has already been done, closing the PR out. Thanks for the PR though.

PR #1263, opened 2018-10-16
Document deprecation of findDOMNode

sebmarkbage:
See https://github.com/facebook/react/pull/13841

We might want to open an RFC for fragment refs which could solve the one remaining use case in a better way but at least let us get the deprecation in.


content/docs/strict-mode.md
```diff
@@ -50,6 +51,32 @@ Since object refs were largely added as a replacement for string refs, strict mo
 
 [Learn more about the new `createRef` API here.](/docs/refs-and-the-dom.html)
 
+### Warning about deprecated findDOMNode usage
+
+React, used to support findDOMNode to search the tree for a DOM node given a class instance. Normally you don't need this because you can [attach a ref directly to a DOM node](/docs/refs-and-the-dom.html#creating-refs).
```
gaearon: Lose the first comma and backtick findDOMNode?

PR #1265, opened 2018-10-16
Document class contextType as the primary consuming mechanism

sebmarkbage:
https://github.com/facebook/react/pull/13728

This documents the Class.contextType API.

The intention here is to document this as the primary way to consume a Context from a class, and since classes are still the mainstream component API, it's the primary way to consume a Context period.

The Consumer API is documented as the way to consume a Context from a function component.

As a result I could remove the following examples:

__Accessing Context in Lifecycle Methods__ - Accessing context in life-cycles is now easy in since the primary way makes that the default. Technically this is still relevant for multiple context in a single class, but it's such an edge case that I feel like it belongs on Stack Overflow or some recipes section. Not the primary documentation.

__Forwarding Refs to Context Consumers__ - While this is still something you might want to do, this is not different than forwarding refs in general. It used to be that this pattern was unique in that when you converted from a class to use Consumer + forwardRef, you would get into this awkward situation. However, if the primary API is to use `.contextType` on a class, then this situation doesn't happen so doesn't seem to add the confusion.

__Consuming Context with a HOC__ - This pattern is possibly still relevant but it's only relevant because the render prop syntax is kind of a mess when you compose many of them. However, it looks like we're not going to be recommending this pattern and instead look to make it easier to consume context in other ways like the [unstable_read](https://github.com/facebook/react/pull/13139) proposal. Since this pattern isn't critical even before that's stable I think we should just remove it.


content/docs/context.md
```diff
@@ -114,46 +112,89 @@ However, sometimes the same data needs to be accessible by many components in th
...
 ```
 
-A React component that allows Consumers to subscribe to context changes.
+Every Context object comes with a Provider React component that allows consuming components to subscribe to context changes.
 
-Accepts a `value` prop to be passed to Consumers that are descendants of this Provider. One Provider can be connected to many Consumers. Providers can be nested to override values deeper within the tree.
+Accepts a `value` prop to be passed to consuming components that are descendants of this Provider. One Provider can be connected to many consumers. Providers can be nested to override values deeper within the tree.
 
-### `Consumer`
+All consumers that are descendants of a Provider will re-render whenever the Provider's `value` prop changes. The propagation from Provider to its descendant consumers is not subject to the `shouldComponentUpdate` method, so the consumer is updated even when an ancestor component bails out of the update.
```
bvaughn: I wonder if we should clarify that context _is_ subject to a component's _own_ `shouldComponentUpdate` method, just not its ancestors. (Seems like this wording could be misleading.)

content/docs/context.md
```diff
@@ -114,46 +112,89 @@ However, sometimes the same data needs to be accessible by many components in th
...
+    /* ... */
+  }
+  componentWillUnmount() {
+    let value = this.context;
+    /* ... */
+  }
+  render() {
+    let value = this.context;
+    /* render soemthing based on the value of MyContext */
+  }
```
bvaughn: Is it worth showing `shouldComponentUpdate` here so we can illustrate `this.context` vs the `nextContext` param? Or would that muddy the waters too much?

If not here, we _should_ mention that param/behavior somewhere now that the "_Accessing Context in Lifecycle Methods_" section is gone.

content/docs/context.md
```diff
@@ -114,46 +112,89 @@ However, sometimes the same data needs to be accessible by many components in th
...
+    let value = this.context;
+    /* ... */
+  }
+  componentWillUnmount() {
+    let value = this.context;
+    /* ... */
+  }
+  render() {
+    let value = this.context;
+    /* render soemthing based on the value of MyContext */
```
markerikson:
```suggestion
    /* render something based on the value of MyContext */
```

alexkrolick:
I think the HOC examples should be kept around. I'm not sure that consuming just a single context from a component is actually that common. The new API works well as a 1-1 replacement for some usages of legacy context, but most apps I've seen use a combination of HOCs and render props to inject data providers and there are usually more than one (for example react-intl `injectIntl` + Redux `connect`, plus some app-specific login context, navigation context, etc.).

gaearon:
You can still use render prop API for cases when you need that. This just attempts to cover the 80% case.

PR #1266, opened 2018-10-16
Document React.pure

sebmarkbage:
https://github.com/facebook/react/pull/13748

Pretty straight forward.

Still undecided whether we should support forwardRef by default or not. https://github.com/facebook/react/pull/13822

### DIFF
--- a/content/docs/reference-react.md
+++ b/content/docs/reference-react.md
@@ -26,6 +26,10 @@ React components let you split the UI into independent, reusable pieces, and thi
 
 If you don't use ES6 classes, you may use the `create-react-class` module instead. See [Using React without ES6](/docs/react-without-es6.html) for more information.
 
+React components can also be defined as functions which can be wrapped by further functionality.
+
+- [`React.pure`](#reactpure)
+
 ### Creating React Elements
 
 We recommend [using JSX](/docs/introducing-jsx.html) to describe what your UI should look like. Each JSX element is just syntactic sugar for calling [`React.createElement()`](#createelement). You will not typically invoke the following methods directly if you are using JSX.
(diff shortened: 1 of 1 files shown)


content/docs/reference-react.md
```diff
@@ -88,6 +92,42 @@ If your React component's `render()` function renders the same result given the
...
+  /*
+  return true if passing nextProps to render would return
+  the same result as passing prevProps to render,
+  otherwise return false
+  */
+}
+const MyComponent = React.pure(render, equals);
+```
+
+This method only exists as a **[performance optimization](/docs/optimizing-performance.html).** Do not rely on it to "prevent" a rendering, as this can lead to bugs.
```
acdlite:
```suggestion
This method only exists as a **[performance optimization](/docs/optimizing-performance.html).** Do not rely on it to "prevent" a render, as this can lead to bugs.
```

content/docs/reference-react.md
```diff
@@ -26,6 +26,10 @@ React components let you split the UI into independent, reusable pieces, and thi
 
 If you don't use ES6 classes, you may use the `create-react-class` module instead. See [Using React without ES6](/docs/react-without-es6.html) for more information.
 
+React components can also be defined as functions which can be wrapped by further functionality.
+
+- [`React.pure`](#reactpure)
```
acdlite:
```suggestion
- [`React.memo`](#reactmemo)
```

content/docs/reference-react.md
```diff
@@ -88,6 +92,42 @@ If your React component's `render()` function renders the same result given the
 
 * * *
 
+### `React.pure`
+
+```javascript
+const MyComponent = React.pure(function(props) {
```
acdlite:
```suggestion
const MyComponent = React.memo(function(props) {
```

content/docs/reference-react.md
```diff
@@ -88,6 +92,42 @@ If your React component's `render()` function renders the same result given the
...
+  /* render using props */
+});
+```
+
+`React.pure` is a [higher order component](/docs/higher-order-components.html). It's similar to [`React.PureComponent`](#reactpurecomponent) but for function components instead of classes.
```
acdlite:
```suggestion
`React.memo` is a [higher order component](/docs/higher-order-components.html). It's similar to [`React.PureComponent`](#reactpurecomponent) but for function components instead of classes.
```

content/docs/reference-react.md
```diff
@@ -88,6 +92,42 @@ If your React component's `render()` function renders the same result given the
...
+
+If your function component renders the same result given the same props, you can wrap it in a call to `React.pure` for a performance boost in some cases by memoizing the result.
```
acdlite:
```suggestion
If your function component renders the same result given the same props, you can wrap it in a call to `React.memo` for a performance boost in some cases by memoizing the result.
```

content/docs/reference-react.md
```diff
@@ -88,6 +92,42 @@ If your React component's `render()` function renders the same result given the
...
+  /* render using props */
+}
+function equals(prevProps, nextProps) {
+  /*
+  return true if passing nextProps to render would return
+  the same result as passing prevProps to render,
+  otherwise return false
+  */
+}
+const MyComponent = React.pure(render, equals);
```
acdlite:
```suggestion
const MyComponent = React.memo(render, equals);
```

content/docs/reference-react.md
```diff
@@ -88,6 +92,42 @@ If your React component's `render()` function renders the same result given the
 
 * * *
 
+### `React.pure`
```
acdlite:
```suggestion
### `React.memo`
```

PR #1269, opened 2018-10-17
add missing semi-colon in conditional rendering codeblock

alex-wilmer:
I'm sure these discussions have been going on, but personally I think running all of the examples through prettier's default settings would go a long way, not just in the react docs, but basically everywhere. seems like the closest thing to a standard js has ever come by.

anyways, you dropped this `;`

😎


gaearon:
sure

PR #1282, opened 2018-10-23
Document React.memo

sebmarkbage:
Supersedes #1266

### DIFF
--- a/content/docs/reference-react.md
+++ b/content/docs/reference-react.md
@@ -95,7 +95,7 @@ If your React component's `render()` function renders the same result given the 
 ### `React.memo`
 
 ```javascript
-const MyComponent = React.memo(function(props) {
+const MyComponent = React.memo(function MyComponent(props) {
   /* render using props */
 });
 ```
@@ -107,7 +107,7 @@ If your function component renders the same result given the same props, you can
 By default it will only shallowly compare complex objects in the props object. If you want control over the comparison, you can also provide a custom comparison function as the second argument.
 
 ```javascript
-function render(props) {
+function MyComponent(props) {
   /* render using props */
 }
 function equals(prevProps, nextProps) {
@@ -117,7 +117,7 @@ function equals(prevProps, nextProps) {
   otherwise return false
   */
 }
-const MyComponent = React.memo(render, equals);
+export default React.memo(MyComponent, equals);
 ```
 
 This method only exists as a **[performance optimization](/docs/optimizing-performance.html).** Do not rely on it to "prevent" a render, as this can lead to bugs.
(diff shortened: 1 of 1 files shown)


content/docs/reference-react.md
```diff
@@ -26,6 +26,10 @@ React components let you split the UI into independent, reusable pieces, and thi
 
 If you don't use ES6 classes, you may use the `create-react-class` module instead. See [Using React without ES6](/docs/react-without-es6.html) for more information.
 
+React components can also be defined as functions which can be wrapped by further functionality.
```
gaearon: functions [...] functionality

a bit repetitive?

content/docs/reference-react.md
```diff
@@ -88,6 +92,42 @@ If your React component's `render()` function renders the same result given the
...
+
+```javascript
+const MyComponent = React.memo(function(props) {
+  /* render using props */
+});
+```
+
+`React.memo` is a [higher order component](/docs/higher-order-components.html). It's similar to [`React.PureComponent`](#reactpurecomponent) but for function components instead of classes.
+
+If your function component renders the same result given the same props, you can wrap it in a call to `React.memo` for a performance boost in some cases by memoizing the result.
```
gaearon: Our readers probably don't know what "memoize" means. Maybe add

(This means that React will skip rendering the component, and reuse the last rendered result.)

content/docs/reference-react.md
```diff
@@ -88,6 +92,42 @@ If your React component's `render()` function renders the same result given the
...
+
+By default it will only shallowly compare complex objects in the props object. If you want control over the comparison, you can also provide a custom comparison function as the second argument.
+
+```javascript
+function render(props) {
```
gaearon: We should probably encourage proper name here so it shows up in DevTools

content/docs/reference-react.md
```diff
@@ -88,6 +92,42 @@ If your React component's `render()` function renders the same result given the
 
 * * *
 
+### `React.memo`
+
+```javascript
+const MyComponent = React.memo(function(props) {
```
gaearon: Also maybe name this too

content/docs/reference-react.md
```diff
@@ -88,6 +92,42 @@ If your React component's `render()` function renders the same result given the
...
+  /* render using props */
+}
+function equals(prevProps, nextProps) {
+  /*
+  return true if passing nextProps to render would return
+  the same result as passing prevProps to render,
+  otherwise return false
+  */
+}
+export default React.memo(MyComponent, equals);
```
sebmarkbage: Thoughts on using `export default` here? I don't want to name it something like `MyComponentMemo` since then people will just copy that pattern which is not how you're supposed to do it.

gaearon: seems fine to me

PR #1283, opened 2018-10-23
Document class contextType as the primary consuming mechanism

sebmarkbage:
Supersedes #1265


content/docs/context.md
```diff
@@ -114,46 +112,89 @@ However, sometimes the same data needs to be accessible by many components in th
...
+</MyContext.Consumer>
+```
+
+A React component that subscribes to context changes. This lets you subscribe to a context within a [function component](/docs/components-and-props.html#function-and-class-components).
+
+Requires a [function as a child](/docs/render-props.html#using-props-other-than-render). The function receives the current context value and returns a React node. The `value` argument passed to the function will be equal to the `value` prop of the closest Provider for this context above in the tree. If there is no Provider for this context above, the `value` argument will be equal to the `defaultValue` that was passed to `createContext()`.
 
 > Note
 > 
-> The way changes are determined can cause some issues when passing objects as `value`: see [Caveats](#caveats).
```
gaearon: Maybe we can move this note to Provider section instead of deleting?

syaau:
@sebmarkbage I think the document should also mention that when using the new `contextType`, they create conflict with libraries (like react-redux < 5.1) that hoist the non-react static properties, making the library unusable.

PR #1284, opened 2018-10-23
Document React.lazy

sebmarkbage:
Suspense is documented here since it's the only built-in way so far to use it. In the future it'll likely be documented on its own more like Error boundaries.

This hoists React.lazy as the primary way to do code splitting. However, before getting into it we keep the recommendation to use react-loadable for server-side rendered apps since Suspense is not yet ready for the server.

### DIFF
--- a/content/docs/code-splitting.md
+++ b/content/docs/code-splitting.md
@@ -111,13 +111,11 @@ If you're setting up Webpack yourself, you'll probably want to read Webpack's
 When using [Babel](http://babeljs.io/), you'll need to make sure that Babel can
 parse the dynamic import syntax but is not transforming it. For that you will need [babel-plugin-syntax-dynamic-import](https://yarnpkg.com/en/package/babel-plugin-syntax-dynamic-import).
 
-## Libraries
+## `React.lazy`
 
-### React Loadable
+> Note: `React.lazy` and Suspense is not yet available for server-side rendering. If you want to do code-splitting in a server rendered app, we still recommend [React Loadable](https://github.com/thejameskyle/react-loadable). It has a nice [guide for bundle splitting with server-side rendering](https://github.com/thejameskyle/react-loadable#------------server-side-rendering).
 
-[React Loadable](https://github.com/thejameskyle/react-loadable) wraps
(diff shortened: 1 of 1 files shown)


content/docs/code-splitting.md
```diff
@@ -111,45 +111,96 @@ If you're setting up Webpack yourself, you'll probably want to read Webpack's
...
+        <AnotherComponent />
+      </section>
+    </Suspense>
+  </div>
+);
+```
+
+### Error boundaries
+
+If the other module fails to load it will trigger an error. For example, due to network failure. To handle these errors to show a nice user experience and manage recovery see the [Error boundaries](/docs/error-boundaries.html) guide. Once you've created your Error boundary. You can use it anywhere above your lazy components to display an error state when there's a network error.
```
gaearon: Breaking up sentences like this is a bit hard to read to me.

Can you rephrase without breaking if ... for example ... to handle these?

content/docs/code-splitting.md
```diff
@@ -163,31 +214,44 @@ re-rendering the entire page at once so your users are unlikely to be
...
-  loader: () => import('./routes/Home'),
-  loading: Loading,
-});
-
-const About = Loadable({
-  loader: () => import('./routes/About'),
-  loading: Loading,
-});
+const Home = lazy(() => import('./routes/Home'));
+const About = lazy(() => import('./routes/About'));
```
gaearon: Have you checked this works? Just in case.

content/docs/code-splitting.md
```diff
@@ -111,45 +111,96 @@ If you're setting up Webpack yourself, you'll probably want to read Webpack's
...
+const MyComponent = () => (
+  <div>
+    <OtherComponent />
+  </div>
+);
+```
+
+This will automatically load the bundle containing the `OtherComponent` when this component gets rendered.
+
+`React.lazy` takes a function that must call a dynamic `import()`. This must return a `Promise` which resolves to a module with a `default` export containing a React component.
```
gaearon: This should probably have (see how to use named exports below) with a link.

PR #1286, opened 2018-10-24
Use 16.6.0

gaearon:


bvaughn:
Can you update [`versions.yml`](https://github.com/reactjs/reactjs.org/blob/master/content/versions.yml) to link to the last build of 16.5 docs?

PR #1294, opened 2018-10-24
Fix inner link for getDerivedStateFromError

kkowalczuk:
Fixed the inner link for getDerivedStateFromError on react-component page


lex111:
@kkowalczuk thank you, but already fixed!

PR #1295, opened 2018-10-24
More complete syntax

bsinkule:
Hello,

I made the props.children section that describes how props.children is available on every component. I think it is more clear to put the example in a variable and add a semi-colon.

Thank you,

Brice


threepointone:
I'm going to push back on this, because this isn't typical usage. But thank you for your PR!

PR #1309, opened 2018-10-25
Removes unused code from useReducer example

mikaelbr:
`initialCount` isn't used in the first `useReducer` example and could be removed.


BartoszKlonowski:
Hello @mikaelbr!
I see that in the latest version of the docs, there's no `initialCount` parameter anymore, so let me close this PR as outdated.
But please feel invited to create another PR for open issues.
Thank you!

PR #1312, opened 2018-10-26
Fix a few instaces of `re-` based words where the hyphen was left out.

Jared-Dev:
It seems that typically words like "re-render" include a hyphen. This PR simply adds those in when they were missing (and I noticed).


BartoszKlonowski:
Hello @Jared-Dev, thank you for the contribution!
Unfortunately, when I check at the latest version of the docs, I don't see any occurrences of `rerender` wording, so let me close this PR as outdated.
However, please feel free to create new PRs for other open issues!

PR #1314, opened 2018-10-26
Update hooks-faq.md

afontcu:
Hi!

This PR just sets the "Hook" word out of the linked text, as in the other instances of "some Hook".

Thanks for your amazing work with both the Hooks and the documentation itself 💯


threepointone:
Looks like this was fixed separately, but thank you for the PR!

PR #1316, opened 2018-10-26
Update hooks-intro.md

baering:
What changed
===========
Made link to Writing Custom Hooks lead to the top of the page rather than to Using a Custom Hook

Before the change
==============
When you clicked the link to _Writing Custom Hooks_ from `hooks-intro.html`:

image

You would be taken to `/docs/hooks-custom.html#using-a-custom-hook`:

image

After the change
============
Now when you click the link, you are taken to the top of the page, `/docs/hooks-custom.html`, as you would expect:

image


alexkrolick:
Thanks!

PR #1317, opened 2018-10-26
Small typo fix

tusharkhatiwada:

### DIFF
--- a/content/docs/hooks-faq.md
+++ b/content/docs/hooks-faq.md
@@ -109,7 +109,7 @@ There are a few more heuristics, and they might change over time as we fine-tune
 
 ### How do lifecycle methods correspond to Hooks?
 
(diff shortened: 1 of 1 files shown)


lex111:
@tusharkhatiwada sorry for the long delay, thanks!

content/docs/hooks-faq.md
```diff
@@ -109,7 +109,7 @@ There are a few more heuristics, and they might change over time as we fine-tune
 
 ### How do lifecycle methods correspond to Hooks?
 
-* `constructor`: Function components don't need a constructor. You can initialize the state in the [`useState`](/docs/hooks-reference.html#usestate) call. If computing it is expensive, you can pass a function to `useState`.
```
tesseralis: I admit the wording is confusing, so perhaps we could change this instead to: 

> if computing the state is expensive, you can pass a function to `useState`.

tesseralis: This isn't actually a typo. It reads as "If computing it (the state) is expensive, you can pass a function to `useState`.

PR #1318, opened 2018-10-26
Removed unused parameter from example code

mjomble:
Looks like a parameter from the next example was accidentally included in this one as well.

I did not edit the last line of the file, there might be a newline at the end of the file that GitHub's editor may have added or removed and it's highlighting that in the diff.


mjomble:
Oops, never mind, duplicate of #1309

PR #1324, opened 2018-10-26
Tip => Note

moroshko:
"Tip" usually contains an action for you to perform.
"Note" is more appropriate here since it's just an FYI.


saranshkataria:
LGTM, Thanks!

PR #1325, opened 2018-10-27
Fix typo

moroshko:


lex111:
Thank you, but already fixed in https://github.com/reactjs/reactjs.org/pull/1435

PR #1326, opened 2018-10-27
Correct Functional Updates example code

r4j4h:
I believe this example should have these anonymous functions accepting a named parameter that is used in their function body, as I don't see prevCount defined anywhere else and this makes sense. I could be wrong though!


r4j4h:
Ack, this change is not good. I did not read the code closely enough, the anonymous with the variable is handed to setCount. Sorry, it's what I get trying to read docs while distracted with people talking to me :blush:

PR #1330, opened 2018-10-28
Added import React for using JSX syntax

pavelkeyzik:


pavelkeyzik:
Closed because fixed by other user

PR #1333, opened 2018-10-28
Fix typo in Hooks FAQ

hnordt:

PR #1334, opened 2018-10-29
Remove astray word

koenpunt:
Should be straightforward.


saranshkataria:
#1333 was there before this. So closing this out since it was created earlier.

PR #1335, opened 2018-10-29
Update hooks-state.md

Lemaro86:
import React, { useState } from 'react'; 

missing React in string 14


Lemaro86:
May be i am stupid...

PR #1336, opened 2018-10-29
Fix miniscule typo in React.lazy documentation

rockymeza:


threepointone:
looks like this was fixed separately, but thank you for your PR!

PR #1337, opened 2018-10-29
Add note about Hooks working alongside Suspense

tylermcginnis:
https://twitter.com/tylermcginnis/status/1057015589393227776

### DIFF
--- a/content/docs/hooks-intro.md
+++ b/content/docs/hooks-intro.md
@@ -81,7 +81,7 @@ To solve these problems, **Hooks let you use more of React's features without cl
 
 >**TLDR: There are no plans to remove classes from React.**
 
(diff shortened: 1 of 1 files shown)


ThisIsMissEm:
Would it make sense to add an alert/warning box to each of the methods (useEffect especially) stating that performing data fetching may best be done with Suspense instead of useEffect, as to make it very clear & not lost between the introduction & the API docs

tylermcginnis:
@ThisIsMissEm I agree but went with the most subtle approach since I wasn't sure the overall sentiment on it. If the premise that Hooks are really only half the picture is actually true, seems like it should be all over.

tylermcginnis:
Closing this because I think the messaging has changed now. https://twitter.com/dan_abramov/status/1090738375705935873

PR #1341, opened 2018-10-29
Fix a broken link to code splitting technique on CRA projects

yuyokk:
Hello,

Here is a fix for create-react-app code-splitting guide link as CRA has a dedicated site now.

Please check.
Thanks 🙏


lex111:
@yuyokk thank you, but already fixed!

PR #1342, opened 2018-10-29
Consistently call the second argument to useEffect dependencies array

philipp-spiess:
This PR is more thought of as a discussion starter. I was trying to find out how we refer to the second argument of the `useEffect` hook and it seems that we use `inputs` in both the API reference as well as the implementation but this name isn't used in the guide article on effects.

This made it hard for someone to explain this concept to me since she didn't know how to call it and made up a name instead. I think we should be more up-front about this name so it's easier to discuss it.

We should probably also mention it in the API signature. I've wrapped it in square brackets like we do for `useImperativeMethods`.


acdlite:
Let's call it the "dependencies" array instead? I think that's the term we agreed on.

philipp-spiess:
@acdlite Sure! This makes a lot of sense for the `[]` case:

> Passing in an empty array `[]` of dependencies tells React that your effect doesn't depend on any values from the component

philipp-spiess:
This was superset by https://github.com/reactjs/reactjs.org/pull/1957 🙂

PR #1346, opened 2018-10-30
Render Props API -> Context API

ryota-murakami:
## Overview
Thank you for quite good post for I could catch up v16.6.0!

I found out 1 nitpick there.(I think `Context API` is correct at here)
New Context API syntax form is similar to render props(children) pattern, so I guess who attempted to mention it here.(e.g. render props like new Context API)

Thank you😀


ryota-murakami:
I mean Render Props is a technique, not a API name.
I have a concern to give confusion which are 
- what's the Render Props API? 
- what's the relational between Context API and Render Props? 
- Is Render Props API exactly same as following doc?
https://reactjs.org/docs/render-props.html

If I was misunderstanding, I'd like to close within at ease😌

PR #1350, opened 2018-10-31
Clarify that you can't return early between Hooks

jamesgpearce:
...Otherwise you get an error like `Uncaught Error: Rendered fewer hooks than expected. This may be caused by an accidental early return statement.` in `finishHooks`.


rickhanlonii:
Sorry for the delay, closing since these docs have moved and we have a note about this now: https://react.dev/reference/rules/rules-of-hooks#only-call-hooks-at-the-top-level

PR #1355, opened 2018-11-01
Fix grammar

henchies:


lex111:
@henchies thank you, but already fixed!

PR #1356, opened 2018-11-01
Be more precise about when false will stop default behaviour in an event handler.

sinkaszab:
The example shows an attached function in React, however doing the same in plain HTML event handler as a function with returning false would simply have no effect.

- Amends wording to me more exact.
- Attaches an external link to help understand the topic for one in need.

### DIFF
--- a/handling-events.md
+++ b/handling-events.md
@@ -29,7 +29,7 @@ is slightly different in React:
 </button>
 ```
 
-Another difference is that you cannot return `false` to prevent default behavior in React. You must call `preventDefault` explicitly. For example, with plain HTML, to prevent the default link behavior of opening a new page, you can write:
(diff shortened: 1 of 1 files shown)


alexkrolick:
Would it be worth differentiating between event.preventDefault() and event.stopPropagation()?

sinkaszab:
On the [Synthetic Events](https://reactjs.org/docs/events.html) page stopPrpagation is mentioned in the 1st note anyway.

I think this topic should be mentioned here, because an inline handler is in the plain HTML and the React example uses a callback. And when you use a plain HTML callback it behaves differently. Also nice for people coming from jQuery.

How would you differentiate preventDefault and stopPropagation?

sinkaszab:
But if you thought about clarifying the difference between the 2 methods, [Event Order](https://www.quirksmode.org/js/events_order.html) article also mentioned on MDN is a nice starting point.

PR #1357, opened 2018-11-01
Frenchdocs

emmadal:
# Translation of React docs in French
I want to translate the react documentation in french to allow some people that doesn't understand the English to learn react. that is will allow at french people to be comfortable with our native language in the learning of React


alexkrolick:
Thanks for asking about the translations. The docs are translated using CrowdIn. You can sign up to to help here: https://crowdin.com/project/react

PR #1358, opened 2018-11-02
Change links to use https

JoshDoug:


lex111:
@JoshDoug thank you, but already fixed!

PR #1362, opened 2018-11-02
Fix anchor link for getDerivedStateFromError

dmmulroy:
The anchor for `getDerivedStateFromError` was previously `#getderivedstatefromerror` when it should be `#static-getderivedstatefromerror`


gaearon:
Thx

PR #1364, opened 2018-11-02
Title change: Custom Hooks => Building Your Own Hooks

gaearon:
See

https://mobile.twitter.com/JoshWComeau/status/1058300101456773121
https://mobile.twitter.com/siddharthkp/status/1058322582024314880

@sophiebits's idea is to deemphasize "custom Hooks" as a name in the navigation.
(But keep calling them custom Hooks on the page.)

### DIFF
--- a/content/docs/hooks-custom.md
+++ b/content/docs/hooks-custom.md
@@ -1,6 +1,6 @@
 ---
 id: hooks-custom
-title: Writing Custom Hooks
+title: Building Your Own Hooks
 permalink: docs/hooks-custom.html
 next: hooks-reference.html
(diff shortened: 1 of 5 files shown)


vramana:
@gaearon @sophiebits Good change. But I think we should not use custom hooks at all and just say hooks. 

We never say write a custom functions in javascript. We just write a function. The custom doesn't add any value. Hook is just an another abstraction. React provides a a few hooks. Just call them built-in hooks. 

Just use the terminology, built-in hooks and hooks similar to built-in functions and functions. 

I can do a PR if you are okay with the idea.

PR #1368, opened 2018-11-03
Replace `let` with `const`

naruaway:
In other places, `const` is used consistently. I think using `let` here might cause unnecessary confusion.

### DIFF
--- a/content/docs/hooks-faq.md
+++ b/content/docs/hooks-faq.md
@@ -287,8 +287,8 @@ Here, we store the previous value of the `row` prop in a state variable so that 
 
 ```js
 function ScrollView({row}) {
-  let [isScrollingDown, setIsScrollingDown] = useState(false);
-  let [prevRow, setPrevRow] = useState(null);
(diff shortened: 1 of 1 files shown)


Haugen:
I think they intentionally used `let` here, since these values need to be able to change during a state update. For example, the value of `isScrollingDown` will toggle between `true` and `false`, and the use of `const` wouldn't allow for that.

naruaway:
Thanks for your reply! But I do not think so :(
The value of `isScrollingDown` is consistent while `ScrollView` is executing.
State changes initiated by `setIsScrollingDown` will cause another execution of `ScrollView` possibly with a different value of `isScrollingDown`.

If you assign some value to `isScrollingDown`, it will be out of sync with "the actual state" managed by React. By using `const` here, we can avoid this situation.

That's why I think using `let` here is confusing and we should fix this.

threepointone:
This was fixed separately, but thank you for the PR!

PR #1370, opened 2018-11-03
Bump gatsby-remark-autolink-headers, fixes sidebar links not correctly jumping to referenced element

fk:
Hey folks! 👋 We recently fixed an issue with `gatsby-remark-autolink-headers` that caused an issue with scrolling to anchor links:

- issue: https://github.com/gatsbyjs/gatsby/issues/9638
- fix: https://github.com/gatsbyjs/gatsby/pull/9657, released as `gatsby-remark-autolink-headers@2.0.10`

Since reactjs.org is using `gatsby-remark-autolink-headers`, it currently is affected by this issue, too:

- Directly opening https://reactjs.org/tutorial/tutorial.html#prerequisites shows the expected, correct behavior—the referenced element is fully visible:  
image
- Clicking one of the sidebar links however does not work as expected—the element referenced in the `href`s hash is not properly scrolled to, but slightly hidden behind the header: 
image

This PR fixes the described behavior by bumping `gatsby-remark-autolink-headers`' version to `2.0.10`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "gatsby-plugin-react-helmet": "^3.0.0",
     "gatsby-plugin-sharp": "^2.0.0",
     "gatsby-plugin-twitter": "^2.0.0",
-    "gatsby-remark-autolink-headers": "^2.0.0",
+    "gatsby-remark-autolink-headers": "^2.0.10",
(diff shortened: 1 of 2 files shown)


gaearon:
Thanks for this!

PR #1372, opened 2018-11-04
Update jsx-in-depth.md

mosheKenshoo:


alexkrolick:
Thanks for the suggestion but it's correct as is - `items` needs to be declared with `let` since `push` mutates the array 😄

PR #1375, opened 2018-11-05
Length of useEffect second argument

thchia:
[From this issue](https://github.com/facebook/react/issues/14090), adds a note to the docs that the second argument to `useEffect` must have a static length.


BartoszKlonowski:
Hello @thchia, thank you for the contribution!
Unfortunately, the part of the documentation you update in this PR is no longer present in the latest version of it. Moreover, the updated part (separated between each hook) already contains the mentioning of the requirement of `a constant number of items and be written inline (...)`.
Let me close this as PR as outdated, however, please feel free to create new PRs for other open issues!

PR #1376, opened 2018-11-05
Typo in useReducer docs, the initialCount prop is unused

einarq:
Unless I'm misunderstanding something, the initialCount prop should be passed as initial state to the useReducer hook.

### DIFF
--- a/content/docs/hooks-reference.md
+++ b/content/docs/hooks-reference.md
@@ -194,7 +194,7 @@ function reducer(state, action) {
 }
 
 function Counter({initialCount}) {
-  const [state, dispatch] = useReducer(reducer, initialCount);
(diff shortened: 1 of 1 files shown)


content/docs/hooks-reference.md
```diff
@@ -194,7 +194,7 @@ function reducer(state, action) {
 }
 
 function Counter({initialCount}) {
-  const [state, dispatch] = useReducer(reducer, initialState);
+  const [state, dispatch] = useReducer(reducer, {count: initialCount});
```
mehiel: This doesn't feel right though. 
In this example `initialState` is only used for reducer's `reset` action. Which will reset to an incorrect state as the `initialCount` may be different. IMHO in this example `Counter` should have *no props* as the correct example for `initialCount` is the one below with the `initialAction`

PR #1377, opened 2018-11-05
The callback function in `useCallback` should return a value

atymchuk:


AlekSi:
Any news on that?

threepointone:
This doesn't seem right, because the exampl implies side effects in doSomething, so the code is correct. Thanks for your PR!

PR #1381, opened 2018-11-08
fix: Small typo in the article

pReya:
Fixed a one-character typo in the documentation. Don't think contribution guidelines should play a role here ;)


lex111:
@pReya thank you, but already fixed!

PR #1382, opened 2018-11-08
docs: revise imports on destructing

evenstensberg:
When looking at the documentation and trying to follow along on using hooks I noticed that React is not getting imported and will make users doing the same as in the documentation to compile to syntax errors. This small PR fixes that :)


threepointone:
We fixed this separately, thanks for the PR!

PR #1384, opened 2018-11-09
Align forwardRef usage in useImperativeMethods example with other exa…

onnovisser:
The `forwardRef` usage in the `useImperativeMethods` example might give the mistaken impression that `function FancyInput` is a component instead of a render function. This PR brings the `forwardRef` usage more in line with other examples.


threepointone:
This isn't right, FancyInput *is* a component. We should probably revisit other examples too. Thanks for the PR!

PR #1391, opened 2018-11-10
Use @babel namespace package

kpman:
Change `@babel-plugin-syntax-dynamic-import` to `@babel/plugin-syntax-dynamic-import`.


harish-sethuraman:
Seems like this is now fixed.
https://reactjs.org/docs/code-splitting.html#import

Thanks for sending a PR. Sorry for the late turn around time.

PR #1392, opened 2018-11-10
Document possible spread attributes side effects

borekb:
Added an example where attribute spreading might lead to unexpected results.


BartoszKlonowski:
Hello @borekb!
Unfortunately the part you modify in this PR no longer exists in the docs, as well as the page. I will close this PR as outdated, but if you would still like to approach this topic feel free to create new PR or tackle other open issues if you would like to contribute.
Thank you!

PR #1394, opened 2018-11-12
Explain how `ref` and `key` are treated with cloneElement

### LINKED ISSUE #3840: The `React.cloneElement` docs are incorrect or very misleading
The [React.cloneElement](https://reactjs.org/docs/react-api.html#cloneelement) docs state:
> key and ref from the original element will be preserved.

This statement makes it seem like any `key` or `ref` passed in the new props will be ignored.

However this is clearly not the case, as can be seen in the [source code](https://github.com/facebook/react/blob/2edf449803378b5c58168727d4f123de3ba5d37f/packages/react/src/ReactElement.js#L496):

```javascript
export function cloneElement(element, config, children) {

  ...

  if (hasValidRef(config)) {
    // Silently steal the ref from the parent.
    ref = config.ref;
    owner = ReactCurrentOwner.current;
  }
  if (hasValidKey(config)) {
    key = '' + config.key;
  }
  
  ...
  
  return ReactElement(element.type, key, ref, self, source, owner, props);
}
```

stevemao:
I had problems understanding how `key` and `ref` are cloned over. Without the change, I don't know if new `key` or `ref` would replace the old ones or not.

Closes #3840

### DIFF
--- a/content/docs/reference-react.md
+++ b/content/docs/reference-react.md
@@ -149,12 +149,12 @@ Code written with [JSX](/docs/introducing-jsx.html) will be converted to use `Re
 ```
 React.cloneElement(
   element,
-  [props],
+  [config],
   [...children]
 )
 ```
 
-Clone and return a new React element using `element` as the starting point. The resulting element will have the original element's props with the new props merged in shallowly. New children will replace existing children. `key` and `ref` from the original element will be preserved.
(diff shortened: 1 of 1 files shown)


eps1lon (approved):
Sounds great, thanks!

PR #1403, opened 2018-11-17
Add information about versioning policy

sophiebits:


sophiebits:
@gaearon approved this via chat. :)

PR #1405, opened 2018-11-18
Update Hooks Intro Example() to `import React`

jsjoeio:
Hi there! 👋 

I was trying to test out Hooks for fun (cheers to the React team!). I copied the first example into my project and tried exporting that function to use as a component in another component and kept getting this error `Uncaught ReferenceError: React is not defined`.

I realized it was because I wasn't import `React` along with the `useState` hook.

So, I thought I would propose this change to update the example to import React so that future readers don't run into the same error.


threepointone:
We fixed this separately, thanks for the PR!

PR #1406, opened 2018-11-18
Mention wrapping custom hooks with render props/hoc in hooks FAQ

suchipi:
Based on a [Twitter conversation](https://twitter.com/dan_abramov/status/1063800913259913216), this PR adds a section to the FAQ explaining that you don't need to rewrite all your class components in order to use custom hooks in them, because they can be wrapped with a render prop component or hoc. And it shows how to use render props to do that.


suchipi:
@gaearon do you still want this? I rely on my "Pull Requests" page to keep track of my work so I kinda wanna close this if not

PR #1407, opened 2018-11-19
Replace unmaintained react-loadable package with loadable-components.

threehams:
`react-loadable` no longer has an active maintainer and has no support for Webpack 4 or Babel 7. Only personal forks are available. This isn't clear at all from the github page, and can only be discovered by digging into issues such as https://github.com/jamiebuilds/react-loadable/pull/151.

This PR links to two actively-developed projects which accomplish the same thing - split on the server and load all bundles in parallel.

Closes https://github.com/reactjs/reactjs.org/issues/1146, which was about the react-loadable license (which has since been reverted).

### DIFF
--- a/content/docs/code-splitting.md
+++ b/content/docs/code-splitting.md
@@ -115,7 +115,7 @@ parse the dynamic import syntax but is not transforming it. For that you will ne
 
 > Note:
 >
-> `React.lazy` and Suspense is not yet available for server-side rendering. If you want to do code-splitting in a server rendered app, we recommend [Loadable Components](https://github.com/smooth-code/loadable-components) or [React Universal Component](https://github.com/faceyspacey/react-universal-component).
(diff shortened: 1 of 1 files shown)


content/docs/code-splitting.md
```diff
@@ -115,7 +115,7 @@ parse the dynamic import syntax but is not transforming it. For that you will ne
 
 > Note:
 >
-> `React.lazy` and Suspense is not yet available for server-side rendering. If you want to do code-splitting in a server rendered app, we still recommend [React Loadable](https://github.com/thejameskyle/react-loadable). It has a nice [guide for bundle splitting with server-side rendering](https://github.com/thejameskyle/react-loadable#------------server-side-rendering).
+> `React.lazy` and Suspense is not yet available for server-side rendering. If you want to do code-splitting in a server rendered app, we recommend [Loadable Components](https://github.com/smooth-code/loadable-components) or [React Universal Component](https://github.com/faceyspacey/react-universal-component).
```
gaearon: What's the difference between them?

It looks like `react-universal-component` depends on legacy context. While this is true, I'd prefer that we don't recommend it.

threehams: Handling of CSS imports (including HMR) is much nicer in `react-universal-component`. They're otherwise pretty similar. Since that's a minor difference, I'll drop it in favor of just `loadable-components` and add a link to their SSR section.

PR #1409, opened 2018-11-19
Update React.lazy first code snippet

nuragic:
`React.lazy` first code snippet was not working out of the box. 

After the snippet it stated: _This will automatically load the bundle containing the `OtherComponent` when this component gets rendered._

But that's simply not true, *it'd simply throw* without `Suspense`. I got confused by this first example and needed to test it out on an isolated [CodeSandbox](https://codesandbox.io/s/mjz3p3w7xy) to verify that Suspense is actually needed. I think it'd be better to change the snippet and make it work out of the box, so copy-pasting will work (and the phrase _This will automatically load the bundle containing the `OtherComponent` when this component gets rendered._ will hold `true` 😄 ).

### DIFF
--- a/content/docs/code-splitting.md
+++ b/content/docs/code-splitting.md
@@ -147,6 +147,8 @@ function MyComponent() {
 }
 ```
 
+_*Note*: in order to make this snippet work you also need `Suspense`, please keep reading._
+
 This will automatically load the bundle containing the `OtherComponent` when this component gets rendered.
 
 `React.lazy` takes a function that must call a dynamic `import()`. This must return a `Promise` which resolves to a module with a `default` export containing a React component.


nuragic:
Yeah, I think it'd be better as you suggested, thanks! Just updated

nuragic:
Hey @gaearon do you think it'd be finally ok to merge as it is now? It's just a _note_ as you suggested. Thanks!

DanielRuf:
What also about importing `Suspense`? This sees to be missing.

nuragic:
Closing because this is almost 1 year old and it doesn't apply anymore; current docs and related examples are explicit now:

> The lazy component should then be rendered inside a Suspense component

PR #1412, opened 2018-11-19
React Norway

juicycleff:
This PR is to add ReactNorway event to the conference list


gaearon:
Can you please move it to be in the chronological order? The rest of conferences are.

juicycleff:
> Can you please move it to be in the chronological order? The rest of the conferences are.

Oops my bad, I thought I did that. Anyway, it's fixed now. Thank you

PR #1414, opened 2018-11-20
Specify articleLayout.container.flexDirection to 'column'

wongmjane:
`articleLayout.container.flexDirection` is supposed to be column direction since the layout is vertical.

Fixes #1404

PR #1416, opened 2018-11-20
articleLayout.container.flexDirection='column' only when media is less than 'small'

### LINKED ISSUE #1404: Horizontal scrolling can break on mobile
Spotted this on Twitter: https://twitter.com/wojtekmaj91/status/1063708337177534465

Didn’t verify myself.

wongmjane:
`articleLayout.container.flexDirection` is will be set to 'column' only when the media is less than 'small'

My #1414 previously broke desktop layout because the `flexDirection: column` applied to desktop as well, where it shouldn't had because `flexDirection: row | initial` is required for the sidebar to be placed next to the article

This should now work on both desktop and mobile:

output2

Fixes #1404

### DIFF
--- a/src/theme.js
+++ b/src/theme.js
@@ -115,6 +115,7 @@ const sharedStyles = {
   articleLayout: {
     container: {
       display: 'flex',
+      flexDirection: 'column',
       minHeight: 'calc(100vh - 60px)',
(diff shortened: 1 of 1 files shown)


gaearon:
Thanks!

PR #1417, opened 2018-11-20
Added useState updater function example to hooks docs

chrisjpatty:
The hooks proposal fails to mention that the update function returned from `useState` accepts an updater function similar to `this.setState()`. I've added a short example to the section titled "Updating State", which shows that an updater function can also be used.

### DIFF
--- a/content/docs/hooks-state.md
+++ b/content/docs/hooks-state.md
@@ -187,6 +187,14 @@ In a function, we already have `setCount` and `count` as variables so we don't n
   </button>
 ```
 
+Similiar to `this.setState()`, `setCount` also accepts an updater function:
+
+```js{1}
+  <button onClick={() => setCount((currentCount) => currentCount + 1)}>
+    Click Me
+  </button>
+```
+
 ## Recap
 
 Let's now **recap what we learned line by line** and check our understanding.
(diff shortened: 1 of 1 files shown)


content/docs/hooks-state.md
```diff
@@ -187,6 +187,14 @@ In a function, we already have `setCount` and `count` as variables so we don't n
   </button>
 ```
 
+Similiar to `this.setState()`, `setCount` also accepts an updater function:
+
+```js{1}
+  <button onClick={() => setCount((currentCount) => currentCount + 1)}>
```
bryndyment: Could be simplified to `currentCount => currentCount + 1` (just dropped the parens).

a8t:
This should really be documented. It would have saved me some headache.

Thanks!

BartoszKlonowski:
Hello @chrisjpatty, thanks so much for taking the time to contribute!
This PR is pretty old and from what I can see there already is a section that describes [updating the state based on previous state](https://react.dev/reference/react/useState#updating-state-based-on-the-previous-state) with proper examples of code.
Let me close this PR as outdated, but please feel invited to create new additions or fixes to docs for any open issues.
Thank you!

PR #1422, opened 2018-11-22
Warn about inlining React.lazy

FredyC:
See https://github.com/facebook/react/issues/14299


FredyC:
@gaearon Just pinging you here in case you missed it. It's the PR as you have asked in https://github.com/facebook/react/issues/14299.

Would like to hear more from you on recent comments there also.

jotto:
I got burned by this too @FredyC thanks for this PR

content/docs/code-splitting.md
```diff
@@ -151,6 +151,20 @@ This will automatically load the bundle containing the `OtherComponent` when thi
...
+
+```js
+function MyComponent() {
+  return (
+    {/* Don't do this */}
+    <Route path="/foo" component={React.lazy(() => import('./OtherComponent'))} />
+  );
+}
+```
+
```
dhedey: Might it be worth explaining why the resulting component gets remounted - ie that the object handle to the lazy Component is changing / being recreated every render?

And can we have also give a solution? Perhaps in the form of a hook?

We should also avoid Memo, because it's not guaranteed not to change (unlike Ref which is). In this case,

```
/* Do this instead. */
function useLazy(callToImport) {
    // We store the handle to the created lazy component so we can use the same component on every render.
    const componentRef = useRef();
    if (!componentRef .current) {
        componentRef .current = React.lazy(callToImport);
    }
    return [componentRef .current];
}

function MyComponent() {
    const [OtherComponent] = useLazy(() => import('./OtherComponent'));
    return (
       <Route path="/foo" component={OtherComponent} />
    );
}
```

dhedey:
Nice thoughts - I'm been investigating an issue which relates to a variant of this myself!  👍

ashkank83:
@dhedey Is your solution working? I'm experiencing this same issue with no solution. Have created a sandbox [here](https://codesandbox.io/embed/dynamic-comp-loading-sample-9zsc6) to demonstrate the issue. also question on [SO](https://stackoverflow.com/questions/56887650/react-lazy-loaded-component-loosing-its-state-gets-unmounted). Would be great if you have a solution and can help me please?

dhedey:
@ashkank83 - In this case you'd want to make use of a generalization of useLazy discussed above which maintains a dictionary of name to lazily loaded component reference, such as the below:
```
function useLazy(name, callToImport) {
  const lazyComponentsStore = React.useRef({}).current;
  if (!lazyComponentsStore[name]) {
    lazyComponentsStore[name] = React.lazy(callToImport);
  }
  return lazyComponentsStore[name];
}
```

See https://codesandbox.io/s/dynamic-comp-loading-sample-3hdb8

FredyC:
Well, I tried to help but seems nobody is merging docs update around here (this is my 2nd PR) so yea, I am not going to bother to keep this open for another year :/

PR #1423, opened 2018-11-22
Update 2018-05-23-react-v-16-4.md example fix

msteward:
Minor "fix" to the suggested fix in the blog post to default back to empty object if `state.prevProps` isn't yet set. This will help anyone copying the example as-is who has stumbled across this blog post.


gaearon:
I think the right fix would be to define `state = {}`? React would warn anyway if you don't specify it.

gaearon:
Seems okay. Not ideal but also probably shouldn't be too distracting.

PR #1425, opened 2018-11-23
Add additional info about React.memo & React.PureComponent semantic d…

Andarist:


gaearon:
I think this will just lead people to use `PureComponent` ("it compares state too, so it must be better!")

Not sure that's the message we want to send, even if it's technically accurate.

Andarist:
Hm, we can clarify it here that React.memo is preferred over PureComponent. We may maybe change the placement of this clarification, but IMHO it should be clarified somewhere - knowing `PureComponent` semantics I was slightly surprised that `React.memo` is working differently (before I've looked at comparator signature - then it was pretty obvious that this doesn't compare state).

Basically it's similar case to what we had in emotion recently https://github.com/emotion-js/emotion/pull/1027#discussion_r235660774 . Even though we dont recommend using something in certain way, we should document how things work - otherwise it's easier to miss it when looking through documentation and this might waste somebody's time pretty badly.

gaearon:
Maybe:

>`React.memo` is a [higher order component](/docs/higher-order-components.html). It's similar to [`React.PureComponent`](#reactpurecomponent) but for function components instead of classes. Unlike `React.PureComponent` (which compares both props and state), `React.memo` only compares component's props when deciding whether to re-render.

Andarist:
@gaearon I've adjusted the text according to ur suggestion

Andarist:
@gaearon friendly 🏓 😉

rickhanlonii:
Sorry for the delay, closing since these docs no longer exist. If you think this note is still needed, you could submit a new PR to the new PureComponent docs, but tbh this doesn't seem relevant now that almost everyone is on hooks.

PR #1430, opened 2018-11-24
Clarify project setup steps

markerikson:
We had a learner in Reactiflux who accidentally deleted the entire `my-project/src/` folder because they mis-typed the "delete all files in src" step.  I tweaked the instructions to hopefully make it less likely that others would do that in the future.

### DIFF
--- a/content/tutorial/tutorial.md
+++ b/content/tutorial/tutorial.md
@@ -83,7 +83,7 @@ create-react-app my-app
 
 3. Delete all files in the `src/` folder of the new project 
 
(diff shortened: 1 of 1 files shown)


gaearon:
This is how it looks btw.

screen shot 2019-02-09 at 11 40 22 pm

To avoid this kind of weird formatting please check deployment previews. :-) In this case, the fix is to add newlines after "Note"

gaearon:
(The change is good though. I just want to make sure we're mindful of formatting)

PR #1432, opened 2018-11-25
Usage of context value on Lifecycle Method

jyash97:
Don't know, if this might be a good change to add. But I remember that react docs were having this section with legacy API.


rickhanlonii:
Sorry for the delay, going to close because the docs have been rewritten. You could submit a new PR but tbh it doesn't seem worth improving the class component docs at this point.

PR #1437, opened 2018-11-26
Add jsx text type to script tag in add-react-to-a-website.md

svnh:


gaearon:
Thanks for the PR! In this example, it's not necessary because the source doesn't use JSX. JSX is discussed further below on the page.

PR #1438, opened 2018-11-27
Update hooks-intro.md

yesmeck:
Fix https://github.com/reactjs/reactjs.org/issues/1426


gaearon:
If we think it's important to fix then it would need to be changed on all pages, not just here.

PR #1442, opened 2018-11-28
Fix typo on React 16 roadmap

MatanBobi:
> The API design doesn't always make sense when looking at **once** piece in isolation

Fix typo in line 31 (once - one)


gaearon:
Thanks

PR #1446, opened 2018-11-29
Add Suspense and lazy to API reference

gaearon:
Fixes https://github.com/reactjs/reactjs.org/issues/1424

### DIFF
--- a/content/docs/reference-react.md
+++ b/content/docs/reference-react.md
@@ -58,6 +58,13 @@ See [Using React without JSX](/docs/react-without-jsx.html) for more information
 - [`React.createRef`](#reactcreateref)
 - [`React.forwardRef`](#reactforwardref)
 
+### Suspense
+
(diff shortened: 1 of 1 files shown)


content/docs/reference-react.md
```diff
@@ -58,6 +58,13 @@ See [Using React without JSX](/docs/react-without-jsx.html) for more information
 - [`React.createRef`](#reactcreateref)
 - [`React.forwardRef`](#reactforwardref)
 
+### Suspense
+
+Suspense lets components "wait" for asynchronous work to be done before rendering. Currently, Suspense only supports one use case: [code splitting with `React.lazy`](/docs/code-splitting.html#reactlazy). In the future, it will be possible to use Suspense for data fetching, but it's not supported yet. Today, Suspense only works on the client. Server-side rendering support is [in the works](/blog/2018/11/27/react-16-roadmap.html#suspense-for-server-rendering).
```
threepointone:
```suggestion
Suspense lets components "wait" for asynchronous work to be done before rendering. Currently, Suspense only supports one use case: [code splitting with `React.lazy`](/docs/code-splitting.html#reactlazy). In the future, it will be possible to use Suspense for data fetching. Today, Suspense only works on the client. Server-side rendering support is [in the works](/blog/2018/11/27/react-16-roadmap.html#suspense-for-server-rendering).
```

threepointone: "In the future" implies it's not supported yet, yes?

gaearon: I guess so.

PR #1447, opened 2018-11-30
Remove exact versions from blog post

gaearon:
The problem with using exact versions, as pointed out by @sebmarkbage, is that this doesn't let us release minors in the middle. It is very likely we'll need more minors in the middle so let's leave some wiggle room.

I figured this might as well introduce a naming convention for our releases? :-)

### DIFF
--- a/content/blog/2018-11-27-react-16-roadmap.md
+++ b/content/blog/2018-11-27-react-16-roadmap.md
@@ -9,10 +9,12 @@ You might have heard about features like "Hooks", "Suspense", and "Concurrent Re
 
 We plan to split the rollout of new React features into the following milestones:
 
-* React 16.6: [Suspense for Code Splitting](#react-166-suspense-for-code-splitting-shipped) (*already shipped*)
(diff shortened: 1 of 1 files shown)


content/blog/2018-11-27-react-16-roadmap.md
```diff
@@ -32,7 +34,7 @@ We have a single vision for how all of these features fit together, but we're re
 
 The gradual release strategy helps us refine the APIs, but the transitional period when some things aren't ready can be confusing. Let's look at what these different features mean for your app, how they relate to each other, and when you can expect to start learning and using them.
 
-### [React 16.6](/blog/2018/10/23/react-v-16-6.html): Suspense for Code Splitting (shipped)
+### [React 16.6](/blog/2018/10/23/react-v-16-6.html) (shipped): The One with Suspense for Code Splitting
```
gaearon: This would have a nicer ring to it if we dropped "Suspense for". :-) But I want to emphasize the staggered rollout.

PR #1449, opened 2018-11-30
add symbol polyfill to the list

probablyup:
Now that react-is is proliferating the ecosystem for things like
ForwardRef handling and such, symbols have become a requirement.


threepointone:
We don't use Symbol inside React itself, so it's not necessary to have this here. Thanks for your PR!

PR #1456, opened 2018-12-03
Add note for early return when using ref

iamandrewluca:
Add note for components that use `ref` but return early from component

https://codesandbox.io/s/ymvxj5pqmx  
https://stackoverflow.com/a/53596308


lex111:
cc @alexkrolick

PR #1457, opened 2018-12-03
Add -webkit-overflow-scrolling to prism styles

crosscompile:
Adding `-webkit-overflow-scrolling: touch` to the code examples will give them nice momentum scrolling in webkit browsers.

## Before

screen recording 2018-12-03 at 2 34 19 pm

## After

untitled


gaearon:
Thanks.

PR #1458, opened 2018-12-03
Fixing id to 'What Are We Building?' section

tv1ster:
Found an issue in ["What are we building?" section](https://reactjs.org/tutorial/tutorial.html#what-are-we-building) - it's not highlighted, despite clicking or scrolling trough it. Fixed it locally and created this HUGE pull request.


gaearon:
Thanks

PR #1460, opened 2018-12-04
Adds example of useState updates with callbacks

### LINKED ISSUE #1459: [Hooks] useState code should include example of setting state with callback.
The documentation for the `useState` hook could use an example of using a callback to update based on the latest state.  This technique [exists in the docs](https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous) for `setState()`, but not for this hook.  

It was not obvious to me that I could pass a callback instead of just a value.  Once I guessed that it was possible and tried it, I was able to manage async race conditions more smoothly.  But if someone is not familiar with passing a callback to `setState`, or does not assume that it was included in this new API, they may miss out on the technique.  I suggest including this in the [Updating State](https://reactjs.org/docs/hooks-state.html#updating-state) section.

JoelCodes:
Fix #1459


chrisjpatty:
For the moderators, this is a similar pull request to #1417. However, I prefer @JoelCodes comment referencing asynchronous code as the primary purpose for the `useState` updater function.

BartoszKlonowski:
Hello @JoelCodes, thanks for spending time to analyze and contribute with the explanation 👍 
As the callback for previous state is already described in the docs I will close this PR as mentioned in the issue.
Thanks!

PR #1461, opened 2018-12-04
Remove references to deprecated APIs

rmadsen:
The docs reference `componentWillReceiveProps()` and `componentWillUpdate()`, but these are now prefixed with `UNSAFE`, and I doubt we want the official documentation referring to them. I removed the offending line and replaced it with a reference to `getDerivedStateFromProps()` though I'm not sure that mentioning this lifecycle method offers much in the context of this doc.

### DIFF
--- a/content/docs/reconciliation.md
+++ b/content/docs/reconciliation.md
@@ -69,9 +69,7 @@ After handling the DOM node, React then recurses on the children.
 
 ### Component Elements Of The Same Type
 
-When a component updates, the instance stays the same, so that state is maintained across renders. React then updates the props of the underlying component instance to match the new element. If the component implements the `getDerivedStateFromProps()` static method, React will similarly update the component instance with any changed state.
-
-Next, the `render()` method is called and the diff algorithm recurses on the previous result and the new result.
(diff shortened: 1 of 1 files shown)


content/docs/reconciliation.md
```diff
@@ -69,7 +69,7 @@ After handling the DOM node, React then recurses on the children.
 
 ### Component Elements Of The Same Type
 
-When a component updates, the instance stays the same, so that state is maintained across renders. React updates the props of the underlying component instance to match the new element, and calls `componentWillReceiveProps()` and `componentWillUpdate()` on the underlying instance.
+When a component updates, the instance stays the same, so that state is maintained across renders. React then updates the props of the underlying component instance to match the new element. If the component implements the `getDerivedStateFromProps()` static method, React will similarly update the component instance with any changed state.
```
gaearon: I'm worried this will encourage people to implement it, which we usually *don't* want. Let's just drop this sentence?

rmadsen:
Hi folks - I think this is good to go. Happy to either add any additional tweaks you'd like or close this request if you think it is no longer necessary.

PR #1464, opened 2018-12-05
Update context.md

dhruvdutt:
The inversion of control pattern makes the low-level components **less** flexible since they are already computed at higher-level. Only the rendering part is done down in the component tree.


BartoszKlonowski:
Hello @dhruvdutt!
The change you propose in this PR is outdated as that part of the code is no longer present in the latest version of it.
Thank you!

PR #1465, opened 2018-12-05
Update tutorial.md

Suizzle:
Changed "to check what ES6 code compiles to" to "to check what JSX code compiles to" because in this context Babel is compiling JSX to raw JS. 

This is consistent with "Try the Babel REPL to see the raw JavaScript code produced by the JSX compilation step" at https://reactjs.org.


BartoszKlonowski:
Hello @Suizzle!
The part of the docs you update in this PR is no longer present in the latest version of it, so this change is no longer needed.
Thank you!

PR #1466, opened 2018-12-05
Added `this` object to props

rachnac57:


threepointone:
`props` are available in this scope, so this change isn't useful. Thanks for the PR!

PR #1468, opened 2018-12-06
Changed wording about UNSAFE_componentWillReceiveProps deprecation

bvaughn:
I tried to compress this a little bit. (It's still a lot of words to have in one of our yellow "notes".) I also removed the misleading deprecation bit.

### DIFF
--- a/content/docs/reference-react-component.md
+++ b/content/docs/reference-react-component.md
@@ -450,17 +450,13 @@ UNSAFE_componentWillReceiveProps(nextProps)
 
 > Note:
 >
-> Using this lifecycle method often leads to bugs and inconsistencies, and for that reason it is going to be deprecated in the future.
+> We added "UNSAFE_" to this lifecycle's name because using it often leads to bugs and inconsistencies.
 >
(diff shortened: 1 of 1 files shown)


gaearon:
Maybe let's combine this with note at the bottom? And make that first?

Same for other methods.

screen shot 2018-12-06 at 5 51 12 pm

bvaughn:
> Maybe let's combine this with note at the bottom? And make that first?

I don't mind moving the note near the bottom higher up, but I'm not sure "combining" them feels right. This is already a _lot_ of text for what (I think) is intended to be a short format note.

> Same for other methods.

Not sure what this means.

gaearon:
I mean that other methods have the same note at the bottom and it’s very easy to miss. But arguably that’s the most important one and the one people most confused over.

bvaughn:
Oh, so~ move those notes higher for each lifecycle? Sure, I'm not opposed to that.

bvaughn:
Okay I moved those comment blocks higher up.

PR #1469, opened 2018-12-07
remove stub function in gatsby-browser

dpyzo0o:
The stub function in gatsby-browser is no longer needed since the [issue](https://github.com/gatsbyjs/gatsby/issues/6759) has been [fixed](https://github.com/gatsbyjs/gatsby/pull/7011).


BartoszKlonowski:
Hello @dpyzo0o!
I can see that the whole script for gatsby is no longer present in the docs, so there no longer is the adjustment you remove in that PR.
Let me close this one as outdated.

PR #1471, opened 2018-12-07
Updated docs to mention .eslintrc WITH extension

ManoelLobo:
Given that .eslintrc without extension is [deprecated](https://eslint.org/docs/user-guide/configuring#configuration-file-formats), it would be better if the docs suggest to use the file with extension.

This PR adds the `.json` extension where `.eslintrc` is mentioned


BartoszKlonowski:
Hello @ManoelLobo!
I will close this PR as outdated because the blogs at that date are no longer present nor maintained in the latest of docs, also the .eslintrc is no longer in the docs.
However, we welcome PRs for other open issues!

PR #1476, opened 2018-12-08
Remove Unnecessary this.bind(this)

olivercarlson:
Hi,

I updated post to use ES6/ES2018 fat arrow functions which fixes the scoping error that requires the this.bind(this).

confirming that works is codepen link:
https://codepen.io/olivercarlson/pen/MZWYRE

### DIFF
--- a/content/docs/conditional-rendering.md
+++ b/content/docs/conditional-rendering.md
@@ -78,16 +78,14 @@ It will render either `<LoginButton />` or `<LogoutButton />` depending on its c
 class LoginControl extends React.Component {
   constructor(props) {
     super(props);
-    this.handleLoginClick = this.handleLoginClick.bind(this);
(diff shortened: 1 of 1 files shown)


alexkrolick:
Thanks for contribution! For now the docs prefer to use non-proposal stage syntax: https://github.com/reactjs/reactjs.org/blob/master/CONTRIBUTING.md#dont-use-features-that-arent-standardized-yet

olivercarlson:
Ok, thanks for the fast response Alex!

This seems a little confusing though to have this line in the reactjs contributor docs though

"However, you should use arrow functions where they provide a tangible improvement (such as preserving this context inside a component)."

alexkrolick:
Arrow functions are standard in ES2015 - class property initializer syntax is not (yet).

PR #1478, opened 2018-12-09
Idea: Expand all nav sections by default

### LINKED ISSUE #1477: [UI] Expand all navigation sections by default
For example, the prettier docs: https://prettier.io/docs/en/index.html

screen shot 2018-12-08 at 9 51 54 pm

alexkrolick:
Closes #1477

**What**

Expands the right navigation section in the docs, like the tutorial page and some other docs sites.

**Why**

To make it easier to discover other sections.

screen shot 2018-12-08 at 10 07 53 pm

### DIFF
--- a/src/templates/components/Sidebar/Sidebar.js
+++ b/src/templates/components/Sidebar/Sidebar.js
@@ -52,7 +52,7 @@ class Sidebar extends Component {
         {sectionList.map((section, index) => (
           <SectionComponent
             createLink={createLink}
-            isActive={activeSection === section || sectionList.length === 1}
+            isActive={true}
             key={index}
             location={location}
             onLinkClick={closeParentMenu}


sophiebits:
I like this overall although I wonder if we could get away with only auto-expanding Installation when visiting a page inside it (but always auto-expanding the rest). However maybe it's good for people to see those and get reminded that you can get React from a CDN/etc.

@gaearon Curious for your thoughts.

ELBEQQAL94:
Hello there,
Can you help me to understand the problem is?

threepointone:
How do we feel about this now? I'm not sold tbh.

alexkrolick:
I think I'm still for it... let me rebase and see how it looks with all the new sections

alexkrolick:
🤔 The main issues this was addressing are:

1. Sometimes you don't know that there is content related to what you are looking at in another, collapsed section
1. You don't really know where the thing you are looking for lives in the hierarchy
1. You are new to the docs and getting your bearings as far as where to start reading

BartoszKlonowski:
@sophiebits @alexkrolick Is this idea still valid? In the latest `react.dev` there already is a navigation panel expanded by default.

rickhanlonii:
> only auto-expanding Installation when visiting a page inside it

This is how the new react.dev site works, closing since it's fixed in the new site and we're not updating the old site.

PR #1483, opened 2018-12-12
Add React Next 2019

adamkleingit:
Add React Next 2019
Move past conferences to relevant section


harish-sethuraman:
The conference page seems to be updated right now. Thanks for sending PR and sorry for the late turn around time.

PR #1488, opened 2018-12-14
Add an example with propTypes on functional component

aibolik:
I mentioned a misconception among JavaScript developers(mostly beginners, but not limiting with them) - they tend to think that properties can be assigned only to classes and there is no way(at least it won't work) if you assign them to functions. This translates to React too, so some developers think that functional components can't have propTypes set to them.

In order to break this misconception, I've added an example of setting propTypes to functional component(I used arrow function, so that they can be sure that it also be used with arrow functions).

If you feel I am write, I will be happy if you merge it.

And of course, any feedback is welcomed.


aibolik:
@facebook-github-bot just sumbitted it. Thanks :)

BartoszKlonowski:
Hello @aibolik, thank you for the contribution!
As the `PropTypes` are being deprecated in v15.5 this addition is no longer required, so let me close this as outdated.
Please feel invited to create new PRs for open issues.

PR #1489, opened 2018-12-15
Update forms.md

pure-js:
Use object destructuring


threepointone:
This seems like a nit, and I'm inclined to reject it. Thanks for your PR!

PR #1491, opened 2018-12-18
Changing state to prevState

kevindelsh:
In the second form of setState() that accepts a function, the first argument to the function is the previous state. In the code example though, it is called state which is counter intuitive to beginners. function(prevState, props) makes more sense IMHO.
I'm new by the way and don't know your rules and conventions. I apologize in advance if I should have proceeded differently.


threepointone:
This hasn't turned out to be a problem in practice. Thank you for your PR!

PR #1496, opened 2018-12-20
update the date for the conference

apherio:
This is to update the date for the conference React India 2019 in Goa, India


manjula91:
@gaearon @sophiebits @alexkrolick

PR #1498, opened 2018-12-22
Misleading useReducer lazy init example

FredyC:
For more context see: https://github.com/reactjs/rfcs/pull/68#issuecomment-449568020

The example just assumes that `initialCount` value will be passed from a parent. When it isn't, the state becomes `undefined` and later `NaN` when trying to inc/dec a value.

### DIFF
--- a/content/docs/hooks-reference.md
+++ b/content/docs/hooks-reference.md
@@ -196,7 +196,7 @@ function reducer(state, action) {
   }
 }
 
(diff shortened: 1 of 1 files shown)


FredyC:
Well, one would think that such small docs fixes would be merged right away. I am not going to bother with resolving conflicts since it would be probably useless.

PR #1502, opened 2018-12-26
updated first example

sathya9897:
fixed the bug in first example.


threepointone:
fixed separately, thanks for your PR!

PR #1506, opened 2018-12-27
Update Hooks Docs to refer to the correct react version

ayox:
Hooks documentation had the wrong version of the alpha release for hooks.


threepointone:
This was updated to point ot the right version (16.8). Thanks for your PR!

PR #1508, opened 2018-12-27
Remove old conferences from the list

ronanmathew:
PR to remove conferences that have now passed from the list.


gaearon:
Thanks for the PR! Seems like this was done separately so I'm closing.

PR #1512, opened 2018-12-28
Update hooks-reference.md

yy8597:


harish-sethuraman:
Thanks for sending a PR. Seems like this was fixed. Sorry for the late turn around time.

PR #1514, opened 2018-12-29
Update web-components.md

arjunyel:
Removed unnecessary line


threepointone:
I think we kept that line explicitly to be clear about the level of support. We'll revisit it soon in some fashion. Thanks for your PR!

PR #1520, opened 2019-01-01
Add React.JS Girls to the list of React Meetups

M0nica:
Added the React.JS Girls Berlin and London chapters to the list of React Meetups.


alexkrolick:
Thanks!

PR #1521, opened 2019-01-02
clarify mountNode with comments

alexkrolick:
This is still a bit opaque but it's difficult to fix with the current way
CodeEditor works:
- long-term CodeEditor.js could take a node reference instead of defining
  mountNode internally
- could also use document.createElement to define the target in the code,
  but this could be mislead people to think this is required instead of
  using an existing reference

see #1017, #1018, #1019


gaearon:
I don't think this is very clear. Could even be more confusing than what we had.

Can we instead just make it target a CSS class?

gaearon:
I'll do it

PR #1526, opened 2019-01-02
Clarify DOM container in homepage examples

### LINKED ISSUE #1017: Explanation for "mountNode" on home page code examples would help

gaearon:
Supersedes https://github.com/reactjs/reactjs.org/pull/1521.

Fixes #1017.

Same as https://github.com/reactjs/reactjs.org/pull/1019#issuecomment-412757325 in spirit but doesn't break the website.

### DIFF
--- a/content/home/examples/an-application.md
+++ b/content/home/examples/an-application.md
@@ -1,6 +1,7 @@
 ---
 title: An Application
 order: 2
+domid: todos-example
 ---
 
 Using `props` and `state`, we can put together a small Todo application. This example uses `state` to track the current list of items as well as the text that the user has entered. Although event handlers appear to be rendered inline, they will be collected and implemented using event delegation.
--- a/src/components/CodeEditor/CodeEditor.js
+++ b/src/components/CodeEditor/CodeEditor.js
@@ -27,8 +27,6 @@ class CodeEditor extends Component {
   }
 
   componentDidMount() {
-    // Initial render() will always be a no-op,
-    // Because the mountNode ref won't exist yet.
     this._render();
   }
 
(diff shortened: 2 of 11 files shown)


content/home/examples/an-application.md
```diff
@@ -1,6 +1,7 @@
 ---
 title: An Application
 order: 2
+domid: todos-example
```
alexkrolick: Perfect

src/components/CodeEditor/CodeEditor.js
```diff
@@ -27,8 +27,6 @@ class CodeEditor extends Component {
   }
 
   componentDidMount() {
-    // Initial render() will always be a no-op,
-    // Because the mountNode ref won't exist yet.
     this._render();
```
gaearon: Technically this would still not work on first render if parents weren't already in the documented. I can add a setTimeout but don't care enough about this (?)

gaearon:
Going to assume this isn't controversial and get it in. I really dislike the current comment solution so I'd like to replace it. :-)

Thanks @alexkrolick for getting the ball rolling!

pcgeek86:
I prefer this solution vs. the comment solution. Thanks to both of you for updating this. @gaearon @alexkrolick

PR #1529, opened 2019-01-03
Correct package number

StJohn3D:
I tried installing `npm i react@v16.7.0-alpha` but that didn't work.
After finding [this](https://stackoverflow.com/a/53009407/5900250) answer I was able to successfully install it.


saranshkataria:
Since the documentation is referring to react version, and not specifying the packages, I think it solves the purpose. Also, they are in 16.8 now and the docs have been updated for the same. So this PR is not needed anymore. Closing this out! Thanks.

PR #1530, opened 2019-01-04
Fix the "Releases" link in the "Getting Started" page

qbahers:
`https://github.com/facebook/react` -> `https://github.com/facebook/react/releases`.


lex111:
@qbahers thank you!

PR #1535, opened 2019-01-06
I don't like the comment text.

agamemnus:
It doesn't just create a new variable. It creates two new variables. Then, let's say what they actually do.


BartoszKlonowski:
Hello @agamemnus, thank you for the contribution!
The code you that you are changing is no longer present in the current documentation of React, so let me close this PR as already outdated.
But please feel invited to create new PRs for other open issues.

PR #1542, opened 2019-01-10
Remove event param from arrow function example

traumverloren:
As a newer JS/React dev, I found it confusing that the event was being passed in the arrow function in the callback, when it is not used in the referenced function. 

I updated the code example and removed the event param, so that other devs might not also get confused that the event needs to be passed in this example, even though it isn't used in the referenced callback.

screenshot 2019-01-10 at 09 08 40


threepointone:
good catch, thank you!

PR #1543, opened 2019-01-10
Add 16.7 to versions

eps1lon:
For the 16.6 docs version I used the last production build on netlify before #1492 was merged.


gaearon:
thanks

PR #1544, opened 2019-01-10
rename useImperativeMethods -> useImperativeHandle

threepointone:
A followup from https://github.com/facebook/react/pull/14565; don't merge this till we land that one. thanks!

### DIFF
--- a/content/docs/hooks-reference.md
+++ b/content/docs/hooks-reference.md
@@ -324,7 +324,7 @@ Note that `useRef()` is useful for more than the `ref` attribute. It's [handy fo
 ### `useImperativeHandle`
 
 ```js
-useImperativeHandle(ref, createInstance, [inputs])
+useImperativeHandle(ref, createHandle, [inputs])
 ```
 
 `useImperativeHandle` customizes the instance value that is exposed to parent components when using `ref`. As always, imperative code using refs should be avoided in most cases. `useImperativeHandle` should be used with `forwardRef`:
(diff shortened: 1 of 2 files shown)


content/docs/hooks-reference.md
```diff
@@ -321,18 +321,18 @@ function TextInputWithFocusButton() {
 
 Note that `useRef()` is useful for more than the `ref` attribute. It's [handy for keeping any mutable value around](/docs/hooks-faq.html#is-there-something-like-instance-variables) similar to how you'd use instance fields in classes.
 
-### `useImperativeMethods`
+### `useImperativeHandle`
 
 ```js
-useImperativeMethods(ref, createInstance, [inputs])
+useImperativeHandle(ref, createInstance, [inputs])
```
gaearon: Maybe `createHandle`

content/docs/hooks-reference.md
```diff
@@ -321,18 +321,18 @@ function TextInputWithFocusButton() {
...
 
-`useImperativeMethods` customizes the instance value that is exposed to parent components when using `ref`. As always, imperative code using refs should be avoided in most cases. `useImperativeMethods` should be used with `forwardRef`:
+`useImperativeHandle` customizes the instance value that is exposed to parent components when using `ref`. As always, imperative code using refs should be avoided in most cases. `useImperativeHandle` should be used with `forwardRef`:
 
 ```js
 function FancyInput(props, ref) {
   const inputRef = useRef();
-  useImperativeMethods(ref, () => ({
+  useImperativeHandle(ref, () => ({
     focus: () => {
```
gaearon: Maybe we should change this to promote the static call style. You’ll need to discuss this with Seb.

threepointone: static call style? like, `React.` ? I'll reach out.

gaearon: No, I mean not using an object at all. Although I guess in that case `useImperativeHandle` would be unnecessary in the first place. Disregard.

bvaughn:
Since I just published a new alpha, I'm rolling out this change also

PR #1553, opened 2019-01-14
Extended JS environment requirements polyfills list with a promise

Kazimirkas:
IE11 needs a Promise polyfill for React 16.7. (not only Set and Map polyfills)
If you are using Lazy and Suspense feature, IE throws a null error. In 'readLazyComponentType' function implementation.

https://github.com/reactjs/reactjs.org/issues/1552

### DIFF
--- a/content/docs/reference-javascript-environment-requirements.md
+++ b/content/docs/reference-javascript-environment-requirements.md
@@ -11,6 +11,7 @@ React 16 depends on the collection types [Map](https://developer.mozilla.org/en-
 A polyfilled environment for React 16 using core-js to support older browsers might look like:
 
 ```js
+import 'core-js/es6/promise';
 import 'core-js/es6/map';
 import 'core-js/es6/set';
 


alexkrolick:
👍thanks

aweary:
We should probably explain this a bit more, as I don’t think Promise isn’t strictly required unless you’re using `React.lazy` with dynamic import. 

Internally React just models it as “thenable”, which could be a regular object with `then` method.

cc @gaearon

Kazimirkas:
@aweary Would It be correct to add after this sentence
`...consider including a global polyfill in your bundled application, such as core-js or babel-polyfill. `
a sentence like: 
`Also, If you’re using React.lazy (that requires a Promise or dynamic import), don't forget to add a polyfilled Promise` ?

gaearon:
@alexkrolick for future reference, "JS requirements" can't change without a major version. If we add requirements midway it looks like a breaking change. We should avoid this kind of confusion.

alexkrolick:
Definitely need to add a note to the lazy docs. Otherwise technically it is broken without a polyfill. It's also in the IE11 polyfill for CRA: https://github.com/facebook/create-react-app/blob/master/packages/react-app-polyfill/ie11.js

PR #1557, opened 2019-01-15
Add docs for useDebugValue

bvaughn:


content/docs/hooks-reference.md
```diff
@@ -353,3 +354,31 @@ Prefer the standard `useEffect` when possible to avoid blocking visual updates.
...
 >
 > If you're migrating code from a class component, `useLayoutEffect` fires in the same phase as `componentDidMount` and `componentDidUpdate`, so if you're unsure of which effect Hook to use, it's probably the least risky.
+
+### `useDebugValue`
+
+```js
+useDebugValue(value)
+```
+
+`useDebugValue` can be used to provide a label for the React DevTools to display beside of a custom hook.
```
threepointone: "can be used to display a label for custom hooks in React DevTools"?

content/docs/hooks-reference.md
```diff
@@ -353,3 +354,31 @@ Prefer the standard `useEffect` when possible to avoid blocking visual updates.
...
+
+For example, consider the `useFriendStatus` custom hook described in ["Building Your Own Hooks"](/docs/hooks-custom.html):
+
+```js{6-8}
+function useFriendStatus(friendID) {
+  const [isOnline, setIsOnline] = useState(null);
+
+  // ...
+
+  // Show a label in DevTools beside of this hook
```
threepointone: "next to this hook"?

content/docs/hooks-reference.md
bvaughn: I dig it

content/docs/hooks-reference.md
bvaughn: I dig it

gaearon:
We haven't cut an alpha that includes it, have we?

bvaughn:
> A screenshot would help explain it.

I thought of this but it seems that we want to be a bit cautious about encouraging use of this hook, so I thought maybe it was better to leave the picture out for now.

PR #1560, opened 2019-01-16
pseudo setState callback tip

escorponox:
It is common to use `setState` callback on class components to ensure the state has been updated. But there is no `setState` callback on state hooks. This is a new way of think and it may be hard to new hooks users.

Maybe this tip should be placed on State Hooks page, but here is where you mention that `useEffect` guarantees the DOM (and the state) has been updated.

This is an example sandbox: https://codesandbox.io/s/k95yv04rr


threepointone:
This isn't actually a great way to do this, since they're not synchronized. Thank you for the PR!

PR #1561, opened 2019-01-16
Add YouTube Meta into examples

alik0211:


gaearon:
Thanks, but we're keeping this page focused on React. No Redux.

PR #1569, opened 2019-01-17
Tweak useMemo wording

gaearon:
In response to feedback.


markerikson:
Commenting after [this discussion on Twitter](https://twitter.com/dan_abramov/status/1085927944210563073).

I'm mostly okay with the wording here, but it would still be helpful to call out particular scenarios where `useMemo` _might_ not memoize (such as the possible future handling of virtualized rows).  I realize this can be tough to do without tying yourself into particular implementation details, but it would really be beneficial.

gaearon:
My concern is then we over-promise and people will think React already treats e.g. React Virtualized specially. Or even that you don't need it.

gaearon:
ok I can't do better than this. I'll roll with it.

screen shot 2019-01-17 at 5 09 01 pm

markerikson:
Yeah, I'm good with that. Thank you for addressing some of my concerns :)

suchipi:
This is great!!

PR #1575, opened 2019-01-19
Change copyright year

swarnava:


swarnava:
Let me know if there is a way to change it to dynamic so that it gets updated every year automatically :)

srsholmes:
I already addressed this in my PR, with dynamic year 👍 
https://github.com/reactjs/reactjs.org/pull/1546

PR #1583, opened 2019-01-22
Update conferences.md | append React Conf Iran for May 30, 2019

ahmadina:
Append React Conf Iran for May 30, 2019


rickhanlonii:
Sorry we missed this, closing since it's 2025 now :(

PR #1587, opened 2019-01-23
Add Columbia, MD meetup group

cicavey:


rickhanlonii:
Sorry for missing this in 2018. Looks like this last met in 2019, so going to close but if it's still active let me know.

PR #1590, opened 2019-01-24
When use Class.contextType, React.createContext should be call in separate files

Sylvenas:
16.6 Context API not working in class component,Issus,[13969](https://github.com/facebook/react/issues/13969),the Class.contextType is a greet improve,but the documentation is easy to guide developers to make mistakes.

### DIFF
--- a/content/blog/2018-10-23-react-v-16-6.md
+++ b/content/blog/2018-10-23-react-v-16-6.md
@@ -72,7 +72,7 @@ class MyClass extends React.Component {
 
 > Note:
 >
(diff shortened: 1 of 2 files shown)


Sylvenas:
Fixed the grammar & add an demo on CodeSandBox.

aruprakshit:
I met the same error, when I have child and context created in the same file. The solution written here solved my problem. But I would like to know if this is the standard or there will be a fix from react team?

Sylvenas:
@aruprakshit My team met this problem, but I think the real issue is the documentation.

BartoszKlonowski:
As the related issue is closed and the addition targets outdated documentation, I will close this PR.

PR #1591, opened 2019-01-24
Uselayouteffect server render warning

threepointone:
Some content to iterate on. This is a warning that shows up when you try to server render a component using 'useLayoutEffect'. related to https://github.com/facebook/react/issues/14363 

I need to copyedit the small things, like capitalizing React etc as well

### DIFF
--- a/content/warnings/uselayouteffect-server-warning.md
+++ b/content/warnings/uselayouteffect-server-warning.md
@@ -1,0 +1,25 @@
+---
+title: useLayoutEffect does nothing on the server 
+layout: single
+permalink: warnings/uselayouteffect-server-warning.html
+---
+
+
+You are probably here because you saw this warning in your server logs when rendering a react component. 
+
+useLayoutEffect does nothing on the server, because its effect cannot
+be encoded into the server renderer's output format. This will lead
+to a mismatch between the initial, non-hydrated UI and the intended
+UI. To avoid this, useLayoutEffect should only be used in
+components that render exclusively on the client.
+
+An example of a component that could do this - 
+```
+function Foo(){
+  useLayoutEffect(() => {
+    
+  })
+  return <div></div>
+}
+```
+
--- a/content/warnings/uselayouteffect-server-warning.md
+++ b/content/warnings/uselayouteffect-server-warning.md
@@ -4,22 +4,46 @@ layout: single
 permalink: warnings/uselayouteffect-server-warning.html
 ---
 
-
 You are probably here because you saw this warning in your server logs when rendering a react component. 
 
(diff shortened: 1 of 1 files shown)


content/warnings/uselayouteffect-server-warning.md
```diff
@@ -0,0 +1,47 @@
...
+> useLayoutEffect does nothing on the server, because its effect cannot be encoded into the server renderer's output format. This will lead to a mismatch between the initial, non-hydrated UI and the intended UI. To avoid this, useLayoutEffect should only be used in components that render exclusively on the client.
+
+`useLayoutEffect` is a hook used to read layout from the dom rerender. They're scheduled to run right after updating the dom, but just before the browser has painted. An example of such a component -
+
+```jsx
+function Foo() {
+  let ref = useRef();
+  useLayoutEffect(() => {
+    if (ref.clientWidth > 100) {
+      ref.style.width = 100;
```
trueadm: I know this is somewhat unrelated to the PR, but is this a pattern we're promoting, I though the user would do a state update with the above width, then alter that as a prop on the `div`?

content/warnings/uselayouteffect-server-warning.md
```diff
@@ -0,0 +1,47 @@
...
+You're probably here because you saw this warning in your server logs when rendering a react component.
+
+> Warning:
+>
+> useLayoutEffect does nothing on the server, because its effect cannot be encoded into the server renderer's output format. This will lead to a mismatch between the initial, non-hydrated UI and the intended UI. To avoid this, useLayoutEffect should only be used in components that render exclusively on the client.
+
+`useLayoutEffect` is a hook used to read layout from the dom rerender. They're scheduled to run right after updating the dom, but just before the browser has painted. An example of such a component -
+
+```jsx
+function Foo() {
```
trueadm: I don't think we should use function component names like `Foo` and `Bar`, we should name them to somewhat descriptive component names so there's a better connection with the user reading this document.

content/warnings/uselayouteffect-server-warning.md
threepointone: I could change it to that. really though, I'd love a better example. I was considering a tooltip-ish thing maybe.

content/warnings/uselayouteffect-server-warning.md
threepointone: agreed

gaearon: Great point @trueadm 👍

rickhanlonii:
Hey @threepointone and @trueadm :) 

Closing since we removed this warning 👍

PR #1592, opened 2019-01-24
Hooks can only be used in function components

threepointone:
Some content to iterate on. This link gets included in the error message when hooks are used outside function component bodies. We also include notes on how it might happen if you've installed react incorrectly.

### DIFF
--- a/content/warnings/hooks-in-function-components.md
+++ b/content/warnings/hooks-in-function-components.md
@@ -1,0 +1,56 @@
+---
+title: Hooks can only be called inside the body of a function component.
+layout: single
+permalink: warnings/hooks-in-function-components.html
+---
+
+You are probably here because you got the following error message:
+
+> Hooks can only be called inside the body of a function component.
+
+This usually means that you've tried to call one on react's [hooks]() from outside the body of a [function component](). This breaks the [Rules of Hooks](). Some examples of code that could trigger this error - 
+
+```jsx
+// calling a hook inside an event handler
+function App(){
+  return <button onClick={() => {
+    useEffect(() => {
+      console.log('clicked')
+    })
+  }}> 
+    click me
+  </button>
+} 
+```
+In this case, you don't need the `useEffect` call, and can log the message directly. Rewriting teh above example - 
+```jsx
+function App(){
+  return <button onClick={() => {
+    console.log('clicked')
+  }}> 
+    click me
+  </button>
+} 
+```
+
+```jsx
+// calling a hook inside useMemo
+function App(){
+  const value = useMemo(() => {
+    const [counter, setCounter] = useState(0)
+    return counter
+  }, [])
+  return value
+}
+```
+
+```jsx
+// calling a hook inside .memo's comparator
+const App = React.Memo((props) => {
+  return props.left + props.right
+}, (a, b) => a.left === b.left && a.right === b.right)
+```
+
+In rare edge cases, you could also get this error in case you've configured your app incorrectly. 2 cases we've come across - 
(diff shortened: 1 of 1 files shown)


content/warnings/hooks-in-function-components.md
```diff
@@ -0,0 +1,78 @@
...
+title: Hooks can only be called inside the body of a function component.
+layout: single
+permalink: warnings/hooks-in-function-components.html
+---
+
+You are probably here because you got the following error message:
+
+> Hooks can only be called inside the body of a function component.
+
+This usually means that you've tried to call one on React's [Hooks](/docs/hooks-intro.html) from outside the body of a [function component](/docs/components-and-props.html#function-and-class-components). This breaks the [Rules of Hooks](/docs/hooks-rules.html). Some examples of code that could trigger this error -
```
gaearon: The dash at the end looks weird. Maybe colon?

gaearon: "one on" typo

gaearon: This paragraph has a lot of links. I'd just leave "rules of Hooks" as a link here.

content/warnings/hooks-in-function-components.md
```diff
@@ -0,0 +1,78 @@
...
+
+### Calling a hook inside an event handler:
```
gaearon: We don't put colons in headers

content/warnings/hooks-in-function-components.md
```diff
@@ -0,0 +1,78 @@
...
+        });
+      }}
+    >
+      click me
+    </button>
+  );
+}
+```
+
+In this case, you don't need the `useEffect` call, and can log the message directly. Rewriting the above example -
```
gaearon: dash -> colon. Also, this sentence is very abrupt. Rewriting the above example.. what?

Alternative suggestion.

>In this case, you don't need the `useEffect` call, and can log the message directly. Here is the correct version:

content/warnings/hooks-in-function-components.md
gaearon: We capitalize Hooks everywhere in the docs

content/warnings/hooks-in-function-components.md
```diff
@@ -0,0 +1,78 @@
...
+        console.log("clicked");
+      }}
+    >
+      click me
+    </button>
+  );
+}
+```
+
+### Calling a hook inside useMemo:
```
gaearon: How to fix?

content/warnings/hooks-in-function-components.md
```diff
@@ -0,0 +1,78 @@
...
+function App() {
+  const value = useMemo(() => {
+    const [counter, setCounter] = useState(0);
+    return counter;
+  }, []);
+  return value;
+}
+```
+
+### Calling a hook inside .memo's comparator:
```
gaearon: This is likely so rare it's not worth mentioning and it's more confusing to see it than not.

content/warnings/hooks-in-function-components.md
```diff
@@ -0,0 +1,78 @@
...
+  (a, b) => {
+    useEffect(() => {
+      // ...
+    });
+    return a.left === b.left && a.right === b.right;
+  }
+);
+```
+
+NB: In rare edge cases, you could also get this error in case you've configured your app incorrectly. 2 cases we've come across -
```
gaearon: These are not rare cases. It's the primary case that causes this error. It should be at the top with an explanation for how to fix it.

threepointone:
This was fixed by https://reactjs.org/warnings/invalid-hook-call-warning.html

PR #1593, opened 2019-01-24
Update docs for stable Hooks

gaearon:
Work in progress. Don't merge before the release.

**ALSO DON'T MERGE UNTIL BLOG POST DATE IS SET TO THE REAL ONE**


content/docs/hooks-faq.md
```diff
@@ -19,7 +19,9 @@ This page answers some of the frequently asked questions about [Hooks](/docs/hoo
 -->
 
 * **[Adoption Strategy](#adoption-strategy)**
+  * [Which versions of React include Hooks?](#which-versions-of-react-include-hooks)
   * [Do I need to rewrite all my class components?](#do-i-need-to-rewrite-all-my-class-components)
+  * [What can I do with Hooks that I couldn't with classes?](#what-can-i-do-with-hooks-that-i-couldnt-with-classes)
   * [How much of my React knowledge stays relevant?](#how-much-of-my-react-knowledge-stays-relevant)
   * [Should I use Hooks, classes, or a mix of both?](#should-i-use-hooks-classes-or-a-mix-of-both)
   * [Do Hooks cover all use cases for classes?](#do-hooks-cover-all-use-cases-for-classes)
```
Avi98: Hi @gaearon, I know its still in progress but still I just wanted to highlight that in the answer of the above question some things need to be changed as I have mentioned in [issue](https://github.com/reactjs/reactjs.org/issues/1585)

gaearon: I think I already edited that?

Avi98: I'm sorry just saw that. I'll close that issue

content/docs/hooks-custom.md
```diff
@@ -6,14 +6,14 @@ next: hooks-reference.html
 prev: hooks-rules.html
 ---
 
-*Hooks* are an upcoming feature that lets you use state and other React features without writing a class. They're currently in React v16.8.0-alpha.1.
+*Hooks* let you use state and other React features without writing a class.
```
sophiebits: Do you think "are a new feature that lets" would be better? I ask mainly because it makes it more obvious why the Hooks docs are in their own section and not properly incorporated through the entire docs.

gaearon: I was a bit annoyed by "new feature that lets you use [...] React features" but I agree with the point. Any other suggestions?

sophiebits: Maybe: “Hooks are a new addition in React 16.8. They let”

content/blog/2019-01-30-react-v16.8.0.md
```diff
@@ -0,0 +1,99 @@
...
+* React Test Renderer
+* React Shallow Renderer
+
+Note that **to enable Hooks, all React packages need to be 16.8.0 or higher**. Hooks won't work if you forget to update, for example, React DOM.
+
+**React Native will fully support Hooks in its next stable release.**
+
+## Tooling Support
+
+React Hooks are now fully supported by React DevTools. They are also supported in the latest Flow and TypeScript definitions for React. We recommend using a new [lint rule](https://www.npmjs.com/package/eslint-plugin-react-hooks) to enforce best practices with Hooks. It will soon be included into Create React App by default.
```
sophiebits: How about:
```We strongly recommend enabling a new [lint rule called `eslint-plugin-react-hooks`](https://www.npmjs.com/package/eslint-plugin-react-hooks) to enforce best practices with Hooks.```

CRA soon? Why not now?

File: content/blog/2019-01-30-react-v16.8.0.md [hunk 4]
```diff
@@ -0,0 +1,99 @@
...
+
+## What's Next
+
+We described our plan for the next months in the recently published [React Roadmap](/blog/2018/11/27/react-16-roadmap.html).
+
+Note that React Hooks don't cover *all* use cases for classes yet but they're [very close](/docs/hooks-faq.html#do-hooks-cover-all-use-cases-for-classes). Currently, only `getSnapshotBeforeUpdate()` and `componentDidCatch()` methods don't have equivalent Hooks APIs, and these lifecycles are relatively uncommon. If you want, you should be able to use Hooks in most of the new code you're writing.
+
+Even while Hooks were in alpha, we saw many [interesting examples](https://medium.com/@drcmda/hooks-in-react-spring-a-tutorial-c6c436ad7ee4) of custom Hooks for animations, forms, subscriptions, integrating with other libraries, and so on. Our goal is to empower the React community to write components and Hooks that deliver great user and developer experience. We can't wait to see what you'll create next!
```
sophiebits: "Our goal…" is a little fluffy. I might write:

> We're excited about Hooks because they make code reuse easier, helping you write your components in a simpler way and make great user experiences.

(Take or leave, I don't mind the current one either.)

content/blog/2019-01-30-react-v16.8.0.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Thanks
+
+We'd like to thank everybody who commented on the [Hooks RFC](https://github.com/reactjs/rfcs/pull/68) for sharing their feedback. We've read all of your comments and made some adjustments to the final API based on them.
+
+## Installation
```
sophiebits: Add lint rule to this section?

content/blog/2019-01-30-react-v16.8.0.md
```diff
@@ -0,0 +1,99 @@
...
+
+### React Test Renderer
+
+* Support Hooks in the shallow renderer. ([@trueadm](https://github.com/trueadm) in [#14567](https://github.com/facebook/react/pull/14567))
+* Fix shallow renderer passing incorrect state to `shouldComponentUpdate` in the presence of `getDerivedStateFromProps`. ([@chenesan](https://github.com/chenesan) in [#14613](https://github.com/facebook/react/pull/14613))
+
+### ESLint Plugin: React Hooks
+
+* Initial [release](https://www.npmjs.com/package/eslint-plugin-react-hooks). ([@calebmer](https://github.com/calebmer) in [#13968](https://github.com/facebook/react/pull/13968))
+* Fix reporting after encountering a loop. ([@calebmer](https://github.com/calebmer) and [@Yurickh](https://github.com/Yurickh) in [#13968](https://github.com/facebook/react/pull/13968))
```
sophiebits: what about https://github.com/facebook/react/pull/14040 🙃

content/blog/2019-01-30-react-v16.8.0.md
gaearon: I guess I was trying to say our job is to focus on the primitives

content/blog/2019-02-04-react-v16.8.0.md
```diff
@@ -0,0 +1,159 @@
...
+
+Note that **to enable Hooks, all React packages need to be 16.8.0 or higher**. Hooks won't work if you forget to update, for example, React DOM.
+
+**React Native will fully support Hooks in its next stable release.**
+
+The final TypeScript and Flow definitions are currently in review and will be available soon.
+
+## Tooling Support
+
+React Hooks are now fully supported by React DevTools. They are also supported in the latest Flow and TypeScript definitions for React. We strongly recommend enabling a new [lint rule called `eslint-plugin-react-hooks`](https://www.npmjs.com/package/eslint-plugin-react-hooks) to enforce best practices with Hooks. It will soon be included into Create React App by default.
```
reaysawa: I'll take the opportunity to ask a question if you don't mind. I noticed that hooks' values aren't easily editable like class properties in the DevTools, which was a surprise for me.

react

Although they could be considered "fully-supported" in the sense of what's possible to do in regards to implementation details, I'd think the passage also implies feature-parity with the way people are used to interact with class components.

The way the sentence is structured implies a solved problem. Is it the case? Or will hooks' values also be as flexible in the future?

content/blog/2019-02-04-react-v16.8.0.md
```diff
@@ -0,0 +1,159 @@
...
+* React DOM
+* React DOM Server
+* React Test Renderer
+* React Shallow Renderer
+
+Note that **to enable Hooks, all React packages need to be 16.8.0 or higher**. Hooks won't work if you forget to update, for example, React DOM.
+
+**React Native will fully support Hooks in its next stable release.**
+
+The final TypeScript and Flow definitions are currently in review and will be available soon.
```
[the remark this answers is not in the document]
Jessidhia: To clarify, the DefinitelyTyped publish bot does an auto-publish every time a PR is merged. There were 2 PRs to update the API so the first publish to have both was 16.8.1.

content/docs/hooks-reference.md
```diff
@@ -202,31 +204,59 @@ function Counter({initialCount}) {
...
+```js{3}
+  const [state, dispatch] = useReducer(
+    reducer,
+    {count: initialCount}
+  );
+```
+
+>Note
+>
+>React doesn’t use the `state = initialState` argument convention popularized by Redux. The initial value sometimes needs to depend on props and so is specified from the Hook call instead. If you feel strongly about this, you can write `state = initialState` both in the reducer and inside the `useReducer` destructuring assignment, but it's not encouraged.
```
sophiebits: FWIW I found this very hard to parse due to my unfamiliarity. Maybe…

> To specify initial state, React doesn't use the default arguments pattern `function reducer(action, state = {count: 0})` popularized by Redux.

---

> you can write `state = initialState` both in the reducer and inside the `useReducer` destructuring assignment

can you not put it only in the reducer and let the initial state default to undefined? why?

content/docs/hooks-faq.md
```diff
@@ -322,6 +323,22 @@ function ScrollView({row}) {
 
 This might look strange at first, but an update during rendering is exactly what `getDerivedStateFromProps` has always been like conceptually.
 
+### Is there something like forceUpdate?
+
+Both `useState` and `useReducer` Hooks [bail out of updates](/docs/hooks-reference.html#bailing-out-of-a-state-update) if the next value is the same as the previous one. Mutating state in place and calling `setState` will not cause a re-render.
```
sophiebits: "bail out" feels jargony – how do you feel about "skip rerendering"?

content/docs/hooks-rules.md
```diff
@@ -28,7 +28,7 @@ By following this rule, you ensure that all stateful logic in a component is cle
 We released an ESLint plugin called [`eslint-plugin-react-hooks`](https://www.npmjs.com/package/eslint-plugin-react-hooks) that enforces these two rules. You can add this plugin to your project if you'd like to try it:
 
 ```bash
-npm install eslint-plugin-react-hooks@next
+npm install eslint-plugin-react-hooks
```
DAB0mB: What about a plugin that automatically provides hooks with the right arguments? Such as `useMemo` and `useCallback`. It's also mentioned in the docs:

```
Note

The array of inputs is not passed as arguments to the callback. Conceptually, though, that’s what they represent: every value referenced inside the callback should also appear in the inputs array. In the future, a sufficiently advanced compiler could create this array automatically.
```

Check out a plugin of mine (still experimental): https://github.com/DAB0mB/babel-plugin-react-persist Would be nice to add it to the docs rather than people just searching around, or even worse, not using hooks when necessary.

Jessidhia: It is a part of https://github.com/facebook/react/pull/14636

bvaughn:
Dan's gone for the day and we don't have write access to his repo so I've forked and opened a new PR #1629

PR #1600, opened 2019-01-26
Fix wrong references in the examples of extracting components section

micnic:
This PR fixes some wrong references in the examples of the "Extracting Components" section, when splitting a bigger component in smaller components the structure of props was different and did not match the original component.


micnic:
was fixed meanwhile or I missed something

PR #1602, opened 2019-01-27
added: notice for many nodes in react

itsjzt:


threepointone:
This doesn't seem right, and doesn't  consider fragments. I don't think we want to add details on the edge case here. Thank you for the PR!

PR #1606, opened 2019-01-30
button outline fix

KARTIK01:
On click of the button, outline comes in docs which doesn't look good.

Before:

screenshot 2019-01-30 at 11 57 08 am

After : 

screenshot 2019-01-30 at 11 58 52 am


lex111:
Please read it :)
http://www.outlinenone.com/

KARTIK01:
@lex111 then i think i should close this PR.

PR #1607, opened 2019-01-30
Update hooks-effect.md

segmentationfaulter:
What I understand, these are the effects which run *after* every re-render, on the contrary, cleanup run *before* every re-render.

### DIFF
--- a/content/docs/hooks-effect.md
+++ b/content/docs/hooks-effect.md
@@ -333,7 +333,7 @@ function FriendStatusWithCounter(props) {
 
 ### Explanation: Why Effects Run on Each Update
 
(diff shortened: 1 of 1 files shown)


segmentationfaulter:
Sorry for badly phrasing things, this is what I meant to say:
Cleanup happens before a component having a `useEffect` hook re-renders, the documentation mentions the opposite of it.

BartoszKlonowski:
This is a valid addition, thank you, @segmentationfaulter, for contributing! 👍 

Looking at the "Synchronizing with effects" page I can see, however, that this is described in https://react.dev/learn/synchronizing-with-effects#step-3-add-cleanup-if-needed
> React will call your cleanup function each time before the Effect runs again, and one final time when the component unmounts (gets removed).

I will close this PR as no longer required as the current wording describes it well, but I invite you to create new contributions by addressing other open issues.
Thank you!

PR #1612, opened 2019-01-31
Add React Static to "More Flexible Toolchains"

tannerlinsley:
@sw-yx What do you think?


sw-yx:
yup

BartoszKlonowski:
Dears, as the "More flexible toolchains" section is no longer present in the docs, and the toolchains are no longer mentioned anywhere, I will close this PR as no longer required.
Thank you!

PR #1614, opened 2019-02-02
Update non secure URLs to https

koba04:
I've noticed that some URLs are non-secure(http) regardless of these pages are providing secure URLs so I've updated the URLs to https.


lex111:
Hi, we already changed the links in #1644, although your PR was earlier, so I close it, but thanks anyway! :+1:

PR #1618, opened 2019-02-03
Update code-splitting.md

cjke:
Document the use of magic comments to yield a friendly bundle name when code splitting


threepointone:
I don't think we want to go into describing webpack's api here. Thank you for your PR!

PR #1622, opened 2019-02-04
Update examples.md

fatiherikli:
I think it is a convenient project to demonstrate how react and redux work together.


gaearon:
Thanks, but we're keeping this page focused on React. No Redux.

PR #1633, opened 2019-02-06
React Native will ship hooks in its 0.59 release.

cpojer:
I was made aware by @grabbou that the plan is for hooks to ship in React Native's 0.59 release.


gaearon:
Thanks

PR #1635, opened 2019-02-06
react-testing-library now supports `act`

kentcdodds:
It will very shortly: https://github.com/kentcdodds/react-testing-library/pull/280


kentcdodds:
Here's what the update looks like:

image

kentcdodds:
Updated screenshot for the typo fix:

image

This is ready to go! react-testing-library has been updated.

PR #1636, opened 2019-02-06
Support Custom-ID Markdown Syntax

### LINKED ISSUE #1608: Support explicit heading IDs
We currently auto-generate IDs for headings and often link to those generated IDs. As [pointed out] by @smikitky, this causes issues with translation (see https://github.com/reactjs/reactjs.org/issues/1605). If article sections are translated (which they should), the generated IDs will be inconsistent with the English ones, which means translators will need to search for links to those sections and change them correspondingly.

There are two steps needed for this:

First, adding support for [custom heading IDs] in the markdown parser (I checked, we don't support the syntax right now). While this isn't strictly necessary, it would be a great aid for translators and prevent them from having to write ugly html (`<h2 id="examples">Amplesexay</h2>`).

Second, create a script and node task to add explicit IDs to all headings, and a test/lint to make sure that all headings have an explicit heading. The task allows English documentation writers to not worry about creating IDs whenever they write new blog posts/documentation. The test helps ensure that IDs are consistent among the English and international versions of the documentation.

We also may want to consider things like warning when generated IDs are too long (`#what-do-hooks-mean-for-popular-apis-like-redux-connect-and-react-router`), or allowing headings in code blocks (```## `Context.Provider` ```) to not have an explicit heading, since those should not be translated anyway.

[pointed out]: https://github.com/reactjs/reactjs.org/issues/1605#issuecomment-458816106
[custom heading IDs]: https://www.markdownguide.org/extended-syntax/#heading-ids

smikitky:
For the reason why this is necessary, see: #1608

This PR enables user-defined IDs on markdown headings, as explained in [this guide](https://www.markdownguide.org/extended-syntax/#heading-ids).

A heading with an explicit custom ID would look like this:

```
## Try React {#try-react}

### What does const [thing, setThing] = useState() mean? {#destructuring}
```

To implement this, we introduce a new **custom** Gatsby plug-in, which will work in place of the official [gatsby-remark-autolink-headers](https://www.gatsbyjs.org/packages/gatsby-remark-autolink-headers/) plugin (MIT Licensed).

- Removed dependency on `gatsby-remark-autolink-headers` and copied most of its contents under `plugins` directory in this repository. The new custom plugin is called `gatsby-remark-header-custom-ids`.
- Added a few lines of code to modify the AST. (e74a555db72ce3400225fbcbfc7159767730cef8)

To check how this works, edit some `*.md` file under `content`, and execute the following:

1. `yarn` (removes the official plugin)
2. `yarn reset` (this seems to be important after something changed in config)
3. `yarn dev` or `yarn build`

Notes:

- If no explicit custom ID is present, nothing changes; we just keep generating a slugified ID from the header text. So we can gradually adopt custom IDs on a per-page or per-heading basis.
- Custom IDs must contain only *lower-case* alphabet, digits, hyphens and underscores.
- This PR does not include a script to add explicit custom IDs to existing headings.

CC @tesseralis


plugins/gatsby-remark-header-custom-ids/gatsby-ssr.js
```diff
@@ -0,0 +1,76 @@
...
+  );
+
+  return setHeadComponents([
+    style,
+    <script
+      key="gatsby-remark-header-custom-ids-script"
+      dangerouslySetInnerHTML={{__html: script}}
+    />,
+  ]);
+};
```
bvaughn: Looks like a formed+formatted version of [`gatsby-remark-autolink-headers/src/gatsby-ssr.js`](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-remark-autolink-headers/src/gatsby-ssr.js) 👍

plugins/gatsby-remark-header-custom-ids/index.js
```diff
@@ -0,0 +1,69 @@
...
+  visit(markdownAST, 'heading', node => {
+    // Support custom-id syntax.
+    const rawHeader = toString(node);
+    const match = /^.+(\s*\{#([a-z0-9\-_]+?)\}\s*)$/.exec(rawHeader);
+    const id = match ? match[2] : slugs.slug(rawHeader, maintainCase);
+    if (match) {
+      // Remove the custom ID part from the text node.
+      const lastNode = node.children[node.children.length - 1];
+      lastNode.value = lastNode.value.replace(match[1], '');
+    }
```
bvaughn: Looks like a formed+formatted version of [`gatsby-remark-autolink-headers/src/index.js`](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-remark-autolink-headers/src/index.js) with the exception of the above ~10 lines 👍

tesseralis:
@bvaughn we could always merge it in, and switch to the open-sourced plugin later!

KyleAMathews:
Would love this PRed upstream! We do merge and release pretty quickly so it might be easier to just PR your improvements to us.

bvaughn:
😄 I've found it pretty easy to get contributes to Gatsby plug-ins merged quickly!

I'll merge this for now and we can revert once the upstream change has been published

smikitky:
Thank you for merging. I believe this works in our use case, but right now this is not suitable for upstream because:

- Probably, the more "correct" way to handle this is to extend the parser. The current regexp-based approach to patch AST is hacky.
- Only lower-cased IDs are accepted, which is a requirement specific to this website.
- I removed tests because sibling custom plugins didn't have ones, either :)

bvaughn:
I don't think I agree that the first two bullet points are blockers, and some basic tests could be written pretty easily 😄

smikitky:
Yes, I'll make a PR to the official plug-in tonight or this weekend :)

bvaughn:
That's awesome 😄 I think other Gatsby users will really appreciate the ability to customize header ids like this

PR #1157, opened 2018-09-05
HTTPS-ify lots of URLs, update some to their newer homes

mgol:
This matters for user security.

I didn't touch blog posts & pages related to past conferences.

The `content/docs` could be HTTPS-ified as well but it contains lots of HTTP
links so maybe that'd be best done in a separate PR.

PR #1644, opened 2019-02-07
HTTPS-ify most insecure URLs in content/docs (& one in Acknowledgments)

mgol:
This is a followup to #1157.


lex111:
@mgol thank you! Safety in first place! :shield:

tesseralis:
@mgol how were you able to verify that all the https links actually work? If you go to an https link that *isn't* actually secure, Google Chrome will tell you it's insecure and won't let you visit it.

mgol:
@tesseralis I visited them all by hand in the browser. Well, not all URLs but at least one URL per domain (when I saw a repeated domain I didn't always check all paths).

tesseralis:
Okay, thanks! Just wanted to make sure :)

PR #1649, opened 2019-02-08
Update hooks-effect.md

Eckhardt-D:
In the `handleStatusChange` function it receives an object as a parameter, but from the `subscribeToFriendStatus` and `unsubscribeFromFriendStatus` methods, these callback functions do not have any arguments passed to them. For the sake of verbosity I propose adding the arguments.


BartoszKlonowski:
Hello @Eckhardt-D, thank you for taking the time to contribute!
I can see that `useEffect` page in react.dev contains updated info and the parts and samples you modify are no longer present there. Let me close this PR as outdated, but please create new PRs to address open issues if you would still like to contribute.

PR #1662, opened 2019-02-10
Update examples.md

Dhroov7:
I love to see my project added on reactjs.org page.


gaearon:
Thanks, but open source is not a competition :-D I don't think the spirit of this example makes a lot of sense for this page.

PR #1671, opened 2019-02-12
adding --save-dev to eslint-plugin-react-hooks bash command so it will ins…

hananbo:
…talled as devDependency

eslint plugins should be installed as devDependencies instead of dependencies.


lex111:
@hananbo thank you, but already fixed!

PR #1675, opened 2019-02-12
Added context around why use componentDidMount

dericgw:
I wanted to give some context around why we recommend using `componentDidMount` to make API calls and why we do not use the constructor, for instance.


dericgw:
I can't imagine this is needed anymore so I am closing it.

PR #1681, opened 2019-02-12
Fix useCallback example snippet to match useMemo equivalence clause

barkmadley:
Based on the line: https://github.com/reactjs/reactjs.org/blob/486dac1c6786de0b4b670c4d4b14123283974f97/content/docs/hooks-reference.md#L288

I get the impression that the `useCallback` example code should be of the form `useCallback(fn, inputs)`.

### DIFF
--- a/content/docs/hooks-reference.md
+++ b/content/docs/hooks-reference.md
@@ -273,12 +273,7 @@ If you return the same value from a Reducer Hook as the current state, React wil
 ### `useCallback` {#usecallback}
 
 ```js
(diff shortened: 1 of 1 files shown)


denis-sokolov:
This is not correct, the example in the PR does not switch to the useMemo. In addition, useMemo and useCallback do not pass the parameters from the dependency array into the function, those are only for cache tracking purposes.

barkmadley:
Thanks for responding.

Indeed it was my early misunderstanding.

PR #1692, opened 2019-02-15
update gatsby, s/glamor/emotion

threepointone:
emotion's maintained well right now, the perf's great, and the team is pretty nice. the api is almost equivalent to glamor's. I decided to see how hard it it would be to port from glamor to emotion, and because its plugin depended on it, I updated gatsby too. It wasn't too hard! One annoyance is that emotion recommends against `:first-child` and `:nth-child` because of the way  it messes with its ssr output, and might flash when loading. I'll fix that if there's interest in this PR. 🍻


plugins/gatsby-transformer-versions-yaml/gatsby-node.js
```diff
@@ -20,10 +20,12 @@ exports.onPostBuild = async ({store}) => {
 
   // versions.yml structure is [{path: string, url: string, ...}, ...]
   createRedirects(
-    versions.filter(version => version.path && version.url).map(version => ({
```
threepointone: prettier reformatted this

PR #1694, opened 2019-02-16
Fixed Header Navbar Shadow on Mobile

vahnag:
The header navbar shadow seems to be there as a UX hint to indicate mobile users that there's more content and they can scroll to view that. 

Unfortunately it's kind of broken now and looks really bad both on Android and iOS and even just viewing the website in narrow browser widths. The shadows on both sides are always there and don't really work well for the thing they are supposed to help with and look plain ugly and broken.

I've added a little bit of spacing inside the navbar on both sides to make shadow go away and look like it's not there when it reaches the ends of the scrollable container and there's nothing else to scroll to.

Here's the before and after screenshots:

reactjs org_
localhost_8000_

And [here's](https://github.com/reactjs/reactjs.org/files/2872358/Archive.zip) an ZIP archive with two quick screen recordings of the scrolling behavior before and after the change.

This is a pretty minor change and shouldn't affect anything, but actually not sure if I've organized the CSS right and it's consistent throughout the codebase. Let me know if there's anything that needs to be tweaked. Thanks!


threepointone:
Happy to revisit this if you can rebase and confirm it's still alright?

vahnag:
@threepointone rebased and tested again. Thank you!

PR #1696, opened 2019-02-16
updated footer to get year based on current date

saranshkataria:
When I was merging the recent PR for updating year, I realized that it would be better to generate it using javascript instead of changing it manually every year


saranshkataria:
haha, none of the lawyers need to know that we are automating things

PR #1701, opened 2019-02-17
Added react-model as model-management

leecade:


leecade:
ping @tesseralis @saranshkataria

PR #1704, opened 2019-02-19
Updated reference-react-component.md

Mayankgupta688:
When going through the documentation, I got confused when I came across this statement:

"Note that returning false does not prevent child components from re-rendering when their state changes"

Confusion is around the fact that what are the conditions where the component is not allowed to re-rendered. I wanted to propose the updation of following lines to: 

"Note that returning `false` prevents child components from re-rendering when the props values are updated however it does not prevent child components from re-rendering when *their* state changes."

This makes it more explicit that only the "props" change is restricted from re-rendering in case "shouldComponentUpdate" returns false

Thanks & Regards
Mayank Gupta


BartoszKlonowski:
Hello @Mayankgupta688!
I can see that in the latest of react.dev docs this part of the sentence you extend is not present, hence I will close this PR as outdated.
If you still find the information insufficient please create a new PR or address open issues if you would still like to contribute.
Thank you!

PR #1705, opened 2019-02-19
Fix command for initializing tsconfig

vvscode:
Project dependency doesn't add itself it global PATH. So call `tsc` will work either with globally installed typescript, or from npm/yarn script.


lex111:
Good catch, just let's add another command for Yarn, for consistency with the text, i.e.

> If you use [Yarn](https://yarnpkg.com/), run:
```bash
yarn run tsc --init
```

If you use [npm](https://www.npmjs.com/), run:
```bash
npx tsc --init

lex111:
@vvscode thank you!

PR #1707, opened 2019-02-19
Fix typo in blog

DenrizSusam:
"collocation" instead of "colocation".


lex111:
@DenrizSusam thank you!

PR #1714, opened 2019-02-20
Add note to code splitting doc clarifying use of lazy and Suspense

mattmurph9-zz:
When using React.lazy without Suspense, I run into an error:

screenshot 2019-02-20 at 11 00 44

This error can be superseded by other errors that stem from this error when working with a larger codebase with more complex Component selection logic.

It should be specified in the docs that React.lazy needs Suspense to specify a fallback UI.

Also, small format and grammar fix

### DIFF
--- a/content/docs/code-splitting.md
+++ b/content/docs/code-splitting.md
@@ -115,7 +115,7 @@ parse the dynamic import syntax but is not transforming it. For that you will ne
 
 > Note:
 >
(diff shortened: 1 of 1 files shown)


mattmurph9:
**After:**

```js
const OtherComponent = React.lazy(() => import('./OtherComponent'));

function MyComponent() {
  return (
    <div>
      <OtherComponent />
    </div>
  );
}
```

This example doesn't work on its own. We should consider adding a note to say this to avoid confusion, or add a Suspense parent into the example

BartoszKlonowski:
Hello @mattmurph9-zz!
I don't see this addition being anymore required, as both got their separate pages describing them in details.
If you would still like to contribute, then please create new PR to any open issue.
Thank you!

PR #1715, opened 2019-02-20
Add React Conf Armenia 2019 to conferences

varszakaryan:
This is the first React Conf in the region and it will gather more than 500 React developers from Armenia and other regional countries.


gaearon:
❤️

varszakaryan:
@gaearon Thanks, Dan ❤️ It would be just amazing if you can make to join us! 👍

PR #1716, opened 2019-02-20
Change wording?

jdetle:
Hi there, it looks like this is talking about functional components, which are stateless, so saying that you have access to the components state is confusing. Am I getting something wrong here?


StepanNaryshkov:
I think that they meant the state from `useState`

threepointone:
I think it's clear what it means, and the last year has shown that people understand that 'state' inside a function component refers to useState. Thank you for your PR!

PR #1718, opened 2019-02-20
Define hooks in positive terms to avoid confusion about classes

infiniteluke:
In response to confusion over whether classes are "bad" or "being deprecated" we should define hooks by what they are not by what they're not (classes).

This PR changes the verbiage of the intro text that appears in every hooks section.

Spawned from the following thread/conversation:
<a href="https://twitter.com/dan_abramov/status/1098147803480670209">A tweet from dan: I think it’s unfortunate that people see Hooks as a statement against JavaScript classes. I’ll happily continue using classes for cases where I want an object factory with some methods. React components seem a bit different to me, that’s all.</a>

See preview of changes [here](https://github.com/reactjs/reactjs.org/pull/1718#issuecomment-465701116) or see the netlify build.

### DIFF
--- a/content/docs/hooks-custom.md
+++ b/content/docs/hooks-custom.md
@@ -6,7 +6,7 @@ next: hooks-reference.html
 prev: hooks-rules.html
 ---
 
-*Hooks* are a new addition in React 16.8. They are the fundamental building blocks of a React app, allowing you to compose state and other React features into sharable logic.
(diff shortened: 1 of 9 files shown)


infiniteluke:
Prior verbiage:
prior-hook-intro-verbiage

New verbiage:
new-hook-intro-verbiage

infiniteluke:
Concluded conversation with Dan on twitter: https://twitter.com/lukeherrington/status/1098300843198013440

PR #1719, opened 2019-02-20
Fix setState to use function

zardav:
Fix setState to use function when using the previous state.
(Read https://reactjs.org/docs/state-and-lifecycle.html 😃)

### DIFF
--- a/content/docs/hooks-state.md
+++ b/content/docs/hooks-state.md
@@ -174,7 +174,7 @@ In a function, we can use `count` directly:
 In a class, we need to call `this.setState()` to update the `count` state:
 
 ```js{1}
-  <button onClick={() => this.setState({ count: this.state.count + 1 })}>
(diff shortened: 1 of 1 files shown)


content/docs/hooks-state.md
```diff
@@ -174,7 +174,7 @@ In a function, we can use `count` directly:
 In a class, we need to call `this.setState()` to update the `count` state:
 
 ```js{1}
-  <button onClick={() => this.setState({ count: this.state.count + 1 })}>
+  <button onClick={() => this.setState(state => ({ count: state.count + 1 }))}>
```
tesseralis: If you're changing this, you probably want to change the function version below on line 185. But I think we might purposely not use the function versions here for simplicity.

gaearon:
It’s not strictly necessary for many cases and here it gets in the way of explaining other things. Since the code works anyway I’d avoid changing that here. (Although you’re right it’s fenerally preferred).

PR #1722, opened 2019-02-21
correct useMemo pseudocode

monkindey:
useMemo need to return something, just like

```js
const memoizedValue = useMemo(() => computeExpensiveValue(a, b), [a, b]);
```


tesseralis:
In this case, it's telling you that `useCallback` is the same as using `useMemo` on a *function*. The current example as-is is correct.

PR #1723, opened 2019-02-21
Added the Ann Arbor Michigan Meetup to the US list.

philjestin:
Added the Ann Arbor Michigan Meetup to the US list.


lex111:
@philjestin thanks!

PR #1725, opened 2019-02-21
Update accessibility.md

nathanlschneider:
Fixed grammar issue.

### DIFF
--- a/content/docs/accessibility.md
+++ b/content/docs/accessibility.md
@@ -306,7 +306,7 @@ as the `window` object never receives a `click` event. This can lead to obscured
 
 <img src="../images/docs/outerclick-with-keyboard.gif" alt="A toggle button opening a popover list implemented with the click outside pattern and operated with the keyboard showing the popover not being closed on blur and it obscuring other screen elements." />
 
-The same functionality can be achieved by using an appropriate event handlers instead, such as `onBlur` and `onFocus`:
+The same functionality can be achieved by using an appropriate event handler instead, such as `onBlur` and `onFocus`:
 
 ```javascript{19-29,31-34,37-38,40-41}
 class BlurExample extends React.Component {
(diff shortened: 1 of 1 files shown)


content/docs/accessibility.md
```diff
@@ -369,7 +369,7 @@ class BlurExample extends React.Component {
 ```
 
 This code exposes the functionality to both pointer device and keyboard users. Also note the added `aria-*` props to support screen-reader users. For simplicity's sake
-the keyboard events to enable `arrow key` interaction of the popover options have not been implemented.
+the keyboard events to enable `arrow key` interaction of the popover options has not been implemented.
```
AlmeroSteyn: This refers to the keyboard events and should, I think, be `have` for plural.

content/docs/accessibility.md
```diff
@@ -306,7 +306,7 @@ as the `window` object never receives a `click` event. This can lead to obscured
 
 <img src="../images/docs/outerclick-with-keyboard.gif" alt="A toggle button opening a popover list implemented with the click outside pattern and operated with the keyboard showing the popover not being closed on blur and it obscuring other screen elements." />
 
-The same functionality can be achieved by using an appropriate event handlers instead, such as `onBlur` and `onFocus`:
+The same functionality can be achieved by using an appropriate event handler instead, such as `onBlur` and `onFocus`:
```
AlmeroSteyn: Yeah this is wrong. In the example below this, both `onFocus` and `onBlur` have been implemented, so perhaps `using appropriate event handlers` will be better.

tesseralis:
Closing because as @AlmeroSteyn said, the current grammar is correct.

AlmeroSteyn:
@tesseralis The one sentence is indeed a little broken:

> The same functionality can be achieved by using an appropriate event handlers instead, such as `onBlur` and `onFocus`:

should become 

The same functionality can be achieved by using an appropriate event handler instead, such as `onBlur` and `onFocus`:

So let's re-open :-)

tesseralis:
@AlmeroSteyn oh, you're right, I didn't see that.

My preference would be

> The same functionality can be achieved by using the appropriate event handlers instead, such as `onBlur` and `onFocus`:

AlmeroSteyn:
@tesseralis Yeah you are quite right! Also along the lines of what I suggested initially. Don't comment from your phone, haha!

@nathanlschneider Will you be able to update?

PR #1726, opened 2019-02-21

nathanlschneider:
Fixed grammar issue.

### DIFF
--- a/content/docs/typechecking-with-proptypes.md
+++ b/content/docs/typechecking-with-proptypes.md
@@ -34,7 +34,7 @@ Greeting.propTypes = {
 
 ### PropTypes {#proptypes}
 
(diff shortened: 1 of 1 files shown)


content/docs/typechecking-with-proptypes.md
```diff
@@ -34,7 +34,7 @@ Greeting.propTypes = {
 
 ### PropTypes {#proptypes}
 
-Here is an example documenting the different validators provided:
+Here is an example of documenting the different validators provided:
```
tesseralis: If this is confusing, I prefer

> Here is an example that documents the different validators provider.

This current version is kind of awkward.

threepointone:
Happy to revisit this if @tesseralis' recommendation is included.

PR #1729, opened 2019-02-22

gcor:
After adding hooks, it may be better to remove the part that classes have more features than functions.


tesseralis:
Classes still have some functionality that hooks don't: https://reactjs.org/docs/hooks-faq.html#do-hooks-cover-all-use-cases-for-classes

This will definitely be true at some point, just not now :)

PR #1736, opened 2019-02-23
Fixed component name on render-props page

ran81:
The docs say you can use a HOC that uses the `<Mouse>` "instead of a `<Mouse>` component". Seems like it should say "instead of a `<MouseTracker>` component"

### DIFF
--- a/content/docs/render-props.md
+++ b/content/docs/render-props.md
@@ -215,7 +215,7 @@ More concretely, **a render prop is a function prop that a component uses to kno
 
 This technique makes the behavior that we need to share extremely portable. To get that behavior, render a `<Mouse>` with a `render` prop that tells it what to render with the current (x, y) of the cursor.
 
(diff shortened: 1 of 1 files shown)


ran81:
@tesseralis 

Thanks! I see what you are saying. Though the wording is a bit off in my opinion. It says:
"For example, if you would prefer to have a ```withMouse``` HOC instead of a ```<Mouse>``` component, you could easily create one using a regular ```<Mouse>``` with a render prop:"

Saying "instead of a ```<Mouse>```" we use a "regular ```<Mouse>```" suggests that they are different.

Note to self - don't read docs late at night

BartoszKlonowski:
Dears! Let me close this PR as per discussion, requested changes, and the fact that the PR is already outdated, as the changed part of the docs is already no longer present in the latest version of it.
@ran81 Thank you for taking the time to create the PR though, please feel free to create new PRs for other open issues.

PR #1741, opened 2019-02-23
Put Blog in top bar before Community

gaearon:
This way it isn't completely obscured on iPhone 7 due to language picker. (Instead, "Community" doesn't quite fit but it's more obviously scrollable.)

It's also official so seems fitting to emphasize it.


bvaughn:
Multi-line header wouldn't be necessary. With a few style tweaks, we could avoid the need to scroll on mobile-sized screens:
screen shot 2019-02-23 at 2 17 44 pm

PR #1747, opened 2019-02-24
Add missing `in` at state-and-lifecycle.md

ariebrainware:


ariebrainware:
I just re-created my PR, because my CLA before is not signed. Now it's done.

threepointone:
This feels like a nit, it's clear either way. But thank you for the PR!

PR #1750, opened 2019-02-24
Update examples.md

klyburke:
We created a boilerplate app, using React, to build any SaaS product. We've used this boilerplate for [builderbook.org](https://builderbook.org) and [async-await.com](https://async-await.com).

Thanks in advance for considering it!


gaearon:
Thanks! We're trying to keep this page focused on vanilla React examples, so MobX doesn't quite fit in there.

klyburke:
Thanks! Completely understand.

PR #1751, opened 2019-02-25
Avoid horizontal scrolling in header for mobile devices

bvaughn:
Wouldn't it be nice if the header didn't require horizontal scrolling on smaller devices? Here's one possible take.

cc @joecritch 

# Screenshots at various sizes
screen shot 2019-02-24 at 6 13 28 pm
screen shot 2019-02-24 at 6 13 40 pm
screen shot 2019-02-24 at 6 14 00 pm
screen shot 2019-02-24 at 6 14 14 pm
screen shot 2019-02-24 at 6 14 24 pm
screen shot 2019-02-24 at 6 14 30 pm

# Possible pro (or con)?
header

### DIFF
--- a/src/components/Container/Container.js
+++ b/src/components/Container/Container.js
@@ -30,6 +30,11 @@ const Container = ({children}: {children: Node}) => (
       [media.size('xxlarge')]: {
         maxWidth: 1260,
       },
+
+      [media.lessThan('small')]: {
+        paddingLeft: 8,
+        paddingRight: 8,
+      },
     }}>
     {children}
   </div>
--- a/src/components/LayoutHeader/DocSearch.js
+++ b/src/components/LayoutHeader/DocSearch.js
@@ -45,14 +45,11 @@ class DocSearch extends Component<{}, State> {
           paddingLeft: '0.25rem',
           paddingRight: '0.25rem',
 
-          [media.lessThan('large')]: {
+          [media.lessThan('xlarge')]: {
             justifyContent: 'flex-end',
             marginRight: 10,
           },
-          [media.between('medium', 'xlarge')]: {
-            //width: 'calc(100% / 6)',
-          },
-          [media.greaterThan('large')]: {
+          [media.greaterThan('xlarge')]: {
             minWidth: 100,
           },
         }}>
@@ -80,7 +77,7 @@ class DocSearch extends Component<{}, State> {
               borderRadius: '0.25rem',
             },
 
-            [media.lessThan('large')]: {
+            [media.lessThan('xlarge')]: {
               fontSize: 16,
               width: '16px',
               transition: 'width 0.2s ease, padding 0.2s ease',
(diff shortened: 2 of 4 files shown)


bvaughn:
Hm. Looks like Dan merged something that touches this codepath already and we clashed.

I think this approach is still worth consideration 😄I'll hold off on dealing with merge conflicts until/unless other people also think it's worthwhile.

atomiks:
The layout reflow is kind of unattractive imo. I think the current approach is still the nicest alternative to a hamburger menu, but with a modification to the gradient so when scrolled right to the left/right edge, it doesn't appear (maybe extra padding?)

gaearon:
Sorry for the conflict, I just wanted to plug the visible scrollbar problem on Windows asap.

I like that in this PR iPhone has a smaller font so everything fits by default. I'd like to see that change.

However, I prefer the current search/top bar behavior that doesn't reflow elements.

bvaughn:
> However, I prefer the current search/top bar behavior that doesn't reflow elements.

I don't think it's common to resize browser windows and actually see this reflow happening. Most users browse at a fixed size, so we're just trying to optimize every fixed size to avoid unnecessary overflow/scrolling.

sophiebits:
* +1 for avoiding splitting the nav on two lines when searching.
* The narrow window that now fits them all on one line looks nice, but I'm worried that it is too specific to your particular browser size and to English; I don't think we can count on things fitting perfectly, so we should make sure we have a design that works well with longer text too.
* The space between lines looks too big on the multi-line screenshot.
* What if instead of 

  > ⚛️ Docs Tutorial Blog Community 🔍 🌍

  we do

  > ⚛️ React ………………… 🔍 🌍
  > Docs Tutorial Blog Community

  on two lines (but with the nav links themselves (often) only on one). Does that fit better?

bvaughn:
> but I'm worried that it is too specific to your particular browser size

No. I tested a fluid range of size between the smallest size Firefox would go (it goes smaller than Chrome/Safari) to a full screen.

> * ⚛️ React ………………… 🔍 🌍
  Docs Tutorial Blog Community

I like this suggestion.

lex111:
Please do something already, otherwise now the search overlaps last menu item, so generally looks awful...

image

atomiks:
@lex111 That seems to be what @sophiebits was talking about regarding adjusting the layout to based around English wording. It's currently is happening on the website, because "Docs" (English version) is much shorter, and the breakpoint assumes the English version's size.

RE two line layout: would the increased height be a problem on mobile since it's a fixed nav? It would reduce the amount of vertical space for viewing content. Maybe when scrolling up it could expand open, when scrolling down to read content, it hides to be one line.

musewick:
Concept where long words fit well:
react_js

threepointone:
Would you like to revisit this?

harish-sethuraman (commented):
Seems like this issue was addressed and is not occurring now. So I'll close the PR. Thanks :)

PR #1754, opened 2019-02-25
Define behavior of changes to `initialState` value

rubencodes:
This PR adds information about what happens if the `initialState` argument to `useState` changes value on subsequent re-renders (which is: nothing, it's ignored).

### DIFF
--- a/content/docs/hooks-reference.md
+++ b/content/docs/hooks-reference.md
@@ -43,7 +43,7 @@ The `setState` function is used to update the state. It accepts a new state valu
 setState(newState);
 ```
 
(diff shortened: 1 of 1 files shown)


BartoszKlonowski:
Hello @rubencodes,
this is a valid note, thank you for the contribution!
Looking at the latest docs describing the `useState` hook I see that the fact of ignoring the further changes to initial state is already mentioned in [Avoiding recreating the initial state](https://react.dev/reference/react/useState#avoiding-recreating-the-initial-state) section.
Let me close this PR as outdated, however please feel encouraged to create new PRs for open issues.

PR #1756, opened 2019-02-26
fix docs word error

cqq626:
In "This is where higher-order components excel" of the higher-order-components.html, "excel" should be "exist"


saranshkataria:
I don't think it should be exist, excel is perfect since that is the use case for HOCs.

cqq626:
> I don't think it should be exist, excel is perfect since that is the use case for HOCs.

You are right, I misunderstand the meaning. Sorry for my bothering.

PR #1758, opened 2019-02-27
Update ESLint Plugin installation command

imagentleman:
This PR changes the installation command to match the instructions in the plugin page on npm (https://www.npmjs.com/package/eslint-plugin-react-hooks), which installs the plugin as a dev dependency (which is probably the expected use case for most people).


lex111:
@imagentleman thank you!

PR #1760, opened 2019-02-27
Stop calling function components stateless

aem:
These docs still refer to function components as "stateless function components" despite the fact that function components aren't necessarily stateless (anymore). The docs shouldn't use this outdated terminology anymore. 

I left the "stateless" in the link to avoid breaking existing links to these docs but can remove if keeping links valid isn't a priority.


gaearon:
Sure.

PR #1761, opened 2019-02-28
Tweaked explanation of unbound function context

dxparker:


threepointone:
I don't think this is correct, the original text seems fine? Thanks for the effort!

PR #1762, opened 2019-02-28
Add ReactJS Girls Conference to Conference List

M0nica:
- add upcoming ReactJS Girls Conf 
- move past conferences to the right place


alexkrolick:
👍thanks

PR #1763, opened 2019-02-28
Better favicon in all browsers/OS

MoOx:
The goal of this commit is to offer a better favicon (& friends) to this site.
This will looks better in all browsers/OS when added to favorites etc.

For example, currently it looks like this in Safari

image

With this change, it will looks better

image

I used http://realfavicongenerator.net to generate this versions from a proper svg using color codes from the site.


threepointone:
Happy to revisit this if you'd rebase and make sure it all still works?

MoOx:
Currently it seems to be good at least for what initiated this PR in my brain

image

I guess I should do this PR for RN now haha

MoOx:
This PR https://github.com/reactjs/reactjs.org/pull/2377 seems to have made the situation better.

PR #1765, opened 2019-02-28
Fix CRA code splitting link

RaedsLab:
This commit should help redirect people to the proper "code splitting" create app page instead of the old github link.


sophiebits:
Thanks!

Issue #1639, opened 2019-02-06
Document TestRenderer.act()


gaearon:
- Like https://reactjs.org/docs/test-utils.html#act but on test renderer page
- Link them with each other (Note: if you use [other thing], it also has act)

threepointone:
Fixed earlier

PR #1766, opened 2019-02-28
Add act to react-test-renderer reference docs

Esemesek:
Added missing `act` method to `react-test-renderer` reference docs with simple `Counter` example similar to `react-dom/test-utils`.

Fixes: #1639


Esemesek:
ping @gaearon

threepointone:
This was added in a separate diff. Thank you for your PR!

PR #1768, opened 2019-03-01
Change const to var in test renderer

itayyehezkel:
Change `const` to `var` in Test Renderer like mentioned in guidelines for code examples. 

> **Use `const` where possible, otherwise `let`. Don't use `var`.**
>
> Ignore this if you're specifically writing about ES5.


alexkrolick:
"ES5" here is misleading, it is really talking about CommonJS `require` syntax

threepointone:
Yeah I'm going to push back on this, the example is fine as is. Thank you for the PR!

PR #1769, opened 2019-03-01
Add default value to useReducer example usage

jkjustjoshing:


jkjustjoshing:
I think this PR is still relevant, but it seems like it won't get merged. So I'm just going to close it.

PR #1771, opened 2019-03-02
Update languages page

tesseralis:
* Japanese is at 100%! おめでとう御座います！
screen shot 2019-03-01 at 4 29 45 pm

* Switch to just two categories: "complete" (100%) and "in progress" (<100%)
* Added new languages

Preview:
screen shot 2019-03-01 at 4 46 18 pm

The markdown styling feels a bit weird to me. Anyone have any ideas?

### DIFF
--- a/content/languages.yml
+++ b/content/languages.yml
@@ -1,12 +1,11 @@
 # Status enums indicate what percentage of "core" content has been translated:
-# 0: Incomplete (0–49%)
-# 1: Partially complete (50–94%)
-# 2: Complete (95–100%)
+# 0: In progress (<100%)
+# 1: Complete (100%)
 
 - name: English
   translated_name: English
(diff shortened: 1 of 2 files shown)


gaearon:
What markdown styling?

Also, why do we change sections?

bvaughn:
None of the partially translated docs for the in-progress languages are viewable after this change. Even ones like Portuguese (Brazil) that are well on their way.

tesseralis:
@bvaughn ohh right that's why that section is there. Okay, reset everything to the old system for now. I'm still kind of iffy on the naming but that's a discussion we can have later. For now, let's make sure our page is updated :)

bvaughn:
Pretty exciting to see zh-hans bumped up to 50%!

PR #1773, opened 2019-03-02
Set Japanese as Completed

gaearon:
@tesseralis Is this a good time or did I misunderstand?


tesseralis:
It's a good time but I already have a PR here (along with adding more languages): https://github.com/reactjs/reactjs.org/pull/1771

PR #1780, opened 2019-03-04
fix(unicode-to-char): Replaced other languages unicode to text

### LINKED ISSUE #1732: Unicode-escaped national characters are shown instead of human readable text
The issue is migrated from the ru.reactjs.org: https://github.com/reactjs/ru.reactjs.org/issues/180

@gcor noticed a bug in the way national characters are displayed in Code Editor **with JSX checkbox disabled (reset)**. **Expected** to see human-readable text in a particular natural language (when used in string literals and comments). **Actually** seeing unicode-escaped symbols. (See images below). 

At least two languages are known to be affected: Russian and Armenian, but the issue must be not specific to just these two.

**Russian translation**

screen shot 2019-02-22 at 12 35 12 pm
screen shot 2019-02-22 at 12 35 19 pm

**Armenian translation**

image

nutboltu:
Fixes #1732 

Solution:
After compiled ES6 convert all unicode to text.

### DIFF
--- a/src/components/CodeEditor/CodeEditor.js
+++ b/src/components/CodeEditor/CodeEditor.js
@@ -11,6 +11,11 @@ import {LiveEditor, LiveProvider} from 'react-live';
 import {colors, media} from 'theme';
 import MetaTitle from 'templates/components/MetaTitle';
 
+// Replace unicode to text for other languages
+const unicodeToText = (text) => text
+  .replace(/\\u[\dA-F]{4}/gi, 
+    match => String.fromCharCode(parseInt(match.replace(/\\u/g, ''), 16)));
+
 const compileES5 = (
   code, // eslint-disable-next-line no-undef
 ) => Babel.transform(code, {presets: ['es2015', 'react']}).code;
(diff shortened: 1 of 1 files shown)


tesseralis:
It looks alright, but my script injection skills are lacking and I'd like someone else with more experience to look over it first.

tesseralis:
Also, thank you so much for doing this!

src/components/CodeEditor/CodeEditor.js
```diff
@@ -11,6 +11,12 @@ import {LiveEditor, LiveProvider} from 'react-live';
 import {colors, media} from 'theme';
 import MetaTitle from 'templates/components/MetaTitle';
 
+// Replace unicode to text for other languages
+const unicodeToText = text =>
+  text.replace(/\\u[\dA-F]{4}/gi, match =>
```
tesseralis: You can simplify this with:
```
text.replace(/\\u([\dA-F]{4})/gi, (match, p1) => String.fromCharCode(parseInt(p1, 16))
```

nutboltu: That's more efficient!. I updated the function.

PR #1781, opened 2019-03-04
Remove implicit boolean conversion from homepage example

xixixao:
We shouldn't be encouraging this pattern which requires good understanding of JS quirks and can lead to bugs in other instances.


threepointone:
Thank you!

PR #1783, opened 2019-03-05
removed line as functions can have state now too

### LINKED ISSUE #1767: [Documentation] Glossary: Only calss components can have state
Hey,
I was reviewing a [PR](https://github.com/reactjs/ru.reactjs.org/pull/200) with Russian translation of the article `Glossary of React Terms` and noticed that there is a sentence about class components and state.  

> Only components defined as classes can have state.

You can find it [here](https://reactjs.org/docs/glossary.html#state) in the second paragraph.

React provides a hook `useState` to declare a state for functional components. So I would like to ask, should it be changed after hooks was shipped?

saranshkataria:
Fixes #1767


saranshkataria:
@alexkrolick done

PR #1789, opened 2019-03-05
fixing typos. should be below instead of above

ImranMH:

### DIFF
--- a/content/docs/lists-and-keys.md
+++ b/content/docs/lists-and-keys.md
@@ -263,7 +263,7 @@ With the example above, the `Post` component can read `props.id`, but not `props
 
 ### Embedding map() in JSX {#embedding-map-in-jsx}
 
-In the examples above we declared a separate `listItems` variable and included it in JSX:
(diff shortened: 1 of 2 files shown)


content/docs/lists-and-keys.md
```diff
@@ -263,7 +263,7 @@ With the example above, the `Post` component can read `props.id`, but not `props
 
 ### Embedding map() in JSX {#embedding-map-in-jsx}
 
-In the examples above we declared a separate `listItems` variable and included it in JSX:
+In the examples below we declared a separate `listItems` variable and included it in JSX:
```
tesseralis: I think it's supposed to be "above" because it's referencing previous examples where we declared `listItems` (which you can see if you searched for "listItems").

BartoszKlonowski:
Hello @ImranMH, thank you for the contribution!
Unfortunately the code that you aim to change in this PR is no longer present in the documentation of React. Also the part listing `ReactJS Girls Conference` is already covered, so let me close this as outdated.
However please feel free to create new PRs for other open issues.

PR #1790, opened 2019-03-05
Update tools-misc.md

jondavidjohn:
Add [packtracker.io](https://packtracker.io) to the tools section.


gaearon:
We've removed this page.

PR #1793, opened 2019-03-06
Update lifting-state-up - minor grammatical correction. 

nkamath:

### DIFF
--- a/content/docs/lifting-state-up.md
+++ b/content/docs/lifting-state-up.md
@@ -201,7 +201,7 @@ Now, when the `TemperatureInput` wants to update its temperature, it calls `this
 
 >Note:
 >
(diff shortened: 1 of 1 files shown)


content/docs/lifting-state-up.md
```diff
@@ -201,7 +201,7 @@ Now, when the `TemperatureInput` wants to update its temperature, it calls `this
 
 >Note:
 >
->There is no special meaning to either `temperature` or `onTemperatureChange` prop names in custom components. We could have called them anything else, like name them `value` and `onChange` which is a common convention.
+>There is no special meaning to either `temperature` or `onTemperatureChange` prop names in custom components. We could have called them anything else. For instance, `value` and `onChange` is a common convention.
```
tesseralis: I would do:

> For instance, a common convention is naming them `value` and `onChange`.

Or if you're feeling fancy:

> We could have called them anything else -- for instance, naming them...

I also like "for example" over "for instance" but 🤷‍♀️

PR #1794, opened 2019-03-06
Update examples.md

TheCodersDream:


gaearon:
Thanks, but we're trying to keep this page focused on vanilla React. No Redux.

PR #1806, opened 2019-03-11
remove a reference to fbjs from codebase overview

koba04:
We can remove this sentence because React no longer has `fbjs` as its dependencies.


threepointone:
Thank you!

PR #1807, opened 2019-03-11
[WIP] Hooks Recipes

gaearon:
This doesn't actually have any content yet.

### DIFF
--- a/content/docs/hooks-recipes.md
+++ b/content/docs/hooks-recipes.md
@@ -1,0 +1,12 @@
+---
+id: hooks-recipes
+title: Hooks Recipes
+permalink: docs/hooks-recipes.html
+prev: hooks-faq.html
+---
+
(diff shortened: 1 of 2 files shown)


acdlite:
Why "Hooks Recipes" instead of "Recipes"? It would be nice to have e.g. recipes for how to do stuff with Suspense at some point, too.

gaearon:
For now I'm just keeping it colocated with Hooks docs so it seemed related. We don't use Hooks outside of Hooks docs yet. I think if we start using them, we'll want to switch the whole thing over (or else offer two versions of each article). I'd want to later break recipes out in separate pages (once per recipe) in their own category.

gaearon:
In other words — by the time we ship Suspense for data fetching this page would be broken down into several anyway.

threepointone:
Want to abandon this?

PR #1813, opened 2019-03-12
Extend useCallback inputs argument documentation

KidkArolis:


KidkArolis:
Not sure how to best word this. Suggestions welcome. But I think even this bit of information would be helpful.

PR #1814, opened 2019-03-12
Add FAQ that shows the Hook Flow diagram

donavon:
Display the hook-flow diagram and provide links to the latest PDF and PNG files

image


gaearon:
Can we remove the header/attribution in the inline version? While I appreciate your effort I think it's a bit out of place in the official docs. Alternatively, we could just leave a repo link.

donavon:
Are you referring to the header in the image or the markdown text?

Why don't I remove the header and the footer in the image and make the footer as actual text in markdown.

donavon:
@gaearon 

Is this what you're looking for? It's a lot cleaner now, IMO. I also fixed the merge conflict.

image

content/docs/hooks-faq.md
```diff
@@ -208,6 +209,17 @@ There are a few more heuristics, and they might change over time as we fine-tune
 
 * `componentDidCatch` and `getDerivedStateFromError`: There are no Hook equivalents for these methods yet, but they will be added soon.
 
+### What is the flow of a function component that uses Hooks? {#hook-flow-diagram}
+
+This diagram visually explains, at a high level, the execution flow of a function component that uses Hooks.
```
ahtee: Can probably remove “, at a high level, “ and just say: “ the diagram below is a basic visual example showing the execution flow of a component using hooks.”

donavon: @gaearon? thoughts? and is there anything else you need from me?

rickhanlonii:
Thanks for your work on this, sorry we never got it merged. 

I'm closing since this is for the old docs and the new docs have some more information [about this](https://react.dev/learn/render-and-commit). If we wanted to add this, the new docs have a diagram design system we'd use.

PR #1815, opened 2019-03-12
Add more info to FAQ about deps

gaearon:
This addresses the common confusion points around the deps array and optimizing effects, and suggests concrete fixes. It might be too wordy but I figured it's probably worth including.

Open to edits.

### DIFF
--- a/content/docs/hooks-custom.md
+++ b/content/docs/hooks-custom.md
@@ -18,11 +18,11 @@ import React, { useState, useEffect } from 'react';
 function FriendStatus(props) {
   const [isOnline, setIsOnline] = useState(null);
 
-  function handleStatusChange(status) {
-    setIsOnline(status.isOnline);
-  }
-
   useEffect(() => {
+    function handleStatusChange(status) {
+      setIsOnline(status.isOnline);
+    }
+
     ChatAPI.subscribeToFriendStatus(props.friend.id, handleStatusChange);
     return () => {
       ChatAPI.unsubscribeFromFriendStatus(props.friend.id, handleStatusChange);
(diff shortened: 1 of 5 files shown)


content/docs/hooks-custom.md
```diff
@@ -18,11 +18,11 @@ import React, { useState, useEffect } from 'react';
 function FriendStatus(props) {
   const [isOnline, setIsOnline] = useState(null);
 
-  function handleStatusChange(status) {
-    setIsOnline(status.isOnline);
-  }
-
   useEffect(() => {
+    function handleStatusChange(status) {
```
gaearon: I moved these since it's harder to tell when it's safe, and we should probably just teach people to put them inside effects as a default.

content/docs/hooks-faq.md
```diff
@@ -416,6 +457,193 @@ If you're not familiar with this syntax, check out the [explanation](/docs/hooks
...
+      setCount(c => c + 1); // ✅ This doesn't depend on `count` variable outside
+    }, 1000);
+    return () => clearInterval(id);
+  }, []); // ✅ Our effect doesn't use any variables in the component scope
+
+  return <h1>{count}</h1>;
+}
+```
+
+(The identify of the `setCount` function is guaranteed to be stable so it's safe to omit.)
```
rickbeerendonk:
```suggestion
(The identity of the `setCount` function is guaranteed to be stable so it's safe to omit.)
```

content/docs/hooks-custom.md
rickbeerendonk: I agree. Functions are generally only called within the effect, so prefer to teach this as the default. As a result a value and not a function will be the effect's dependency, which is slightly easier to understand for beginners.

content/docs/hooks-faq.md
```diff
@@ -362,6 +365,44 @@ It's possible that in the future React will provide a `usePrevious` Hook out of
 
 See also [the recommended pattern for derived state](#how-do-i-implement-getderivedstatefromprops).
 
+### Why am I seeing stale props or state inside my function? {#why-am-i-seeing-stale-props-or-state-inside-my-function}
+
+Any function inside a component, including event handlers and effects, “sees” the props and state from the render it was defined in. For example, consider code like this:
```
rickbeerendonk: A few suggestions.
* "sees" is clear enough for me, but due to the quotes maybe a little imprecise. 
* Functions are defined once, but are potentially created multiple times (during renders).

```suggestion
Any function inside a component, including event handlers and effects, contain ("sees") the props and state from the render it was created in. For example, consider code like this:
```

gaearon:
I'll get this in for now. I'd love if someone can figure out a way to make it shorter but at least it's in the docs now.

PR #1824, opened 2019-03-13
Fixing incorrect value for eslint rule

manuelbieh:
The name for the "warning" level in ESLint is `warn` not `warning` (https://eslint.org/docs/user-guide/configuring#configuring-rules)


alexkrolick:
Good catch

PR #1830, opened 2019-03-15
Fix overflowing headers on small screens

nelsonreitz:
Long words in headers will overflow and break the layout on mobile screens (e.g. [https://reactjs.org/acknowledgements.html](https://reactjs.org/acknowledgements.html)). This fix allows them to break on small screens. 

Before:
reactjs org_(iPhone 5_SE)

After:
0 0 0 0_8001_(iPhone 5_SE)

Tested on mobile Safari, Chrome & Firefox

### DIFF
--- a/src/theme.js
+++ b/src/theme.js
@@ -89,6 +89,11 @@ const fonts = {
     lineHeight: '65px',
     fontWeight: 700,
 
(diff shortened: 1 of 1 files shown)


gaearon:
Thanks

PR #1843, opened 2019-03-19
Add a DOM measurement recipe to Hooks FAQ

gaearon:
This keeps coming up. The first impulse is to useRef + useLayoutEffect but they're actually not what you want. So I documented it.

### DIFF
--- a/content/docs/hooks-faq.md
+++ b/content/docs/hooks-faq.md
@@ -475,7 +475,7 @@ function MeasureExample() {
 }
 ```
 
-We didn't choose `useRef` in this example because an object ref doesn't notify us about *changes* to the current ref value. Using a *callback ref* ensures that [even if a child component displays the measured node later](https://codesandbox.io/s/818zzk8m78) (e.g. in response to a click), we still get notified about it in the parent component and can update the measurements.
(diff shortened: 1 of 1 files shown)


content/docs/hooks-faq.md
```diff
@@ -451,6 +452,60 @@ Try to avoid this pattern if possible.
...
+  return (
+    <>
+      <h1 ref={measuredRef}>Hello, world</h1>
+      <h2>The above header is {Math.round(height)}px tall</h2>
+    </>
+  );
+}
+```
+
+We didn't choose `useRef` in this example because an object ref doesn't notify us about *changes* to the current ref value. Using a *callback ref* ensures that [even if a child component displays the measured node later](https://codesandbox.io/s/818zzk8m78) (e.g. in response to a click), we still get notified about it in the parent component and can update the measurements.
```
threepointone: - the italics don't seem necessary of relevant, since you just introduced it. 
```suggestion
We didn't choose `useRef` in this example because an object ref doesn't notify us about *changes* to the current ref value. Using a callback ref ensures that [even if a child component displays the measured node later](https://codesandbox.io/s/818zzk8m78) (e.g. in response to a click), we still get notified about it in the parent component and can update the measurements.
```

PR #1846, opened 2019-03-19
Issue solution added

Rossolson:
As this is still the current Google Search result for Keyed Fragments, a link to the reason why the short syntax can't be used should be added.


BartoszKlonowski:
Hello @Rossolson, thank you for the contribution!
Unfortunately, the part of the docs you are updating in this PR is no longer present in the latest version of docs, so let me close this PR as outdated.

PR #1848, opened 2019-03-19
Document useContext bailout strategy

gaearon:
This keeps coming up, might as well document it.


gaearon:
Yeah, meh.

gaearon:
(We can do a pass to clean them up all together if we care someday)

PR #1849, opened 2019-03-20
Set a context for the external example.

sunnykgupta:
Previously, the text mentioned - here is a problematic codepen link. Here is a codepen link that solves the problem.

As a developer, I had to figure out the real problem with the first codepen link. Which was to actually add some text to the input fields and then try to sort the output.

Added a short description along with the link to the example.

This is an important topic to understand and the examples need to be understood well be developers to write bug-free code with `keys`.


BartoszKlonowski:
Hello @sunnykgupta!
As this sentence is no longer in the docs, I will close this PR as outdated.
Thank you!

PR #1850, opened 2019-03-20
swapped condition position in handleClick function

pnutmath:
swapped conditions within  if condition that returns function early to ignore a click if someone has won the game or if a Square is already filled

### DIFF
--- a/content/tutorial/tutorial.md
+++ b/content/tutorial/tutorial.md
@@ -739,7 +739,7 @@ We can now change the Board's `handleClick` function to return early by ignoring
 ```javascript{3-5}
   handleClick(i) {
     const squares = this.state.squares.slice();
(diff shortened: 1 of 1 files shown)


content/tutorial/tutorial.md
```diff
@@ -739,7 +739,7 @@ We can now change the Board's `handleClick` function to return early by ignoring
 ```javascript{3-5}
   handleClick(i) {
-    if (calculateWinner(squares) || squares[i]) {
```
pnutdev: This change is to avoid call to calculateWinner function if game is not finished

rickhanlonii:
Closing since this page has moved, but you could submit a new PR against the new docs. Seems like a reasonable optimization.

PR #1860, opened 2019-03-25
Restyle nav bar positioning to support translations

### LINKED ISSUE #1817: Restyle logo width to make translations easier
Currently, the width of the logo is fixed at `calc(100% / 6)` so that the nav links are flush with the search bar:

Screen Shot 2019-03-12 at 7 33 34 PM

Unfortunately, this leads to cut-offs in some translations with longer words in the nav bar:

Screen Shot 2019-03-12 at 7 38 26 PM

We're currently solving this by just adjusting the width of the logo, but it would be nice to have a solution that doesn't require hard-coding the logo width (e.g. just making the nav bar flush against the search bar)

I'll probably fix it myself when I have time but I figured I'd put this up here in case anyone wants to take a look.

kevjin:
### **About:**
This PR fixes #1817. 

Currently for Ukrainian translation, a hard-coded logo width is used to prevent nav bar cut offs w/ the search bar. 
The change in this PR centers the nav bar between the logo and search bar.

### **Visuals:**
**Before Changes:**
[uk.reactjs.org](https://uk.reactjs.org/):
Screen Shot 2019-03-24 at 10 03 58 PM

**After Changes:**
Screen Shot 2019-03-24 at 11 03 40 PM

(Decided to go for centering of nav bar so the style would fit with the default English translation)
[reactjs.org](https://reactjs.org/):
Screen Shot 2019-03-24 at 10 04 37 PM

### **Other:**
Flexbox auto margins were used over `justify-content: center` so [scrolling of nav bar wouldn't be cut off in mobile display](https://stackoverflow.com/questions/33454533/cant-scroll-to-top-of-flex-item-that-is-overflowing-container).

### DIFF
--- a/src/components/LayoutHeader/Header.js
+++ b/src/components/LayoutHeader/Header.js
@@ -54,9 +54,6 @@ const Header = ({location}: {location: Location}) => (
               color: colors.white,
             },
 
-            [media.greaterThan('small')]: {
-              width: 'calc(100% / 6)',
-            },
             [media.lessThan('small')]: {
               flex: '0 0 auto',
             },
@@ -94,7 +91,6 @@ const Header = ({location}: {location: Location}) => (
           css={{
             flex: '1',
             display: 'flex',
-            flexDirection: 'row',
             alignItems: 'stretch',
             overflowX: 'auto',
             overflowY: 'hidden',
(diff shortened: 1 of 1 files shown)


kevjin:
:(

PR #1865, opened 2019-03-25
remove pleonasm: 0 side effects

blistok:
Required property of the pure function is to have no side effects, so emphasizing it feels superfluous.


threepointone:
Sure... but this feels like a nit, and I'm inclined to reject. The extra words provide a good emphasis. Thank you for your PR!

PR #1871, opened 2019-03-26
Update tutorial.md

ch-ricardor:
This change is irrelevant for the tutorial but could be useful to improve performance in complicated conditions and calculations.

I confirmed a behavior, in an IF-THEN-ELSE with OR conditions, the "if" will test the state of the sentences to get a TRUE and omit the other clauses. _(Considering this could improve performance and reduce processor cycles in more complex calculations)_

Could be added a recommendation: _- If the result of a function will be tested more than once, run it once and store the result in a variable, considering the scope of that variable._

In this case:
```
if ( squares[i] || calculateWinner(squares)) {
```

Test square first, if it has a move, won't test the "calculateWinner".

In this case the "for" has few cycles, 

image

If you test:

	if ( calculateWinner(squares) || squares[i]) {

With this additional cycle in calculateWinner function will notice the difference if the user clicks on a played square.

```
  for (let i = 0; i < 10000000000; i++) {
  }
```

image

or adding an:

`alert('testing winner');`

Nice explanation in how to build and refactor classes.

### DIFF
--- a/content/tutorial/tutorial.md
+++ b/content/tutorial/tutorial.md
@@ -739,7 +739,7 @@ We can now change the Board's `handleClick` function to return early by ignoring
 ```javascript{3-5}
   handleClick(i) {
     const squares = this.state.squares.slice();
-    if (calculateWinner(squares) || squares[i]) {
+    if (squares[i] || calculateWinner(squares)) {
       return;
     }
     squares[i] = this.state.xIsNext ? 'X' : 'O';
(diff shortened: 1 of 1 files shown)


BartoszKlonowski:
In this case the readability stays the same (and can even be improved by putting `squares[i]` into some const) while performance is indeed improved a little.
To have it merged though this PR needs to be rebased as this tutorial was moved to other file.

PR #1882, opened 2019-03-29
Clarify return value of ReactDOM.render

bouzlibop:
*   changed `stateless components` to `function components`
*   added rule for built-in components


threepointone:
We're moving away from having a .render() return a value at all anyway, so I have to push back on your effort. Thank you for the PR!

PR #1884, opened 2019-03-30
Fixes outline style in Section

suyalcinkaya:
Fixes:
issue image


kevjin:
@lex111 just randomly came across this PR, but personally I think having no outline looks better. I can understand why it would be there for accessibility purposes though, especially for users navigating the website on a keyboard.

lex111:
@kevjin I understand this, but just here accessibility is of great importance.

suyalcinkaya:
Because it looks much better visually without the outline? :)

threepointone:
The outline is an accessibility concern, and doesn't affect visually atypical readers. Thanks for the PR!

PR #1888, opened 2019-04-01
Add new routing framework

pickypg:
This adds the `hookrouter` routing framework to the list of known community routing frameworks.

I happened to come across this framework while looking at alternatives. It's quite small and simple
to use.

/cc @Paratron


gaearon:
We've removed this page.

PR #1892, opened 2019-04-03
Update hooks-rules.md

thomasjohn:
Without [changeVar] why not to remove Effects list and create new one on every render call.


threepointone:
It's absolutely fine to run it like so. Thank you for the PR!

PR #1895, opened 2019-04-03
Update useReducer hook reference example

henrikgs:
The `initialState` prop that may be passed to the `Counter` component shadows the `initialState` constant declared at the top of the example. If the example code is run without passing `initialState` the code will fail with the error `Uncaught TypeError: Cannot read property 'count' of undefined`.

Before: https://codesandbox.io/s/jzm0k3o4y9
After: https://codesandbox.io/s/506v6j107k

### DIFF
--- a/content/docs/hooks-reference.md
+++ b/content/docs/hooks-reference.md
@@ -226,7 +226,7 @@ function reducer(state, action) {
   }
 }
 
-function Counter({initialState}) {
+function Counter() {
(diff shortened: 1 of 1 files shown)


henrikgs:
@lex111 https://github.com/reactjs/reactjs.org/pull/1992 fixed this (even though my PR was more than a month earlier 😞). Can be closed.

lex111:
@henrikgs sorry man :(

PR #1896, opened 2019-04-03
Better efficiency

AkashPE:
I think creating a new function `handleStatusChange` on every render might be unnecessary.


threepointone:
Your change doesn't 'fix' the problem, since it'll still be created for every render. 

That said, this is actually fairly cheap to do, and we don't recommend hoisting in this scenario. Thank you for the PR!

PR #1905, opened 2019-04-05
Add explicit heading ids

taehwanno:
Related to reactjs/ko.reactjs.org#74

I add explicit heading ids with `yarn generate-ids`.


taehwanno:
resolved by 066479b. I will close this.

PR #1910, opened 2019-04-08
Mark Xes on X's turn and Os on O's turn

ddohler:
The current code seems to have its logic backwards: when X plays next, it marks Xes, but it should be marking Os because if X's turn is _next_ then the current turn is O. An alternative way to address this would be to adjust the status text to say something like "Current player: " rather than "Next player: ". I'm happy to adjust this PR or close it if that would be a preferable solution.

I think I've found all the spots in the document that contain this code snippet, but if this is accepted it would also require updates to the Codepen, which I don't have access to. It might also be good to add a note explaining the logic; I'm happy to include that here if someone will point me to the best place to put it.

### DIFF
--- a/content/tutorial/tutorial.md
+++ b/content/tutorial/tutorial.md
@@ -611,7 +611,7 @@ Each time a player moves, `xIsNext` (a boolean) will be flipped to determine whi
 ```javascript{3,6}
   handleClick(i) {
     const squares = this.state.squares.slice();
-    squares[i] = this.state.xIsNext ? 'X' : 'O';
+    squares[i] = this.state.xIsNext ? 'O' : 'X';
     this.setState({
       squares: squares,
       xIsNext: !this.state.xIsNext,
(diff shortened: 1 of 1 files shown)


BartoszKlonowski:
Hello @ddohler, thank you for the contribution!
Unfortunately the latest version of docs do not have this example, so this change is no longer needed.
Let me close this PR as outdated, but please feel invited to create new PRs for other open issues.
Thank you!

content/tutorial/tutorial.md
```diff
@@ -611,7 +611,7 @@ Each time a player moves, `xIsNext` (a boolean) will be flipped to determine whi
 ```javascript{3,6}
   handleClick(i) {
-    squares[i] = this.state.xIsNext ? 'X' : 'O';
+    squares[i] = this.state.xIsNext ? 'O' : 'X';
```
BartoszKlonowski, line 614: The `xIsNext` variable points the next player, but is switched only **after** the move is done
(see line 617: `xIsNext: !this.state.xIsNext,`).
So, we have next player being "X", -> move is done, -> "X" is marked, -> next player is switched to "O" -> and loop.

I compared this implementation to the latest of what is in docs, and see the same mechanism being implemented, and checking the example running I also verify it works as expected.

BartoszKlonowski:
@ddohler,
I checked the implementation and I see no reason for having it rebased and merged into the latest docs, as the implementation is correct and example works as expected.
Let me close this PR as no longer needed, however, like already said, please feel encouraged to create new PR for any open issue.
Thanks again!

PR #1914, opened 2019-04-10
Fix anchor link offset with fixed header

abdelrahmanrifai:
Anchor links overlaps behind the fixed header on all viewport except `small`. I fixed it by matching `paddingTop` with [the header height on different viewport](https://github.com/reactjs/reactjs.org/blob/bab2a3491a18e7074799272d60c8a159473bb7f5/src/components/LayoutHeader/Header.js#L36-L42).

Open [this URL](https://reactjs.org/docs/error-boundaries.html#how-about-event-handlers) for example in a new tab/window to see the issue.

reactorg

### DIFF
--- a/src/theme.js
+++ b/src/theme.js
@@ -350,8 +350,8 @@ const sharedStyles = {
       '::before': {
         content: ' ',
         display: 'block',
-        paddingTop: 90,
-        marginTop: -65,
+        paddingTop: 85,
+        marginTop: -60,
       },
 
       fontSize: 20,
--- a/src/theme.js
+++ b/src/theme.js
@@ -300,8 +300,8 @@ const sharedStyles = {
           content: ' ',
           display: 'block',
           borderBottom: 0,
-          marginTop: -80,
           paddingTop: 40,
+          marginTop: -80,
         },
       },
 
--- a/src/theme.js
+++ b/src/theme.js
@@ -341,8 +341,16 @@ const sharedStyles = {
       fontSize: 20,
       color: colors.subtle,
       lineHeight: 1.3,
-      marginTop: 50,
+      paddingTop: 65,
       fontWeight: 400,
+
+      [media.between('small', 'large')]: {
+        paddingTop: 55,
+      },
+
+      [media.lessThan('small')]: {
+        paddingTop: 45,
+      },
     },
 
     '& h4 + p': {
--- a/src/theme.js
+++ b/src/theme.js
@@ -285,7 +285,7 @@ const sharedStyles = {
     },
 
     '& h2': {
-      '&::before': {
+      '::before': {
         content: ' ',
         display: 'block',
         borderBottom: `1px solid ${colors.divider}`,
@@ -296,9 +296,13 @@ const sharedStyles = {
       lineHeight: 1.2,
 
       ':first-child': {
-        borderTop: 0,
-        marginTop: 0,
-        paddingTop: 0,
+        '::before': {
+          content: ' ',
+          display: 'block',
+          borderBottom: 0,
+          marginTop: -80,
+          paddingTop: 40,
+        },
       },
 
       [media.lessThan('large')]: {
@@ -315,11 +319,11 @@ const sharedStyles = {
     },
 
     '& h3': {
-      '&::before': {
+      '::before': {
         content: ' ',
         display: 'block',
-        paddingTop: 60,
-        marginTop: -15,
+        paddingTop: 90,
+        marginTop: -45,
       },
 
       [media.lessThan('small')]: {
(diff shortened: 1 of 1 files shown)


EPolovniy:
This fix work only for h2.
I propose deepest changes for all headers https://github.com/reactjs/reactjs.org/pull/1855

abdelrahmanrifai:
@EPolovniy you're right, however making the rule more generic with all headlines should be relatively easy, I will revise it.

_I will have a look at your PR too, not straightforward from first look!_

EPolovniy:
#1944 I created  more clear request for this problem. Review my pull request pls!

EPolovniy:
#1944

abdelrahmanrifai:
@EPolovniy @lex111 

My the new changes in https://github.com/reactjs/reactjs.org/pull/1914/commits/88f04250a1c188b090956adb1063dfeea2a8ebbb is now handling both `h2` `h3` or even nested `h3` inside `h2`.

_**You can clone and run locally to test if needed!**_ because I was expecting @reactjs-bot to create a new build for https://github.com/reactjs/reactjs.org/pull/1914/commits/88f04250a1c188b090956adb1063dfeea2a8ebbb so you review my fixes live, but it didn't!

---

Also I noticed some inconsistency of using `h3` directly after `h1` instead of `h2`, examples:

- https://reactjs.org/docs/introducing-jsx.html
- https://reactjs.org/docs/lists-and-keys.html

You will also notice that the border between sections is not there as in other pages, for example:

- https://reactjs.org/docs/lifting-state-up.html
- https://reactjs.org/docs/forms.html

However that's out of this PR scope. I will adjust that in a following PR.

lex111:
> because I was expecting @reactjs-bot to create a new build for 88f0425 so you review my fixes live, but it didn't!

In fact, it is. Your new changes are available at the same link!

lex111:
There is a problem with `h4`.

https://deploy-preview-1914--reactjs.netlify.com/docs/faq-functions.html#bind-in-constructor-es2015

abdelrahmanrifai:
@lex111 nice! maybe it takes few minutes, because I checked right after I pushed my changes wasn't there.

I was wondering if h4 is being used as an anchor, good catch, will update it now!

lex111:
@abdelrahmanrifai I've been thinking, maybe there is another solution to the problem. For example, look please this one https://stackoverflow.com/questions/4086107/fixed-page-header-overlaps-in-page-anchors/28824157#28824157

abdelrahmanrifai:
@lex111 looking into applying a solution with `:target`, looks more simple and consistent!

lex111:
@abdelrahmanrifai now let's see it in action! :yum:

abdelrahmanrifai:
@lex111 I tested both `:target` and `:target::before`, none worked well because:

Target elements `h2`, `h3`, and `h4` already have different `margin` and `padding` applied that is part of the design, and sometimes even we have special conditions like [here](https://github.com/reactjs/reactjs.org/blob/master/src/theme.js#L293)

To apply this `:target` trick, it will need much more refactoring efforts, that I'm not sure will work well.

I see the current status of my PR works good enough to fix the problem.

abdelrahmanrifai:
Another drawback related to `:target`: it only work on click, not when opening the URL in a new window/tab.

lex111:
@abdelrahmanrifai okay, but what about the alternative?

abdelrahmanrifai:
@lex111 I see [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/:target#Syntax), [CSS-Tricks](https://css-tricks.com/on-target/#article-header-id-3) and [Stack Overflow](https://stackoverflow.com/questions/4086107/fixed-page-header-overlaps-in-page-anchors/28824157#28824157) are using same technique as in my PR.

Actually that technique is already applied for [reactjs.org](https://reactjs.org/) only for mobile, my PR is just adding more media queries and using [corresponding header heights](https://github.com/reactjs/reactjs.org/blob/bab2a3491a18e7074799272d60c8a159473bb7f5/src/components/LayoutHeader/Header.js#L36-L42) + `5px` so the headlines don't touch header edge.

I find it so interesting issue, I'd spend more time researching it later. However I still see my PR is good enough for now, what do you think?

EPolovniy:
This PR change margins between blocks.

abdelrahmanrifai:
@EPolovniy yes it does add few pixels more, however nothing harmful by any means. In contrast it's better for readability check [CSS-Tricks](https://css-tricks.com/on-target/#article-header-id-3) for example. It can be addressed and get decreased as it was before if required anyway.

lex111:
@abdelrahmanrifai note that the margin of the `h4` tag has changed:

https://deploy-preview-1914--reactjs.netlify.com/docs/faq-functions.html#how-do-i-pass-an-event-handler-like-onclick-to-a-component
https://reactjs.org/docs/faq-functions.html#how-do-i-pass-an-event-handler-like-onclick-to-a-component

abdelrahmanrifai:
@lex111 Thanks! I just noticed some special conditions flaws too with `:first-child` or `:first-of-type`, addressing them all now.

abdelrahmanrifai:
@EPolovniy @lex111 My PR is ready to be reviewed now!

No single UI change should be there now, same distance above all heading. 🙌

final

lex111:
@abdelrahmanrifai this PR is ready for merge?

abdelrahmanrifai:
@lex111 Yes it's ready!

I just noticed this so tiny extra pixel in my previous screenshot. that's what this `pixel perfection` commit all about!

_You may need to zoom to see it!_
pixel3
pixel1
pixel2

lex111:
@abdelrahmanrifai :confused: 

Before:

image



After

image

lex111:
@abdelrahmanrifai and @EPolovniy, thank you so much both for your help! :hugs:

PR #1922, opened 2019-04-14
add add-react-redux-thunk

tibabit:
add add-react-redux-thunk in other starter kits section


gaearon:
We've removed this page.

PR #1925, opened 2019-04-15
Don't recommend shallow rendering inside TestUtils docs

threepointone:
They're already using react-dom, so shallow rendering is a drastic change to their setup. `jest.mock` is more suitable in this scenario. 

(should I mention "or your framework's mocking helpers" to not be jest specific?)


alexkrolick:
Adding the bit for non-Jest users would be nice

PR #1927, opened 2019-04-16
Changed unicode hyphen to normal hyphen

BrainBoxBrown:
When copy and pasting `npm install jest-cli —save-dev` you get this error
`Invalid tag name "—save-dev": Tags may not have any characters that encodeURIComponent encodes`

I replaced the unicode characters '—' with ascii hypens '-'


lex111:
@BrainBoxBrown thank you!

Issue #1799, opened 2019-03-09
Amend documentation for skipping useEffect 


Pomax:
This is a documentation improvement request. 

Currently https://reactjs.org/docs/hooks-effect.html#tip-optimizing-performance-by-skipping-effects explains skipping a `useEffect` based on a state variable by showing the following code:

```javascript
useEffect(() => {
  document.title = `You clicked ${count} times`;
}, [count]); // Only re-run the effect if count changes
```

With the following explanation:

> In the example above, we pass [count] as the second argument. What does this mean? If the count is 5, and then our component re-renders with count still equal to 5, React will compare [5] from the previous render and [5] from the next render. Because all items in the array are the same (5 === 5), React would skip the effect. That’s our optimization.

However, as the array we're shown only has a single element it, it is ambiguous what "all the items in the array" means. E.g. we cannot tell whether that's intentional phrasing, or just plain English that isn't actually tied to any technical precision.

Could this phrasing either be made more explicit, or (ideally) can an extra paragraph be added to explain what happens to an array with more than one element?

(originally filed as from https://github.com/facebook/react/issues/15068)

enBonnet:
I was thinking the same thing

knanco:
Had the same question!

PR #1932, opened 2019-04-17
Add multiple items on useEffect example

enBonnet:
Related to issue #1799


content/docs/hooks-effect.md
```diff
@@ -442,15 +442,19 @@ componentDidUpdate(prevProps, prevState) {
...
 
-```js{3}
+```js{3,7}
 useEffect(() => {
   document.title = `You clicked ${count} times`;
 }, [count]); // Only re-run the effect if count changes
+
+useEffect(() => {
+  document.title = `You clicked ${count} times`;
+}, [count, other]); // Only re-run if count or other changes
```
Pomax, line 452: this probably makes more sense as a single example,
```
useEffect(() => {
  document.title = `${name} clicked ${count} times`;
}, [name, count]); // Only re-run when any of the indicated variables change
```

PR #1933, opened 2019-04-17
Update renderToString()

dnkodi:
It is important to keep in mind that renderToString() is synchronous and it will block the server while it runs causing slightly poor performance than expected.


threepointone:
I don't think this is necessary to mention here, especially since we mention renderToNodeStream right after. Thank you for the PR!

PR #1936, opened 2019-04-19
Updated input text field state value field name to avoid confusion

dandoman:
When the word 'value' is used as the state field name, it is possible that the reader will get confused and assume that they need to use the same variable name in the event.target.x field when this is not the case.


rickhanlonii:
Sorry for missing this, closing since the page was rewritten and doesn't have this flaw https://react.dev/reference/react-dom/components/input#controlling-an-input-with-a-state-variable

PR #1937, opened 2019-04-19
change 'lifecycle' to 'lifecycle event'

dcwilliams1:
The lifecycle is not the same as a lifecycle event, so you can't abbreviate 'lifecycle even' with 'lifecycle'. There are many events in one lifecycle. If you're writing about one of those events, you should write 'lifecycle event' to be clear.


threepointone:
I want to push back and disagree, we use the term 'lifecycle' to mean the function/method that's called. With that context, the doc is actually fine. Thank you for your PR!

PR #1940, opened 2019-04-20
Removed unnecessary wrapper component

plug-n-play:
Removed Unnecessary wrapper component. While copy-pasting this e.g. to local create-react-app, I had to do this change

(Apart from adding following to index.css:)
```css
html,
body {
  height: 100%;
}
 #root {
  height: 100%;
} 
```


threepointone:
Thank you!

PR #1942, opened 2019-04-22
docs: reorder past conference

jepser:
It makes sense to have ascending order for future conferences, but for past conferences, the latest it would be more relevant since it has more updated content and topics that interest the community.


harish-sethuraman:
The conference page seems to be updated right now. Thanks for sending PR and sorry for the late turn around time.

PR #1943, opened 2019-04-24
add useRef Note

nikgraf:
In a workshop the question came up why not use `createRef` instead of `useRef`


rickhanlonii:
Sorry we missed this, closing since the page is rewritten and there are caveats and notes in the new docs: https://react.dev/reference/react/createRef

rickhanlonii:
Also hi @nikgraf!

PR #1948, opened 2019-04-25
adds React Boston 2019 to conference page

ehermanson:


lex111:
@ehermanson thank you!

PR #1951, opened 2019-04-28
Broken anchor link in Advanced Concepts.

0xnoob:
The link should probably either go to directly to the beginning of main concepts (which is what my proposes change does), or to the anchor  -tag of "Step-by-Step Guide" (#step-by-step-guide). There is no other link in the article which refers to another anchor-tag, that's why I chose to directly link to the main-concepts.


lex111:
@0xnoob thank you!

PR #1952, opened 2019-04-29
Document the [deps] argument to useEffect()

### LINKED ISSUE #1825: Suggest improving hook API reference signatures
A user on Reddit pointed out that the Hooks API Reference page section for `useEffect` doesn't include the deps array in the signature - it's just `useEffect(callback)`.  The deps array is only described later on under the heading "Conditionally Firing an Effect".

Similarly, the entry for `useLayoutEffect()` doesn't even show a signature at all - it just says "the signature is identical to `useEffect`".

At a minimum, the `useEffect` signature should show that there's an optional second argument - the deps array.  It would also help to show the signature for `useLayoutEffect` as well.

Reddit link reference: https://www.reddit.com/r/reactjs/comments/b0oh20/usedimensions_a_react_hook_to_measure_dom_nodes/eigiqud/

tmcw:
This PR treads similar ground as #1342, but is much more constrained and is a straightforward clarification rather than a jumping-off point. It fixes #1825.

This documents the optional `[deps]` second argument to `useEffect()` the same way that it is documented for `useImperativeHandle`.

### DIFF
--- a/content/docs/hooks-reference.md
+++ b/content/docs/hooks-reference.md
@@ -102,7 +102,7 @@ Note that React may still need to render that specific component again before ba
 ### `useEffect` {#useeffect}
 
 ```js
-useEffect(didUpdate);
+useEffect(didUpdate, [deps]);
 ```
 
 Accepts a function that contains imperative, possibly effectful code.


tmcw:
Ping? This seems like a critical missing piece of documentation, right? It's one of the most-used hooks, documenting an often-used argument to it?

alexkrolick:
Isn't this covered in the section below? (#conditionally-firing-an-effect)

tmcw:
Yes, but the other function signatures in the list (like `useImperativeHandle`) don't rely on body text to explain. If this section of documentation is supposed to be an API signature, it should include the arguments, right?

PR #1957, opened 2019-04-30
inputs array > dependencies array

kentcdodds:
I think initially this was referred to as the "inputs array" in the docs, then it was changed to the "dependencies array" everywhere except it looks like a few were missed.


kentcdodds:
I almost tweeted this 😬

image

Then decided to look around and I _think_ it's `dependencies` :) Better to keep things consistent and avoid confusion!

kentcdodds:
Ping. This should be a quick merge.

kentcdodds:
Oh! @alexkrolick! Fancy meeting you here. I had no idea you were a maintainer of reactjs.org too. Very cool 💯

PR #1959, opened 2019-05-01
previous page link not linking to previous page, but to the second last.

0xnoob:
The example code is on both, the `Introducing Hooks` as well as `Hooks at a Glance` page, but the linked page is the `Introducing Hooks` and I think that's the right one. The `Hooks at a Glance` page probably got added later?

Not sure about the phrasing so, having "introduce" twice in one sentence ...

### DIFF
--- a/content/docs/hooks-state.md
+++ b/content/docs/hooks-state.md
@@ -8,7 +8,7 @@ prev: hooks-overview.html
 
 *Hooks* are a new addition in React 16.8. They let you use state and other React features without writing a class.
 
-The [previous page](/docs/hooks-intro.html) introduced Hooks with this example:
+The [introduction page](/docs/hooks-intro.html) used this example to introduce Hooks:
 
 ```js{4-5}
 import React, { useState } from 'react';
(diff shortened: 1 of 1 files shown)


content/docs/hooks-state.md
```diff
@@ -8,7 +8,7 @@ prev: hooks-overview.html
 
 *Hooks* are a new addition in React 16.8. They let you use state and other React features without writing a class.
 
-The [previous page](/docs/hooks-intro.html) introduced Hooks with this example:
+The [introduction page](/docs/hooks-intro.html) used this example to introduce Hooks:
```
lex111: How about you rephrase it?

```suggestion
The [introduction page](/docs/hooks-intro.html) used this example to get familiar with Hooks:
```

0xnoob: I like that rephrasing!

lex111: Great, you can apply it if you click on the "Commit suggestion" button.

lex111:
@0xnoob thank you!

PR #1962, opened 2019-05-02
Mirror optimize demo code

Kaiyokun:
1. Exchange the order of short-circuiting operands in `calculateWinner(squares) || squares[i]`
which function calling is expensive than indexing an array.
2. Clarify the conditional(ternary) expression in `history.map((step, move) => { const desc = move ? ...`
which `move` is in `0, 1, 2, ...`, so `'Go to game start'` should appear before `'Go to move #' + move`, and
`move ? ...` may change to `(0 === move) ? ...` respectively.

### DIFF
--- a/content/tutorial/tutorial.md
+++ b/content/tutorial/tutorial.md
@@ -739,8 +739,8 @@ We can now change the Board's `handleClick` function to return early by ignoring
 ```javascript{3-5}
   handleClick(i) {
     const squares = this.state.squares.slice();
-    if (calculateWinner(squares) || squares[i]) {
-      return;
+    if (squares[i] || calculateWinner(squares)) {
+      return; 
     }
     squares[i] = this.state.xIsNext ? 'X' : 'O';
     this.setState({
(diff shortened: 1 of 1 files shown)


content/tutorial/tutorial.md
```diff
@@ -1007,10 +1007,10 @@ Let's `map` over the `history` in the Game's `render` method:
     const current = history[history.length - 1];
     const winner = calculateWinner(current.squares);
 
-    const moves = history.map((step, move) => {
-      const desc = move ?
-        'Go to move #' + move :
-        'Go to game start';
+    const moves = history.map((_, move) => {
```
lex111: How about to leave the name first parameter (`step`) as it was before?

Kaiyokun: Sounds good, I'd like to suggest you also add a prefix `_` to it (`_step`) as jslint suggestion, which may act as a good guideline to audience that implies the first parameter is unused.

lex111: Let's leave it for now. Okay?

Kaiyokun: At your disposal.

PR #1966, opened 2019-05-03
Removing props params from functional component

siddtheone:
It is causing confusion with initialState defined on top, and expecting Counter component to have prop with same name.

### DIFF
--- a/content/docs/hooks-reference.md
+++ b/content/docs/hooks-reference.md
@@ -226,7 +226,7 @@ function reducer(state, action) {
   }
 }
 
-function Counter({initialState}) {
+function Counter() {
   const [state, dispatch] = useReducer(reducer, initialState);
   return (
     <>


content/docs/hooks-reference.md
```diff
@@ -226,7 +226,7 @@ function reducer(state, action) {
   }
 }
 
-function Counter({initialState}) {
+function Counter() {
```
lex111: This is necessary, look at line 216.

siddtheone: This is required for line
`const [state, dispatch] = useReducer(reducer, initialState);`

Since Counter is a functional component, parameter it should accept is its **props** not the **initial state**

threepointone:
As pointed out, this is expected. Thank you for the PR!

PR #1970, opened 2019-05-06
Add example website to Render props

lekoaf:
Added Apollo GraphQL as an example of [Render props](https://reactjs.org/docs/render-props.html), since it's a high profile library for Facebooks GraphQL API.


lekoaf:
CLA signed.

yvesgurcan:
Good idea!

PR #1976, opened 2019-05-07
Update docs/hooks-state.md

kambleaa007:
Prefer functional updates version in hooks documentation. 
Updating State: second example is functional, 
strongly seeks reference to [functional-updates](https://reactjs.org/docs/hooks-reference.html#functional-updates)


threepointone:
This doesn't appear to add any value since it's already explained right before it. Thank you for the PR!

PR #1977, opened 2019-05-07
fix(terms): treeshaking -> tree shaking

neighborhood999:


lex111:
@neighborhood999 thank you!

PR #1978, opened 2019-05-07
Substitute uglify-js with terser in Browserify section

tondi:
Package uglify-js referenced in docs is not longer maintained and has a good replacement in [terser](https://github.com/terser-js/terser)

From [terser readme](https://github.com/terser-js/terser):
> terser is a fork of uglify-es that retains API and CLI compatibility with uglify-es and uglify-js@3.

### DIFF
--- a/content/docs/optimizing-performance.md
+++ b/content/docs/optimizing-performance.md
@@ -71,35 +71,28 @@ Remember that you only need to do this for production builds. You shouldn't pass
 
 ### Browserify {#browserify}
 
-For the most efficient Browserify production build, install a few plugins:
+For the most efficient Browserify production build, install a couple of plugins:
 
 ```
 # If you use npm
-npm install --save-dev envify terser uglifyify 
+npm install --save-dev envify terser
 
 # If you use Yarn
-yarn add --dev envify terser uglifyify 
+yarn add --dev envify terser
 ```
 
 To create a production build, make sure that you add these transforms **(the order matters)**:
 
 * The [`envify`](https://github.com/hughsk/envify) transform ensures the right build environment is set. Make it global (`-g`).
-* The [`uglifyify`](https://github.com/hughsk/uglifyify) transform removes development imports. Make it global too (`-g`).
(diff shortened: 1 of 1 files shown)


content/docs/optimizing-performance.md
```diff
@@ -75,31 +75,31 @@ For the most efficient Browserify production build, install a few plugins:
...
-yarn add --dev envify uglify-js uglifyify 
+yarn add --dev envify terser uglifyify 
 ```
 
 To create a production build, make sure that you add these transforms **(the order matters)**:
 
 * The [`envify`](https://github.com/hughsk/envify) transform ensures the right build environment is set. Make it global (`-g`).
 * The [`uglifyify`](https://github.com/hughsk/uglifyify) transform removes development imports. Make it global too (`-g`).
-* Finally, the resulting bundle is piped to [`uglify-js`](https://github.com/mishoo/UglifyJS2) for mangling ([read why](https://github.com/hughsk/uglifyify#motivationusage)).
+* Finally, the resulting bundle is piped to [`terser`](https://github.com/terser-js/terser) for mangling ([read why](https://github.com/hughsk/uglifyify#motivationusage)).
```
mterczynski: Do you want to put link with title "read why" that points to uglify documentation?

content/docs/optimizing-performance.md
```diff
@@ -75,31 +75,31 @@ For the most efficient Browserify production build, install a few plugins:
...
 
 For example:
 
 ```
 browserify ./index.js \
   -g [ envify --NODE_ENV production ] \
   -g uglifyify \
```
gaearon: What about `uglifyify`?

tondi: It seems that it also uses `terser` under the hood as stated [here](https://github.com/hughsk/uglifyify#uglifyify). 
I tried removing it from these transforms and looks like `terser` handles dead code removal as well - there is no difference in resulting bundle size and therefore no need to use `uglifyify`

I'll update the doc

content/docs/optimizing-performance.md
tondi: Removed 👍

tondi:
Outdated, closing

PR #1979, opened 2019-05-09
drop unused React import for hooks example

rajington:
probably just an accidental copy/paste from the JSX example


threepointone:
good catch, thanks!

PR #1985, opened 2019-05-11
Update regarding JSX in-browser

rajputji:
added a note regarding the babel standalone for trying JSX in browser itself


rickhanlonii:
Sorry we missed this. Closing since we rewrote all of these pages, so it doesn't seem relevant any more. If you think there's something we missed, feel free to open a new PR.

PR #1986, opened 2019-05-11
Minesweeper implementation with react and redux

mjamilfarooq:


gaearon:
Thanks, but we're trying to keep this page focused on plain React examples.

PR #1988, opened 2019-05-12
Add ReactQL to starter kit list

leebenson:
Adds [ReactQL](https://github.com/leebenson/reactql) to starter kit list


gaearon:
We've removed this page.

Issue #1938, opened 2019-04-20
Higher-Order Components


Learner16:
Under the explanation of "[Don’t Mutate the Original Component. Use Composition.](https://reactjs.org/docs/higher-order-components.html#dont-mutate-the-original-component-use-composition)"

> The comment shall say : The fact that we're **NOT** returning the original input is a hint that it has been mutated.

```
function logProps(InputComponent) {
  InputComponent.prototype.componentWillReceiveProps = function(nextProps) {
    console.log('Current props: ', this.props);
    console.log('Next props: ', nextProps);
  };
  // The fact that we're not returning the original input is a hint that it has
  // been mutated.
  return InputComponent;
}

// EnhancedComponent will log whenever props are received
const EnhancedComponent = logProps(InputComponent);
```

kambleaa007:
Yes, good catch,
you can check PR,
https://github.com/reactjs/reactjs.org/pull/1989

threepointone:
As pointed out on the PR, the docs are fine

PR #1989, opened 2019-05-13
Update lifting-state-up.md

kambleaa007:

### DIFF
--- a/content/docs/higher-order-components.md
+++ b/content/docs/higher-order-components.md
@@ -181,7 +181,7 @@ function logProps(InputComponent) {
     console.log('Current props: ', this.props);
     console.log('Next props: ', nextProps);
   };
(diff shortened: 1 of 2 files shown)


kambleaa007:
https://github.com/reactjs/reactjs.org/issues/1938
#1938

content/docs/higher-order-components.md
```diff
@@ -181,7 +181,7 @@ function logProps(InputComponent) {
     console.log('Current props: ', this.props);
     console.log('Next props: ', nextProps);
   };
-  // The fact that we're returning the original input is a hint that it has
+  // The fact that we're not returning the original input is a hint that it has
```
threepointone: this doesn't seem right, we *are* returning the original input.

PR #1990, opened 2019-05-13
Clarify useEffect behavior for empty dependency array

ryancogswell:
The current wording "If you want to run an effect and clean it up only once (on mount and unmount), you can pass an empty array ([]) as a second argument." is confusing some people. They are interpreting this as meaning that the effect will run on mount **and** on unmount. See cbdev420's comment on my answer [here](https://stackoverflow.com/questions/56115295/how-to-set-up-an-event-listener-and-remove-it-after-first-fired-event-using-reac/56115564?noredirect=1#comment98864939_56115564).


rickhanlonii:
Sorry we missed this. Closing since we rewrote all of these docs, but if you see this issue still present in the new docs feel free to submit a new PR.

PR #1994, opened 2019-05-14
update faq on perf with function props

alexkrolick:


sophiebits:
This looks perfect, thanks!

PR #1996, opened 2019-05-14
Updated minor comments

suryarsp:
Beginners might struggle without that comment


threepointone:
This doesn't seem as helpful, but thank you for the PR!

PR #1998, opened 2019-05-14
Access second result item to avoid `ignored`

agorf:


threepointone:
This isn't typically how we destructure this hook, and would probably confuse more. The proper fix is to rename `ignored` to `_ignored`. That said, I wouldn't recommend changing that here anyway. Thank you for the PR!

PR #2009, opened 2019-05-17
Update cdn-links.md - clarifying what a CDN is

chevybowtie:


lex111:
Agree with @starkblaze01.

@chevybowtie thanks for the PR, but I will close it because there is no point in it.

PR #2011, opened 2019-05-18
Fix broken webpack production guide link in "Optimizing Performance" page.

MatthewWid:
**Before**

[The link to the webpack production build guide](https://reactjs.org/docs/optimizing-performance.html#webpack) went to https://webpack.js.org/guides/production-build/ which results in a 404 Error.

image

**After**

The link now directs to https://webpack.js.org/guides/production/ which is the correct URL.

image


lex111:
@MatthewWid thank you!

PR #2012, opened 2019-05-18
Add Real World React to meetups.md

dunphyben:
Add Real World React (San Francisco) meetup group


lex111:
@dunphyben thank you!

PR #2013, opened 2019-05-19
Add function call to useEffect docs example

cdock1029:


cdock1029:
Exists already
https://github.com/reactjs/reactjs.org/pull/1999

PR #2016, opened 2019-05-20
Change 'operator' to 'method'

carstonhenderson:
I believe `.slice()` is an array prototype method, rather than an operator.


lex111:
@carstonhenderson thank you!

PR #2018, opened 2019-05-20
Update hooks-reference.md confusion about `useReducer`.

mycolaos:

### DIFF
--- a/content/docs/hooks-reference.md
+++ b/content/docs/hooks-reference.md
@@ -242,6 +242,28 @@ function Counter() {
 >
 >React guarantees that `dispatch` function identity is stable and won't change on re-renders. This is why it's safe to omit from the `useEffect` or `useCallback` dependency list.
 
+To avoid confusion of the Redux users, it worth noting that `useReducer` holds an isolated state just like `useState`, and is not connected to a single combined store. Using the `useReducer(todosReducer)` between different components will produce separated states, and the **`dispatch` function dispatches only to its own reducer**.
+
+```
+function MultiCounter() {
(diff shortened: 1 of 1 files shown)


content/docs/hooks-reference.md
```diff
@@ -242,6 +242,28 @@ function Counter() {
 >
 >React guarantees that `dispatch` function identity is stable and won't change on re-renders. This is why it's safe to omit from the `useEffect` or `useCallback` dependency list.
 
+To avoid confusion of the Redux users, it worth noting that `useReducer` holds an isolated state just like `useState`, and is not connected to a single combined store. Using the `useReducer(todosReducer)` between different components will produce separated states, and the **`dispatch` function dispatches only to its own reducer**.
```
ahtee: it's

content/docs/hooks-reference.md
```diff
@@ -242,6 +242,28 @@ function Counter() {
...
+
+```
+function MultiCounter() {
+  // Isolated states. Each `dispatch` notifies only its own `reducer`.
```
ahtee: you can probably remove template literals here (\` \`)

content/docs/hooks-reference.md
```diff
@@ -242,6 +242,28 @@ function Counter() {
...
+  const [state1, dispatch1] = useReducer(countReducer, initialState1);
+  const [state2, dispatch2] = useReducer(countReducer, initialState2);
```
ahtee: possibly a real-world use case would be better for these var names

ahtee:
Also, make sure you sign the SLA

PR #2019, opened 2019-05-20
Add ReactNext 2019

adamkleingit:
Add ReactNext 2019 conference to list of conferences. Thanks


adamkleingit:
Thanks!
Do we need to sign the CLA?

gaearon:
Wouldn't hurt. It's a technicality with this kind of change and not strictly necessary but it would probably make our internal systems happier.

adamkleingit:
If your internal systems are happier then I'm happier. We'll sign

PR #2033, opened 2019-05-24
Add lang and hreflang to language pages for screen reader UX

M0nica:
Previous HTML
`<a href="https://fr.reactjs.org/" rel="nofollow">Français</a>`
Updated HTML
`<a href="https://fr.reactjs.org/" rel="nofollow" lang="fr" hreflang="fr">Français</a>`

I wanted to add `lang` and `hreflang` to the link element for translated languages at https://reactjs.org/languages.  `Lang` is used by screen readers to switch language profiles to provide the correct accent and pronunciation and `hreflang` seems to help with SEO

>If you have multiple versions of a page for different languages or regions, tell Google about these different variations. Doing so will help Google Search point users to the most appropriate version of your page by language or region.
> Note that even without taking action, Google might still find alternate language versions of your page, but it is usually best for you to explicitly indicate your language- or region-specific pages.

Source: https://support.google.com/webmasters/answer/189077?hl=en

### DIFF
--- a/src/components/TitleAndMetaTags/TitleAndMetaTags.js
+++ b/src/components/TitleAndMetaTags/TitleAndMetaTags.js
@@ -29,7 +29,7 @@ const alternatePages = canonicalUrl => {
   const pages = languagesToLinkTo.map(language => (
     <link
       rel="alternate"
-      hreflang={language.code}
+      hreflang={language.code === 'en' ? 'x-default' : language.code}
       href={canonicalUrl.replace(
         urlRoot,
         `https://${
--- a/src/components/TitleAndMetaTags/TitleAndMetaTags.js
+++ b/src/components/TitleAndMetaTags/TitleAndMetaTags.js
@@ -28,8 +28,9 @@ const completeLanguages = languages.filter(language => {
 const alternatePages = canonicalUrl => {
   return completeLanguages.map(language => (
     <link
+      key={('alt-', language.code)}
       rel="alternate"
-      hreflang={language.code === 'en' ? 'x-default' : language.code}
+      hreflang={language.code}
       href={canonicalUrl.replace(
         urlRoot,
         `https://${
@@ -38,6 +39,10 @@ const alternatePages = canonicalUrl => {
       )}
     />
   ));
+};
+
+const defaultPage = canonicalUrl => {
+  return canonicalUrl.replace(urlRoot, 'https://reactjs.org');
 };
 
 const TitleAndMetaTags = ({title, ogDescription, canonicalUrl}: Props) => {
@@ -53,6 +58,13 @@ const TitleAndMetaTags = ({title, ogDescription, canonicalUrl}: Props) => {
       />
       <meta property="fb:app_id" content="623268441017527" />
       {canonicalUrl && <link rel="canonical" href={canonicalUrl} />}
+      {canonicalUrl && (
+        <link
+          rel="alternate"
+          href={defaultPage(canonicalUrl)}
+          hreflang="x-default"
+        />
+      )}
       {canonicalUrl && alternatePages(canonicalUrl)}
     </Helmet>
   );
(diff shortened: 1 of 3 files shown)


M0nica:
> > Each variation of the page should include a set of `<link>` elements in the `<head>` element, one link for each page variant including itself. The set of links is identical for every version of the page. See the additional guidelines.
> 
> Can we add to the `head`?

@lex111 I just updated to add the alternate links to the head of the page:
Screen Shot 2019-05-25 at 9 45 35 AM

src/components/TitleAndMetaTags/TitleAndMetaTags.js
```diff
@@ -16,6 +19,28 @@ type Props = {
...
 };
 
+// only provide alternate links to pages in languages where 95-100% of core content has been translated
+// which is determined by status enum of 2
+const languagesToLinkTo = languages.filter(language => {
+  return language.status == 2;
+});
+
+const alternatePages = canonicalUrl => {
+  const pages = languagesToLinkTo.map(language => (
```
lex111:
```suggestion
  return completeLanguages.map(language => (
```

src/components/TitleAndMetaTags/TitleAndMetaTags.js
```diff
@@ -16,6 +19,28 @@ type Props = {
   canonicalUrl: string,
 };
 
+// only provide alternate links to pages in languages where 95-100% of core content has been translated
+// which is determined by status enum of 2
+const languagesToLinkTo = languages.filter(language => {
```
lex111:
```suggestion
const completeLanguages = languages.filter(language => {
```

src/components/TitleAndMetaTags/TitleAndMetaTags.js
```diff
@@ -16,6 +19,28 @@ type Props = {
...
+      hreflang={language.code}
+      href={canonicalUrl.replace(
+        urlRoot,
+        `https://${
+          language.code === 'en' ? '' : `${language.code}.`
+        }reactjs.org`,
+      )}
+    />
+  ));
+  return pages;
```
lex111:
```suggestion
```

M0nica:
@lex111 I just updated the `alternatePages()` logic to make English the default
Screen Shot 2019-05-27 at 4 11 52 PM

src/components/TitleAndMetaTags/TitleAndMetaTags.js
```diff
@@ -16,6 +19,27 @@ type Props = {
...
+// which is determined by status enum of 2
+const completeLanguages = languages.filter(language => {
+  return language.status == 2;
+});
+
+const alternatePages = canonicalUrl => {
+  return completeLanguages.map(language => (
+    <link
+      rel="alternate"
+      hreflang={language.code === 'en' ? 'x-default' : language.code}
```
lex111: Considering the subdomains, is this the best way?

```suggestion
      hreflang={language.code === window.location.host.split('.')[0] ? 'x-default' : language.code}
```

M0nica: My understanding is that `x-default` is a global definition and shouldn’t be changed due to locale context. It’s treated by Google as the default version of the site if a site doesn’t exist in a certain language or they are unable to detect a user’s language. https://www.hreflangbuilder.com/complete-guide-to-using-the-using-x-default-hreflang-element/

lex111: @M0nica thanks, I get it, mistakenly thought that in the case of subdomains, we need to mark it as `x-default`.

However, my first question worries me, look at the code of https://www.mozilla.org/en-US/
You see, as I suggested earlier to do two tags - `en` and `x-default` with the same values, I have seen similar on other sites. That is, `x-default` does **not** serve as a replacement for the original language (en).

image

M0nica: Ok, I updated to have all languages that are 100% translated incl. English show up as alternate with their hreflang. And then will also set the `default-x` page as the English version of the page separately. Is that what you had in mind? It seems aligned with Mozilla now.

lex111:
@M0nica thank you so much! :heart:

PR #2037, opened 2019-05-25
Say "disconnecting" not "plugging out" your mouse

brendanfalkowski:
The previous phrase "plugging out" is uncommon in English. Saying "disconnecting" is better than "unplugging" because many mouse users have wireless mice, and is more accurate.


alexkrolick:
Thanks

Issue #2038, opened 2019-05-26
"Quickly Try JSX" does not work on a local file


fuyemi:
**In Docs > Installation > Add React to a Website > Optional : Try React with JSX > Quickly Try JSX,** the instructions say:

> ...add this `<script>` tag to your page:
> ```html
> <script src="https://unpkg.com/babel-standalone@6/babel.min.js"></script>
> ```
> Now you can use JSX in any `<script>` tag by adding `type="text/babel"` attribute to it.

However, after adding these lines to a local HTML file, opening it in browser gives the following error:

> Access to XMLHttpRequest at 'file:///C:/Users/.../like_button.js' from origin 'null' has been blocked by CORS policy: Cross origin requests are only supported for protocol schemes: http, data, chrome, chrome-extension, https.

kambleaa007:
Try to run on firefox,
Run a Local Server
Run a server on your computer, like Apache or Python
Python isn't a server, but it will run a simple server

fuyemi:
Yes, I eventually ended up making a Python server and it resolved the issue. But I think this should be mentioned in the tutorial, because currently, the docs make it seem like simply adding the script to a local file is enough, which isn't the case. This might be a hard issue for a beginner to identify if it's not mentioned explicitly.

alexkrolick:
Adding the babel script from CDN works on an HTML file if you _write the code_ inside script tags, not if you add script tags for external sources. PR welcome to clarify that distinction 😄

trrapp12:
I would like to add that I have been beating my head against the wall for a week on this same issue.  It really does need to be added to the documentation.

PR #2040, opened 2019-05-29
Add clarification for trying JSX on local file

fuyemi:
For issue #2038


joshuasorkin:
Hi, just dropping by to say that I ran into the problem described in https://github.com/reactjs/reactjs.org/issues/2038, and would have been able to diagnose the problem sooner if this commit had been merged.  Hopefully someone with write access can resolve the merge conflicts!

PR #2050, opened 2019-06-03
Use imperative tone for suggested commit message

hugo:
This is both widely recommended as the preferred style and also that in which actual commit messages on this project are written.


alexkrolick:
indeed it is

PR #2052, opened 2019-06-03
Introduce useReducer with the simpler function signature, defer lazy initialization signature to it's section.

danieledler:
**Update:**
I got confused by the changing role of the second argument to `useReducer` from `initialState` to `initialArg` if a third argument is provided. Pull request is updated to instead reduce this risk of confusion.

In the first commit I changed 
from:
```js
  const [state, dispatch] = useReducer(reducer, initialCount, init);
```
to:
```js
  const [state, dispatch] = useReducer(reducer, {count: initialCount}, init);
```
as the second argument was used as `initialState` in the previous examples.

In the updated pull request, the `useReducer` hook is no longer introduced with the signature:
```js
const [state, dispatch] = useReducer(reducer, initialArg, init);
```
but with the simpler form used in the first example:
```js
const [state, dispatch] = useReducer(reducer, initialState);
```
The lazy initialization signature is instead moved down under the lazy initialization section where the connection between `init` and `initialArg` is first explained and exemplified.

### DIFF
--- a/content/docs/hooks-reference.md
+++ b/content/docs/hooks-reference.md
@@ -288,7 +288,7 @@ function reducer(state, action) {
 }
 
 function Counter({initialCount}) {
-  const [state, dispatch] = useReducer(reducer, {count: initialCount}, init);
+  const [state, dispatch] = useReducer(reducer, initialCount, init);
   return (
     <>
       Count: {state.count}
--- a/content/docs/hooks-reference.md
+++ b/content/docs/hooks-reference.md
@@ -282,7 +282,7 @@ function reducer(state, action) {
 }
 
 function Counter({initialCount}) {
-  const [state, dispatch] = useReducer(reducer, initialCount, init);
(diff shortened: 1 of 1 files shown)


alexkrolick:
Is the second arg ignored when using the init function (3rd arg) as shown in the example?

danieledler:
Oh, thanks for the question, apparently I got a bit confused here. The second arg is actually not ignored but gets another role when the 3rd argument is used. Instead of being an `initialState`, it becomes the `initialArg` to the state initializer function if that is provided:
> The initial state will be set to `init(initialArg)`

I think this can be clarified in the beginning, as `useReducer` is introduced with the 3-parameter signature
```js
const [state, dispatch] = useReducer(reducer, initialArg, init);
```
but without any info about the connection between `initialArg` and `init`. Instead, the first examples uses the 2-parameter version `useReducer(reducer, initialState)`.

I can update my pull request with a suggestion on that.

PR #2054, opened 2019-06-04
Replaced nonexistent document.root

kyarik:
In the "Context" guide, `document.root` is specified as the second argument to `ReactDOM.render()`. However, `document.root` seems to be undefined and I couldn't find anything about it on the web. So, I thought this was an error and I replaced `document.root` with `document.getElementById('root')`, which is consistent with what other guides in the docs specify as the second argument to `ReactDOM.render()`.

### DIFF
--- a/examples/context/theme-detailed-app.js
+++ b/examples/context/theme-detailed-app.js
@@ -46,4 +46,4 @@ class App extends React.Component {
   }
 }
 
-ReactDOM.render(<App />, document.root);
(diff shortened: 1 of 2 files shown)


BartoszKlonowski:
Hello @kyarik, thank you for taking the time to contribute!
As of React 18 the `ReactDOM.render` is deprecated, so I don't see this PR to be "rebasable".
Let me close this as outdated.

PR #2059, opened 2019-06-06
Document setting state with callback

### LINKED ISSUE #2039: useState hook updating state using a callback is not documented
Currently the following behavior exists with the useState hook as an undocumented feature.

```
  it("calls function passed to useState", () => {
    let didCall = false;

    const Component = () => {
      const [state, setState] = React.useState(null);
      React.useLayoutEffect(() => {
        setState(() => { didCall = true; });
      }, []);
      return null;
    };

    usingMount(<Component />, wrapper => {
      expect(didCall).toBe(true);
    });
  });
```

The section dedicated to documenting updating state does not mention that if a function is passed to setState it will execute it and set the state to the result. https://reactjs.org/docs/hooks-state.html#updating-state

tomecho:
fixes #2039

### DIFF
--- a/content/docs/hooks-state.md
+++ b/content/docs/hooks-state.md
@@ -187,6 +187,14 @@ In a function, we already have `setCount` and `count` as variables so we don't n
   </button>
 ```
 
+We could also call setCount with a function which will be invoked on call.
+
+```js{1}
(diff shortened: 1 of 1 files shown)


tomecho:
Somebody else completed this work.

PR #2063, opened 2019-06-08
Updates hooks-state.md

SinaKhorami:
fix asynchronous state update based on [doc](https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous)


lex111:
@SinaKhorami in this case, this does not make much sense, since `props` is not used, right?

SinaKhorami:
> @SinaKhorami in this case, this does not make much sense, since `props` is not used, right?

Aw yes you are right. i didn't realize it in that time.

PR #2064, opened 2019-06-09
chore/and operator instead of ternary operator

jonny22094:


lex111:
@jonny22094 thank you!

PR #2067, opened 2019-06-11
Switch single-file-example to Babel 7

Daniel15:
Babel 6 is old; let's use Babel 7 here instead.


threepointone:
This has been updated to reference https://unpkg.com/@babel/standalone/babel.min.js instead. Thanks for the PR!

PR #2071, opened 2019-06-12
Add Kansas City React Meetup

jsmapr1:
Add Kansas City React Meetup to documentation


lex111:
@jsmapr1 nice, thank you!

PR #2074, opened 2019-06-14
Update Refs documentation

### LINKED ISSUE #2066: after introducing hooks this phrase is misleading
after introducing hooks this part is misleading, as converting into class is no longer the only option, that can be used if state needed for the component

> You should convert the component to a class if you need a ref to it, just like you do when you need lifecycle methods or state.

https://github.com/reactjs/reactjs.org/blob/aada3a308493614b7d5b4b438b5c345d7ecc6c53/content/docs/refs-and-the-dom.md#refs-and-function-components-refs-and-function-components

Ruffeng:
This PR fixes #2066. Its intention is to inform the user about the
`useRef()` hook, and not confuse the user thinking that you can just
`ref` with class components.

### DIFF
--- a/content/docs/refs-and-the-dom.md
+++ b/content/docs/refs-and-the-dom.md
@@ -140,7 +140,12 @@ class CustomTextInput extends React.Component {
 
 #### Refs and Function Components {#refs-and-function-components}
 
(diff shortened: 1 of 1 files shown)


joechesney:
LGTM

PR #2078, opened 2019-06-15
Add an example for using the useReducer hook without using a redux style

wilsonrocks:
reducer and action.

I'm not very experienced in pull requests and open source contribution, so very happy to take and apply feedback.

James

### DIFF
--- a/content/docs/hooks-reference.md
+++ b/content/docs/hooks-reference.md
@@ -297,6 +297,27 @@ function Counter({initialCount}) {
 }
 ```
 
+#### useReducer with any function
+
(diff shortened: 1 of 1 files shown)


alexkrolick:
I think the idiomatic way to do this is using a callback for setting state. Here's a PR to document it:

https://github.com/reactjs/reactjs.org/pull/2059/files

wilsonrocks:
Hmm, I have used that pattern too. My concern is that it puts logic in render() that doesn't need to be there. The point I was trying to make with this example is that you can use useReducer without the redux style function that takes an object with a 'type' property. Happy to come up with a different example if this isn't clear, but thought a typing logger was something that people might want to use - for searches etc.

wilsonrocks:
If both of these were merged, it would cover both cases, though?

wilsonrocks:
is there any progress on this?

wilsonrocks:
I'm guessing this is not accepted... should I close it? @reactjs-bot

PR #2079, opened 2019-06-16
* CodeExample component converted to stateless function

abidhkm:
- converted State component to stateless component


threepointone:
This seems unnecessary, but thank you for the effort!

PR #2084, opened 2019-06-17
Remove warning

BackEndTea:
This syntax is no longer experimental, so the warning can be removed.


alexkrolick:
It is a stage 3 proposal https://github.com/tc39/proposal-class-fields

BackEndTea:
My bad, i misunderstood what it was warning about 😅

PR #2085, opened 2019-06-18
Update tutorial.md

ashutoshhans611:
Removing the parentheses from 'this.props.onClick' when passed to square component otherwise it tries to call the function even before the click event


ashutoshhans611:
missed some code. The as is code is correct

PR #2090, opened 2019-06-20
handling-events.md

121371:
I have proposed the third method to get rid of calling the bind method.


121371:
looks like this is already covered which i didn't notice earlier. hence, closing the pull request.

PR #2091, opened 2019-06-20
animation plugin moved

himanshupathakpwd:
`react-addons-css-transition-group` moved to `react-transition-group`


threepointone:
Thank you!

PR #2092, opened 2019-06-21
Fix example about context

econavi:
Hello!
Sorry for my English.
I started learning the React and it seemed to me that I had found a mistake.
Sorry if this is not the case.
All the best.


econavi:
I was wrong
sorry

PR #2093, opened 2019-06-21
Add explicit function argument in useCallback

VitGottwald:


content/docs/hooks-reference.md
```diff
@@ -307,8 +307,8 @@ Note that React may still need to render that specific component again before ba
 
 ```js
 const memoizedCallback = useCallback(
-  () => {
-    doSomething(a, b);
+  (c) => {
```
alexkrolick:
```suggestion
// callbacks can take arguments
(c) => {
```

sophiebits:
I'm worried that the a, b, argA, argB is hard to follow since the variables don't have meaning. What do you think of doing something slightly "more realistic"?

```
const handleClick = useCallback(
  (event) => {
    event.preventDefault();
    console.log('Hello, ' + props.name);
  },
  [props.name],
);
```

I'm not attached to it being an event handler but easily couldn't think of anything clearer.

alexkrolick:
@sophiebits sounds good to me as long as it is clear it's not only for events and an event isn't somehow the default argument

sophiebits:
That's a good point. Could do something like this, though I fear it's harder to follow:
```js
const getFormattedAge = useCallback(
  (birthDate) => {
    const age = calculateAge(birthDate);
    return formatDuration(age, props.dateFormat);
  },
  [props.dateFormat],
);
// `getFormattedAge` is the same reference on every render
// until `props.dateFormat` changes
```

Other thoughts?

alexkrolick:
I think that last example would be great


@baremetalfreak what do you think? Change it to that?

baremetalfreak:
I actually like the former example better. It feels artificial to to have hook depend only on `props.dateFormat` and not on `birthDate`. It is easy to imagine `birthDate` comes as `prop.birthDate` and then we are showing an example that would not pass [eslint-plugin-react-hooks](https://www.npmjs.com/package/eslint-plugin-react-hooks).

I think the arguments of the function returned by `useCallback` are always going to be something that comes from the outside. Some kind of event from a user interaction (like clicking), a change in an external store (like redux, mobx, or graphql cache), or some other kind of event emitter.

I would slightly modify the example to actually use the variable:

```js
const handleClick = useCallback(
  (event) => {
    event.preventDefault();
    console.log(`Hello ${props.name}, you clicked`, event.target);
  },
  [props.name],
);
// `handleClick` is the same reference on every render
// until `props.name` changes
```

What do you think @sophiebits and @alexkrolick?

BartoszKlonowski:
Hello @VitGottwald,
Unfortunately the part of the docs you update in this PR no longer exists in the latest docs.
Let me close this PR as outdated, but you can create new PR for other open issues if you would still like to contribute.
Thank you!

PR #2094, opened 2019-06-22
Improve Web Components in React Section

rwieruch:
Hello there,

for a recent client of mine I had to dive into Web Components and how to use them in React. Their goal is to create a UI library, based on Web Standards, which shares a common structure, style  **and behavior**. This way they would enable all their developer teams to pick their frameworks of choice (React, Angular, ...) to wrap these common UI components. Long story short: I had to come up with a way to use Web Components in React. 

It's fairly straight forward, but I didn't find lots of documentation in the official React docs. Since I wrote about [Web Components](https://www.robinwieruch.de/web-components-tutorial/) and [how to use them in React](https://www.robinwieruch.de/react-web-components/) myself, I though it may be helpful for people who face the same problem. It comes also with a [custom React hook](https://github.com/the-road-to-learn-react/use-custom-element) which takes care about formatting information to JSON and attaching event listeners. I hope it's useful for the community 🚀 🙂

### DIFF
--- a/content/docs/web-components.md
+++ b/content/docs/web-components.md
@@ -12,20 +12,56 @@ Most people who use React don't use Web Components, but you may want to, especia
 
 ## Using Web Components in React {#using-web-components-in-react}
 
+Web components are just custom (HTML) elements which can be used directly in React:
+
 ```javascript
-class HelloMessage extends React.Component {
-  render() {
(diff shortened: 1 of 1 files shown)


rwieruch:
Closed in favor of https://github.com/reactjs/reactjs.org/pull/2631

PR #2096, opened 2019-06-23
Remove tooling support info in fragment docs

sibiraj-sr:
I think these lines can be safely removed now. Most tools now support the short syntax for react fragment.


lex111:
@sibiraj-sr thank you!

PR #2100, opened 2019-06-24
Correct the description of when gDSFP gets called

cyan33:
There's a bit of inconsistency between the doc of gDSFP. According to [static getDerivedStateFromProps](https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops):

> Note that this method is fired on every render, regardless of the cause. This is in contrast to UNSAFE_componentWillReceiveProps, which only fires when the parent causes a re-render and not as a result of a local setState.

But in the [blog](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html):

> That lifecycle is called when a component is created and each time it receives new props

Feel free to propose a better wording. :)

### DIFF
--- a/content/blog/2018-03-27-update-on-async-rendering.md
+++ b/content/blog/2018-03-27-update-on-async-rendering.md
@@ -134,7 +134,7 @@ Here is an example of a component that uses the legacy `componentWillReceiveProp
 
 Although the above code is not problematic in itself, the `componentWillReceiveProps` lifecycle is often mis-used in ways that _do_ present problems. Because of this, the method will be deprecated.
 
-As of version 16.3, the recommended way to update `state` in response to `props` changes is with the new `static getDerivedStateFromProps` lifecycle. (That lifecycle is called when a component is created and each time it receives new props):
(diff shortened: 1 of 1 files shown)


content/blog/2018-03-27-update-on-async-rendering.md
```diff
@@ -134,7 +134,7 @@ Here is an example of a component that uses the legacy `componentWillReceiveProp
 
 Although the above code is not problematic in itself, the `componentWillReceiveProps` lifecycle is often mis-used in ways that _do_ present problems. Because of this, the method will be deprecated.
 
-As of version 16.3, the recommended way to update `state` in response to `props` changes is with the new `static getDerivedStateFromProps` lifecycle. (That lifecycle is called when a component is created and each time it receives new props):
+As of version 16.3, the recommended way to update `state` in response to `props` changes is with the new `static getDerivedStateFromProps` lifecycle. It is called when a component is created and each time it re-renders (props or state changes):
```
gaearon: maybe "due to changes to props or state"

cyan33: Done! Thanks for the suggestion!

cyan33:
@gaearon Shall we get this merged? :)

lex111:
@cyan33 thank you!

PR #2101, opened 2019-06-25
Update docs re dev vs prod

VadimBrodsky:
The fact that error boundaries behave differently depending on the environment is surprising, this adds more information about the differences.

Found a nice explanation by @gaearon  about the differences here https://github.com/facebook/react/issues/12897#issuecomment-410036991


lex111:
@alexkrolick could you take a look here, please?

Issue #1101, opened 2018-07-30
Document how to update nested state properties


dandv:
In [setState](https://reactjs.org/docs/react-component.html#setstate), what should one return from the updater function in order to update a nested property of the state? This isn't documented, and it trips a **lot** of people:

https://stackoverflow.com/questions/18933985/this-setstate-isnt-merging-states-as-i-would-expect
https://stackoverflow.com/questions/43040721/how-to-update-a-nested-state-in-react
https://stackoverflow.com/questions/34071160/how-to-update-object-in-react-js
etc.

zacharymorel:
I second this, I would love to read some detailed docs from y'all.

cbdeveloper:
I investigated on this subject a little bit:
[https://stackoverflow.com/questions/54445040/whats-the-best-alternative-to-update-nested-react-state-property-with-setstate](https://stackoverflow.com/questions/54445040/whats-the-best-alternative-to-update-nested-react-state-property-with-setstate)

yvesgurcan:
Hey guys! I opened a PR to document how to handle nested state properties. First contribution to the React project :)

laozhizi:
@yvesgurcan where is the PR for handle nested state properties?

bk973:
It's better to write your updater function to return new state based on old state. If  you do this, then updating state for nested data is pretty straightforward.

silver-snoopy:
Has this issue been abandoned? I can pick it up and write the documentation based on the previous PR.

yvesgurcan:
Hey @silver-snoopy, you're welcome to work based on the previous PR!

Kanha-13:
Hey I wanted to contribute to this. As I have worked on nested state update. Can anyone help me how to contribute to this I am new to the opensource. Thank you

tbroadley:
https://github.com/reactjs/reactjs.org/pull/4257#issuecomment-1025374675

> This has been addressed in the newer version of docmentation site.
>
> link: https://beta.reactjs.org/learn/updating-objects-in-state#updating-a-nested-object.

I'd suggest closing this issue and marking it as one that won't be completed.

Leddfoot:
> [#4257 (comment)](https://github.com/reactjs/reactjs.org/pull/4257#issuecomment-1025374675)
> 
> > This has been addressed in the newer version of docmentation site.
> > link: https://beta.reactjs.org/learn/updating-objects-in-state#updating-a-nested-object.
> 
[...]

PR #2108, opened 2019-06-29
Add nested state property FAQ

yvesgurcan:
Fixes issue #1101 by adding a simple example on how to updated a nested property within the state of a component thanks to the spread operator.

### DIFF
--- a/content/docs/faq-state.md
+++ b/content/docs/faq-state.md
@@ -92,7 +92,7 @@ state = {
 }
 ```
 
-If you want to update `user.lastname` without replacing the whole `user` object in the state, use the [spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) in the update function:
+If you want to update `user.lastname` while keeping the other properties the same, use the [spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) in the update function to copy the original keys to a new object:
 
 ```jsx{4-7}
 changeLastName() {
--- a/content/docs/faq-state.md
+++ b/content/docs/faq-state.md
@@ -94,7 +94,7 @@ state = {
 
 If you want to update `user.lastname` without replacing the whole `user` object in the state, use the [spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) in the update function:
 
-```js{4-7}
+```jsx{4-7}
 changeLastName() {
   this.setState((state) => {
     return {
--- a/content/docs/faq-state.md
+++ b/content/docs/faq-state.md
@@ -78,6 +78,39 @@ handleSomething() {
 
 [Learn more about setState](/docs/react-component.html#setstate)
 
+### How do I update nested state properties?
+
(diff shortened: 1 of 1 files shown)


content/docs/faq-state.md
```diff
@@ -78,6 +78,39 @@ handleSomething() {
...
+state = {
+    user: {
+        firstname: 'Bob',
+        lastname: 'Dylan',
+        age: 31
+    }
+}
+```
+
+If you want to update `user.lastname` without replacing the whole `user` object in the state, use the [spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) in the update function:
```
alexkrolick: To be clear, this does replace the user object.

```suggestion
If you want to update `user.lastname` while keeping the other properties the same, use the [spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) in the update function to copy the original keys to a new object:
```

yvesgurcan: Very good point! Thank you for the suggestion.

content/docs/faq-state.md
```diff
@@ -78,6 +78,39 @@ handleSomething() {
...
+
+If you want to update `user.lastname` while keeping the other properties the same, use the [spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) in the update function to copy the original keys to a new object:
+
+```jsx{4-7}
+changeLastName() {
+  this.setState((state) => {
+    return {
+        user: {
+            ...state.user,
+            lastName: 'Doe'
```
emileber: There's a typo here, where you update `user` with `lastName`, but initially, it was `lastname`.

content/docs/faq-state.md
```diff
@@ -78,6 +78,39 @@ handleSomething() {
...
+        lastname: 'Dylan',
+        age: 31
+    }
+}
+```
+
+If you want to update `user.lastname` while keeping the other properties the same, use the [spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) in the update function to copy the original keys to a new object:
+
+```jsx{4-7}
+changeLastName() {
```
emileber: To be closer to a real use-case (which would be less confusing), `changeLastName()` should take a new value as a parameter.

    changeLastName(newLastname) {

content/docs/faq-state.md
```diff
@@ -78,6 +78,39 @@ handleSomething() {
...
+state = {
+    user: {
+        firstname: 'Bob',
+        lastname: 'Dylan',
+        age: 31
+    }
+}
+```
+
+If you want to update `user.lastname` while keeping the other properties the same, use the [spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) in the update function to copy the original keys to a new object:
```
emileber: There's no _spread operator_, it's the _spread **syntax**_ which you're correctly linking to already.

BartoszKlonowski:
Closing as per discussion in the issue.
Thank you!

PR #2113, opened 2019-07-01
Update hooks-faq.md

n3dst4:
I tripped up slightly while reading this example for using the callback form of a state setter inside an effect. I've added a few lines that might help a hook newbie grok the differences between the examples.


n3dst4:
Sorry, copyediting myself here...

alexkrolick:
Thanks that is helpful

PR #2114, opened 2019-07-01
Updated error-boundaries.md

siddrc:
Just added a one liner, to be more verbose.

### DIFF
--- a/content/docs/error-boundaries.md
+++ b/content/docs/error-boundaries.md
@@ -157,7 +157,7 @@ class MyComponent extends React.Component {
 }
 ```
 
-Note that the above example is demonstrating regular JavaScript behavior and doesn't use error boundaries.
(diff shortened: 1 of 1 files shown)


content/docs/error-boundaries.md
```diff
@@ -157,7 +157,7 @@ class MyComponent extends React.Component {
 }
 ```
 
-Note that the above example is demonstrating regular JavaScript behavior and doesn't use error boundaries.
+Note that the above example is demonstrating regular JavaScript behavior and doesn't use error boundaries. However, throwing an error like ```js throw new Error(`ERROR:${this.state.error}`)``` will trigger the surrounding Error Boundary.
```
ahtee: possibly say:

" However, throwing the following error will trigger the surrounding Error Boundary.
```js 
throw new Error(`ERROR:${this.state.error}`)
```
"

BartoszKlonowski:
Hello @siddrc!
As the sentence you modify no longer exists in the latest of react.dev, I will close this PR as outdated.
Thank you for taking time to improve the docs!

PR #2116, opened 2019-07-02
Update in the Avatar component

SqBrain:
It seems wrong to have both props and user in it. In the next step we already pass in the component props.author, which have name and avatarUrl attributes. So the user must be instead of props, isn't it?


alexkrolick:
Nope, all attributes in JSX become named keys in the props object.

You could destructure like this though:

```jsx
function Avatar({user}) {
// ...
}
```

PR #2129, opened 2019-07-05
Updated line number-73

Sagar-Vinayak:
Updated the content for better readability.


ahtee:
pitfall is a better word for this, in my opinion

lex111:
> pitfall is a better word for this, in my opinion

Agree! :+1:

PR #2135, opened 2019-07-09
fix outline in sideBar button

israelKusayev:
fix outline button in small devices.

image


ahtee:
What browser and device are you using? I am not seeing an outline on the live version of the site on safari or chrome on iOS.

israelKusayev:
I'm using chrome
And I have galaxy s10

israelKusayev:
Any comments?

ahtee:
If im not mistaken, the outline offered by the browser is there for accessibility purposes. 
More reading: https://a11yproject.com/posts/never-remove-css-outlines/

alexkrolick:
Correct, it should not be removed

PR #2137, opened 2019-07-10
chore: upgrade gatsby to latest version

wardpeet:
I've noticed reactjs.org is still running on gatsby .... Gatsby is currently on version 2.18.20. So it's time to upgrade.

I might upgrade the plugins in another PR


BartoszKlonowski:
Hello @wardpeet, thanks for taking the time to contribute!
The gatsby is no longer in the packages used by these docs, so I will close this PR as no longer needed.

PR #2144, opened 2019-07-13
Added option for more cdns.

praisedpk:
unpkg CDN is used throughout the react docs. The CDN Links page on documentation again uses the same unpkg code without providing option for other CDNs.

Added 2 new CDNs. CDNJS is **insanely popular**, while PageCDN **compresses JS/CSS more tightly**. Both can be a good addition to this page.

### DIFF
--- a/content/docs/cdn-links.md
+++ b/content/docs/cdn-links.md
@@ -22,6 +22,8 @@ The versions above are only meant for development, and are not suitable for prod
 
 To load a specific version of `react` and `react-dom`, replace `16` with the version number.
 
+Both React and ReactDOM are available on PageCDN ([React](https://pagecdn.com/lib/react), [ReactDOM](https://pagecdn.com/lib/react-dom)) and CDNJS ([React](https://cdnjs.com/libraries/react), [ReactDOM](https://cdnjs.com/libraries/react-dom)).
+
 ### Why the `crossorigin` Attribute? {#why-the-crossorigin-attribute}
 
 If you serve React from a CDN, we recommend to keep the [`crossorigin`](https://developer.mozilla.org/en-US/docs/Web/HTML/CORS_settings_attributes) attribute set:


lex111:
@praisedpk thank you!

MartinKolarik:
@lex111 I'd like to point out that while CDNJS is definitely a popular and trustworthy option which has been around for a long time, [PageCDN](https://pagecdn.com/) is a completely new, unknown service, which has just launched [this week](https://twitter.com/pagecdn/status/1146407076698177536). It looks like they are sending PRs like this to the most popular projects just to promote their business.

praisedpk:
@lex111 thanks.

@MartinKolarik please see to the PR comment above.

jimaek:
The only under-performance point here is regarding the compression. That not the main performance metric of a CDN. For example https://www.cdnperf.com/

praisedpk:
@jimaek Compression **IS** the main performance metric, but its not the only one.

1. I explained the compression in another thread with font-awesome example. In some other cases, we observed reduced file size of upto 29% for compressible resources compared to other compression formats.
````
 331.12 KB    https://pagecdn.io/lib/font-awesome/5.10.0-11/js/all.min.js
 385.72 KB    https://cdnjs.cloudflare.com/ajax/libs/font-awesome/5.10.0-11/js/all.min.js
 459.39 KB    https://use.fontawesome.com/releases/v5.10.0-11/js/all.js (named all.js on this CDN)
1105.92 KB    Raw all.min.js file
````
You can see the **difference of upto 128 KB** between different compression levels. 128 KB can influence the decision of many users when selecting a CDN. And that's why I submitted PageCDN to react's docs just to give users more options to decide from. (Sure, react javascript files would not get the same level of savings from aggressive compression.)

2. Immutable Caching. Generally, 1 year expiry is sufficient for resources on Public CDN, but the problem is that browsers keep issuing conditional requests that is also an overhead. In my tests, none of the popular CDNs set the immutable caching header.

3. Server Push is a feature that usually does not related to most of the libraries, but there are some libraries that can get benefit from it. We are planning to use HTTP/2 server push with some of the libraries, and most of the themes that we host through public CDN. We are collecting data to assist our research on this. We believe this can make noticeable difference in performance of PageCDN soon.

4. HTTP/2 multiplexing. Today almost all popular CDNs (except for `code.jquery.com`) use HTTP/2. But the problem is, loading just few resources over a connection is not a good use of multiplexing that comes with HTTP/2. For example, if you look at the source code of http://www.piston.rs/dyon-tutorial/ (an example site), you will see Google Fonts CDN, MathJax CDN, BootstrapCDN, jQuery CDN, etc used on this single page. Almost all of these are HTTP/2, but they do not share connection. Plus, there are some CSS, JS and image files that load over the primary host. PageCDN aims to collect every such resource over a single host `pagecdn.io` so that whatever you include on your site, gets delivered in parallel... including your private files too.

5. WordPress optimization. WordPress is an area where we see more room for optimizations compared to other websites. Our WordPress plugin beautifully mixes public and private cdn and provides option for minification, fonts loading, etc too. Plus there are some other optimizations that we plan to introduce in future that I cannot talk about before time.

6. We are working on some other areas of website performance optimization that are never discussed in the past. I cannot provide more details beforehand.

Overall, we are using existing technologies and want to make the optimal use of it to fuel our research based on the data we collect. Yes, PageCDN is our business, but we plan to return a lot back to the community. Its not that we want to take advantage of popular libraries. The benefit is two ways. A lot will be returned back to the community.

@jimaek I am following you for last so many years and learnt a lot reading your comments on github. I admit that following you and jsDelivr helped me a lot in my research. If you have any feedback that can help us improve, we can discuss here or on twitter.

Thanks.

praisedpk:
cdnperf.com is a good resource. Thanks.

alexkrolick:
Last time a new service (Codesandbox) came out that someone wanted to add to the docs, we waited several months to allow it to mature. Let's stick with that precedent.

lex111:
@alexkrolick I'm so sorry. It's my fault. That's good to know...

praisedpk:
Hi @lex111 @alexkrolick 

Maybe its time to merge this PR again that was reverted by #2146

PR #2145, opened 2019-07-13
Update docs about an existence of .elementType

lonyele:
Hi, I've met the same issue on [here](https://github.com/facebook/prop-types/issues/200) at storybook and found the following [PR](https://github.com/facebook/prop-types/pull/211) that adds the `elementType` feature. It could find the doc on [npm](https://www.npmjs.com/package/prop-types), but not the official react site.


lex111:
@lonyele thank you!

PR #2157, opened 2019-07-20
update the status of Arabic translation ..

3imed-jaberi:
arabic-status-prove


3imed-jaberi:
Please @tesseralis  , check this ..

PR #2165, opened 2019-07-23
Testing Docs

threepointone:
This expands the testing docs for reactjs.org. It adds 3 main docs -

- testing.md: An entry point to the testing docs
- testing-recipes.md: Common patterns when writing tests for React Components.
- testing-environments.md: Setting up your testing environment.

https://deploy-preview-2165--reactjs.netlify.com/docs/testing.html
https://deploy-preview-2165--reactjs.netlify.com/docs/testing-recipes.html
https://deploy-preview-2165--reactjs.netlify.com/docs/testing-environments.html


gaearon:
Can you add them temporarily to sidebar *somewhere* so that it’s easy to preview what you already posted on a live site?

content/docs/testing-recipes.md
```diff
@@ -0,0 +1,358 @@
...
+    render(<Hello name="Sophie" />, container);
+  });
+  expect(container.textContent).toBe("Hello Sophie!");
+
+  act(() => {
+    render(<Hello name="Flarnie" />, container);
+  });
+  expect(container.textContent).toBe("Hello Flarnie!");
+  // cleanup on exiting
+  container.remove();
```
gaearon: This seems like an anti pattern since it won’t run on failure.

threepointone:
It's now under Advanced Guides > Testing

content/docs/testing-recipes.md
threepointone: added setup/teardown for all the tests

content/docs/testing-recipes.md
```diff
@@ -0,0 +1,403 @@
...
+  act(() => {
+    render(<Hello name="Flarnie" />, container);
+  });
+  expect(container.textContent).toBe("Hello Flarnie!");
+});
+```
+
+### Data fetching {#data-fetching}
+
+Mocking data fetching with 'fake' data prevents flaky tests, and makes them run faster.
```
alexkrolick:
```suggestion
Instead of calling real APIs in all your tests, you can mock requests with dummy data. Mocking data fetching with 'fake' data prevents flaky tests due to an unavailable backend, and makes them run faster. Note: you may still want to run a subset of tests using an "end-to-end" framework that tells whether the whole app is working together.
```

content/docs/testing-recipes.md
```diff
@@ -0,0 +1,403 @@
...
+      .dispatchEvent(new MouseEvent("click", { bubbles: true }));
+  });
+
+  expect(yields).toBe([2]);
+});
+```
+
+### Snapshot testing {#snapshot-testing}
+
+Frameworks like Jest also let you save 'snapshots' of data. Combined with react-test-renderer, we can save component structures, and verify they don't break in the future. It's typically better to make more specific assertions than to use snapshots (or at least ensure that the snapshot is small) because these kinds of tests include implementation details meaning they break easily and teams can get desensitized to snapshot breakages.
```
alexkrolick: Snapshots also work with regular DOM nodes and JSON data structures. Might be worth showing an example of that and maybe mention how inline snapshots can reduce "desensitization" to snapshot changes.

content/docs/act.md
```diff
@@ -0,0 +1,105 @@
...
+You could use this in tests:
+
+```jsx
+// ...
+click(element);
+// then assert on the side effects of the click
+// ...
+```
+
+However, if we need to batch multiple clicks as one 'unit', we can wrap them with another `act()` call.
```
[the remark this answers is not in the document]
threepointone: Noted, I'll think about this a bit more. Not sure if it'll be in the PR.

content/docs/testing-environments.md
```diff
@@ -0,0 +1,61 @@
...
+
+- If you're writing a library that tests mostly browser-specific behaviour, and requires 'native' browser behaviour like layout, you could use a framework like [mocha](https://mochajs.org/) in combination with [Karma](https://karma-runner.github.io) to run your tests in a browser.[<sup>(example)</sup>](packages/mocha/README.md)
+
+- In an environment where you _can't_ simulate a DOM (eg: testing React Native components on Node.js), you could use [event simulation helpers](https://reactjs.org/docs/test-utils.html#simulate) to simulate interactions with elements. Alternatively, you could use the `fireEvent` helper from [`@testing-library/react-native`](https://testing-library.com/docs/native-testing-library).
+
+- Frameworks like [Cypress](https://www.cypress.io/), [puppeteer](https://github.com/GoogleChrome/puppeteer) and [webdriver](https://www.seleniumhq.org/projects/webdriver/) are useful for running [end-to-end tests](#end-to-end-tests-aka-e2e-tests) (puppeteer and webdriver both run in a separate process from the actual code being tested).
```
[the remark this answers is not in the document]
threepointone: removing all the bits about karma, and I'll leave just this part here.

content/docs/testing-recipes.md
```diff
@@ -0,0 +1,403 @@
...
+We can write a test for this component:
+
+```jsx
+// hello.test.js
+import Hello from "./hello";
+import { render } from "react-dom";
+import { act } from "react-dom/test-utils";
+
+let container = null
+beforeEach(() => {
```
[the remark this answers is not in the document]
threepointone: good idea.

content/docs/testing-environments.md
```diff
@@ -0,0 +1,57 @@
...
+permalink: docs/testing-environments.html
+---
+
+<!-- This document is intended for folks who are comfortable with Javascript, and have probably written tests with it. It acts as a reference for the differences in testing environments for React components, and how those differences affect the tests that they write. This document also assumes a slant towards web-based react-dom components, but has notes for other renderers. -->
+
+Testing environments can affect the way you write and run tests for React Components. 
+
+This document goes through the major factors that can affect your environment, the implications of those decisions, and recommendations for some scenarios.
+
+### Test frameworks {#test-frameworks}
```
gaearon: "Test runners"

content/docs/testing-environments.md
```diff
@@ -0,0 +1,57 @@
...
+
+Test frameworks like [Jest](https://jestjs.io/), [mocha](https://mochajs.org/), [ava](https://github.com/avajs/ava) let you write test suites as regular Javascript, and run them as part of your development process. Additionally, test suites are run as part of workflows like builds, CI/CD, etc.
+
```
gaearon: It's probably worth calling out here that some setups like Create React App already include a test runner. https://facebook.github.io/create-react-app/docs/running-tests

Maybe even before this section. We should avoid the situation where people spend an hour configuring Jest in a CRA project (which happened before).

content/docs/nav.yml
```diff
@@ -73,6 +73,8 @@
       title: Static Type Checking
     - id: strict-mode
       title: Strict Mode
+    - id: testing
```
gaearon: Lets make it top level with all relevant pages.

content/docs/testing-recipes.md
```diff
@@ -0,0 +1,403 @@
+---
+id: testing-recipes
+title: Test recipes 
+permalink: docs/testing-recipes.html
+---
+
+This document lists some common testing patterns for React components.
+
+- This assumes tests are running in jest. For details on setting up a testing environment, read [the Environments doc](/docs/testing-environments.html).
```
gaearon: I'd change to something like

>This page assumes you're using Jest as a test runner. If you use a different test runner, you may need to adjust the API, but the overall shape of the solution will likely be the same.

content/docs/testing-recipes.md
```diff
@@ -0,0 +1,403 @@
...
+  expect(button.innerText).toBe("Turn off!");
+  act(() => {
+    for (let i = 0; i < 5; i++) {
+      button.dispatchEvent(new MouseEvent("click", { bubbles: true }));
+    }
+  });
+  expect(clicks).toBe(6);
+  expect(button.innerText).toBe("Turn on!");
+});
+```
```
gaearon: Call out here that it *needs to be attached to document* for events to fire. This is easy to miss.

content/docs/testing-recipes.md
```diff
@@ -0,0 +1,403 @@
...
+  return root.toJSON();
+}
+
+it("should preserve it's structure", () => {
+  expect(toJSON(<Hello />)).toMatchSnapshot();
+  expect(toJSON(<Hello name="Sophie" />)).toMatchSnapshot();
+  expect(toJSON(<Hello name="Flarnie" />)).toMatchSnapshot();
+});
+```
+
```
gaearon: Maybe call out mocking a dependency as a common solution to avoid snapshots growing too much.

content/docs/testing-recipes.md
```diff
@@ -0,0 +1,403 @@
...
+      .dispatchEvent(new MouseEvent("click", { bubbles: true }));
+  });
+
+  expect(yields).toBe([2]);
+});
+```
+
+### Snapshot testing {#snapshot-testing}
+
+Frameworks like Jest also let you save 'snapshots' of data. Combined with react-test-renderer, we can save component structures, and verify they don't break in the future. It's typically better to make more specific assertions than to use snapshots (or at least ensure that the snapshot is small) because these kinds of tests include implementation details meaning they break easily and teams can get desensitized to snapshot breakages.
```
gaearon: >It's typically better to make more specific assertions than to use snapshots

This is a very strong recommendation. In practice snapshots might be better overall than over-specific tests that make *too* "specific" assertions. Can we relax this wording a little bit? Try to clarify pitfalls and guiding principles rather than make value judgements ("better").

content/docs/testing-recipes.md
```diff
@@ -0,0 +1,403 @@
...
+})
+
+function render(ui){
+  act(() => {
+    ReactDOM.render(ui, container);
+  });
+}
+
+it("should select null after timing out", () => {
+  const yields = [];
```
gaearon: I think most people will be confused by the use of the word "yields" here

content/docs/testing-environments.md
```diff
@@ -0,0 +1,57 @@
...
+- Jest is widely compatible with React projects, supporting features like mocked [modules](#mocking-modules) and [timers](#mocking-timers), [jsdom](#mocking-a-rendering-surface) support, etc.
+- Libraries like [mocha](https://mochajs.org/) work well in real browser environments, and could help for tests that explicitly need it.
+- End-to-End tests are used for testing longer flows across multiple pages, and require a [different setup](#end-to-end-tests-aka-e2e-tests)
+
```
gaearon: Maybe add a sentence that suggests how much to test UI components, and when we recommend to use unit / e2e testing.

content/docs/testing-environments.md
```diff
@@ -0,0 +1,57 @@
...
+- [`jsdom`](https://github.com/jsdom/jsdom) is a lightweight browser implementation that's _mostly_ spec compliant, but runs inside Node.js. It behaves like a regular browser would, but doesn't have features like [layout and navigation](https://github.com/jsdom/jsdom#unimplemented-parts-of-the-web-platform). This is still useful for most web-based component tests, since it runs quicker than having to start up a browser for each test. It also runs in the same process as your tests, so you can write code to examine and assert on the rendered DOM.
+
+- Just like in a real browser, `jsdom` lets us model user interactions; tests can dispatch 'real' events on DOM nodes, and then observe and assert on the side effects of these actions on a rendered surface.[<sup>(example)</sup>](/docs/testing-recipes.html#events)
```
gaearon: These sup tags are pretty hard to read in the rendered version. They feel very noisy. Can we restructure this so we won't need them?

content/docs/testing-environments.md
```diff
@@ -0,0 +1,57 @@
...
+
+- Jest is widely compatible with React projects, supporting features like mocked [modules](#mocking-modules) and [timers](#mocking-timers), [jsdom](#mocking-a-rendering-surface) support, etc.
+- Libraries like [mocha](https://mochajs.org/) work well in real browser environments, and could help for tests that explicitly need it.
+- End-to-End tests are used for testing longer flows across multiple pages, and require a [different setup](#end-to-end-tests-aka-e2e-tests)
+
+### Mocking a rendering surface {#mocking-a-rendering-surface}
+
+Tests could run in an environment without access to a 'real' rendering surface. (eg: React Native or ReactDOM based components could be tested on Node.js) Depending on the type of test, you could simulate a browser-like DOM with something like [jsdom](https://github.com/jsdom/jsdom), or use [`react-test-renderer`](https://reactjs.org/docs/test-renderer.html) to render your components.
```
gaearon: If I were new to DOM testing, and never heard of RN, I'd get confused in this section. Can we rephrase this with a more clear recommendation path? Most people reading this probably aren't investigating RN. "something like" doesn't inspire confidence, we should use stronger wording here.

content/docs/testing-environments.md
```diff
@@ -0,0 +1,57 @@
+---
+id: testing-environments
+title: Testing environments
+permalink: docs/testing-environments.html
+---
+
+<!-- This document is intended for folks who are comfortable with Javascript, and have probably written tests with it. It acts as a reference for the differences in testing environments for React components, and how those differences affect the tests that they write. This document also assumes a slant towards web-based react-dom components, but has notes for other renderers. -->
+
+Testing environments can affect the way you write and run tests for React Components.
```
gaearon: This sounds a bit like a vague truism. "Can affect", so..?

Maybe could be a bit more concrete about what people will get out of this document.

content/docs/act.md
```diff
@@ -0,0 +1,109 @@
+---
+id: act
+title: act
+permalink: docs/act.html
+---
+
+`act()` provides an abstraction to make sure what's rendered reflects a state where pending actions have been 'committed' and the DOM has been updated. When writing UI tests, tasks like rendering, user events, or data fetching can be considered as 'units' of interaction with a user interface. `act` allows you be sure all effects related to a unit have been processed before asserting on the result.
```
gaearon: Can we avoid jargon like "committed", "units", "processing"? This is a bit heavy. There's probably a way to rephrase this with a focus on use cases.

content/docs/act.md
```diff
@@ -0,0 +1,109 @@
...
+```
+
+### Guarantees {#guarantees}
+
+`act()` provides some guarantees when it's run.
+
+- It batches updates triggered inside its scope, and applies them all in one go.
+- It flushes any queued effects and updates.
+- When nested, the work is flushed only on exiting the outermost `act()`.
+- The asynchronous version of `act()` also flushes any resolved promises.
```
gaearon: This implies the user knows what "batches", "flushes", "effects" mean. I don't think we introduce all of these terms explicitly. Can probably rewrite this in a more concrete way.

content/docs/act.md
```diff
@@ -0,0 +1,109 @@
...
+If you are seeing these warnings, it's possible that your test will is making assertions on an unreliable state of the UI, before it's had a chance to stabilise. Wrapping the test code that triggers these conditions with `act()` will avoid this problem and silence the warnings.
+
+### Asynchronous Testing {#asynchronous-testing}
+
+When working with asynchronous code, an async version of `act()` can be used to flush resolved promises and write assertions on asynchronous updates.
+
+```jsx
+import {act} from 'react-dom/test-utils'
+import App from './app'
+it('flushes microtasks', () => {
```
gaearon: this sounds like you're encouraging people to write a test verifying that it "flushes microtasks"

content/docs/act.md
```diff
@@ -0,0 +1,109 @@
...
+import { act as testAct, create } from "react-test-renderer";
+// ...
+let root;
+domAct(() => {
+  testAct(() => {
+    root = create(<App />);
+  });
+});
+expect(root).toMatchSnapshot();
+```
```
gaearon: Interestingly most of the things I wanted to learn about `act` aren't on this page. They're on Recipes. If you don't want to unify them, maybe let's have a section at the end that crosslinks specific scenarios to Recipes? So people can find them.

content/docs/testing-recipes.md
```diff
@@ -0,0 +1,403 @@
...
+  useEffect(
+    props => {
+      const timeoutID = setTimeout(() => {
+        props.onSelect(null);
+      }, 5000);
+      return () => {
+        clearTimeout(timeoutID);
+      };
+    },
+    [props.onSelect]
```
gaearon: This doc uses Hooks for everything. None of other docs outside Hooks section do. Will this give an impression that these examples don't work for classes? Maybe we should have a mix here.

content/docs/act.md
Eddie-CooRo: Also, I think you forgot to make the `it`'s callback function `async`.

content/docs/testing-recipes.md
kentcdodds: Instead of a mix, a single sentence that says: "Note: These testing strategies work just as well for class components as they do for function components."

content/docs/testing-recipes.md
kentcdodds: Agreed. I think it'd be better to use a mock function instead.

content/docs/testing-recipes.md
kentcdodds: I can't tell you how many teams I've talked with who regret diving into snapshot testing. It's been a lot of teams.

I'm not opposed to changing the wording up a bit to be less strong I guess, but I think we should try to encourage effective use of snapshots and warn of the dangers of over-using snapshots.

I agree with @alexkrolick, using an inline snapshot in this example could go a long way toward helping people avoid the pitfalls naturally.

threepointone: I'll make this a *little* less sever, but I've also been thinking of recommending `toMatchInlineSnapshot`
- it's inline, so plainly visible when reading the test
- if it grows big, it's a sign that the test is losing it's efficacy, and will prompt a more focused test.

content/docs/testing-recipes.md
threepointone: I don't think dependencies should be mocked to prevent big snapshots. It could subtly break the code (eg - an effect in a child that called a callback would not be called any more, changing the observed result)

threepointone: ie - dependencies should only be mocked if there's a clear reason they don't belong in the test: they render a canvas, or can't run in test environments, etc. This is covered by the mocking-modules section.

content/docs/testing-environments.md
threepointone: We don't really recommend 'unit' testing in the traditional sense anyway. On L29 I mention how most UI tests are done with a jest/jsdom setup.

content/docs/testing-environments.md
threepointone: Yeah I dislike them too. Will rewrite.

content/docs/testing-environments.md
threepointone: Doesn't the next line cover that?

content/docs/act.md
threepointone: This is now merged with the recipes doc.

PR #2170, opened 2019-07-25
Add a relevant FAQ link in "Thinking In React"

colinmorris:
I think this FAQ entry is more directly relevant to the question of props vs. state. The linked "State and Lifecycle" section (I updated the url from `/docs/interactivity-and-dynamic-uis.html`, which is a redirect, I assume from a previous name) has one sentence comparing them (*State is similar to props, but it is private and fully controlled by the component.*), but it's tucked away and easy to miss.


alexkrolick:
👍 thanks

PR #2173, opened 2019-07-26
Fix false description about how setState works.

Rick-Kirkham:
The sentence that I'm replacing here says that `this.setState` merges new values into an existing state object rather than replace the object. This blatantly contradicts the explanations elsewhere in this documentation of how `setState` works. React treats the state object as **immutable**.  Accordingly, `setState` does _not_ change it. Instead, it creates a whole new object that duplicates the original state except that the properties of the object that are passed to `setState` are used to set the corresponding properties of the new state object. So, it _replaces_ the state object, the exact opposite of what the sentence that I want to fix is claiming.

### DIFF
--- a/content/docs/hooks-faq.md
+++ b/content/docs/hooks-faq.md
@@ -276,7 +276,7 @@ Now let's say we want to write some logic that changes `left` and `top` when the
   // ...
 ```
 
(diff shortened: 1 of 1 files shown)


BartoszKlonowski:
Hello @Rick-Kirkham, thank you for contributing!
Unfortunately the sentence you modify in this PR no longer exists in the docs, so I will close this PR as outdated.
But we welcome opening PRs for other open issues.
Thank you!

PR #2176, opened 2019-07-27
Profiler API docs

bvaughn:
Includes a 16.9 release notes section **with an arbitrary/random date** mentioning the new `Profiler` API and a reference page with examples and docs.

* [Release blog section](https://deploy-preview-2176--reactjs.netlify.com/blog/2019/08/01/react-v-16-9.html)
* [Profiler reference docs](https://deploy-preview-2176--reactjs.netlify.com/docs/profiler.html)

### Open Questions
* I didn't go into as much depth as I could (or as I did in [the RFC](https://github.com/reactjs/rfcs/blob/master/text/0051-profiler.md) for example) because I feel like the profiling story is currently incomplete until I've been able to tackle the root events API. Are these docs okay for now or are they too shallow?
* I just linked to the fb.me URL for interactions, since that API is currently unstable.
* I linked to fb.me/react-profiling for instructions on enabling the production+profiling bundle. Should I inline these instructions into the reference docs somewhere instead?


package.json
```diff
@@ -86,7 +86,8 @@
     "nit:examples": "prettier --config examples/.prettierrc --list-different \"examples/**/*.js\"",
     "prettier": "yarn format:source && yarn format:examples",
     "prettier:diff": "yarn nit:source && yarn nit:examples",
-    "reset": "rimraf ./.cache"
+    "reset": "rimraf ./.cache",
+    "start": "yarn dev"
```
gaearon: lol

bvaughn: lol you caught me 😉  this always trips me up

gaearon: Sme

PR #2177, opened 2019-07-29
Added Render-Atlanta to list of conferences

js1599:
The first React focused conference around culture & inclusion will be held in the peach state, May 4 - 6, 2020. Full website is going live August 1, but signup page is currently active via our URL


js1599:
CLA signed

gaearon:
Thanks. I moved it to match the chronological order for upcoming confs.

PR #2182, opened 2019-07-31
Unmount in tests to ensure Portals are cleaned up

ssorallen:
When following the example, my tests never cleaned up Portals that were
created by my components. To make sure that problem is bypassed all
together, make sure the component is unmounted via
`ReactDOM.unmountComponentAtNode` before removing the container from
the DOM.

### DIFF
--- a/content/docs/addons-test-utils.md
+++ b/content/docs/addons-test-utils.md
@@ -99,6 +99,9 @@ beforeEach(() => {
 });
 
 afterEach(() => {
+  act(() => {
+    ReactDOM.unmountComponentAtNode(container);
+  });
   document.body.removeChild(container);
   container = null;
 });


alexkrolick:
@gaearon @threepointone this is something to consider in the WIP testing docs. For what it's worth, it's also what the [cleanup function](https://github.com/testing-library/react-testing-library/blob/929748f092013b4045f65866edea5df9680f1f3a/src/index.js#L91-L99) from React Testing Library does.

ssorallen:
I wasn't sure whether the wrapping `act` was required or that was only needed inside tests; my tests pass whether the wrapping `act` is there or not. I can remove that if it's not needed here.

ssorallen:
This was addressed at some point and is now called out in [Setup/Teardown](https://reactjs.org/docs/testing-recipes.html#setup--teardown) in the testing docs. Closing this one out.

PR #2188, opened 2019-08-02
Update examples.md

piyushgarg195:


gaearon:
Thanks but this doesn't actually link to the source code.

PR #2192, opened 2019-08-04
Fix 'Going Big with React' part in community/videos.md ..

3imed-jaberi:


3imed-jaberi:
@tesseralis , please check this ...

PR #2201, opened 2019-08-06
Update an example to use a more accurate abbreviation.

nirvdrum:
"I.e." is short for "id est", which means "that is". It should be used to clarify something, perhaps by rephrasing. Unfortunately, here, it serves to limit the what it means for the header to grow complex. Using "e.g." makes it clear that the hypothetical suggestion is merely an example of how the header may grow to be more complex.


alexkrolick:
Indeed

PR #2202, opened 2019-08-07
Links to React Redux hooks api

lkostrowski:
Since Hooks are already released for Redux, link the docs


lex111:
@lkostrowski thank you!

PR #2207, opened 2019-08-08
Added missing function bracket in hooks-effect.md 

h-dong:
Added in a missing function bracket in one of the code snippet.


lex111:
@h-dong thank you!

PR #2226, opened 2019-08-10
Update testing-recipes.md fixing typo

GonzaloGM:
The delay is 500ms in `setTimeout`. I think that the description above the example might be incorrect because it's stating it's 5 seconds.


threepointone:
This was fixed yesterday, it’s now consistently 5 seconds everywhere

GonzaloGM:
@threepointone Oh, my bad, I had that page open from yesterday and I was reading it today when I saw the typo. Looks like it's fixed as you say, sorry about that!

PR #2236, opened 2019-08-13
Link accessible only on Open Networks.

arshita-kakkar:
Added a note under  : "If you haven't worked with JavaScript in the last few years, [these three points]..." as the link given (https://gist.github.com/gaearon/683e676101005de0add59e8bb345340c) works only on Open Networks and is inaccessible on Organizational Networks.

### DIFF
--- a/content/docs/hello-world.md
+++ b/content/docs/hello-world.md
@@ -40,7 +40,7 @@ React is a JavaScript library, and so we'll assume you have a basic understandin
 
 >Note
 >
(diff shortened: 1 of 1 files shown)


alexkrolick:
Do you mean that your organization's firewall is blocking GitHub gists?

arshita-kakkar:
Yes. The link won't work when firewalls are enabled by organization.
Error message : "This site can’t be reached The webpage at https://gist.github.com/gaearon/683e676101005de0add59e8bb345340c might be temporarily down or it may have moved permanently to a new web address."
ERR_TUNNEL_CONNECTION_FAILED

alexkrolick:
OK, that's not something that would affect everyone - maybe you can discuss unblocking GitHub gists with your administrator?

kambleaa007:
Does #Infosys blocks github gists, Its bad,
you can try, using your cellular network to access,
If its really urgent.

PR #2238, opened 2019-08-13
Fix sentence negation in Step 2

navsgh:


tesseralis:
No, the original version is correct. It's saying "(Even if) you're familiar with state, don't use it for this project". I agree it might cause some confusion. What do you think we should reword it to? Should we just delete that clause?

navsgh:
If that's the case, I think we can delete the sentence "If you’re familiar with the concept of state, don’t use state at all to build this static version".  Also, the last sentence "Since this is a static version of the app, you don’t need it." clarifies it. 

If we want to keep it we can change it to "Don’t use state at all to build this static version".

navsgh:
@tesseralis Please review the new commit. If that's better can you please close this? Thanks.

alexkrolick:
It's trying to say "don't use state to build this part (and if this is the first time you've heard of 'state' don't worry about it, we'll get to it".

Can you come up with something that preserves that meaning?

navsgh:
Updated the description. Please review

PR #2240, opened 2019-08-13
Update examples.md

kambleaa007:


gaearon:
Thanks but we're keeping this page focused on vanilla React. No Redux.

PR #2241, opened 2019-08-13
Change variable declaration of ‘newState’ in CodeEditor.js

Pewww:
I used 'const' instead of 'let' because it was not reassigning a variable 'newState'.

Thank you!


lex111:
@Pewww thank you!

PR #2242, opened 2019-08-14
Replacing incorrect link - Forms documentation

Guttz:
A link was redirecting to the same section it was in.

Now it properly forwards to the right part of the documentation, uncontrolled components.


tesseralis:
@Guttz how is the link incorrect? The original link was to the [file input tag section](https://reactjs.org/docs/uncontrolled-components.html#the-file-input-tag) of Uncontrolled Components. It's not the same section -- it's a section with the same name in a different page.

PR #2244, opened 2019-08-14
Hooks intro: Remove message about React Native support

MoOx:
Already landed for a while now.


lex111:
@MoOx thank you!

gaearon:
Instead of removing, can we add something like "React Native supports Hooks since the [React Native 0.59](https://facebook.github.io/react-native/blog/2019/03/12/releasing-react-native-059) release." please? We don't want RN to appear second class.

MoOx:
You are right @gaearon! Done in https://github.com/reactjs/reactjs.org/pull/2250

PR #2246, opened 2019-08-14
Rephrase confusing documentation about TestRenderer.create return value

TPXP:
Hello,

The addition of the `TestRenderer.act()` method to the documentation broke the logical flow thus saying "The returned instance has the following methods and properties." is confusing now since the `act()` method is documented below (while it belongs to the TestRenderer class). This commit aims at clarifying what is returned.

By the way, react rocks, thanks a lot! <3


gaearon:
Thanks, this is great.

PR #2248, opened 2019-08-15
Updated babel installation instructions for Flow

aiveneh:
updated how to install babel presets to use the newer namespace conventions that comes with newer versions of Babel, version 7.
My react app kept crashing when I tried installing Flow.
I have updated the docs on Flow's documentation, then I saw the same issue here.

Here is the PR on Flow's docs just in case - https://github.com/facebook/flow/pull/7856


alexkrolick:
💯

PR #2255, opened 2019-08-17
Language dropdown

tesseralis:
Improvement on the reactjs.org/languages page. Allows users to more easily see the same page in different languages.

Desktop:
Screen Shot 2019-08-16 at 9 05 41 PM

Mobile:
Screen Shot 2019-08-16 at 9 05 58 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
     "url": "https://github.com/reactjs/reactjs.org"
   },
   "dependencies": {
+    "@reach/menu-button": "^0.1.18",
     "babel-eslint": "^8.0.1",
     "eslint": "^4.8.0",
     "eslint-config-fbjs": "^2.0.0",
@@ -18,7 +19,7 @@
     "eslint-plugin-react": "^7.4.0",
     "eslint-plugin-relay": "^0.0.19",
     "flow-bin": "^0.56.0",
-    "gatsby": "^2.0.0",
+    "gatsby": "^2.13.67",
     "gatsby-plugin-catch-links": "^2.0.0",
     "gatsby-plugin-feed": "^2.0.0",
     "gatsby-plugin-glamor": "^2.0.0",
--- a/src/components/LayoutHeader/LanguageDropdown.js
+++ b/src/components/LayoutHeader/LanguageDropdown.js
@@ -2,7 +2,6 @@
  * Copyright (c) 2013-present, Facebook, Inc.
  *
  * @emails react-core
- * @flow
  */
 
 import React from 'react';
--- a/src/components/LayoutHeader/LanguageDropdown.js
+++ b/src/components/LayoutHeader/LanguageDropdown.js
@@ -1,3 +1,10 @@
+/**
+ * Copyright (c) 2013-present, Facebook, Inc.
+ *
+ * @emails react-core
+ * @flow
+ */
+
 import React from 'react';
 import {Link} from 'gatsby';
 
@@ -11,9 +18,9 @@ const completeLanguages = languages.filter(language => {
   return language.status == 2;
 });
 
-function getPageLink(code, pathname) {
+function getPageLink(code, {pathname, hash}) {
   const prefix = code === 'en' ? '' : `${code}.`;
-  return `https://${prefix}reactjs.org${pathname}`;
+  return `https://${prefix}reactjs.org${pathname + hash}`;
 }
 
 export default function LanguageDropdown({children, location}) {
@@ -49,7 +56,7 @@ export default function LanguageDropdown({children, location}) {
             css={linkStyle}
             key={lang.code}
             target="_blank"
-            href={getPageLink(lang.code, location.pathname)}>
+            href={getPageLink(lang.code, location)}>
             {lang.translated_name}
           </MenuLink>
         ))}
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -18,7 +18,7 @@
     "eslint-plugin-react": "^7.4.0",
     "eslint-plugin-relay": "^0.0.19",
     "flow-bin": "^0.56.0",
-    "gatsby": "^2.0.0",
+    "gatsby": "^2.13.67",
```
tesseralis: Old Gatsby (and scheduler) creates a version mismatch which makes it hard to use hooks!

tesseralis:
Hmm... I had a brainfart and thought this was a tooltip rather than a menu button/dropdown so I implemented it myself. I wonder if I should just use Reach UI (or another component library) instead since it has the accessibility built-in: https://ui.reach.tech/menu-button

alexkrolick:
> I wonder if I should just use Reach UI (or another component library) instead since it has the accessibility built-in

Seems like this would work!

src/components/LayoutHeader/LanguageDropdown.js
```diff
@@ -0,0 +1,62 @@
...
+          backgroundColor: `${colors.darker} !important`,
+          color: 'white !important',
+          fontSize: '1rem !important',
+          border: 'none !important',
+        }}>
+        {completeLanguages.map(lang => (
+          <MenuLink
+            css={linkStyle}
+            key={lang.code}
+            target="_blank"
```
lex111: Can remove it? Like this at `developer.mozilla.org`...

tesseralis: The difference is that in MDN, all the languages are under the same domain. vuejs.org opens a new window for each language, since it's technically different subdomains.

tesseralis:
Would someone be able to explain to me why CircleCI is failing? It seems that prettier is failing on files that I haven't touched, and I can't re-run the test.

3nuc:
Regarding #2261, I think there should be special handling of language redirections to blog entries. These are less likely to be already translated (eg. if the post was published a day ago) compared to existing docs that have been around for a while. I think if you change the language on a blog post just redirect to /blog

Right now, the dropdown will try to redirect you to a page that doesn't exist yet.

For example:
1. Go to https://deploy-preview-2255--reactjs.netlify.com/blog/2019/08/15/new-react-devtools.html
2. Change the language to eg. Polish (you'll get a 404)

src/components/LayoutHeader/LanguageDropdown.js
```diff
@@ -2,7 +2,6 @@
  * Copyright (c) 2013-present, Facebook, Inc.
  *
  * @emails react-core
- * @flow
```
3nuc: I assume this is why Circle-CI is failing. Re-adding @flow here and fixing the errors (npm run flow) will probably make Circle-CI pass. I would fix them myself as they are just code format errors but I'm not sure if I can edit somebody else's branch?

tesseralis: Done but it doesn't seem to be re-running

3nuc:
@bvaughn Could you please take a look? Doesn't seem like we can restart CircleCI builds on our own

bvaughn:
I re-ran, but CI failed again on Prettier check step. Does `yarn prettier:diff` fail for you locally?

3nuc:
It does fail locally for me.
As for `yarn prettier:diff`, in package.json it's `yarn nit:examples && yarn nit:source`.

`yarn nit:source` is `prettier --config .prettierrc --list-different "{gatsby-*.js,{flow-typed,plugins,src}/**/*.js}"`, so I ran `npx prettier [...]` with different loglevels to see what's going on and it doesn't return anything interesting on any of the loglevels. 

[Only `--loglevel debug` is a bit suspicious](https://gist.github.com/3nuc/434e27c8d486815f729c2ab39eef619b), looks like it's stuck in a loop but I'm not sure if that's how those debug logs normally look

bvaughn:
I see. I assumed that the ping for me to re-run CI meant that everything worked for you when you ran it locally :)

I don't really have time to dig into this at the moment :( I'm sorry. Pretty swamped at the moment with DevTools launch stuff. Maybe someone else can dig in and take a look. Excluding the occasional blog post, I don't really actively work in this repo.

tesseralis:
@3nuc It's not failing for me locally. I wonder why that is...

@bvaughn who would be the best person with permissions to contact about this?

bvaughn:
@tesseralis There really isn't anyone. This repo is something Dan and I try to help keep up with as we have time. Maybe one of us to pick it up and help look in a couple of days.

BartoszKlonowski (changes requested):
I wonder if we still want to have that dropdown 🤔 
Currently the language button leads to a page describing what translations are ready, and which are in progress, also providing one with opportunity to visit the repo (in case of wanting to contribute or check the progress).
Having the dropdown would remove that possibility, or force us to make yet another button to lead to that page, making the navigation confusing. Plus, if we would go for the dropdown we would have to decide which languages to show there, and which to not, also, the further maintenance effort would increase as we would have to check which languages are still up-to-date and which are not.
In short: I'm afraid I don't see any gains by having the dropdown for languages considering how those translations look like right now, and the approach the docs website has currently.

rickhanlonii:
Yeah the dedicated page replaces the need for a drop down. Thanks for the PR though!

PR #2258, opened 2019-08-17
Update testing-recipes.md

hontas:
Move mock restoration in Data Fetching-tests to afterEach to prevent it from polluting test environment if test was unsuccessful.


hontas:
Signed cla 👍

PR #2260, opened 2019-08-17
Handling Multiple Inputs

rajinder-yadav:
Text says to use "__name__" but code shows usage of "__type__".


threepointone:
Thank you!

PR #2261, opened 2019-08-17
Stay on the same page when switching languages

j-f1:
If you go to https://deploy-preview-2261--reactjs.netlify.com/docs/hooks-reference.html#usestate, then click the language button, then pick another language, you’ll be redirected to [the `useState` docs for that language](https://es.reactjs.org/docs/hooks-reference.html#usestate) rather than the homepage.

### DIFF
--- a/src/components/LayoutHeader/Header.js
+++ b/src/components/LayoutHeader/Header.js
@@ -199,7 +199,9 @@ const Header = ({location}: {location: Location}) => (
                 borderRadius: 15,
               },
             }}
-            to="/languages">
+            to={`/languages#!${location.pathname +
+              location.search +
+              location.hash}`}>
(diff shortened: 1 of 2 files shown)


j-f1:
I just noticed this is very similar to #2255 (cc @tesseralis). Should I close this PR in favor of that one?

One improvement my variant makes over #2255 is that it preserves query params and hash when switching languages, although that would be trivial to add to the other PR.

tesseralis:
@j-f1 I think I prefer having a dropdown, but good point on keeping the query params/hash when switching. I'll implement that

tesseralis:
@j-f1 if you want to contribute more to stuff relating to translation there's this: https://github.com/reactjs/reactjs.org/issues/1703

PR #2262, opened 2019-08-18
Changed binding naming convention for less confusion

MGhandour92:
So the reader will be less confused, he can now know the difference between what is referring to the function and what is to be binded to the button.


alexkrolick:
"Binding in the constructor" without renaming is kind of a community convention that's been around for a while. It seems like it should be shown in it's most common form here in case anyone runs across it in existing code. Nowadays class property syntax and hooks make this less pattern less likely to used.

PR #2264, opened 2019-08-18
Clarify effect of ‘only’ language

benfletcher:
Clarify that ‘only’ does not preclude `<Content />` from rendering.


alexkrolick:
Thanks

PR #2265, opened 2019-08-19
Flow typings for yaml files

tesseralis:
This means we no longer need to put a `$FlowFixMe` every time we import a yaml file.

This *does* require us to use [`gatsby-plugin-root-import`](https://www.gatsbyjs.org/packages/gatsby-plugin-root-import/), since yaml module declarations don't work on relative paths. If this isn't desirable then we don't have to do this.

### DIFF
--- a/flow-typed/yaml.js
+++ b/flow-typed/yaml.js
@@ -1,0 +1,47 @@
+declare module 'content/languages.yml' {
+  declare module.exports: {
+    name: string,
+    translated_name: string,
+    code: string,
+    status: 0 | 1 | 2,
+  }[];
+}
+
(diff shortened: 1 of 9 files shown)


threepointone:
Happy to merge this if you'd like to rebase and make sure it's still good?

bvaughn:
Oh this change looks nice. We should make sure it lands, even if Nat's too busy to work on it.

BartoszKlonowski:
Hello @tesseralis!
I fear this PR is no longer required, or able to be rebased as Gatsby is no longer used in the latest of react.dev.
Let me close it as outdated, but thank you for taking the time to contribute!

PR #2269, opened 2019-08-20
Devtools Docs: Add step to checkout the correct branch

ThatTobMate:
Include step to checkout v3 branch before building the old devtools extension. The current docs don't specify which branch to checkout before building the old extension.
It is specified in the README at https://github.com/facebook/react-devtools but should probably be specified here too.

https://github.com/reactjs/reactjs.org/blob/master/content/blog/2019-08-15-new-react-devtools.md#how-do-i-get-the-old-version-back

Screenshot 2019-08-20 at 10 59 05

### DIFF
--- a/content/blog/2019-08-15-new-react-devtools.md
+++ b/content/blog/2019-08-15-new-react-devtools.md
@@ -60,6 +60,9 @@ git clone https://github.com/facebook/react-devtools
 
 cd react-devtools
 
+# Checkout the previous release branch
+git checkout v3
+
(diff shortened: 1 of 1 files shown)


gaearon:
What’s wrong with building from master? Master isn’t v4.

ThatTobMate:
> What’s wrong with building from master? Master isn’t v4.

You can't build master, the source code has been removed.

gaearon:
Thanks for explaining. I forgot.

PR #2270, opened 2019-08-20
Extracting strings from source files for easier translations

GasimGasimzada:
This PR started from a conversation about creating a Gatsby theme and using that theme in localization pages. However, there are some strings in source files that need to be translatable before getting into creating the Gatsby theme. So, this PR is my attempt at extracting these translatable strings somewhere outside the src directory.

### Things To Do

- [ ] Move Footer links into content/footerNav.yml
- [ ] Move Header links into content/headerNav.yml
- [ ] Extract strings in following components
- - [ ] CodeEditor
- - [ ] CodeExample
- - [ ] ErrorDecoder
- - [ ] LayoutFooter
- - [ ] LayoutHeader (Languages string is not extracted)
- - [ ] MarkdownPage
- [ ] Extract strings from Gatsby config into a separate

So far, in this PR, I have extracted footer and header links into their respective YAML files and loaded them into appropriate components. However, there are some caveats (for example, I think it would make more sense if the Github link in Header and Footer are loaded from the same place.)

### DIFF
--- a/src/utils/sectionList.js
+++ b/src/utils/sectionList.js
@@ -11,6 +11,10 @@ import navCommunity from '../../content/community/nav.yml';
 import navDocs from '../../content/docs/nav.yml';
 // $FlowExpectedError
 import navTutorial from '../../content/tutorial/nav.yml';
+// $FlowExpectedError
+import navFooter from '../../content/footerNav.yml';
+// $FlowExpectedError
+import navHeader from '../../content/headerNav.yml';
 
 const sectionListDocs = navDocs.map(
   (item: Object): Object => ({
@@ -30,4 +34,6 @@ export {
   sectionListCommunity,
   sectionListDocs,
   navTutorial as sectionListTutorial,
+  navFooter as sectionListFooter,
+  navHeader as sectionListHeader,
 };
--- a/content/footerNav.yml
+++ b/content/footerNav.yml
@@ -1,0 +1,43 @@
+community:
+  title: Community
+
+docs:
+  title: Docs
+
+more:
+  title: More
+  items:
+    - title: Tutorial
+      to: /tutorial/tutorial.html
+    - title: Blog
+      to: /blog
+    - title: Acknowledgements
+      to: /acknowledgements.html
+    - title: React Native
+      to: https://facebook.github.io/react-native/
+      external: true
+
+channels:
+  title: Channels
+  items:
+    - title: Github
+      to: https://github.com/facebook/react
+      external: true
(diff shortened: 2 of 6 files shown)


src/utils/sectionList.js
```diff
@@ -30,4 +34,6 @@ export {
   sectionListCommunity,
   sectionListDocs,
   navTutorial as sectionListTutorial,
+  navFooter as sectionListFooter,
```
tesseralis: I don't think it makes sense to put these here. The other three are "section lists" because they appear in the sidebar for each type (Docs, Tutorial, and Community). I think it's best to just do

`import headerLinks from 'content/headerNav.yml`

in `Header.js`.

GasimGasimzada: > Just one nit about placement and naming.
> 
> @GasimGasimzada do you want to just have this PR be for the header/footer links and do the rest in another one so we can keep things simple?

Okay, let’s keep things simple. I will rename this PR to reflect that it is only about Footer and Header links.

GasimGasimzada:
> @GasimGasimzada do you want to just have this PR be for the header/footer links and do the rest in another one so we can keep things simple?

Okay, let’s keep things simple. I will rename this PR to reflect that it is only about Footer and Header links.

src/utils/sectionList.js
GasimGasimzada: Done.

src/utils/sectionList.js
```diff
@@ -11,6 +11,10 @@ import navCommunity from '../../content/community/nav.yml';
 import navDocs from '../../content/docs/nav.yml';
 // $FlowExpectedError
 import navTutorial from '../../content/tutorial/nav.yml';
+// $FlowExpectedError
```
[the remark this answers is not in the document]
GasimGasimzada: Done.

tesseralis:
@GasimGasimzada I think you still need the `$FlowFixMe`s for now since the other PR hasn't gone in yet (or would you like to take a look at it?)

GasimGasimzada:
@tesseralis sorry. I thought it was fixed when you referenced the PR. I added them back for YAML files. Regarding looking at it, let's do one thing at a time because I have an idea on how to go further regarding extracting strings. However, I need to spend some time investigating the problem before I can fix it.

GasimGasimzada:
@tesseralis hello, if everything is okay here, could you please merge this; so that, I can move to extracting translatable strings. These two files rely on these strings; so, there are going to be merge conflicts.

PR #2271, opened 2019-08-20
Added premium support for React.js projects

nishantrpai:
In case there are users, who are still finding some issues regarding solving problems. You could check out a service I started to solve fixes for React.js websites called Quickly react. (https://quicklyreact.com)


alexkrolick:
Sorry, this could potentially confuse people that this is a first-party support option from the react team

PR #2272, opened 2019-08-20
Updated Render-Atlanta's Social Media

js1599:
Updated the social media accounts for Render-Atlanta 2020


lex111:
@js1599 thank you!

PR #2274, opened 2019-08-21
Add reference to persist to SyntheticEvent's API

Sidnioulz:
Since persist() is referred to in the doc as a method for disabling event pooling, that must mean it's part of the SyntheticEvent API.

I'm proposing to state it more explicitly so that there's no doubt for newcomers like me whether it is the SyntheticEvent or the actual JS event that has the method.


threepointone:
Thank you!

PR #2277, opened 2019-08-22
Update hooks-reference.md

ScreamZ:
Hello,

I had some problem to understand what was lazy state, now that I figured out after some tests, I made an update to the hook documentation to clarify the lazy computation of initial state.

I think documentation should be the clearest as possible to help developpers learn the framework.

Hope you like it,

regards,
Andréas

### DIFF
--- a/content/docs/hooks-reference.md
+++ b/content/docs/hooks-reference.md
@@ -84,7 +84,9 @@ The "+" and "-" buttons use the functional form, because the updated value is ba
 
 #### Lazy initial state {#lazy-initial-state}
 
-The `initialState` argument is the state used during the initial render. In subsequent renders, it is disregarded. If the initial state is the result of an expensive computation, you may provide a function instead, which will be executed only on the initial render:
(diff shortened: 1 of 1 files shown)


content/docs/hooks-reference.md
```diff
@@ -84,7 +84,9 @@ The "+" and "-" buttons use the functional form, because the updated value is ba
 
 #### Lazy initial state {#lazy-initial-state}
 
-The `initialState` argument is the state used during the initial render. In subsequent renders, it is disregarded. If the initial state is the result of an expensive computation, you may provide a function instead, which will be executed only on the initial render:
+The `initialState` argument is the state used during the initial render. In subsequent renders, it is disregarded. This mean that if you are using a function to compute this initial state, the function will be invoked, but the return content will be ignored. This leads to unnecessary computation, which in some circumstances can cause performances issues.
```
Aprillion, line 87: The proposed change does not sound accurate: `useState(myFn)` works the same as `useState(() => myFn())`, it won't be executed in subsequent renders.

I am not sure what would be a good way to address `useState(myFn())` here, since that is how JavaScript works that `myFn()` expression is executed before calling `useState(returnValue)`, not any special behaviour of useState...

PR #2278, opened 2019-08-22
Improve examples in faq-functions.md

golopot:
It is useless to `.bind(this)` a function if that function does not use `this`. With this in mind the examples look weird.


threepointone:
I disagree, I think it's good to encourage binding on handlers regardless, avoiding confusion *later* when the dev references `this`. That said, this still seems like a nit and, and I'm inclined to reject it. Thank you very much fort he effort!

PR #2285, opened 2019-08-24
Add act() detail reference

ryota-murakami:
# Overview
I can understand that after act() all task is complete and ensure those results were applied into the DOM from this doc description.

However I was not sure about few points and then I found @threepointone's [excellent resource](https://github.com/threepointone/react-act-examples/blob/master/sync.md) and solve my doubt.

Actually current description isn't enough to guide concrete usage/usecase I guess.

AFAIK @threepointone is member of React official team so I think his resource is trusty too much.

Than you 😀

### DIFF
--- a/content/docs/testing-recipes.md
+++ b/content/docs/testing-recipes.md
@@ -56,6 +56,8 @@ You may use a different pattern, but keep in mind that we want to execute the cl
 
 ### `act()` {#act}
 
+> Please refer to the [react-act-examples/sync.md](https://github.com/threepointone/react-act-examples/blob/master/sync.md) for details on how act () works.
+
 When writing UI tests, tasks like rendering, user events, or data fetching can be considered as "units" of interaction with a user interface. React provides a helper called `act()` that makes sure all updates related to these "units" have been processed and applied to the DOM before you make any assertions:
 
 ```js


content/docs/testing-recipes.md
```diff
@@ -56,6 +56,8 @@ You may use a different pattern, but keep in mind that we want to execute the cl
 
 ### `act()` {#act}
 
+> Please refer to the [react-act-examples/sync.md](https://github.com/threepointone/react-act-examples/blob/master/sync.md) for details on how act () works.
```
alexkrolick: @threepointone this still the best resource?

threepointone:
Probably not. We can do better by just making the act() section on this better instead of linking to an unofficial/informal/long explanation. @ryota-murakami et al: what specific part from that repo do you believe is missing from the explanation in the official docs?

ryota-murakami:
@threepointone 

> Probably not. 

Oh, I see. 😭

>  what specific part from that repo do you believe is missing from the explanation in the official docs?

- Relationship between Hooks and `act ()`

So user can't recognize such as [usesace](https://github.com/threepointone/react-act-examples/blob/master/sync.md#secrets-of-the-act-api) from the doc.

content/docs/testing-recipes.md
threepointone: I'll work on something better.

ryota-murakami:
I think #3237 achieved well enough to close this PR.
At least adding more precise detail is @threepointone matter actually. 😅

@eps1lon I guess there are some inactive and stale similar Issues/PR  in the repo, could we close to clean some of the those? 

@koba04 I wanted to improvement like you at the time I submit the Issue, thank you so much! 🤗

eps1lon:
>  I guess there are some inactive and stale similar Issues/PR in the repo, could we close to clean some of the those?

Didn't find any related to "act". Though I am occasionally going through older issues PR/issue and close them when outdated.

ryota-murakami:
@eps1lon I see, thank you for digging into about it!

PR #2296, opened 2019-08-27
Removed hyperlink from 'a'

arshita-kakkar:
The sentence in line114 - "For this change, we’ve provided a [codemod](https://github.com/reactjs/react-codemod#error-boundaries) to automatically migrate your code" had hyperlink added to [a codemod] instead of only to [codemod]. Updated this.


lex111:
@arshita-kakkar thank you!

PR #2298, opened 2019-08-27
add mount and shallow descriptions

iqbal125:
I thought it might be helpful to briefly describe shallow and mount rendering.


threepointone:
We actively wanted to avoid mentioning shallow rendering in these docs, since we don't really recommend it anymore. Thank you for the PR!

PR #2301, opened 2019-08-28
Doc error

arshita-kakkar:
Changed Line 94- You can't use Hooks *inside* of a class component, but you can definitely mix classes and function components with Hooks in a single tree.  
to 
"You can't use Hooks *inside* a class component, but you can definitely mix classes and function components with Hooks in a single tree. "


threepointone:
Thank you!

PR #2304, opened 2019-08-28
Documentation correction

arshita-kakkar:
Line 290 - ". However, instead we recommend to split state into multiple state variables based on which values tend to change together."
Both 'however' and 'instead' are contrasting words and only one of them should be used. 
Either "Instead we recommend to split state into multiple state variables based on which values tend to change together." OR "However, we recommend to split state into multiple state variables based on which values tend to change together."
Updated the sentence.


arshita-kakkar:
Hey @alexkrolick the sentence has been updated and can be merged now.

threepointone:
Thank you! Tweaked it a little myself.

PR #2306, opened 2019-08-28
Update CodeSplitting to clarify that lazy() must be used with Suspense

jtbandes:
It seems that a `React.lazy()` **must** be wrapped in a `<Suspense>`. Update the wording of the documentation to clarify this (and remove a confusing example which shows the lazy component being rendered without any Suspense around it).

### DIFF
--- a/content/docs/code-splitting.md
+++ b/content/docs/code-splitting.md
@@ -124,37 +124,21 @@ The `React.lazy` function lets you render a dynamic import as a regular componen
 
 ```js
 import OtherComponent from './OtherComponent';
-
-function MyComponent() {
-  return (
-    <div>
-      <OtherComponent />
-    </div>
-  );
-}
 ```
 
 **After:**
 
 ```js
 const OtherComponent = React.lazy(() => import('./OtherComponent'));
-
-function MyComponent() {
-  return (
-    <div>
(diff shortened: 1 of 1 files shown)


jtbandes:
@lex111 Could you please advise on the best way to get this change reviewed? Thanks 🙂

lex111:
@jtbandes this is not small change, so let's ask to review by @alexkrolick ?

content/docs/code-splitting.md
```diff
@@ -124,37 +124,21 @@ The `React.lazy` function lets you render a dynamic import as a regular componen
...
-  );
-}
 ```
 
-This will automatically load the bundle containing the `OtherComponent` when this component gets rendered.
+This will automatically load the bundle containing the `OtherComponent` when this component is first rendered.
 
 `React.lazy` takes a function that must call a dynamic `import()`. This must return a `Promise` which resolves to a module with a `default` export containing a React component.
 
 ### Suspense {#suspense}
```
alexkrolick: I would say we should remove this subheading because it creates a separation between the sections. Since both Suspense and the import() change are required we should keep them together.

jtbandes: Done

PR #2307, opened 2019-08-29
Update hooks-faq.md

arshita-kakkar:
"Whether a component is a class or a function that uses Hooks is an implementation detail of that component." wasn't framed correctly and had an ambiguous implication. 
Updated it to -  Hooks used by a component (class or function) is an implementation detail of that component.


threepointone:
Your suggestion isn't right either, since Hooks cannot be used in classes.

PR #2314, opened 2019-08-31
Add Translatable component

GasimGasimzada:
In this ongoing PR, I am extracting strings to into a separate YAML file called "content/strings.yml"

In order to use the translation, I have added a `Translatable` component. This component accepts strings as children (enforced via Flow) and returns the translatable text if it exists. If the key does not exist in strings.yml or if the value for the key is an empty string (`""`), it will return the original, untranslated text. This way, we can easily apply the translations into existing texts without a hassle. 

Some things that I am considering but would like a feedback before I can move forward. Add `as` prop to `Translatable` component that accepts React Type or DOM Element Type. So, instead of writing the following:

```
<p>
  <Translatable>When you encounter an error, you'll receive a link to this page for that specific error and we'll show you the full error text.</Translatable>
</p>
```

We can write down the following:

```
<Translatable as="p">When you encounter an error, you'll receive a link to this page for that specific error and we'll show you the full error text.</Translatable>
```

---

Components that need to be translated

- [x] components/CodeEditor
- [x] components/CodeExample
- [x] components/ErrorDecoder
- [ ] components/LayoutFooter
- [ ] components/LayoutHeader
- [x] components/MarkdownPage
- [ ] components/TitleAndMetaTags
- [ ] pages/blog/all.html.js
- [ ] pages/404.js
- [ ] pages/acknowledgements.html.js
- [ ] pages/index.js
- [ ] pages/jsx-compiler.html.js
- [ ] pages/languages.js
- [ ] pages/versions.js
- [ ] templates/components/NavigationFooter.js
- [ ] templates/codepen-example.js
- [ ] templates/blog.js (Not sure)

### DIFF
--- a/content/strings.yml
+++ b/content/strings.yml
@@ -1,0 +1,6 @@
+"Babel could not be loaded": ""
+"This can be caused by an ad blocker. If you're using one, consider adding reactjs.org to the whitelist so the live code examples will work.": ""
+"Loading code example...": ""
+"When you encounter an error, you'll receive a link to this page for that specific error and we'll show you the full error text.": ""
+"The full text of the error you just encountered is:": ""
+"Edit this page": ""
--- a/src/components/CodeEditor/CodeEditor.js
+++ b/src/components/CodeEditor/CodeEditor.js
@@ -10,6 +10,7 @@ import Remarkable from 'remarkable';
 import {LiveEditor, LiveProvider} from 'react-live';
 import {colors, media} from 'theme';
 import MetaTitle from 'templates/components/MetaTitle';
+import Translatable from '../Translatable';
 
 // Replace unicode to text for other languages
 const unicodeToText = text =>
@@ -62,12 +63,14 @@ class CodeEditor extends Component {
     if (showBabelErrorMessage) {
       errorMessage = (
         <span>
-          Babel could not be loaded.
+          <Translatable>Babel could not be loaded.</Translatable>
           <br />
           <br />
-          This can be caused by an ad blocker. If you're using one, consider
-          adding reactjs.org to the whitelist so the live code examples will
-          work.
+          <Translatable>
+            This can be caused by an ad blocker. If you're using one, consider
+            adding reactjs.org to the whitelist so the live code examples will
+            work.
+          </Translatable>
         </span>
       );
     } else if (error != null) {
--- a/src/components/CodeExample/CodeExample.js
+++ b/src/components/CodeExample/CodeExample.js
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 
 import {colors, media} from 'theme';
 import CodeEditor from '../CodeEditor/CodeEditor';
+import Translatable from '../Translatable';
 
 class CodeExample extends Component {
   render() {
@@ -61,7 +62,9 @@ class CodeExample extends Component {
         {loaded ? (
           <CodeEditor code={code} containerNodeID={containerNodeID} />
         ) : (
-          <h4>Loading code example...</h4>
+          <h4>
+            <Translatable>Loading code example...</Translatable>
+          </h4>
         )}
       </div>
     );
--- a/src/components/ErrorDecoder/ErrorDecoder.js
+++ b/src/components/ErrorDecoder/ErrorDecoder.js
@@ -8,6 +8,7 @@
 import React from 'react';
 
 import type {Node} from 'react';
+import Translatable from '../Translatable';
 
 function replaceArgs(msg: string, argList: Array<string>): string {
   let argIdx = 0;
@@ -69,8 +70,10 @@ function ErrorResult(props: {|code: ?string, msg: string|}) {
   if (!code) {
     return (
       <p>
-        When you encounter an error, you'll receive a link to this page for that
-        specific error and we'll show you the full error text.
+        <Translatable>
+          When you encounter an error, you'll receive a link to this page for
+          that specific error and we'll show you the full error text.
+        </Translatable>
       </p>
     );
   }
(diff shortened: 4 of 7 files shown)


content/strings.yml
```diff
@@ -0,0 +1,48 @@
...
+"Next article": ""
+"React - Acknowledgements": ""
+"We'd like to thank all of our contributors:": ""
+"In addition, we're grateful to": ""
+"<0>Jeff Barczewski</0> for allowing us to use the <1>react</1> package name on npm.": ""
+"<0>Christopher Aue</0> for letting us use the <1>reactjs.com</1> domain name and the <2>@reactjs</2> username on Twitter.": ""
+"<0>ProjectMoon</0> for letting us use the <1>flux</1> package name on npm.": ""
+"Shane Anderson for allowing us to use the <1>react</1> org on GitHub.": ""
+"<0>Dmitri Voronianski</0> for letting us use the <1>Oceanic Next</1> color scheme on this website.": ""
+"React - All Posts": ""
```
tesseralis: We should change these page titles to be consistent with the way they're displayed for the Docs/Tutorial section (e.g. "Getting Started - React"). That way we can just have one string (the page title) and append React to the end.

content/strings.yml
```diff
@@ -0,0 +1,48 @@
...
+"Redirecting to Codepen...": ""
+"Not automatically redirecting?": ""
+"All posts ...": ""
+"Recent Posts": ""
+"Previous article": ""
+"Next article": ""
+"React - Acknowledgements": ""
+"We'd like to thank all of our contributors:": ""
+"In addition, we're grateful to": ""
+"<0>Jeff Barczewski</0> for allowing us to use the <1>react</1> package name on npm.": ""
```
tesseralis: Could we turn this section into markdown? That way we can just rely on the markdown parser and not have roll out our own markup.

content/strings.yml
```diff
@@ -0,0 +1,48 @@
+"Babel could not be loaded": ""
```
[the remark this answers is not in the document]
GasimGasimzada: I will add `i18nKey` prop to Translatable component and keep `children` text as default.

Regarding loading translations per section, to be honest, I do not like this approach. Even if translations are separated per file, components or pages should not care where the translations are coming from. That was the main reason I separated translation management into its own component. Keeping them in one file or in many files shouldn't require us to change any of the components. We can do changes to Translatable component and be done with the changes.

I have checked out i18next docs to understand how they are managing it I liked using strings to separate modules (e.g "blog:allPosts"). We can just write our strings this way:

```
<Translatable i18nKey="codeEditor:babelNotLoaded" />
```

or

```
<Translatable i18nKey="babelNotLoaded" module="codeEditor" />
```

However, whether codeEditor is in its own YAML file, a dictionary in a single YAML file, or just a dictionary key named "codeEditor:babelNotLoaded" will not matter to the components who call `Translatable`.

In order to conform to ESLint rule, we can move default text into its own prop:

    <Translatable i18nKey="codeEditor:babelNotLoaded" defaultText="Babel Not Loaded" />

or we remove `Translatable` component but keep the translate function and just call it to get texts:

   {translate('codeEditor:babelNotLoaded', 'Babel Not Loaded')}

Please, let me know how you want to approach this.

content/strings.yml
GasimGasimzada: Helmet has titleTemplate prop; so, I utilized it to solve this issue. Now, there are two new strings:

    "titleTemplate": "%s – React"
    "blogTitleTemplate": "%s – React Blog"

I also added title template for <TitleAndMetaTags> component. This way, different pages can have different title templates. I kept it very flexible for simplicity. If you want to, I can keep the same template and add a postfix (like how it used to) that will change based Blog or other pages.

I kept "React – A Javascript Library..." for homepage.

content/strings.yml
GasimGasimzada: That would be great. This way, we can move Acknowledgements and Versions into content. However, I am not sure how they will work because currently there is content and YAML file for both of these pages. What if we use MDX and render the component in the files? Would that be a thing that you will be interested in. So, the resulting MDX will look like this:

```mdx
---
title: Acknowledgement
---

We'd like to thank all of our contributors:

<ContributorsList />

* [Jeff Barczewski](https://github.com/jeffbski") for allowing us to use the [react](https://www.npmjs.com/package/react) package name on npm.
...
```

However, I am not sure how this is going to work if we create a Gatsby theme. Is it possible to inject a component into Markdown Remark per page? If yes, this would actually work (for example, only `ContributorList` will be accessible from Acknowledgements page).

tesseralis: @kentcdodds was working on this here: https://github.com/reactjs/reactjs.org/pull/2198

Kent, do you think you could shed some light on Gasim's concerns?

I think MDX would be cool but you're right and I don't know how custom components would work. I think my preference would be:

* If a component is only used in one page, it's okay if it stays in the /content folder next to the page that uses it
* If it's used in multiple pages, put it in custom-components of the Gatsby theme.

kentcdodds: I feel like there are two routes you can go with enhancing the capabilities of markdown (to make things interactive for example):

1. Make a bunch of custom plugins
2. Use MDX

I think that using MDX is fantastic and reduces more complexity than it adds.

GasimGasimzada: Can someone try installing `gatsby-plugin-mdx`. I have been trying to make it work but it keeps failing (Node: v8.16.1):

```bash
yarn add gatsby-plugin-mdx @mdx-js/mdx @mdx-js/react
```

In gatsby-config.js I added `gatsby-plugin-mdx`

Running `yarn start` throws the following errors (I will write the full error message later):

> `createTypes` is undefined
> Cannot call `filter` on undefined

PR #2329, opened 2019-09-08
Wrongly placed comma.

guryashzone:
Wrongly placed comma, which may confuse many novice developers.


alexkrolick:
Trailing commas are valid syntax and are inserted automatically by the prettier formatter

https://github.com/reactjs/reactjs.org/blob/master/.prettierrc

What makes you think this is confusing?

guryashzone:
A trail comma was a bit confusing as it was not in previous ES formats. May
a brief description on such new syntax is appreciated at the side of the
quoted code.

PR #2330, opened 2019-09-09
Filled in details for multi-select element

carlin-q-scott:
I had to play around with the select tag's code pen and read several SO answers and comments before I figured how to actually use the multi-select element that was briefly mentioned in this document. So I updated the docs to provide a full-blown description of it with example code.

Thanks for providing such excellent documentation.

### DIFF
--- a/content/docs/forms.md
+++ b/content/docs/forms.md
@@ -182,13 +182,49 @@ class FlavorForm extends React.Component {
 
 Overall, this makes it so that `<input type="text">`, `<textarea>`, and `<select>` all work very similarly - they all accept a `value` attribute that you can use to implement a controlled component.
 
-> Note
->
-> You can pass an array into the `value` attribute, allowing you to select multiple options in a `select` tag:
->
->```js
-><select multiple={true} value={['B', 'C']}>
->```
+*** select multiple ***
+Just like in HTML, select elements can become multi-select elements through the addition of the *multiple* attribute to the select tag. But handling of the value attribute in react is more complex due to having an array of values. 
+
+Here's our example select element with multi-select support:
+
+```js{4,11,24}
+class FlavorForm extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {value: ['coconut']};
+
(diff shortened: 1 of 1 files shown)


content/docs/forms.md
```diff
@@ -182,13 +182,49 @@ class FlavorForm extends React.Component {
...
 
-> Note
->
-> You can pass an array into the `value` attribute, allowing you to select multiple options in a `select` tag:
->
->```js
-><select multiple={true} value={['B', 'C']}>
->```
+### The select multiple Tag
+Just like in HTML, select elements can become multi-select elements through the addition of the *multiple* attribute to the select tag. But handling of the value attribute in react is more complex due to having an array of values.
```
alexkrolick:
```suggestion
Just like in HTML, select elements can become multi-select elements through the addition of the *multiple* attribute to the select tag. In React, the `value` attribute for `multiple` is handled as an array. 
```

content/docs/forms.md
```diff
@@ -182,13 +182,49 @@ class FlavorForm extends React.Component {
...
+    this.handleChange = this.handleChange.bind(this);
+    this.handleSubmit = this.handleSubmit.bind(this);
+  }
+
+  handleChange(event) {
+    this.setState({value: Array.from(event.target.options).filter(o => o.selected).map(o => o.value)});
+  }
+
+  handleSubmit(event) {
+    alert('Your favorite flavor is: ' + this.state.value);
```
alexkrolick: It might make a bit more sense if this used a more multiple + plural example, like "Your favorite flavors are" or "You picked: ..."

carlin-q-scott: I thought about that, but then I thought if this is my favorite flavor of ice cream then it would be multiple flavors. So the language actually makes sense in the current context.

content/docs/forms.md
carlin-q-scott: I see where you're going with that but I don't like your wording of it either. How about: 
> In React, this results in the `value` attribute being handled as an array of selected values, rather than just one value.

PR #2332, opened 2019-09-09
Memo: Document behavior with useContext

jarstelfox:
It is not obvious that useContext and Memo work well together. Namely, if a child component listens to a context and updates out from under them memo'd function, does the memo'd component have the right state if re-rendered?

yes, actually, but this is undocumented. Now it's not!

Closes: #1923

### DIFF
--- a/content/docs/reference-react.md
+++ b/content/docs/reference-react.md
@@ -144,6 +144,10 @@ function areEqual(prevProps, nextProps) {
 export default React.memo(MyComponent, areEqual);
 ```
 
+> Note
+>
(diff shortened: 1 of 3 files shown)


content/docs/reference-react.md
```diff
@@ -144,6 +144,10 @@ function areEqual(prevProps, nextProps) {
 export default React.memo(MyComponent, areEqual);
 ```
 
+> Note
+>
+> [`useContext`](/docs/hooks-reference.html#usecontext) will tigger a render of the memoized component. This render will also cause the `React.memo` [higher order component](/docs/higher-order-components.html) to re-cache the result.
```
alexkrolick: typo: trigger

sophiebits:
Added a bit more detail and merged, thank you!

PR #2335, opened 2019-09-10
Add warning about cache purging to useCallback reference

### LINKED ISSUE #2334: useCallback docs don't warn that the callback might run more than once
From the [useCallback docs](https://reactjs.org/docs/hooks-reference.html#usecallback): 

>`useCallback` will return a memoized version of the callback *that only changes if one of the dependencies has changed*.

(emphasis added). Then, the docs also say:

> `useCallback(fn, deps)` is equivalent to `useMemo(() => fn, deps)`.

but, in the [useMemo docs](https://reactjs.org/docs/hooks-reference.html#usememo)

> In the future, React may choose to “forget” some previously memoized values and recalculate them on next render, e.g. to free memory for offscreen components. 

Currently, the `useCallback` docs could mislead developers into writing code that will break in the future (eg, using `useCallback` for a function that must only run once).

It would be an improvement to have the same warning in the `useCallback` docs, even if only a small mention or warning to check the `useMemo` docs for the caveat.

TimothyJones:
This PR fixes #2334 by adding a warning about cache purging to the `useCallback` reference documentation.

### DIFF
--- a/content/docs/hooks-reference.md
+++ b/content/docs/hooks-reference.md
@@ -318,7 +318,7 @@ Returns a [memoized](https://en.wikipedia.org/wiki/Memoization) callback.
 
 Pass an inline callback and an array of dependencies. `useCallback` will return a memoized version of the callback that only changes if one of the dependencies has changed. This is useful when passing callbacks to optimized child components that rely on reference equality to prevent unnecessary renders (e.g. `shouldComponentUpdate`).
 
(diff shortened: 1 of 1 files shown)


TimothyJones:
This has been open for quite a while - I'll close it in a month or two if it's not wanted, just to get it off my list.

BartoszKlonowski:
Closing as per comment in the issue.
Thank you, @TimothyJones for taking the time to contribute!

PR #2338, opened 2019-09-11
Add alt prop for img element.

petrewoo:

### DIFF
--- a/content/docs/introducing-jsx.md
+++ b/content/docs/introducing-jsx.md
@@ -98,7 +98,7 @@ const element = <div tabIndex="0"></div>;
 You may also use curly braces to embed a JavaScript expression in an attribute:
 
 ```js
-const element = <img src={user.avatarUrl}></img>;
+const element = <img src={user.avatarUrl} alt=''></img>;
 ```
 
 Don't put quotes around curly braces when embedding a JavaScript expression in an attribute. You should either use quotes (for string values) or curly braces (for expressions), but not both in the same attribute.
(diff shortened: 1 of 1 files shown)


content/docs/introducing-jsx.md
```diff
@@ -98,7 +98,7 @@ const element = <div tabIndex="0"></div>;
 You may also use curly braces to embed a JavaScript expression in an attribute:
 
 ```js
-const element = <img src={user.avatarUrl}></img>;
+const element = <img src={user.avatarUrl} alt='introducing-jsx'></img>;
```
fanny: The `alt` would be better, if it related with the img content, so:
```suggestion
const element = <img src={user.avatarUrl} alt='avatar-url'></img>;
```

content/docs/introducing-jsx.md
```diff
@@ -114,7 +114,7 @@ Don't put quotes around curly braces when embedding a JavaScript expression in a
 If a tag is empty, you may close it immediately with `/>`, like XML:
 
 ```js
-const element = <img src={user.avatarUrl} />;
+const element = <img src={user.avatarUrl} alt='introducing-jsx'/>;
```
fanny:
```suggestion
const element = <img src={user.avatarUrl} alt='avatar-url'></img>;
```

content/docs/introducing-jsx.md
petrewoo: I re-read the doc and i think you are right. 
`alt`  attribute specifies an alternate text for an image.
I have tried VoiceOver and i think code `src={user.avatarUrl}` is clear enough for everyone, so i removed value of `alt` prop.

fanny: Yeah, but the `alt` attribute is a alternative text, when image is not rendered, so it's still cool keep the attribute.

BartoszKlonowski:
Hello @petrewoo, thank you for the contribution!
Unfortunately the part you modify in this PR no longer exists in the latest version of docs, so I will close this PR as outdated.
But please feel invited to create new PRs for open issues!

PR #2341, opened 2019-09-12
update state hook docs with function argument

kajas90:
Update of state hook documentation with function as a possible argument to set function.


kajas90:
makes sense, I haven't noticed the functional updates part in hooks API reference - people tend to look for the whole info about useState in it's own page I believe. I will take a look if there is any spot to put reference there. Thanks for the review Sophie!

PR #2347, opened 2019-09-16
fixed search bar overlapping problem

nikhilgupta30:
This pull request fixes the search bar overlapping navbar options problem. 
Now there are 3 different search bar sizes small, medium and large for every window size.

Below are the screenshots in every mode of search bar.

largeSearch
mediumSearch
smallSearch


jcserracampos:
I am not member here, but LGTM.

PR #2349, opened 2019-09-16
fix broken link to babel repl on jsx-compiler page

joedel:
Babel's site doesn't like the trailing slash and 404's, removing it links correctly to the repl.


joedel:
From my testing the link as is may only 404 intermittently, but without slash works consistently.

Direct link to page with change:

Before: https://reactjs.org/jsx-compiler.html
After: https://deploy-preview-2349--reactjs.netlify.com/jsx-compiler.html/

This page is first search result for 'jsx compiler', so would be ideal to have a working link.

threepointone:
Thank you!

PR #2360, opened 2019-09-23
Fix a bug in the TestRenderer.act() example code

carlbennettnz:
`TestRenderer.act()` is a void function, so assigning its result to `root` breaks the final assertion.


threepointone:
fixed in a separate PR, thank you!

PR #2367, opened 2019-09-24
Add doc for context displayaName

M-Izadmehr:
This PR adds a description of `Context.displayName` which is used to show a specific name for the React DevTools.
Also solved [this Issue](https://github.com/facebook/react/issues/16851)

### DIFF
--- a/content/docs/context.md
+++ b/content/docs/context.md
@@ -203,7 +203,7 @@ Context object accepts a `displayName` string property. React DevTools uses this
 
 For example, the following component will appear as MyContext in the DevTools:
 
-```js
+```js{2}
 const MyContext = React.createContext(/* some value */);
 MyContext.displayName = 'MyDisplayName';
 
--- a/content/docs/context.md
+++ b/content/docs/context.md
@@ -15,7 +15,7 @@ In a typical React application, data is passed top-down (parent to child) via pr
   - [Context.Provider](#contextprovider)
   - [Class.contextType](#classcontexttype)
   - [Context.Consumer](#contextconsumer)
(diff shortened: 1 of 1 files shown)


content/docs/context.md
```diff
@@ -196,6 +197,20 @@ Requires a [function as a child](/docs/render-props.html#using-props-other-than-
 > 
 > For more information about the 'function as a child' pattern, see [render props](/docs/render-props.html).
 
+### `Context.displayName` {#contextdisplayName}
```
eps1lon: This doesn't get parsed properly it seems. Maybe only lowercase is allowed?

M-Izadmehr: Yeah it seems that all links are in lowercase, I changed it to lowercase and pushed it.

content/docs/context.md
```diff
@@ -196,6 +197,20 @@ Requires a [function as a child](/docs/render-props.html#using-props-other-than-
...
+
+Context object accepts a `displayName` string property. React DevTools uses this string to determine what to display for the context.
+
+For example, the following component will appear as MyContext in the DevTools:
+
+```js{2}
+const MyContext = React.createContext(/* some value */);
+MyContext.displayName = 'MyDisplayName';
+
+<MyContext.Provider> // "MyContext.Provider" in DevTools
```
alexkrolick: shouldn't this be `// "MyDisplayName.Provider" in DevTools`?

M-Izadmehr: Done:)

alexkrolick:
thank you!

PR #2368, opened 2019-09-24
Update Versioning FAQ

acdlite:
Includes a new section describing our approach to minor releases, and why they don't always include new features.

### DIFF
--- a/content/docs/faq-versioning.md
+++ b/content/docs/faq-versioning.md
@@ -55,7 +55,7 @@ It's possible that a minor release will not include new features. [This is allow
 
 However, it does raise the question of why these releases aren't versioned as patches instead.
 
-The answer is that any change to React (or other software) carries some risk of breaking in unexpected ways. Imagine a scenario where a patch release that fixes one bug accidentally introduces a different bug. This would not only be disruptive to developers, but also likely harm their confidence in future patch releases. It's especially regrettable if the original fix is for a bug that is rarely encountered in practice.
+The answer is that any change to React (or other software) carries some risk of breaking in unexpected ways. Imagine a scenario where a patch release that fixes one bug accidentally introduces a different bug. This would not only be disruptive to developers, but also harm their confidence in future patch releases. It's especially regrettable if the original fix is for a bug that is rarely encountered in practice.
 
 We have a pretty good track record for keeping React releases free of bugs, but patch releases have an even higher bar for reliability because most developers assume they can be adopted without adverse consequences.
 
--- a/content/docs/faq-versioning.md
+++ b/content/docs/faq-versioning.md
@@ -51,15 +51,13 @@ That said, if we expect that a change on this list will cause broad problems in 
 
 ### If a Minor Release Includes No New Features, Why Isn't It a Patch? {#minors-versus-patches}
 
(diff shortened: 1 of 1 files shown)


content/docs/faq-versioning.md
```diff
@@ -10,12 +10,14 @@ React follows [semantic versioning (semver)](https://semver.org/) principles.
 
 That means that with a version number **x.y.z**:
 
+* When releasing **critical bug fixes**, we make a **patch release** by changing the **z** number (ex: 15.6.2 to 15.6.3).
+* When releasing **new features** or **non-critical fixes**, we make a **minor release** by changing the **y** number (ex: 15.6.2 to 15.7.0).
```
acdlite: I reversed the order of these items so that "critical bug fixes" comes before "non-critical"

content/docs/faq-versioning.md
```diff
@@ -46,3 +48,19 @@ In general, we *don't* bump the major version number for changes to:
 This policy is designed to be pragmatic: certainly, we don't want to cause headaches for you. If we bumped the major version for all of these changes, we would end up releasing more major versions and ultimately causing more versioning pain for the community. It would also mean that we can't make progress in improving React as fast as we'd like.
 
 That said, if we expect that a change on this list will cause broad problems in the community, we will still do our best to provide a gradual migration path.
+
+### If a Minor Release Includes No New Features, Why Isn't It a Patch? {#minors-versus-patches}
+
+It's possible that a minor release will not include new features. [This is allowed by semver](https://semver.org/#spec-item-7):
+
+> [A minor version] MAY be incremented if substantial new functionality or improvements are introduced within the private code. It MAY include patch level changes.
```
acdlite: Is there a way I can render this blockquote without the yellow box treatment? Or should I rewrite to use an inline quote?

Screen Shot 2019-09-24 at 4 55 17 PM

alexkrolick: no, all the MD blockquotes have that

content/docs/faq-versioning.md
```diff
@@ -46,3 +48,19 @@ In general, we *don't* bump the major version number for changes to:
...
+
+However, it does raise the question of why these releases aren't versioned as patches instead.
+
+The answer is that any change to React (or other software) carries some risk of breaking in unexpected ways. Imagine a scenario where a patch release that fixes one bug accidentally introduces a different bug. This would not only be disruptive to developers, but also likely harm their confidence in future patch releases. It's especially regrettable if the original fix is for a bug that is rarely encountered in practice.
+
+We have a pretty good track record for keeping all React releases free of bugs, including minor and major versions. But patch releases have an even higher bar for reliability, because most developers assume they can be adopted without adverse consequences — indeed, that's why patch releases exist. If we push out an important patch, we need developers to trust that it can be safely applied.
```
bvaughn: nit suggestion:

> We have a pretty good track record for keeping React releases free of bugs, but patch releases have an even higher bar for reliability because most developers assume they can be adopted without adverse consequences

content/docs/faq-versioning.md
bvaughn: Alex is right. This is the default `<blockquote>` styling for the website.

content/docs/faq-versioning.md
acdlite: Good edit

PR #2377, opened 2019-09-30
WebApp manifest added to allow users "Add to Homescreen" documentations

### LINKED ISSUE #2259: Suggestion to add WebApp manifest in reactjs.org
If we allow users to **Add to homescreen** the https://reactjs.org I think it would be helpful for people who want to have react docs on their homescreen and it would allow them to easily access the docs anytime they want.
Along with this we can even set the theme color for URL bar.

we can add gatsby-plugin-manifest to achieve this.

saurabhdaware:
This solves issue #2259 

## Then

Currently there is not manifest set for the website so logo is blank and title is set as a name. Also the color of URL bar in chrome is plain white. Here's how it will look after this PR.

## Now
   

Image 1 is to show the new color of URL bar in chrome
Image 2 is the screen after selecting "Add to homescreen"
Image 3 is the splash screen after opening it as WebApp

I've added color to the URL bar, Configured name, short_name and other properties. and added a logo file in `src/images/react_logo.png` (Needed 512px png logo so had to add a new logo)

This is my first contribution here so I'm sorry if I missed out on something


gatsby-config.js
```diff
@@ -159,5 +159,17 @@ module.exports = {
...
+    {
+      resolve: `gatsby-plugin-manifest`,
+      options: {
+        name: 'React Docs',
+        short_name: 'React',
+        start_url: '/',
+        background_color: '#20232a',
+        theme_color: '#20232a',
+        display: 'standalone',
+        icon: 'src/images/react_logo.png',
```
bvaughn: I don't think this is right. The URL needs to be a web accessible address. This is the source location, but you can't load e.g. https://deploy-preview-2377--reactjs.netlify.com/src/images/react_logo.png

So this just makes the "Add to Homescreen" popup spin for a while before it gives up on trying to load the icon.

saurabhdaware: I am using plugin gatsby-plugin-manifest so the plugin is handling that path.

image

 So it is simply taking that image and generating required variant sizes before it builds. As you can see in the screenshot it generated those images.

bvaughn: Hm, I guess I'm just saying that it doesn't seem to work.

iOS+Safari ignores it entirely and shows a thumbnail of the homepage, and Android+Chrome shows the correct icon for reactjs.org (without this PR even being added).

What am I overlook?

saurabhdaware: I've added some commits can you check now? 
So this repository already had `gatsby-plugin-manifest` installed but it was not used so instead of installing another version I used the same plugin. 
Anyway now I've added a `legacy: true` which is supposed to fix issue for IOS. If that doesn't work then installing latest version of `gatsby-plugin-manifest` would be a better choice.

bvaughn: Seems like the behavior is the same as I described before.

saurabhdaware: So the problem was currently there is `gatsby-plugin-manifest@2.0.5` installed which seems to not add apple meta tags in html and since the newer versions of the plugin are dependent on newer versions of other packages. we would have to update a lot of plugins (which may break stuff)

So instead of that I managed to fix things by adding meta tags manually which should work on both ios and android.

Commit: 
7fe5098a6113dfe4546e2ab0a917d81b1af91195

Also, I moved `src/images/react_logo.png` to `static/logo-512x512.png` and added `static/logo-180x180.png` for apple.

gatsby-config.js
```diff
@@ -159,5 +159,18 @@ module.exports = {
...
+    {
+      resolve: `gatsby-plugin-manifest`,
+      options: {
+        name: 'React Docs',
+        short_name: 'React',
+        start_url: '/',
+        background_color: '#20232a',
+        theme_color: '#20232a',
+        display: 'standalone',
+        icon: 'static/logo-512x512.png',
```
bvaughn: What's this `icon` actually being used for at this point? Can we remove it, given that you've added the static 180x180 icon and that's the only one referenced in the `<link>`?

saurabhdaware: the one added in `<link>` is for apple. Apparently, some versions of apple do not read icons from webapp manifest so they need a special tag in html head but Chrome and other browsers use webapp manifest for icons.

So we'll still need `icon` for chrome

bvaughn: > So we'll still need `icon` for chrome

Ah, is this the icon shown as the splash screen when the app is opening?

TBH I'm not sure if this change is a net positive for Chrome. It's nice that the default title is just "React", and the splash screen is nice- but the "Add to homescreen" action takes a couple of seconds (during which I see a loading spinner) now instead of just being instant (like it is on reactjs.org, without this change) -

Screenshot_20190930-140116_Chrome

saurabhdaware: `icon` takes that image and generates multiple images so `512x512` is used in homescreen and `192x192` is usually used to create that logo on android and other sizes are for different screen sizes I guess.

I am not sure but I think the reason logo loads fast without `icon` is that it uses `favicon.ico` which is already loaded when you open website but when you specifically define icon it probably loads when you click 'Add to Homescreen'.

Also, we can't really rely on favicon to provide logo since it does not work in some mobiles (does not work in mine at least)

bvaughn: I don't really know much about this particular stuff. I'd love for someone else to weigh in.

saurabhdaware: Cool, thank you for the quick response :)

saurabhdaware:
Hi @alexkrolick , Can you (or request someone to) review this please? thanks.

saurabhdaware:
This is so cool! thank you so much 🌻

twhite96:
Hey! There is one plugin missing for this to work properly as a PWA. I've added it: `gatsby-plugin-offline`. Without it, you cannot install the PWA on Chrome desktop or iOS, etc. 

I went to the React website and looked for the little + install button in the Omnibox and it wasn't there. Adding this should make it work as it should. @saurabhdaware thanks for doing this! I wish I had thought of it. 😄

saurabhdaware:
Hi @twhite96  Since it was my first PR I wanted to keep things simple so I didn't add it but yes , it would be really cool to have service worker to make it work offline! Good luck with the PR :D

PR #2385, opened 2019-10-02
Use gist URLs instead of RawGit in React docs

davidjb:
RawGit is being shut down (eg Oct 2019, https://rawgit.com/) and this change uses the direct links to the files as served from gist.github.com.  The same files are being served, just that the links are only dependent on GitHub's gist service.


alexkrolick:
Thanks

PR #2391, opened 2019-10-03
Update site-constants

murdochd:
At the time of writing, the latest version of react is v16.10.1


lex111:
@murdochd good catch, thank you!

PR #2395, opened 2019-10-04
Fix Typo in Lifting State Up Docs

conradbeach:
There's a line in the document that says:
"When it previously rendered, the `Calculator` has specified that"

The sentence is using a present tense verb ("has"), but the statement is in the past tense.

I think "When it previously rendered, the `Calculator` had specified that" would be correct.


threepointone:
good catch, thank you!

PR #2401, opened 2019-10-06
Fix broken test code

### LINKED ISSUE #2386: [Docs] Testing Recipes: broken test code
Hi, I think there's a little bug in the example code in the Testing Recipes section.

https://reactjs.org/docs/testing-recipes.html#events

When I copy the code for `toggle.js` and `toggle.test.js` into a fresh sample project and run `npm run test`, I get a failed test.

It looks like the expectations are backwards vs. the initial state in the `toggle` component.

altyaper:
I notices that since the state initially is false, the label should start with `turn off` instead of `turn on`, I change that, and also I did change the final expected state for the button, which if you click it 6 times it will be off again.
```
"turn off" - 0 - INITIAL STATE
"turn on" - 1
"turn off" - 2
"turn on" - 3
"turn off" - 4
"turn on" - 5
"turn off" - 6
```

Fix #2386

### DIFF
--- a/content/docs/testing-recipes.md
+++ b/content/docs/testing-recipes.md
@@ -356,7 +356,7 @@ export default function Toggle(props) {
       }}
       data-testid="toggle"
     >
-      {state === true ? "Turn off" : "Turn on"}
+      {state === true ? "Turn on" : "Turn off"}
     </button>
   );
 }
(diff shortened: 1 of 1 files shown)


alexkrolick:
In this case shouldn't the fix be to change the initial value rather than the meaning of the label?

echavezNS:
The solution can be both, I think the initial value should be `false`, it's more intuitive I guess. But I can change the initial value instead if you want @alexkrolick

echavezNS:
any update about this @alexkrolick ?

BartoszKlonowski:
Hello @altyaper, sorry for the late response!
I couldn't find that certain part of docs you update in the latest version of it, so I will close this PR as outdated.
Thank you for taking the time to contribute!

PR #2402, opened 2019-10-06
Update hooks-effect.md

imran-mind:


alexkrolick:
"stale" is correct, it means out of date

PR #2404, opened 2019-10-06
Update examples.md

GermaVinsmoke:


GermaVinsmoke:
@hg-pyun If you can suggest something better then I'll change it to that 😁

gaearon:
Thanks but we're trying to keep examples focused on vanilla React.

PR #2405, opened 2019-10-07
Add tipp on dependency arrays for custom hooks

clentfort:
The tipp is based on this tweet by Dan https://twitter.com/dan_abramov/status/1140316424898043904

### DIFF
--- a/content/docs/hooks-custom.md
+++ b/content/docs/hooks-custom.md
@@ -110,9 +110,9 @@ function useFriendStatus(friendID) {
 }
 ```
 
->Tip
+> **Note**
 >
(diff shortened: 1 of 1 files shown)


sophiebits:
This is definitely not the right place in the docs for it. It is too early. This section is someone's very first exposure to the idea of a custom Hook. They probably don't even remember what a dependency array is.

clentfort:
I agree, but I didn't find any other places that felt right for this kind of information. Maybe an extra entry inside a new section of the Hooks FAQ? 

Scrolling through the FAQs I found this entry https://deploy-preview-2405--reactjs.netlify.com/docs/hooks-faq.html#how-to-read-an-often-changing-value-from-usecallback which uses a custom deps array. Should this entry be updated to not use this pattern? I'm not sure how this should look. Maybe a warning would be enough?

PR #2406, opened 2019-10-07
Fix variables name inconsistency in jumpTo()

voskresla:
I think it will be better to understand if we change `const moves = history.map((step, move) => {...})` to `const moves = history.map((_,step) => {...})` because we don't use `step` variable in map method, but later we use `jumpTo(step)` and it's confusing.

(I do not know if I can fix the code on codepen?)

### DIFF
--- a/content/tutorial/tutorial.md
+++ b/content/tutorial/tutorial.md
@@ -999,7 +999,7 @@ const doubled = numbers.map(x => x * 2); // [2, 4, 6]
 
 Using the `map` method, we can map our history of moves to React elements representing buttons on the screen, and display a list of buttons to "jump" to past moves.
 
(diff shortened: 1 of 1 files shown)


content/tutorial/tutorial.md
```diff
@@ -1091,16 +1091,16 @@ Keys do not need to be globally unique; they only need to be unique between comp
...
 ```js{6}
-    const moves = history.map((step, move) => {
-      const desc = move ?
-        'Go to move #' + move :
+    const moves = history.map((_, step) => {
+      const desc = step ?
+        'Go to move #' + step :
         'Go to game start';
       return (
         <li key={move}>
```
voskresla:
```suggestion
        <li key={step}>
```

PR #2409, opened 2019-10-08
media query to hide header on print

### LINKED ISSUE #2322: When printing the Tutorial... the top line of each page is missing.
When in Chrome and you print the Tutorial (ctrl-P), the top line of each page is missing (hidden under the page header).  Can this be fixed?

flurmbo:
Fixes #2322 by adding a print media query to the Header component which hides it. This way, the first line of every page doesn't get covered up


flurmbo:
@bvaughn that wasn't causing the initial bug but a quick survey of a couple websites (wikipedia and vue js docs) suggests it's a common thing to do, and it would save some paper. I can add a commit for that as well

PR #2413, opened 2019-10-09
Adding ReactJS PH

franreysaycon:
We have recently formalized a group here in the Philippines with the aim to have monthly meetups, study jams, and eventually(hopefully) a conference. We are on our 5th meetup for the year now! ever since we started the core team 5-6 months ago.


alexkrolick:
👍 if you can, it would be good to clean up the old group in a followup if you confirm it is no longer active

PR #2431, opened 2019-10-14
Update composition-vs-inheritance.md

sebastian-sugarai:


threepointone:
This isn't right, the docs are correct. Thank you for the effort!

PR #2433, opened 2019-10-14
Fix typo on React 16 Roadmap page

bobziroll:
I believe it was meant to be "news feed story", but I could be misunderstanding something in the context of the paragraph.


threepointone:
Thank you!

PR #2438, opened 2019-10-15
Language redirect feature and fixed Search bar overlapping

nikhilgupta30:
This pr resolves 2 issues

first if a user has opened a particular page and then clicks on languages and select one it redirects to home page of that language so user has to search again for previous page. Now if user clicks on language and select any language user will be redirected to same path of selected language.
for https://github.com/reactjs/reactjs.org/issues/2324

Second it fixes search bar overlapping problem in navbar
for https://github.com/reactjs/reactjs.org/issues/2327


BartoszKlonowski:
Closing as per comments in the issues.
Thank you!

PR #2440, opened 2019-10-16
FrameworkTraining/reactjs.org/content/community/courses.md

FrameworkTraining:
We would appreciate it if our React course could be added. Thanks, Tom, Framework Training


FrameworkTraining:
Success!

alexkrolick:
@FrameworkTraining the other courses listed are online-only/video - is that an option for this one?

FrameworkTraining:
@alexkrolick 

Hi Alex, thanks for following up - we're very much advocates of instructor-led face-to-face training - remote attendance is also possible but we really prefer having an expert in the room with the students.

PR #2441, opened 2019-10-16
Revise useState hook docs and examples to account for batched state updates

ndresx:
I recently went through the Hook docs to see how they work and was a little bit confused that a counter was used to demonstrate the `useState` hook, not because of the counter, but because it was used without callback function.

I think it's important that these few intro example(s) convey the correct usage which why I have replaced the counter examples with a simple random number generator.

```js
function Example() {
  // Declare a new state variable, which we'll call "randomNumber"
  const [randomNumber, setRandomNumber] = useState(0);

  return (
   <div>
     <p>Random number: {randomNumber}</p>
     <button onClick={() => setRandomNumber(Math.random())}>
       Randomize
     </button>
   </div>
 );
}
```

It should still be easy to understand and doesn't speak of batched state updates that especially depend on the previous state in these early documentation stages yet (just mentions it at the very bottom like the `setState` page is doing it).

However, I've left the counters untouched where it makes sense (especially when the API gets explained in more detail).

### DIFF
--- a/content/blog/2018-11-27-react-16-roadmap.md
+++ b/content/blog/2018-11-27-react-16-roadmap.md
@@ -78,17 +78,17 @@ Code splitting is just the first step for Suspense. Our longer term vision for S
 
 ```js
 function Example() {
-  // Declare a new state variable, which we'll call "count"
-  const [count, setCount] = useState(0);
+  // Declare a new state variable, which we'll call "randomNumber"
+  const [randomNumber, setRandomNumber] = useState(0);
 
   return (
-   <div>
-     <p>You clicked {count} times</p>
-     <button onClick={() => setCount(count + 1)}>
-       Click me
-     </button>
-   </div>
- );
+    <div>
+      <p>Random number: {randomNumber}</p>
+      <button onClick={() => setRandomNumber(Math.random())}>
+        Randomize
+      </button>
+    </div>
+  );
 }
 ```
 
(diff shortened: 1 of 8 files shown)


gaearon:
The focus of these docs is to make the first introduction crystal clear. Batching is a nuance that most people reading them don’t think about until many months in. So I don’t think we should be sacrificing the clarity here.

I don’t think in example with random numbers the purpose of state is clear. People will think: why not just generate a random number during rendering? It is distracting from the main concept so I don’t think this is a right direction.

ndresx:
> Batching is a nuance that most people reading them don’t think about until many months in. 

That was one of the reasons why I wanted to do it. There are people that haven't even heard of this possibility and while it will work out fine for the most as you said, some will never find out that there's a potential problem in their code, or will randomly stumble upon it when it's maybe already too late.

> I don’t think in example with random numbers the purpose of state is clear.

I agree that the random number example is not the best to showcase this feature, but it's the most basic/interactive one I could think of without having to sacrifice the correctness as well as the simplicity of the example. I was also thinking about showing the current time, but it's not as interesting without a dynamic timer in the background like the `setState` examples are using it. _Maybe there's another one?_

Would you still be interested in keeping the tip at the bottom to mention this asynchronous behavior, or the changes in https://github.com/reactjs/reactjs.org/pull/2441/files#diff-c38ef803fe7e7fc860f577c958440873?

PR #2444, opened 2019-10-17
Remove mention of state only in class component

obedparla:
The phrase `just like you do when you need lifecycle methods or state.` sounds like only Class components have state or lifecycles, which is not true with the introduction of Hooks. It can introduce confusion for people that know about Hooks, better to not mention it at all.

### DIFF
--- a/content/docs/refs-and-the-dom.md
+++ b/content/docs/refs-and-the-dom.md
@@ -161,7 +161,7 @@ class Parent extends React.Component {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


eps1lon (changes requested):
With the current flow of the docs the current sentence makes sense because it refers to prior parts of the docs where we still say that for state or lifecycles you'll rewrite to a class component.

I think this will be resolved after the big docs rewrite Dan and Rachel are doing.

Anyway, this PR only makes sense with a big picture approach.

obedparla:
Makes sense!

PR #2448, opened 2019-10-19
Add npm command in 'Getting started' topic in README.md

neung123:
Add npm command in 'Getting started' topic in README.md

1. Add `npm i` or `npm install` instead of only `yarn` 
2. Add `npm run dev` instead of only `yarn dev`


threepointone:
We only use `yarn` because of its lockfile support. It might very well still work with `npm`, but we don't support that at the moment. Thank you for the PR!

PR #2449, opened 2019-10-21
Documentation

arshita-kakkar:
Line 41 - 
When starting a React project, [a simple HTML page with script tags](/docs/add-react-to-a-website.html) might still be the best option. It only takes a minute to set up!
Removed the hyperlink from "a" as we're focusing on simple HTML page so a shouldn't be hyperlinked.


lex111:
@arshita-kakkar thanks.

PR #2454, opened 2019-10-21
Move {...props} at the end of Component

HardLight00:
As I had read in React docs about HOCs approach when {...props} stay at the end of <Component /> props is better because it avoids redeclaration of original props in <Component /> and unexpected behavior

### DIFF
--- a/content/docs/render-props.md
+++ b/content/docs/render-props.md
@@ -225,7 +225,7 @@ function withMouse(Component) {
     render() {
       return (
         <Mouse render={mouse => (
-          <Component {...this.props} mouse={mouse} />
+          <Component mouse={mouse} {...this.props} />
         )}/>
       );
     }


content/docs/render-props.md
```diff
@@ -225,7 +225,7 @@ function withMouse(Component) {
     render() {
       return (
         <Mouse render={mouse => (
-          <Component {...this.props} mouse={mouse} />
+          <Component mouse={mouse} {...this.props} />
```
alexkrolick: Leaving the `{...props}` before the `mouse={mouse}` ensures that passing a `mouse` prop can't override what the render prop is trying to pass to the component.

HardLight00: I am a bit confused. I thought that if `mouse={mouse}` stay after `{...props}` contrariwise can override attributes in primordial props

alexkrolick: that's what we want, the render prop should override whatever is passed in

PR #2464, opened 2019-10-24
Minor typo fix in concurrent mode overview.

Shelob9:
There is a y missing in the word "displaying"


lex111:
@Shelob9 thank you!

PR #2472, opened 2019-10-24
Fix minor typo in concurrent mode adoption docs

mattfwood:


lex111:
@mattfwood thank you!

PR #2473, opened 2019-10-24
Fix minor typo in concurrent mode reference docs

mattfwood:


lex111:
@mattfwood thank you!

PR #2486, opened 2019-10-25
add example project radix converter

kumom:
This is the same PR as #2468 which becomes orphaned because I accidentally deleted my forked repo.


lex111:
@kumom thx!

PR #2488, opened 2019-10-25
added new stackoveflow app

Chandra6160:


gaearon:
Hi, this is a good idea but the code formatting is very inconsistent and there's a lot of copy paste between components. So it's not a very good example of using React components. Maybe if you tidy it up we can take another look.

PR #2490, opened 2019-10-25
Missing highlighted line in concurrent mode patterns example

Drakota:
In the Concurrent UI Patterns article section "Reviewing the Changes", the text was saying "It took us only seven lines of code to add this transition" even though there was only 6 lines highlighted.


lex111:
@Drakota  I think you're right, merged this PR, thanks!

PR #2494, opened 2019-10-25
Update meetups.md adding Nantes (France ) meetup

LoicGoyet:
Add Nantes, France meetup group


lex111:
@LoicGoyet thank you!

PR #2495, opened 2019-10-26
Update concurrent-mode-suspense.md

wolverineks:
According to [ProofreadNOW](https://www.proofreadnow.com/blog/bid/101485/If-I-Were-or-If-I-Was-Which-is-Correct),
> Guideline: Use were (instead of was) in statements that are contrary to fact.


threepointone:
Thank you!

PR #2496, opened 2019-10-26
added easy to use reactjs timer stopwatch

mohamedhassanmn:
code and features Updated


gaearon:
I think I already reviewed a PR for this one before

PR #2498, opened 2019-10-26
Fix typo on concurrent mode api

hirofumii:
fowards -> forwards


lex111:
@hirofumii good catch, thank you!

PR #2502, opened 2019-10-27
Add wix-style-react to Free Components list

yershalom:
Add wix-style-react to Free Components list


lex111:
@yershalom thank you!

PR #2503, opened 2019-10-27
Update Conferences

TejasQ:
ReactConf just finished, so making a quick update to the upcoming conference list. 💞I also updated the list of upcoming conferences to be in ascending order relative to today.


alexkrolick:
Thanks!

PR #2506, opened 2019-10-27
Update tools-ui-components.md, Added MTCaptcha React Component

batou-mtcapthca:
Added MTCaptcha React Component


gaearon:
We removed this page

PR #2508, opened 2019-10-28
Update example code: set state by passing function

gossie:
I went through the tic tac toe tutorial to refresh my react knowledge. Shouldn't updates to the state, that depend on the previous state, be performed by passing a function to setState()?


Daniel15:
@gaearon Any issues with this PR? It looks good to me. Otherwise I'll merge it. 😃

PR #2519, opened 2019-10-30
Add experimental release link to Concurrent Mode Intro

gnbaron:


gnbaron:
Actually, since the adoption link is right above, I'm not sure whether adding another one is worth it. Feel free to drop it if that's the case.

That's my bad, I've instantly looked for the version label and could find it just on the release channels page, while it was right on the next sections.

PR #2520, opened 2019-10-30
updates the example code blocks

amirsaeed671:
there's missing div tags on the code blocks example ........ which I've added


content/docs/hooks-reference.md
```diff
@@ -57,12 +57,12 @@ If the new state is computed using the previous state, you can pass a function t
 function Counter({initialCount}) {
   const [count, setCount] = useState(initialCount);
   return (
-    <>
+    <div>
```
alexkrolick: `<>` is a shorthand syntax for React.Fragment

PR #2521, opened 2019-10-30
Add Detroit React User Group to list of meetups

eruby94:
Just made the group back in June -- its still fairly small, but hopefully listing here will help improve awareness and increase membership.


lex111:
@eruby94 cool, thank you!

PR #2524, opened 2019-10-31
Updated hooks reference

JaffParker:
Added a note about throwing an error in default switch case in `useReducer`, as per Dan's [tweet](https://twitter.com/dan_abramov/status/1093969005675773954) and [this issue](https://github.com/reactjs/reactjs.org/issues/2159)

### DIFF
--- a/content/docs/hooks-reference.md
+++ b/content/docs/hooks-reference.md
@@ -286,6 +286,10 @@ function Counter() {
 >
 >React guarantees that `dispatch` function identity is stable and won't change on re-renders. This is why it's safe to omit from the `useEffect` or `useCallback` dependency list.
 
+>Note
+>
+>Unlike reducers in Redux or similar state managing solution, actions in `useReducer` are specific to a component. Hence, it is recommended that you **throw an error in the default switch case**. It can help you track bugs in early stages (for example, mistakes in action references).
+
 #### Specifying the initial state {#specifying-the-initial-state}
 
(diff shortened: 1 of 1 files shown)


content/docs/hooks-reference.md
```diff
@@ -286,6 +286,10 @@ function Counter() {
 >
 >React guarantees that `dispatch` function identity is stable and won't change on re-renders. This is why it's safe to omit from the `useEffect` or `useCallback` dependency list.
 
+>Note
+>
+>Unlike reducers in Redux or similar state managing solution, actions in `useReducer` are specific to a component. Hence, it is recommended that you **throw an error in the default switch case**. It can help you track bugs in early stages (for example, mistakes in action references).
```
alexkrolick: Switch case is one way to write a reducer, but there are others, like an object literal.

```suggestion
Unlike reducers in Redux or similar state managing solution, actions in `useReducer` are specific to a component. Hence, it is recommended that you **treat unknown action types as errors** rather than handling them with a default value. It can help you track bugs in early stages (for example, mistakes in action references).
```

alexkrolick: Also, might be work mentioning how Flow and Typescript can help prevent dispatching unknown action types.

JaffParker: I find React docs doesn't make many references to Typescript, they stay pretty agnostic (although I do agree). Do you still think it's a good idea to add it? I could work it in.

alexkrolick: There are a few and this area is one where a type system can specifically help either an unhandled state or runtime error.

PR #2532, opened 2019-11-05
Update code-splitting note that import() is part of ES2020

hycner:
This updates text in the code-splitting section indicating that the `import()` syntax has finished the proposal process and is a part of the ES2020 specification.

- [Specification](https://tc39.es/ecma262/) that lists in multiple sections if you search for `import()` on the page
- The [v8 site](https://v8.dev/features/tags/es2020) also independently lists as part of ES2020

### DIFF
--- a/content/docs/code-splitting.md
+++ b/content/docs/code-splitting.md
@@ -96,10 +96,8 @@ import("./math").then(math => {
 
 > Note:
 >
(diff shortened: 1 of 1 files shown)


threepointone:
Looks like we've removed the line altogether in the docs. Thank you for the PR!

PR #2535, opened 2019-11-05
Remove the Tools doc section

gaearon:
We haven't really been maintaining this section. It's full of old unmaintained projects, and random recommendations with no editorial oversight. Originally it was migrated from the wiki. It's also barely visited. So I'm just removing it.

Additionally I'm removing some pages from localization effort many years ago. They're very stale and aren't linked to from anywhere.


alexkrolick:
Yep. There's hardly any oversight on what's in there.

PR #2537, opened 2019-11-06
Update static-type-checking.md

mos-adebayo:
Fix yarn command to install flow as dev dependency


alexkrolick:
Works either way! Flags usually go before arguments.

PR #2539, opened 2019-11-06
Fixing some typos

altano:
I'm actually not even sure your `synchronous` typo is actually a typo, but calling a function that isn't literally `async` that immediately returns an object with nested, unresolved `Promises` doesn't seem to convey the right idea, so I'm assuming it was a typo?


altano:
cc @josephsavona

josephsavona:
Thanks for asking, but this is intentional. When used to describe functions in JavaScript, "asynchronous" usually means either a function that either returns a `Promise` or accepts a callback that is *not* called in the same tick of the event loop. The idea of the `prepare()` function  described in the post is that it *should* immediately return a non-Promise value that is **synchronous**. Certainly the work started by preloading *may* finish asynchronously, but it could also finish synchronously, and either way it immediately returns a value that we can render with.

This is an important distinction, so we'd prefer to keep the wording as-is.

Issue #2540, opened 2019-11-07
Sample code question on "Building Great User Experiences with Concurrent Mode and Suspense" blog post


swimyoung:
Thanks for a great blog post.

I have a question about a code sample in [this chapter](https://reactjs.org/blog/2019/11/06/building-great-user-experiences-with-concurrent-mode-and-suspense.html#load-data-incrementally).
what is `post` in `<CommentList post={post} />`? 
Is it `postData` or `props.post`?

![image](https://user-images.githubusercontent.com/8675529/68356164-e93a1580-0154-11ea-9364-0f763e96727e.png)

swimyoung:
answered and merged at #2541

PR #2541, opened 2019-11-07
Fix sample code variable name on "Building Great User Experiences with Concurrent Mode and Suspense" blog post

swimyoung:
#2540 
I think `post` is `props.post`


swimyoung:
Thank you for comment :)
I've updated

PR #2547, opened 2019-11-08
Update components docs for functional components

mattfwood:
I noticed in the "Main Concepts" section under components, it mentions class components have "additional features" compared to function components. I reworded this slightly since hooks allow function components to (mostly) have parity with class components and it might be confusing later.

I searched for a bit and wasn't able to find a larger plan to update these older docs to clarify things like this, or if there's a larger effort to include hook-based examples alongside class component ones.

I know it might still be up for debate, but I do recall the React core team saying they foresee hooks being the primary way to learn React going forward.

### DIFF
--- a/content/docs/components-and-props.md
+++ b/content/docs/components-and-props.md
@@ -44,7 +44,7 @@ class Welcome extends React.Component {
 
 The above two components are equivalent from React's point of view.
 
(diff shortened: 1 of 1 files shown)


threepointone:
Thank you very much!

PR #2557, opened 2019-11-12
Update examples.md

hassankhalid1:


gaearon:
Thanks but we're trying to keep examples focused on vanilla React.

PR #2560, opened 2019-11-12
Minor correction in function component example

sellandb:
Added parenthesis to function call in function component parameter.


sellandb:
Noticed that I was wrong!

PR #2562, opened 2019-11-12
Update proposed solution for cancellable promises

filipemir:
As discussed in [this thread](https://github.com/facebook/react/issues/5465#issuecomment-339544623) and expanded [here](https://github.com/hjylewis/trashable/blob/master/PROOF.md), the solution proposed for cancelling promises when a component unmounts doesn't actually address possible memory leaks. Interestingly, I don't believe @hjylewis's approach does so fully either (see [this issue](https://github.com/hjylewis/trashable/issues/6)) but it does so in the vast majority of cases. If I'm following the conversation along correctly I think the docs should be updated to reflect the better answer. And if his is not the better answer I would love to hear the arguments against it.

### DIFF
--- a/content/blog/2015-12-16-ismounted-antipattern.md
+++ b/content/blog/2015-12-16-ismounted-antipattern.md
@@ -47,33 +47,41 @@ const cancelablePromise = makeCancelable(
 );
 
 cancelablePromise
-  .promise
   .then(() => console.log('resolved'))
   .catch((reason) => console.log('isCanceled', reason.isCanceled));
 
 cancelablePromise.cancel(); // Cancel the promise
 ```
 
-Where `makeCancelable` was originally [defined by @istarkov](https://github.com/facebook/react/issues/5465#issuecomment-157888325) as:
+Where `makeCancelable` was originally [defined by @hjylewis](https://github.com/facebook/react/issues/5465#issuecomment-339544623) as:
 
 ```js
 const makeCancelable = (promise) => {
-  let hasCanceled_ = false;
+    let cancel = () => {};
 
-  const wrappedPromise = new Promise((resolve, reject) => {
-    promise.then(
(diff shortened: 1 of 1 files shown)


ekilah:
I've followed the discussion that has taken place over several years and ended up here. It seems odd that there isn't more discussion about real garbage collection here / on https://github.com/facebook/react/issues/5465 , which (in the absence of replies / changes to the docs like this PR wants to make) seems to implicitly suggest that the React team disagrees with the logic that is presented at the end of 5465 re: garbage collection, e.g. [trashable](https://github.com/hjylewis/trashable). 

Is that the case, or is this just mostly a case of bit rot on the 2015 blog post? The blog post is still referenced a lot (presumably, certainly I have references to it in my codebases), so if its solution is wrong/not fully efficient it seems like it should be fixed, but this PR to do that is pretty stale / discussionless.

I don't know much about how GC in Javascript works, but it seems obvious to me from a CS-fundamentals POV that if `this` is referenced in a `Promise`s `then`/`catch`, then canceling the way the blog post suggests won't actually do anything re: memory savings until the promise resolves/rejects on its own. [trashable's proof doc](https://github.com/hjylewis/trashable/blob/master/PROOF.md) goes into this more, though again, I am not familiar with the inner workings w/ JS here enough to express my own confidence past reading this and and agreeing with its hypothesis. 

Since it's been quite a while, I hope no one minds me tagging some of the original authors / debaters on the original issue to see if this gains any traction.

@jimfb from https://github.com/reactjs/reactjs.org/commit/eeaf72f9aa4f9b12e4d33f9a2dcd495453c81aad
@gaearon from https://github.com/reactjs/reactjs.org/commit/35bb6edcbb924029f4fcd4f7ee74bc10a0662eaa
@benmmurphy and @istarkov from https://github.com/facebook/react/issues/5465
@hjylewis from [trashable](https://github.com/hjylewis/trashable)

PR #2564, opened 2019-11-13
Remove dynamic import proposal text

jkjustjoshing:
Dynamic imports have reached stage 4 and are now part of ECMAScript.


lex111:
@jkjustjoshing seems legit, thanks!

PR #2567, opened 2019-11-14
Update handling-events.md

HowMade:


content/docs/handling-events.md
```diff
@@ -140,6 +140,25 @@ class LoggingButton extends React.Component {
...
+  render() {
+    // This syntax ensures `this` is bound within handleClick
+    return (
+      <button onClick={this.handleClick.bind(this)}>
+        Click me
+      </button>
+    );
+  }
+}
+```
```
HowMade: And the fourth approach is that you can use [.bind] in callback:

```js{9}
class LoggingButton extends React.Component {
  handleClick() {
    console.log('this is:', this);
  }
  render() {
    // This syntax ensures `this` is bound within handleClick
    return (
      <button onClick={this.handleClick.bind(this)}>
        Click me
      </button>
    );
  }
}
```

PR #2573, opened 2019-11-17
Update testing-recipes.md

yosuheee:
- correcting assertion mistakes

With previous code, the test fails.


threepointone:
Thank you, nice catch!

PR #2574, opened 2019-11-18
Add React Loop 2020

joemaddalone:
This PR adds React Loop 2020 to the conference list


lex111:
@joemaddalone thank you!

PR #2577, opened 2019-11-19
Update fragments.md

yudium:
This is confusing for beginner like me if not give the knowledge about JSX behavior.

### DIFF
--- a/content/docs/fragments.md
+++ b/content/docs/fragments.md
@@ -38,7 +38,7 @@ class Table extends React.Component {
 }
 ```
 
-`<Columns />` would need to return multiple `<td>` elements in order for the rendered HTML to be valid. If a parent div was used inside the `render()` of `<Columns />`, then the resulting HTML will be invalid.
+`<Columns />` would need to return multiple `<td>` elements in order for the rendered HTML to be valid. If a parent div was used inside the `render()` of `<Columns />` (note that we add `<div>` because JSX needs us have a parent element, otherwise it will throw SyntaxError), then the resulting HTML will be invalid.
 
 ```jsx
 class Columns extends React.Component {


content/docs/fragments.md
```diff
@@ -38,7 +38,7 @@ class Table extends React.Component {
 }
 ```
 
-`<Columns />` would need to return multiple `<td>` elements in order for the rendered HTML to be valid. If a parent div was used inside the `render()` of `<Columns />`, then the resulting HTML will be invalid.
+`<Columns />` would need to return multiple `<td>` elements in order for the rendered HTML to be valid. If a parent div was used inside the `render()` of `<Columns />` (note that we add `<div>` because JSX needs us have a parent element, otherwise it will throw SyntaxError), then the resulting HTML will be invalid.
```
BartoszKlonowski, line 41:
```suggestion
`<Columns />` would need to return multiple `<td>` elements in order for the rendered HTML to be valid. If a parent div was used inside the `render()` of `<Columns />` (note that we add `<div>` because JSX needs us to have a parent element, otherwise it will throw SyntaxError), then the resulting HTML will be invalid.
```

BartoszKlonowski (changes requested):
Hello @yudium!
I understand that sometimes parts of the docs, or articles may be found confusing or difficult to understand when going through that topic for the first time, or without a prerequisite knowledge. Thing is that in such cases, the best practice in docs is to add a link (in a "See also", "Refer to", or "Read more" etc.) to that prerequisite or another part of docs explaining the mentioned keyword or topic in more details.
Adding an inline, still brief, and not always sufficient, explanation in brackets is unfortunately not really the way to go.
Please consider the above as a hint for future contributions to the docs.

From what I can see that part of the Fragments-related article is no longer present in the docs, so I will close this PR as outdated.
However, please create new PRs to other open [issues](https://github.com/reactjs/react.dev/issues) if you would still like to contribute.

Thank you!

PR #2578, opened 2019-11-20
Add gatsby-plugin-offline for offline PWA

twhite96:
I wasn't able to fully install the PWA referenced in this issue https://github.com/reactjs/reactjs.org/issues/2259#issue-481864203 on Chrome desktop so I added `gatsby-plugin-offline`.

In Chrome, if an app is installable, it will have a plus button in the Omnibox. The current docs are not installable but should be after installing this plugin.

I was unable to verify as service workers are only able to be resolved on secure origins which localhost is not.

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -159,6 +159,7 @@ module.exports = {
     },
     'gatsby-plugin-react-helmet',
     'gatsby-plugin-catch-links',
+    `gatsby-plugin-offline`,
     {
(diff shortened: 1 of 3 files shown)


twhite96:
This is working in the preview as expected.


There seems to be a failing test however I am not sure why.

BartoszKlonowski:
Hello @twhite96!
As the Gatsby is no longer used in the repo I will close this PR as no longer required.
Thank you!

PR #2584, opened 2019-11-23
add dependency array to examples

donavon:
The `usePrevious` example is missing a dependency on `value`.

### DIFF
--- a/content/docs/hooks-faq.md
+++ b/content/docs/hooks-faq.md
@@ -342,7 +342,7 @@ function Counter() {
   const prevCountRef = useRef();
   useEffect(() => {
     prevCountRef.current = count;
-  });
+  }, [count]);
   const prevCount = prevCountRef.current;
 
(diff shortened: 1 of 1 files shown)


donavon:
BTW, This was prompted by this Twitter thread.

https://twitter.com/GabeRicard/status/1198316044957421568

rimunroe:
This hook shouldn’t need a dependency array, and I don’t think we want people adding dependency arrays when they don’t need to.

RebekahEloise:
The useEffect hook takes 2 arguments, a callback, and a dependency array. Therefore, it seems reasonable to include a dependency array for these two instances of useEffect.

BartoszKlonowski:
Hello @donavon, thanks for taking the time to contribute!
I don't see this example anymore in the docs, so I will close this PR as no longer required, but feel invited to create new PRs to other open [issues](https://github.com/reactjs/react.dev/issues).

PR #2586, opened 2019-11-24
Added a Hooks implementation of the example

rebot:
Added a Hooks implementation because the same logic applies there, but users who want to implement it using Hooks might get stuck and an worked out example would definitely help in those cases.


gaearon:
Thanks, but we'll need to rewrite the whole tutorial instead of adding ad-hoc links.

PR #2588, opened 2019-11-25
Fix usePrevious

VdustR:
Fix initial value

The initial value should not be `undefined` at start to prevent triggering the `useEffect`.

Please check the demo: <https://codesandbox.io/s/6rrey>


VdustR:
[@react-hook/previous](https://www.npmjs.com/package/@react-hook/previous) is great enough for me.

PR #2589, opened 2019-11-26
Added React Supers Example App

shaz13:
Added example React App.
- Uses state management
- Components best practises
- Updating state on search query. 

A good starter projects for the beginners to understand React Concepts.


gaearon:
The link needs to point to the code. The code doesn't look polished: there's commented out lines, the formatting is inconsistent. The images are missing / don't always load (API issue?) Please feel free to submit again if you polish it.

PR #2590, opened 2019-11-26
docs: Unify identifiers in "useImperativeHandle" example

pixelass:
The example was using `inputRef` while the description used `fancyInputRef`. This was irritating.


lex111:
@pixelass thank you!

PR #2596, opened 2019-11-28
Docs: Update wrong assignment in Test Renderer reference example

ntban15:
The existing code reassigns `root` on `root.update`, although `update` does not return the current instance of test renderer as seen [here](https://github.com/facebook/react/blob/969f4b5bb8302afb3eb1656784130651047c3718/packages/react-test-renderer/src/ReactTestRenderer.js#L497-L502).


threepointone:
Updated in a separate PR. Thank you!

PR #2598, opened 2019-11-28
Fix typo in useReducer docs

mxstbr:


content/docs/hooks-reference.md
```diff
@@ -299,7 +299,7 @@ There are two different ways to initialize `useReducer` state. You may choose ei
 
 >Note
 >
->React doesn’t use the `state = initialState` argument convention popularized by Redux. The initial value sometimes needs to depend on props and so is specified from the Hook call instead. If you feel strongly about this, you can call `useReducer(reducer, undefined, reducer)` to emulate the Redux behavior, but it's not encouraged.
+>React doesn’t use the `state = initialState` argument convention popularized by Redux. The initial value sometimes needs to depend on props and so is specified from the Hook call instead. If you feel strongly about this, you can call `useReducer(reducer, undefined)` to emulate the Redux behavior, but it's not encouraged.
```
giuseppeg: @mxstbr sorry, I am not stalking you :D this just popped in my gh feed :)

I think the docs are correct but make the assumption that the `reducer` returns `state` (unchanged) instead of throwing an error.

The 3rd parameter behaves as an initializer function.

```js
const initialState = { name: "max" };
function reducer(state = initialState, action = {}) {
  switch (action.type) {
    default:
      return state;
  }
}

function App() {
  const [state, dispatch] = React.useReducer(reducer, undefined, reducer);
 // ...
}
```

The odd thing for the Redux folks who use the `switch` would be that they'd have to specify a default `action` otherwise `action.type` will throw an error (something like cannot access `type` of `undefined`).

BartoszKlonowski:
Hello @mxstbr!
I'm sorry, but that part you update is no longer in the latest version of docs, so I will close this PR as outdated.

PR #2618, opened 2019-12-17
Fix React events package link in Codebase Overview

Simek:
Currently on the Codebase Overview page link to the React events package uses old package name so it leads to 404 page on GitHub. This PR fixes that issue.


lex111:
@Simek thank you!

PR #2621, opened 2019-12-18
Correcting lifting-state-up.md

dhrobinson:
Water boils at temperatures *over* 100 degrees Celsius.


threepointone:
This feels like nitpicking :D Changing this would also require use to change the gif. I'm inclined to reject this PR, even though you're right. Thank you for the PR!

PR #2641, opened 2019-12-24
Update state-and-lifecycle.md

vnoitkumar:


threepointone:
This seems like a nit, so I'm inclined to reject this. Thank you for the effort!

PR #2643, opened 2019-12-27
prefer https to http for links

koba04:
These links are being provided as `https` not only `http`.
So it would be better to use `https`.


lex111:
@koba04 thank you!

koba04:
Thank you!

PR #2647, opened 2019-12-28
Update hooks-rules.md

gergelypap:
Updated information on the ESLint plugin being included in Create React App.


lex111:
@gergelypap thanks!

PR #2648, opened 2019-12-29
Fix sample test code

numb86:
Since the initial value of `state` is `false`, the first displayed text is `Turn on`.
I actually wrote the code and confirmed it.


sophiebits:
Thanks!

PR #2649, opened 2019-12-29
Fix typo in hooks-faq.md

mknapper1:

### DIFF
--- a/content/docs/hooks-faq.md
+++ b/content/docs/hooks-faq.md
@@ -578,7 +578,7 @@ Depending on your use case, there are a few more options described below.
 
 Let's see why this matters.
 
-If you specify a [list of dependencies](/docs/hooks-reference.html#conditionally-firing-an-effect) as the last argument to `useEffect`, `useMemo`, `useCallback`, or `useImperativeHandle`, it must include all values used within the function to participate in the React data flow. That includes props, state, and anything derived from them.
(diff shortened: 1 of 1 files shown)


sophiebits:
I think this is correct as written (although maybe not easy to read), and your suggested edit changes the meaning.

content/docs/hooks-faq.md
```diff
@@ -578,7 +578,7 @@ Depending on your use case, there are a few more options described below.
 
 Let's see why this matters.
 
-If you specify a [list of dependencies](/docs/hooks-reference.html#conditionally-firing-an-effect) as the last argument to `useEffect`, `useMemo`, `useCallback`, or `useImperativeHandle`, it must include all values used inside that participate in the React data flow. That includes props, state, and anything derived from them.
+If you specify a [list of dependencies](/docs/hooks-reference.html#conditionally-firing-an-effect) as the last argument to `useEffect`, `useMemo`, `useCallback`, or `useImperativeHandle`, it must include all values used within the function to participate in the React data flow. That includes props, state, and anything derived from them.
```
sophiebits:
```suggestion
If you specify a [list of dependencies](/docs/hooks-reference.html#conditionally-firing-an-effect) as the last argument to `useEffect`, `useMemo`, `useCallback`, or `useImperativeHandle`, it must include all values that are used inside the function and that participate in the React data flow. That includes props, state, and anything derived from them.
```

threepointone:
Thank you!

PR #2651, opened 2019-12-31
Add example to community examples list

chroline:


gaearon:
Thanks, but we're keeping this page focused on vanilla React. No MobX.

PR #2656, opened 2020-01-04
Update examples.md

winwiz1:


gaearon:
Thanks but we're trying to keep this focused on demos rather than boilerplates.

PR #2657, opened 2020-01-05
Update "Refs and the DOM" documentation

evrimfeyyaz:
As function components can now have state through hooks, remove the part that mentions converting a function component to a class component to be able to use state.


sophiebits:
Merged with a couple tweaks, thanks!

PR #2658, opened 2020-01-05
Added "WordPress Theme Browser" example app

maheshwaghmare:
Added the link of example ReactJS application called "[ReactJS WordPress Theme Browser](https://maheshwaghmare.github.io/reactjs-wp-org-themes/)".

It searches the WordPress themes which are hosted on [https://wordpress.org/themes](https://wordpress.org/themes).


gaearon:
Hi. The code formatting is very inconsistent. This also uses classes but we'll soon rewrite docs to Hooks. If you tidy it up, convert to Hooks, and resubmit, we can take it.

PR #2663, opened 2020-01-06
Replace var with const

tanabee:
I replaced `var` with `const` in the document below.
https://reactjs.org/docs/hooks-state.html#tip-what-do-square-brackets-mean

Reference: https://github.com/reactjs/reactjs.org/blob/master/CONTRIBUTING.md#use-const-where-possible-otherwise-let-dont-use-var

### DIFF
--- a/content/docs/hooks-state.md
+++ b/content/docs/hooks-state.md
@@ -235,9 +235,9 @@ The names on the left aren't a part of the React API. You can name your own stat
 This JavaScript syntax is called ["array destructuring"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Array_destructuring). It means that we're making two new variables `fruit` and `setFruit`, where `fruit` is set to the first value returned by `useState`, and `setFruit` is the second. It is equivalent to this code:
 
 ```js
-  var fruitStateVariable = useState('banana'); // Returns a pair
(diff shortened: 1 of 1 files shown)


content/docs/hooks-state.md
```diff
@@ -235,9 +235,9 @@ The names on the left aren't a part of the React API. You can name your own stat
 This JavaScript syntax is called ["array destructuring"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Array_destructuring). It means that we're making two new variables `fruit` and `setFruit`, where `fruit` is set to the first value returned by `useState`, and `setFruit` is the second. It is equivalent to this code:
 
 ```js
-  var fruitStateVariable = useState('banana'); // Returns a pair
-  var fruit = fruitStateVariable[0]; // First item in a pair
-  var setFruit = fruitStateVariable[1]; // Second item in a pair
```
vnoitkumar: Hahah

alexkrolick:
I think this was written this way because people unfamiliar with array destructuring are probably not familiar with the latest JS syntax

sophiebits:
Agreed; it was intended to be written this way. Thank you for contributing though!

tanabee:
I understand. Thank you for your feedback!

PR #2664, opened 2020-01-07
Move 2019 conferences to Past Conferences list

benhodgson87:
Move from upcoming to past:
* React Summit 2019
* React Day Berlin 2019


lex111:
@benhodgson87 thanks!

PR #2672, opened 2020-01-10
'setState' to 'useState'

vlad3k:
Change names hooks.


sophiebits:
It’s correct as written. The text is describing the second value that useState returns, which is a setState function.

PR #2674, opened 2020-01-14
Update versions.yml

artdxl:

### DIFF
--- a/content/versions.yml
+++ b/content/versions.yml
@@ -1,52 +1,52 @@
 - title: '16.12.0'
   changelog: https://github.com/facebook/react/blob/master/CHANGELOG.md#16120-november-14-2019
-- title: '16.11'
+- title: '16.11.0'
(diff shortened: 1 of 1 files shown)


content/versions.yml
```diff
@@ -1,52 +1,52 @@
...
   path: /version/16.1
   url: https://5a1dbcf14c4b93299e65b9a9--reactjs.netlify.com
   changelog: https://github.com/facebook/react/blob/master/CHANGELOG.md#1611-november-13-2017
-- title: '16.0'
+- title: '16.0.0'
   path: /version/16.0
   url: https://5a046bf5a6188f4b8fa4938a--reactjs.netlify.com
   changelog: https://github.com/facebook/react/blob/master/CHANGELOG.md#1600-september-26-2017
-- title: '15.6'
+- title: '15.6.0'
```
alexkrolick: I think these links are OK to change in principle, except for this one that is actually for 15.6.3

threepointone:
Maybe we should instead change .12.0 and .13.0 to .12 and .13

PR #2680, opened 2020-01-15
chore: add a "deploy to Netlify" button

jlengstorf:
This allows folks to fork this repo and deploy these docs to Netlify. 💜

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,8 @@
 
 This repo contains the source code and documentation powering [reactjs.org](https://reactjs.org/).
 
+[![Deploy to Netlify](https://www.netlify.com/img/deploy/button.svg)](https://app.netlify.com/start/deploy?repository=https://github.com/netlify/netlify-statuskit)
+
 ## Getting started
 
 ### Prerequisites


alexkrolick:
I assume this is for people that want to copy the look and feel of reactjs.org for their own docs site? Might be better to move some of the config from the web interface to a netlify.toml config file, since the build commands and output directory are not obvious. Only a netlify owner can do that.

jlengstorf:
@alexkrolick yeah, this would automate the process of forking and deploying the repo — there was a conversation between @bvaughn and @sdras about this, but I'm not sure if updating the `netlify.toml` is something the team wants to do

if it _is_ something we want to tackle, I can add that here for sure

threepointone:
Can you explain the usecase here? How often do people want to fork the react docs and deploy to netlify? Doesn't seem very likely.

sdras:
This PR is the result of two miscommunications- the original task was to put a "deploys by netlify" badge on the site because it's a term of our OSS offering which gives free Pro and advanced services. That's what Brian and I talked about, but Facebook is not actually a free account so it doesn't apply after all (sorry!)

The `netlify.toml` wasn't originally part of our discussions, we can close out from here, thank you for your time in reviewing Sunil! 👋

threepointone:
Thanks for clarifying! Closing this PR.

jlengstorf:
sorry @threepointone! I thought I’d closed this already 😅

PR #2686, opened 2020-01-17
Update examples.md

telecomsxchange:
Added new react example


gaearon:
This seems like an ad.

PR #2688, opened 2020-01-17
missing very important syntax update

amitshahc:
Missing very important syntax update when replacing navite JavaScript with JSX syntax.


threepointone:
The code there specifically focuses on the lines where jsx is used. Besides, your addition isn't actually correct, since it doesn't include the component boundaries. Thanks for the PR!

PR #2693, opened 2020-01-20
I got this warning when using the former version

yoeldovidcohen:
The --typescript option has been deprecated and will be removed in a future release.
In future, please use --template typescript.


lex111:
@yoeldovidcohen thanks!

PR #2695, opened 2020-01-21
Use correct method of updating state in useState docs

jgarciabengochea:
This pr updates a code snippet in the docs for `useState` in which the snippet show the analogous method of updating state in class components.
This follows the practice of updating state shown in https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous


sophiebits:
React will make sure that the state updates are flushed after each click, and we're intentionally don't want to overcomplicate this example, so it's good as-is. Thanks for asking!

PR #2696, opened 2020-01-21
Update docs on how to contribute

necolas:
* Remove outdated content.
* Add JDK installation requirement


sebmarkbage:
Is this new content or existing links taken from elsewhere?

necolas:
It's removal of old content

sebmarkbage:
Ah. I read it in reverse. 😂

PR #2704, opened 2020-01-30
Fix code

hussainarsh:
in componentDidMount function  "**focusTextInput**" is never defined manually so it is better to put "**focus**" in "**this.textInput.current**".

**Wrong:**
```
componentDidMount() {
    this.textInput.current.focusTextInput();
}
```

**Right:**
```
componentDidMount() {
    this.textInput.current.focus();
}
```

### DIFF
--- a/content/docs/refs-and-the-dom.md
+++ b/content/docs/refs-and-the-dom.md
@@ -119,7 +119,7 @@ class AutoFocusTextInput extends React.Component {
   }
 
   componentDidMount() {
-    this.textInput.current.focusTextInput();
+    this.textInput.current.focus();
   }
 
   render() {


bvaughn:
Hey! Thanks for the PR but I think this is a misunderstanding.

This method name was intentional 🙂 In the example you're updating, the `textInput` ref doesn't actually point to an `HTMLInputElement`, but rather a `CustomTextInput` React component:
```js
class AutoFocusTextInput extends React.Component {
  constructor(props) {
    super(props);
    this.textInput = React.createRef();
  }
  componentDidMount() {
    this.textInput.current.focusTextInput();
  }
  render() {
    return (
      <CustomTextInput ref={this.textInput} />
    );
  }
}
```

Elsewhere on the page is an example showing what that component may look like:
```js
class CustomTextInput extends React.Component {
  constructor(props) {
    super(props);

    this.textInput = null;

    this.setTextInputRef = element => {
      this.textInput = element;
    };

    this.focusTextInput = () => {
      // Focus the text input using the raw DOM API
      if (this.textInput) this.textInput.focus();
    };
  }

  componentDidMount() {
    // autofocus the input on mount
    this.focusTextInput();
  }

  render() {
    // Use the `ref` callback to store a reference to the text input DOM
    // element in an instance field (for example, this.textInput).
    return (
      <div>
        <input
          type="text"
          ref={this.setTextInputRef}
        />
        <input
          type="button"
          value="Focus the text input"
          onClick={this.focusTextInput}
        />
      </div>
    );
  }
}
```

PR #2709, opened 2020-02-01
Update examples.md

Usamaahmad751:


gaearon:
Thanks. We're trying to avoid Redux examples on this page. The Todo app looks all right, but we'll probably want a Hooks version of this instead.

PR #2718, opened 2020-02-05
Update conferences.md, add "Frontend United 2020"

chilic:
Adds details for the May "Frontend United 2020" conference in Minsk, Belarus.
Thanks.


lex111:
I'm not sure that this conference needs to be added to the _list of React conferences_, because such a conference should be devoted entirely to React, in this case I see just a frontend conference.

chilic:
@lex111 Thank you.

PR #2722, opened 2020-02-07
added Ingredient project

Prottoy2938:
Ingredient project link: https://ingredient-d4129.firebaseapp.com/


Prottoy2938:
Added my recipe search webapp in the reactjs example project list.

gaearon:
Thanks, but this is a link to the app. I can't view the source so it's not helpful as an example.

PR #2728, opened 2020-02-08
Update hooks-state.md

Aliaksandrsen:
In the chapter API REFERENCE  React.Component setState()
https://reactjs.org/docs/react-component.html#setstate 
I read :

If the next state depends on the current state, we recommend using the updater function form, instead:
this.setState((state) => {
  return {quantity: state.quantity + 1};
});

Isn't that this case?


sophiebits:
React will make sure that the state updates are flushed after each click, and we're intentionally don't want to overcomplicate this example, so it's good as-is. Thanks for asking!

PR #2731, opened 2020-02-10
explicitly remove global create-react-app before npx command

micahgodbolt:
Having an old version of create-react-app installed globally makes this command fail in a way that is hard to understand. https://stackoverflow.com/questions/59188624/template-not-provided-using-create-react-app


alexkrolick:
Sounds like there are a lot of ways this might not quite work (esp if you used `yarn`). May want to link to the stack overflow.

image

threepointone:
This page is for people who are new to React, and I think this addition might confuse. I think the right thing to do here, is for create-react-app to detect when it's being called in this scenario, and warn/log to the user to delete their global version.

PR #2742, opened 2020-02-15
adding next section url path

Aziiiz:
adding next section URL path at the end in order to ease readers follow steps


threepointone:
Thank you!

PR #2746, opened 2020-02-17
add the empty array

KennethKinLum:
or else when the code is tried by other readers, the `setInterval()` will be executed every time `Timer()` is called, instead of the very first time, and set up a series of timers.


KennethKinLum:
the empty array should down below... created another PR https://github.com/reactjs/reactjs.org/pull/2750

PR #2751, opened 2020-02-18
"Here" as the title for link has no semantics

KennethKinLum:
A page could have 50 links, every one with the link text "Here" and it doesn't mean anything.
Search engines rely on the fact that the link text says something about the link. The text "Here" doesn't mean anything to search engines.  Using "Here" as link text has no semantic value.


alexkrolick:
Search engines aside, it's helpful for screen readers to have more descriptive link content. The link text here is a bit long but let's go with it.

PR #2756, opened 2020-02-19
add the empty array

KennethKinLum:
or else when the code is tried by other readers, the `setInterval()` will be executed every time `Timer()` is called, instead of the very first time, and set up a series of timers.


threepointone:
I think that's fine for this specific example, since previous intervals get cleared when the effect runs again. A much better example requires some reading, like https://overreacted.io/making-setinterval-declarative-with-react-hooks/. Thank you for this PR!

PR #2758, opened 2020-02-19
Explaining more about code splitting

IngaleJ:
Hi team,
Here are some input I want to add in react.doc about code splitting.
I hope this will help give some more explanation about feature.

### DIFF
--- a/content/docs/code-splitting.md
+++ b/content/docs/code-splitting.md
@@ -118,7 +118,7 @@ The `React.lazy` function lets you render a dynamic import as a regular componen
 ```js
 import OtherComponent from './OtherComponent';
 ```
-Regular import will load `OtherCompont` along with other imports in single bundle.
+Regular import will load `OtherCompont` along with other imports in bundle at initial load.
 
 **After:**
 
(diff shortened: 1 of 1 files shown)


content/docs/code-splitting.md
```diff
@@ -118,14 +118,15 @@ The `React.lazy` function lets you render a dynamic import as a regular componen
 ```js
 import OtherComponent from './OtherComponent';
 ```
+Regular import will load `OtherCompont` along with other imports in bundle at initial load.
```
IngaleJ: There is no any line about explanation for regular import so I added here

content/docs/code-splitting.md
```diff
@@ -118,14 +118,15 @@ The `React.lazy` function lets you render a dynamic import as a regular componen
...
 
 **After:**
 
 ```js
 const OtherComponent = React.lazy(() => import('./OtherComponent'));
 ```
 
-This will automatically load the bundle containing the `OtherComponent` when this component is first rendered.
+With React.lazy it will automatically load the chunk of code containing the `OtherComponent` when this component is first rendered and cached.
```
IngaleJ: Here try to explain more about `React.lazy`

content/docs/code-splitting.md
```diff
@@ -145,7 +146,7 @@ function MyComponent() {
 }
 ```
 
-The `fallback` prop accepts any React elements that you want to render while waiting for the component to load. You can place the `Suspense` component anywhere above the lazy component. You can even wrap multiple lazy components with a single `Suspense` component.
+The `fallback` prop accepts any React elements that you want to render while waiting for the component to load. You can place the `Suspense` component anywhere above the lazy component. You can even wrap multiple lazy components with a single `Suspense` component. All of them will be rendered same time and this is useful to show only single loading indicator, instead of showing multiple loading indicator for invidual component.
```
IngaleJ: Added explanation about what will happen in race condition.(One lazy loaded component load faster/slower than other)
[reference](https://web.dev/code-splitting-suspense/)

IngaleJ:
Hi team, please review my changes

PR #2761, opened 2020-02-20
Docs: "async" mode to "concurrent" mode

BurntCaramel:
Changes mention of "async" mode to "concurrent" mode on strict-mode page.

Should hooks such as `useLayoutEffect`, `useEffect`, etc also be listed in the "Render phase lifecycles" section?


bvaughn:
> Should hooks such as useLayoutEffect, useEffect, etc also be listed in the "Render phase lifecycles" section?

No. Those are commit phase hooks. The strict mode docs only mention class lifecycle methods though.

PR #2762, opened 2020-02-20
Make React.memo documentation more clear

Sonicrida:
Just a small tweak to the language explaining `React.memo`. 

I had a discussion on Reactiflux that left me thinking that the language could be tweaked a bit here to further emphasize that it only cares about Props + "affects" seems misleading since it doesn't do anything to your props.


threepointone:
Thanks! Tweaked it a little myself.

Sonicrida:
Works for me!

PR #2763, opened 2020-02-20
New example project added - Tabata

KaterinaLupacheva:


krishnadevz:
why team is not merging our projects in reactjs repo? i am not getting because here two many projects PR's are pending .if you have any answer or reply to me please ASAP comment.

gaearon:
Thank you for the PR! We're trying to keep this page focused on plain React examples (no Redux) so it doesn't quite fit it.

PR #2764, opened 2020-02-20
s/'a myriad of'/'myriad'

dangerismycat:
Think of "myriad" as a number. You wouldn't say "a nine of things", you would say "nine things".


alexkrolick:
This is @acdlite content for sure

dangerismycat:
Anyone wanna merge this? I think after two weeks it's clear @acdlite isn't too concerned about reviewing 😅

PR #2773, opened 2020-02-24
Updated hooks-state docs to use setState with updater function

### LINKED ISSUE #2739: Incorrect setState usage in example
https://reactjs.org/docs/hooks-state.html#equivalent-class-example shows 

```js
render() {
    return (
      <div>
        <p>You clicked {this.state.count} times</p>
        <button onClick={() => this.setState({ count: this.state.count + 1 })}>
          Click me
        </button>
      </div>
    );
  }
```

However, https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous provides the following examples, indicating that we should not use  ```this.state.count + 1``` when updating state dues to async updates:

```js
// Wrong
this.setState({
  counter: this.state.counter + this.props.increment,
});

// Correct
this.setState((state, props) => ({
  counter: state.counter + props.increment
}));
```

akshayymahajan:
[Docs](https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous) mention that setState calls are batched and the state may be updated asynchronously. The example shown on https://reactjs.org/docs/hooks-state.html#equivalent-class-example uses the incorrect way of updating the state. 

**How it looks right now**
image

**Correct Way**
image

This PR updates the example to use the correct implementation using an updater function. 

Resolves #2739

### DIFF
--- a/content/docs/hooks-state.md
+++ b/content/docs/hooks-state.md
@@ -47,7 +47,7 @@ class Example extends React.Component {
     return (
       <div>
         <p>You clicked {this.state.count} times</p>
-        <button onClick={() => this.setState({ count: this.state.count + 1 })}>
+        <button onClick={() => this.setState(state => ({ count: state.count + 1 }))}>
           Click me
         </button>
       </div>


sophiebits:
React flushes state updates after a click so this is actually guaranteed to work. In addition, we're trying to keep this example as simple as possible to understand.

Thanks for the contribution though!

welll:
@sophiebits , since the page  [setState](
https://reactjs.org/docs/react-component.html#setstate) doesn't mention it. I would like to suggest two possible solutions to avoid misunderstanding.

1) Updating the `setState` section to include an external link that could explain your statement `React flushes state updates after a click so this is actually guaranteed to work....`


Screen Shot 2021-06-23 at 5 13 53 PM
 
 2) Update te `hooks-state docs` to update the state using a different example, maybe something like this:
 
 ```javascript
class Example extends React.Component {
  constructor(props) {
    super(props);
    this.state = {
      number: 0
    };
  }

  render() {
    return (
      <div>
        <p>Your lottery number is {this.state.number} </p>
        <button onClick={() => this.setState({ number: Math.random() )}>
          Click me
        </button>
      </div>
    );
  }
}
```

Please, let me know what is your input.
I would love to contribute updating the doc.

PR #2775, opened 2020-02-24
Update introducing-jsx.md

Materuilist:


threepointone:
This feels like nit picking, constants are *technically* variables wrt general vocabulary. I think I'll leave it as is. Thank you for the PR!

PR #2777, opened 2020-02-25
Removed redundant words from Hyperlink text

arshita-kakkar:


threepointone:
Why do you think those words are redundant? I don't feel strongly about it, but happy to hear the reasoning before merging. Thanks!

PR #2782, opened 2020-02-26
fix 16.13 blog post date

threepointone:
(as well the right link to the changelog)


gaearon:
This broke links. We should leave a redirect for such cases.

PR #2788, opened 2020-02-28
Update reference-react.md wording

balazsorban44:
I believe something was off with the wording, but I let someone else prove me wrong.

### DIFF
--- a/content/docs/reference-react.md
+++ b/content/docs/reference-react.md
@@ -110,7 +110,7 @@ If your React component's `render()` function renders the same result given the 
 
 > Note
 >
-> `React.PureComponent`'s `shouldComponentUpdate()` only shallowly compares the objects. If these contain complex data structures, it may produce false-negatives for deeper differences. Only extend `PureComponent` when you expect to have simple props and state, or use [`forceUpdate()`](/docs/react-component.html#forceupdate) when you know deep data structures have changed. Or, consider using [immutable objects](https://facebook.github.io/immutable-js/) to facilitate fast comparisons of nested data.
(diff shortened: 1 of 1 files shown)


content/docs/reference-react.md
```diff
@@ -110,7 +110,7 @@ If your React component's `render()` function renders the same result given the
 
 > Note
 >
-> `React.PureComponent`'s `shouldComponentUpdate()` only shallowly compares the objects. If these contain complex data structures, it may produce false-negatives for deeper differences. Only extend `PureComponent` when you expect to have simple props and state, or use [`forceUpdate()`](/docs/react-component.html#forceupdate) when you know deep data structures have changed. Or, consider using [immutable objects](https://facebook.github.io/immutable-js/) to facilitate fast comparisons of nested data.
+> `React.PureComponent`'s `shouldComponentUpdate()` only shallowly compares the objects. If these contain complex data structures, it may produce false-positives for deeper differences. Only extend `PureComponent` when you expect to have simple props and state, or use [`forceUpdate()`](/docs/react-component.html#forceupdate) when you know deep data structures have changed. Or, consider using [immutable objects](https://facebook.github.io/immutable-js/) to facilitate fast comparisons of nested data.
```
danielkov: I believe the wording was actually correct, e.g.:
```js
const originalProps = { some: 'value' };
const instance = new MyPureComponent(originalProps);
// ... render, etc.
originalProps.some = 'different value';
// roughly what happens internally
instance.shouldComponentUpdate(originalProps); // `false`, which is a false-negative
```
It's meant to state that unless there is referential inequality between deep values ("complex data structures"), `shouldComponentUpdate()` is implemented to return false when extending `React.PureComponent`.

The component says: "no I shouldn't update", when your real intention is to update it (trigger a new render + lifecycles), thus `shouldComponentUpdate()` yields a false-negative (false, because a positive return value is what we were hoping for.

Hope this helps understand the wording.

balazsorban44: Thanks!

PR #2791, opened 2020-03-01
Update examples.md

harshit860:
Adding Restaurant Food Ordering Application.

This application is made from only react state uplifting. The styling used is MaterialUi + Bootstrap. It has some good transitions. It is my hobby project I always keep it updated with new stuff. Fully responsive for mobile devices.


gaearon:
Thanks for the PR. The code formatting and naming is rather inconsistent. An app like this is also a bit confusing without a backend so I'm not sure this makes sense as an example.

PR #2794, opened 2020-03-01
Update hooks-state.md

YaroslavTER:
HI, as far as I understand this way to set state based on previous value is not correct
```js
this.setState({ count: this.state.count + 1 })
```
and you should use callback instead.
Thanks.

### DIFF
--- a/content/docs/hooks-state.md
+++ b/content/docs/hooks-state.md
@@ -47,7 +47,7 @@ class Example extends React.Component {
     return (
       <div>
         <p>You clicked {this.state.count} times</p>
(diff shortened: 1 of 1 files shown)


threepointone:
I'm going to push back and not accept this PR because passing an expression is indeed a supported api and works just fine. The callback version is useful when you're not certain what the captured value of state is. Thank you for the PR!

davelnewton:
@threepointone Can you expand on that? The [React Docs](https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous) explicitly state that referencing state directly may not work.

PR #2800, opened 2020-03-03
Added example project to examples.md

jamesefhawkins:
Suggested adding [PostHog](https://github.com/posthog/posthog/) to the example React projects


gaearon:
Thanks for the PR! We're keeping this page focused on vanilla React examples so this doesn't match it very well.

PR #2803, opened 2020-03-03
Add ajax with hooks example

### LINKED ISSUE #2799: The FAQ page on AJAX and APIs should be refactored to use hooks
The FAQ page describing [how to interact with APIs via AJAX](https://reactjs.org/docs/faq-ajax.html) still uses class components. This should probably be updated to use hooks

lgordey:
Fix #2799

### DIFF
--- a/content/docs/faq-ajax.md
+++ b/content/docs/faq-ajax.md
@@ -91,6 +91,9 @@ function MyComponent() {
   const [isLoaded, setIsLoaded] = useState(false);
   const [items, setItems] = useState([]);
 
+  // Note: the empty deps array [] means
+  // this useEffect will run once
+  // similar to componentDidMount()
   useEffect(() => {
     fetch("https://api.example.com/items")
       .then(res => res.json())
--- a/content/docs/faq-ajax.md
+++ b/content/docs/faq-ajax.md
@@ -83,7 +83,7 @@ class MyComponent extends React.Component {
 }
 ```
 
-Has the following equivalent with [Hooks](https://reactjs.org/docs/hooks-intro.html):
+Here is the equivalent with [Hooks](https://reactjs.org/docs/hooks-intro.html): 
 
 ```js
 function MyComponent() {
--- a/content/docs/faq-ajax.md
+++ b/content/docs/faq-ajax.md
@@ -82,3 +82,47 @@ class MyComponent extends React.Component {
   }
 }
 ```
+
+Has the following equivalent with [Hooks](https://reactjs.org/docs/hooks-intro.html):
+
+```js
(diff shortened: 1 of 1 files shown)


content/docs/faq-ajax.md
```diff
@@ -82,3 +82,47 @@ class MyComponent extends React.Component {
...
 ```
+
+Has the following equivalent with [Hooks](https://reactjs.org/docs/hooks-intro.html):
+
+```js
+function MyComponent() {
+  const [error, setError] = useState(null);
+  const [isLoaded, setIsLoaded] = useState(false);
+  const [items, setItems] = useState([]);
+
```
brettdewoody: Perhaps we should include a comment here to remind readers the empty deps array `[]` means this `useEffect` will run once similar to `componentDidMount()`?

lgordey: Done

lgordey:
@brettdewoody done!

content/docs/faq-ajax.md
```diff
@@ -82,3 +82,50 @@ class MyComponent extends React.Component {
...
 }
 ```
+
+Has the following equivalent with [Hooks](https://reactjs.org/docs/hooks-intro.html):
+
+```js
+function MyComponent() {
+  const [error, setError] = useState(null);
+  const [isLoaded, setIsLoaded] = useState(false);
+  const [items, setItems] = useState([]);
```
alexkrolick: This would be a good use case for a reducer (loading/error/hasItems are exclusive states). Thoughts on showing that here?

content/docs/faq-ajax.md
```diff
@@ -82,3 +82,50 @@ class MyComponent extends React.Component {
   }
 }
 ```
+
+Has the following equivalent with [Hooks](https://reactjs.org/docs/hooks-intro.html):
```
alexkrolick:
```suggestion
Here is the equivalent with [Hooks](https://reactjs.org/docs/hooks-intro.html): 
```

content/docs/faq-ajax.md
brettdewoody: I thought that as well at first. But going from a Class example using state, to a functional example using useReducer seems like a big jump. While useReducer is a better choice here in a real app, imo sticking with useState is better for this particular example.

alexkrolick: Maybe... they don’t need to be, though, so much as represent the recommended pattern for each

@sophiebits @gaearon thoughts?

alexkrolick: We'll go with this for now

PR #2807, opened 2020-03-04
Update examples.md

aman1106:


gaearon:
Thanks, but this one is a bit too simple for this page IMO. Appreciate the effort though.

aman1106:
@gaearon Yes, even I think it is more on CSS side. Anyway thanks for reviewing. :)

PR #2810, opened 2020-03-04
Add shouldComponentUpdate to list of methods double called in Strict Mode

threepointone:
This landed in 16.13 via https://github.com/facebook/react/pull/17942


bvaughn:
Ah, thanks Sunil.

PR #2813, opened 2020-03-04
fix: extract instantiation of Remarkable

danielkov:
Hello 👋!

Originally in the example **A Component Using External Plugins** `Remarkable` was instantiated each time render was called, while I believe the whole point of this example is to show how React can be integrated with other libraries, taking into account different lifecycles and how they're tied together.

With this change the plugin will be instantiated in the constructor, because it makes sense for it to exist throughout the full lifecycle of the component and the example will showcase the more complete and usual use-case this way.


threepointone:
Thank you!

PR #2814, opened 2020-03-05
Remove mention of React.createFactory() in docs

valscion:
In React v16.13.0, the React.createFactory() [got deprecated](https://reactjs.org/blog/2020/02/26/react-v16.13.0.html#deprecating-reactcreatefactory) so it would be useful to not showcase it in a warning page about legacy factories usage.

This warning page (https://reactjs.org/warnings/legacy-factories.html) was one of the top Google results for me when I searched for "react createfactory" in Google after upgrading to React v16.13.0 and got confused.

react_createfactory_-_Google-haku_-_Mozilla_Firefox

I'm open to any wording suggestions. Feel free to change the text directly, too, if you have any ideas on how I could describe this better :relaxed:

Deploy preview of this change: https://deploy-preview-2814--reactjs.netlify.com/warnings/legacy-factories.html

### DIFF
--- a/content/warnings/legacy-factories.md
+++ b/content/warnings/legacy-factories.md
@@ -27,24 +27,9 @@ function render() {
 }
 ```
 
-## Without JSX {#without-jsx}
-
-If you don't want to, or can't use JSX, then you'll need to wrap your component in a factory before calling it:
-
-```javascript
-var React = require('react');
-var MyComponent = React.createFactory(require('MyComponent'));
-
-function render() {
-  return MyComponent({ foo: 'bar' });
-}
-```
-
-This is an easy upgrade path if you have a lot of existing function calls.
-
 ## Dynamic components without JSX {#dynamic-components-without-jsx}
 
-If you get a component class from a dynamic source, then it might be unnecessary to create a factory that you immediately invoke. Instead you can just create your element inline:
+If you get a component class from a dynamic source, you can also create your element inline:
 
 ```javascript
 var React = require('react');


danielkov:
Hello! I am not a maintainer of this or anything, just curious. I guess I don't really see the point of this change. That page is there for the exact purpose of providing guidance for developers handling legacy codebases, which is why it's under `warnings/legacy-factories`. The method described there shows a pattern that used to be common for people not using JSX, for example in codebases that combined Backbone with React or another library.

I might be wrong about this one, because I haven't seen one of those warnings in a while, but I also remember the minified error link pointing to this page too, so I think these pages can't really be edited, without taking away meaningful information from users of React.

valscion:
Good point! Maybe instead of removing the mention, I should amend the page instead to show that even `React.createFactory` is obsolete?

The reason why I thought this page would be nice if it were edited is because it's among the first results in Google, and it isn't clear that all the instructions shown on this page are good only for legacy React versions.

I'll think about this. Thanks for bringing up your concern!

danielkov:
I see your point about Google search results. I wonder, what would you expect to be on top of search results when searching for `react createfactory`? I can imagine a situation when a developer familiar with React 16+ will inherit a codebase that's using an older version and so they would Google stuff like `react createclass`, `react mixins`, `react createfactory`, etc to bring them up to speed. I think right now the search results for those would provide help in that situation. 

What do you have in mind?

valscion:
¯\\\_(ツ)_/¯ I don't really know, or have an opinion on what I'd expect to find.

My point in here is to merely update a section of documentation that is suggesting to use `React.createFactory` to, well, not suggest it. As it now displays deprecation warnings.

I merely do not want people to be confused when they do as the documentation says. Not everybody understands that the documentation they're looking at has outdated suggestions

valscion:
I'll close this PR for now. I might get back to this, but the solution to this docs change will anyways be different to what I originally proposed here ☺️

PR #2818, opened 2020-03-08
Move Hungarian to in progress

balazsorban44:
I just realized that Hungarian was in the wrong category. I now moved it to In Progress. In fact, the `Core` is nearly finished, only reviewing of 3 pages remain.


threepointone:
Thank you!

PR #2825, opened 2020-03-11
remove parantheses from onClick method

nodermatt:
writing this.props.onClick() (with parantheses) invokes the passed method immediately. This causes an update of the component. In turn the render() method calls the method again, resulting in an infinite loop that finally crashes in a "Maximum depth exceeded" error.

### DIFF
--- a/content/tutorial/tutorial.md
+++ b/content/tutorial/tutorial.md
@@ -440,7 +440,7 @@ class Square extends React.Component {
     return (
       <button
         className="square"
-        onClick={() => this.props.onClick()}
+        onClick={() => this.props.onClick}
       >
         {this.props.value}
       </button>


threepointone:
This isn’t right, `()=>...` is a function and don’t get immediately called. Can you reproduce the problem in a codesandbox?

danielkov:
The code in your PR is in fact incorrect, the first version does not create an infinite render cycle, however the use of an anonymous lambda is unnecessary (unless to exclude the `SyntheticEvent` that is the parameter of the click callback. Perhaps you could update it to indicate this:
```jsx
class Square extends React.Component {
  render() {
    return (
      <button
        className="square"
        onClick={this.props.onClick}
      >
        {this.props.value}
      </button>
    );
  }
}
```

threepointone:
This feels like a nit at this point. I’m going to close this PR. Thanks for your input!

PR #2828, opened 2020-03-13
currency converter example

dipanshuraz:
added the easiest currency converter with react js


gaearon:
Thanks. This is a good start but we're hesitant to merge more class examples. If you convert this to Hooks, feel free to resubmit.

dipanshuraz:
@gaearon Sure I will convert this to a Hooks example. Thanks

PR #2830, opened 2020-03-13
Update introducing-jsx.md

david-mateogit:
The img element does not need a closing tag.


threepointone:
This is by design, since the very next example is about how that element doesn't need a closing tag. Thanks for the PR!

PR #2835, opened 2020-03-15
Reorganize information about the plugin

eluchsinger:
Having the information that the plugin is already installed by default after the installation instructions is probably wasting a lot of people's time. Moving it up probably avoids multiple installs of the plugin.


lex111:
@eluchsinger thanks!

PR #2837, opened 2020-03-15
Fix node version in readme

Zeko369:
I forgot to change the Node version in README when upgrading to Node 12 in #2831 , so here is the fix


Zeko369:
Was @facebook-github-bot broken yesterday?, I created a PR that since got merges and the bot didn't notice it 😕

threepointone:
Thank you!

Issue #2833, opened 2020-03-14
reactjs.org homepage examples are outdated


slk333:
The examples are located here:
https://github.com/reactjs/reactjs.org/tree/master/content/home/examples
We can see that they received little to no change since their creation, more than 3 years ago.

There are two problems:

- The examples don't use hooks
- The examples use classes

More generally, reactjs.org introduction largely ignores hooks, while using classes throughout. As a result, **hooks are largely ignored by newcomers.**

Suggestion: the Clock example and the Markdown Editor examples should come in two alternative versions: one using Class, one using function components and useState() and useEffect() hooks.

Zeko369:
I'm already working on this, will have a PR today

slk333:
https://github.com/reactjs/reactjs.org/issues/2874#issuecomment-607930242

PR #2838, opened 2020-03-15
Home page examples hooks and classes

Zeko369:
Upgrade Flow version and add a tab switcher to switch between hooks and class examples on homepage

Closes: #2833

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -17,7 +17,6 @@ module.system.node.resolve_dirname=src
 
 esproposal.class_static_fields=enable
 esproposal.class_instance_fields=enable
-unsafe.enable_getters_and_setters=true
 
 munge_underscores=false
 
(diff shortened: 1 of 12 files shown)


Zeko369:
The second example (`setInterval`) requires the use of useEffect, so I'd rather add a simple counter example (+, - buttons and a value) for the stateful example, add use this for "lifecycle" example.

Any ideas/comments?

slk333:
a welcomed change that will probably require the approval of managers because it involves the frontpage of the website. Thank you for this pull request and i hope it will bring things forward!

ImedAdel:
@gaearon Are you rewriting all the examples using hooks and functional components? 

I'd like to help if that's the case :) The tutorial, with classes, is a bit intimidating

krishnadevz:
i just updated hooks section for useState in functional component if anyone is maintainer here they can merge/review my pull request here 👉https://github.com/reactjs/reactjs.org/pull/2842

PR #2839, opened 2020-03-16
ReactConf Japan 2020 is cancelled due to COVID-19

visshaljagtap:


lex111:
@visshaljagtap thank you for update!

PR #2841, opened 2020-03-17
'MyButton'  would be a string name

scazzy:


Zeko369:
This wouldn't actually happen. You can easily test my examples on the [landing page](reactjs.org). The first param is a string (referring to html tag)  or a component that you want to render.

```jsx
const Foobar = () => <h1>Hello world</h1>

ReactDOM.render(
  <Foobar />,
  document.getElementById('hello-example')
);
```
compiles down to
```js
const Foobar = () => React.createElement(
  'h1',
  null,
  'Hello world'
);

ReactDOM.render(React.createElement(Foobar, null), document.getElementById('hello-example'));
```

You can see that `Foobar` stays `Foobar` refering to the function above

PR #2842, opened 2020-03-18
added basic example of useState hook in functional component

krishnadevz:

### DIFF
--- a/content/docs/hooks-state.md
+++ b/content/docs/hooks-state.md
@@ -81,6 +81,20 @@ function Example(props) {
   return <div />;
 }
 ```
+**Basic Example of useStae hook in functional component**
+
+```js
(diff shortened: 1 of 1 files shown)


alexkrolick:
this is introduced later - this section is just reminding the reader of the ways to define function components 😄

krishnadevz:
@alexkrolick can we push this thing as example code snippet .👋

alexkrolick:
There are already several examples on this page, this area does not need one

krishnadevz:
> 
> 
> There are already several examples on this page, this area does not need one

i think this one is a simple example that can help to the community that's why i made pull request and added example projects please go through 👉https://github.com/reactjs/reactjs.org/pull/2840

PR #2855, opened 2020-03-22
Superheros Example

aman1106:
List of superheros with live search and details.


gaearon:
Thanks but we don't want Redux examples on this page.

aman1106:
@gaearon removed redux from the project and raised the PR again #2897

PR #2859, opened 2020-03-24
Update dependencies

MichaelDeBoey:


package.json
```diff
@@ -7,58 +7,58 @@
...
+    "react-live": "1.12.0",
+    "remarkable": "^1.7.4",
+    "request-promise": "^4.2.5",
+    "rimraf": "^2.7.1",
+    "slugify": "^1.4.0",
+    "unist-util-visit": "^1.4.1"
   },
   "engines": {
-    "node": "12.x.x",
+    "node": ">8.4.0",
```
gaearon: ?

MichaelDeBoey: Oops that's something left from rebasing 😅

MichaelDeBoey:
Ping @gaearon

package.json
```diff
@@ -7,55 +7,55 @@
...
-    "eslint-plugin-flowtype": "^2.39.1",
-    "eslint-plugin-jsx-a11y": "^6.0.2",
-    "eslint-plugin-prettier": "^2.3.1",
-    "eslint-plugin-react": "^7.4.0",
+    "eslint-plugin-flowtype": "^2.50.3",
+    "eslint-plugin-jsx-a11y": "^6.2.3",
+    "eslint-plugin-prettier": "^2.7.0",
+    "eslint-plugin-react": "^7.19.0",
     "eslint-plugin-relay": "^0.0.19",
     "flow-bin": "^0.56.0",
```
Zeko369, line 20: Bump flow to at least 0.75 to get the LSP protocol and all the nice autocompletion that comes with that

MichaelDeBoey: Bumping it to latest caused some errors so that's why I reverted it

Zeko369: Flow changed a lot since then, it works if you update some suff in the `.flowconfig`.

Comment out or remove this
```
unsafe.enable_getters_and_setters=true
```

And update the flow version at the bottom

```
[version]
^0.121.0
```

Then reload the vscode plugin and it should work.

Side note, it would be really cool if you could comment on other files in a PR, can you do it and I just don't know about it??

MichaelDeBoey:
Ping @gaearon @Zeko369

PR #2860, opened 2020-03-24
 Updated the conferences

visshaljagtap:
- Updated the conference date of React Summit Amsterdam.
- App.js Conf 2020 is cancelled due to the COVID-19


lex111:
@visshaljagtap thanks!

PR #2873, opened 2020-03-31
Fix incorrect highlighting in uncontrolled components example

CharlieStras:
In the example of using file input tag, incorrect line is highlighted.

Previous:
571585658962_ pic

Current:
581585659314_ pic


alexkrolick:
Thanks!

PR #2879, opened 2020-04-04
Fix object syntax error

jhernandez-c:
The object passed into dispatch() function is missing the "text" key.

### DIFF
--- a/content/docs/hooks-custom.md
+++ b/content/docs/hooks-custom.md
@@ -242,7 +242,7 @@ function Todos() {
   const [todos, dispatch] = useReducer(todosReducer, []);
 
   function handleAddClick(text) {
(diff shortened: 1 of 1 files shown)


content/docs/hooks-custom.md
```diff
@@ -242,7 +242,7 @@ function Todos() {
   const [todos, dispatch] = useReducer(todosReducer, []);
 
   function handleAddClick(text) {
-    dispatch({ type: 'add', text });
+    dispatch({ type: 'add', text: text });
```
alexkrolick, line 245: both ways work

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer

PR #2880, opened 2020-04-04
Update examples.md

Usamaahmad751:


gaearon:
Thanks, but we'd like to keep this page focused on vanilla React examples.

PR #2884, opened 2020-04-06
Explain why useMemo with stable deps runs 2x in dev

tomhicks:
As explained here:

https://twitter.com/acdlite/status/1067541668797145090

React will always run useMemo twice, even if the dependencies are identical. This is missing in the docs, and can make for a very frustrating developer experience.


gaearon:
This only happens in Strict Mode. I don't think it always happens in development. And `useMemo` is not unique here — we double-call any pure method. The list of them is [here](https://reactjs.org/docs/strict-mode.html#detecting-unexpected-side-effects). I agree it's worth pointing this out but needs to be done for all places and not just `useMemo` alone.

tomhicks:
Unsure how much to simply refer to the link you gave vs explain inline.

As a consumer of these docs, I went to useMemo and didn't find anything relevant. A simple signpost to the link you gave would have been enough to save me several hours.

PR #2890, opened 2020-04-08
Add link to useContext

k1sul1:
I haven't created any contexts before, and I didn't know how to use the context without resorting to the ugly ctx => (<div><Component x={ctx.y} /></div>). I've read the hooks manual like a dozen times but that didn't help me remember the existence of useContext.

### DIFF
--- a/content/docs/context.md
+++ b/content/docs/context.md
@@ -19,6 +19,7 @@ In a typical React application, data is passed top-down (parent to child) via pr
 - [Examples](#examples)
   - [Dynamic Context](#dynamic-context)
   - [Updating Context from a Nested Component](#updating-context-from-a-nested-component)
+  - [Usage with useContext](#usage-with-usecontext)
   - [Consuming Multiple Contexts](#consuming-multiple-contexts)
 - [Caveats](#caveats)
 - [Legacy API](#legacy-api)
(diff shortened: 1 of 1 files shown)


content/docs/context.md
```diff
@@ -19,6 +19,7 @@ In a typical React application, data is passed top-down (parent to child) via pr
 - [Examples](#examples)
   - [Dynamic Context](#dynamic-context)
   - [Updating Context from a Nested Component](#updating-context-from-a-nested-component)
+  - [Usage with useContext](#usage-with-usecontext)
```
alexkrolick, line 22:
```suggestion
  - [Usage with the useContext Hook](#usage-with-the-usecontext-hook)
```

content/docs/context.md
```diff
@@ -239,6 +240,10 @@ It is often necessary to update the context from a component that is nested some
 **app.js**
 `embed:context/updating-nested-context-app.js`
 
+### Usage with useContext
+
+https://reactjs.org/docs/hooks-reference.html#usecontext
```
alexkrolick, line 245: We should probably move this to an embed like the other context examples, but it would be good to show an inline example here. The one on the Hooks page is consistent with the example context already on this page (a "theme" provider).

```suggestion
### Usage with the useContext Hook

```jsx
function ThemedButton() {
  const theme = useContext(ThemeContext);
  return (
      <button style={{ background: theme.background, color: theme.foreground }}>
      I am styled by theme context!
    </button>
  );
}
```

[useContext Hook API Reference](/docs/hooks-reference.html#usecontext)
```

k1sul1: You're probably right. I didn't write anything as I felt it would only be redundant, and I'm not much of a technical writer anyway. 

For a moment I thought that I couldn't use context in function components, because this page didn't even mention them, and most answers I found via Google were either wrong or written in the React.Component era.

I feel like just adding a link would be enough to mitigate that problem, but maybe more is more in this case.

PR #2903, opened 2020-04-10
An extra comma in the object containing one element

mohasarc:
there is an extra comma in the object that is being passed to setState function


alexkrolick:
It's added by Prettier to reduce diffs when you add more lines, and is valid syntax

PR #2907, opened 2020-04-12
Be more specific about empty dependency array vs omitting the array f…

Xunnamius:
…or useMemo


Xunnamius:
Might also add a warning about omitting dependencies in light of https://github.com/facebook/react/pull/15025

PR #2908, opened 2020-04-12
correcting typo from .js to .jsx

SamratGantayat:
correcting typo from .js to .jsx


rachelnabors:
Actually, this isn't necessary! `.js` is fine :) Thanks for submitting, though!

PR #2920, opened 2020-04-19
Fix misspelling

henryStelle:


Dergash:
It doesn't seem to be a misspelling 😕 Aren't they just referring to "their" expirience, as in "React was  used internally by our own teams working on Facebook/Instagram", hence "us"?

henryStelle:
Good catch. I was reading too quickly and forgot that the react team works at Facebook.

PR #2931, opened 2020-04-24
Remove unnecessary state and function from effect hook example

vadimshvetsov:
Presence of `handleStatusChange` function and `isOnline` state in general are unclear at this effect hook example. Also `isOnline` looks as `Boolean` by name. It might be `[status, setStatus]` state for consumers.

I could try to improve all hooks examples with this intention if it's appreciated.

### DIFF
--- a/content/docs/hooks-effect.md
+++ b/content/docs/hooks-effect.md
@@ -203,13 +203,7 @@ You might be thinking that we'd need a separate effect to perform the cleanup. B
 import React, { useState, useEffect } from 'react';
 
 function FriendStatus(props) {
-  const [isOnline, setIsOnline] = useState(null);
-
   useEffect(() => {
-    function handleStatusChange(status) {
-      setIsOnline(status.isOnline);
-    }
-
     ChatAPI.subscribeToFriendStatus(props.friend.id, handleStatusChange);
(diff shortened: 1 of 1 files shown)


content/docs/hooks-effect.md
```diff
@@ -203,13 +203,7 @@ You might be thinking that we'd need a separate effect to perform the cleanup. B
 import React, { useState, useEffect } from 'react';
 
 function FriendStatus(props) {
-  const [isOnline, setIsOnline] = useState(null);
-
   useEffect(() => {
-    function handleStatusChange(status) {
```
alexkrolick, line 209: This functions is needed as the callback below. 

```js
ChatAPI.subscribeToFriendStatus(props.friend.id, handleStatusChange); 
```

The example is showing why it's important to clean up after effects; otherwise, this `handleStatusChange` will subscription will keep being called after the component is unmounted or when the friend id changes

vadimshvetsov: @alexkrolick looks like `setIsOnline` is sufficient in this way, what's the point of separate function in `useEffect`? And `isOnline` has a boolean naming for the string value. Let me know if you want me to change it to the `status`.

PR #2938, opened 2020-04-28
Fix typos in callback as parameters

danieldanielecki:
This PR fixes typos in the docs when callback are being passed as parameters in 3 cases.

### DIFF
--- a/content/docs/reference-react-component.md
+++ b/content/docs/reference-react-component.md
@@ -533,7 +533,7 @@ The second parameter to `setState()` is an optional callback function that will 
 You may optionally pass an object as the first argument to `setState()` instead of a function:
 
 ```javascript
-setState(stateChange[, callback])
+setState(stateChange, [callback])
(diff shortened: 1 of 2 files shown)


content/docs/reference-react-component.md
```diff
@@ -533,7 +533,7 @@ The second parameter to `setState()` is an optional callback function that will
 You may optionally pass an object as the first argument to `setState()` instead of a function:
 
 ```javascript
-setState(stateChange[, callback])
```
alexkrolick, line 536: this syntax indicates an optional argument, not an array

danieldanielecki: @alexkrolick if that's the case in both cases then reject this PR.

alexkrolick: closing for now, but if you have a suggestion for an alternative way of showing the optional args, please submit another one!

PR #2940, opened 2020-04-29
updated tutorial.md

Timibadass:
The method is not used anywhere inside render()

### DIFF
--- a/content/tutorial/tutorial.md
+++ b/content/tutorial/tutorial.md
@@ -845,18 +845,6 @@ The Board component now looks like this:
 
 ```javascript{17,18}
 class Board extends React.Component {
-  handleClick(i) {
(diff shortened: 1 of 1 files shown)


content/tutorial/tutorial.md
```diff
@@ -845,18 +845,6 @@ The Board component now looks like this:
 
 ```javascript{17,18}
 class Board extends React.Component {
-  handleClick(i) {
```
alexkrolick, line 848: should this replace "this.props.onClick" in renderSquaure?

Timibadass: `this.props.onClick` can either be inside the handleClick method, or the `handleClick` method should also be deleted from the Board component since it isn't being called anywhere at that point so learners(like me) can follow along properly.

PR #2946, opened 2020-05-02
The initialState should be an object

chpantoleon:

### DIFF
--- a/content/docs/hooks-custom.md
+++ b/content/docs/hooks-custom.md
@@ -239,7 +239,7 @@ Now we could use it in our component, and let the reducer drive its state manage
 
 ```js{2}
 function Todos() {
-  const [todos, dispatch] = useReducer(todosReducer, []);
+  const [todos, dispatch] = useReducer(todosReducer, {});
 
   function handleAddClick(text) {
     dispatch({ type: 'add', text });


hzhu:
Could you clarify the reason the `initialState` has to be an object?

Depending on the use-case, the initial value for `useReducer` can be of any value — an array or an object. In this particular case, it should be an array because the developer is choosing an array to drive the reducer and the UI*. Notice the next state is the previous todos and the new todo spread onto a new array.

```jsx
function todosReducer(state, action) {
  switch (action.type) {
    case 'add':
      return [...state, {
        text: action.text,
        completed: false
      }]; // returns a new array
    // ... other actions ...
    default:
      return state;
  }
}
```

\* The component code which is driven by an array based state might look like this:

```jsx
function Todos() {
  const [todos, dispatch] = useReducer(todosReducer, []);

  return todos.map((todo, index) => (
    <div>
      <div>Todo: {todo.text}</div>
      <div>Is completed: {todo.completed}</div>
    </div>
  ))

  // ...
}
```

Feel free to close this pull request since the `initialState` in the docs is correct as an array.

chpantoleon:
Thanks you for taking the time to explain what I missed!

PR #1395, opened 2018-11-12
Add constructor usage example for Class.contextType

BRKalow:
Adds an example for accessing `this.context` inside of a constructor using the new `Class.contextType` introduced in `16.6`. It didn't seem necessary to document it beyond just including example constructor access in the existing `Class.contextType` code snippet.

Adding to the documentation based on the conversation here: https://github.com/facebook/react/issues/13944

Thanks!

PR #2952, opened 2020-05-06
docs(context): pass `context` to `super` when using `class` `constructor`

iamandrewluca:
Related #1395


iamandrewluca:
React internally is setting `props`, `context` and other fields eplicit on a instance.

PR #2957, opened 2020-05-07
Update cdn-links.html

AreRex14:
link to next article is wrong. It should be release-channels.html, not hello-world.html


lex111:
@AreRex14 thanks!

PR #2962, opened 2020-05-09
Rename createRoot & createBlockingRoot  to unstable_createRoot & unst…

raghav0622:
[concurrent mode docs] Rename createRoot & createBlockingRoot with unstable_createRoot & unstable_createBlockingRoot with refrence to latest changes in experimental apis in pr 
https://github.com/facebook/react/pull/18825

also closes https://github.com/reactjs/reactjs.org/issues/2960#issue-614500217

Issue Number: 2960

### DIFF
--- a/content/docs/concurrent-mode-adoption.md
+++ b/content/docs/concurrent-mode-adoption.md
@@ -71,14 +71,14 @@ import ReactDOM from 'react-dom';
 //
 // You can opt into Concurrent Mode by writing:
 
-ReactDOM.createRoot(
+ReactDOM.unstable_createRoot(
   document.getElementById('root')
 ).render(<App />);
 ```
 
 >Note:
 >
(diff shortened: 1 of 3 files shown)


gaearon:
Hmm why didn't CI trigger a website build..

gaearon:
It's not just `createRoot`, there's also `useTransition`, etc. It gets pretty noisy if you add it everywhere in the docs.

I think what we should do instead is to add a new section that explains each export is behind `unstable_`. And has a map of them.

raghav0622:
> It's not just `createRoot`, there's also `useTransition`, etc. It gets pretty noisy if you add it everywhere in the docs.
> 
> I think what we should do instead is to add a new section in Adopting that explains each export is behind `unstable_`. And has a map of them.

but still the code examples would need unstable_ as it would be ambigious if unstable is omitted from code examples. i'll create a new pr with separate section explaining unstable

PR #2971, opened 2020-05-13
Added regexpfinder app to examples

LaurentDhont:


LaurentDhont:
When would this be added to the website? 
If not what can I improve so it would be added?

PR #2973, opened 2020-05-14
Change links to React lifecycle methods diagram to use HTTPS

wojtekmaj:
Why HTTP 301 if we can have HTTP 200 :)


lex111 (approved):

lex111:
@wojtekmaj thanks!

Issue #2990, opened 2020-05-25
[docs] Outdated react transition group repo link


vuki656:
In the [animation add-ons](https://reactjs.org/docs/animation.html) doc page the links for [react transition group](https://github.com/reactjs/react-transition-group) repo lead to an outdated branch **v1-stable** that is no longer updated.

rachelnabors:
Actually, upon closer inspection, this is an intentional link to help people migrating. The 1.x branch copies this helper's original API. The latest version's API is different, so linking to it would cause confusion.

PR #2991, opened 2020-05-25
Fix outdated links for react transition group

vuki656:
The current links for react transition group lead to the **v1-stable** branch which is no longer updated.

I changed the links to point to master.

#2990

### DIFF
--- a/content/docs/addons-animation.md
+++ b/content/docs/addons-animation.md
@@ -12,7 +12,7 @@ redirect_from:
 
 > Note:
 >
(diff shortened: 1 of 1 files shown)


rachelnabors:
Actually, upon closer inspection, this is an intentional link to help people migrating. The 1.x branch copies this helper's original API. The latest version's API is different, so linking to it would cause confusion :( Ideally we should try to remove this page altogether! But thanks for trying!

PR #3000, opened 2020-05-30
Docs/update forms.md with hooks examples

ugglr:
Since the release of hooks it has been ever clear that we don't really need class based components to handle state, or lifecycle methods. There are many examples of in the documentation where the undertone is clear that hooks is the modern way to write components. 

Therefore I think it should also be reflected in the forms tutorial to include hooks examples.

It's not perfect as it is, but I wanted to do some work on it and then see what the maintainers think first before throwing more work at it.


ugglr:
Why is this PR just left here? No one has commented or given feedback after such a long time?

rickhanlonii:
Thanks for submitting this PR @ugglr and sorry for the delayed review.

We're in the process of updating the docs to be more hooks based, so I'm going to close this in favor of the larger effort. Thanks again for your work on this!

PR #3006, opened 2020-06-01
Merge duplicate props for noopener and noreferrer

nickmccurdy:
I noticed that `rel=` is used twice here, so the resulting HTML would only be `rel="noreferrer"`. This fixes it to be `rel="noopener noreferrer"` for security.


threepointone:
This has already been fixed to just be a single prop https://github.com/reactjs/reactjs.org/blob/master/src/components/LayoutHeader/Header.js#L58 Thanks for the concern!

PR #3013, opened 2020-06-02
<Avatar /> prop

SardorMutalov:
The string <Avatar user = {props.author} /> in the <Avatar /> component is confusing and should look like this:  <Avatar user = {props.user} />

### DIFF
--- a/content/docs/components-and-props.md
+++ b/content/docs/components-and-props.md
@@ -182,7 +182,7 @@ function Comment(props) {
   return (
     <div className="Comment">
       <div className="UserInfo">
-        <Avatar user={props.author} />
+        <Avatar user={props.user} />
         <div className="UserInfo-name">
           {props.author.name}
         </div>


content/docs/components-and-props.md
```diff
@@ -182,7 +182,7 @@ function Comment(props) {
   return (
     <div className="Comment">
       <div className="UserInfo">
-        <Avatar user={props.author} />
+        <Avatar user={props.user} />
```
iamdarshshah, line 185:
```suggestion
        <Avatar user={props.author} />
```
This is good. It doesn't require any changes. I would suggest you go through the documentation first. ([here](https://reactjs.org/docs/components-and-props.html#extracting-components)).

For a better understanding of what's actually happening and why we are using `author` and not `user` in `Comment` component, refer ([this](https://codepen.io/pen?&editable=true&editors=0010)) CodePen.

PR #3018, opened 2020-06-04
fixes color name issue

arihant-2310:


threepointone:
We removed the grayscale treatment a while ago, so the boxes are back to yellow. Thanks for the concern!

PR #3020, opened 2020-06-04
Clarify language in components-and-props.md

rickhanlonii:
This is a nit for a few reasons:
- The point that extracting a component that "is complex enough on its own" means that it's not necessarily reusable.
- Extracting the components listed in the same sentence (e.g. App) are not necessarily reusable.
- This ends the section "Extracting Components" with a reference to extraction instead of re-usability.
- This delimits extracting into a separate component from creating a re-usable component.


sophiebits:
Thanks!

PR #3024, opened 2020-06-05
remove grayscale treatment

threepointone:
As per https://twitter.com/sliminality/status/1269033413488459776, there's feedback that the contrast isn't great for reading. While I myself have trouble with low contrast text, (and long sightedness), and thought it was fine, I was probably mistaken. Removing this for now to get back to a non harmful position, and we can decide later what to do instead.

NOTE: I was unable to run this locally, having a clash with node versions and I'm not able at the moment to fix it. I hope CI generates a preview URL, let's see.


threepointone:
Ok looks like CI still doesn't generate preview URLs. Feel free to check locally.

PR #3026, opened 2020-06-08
"Must Only Be Unique" -> "Only Need To Be Unique"

cincodenada:
In the former, "must only" implies that they can't be unique if not among siblings, which is not the case. Could be "Only Must Be", but that's an awkward construction, so I used "need to be" in parallel with the opening paragraph.

Should I change the anchor text as well? Does that need to be updated elsewhere?

### DIFF
--- a/content/docs/lists-and-keys.md
+++ b/content/docs/lists-and-keys.md
@@ -205,7 +205,7 @@ ReactDOM.render(
 
 A good rule of thumb is that elements inside the `map()` call need keys.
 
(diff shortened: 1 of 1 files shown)


BartoszKlonowski:
Hello @cincodenada!
I don't see this sentence anymore in the latest of react.dev docs, so let me close this PR as outdated.
However, if you still would like to contribute please create new PRs for other open issues.
Thank you!

PR #3039, opened 2020-06-15
added more technically correct answer to "what does it return"

tomprogers:
"What does it [useState] return? A pair of values" is not the best way to describe the facts.

It returns an array, which is a formal datatype. We should tell developers the truth using the most precise, most accurate terminology available. "A pair of values" is not really a thing.


tomprogers:
I'm not signing anything with Facebook.

PR #3040, opened 2020-06-16
Add new codepen link for error boundaries

NorfeldtAbtion:
The new link contains also nested error boundaries. 

@gaearon could also update his code pen - I dont' have access to save an updated version on his codepen.


threepointone:
Why is it necessary to show nested error boundaries in this example?

NorfeldtAbtion:
@threepointone why don't you think it's relevant to know this type of relation? 
To me it shows that you can change your error bounding down the "road"/nesting and that it does not cascade all the way up.

PR #3043, opened 2020-06-17
Translate to Persian cdn-links.md

hshoja:
translate to persian.

### DIFF
--- a/content/docs/cdn-links.md
+++ b/content/docs/cdn-links.md
@@ -6,32 +6,33 @@ prev: create-a-new-react-app.html
 next: release-channels.html
 ---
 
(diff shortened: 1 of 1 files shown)


threepointone:
There is a farsi version of the documentation here https://fa.reactjs.org/ Isn't that enough?

hshoja:
It seems most of the pages have not translated.
eg: https://fa.reactjs.org/docs/introducing-jsx.html

threepointone:
You should submit fixes for that in this repo here https://github.com/reactjs/fa.reactjs.org/

Closing this PR here, but thanks for checking!

PR #3044, opened 2020-06-18
More docs onFocus and onBlur events

ctrlaltdylan:
Including more substantial documentation around the `onFocus` and `onBlur` events that aren't so common sense on first look.

### DIFF
--- a/content/docs/reference-events.md
+++ b/content/docs/reference-events.md
@@ -169,16 +169,16 @@ These focus events work on all elements in the React DOM, not just form elements
 
 The `onFocus` prop is initiated when focus is on the element. For example, when the user clicks on a text input element, the function defined with the `onFocus` prop is called.
 
-```
+```javascript
   function onFocusExample() {
-     return (
+    return (
       <input
         onFocus={(e) => {
           console.log('Focused on input');
         }}
         placeholder="onFocus is triggered when you click this input."
       />
-     )
+    )
   }
 ```
 
@@ -186,17 +186,17 @@ The `onFocus` prop is initiated when focus is on the element. For example, when 
 
 `onBlur` prop is initiated when focus has left on the element. For example, when the user clicks into and then clicks outside of on a text input element, the function defined with the `onBlur` prop is called.
 
-```
-  function onBlurExample() {
-     return (
-      <input
-        onBlur={(e) => {
-          console.log('Triggered because this input lost focus');
-        }}
-        placeholder="onBlur is triggered when you click this input and then you click outside of it."
-      />
-     )
-  }
+```javascript
+function onBlurExample() {
+  return (
+    <input
+      onBlur={(e) => {
+        console.log('Triggered because this input lost focus');
+      }}
+      placeholder="onBlur is triggered when you click this input and then you click outside of it."
+    />
+  )
+}
 ```
 
 Properties:
@@ -208,7 +208,7 @@ DOMEventTarget relatedTarget
 
 You can use the `currentTarget` and `relatedTarget` to differentiate if the focusing or blurring events originated from _outside_ of the parent element:
 
-```
+```javascript
 export default function App() {
   return (
     <div
--- a/content/docs/reference-events.md
+++ b/content/docs/reference-events.md
@@ -165,6 +165,14 @@ onFocus onBlur
 
 These focus events work on all elements in the React DOM, not just form elements.
 
+#### onFocus
+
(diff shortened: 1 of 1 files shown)


content/docs/reference-events.md
```diff
@@ -165,6 +165,14 @@ onFocus onBlur
 
 These focus events work on all elements in the React DOM, not just form elements.
 
+#### onFocus
+
+The `onFocus` prop is initiated when focus is on the element. For example, when the user clicks on a text input element, the function defined with the `onFocus` prop is called.
+
+#### onBlur
+
+`onBlur` prop is initiated when focus has left on the element. For example, when the user clicks into and then clicks outside of on a text input element, the function defined with the `onFocus` prop is called.
```
iamdarshshah, line 174:
```suggestion
`onBlur` prop is initiated when focus has left on the element. For example, when the user clicks into and then clicks outside of on a text input element, the function defined with the `onBlur` prop is called.
```

crypt096 (approved):

gaearon:
How do you feel about incorporating this example? Seems like it’s a useful collection of behaviors people want. https://github.com/facebook/react/issues/6410#issuecomment-671915381

content/docs/reference-events.md
```diff
@@ -165,6 +165,14 @@ onFocus onBlur
 
 These focus events work on all elements in the React DOM, not just form elements.
 
+#### onFocus
+
+The `onFocus` prop is initiated when focus is on the element. For example, when the user clicks on a text input element, the function defined with the `onFocus` prop is called.
```
eps1lon, line 170: That is certainly true for chrome on windows and linux as far as I know. But does not apply to e.g. firefox macOS. 

More specifically: By spec it is up to the user agent (e.g. a browser) to determine whether a clicked element is also focused.

eps1lon (changes requested):

content/docs/reference-events.md
eps1lon: Or it might be true for inputs. I was a bit suspicious because of the "click-focus" association which is a bit tricky on other platforms.

ctrlaltdylan:
> How do you feel about incorporating this example? Seems like it’s a useful collection of behaviors people want. [facebook/react#6410 (comment)](https://github.com/facebook/react/issues/6410#issuecomment-671915381)

@gaearon  Forked your explanation and changed the logging events to be a tad more people friendly and specific. Added a few more examples to cover vanilla `onFocus` and `onBlur` on a solo element.

content/docs/reference-events.md
ctrlaltdylan: @eps1lon I just booted up Firefox + Mac and ran through a sample:

https://www.loom.com/share/34f1a18300b44187b7a65e2c01b73f8b

Sandbox: https://codesandbox.io/s/busy-browser-0me16?file=/src/App.js:146-191

It _appears_ to be 1:1 to Chrome, at least for the basic examples I wanted to document. If you find something else different please feel free to edit/tack on.

content/docs/reference-events.md
```diff
@@ -165,11 +165,82 @@ onFocus onBlur
...
+        }
+        if (!e.currentTarget.contains(e.relatedTarget)) {
+          console.log(
+            "Now focused on parent <div>, not triggered when swapping between children."
+          );
+        }
+      }}
+      onBlur={(e) => {
+        console.log("left focus of", e.target);
+        if (e.currentTarget === e.target) {
```
iamdarshshah, line 229: I think this condition needs to be changed, because `e.currentTarget === e.target` will never be true for `onBlur` events as it is triggered when _self_ loses its focus or when the user leaves a specific field. Thus, this won't be true in any case.  

We can update it to this: 
```suggestion
        if (e.currentTarget !== e.target) {
```
> This will be triggered when we not focus on _self_ element anymore.

iamdarshshah (changes requested):

iamdarshshah (approved):
LGTM! 👍

gaearon:
Code blocks need to be consistently indented and should use JS language

ctrlaltdylan:
@gaearon thanks, missed that. Updated `onBlur` and `onFocus` examples with the `javascript` tag and fixed spacing so it matches the other examples.

content/docs/reference-events.md
gaearon: I'm not sure what you mean but this condition is definitely true when you're blurring the node itself. https://codesandbox.io/s/strange-albattani-7tqr7?file=/src/App.js The proposed edit is wrong and contradicts what the code is supposed to do.

PR #3055, opened 2020-06-21
Update refs-and-the-dom.md

browniefed:


coryhouse:
I'd suggest changing to "You may only use the `ref` attribute on function components with `React.forwardRef`, optionally in conjunction with the `useImperativeHandle` hook."

PR #3067, opened 2020-06-29
Update hooks-custom.md

hiteshagja:


threepointone:
This doesn't seem right, the component still receives props as expected.

PR #3071, opened 2020-06-30
Mention re-throwing from error boundaries

vazitron:
I had a hard time finding out that in development mode, react re-throws errors up from the boundary.
I only found it mentioned in the Sentry.io's docs.

So I thought that adding this to the official docs might be helpful.


OCmilo:
Hey @vazitron, I looked into the error boundaries docs and it is mentioned in the first line [here](https://reactjs.org/docs/error-boundaries.html#component-stack-traces). Let me know if I missed something!

PR #3079, opened 2020-07-03
Update to parentheses

sufferingAnt638:
Removed parentheses


sophiebits:
It's correct as written.

PR #3083, opened 2020-07-05
Fixed a spelling mistake

amoghrajesh:


amoghrajesh:
> This isn’t right, `true && x` will be the value of `x`, not `true`. You might be confusing it with `true || x`, which will always be `true`.

Yes I probably read it too fast. Thanks for correcting me

PR #3085, opened 2020-07-06
code refactor

adityamandhare:
making the setting new state object simpler.


rickhanlonii:
Hey @adityamandhare, thanks for the PR!

For this use case we [actively recommend](https://reactjs.org/docs/faq-state.html#how-do-i-update-state-with-values-that-depend-on-the-current-state) using an updater function to setState to ensure the intended value is set.

PR #3087, opened 2020-07-07
Added Privacy and Terms footer links

bvaughn:
These links don't really fit in to the existing categories very well, so I added them under "More"

Screen Shot 2020-07-07 at 10 58 16 AM


bvaughn:
Going to merge this since it's a requirement. If there are concerns about the placement or UI, we can iterate.

PR #3091, opened 2020-07-08
Change terminology from 'asynchronous' to something else

adamkleingit:
I think it might be confusing to say 'asynchronous', since that usually implies some promise or timeout - outside the current call stack.
I've talked with many developers who think the state updates won't happen during this event loop because state updates are asynchronous.
In fact, that's what I thought until I started experimenting with it and reading the source code.

I suggested a few options, don't know if any of them make sense to you more.

Thanks

### DIFF
--- a/content/docs/faq-state.md
+++ b/content/docs/faq-state.md
@@ -22,7 +22,7 @@ Here are some good resources for further reading on when to use `props` vs `stat
 
 In React, both `this.props` and `this.state` represent the *rendered* values, i.e. what's currently on the screen.
 
-Calls to `setState` are asynchronous - don't rely on `this.state` to reflect the new value immediately after calling `setState`. Pass an updater function instead of an object if you need to compute values based on the current state (see below for details).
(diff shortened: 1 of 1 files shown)


rickhanlonii:
You're correct that today the state updates can be synchronously flushed in the same event loop, but that's not always the case. For example, you can debug clicking the button in [this sandbox](https://codesandbox.io/s/jolly-https-qygzs) that runs concurrent mode the latest experimental build to see state updates happening asynchronously.

It's best to think of setState as running asynchronously, even if the implementation details of a particular update run in the same event loop.

adamkleingit:
Yes, I realize that it might be either. Just saying it was confusing when the docs say asynchronous, and I run a sandbox without concurrent and see that it's synchronous.

rickhanlonii:
It is asynchronous with respect to the setState call, but can be synchronous within event loop. React has used this language for awhile, and in the future it will also be asynchronous within the event loop so I don't think it makes sense to churn the language now.

adamkleingit:
Thanks for the detailed explanation :)

PR #3093, opened 2020-07-09
Update Rollup example for Optimizing Performance

pioug:
- Update the name of the plugins since Rollup has a monorepo for plugins now (https://github.com/rollup/plugins)
- Use named import for terser plugin (https://github.com/TrySound/rollup-plugin-terser#why-named-export)
- Make the configuration more complete. Most examples for configuration files I found are written with ESM (https://rollupjs.org/guide/en/#configuration-files)

I hope those changes are fine 🙏

### DIFF
--- a/content/docs/optimizing-performance.md
+++ b/content/docs/optimizing-performance.md
@@ -104,28 +104,34 @@ For the most efficient Rollup production build, install a few plugins:
 
 ```bash
 # If you use npm
-npm install --save-dev rollup-plugin-commonjs rollup-plugin-replace rollup-plugin-terser
+npm install --save-dev @rollup/plugin-commonjs @rollup/plugin-replace rollup-plugin-terser
 
 # If you use Yarn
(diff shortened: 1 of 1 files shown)


pioug:
In the case of Rollup config files specifically, I would say it is widely used, though it's always transpiled:

> By default, Rollup will expect config files to be ES modules and bundle and transpile them and their relative imports to CommonJS before requiring them. - https://rollupjs.org/guide/en/#using-untranspiled-config-files

threepointone:
Great to hear! (Also, good feature for rollup to have, very nice)

pioug:
Sorry for insisting 💦💦 I was checking that part of the documentation today, and realized that I made a PR to update it 😅 Someone else made a similar PR https://github.com/reactjs/reactjs.org/pull/3489 since then. Let me know if any changes are needed, I can make time for it.

pioug:
Closing because Irrelevant since react.dev

PR #3097, opened 2020-07-10
Update examples on homepage to use hooks

fabiankaegy:
I have updated the examples on the home page to use functional components and hooks rather than class components.


eps1lon:
There's now a larger effort to rewrite the docs with hooks in mind: https://github.com/reactjs/reactjs.org/issues/3308

Until then you can use https://reactwithhooks.netlify.app/ which is a version of the React docs documenting only hooks.

PR #3100, opened 2020-07-11
Use remark-codesandbox for the code examples

CompuIves:
In this PR I have replaced some code examples with [`remark-codesandbox`](https://github.com/kevin940726/remark-codesandbox). In the repo we specify some template projects, which we then can refer to in our code blocks. For example:

image

This code will take the `umd` template (defined in `/codesandbox/examples/umd`), and replace the entry file with this contents, and then create an embed with those contents included. Like this:

image

We can specify as many templates as we want. For example, right now I've defined `umd` (using script tags), `simple` (single file) and `cra` (CRA).

I've also included some examples with line highlights! 😄 

I'm curious what you think, I think another improvement would be to let this plugin replace the code blocks with something created with [`react-sandpack`](https://github.com/codesandbox/codesandbox-client/tree/master/standalone-packages/react-sandpack). The advantages of using this are:

1. They're "just" React components, so you can customize visuals/behaviour as much as you want
2. Faster performance, as there's no iframe to load for the editor
3. Self-hosted, the bundler can be self-hosted here and will allow you to apply versioning to the bundler + if CSB ever would go down, the examples will stay up

I can work on a second version where we use `react-sandpack` if you're interested!


codesandbox/.gitignore
```diff
@@ -0,0 +1 @@
+!*
```
rickhanlonii, line 1: Is this file needed?

codesandbox/examples/cra/package.json
```diff
@@ -0,0 +1,30 @@
+{
```
rickhanlonii, line 1: This is the CRA template, right? Should the dir be /templates not /examples?

codesandbox/examples/cra/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "start": "react-scripts start",
+    "build": "react-scripts build",
+    "test": "react-scripts test --env=jsdom",
+    "eject": "react-scripts eject"
+  },
+  "browserslist": [
+    ">0.2%",
+    "not dead",
+    "not ie <= 11",
+    "not op_mini all"
```
rickhanlonii, line 28: Do we know what will happed to the embedded sandbox when people browse embedded codesandboxes on unsupported browsers?

content/docs/handling-events.md
```diff
@@ -60,8 +60,10 @@ When using React, you generally don't need to call `addEventListener` to add lis
 
 When you define a component using an [ES6 class](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Classes), a common pattern is for an event handler to be a method on the class. For example, this `Toggle` component renders a button that lets the user toggle between "ON" and "OFF" states:
 
-```js{6,7,10-14,18}
-class Toggle extends React.Component {
+```js codesandbox=cra?highlights=8,9,12,13,14,15,16,20&height=400
```
rickhanlonii, line 63: 🤯

bvaughn:
Nice work @CompuIves !

codesandbox/.gitignore
CompuIves: I added this one to "unignore" the public/ folder in the examples.

codesandbox/examples/cra/package.json
CompuIves: Yeah good point! I'll change it to templates, much more descriptive!

Another idea I had is that we could also save the react version in gatsby config, then you only need to change the react version in one file to update all templates.

codesandbox/examples/cra/package.json
CompuIves: Hmm, yes this is a good point. Most probably it will just be a console error, but I can catch this maybe and show a better message. By the way, which browsers do you aim to support with the runnable examples? I'll make sure they'll work in those browsers.

content/docs/handling-events.md
CompuIves: Cool right?! Would be even nicer if we can update the line highlight setting to copy the exact highlights!

codesandbox/examples/cra/package.json
rickhanlonii: That's a good question, these supported browser options seem fair to me but I defer to @gaearon.

PR #3104, opened 2020-07-13
update note about exhaustive deps

### LINKED ISSUE #3076: Document Why a "complex expression in the dependency array" is a Problem
this is about the exhaustive-deps eslint rule, specifically her: https://github.com/facebook/react/blob/master/packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js#L779. i've been trying to find out why having a complex expression inside of the dependency array is bad, but haven't found anything saying why something like `[err?.message]` would cause issues. is there somewhere it's listed in the docs that i haven't seen?

TomPridham:
adds some more context about what the rule guards against as well as a link to a more thorough explanation of why those things are considered harmful

@rickhanlonii, is this what you had in mind? i couldn't find a nice way to include why complex expressions were a problem without expanding on the whole thing

closes #3076


rickhanlonii:
Thanks for updating this @TomPridham. Now that I see it, I don't think this is the right place to explain it because it already links to the lint plugin which is where the description of how the linter works should live (otherwise this doc could get stale).

PR #3110, opened 2020-07-16
Update hooks-overview.md

amodsachintha:
Changed 'is' -> 'are'. A simple grammar fix.

### DIFF
--- a/content/docs/hooks-overview.md
+++ b/content/docs/hooks-overview.md
@@ -229,7 +229,7 @@ function FriendListItem(props) {
 }
 ```
 
-The state of these components is completely independent. Hooks are a way to reuse *stateful logic*, not state itself. In fact, each *call* to a Hook has a completely isolated state -- so you can even use the same custom Hook twice in one component.
+The state of these components are completely independent. Hooks are a way to reuse *stateful logic*, not state itself. In fact, each *call* to a Hook has a completely isolated state -- so you can even use the same custom Hook twice in one component.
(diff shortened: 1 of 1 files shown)


content/docs/hooks-overview.md
```diff
@@ -229,7 +229,7 @@ function FriendListItem(props) {
 }
 ```
 
-The state of these components is completely independent. Hooks are a way to reuse *stateful logic*, not state itself. In fact, each *call* to a Hook has a completely isolated state -- so you can even use the same custom Hook twice in one component.
+The state of these components are completely independent. Hooks are a way to reuse *stateful logic*, not state itself. In fact, each *call* to a Hook has a completely isolated state -- so you can even use the same custom Hook twice in one component.
```
rickhanlonii, line 232: The noun here is "state" which is singular so we should use "is", but I agree that this is awkward.

What do you think about this:

```suggestion
The state of each component is completely independent. Hooks are a way to reuse *stateful logic*, not state itself. In fact, each *call* to a Hook has a completely isolated state -- so you can even use the same custom Hook twice in one component.
```

amodsachintha: I agree, the latter is much more proper.

rickhanlonii:
Thanks @amodsachintha!

amodsachintha:
> Thanks @amodsachintha!

Thank you too @rickhanlonii !

PR #3115, opened 2020-07-17
Hint at the fact that useState is pure

bripkens:
Why
===

When migrating from React classes with state it can easily
be assumed that React classes' state behaves just like
`useState`. However this is not exactly right. The state
updater returned by `useState` does not cause a re-render
when the state value hasn't changed. In that sense it is
more similar to a class extending from `PureComponent`.

Currently it is not mentioned/hinted at the fact that these
are different.

What
====

Add a hint that points at this behavior. For a lot of users
this piece of information is probably irrelevant, but for
users migrating older codebases to hooks this is important
to know.

### DIFF
--- a/content/docs/hooks-reference.md
+++ b/content/docs/hooks-reference.md
@@ -37,7 +37,7 @@ Returns a stateful value, and a function to update it.
 
 During the initial render, the returned state (`state`) is the same as the value passed as the first argument (`initialState`).
 
-The `setState` function is used to update the state. It accepts a new state value and enqueues a re-render of the component.
+The `setState` function is used to update the state. It accepts a new state value and enqueues a re-render of the component when the value has changed.
 
 ```js
 setState(newState);


rickhanlonii:
Thanks for the PR @bripkens!

This is an implementation detail and an optimization that's not always true and shouldn't be relied on. 

I'm going to close this as we don't want to solidify the behavior in the docs.

bripkens:
@rickhanlonii: Right now the end-user visible behavior is that this comparison/"pureness check" is happening. It could therefore be argued that it is not an implementation detail, but instead an end-user noticeable implementation artifact. For everyone else executing a migration to hooks it would be nice if this surprise could be avoided.

Personally I have no gripe with this behavior. It would just be nice if the docs could point this out. Especially since this behavior is different to the established knowledge of many of your end-users (aside: none of our UI engineers were aware of this).

rickhanlonii:
Right now the behavior is that the bailout _could_ happen. It doesn't always happen, so we can't document it as if it always happens. What's a use case where this would break a migration to hooks?

bripkens:
> What's a use case where this would break a migration to hooks?

Our use case has been the translation of a higher-order component that facilitates binding React components to observables (in the RxJS sense – but it could also have been any other external event stream). The observables did not execute any change detection logic and hence every value downstreamed resulted in an eventual re-render. For efficiency sake some hot code paths chose to use mutable objects. With the HOC this was not a problem – with the hook however it was.

Unfortunately we only learned about this difference after the change to hooks was already rolled out to production :(.

To acknowledge this `useState` behavior will currently implemented a helper for impure state:

```js
function useImpureState(initialState) {
  const [[state], setState] = useState([initialState]);
  return [state, v => setState([v])];
}
```

PR #3119, opened 2020-07-19
Authenticate On Reload Using Context API

inblack67:


rickhanlonii:
@inblack67 thanks for submitting the new guide.

This doesn't make sense to exist as a guide in core because of how niche it is, but I encourage you to publish this as a blog post or elsewhere so that others can learn from it!

inblack67:
@rickhanlonii Sure, will do.

PR #3131, opened 2020-07-22
Channel naming was "Stable" instead of "Latest"

anarafaelaferro:


gaearon:
Ah, good catch. Thanks.

PR #3141, opened 2020-07-27
Adding padding around h2 and h3 titles

### LINKED ISSUE #3113: Anchors are hidden underneath the navbar when navigating to them
I noticed that anchor links slide under the navbar. This is similar to https://github.com/facebook/docusaurus/issues/2907 and likely caused by the increase of the height of the navbar caused by the BLM banner.

# Steps to reproduce

1. Create an anchor and a link to the anchor in React docs
2. Publish the website
3. Tap the anchor link

# Expected result
The anchor appears in view at the top of the page

# Actual result
The anchor appears at the top of the page hidden under the navbar

Screen Shot 2020-07-16 at 9 39 25 PM

kwaimind:
This PR offers a fix to close #3113 and adds padding around `h2` and `h3` elements. These were previously hidden under the header navigation.

Since this change is to the `theme.js` file, take a look around the rest of the site and check you are happy with how the padding affects other pages and headlines.

I've also add some intro text for "Basic hooks". This gives a consistent intro just like the "Additional Hooks" section, but also adds some padding between the section title and the useState content. The text comes from [here](https://reactjs.org/docs/hooks-overview.html#:~:text=Hooks%20are%20functions%20that%20let,you%20use%20React%20without%20classes.&text=You%20can%20also%20create%20your,stateful%20behavior%20between%20different%20components.) and some of my own wording. I'm happy to change this as needed.

> Hooks are functions that let you “hook into” React state and lifecycle features from function components. Here are 3 of the most basic and most often used Hooks in React.

Below are some screenshots to show how these new padding values look on desktop and mobile.

### Desktop
Screenshot 2020-07-27 at 11 24 52

### Mobile
Screenshot 2020-07-27 at 11 25 00

### DIFF
--- a/content/docs/hooks-reference.md
+++ b/content/docs/hooks-reference.md
@@ -27,6 +27,8 @@ If you're new to Hooks, you might want to check out [the overview](/docs/hooks-o
 
 ## Basic Hooks {#basic-hooks}
 
+Hooks are functions that let you “hook into” React state and lifecycle features from function components. Here are 3 of the most basic and most often used Hooks in React.
+
 ### `useState` {#usestate}
 
 ```js
--- a/src/theme.js
+++ b/src/theme.js
@@ -320,7 +320,7 @@ const sharedStyles = {
       '::before': {
         content: ' ',
         display: 'block',
-        paddingTop: 90,
+        paddingTop: 120,
         marginTop: -45,
       },
 
--- a/src/theme.js
+++ b/src/theme.js
@@ -287,7 +287,7 @@ const sharedStyles = {
         content: ' ',
         display: 'block',
         borderBottom: `1px solid ${colors.divider}`,
-        paddingTop: 44,
+        paddingTop: 85,
         marginBottom: 40,
       },
 
@@ -320,7 +320,7 @@ const sharedStyles = {
       '::before': {
         content: ' ',
         display: 'block',
-        paddingTop: 120,
+        paddingTop: 125,
         marginTop: -45,
       },
 


kwaimind:
☝️ I've signed the CLA now. Hopefully this will recheck itself.

content/docs/hooks-reference.md
```diff
@@ -27,6 +27,8 @@ If you're new to Hooks, you might want to check out [the overview](/docs/hooks-o
 
 ## Basic Hooks {#basic-hooks}
 
+Hooks are functions that let you “hook into” React state and lifecycle features from function components. Here are 3 of the most basic and most often used Hooks in React.
```
iamdarshshah, line 30:
```suggestion
Hooks are functions that `hook` into a functional component and allow us to use state and component features like `componentDidUpdate`, `componentDidMount`, and more. Here are 3 of the most basic and most often used Hooks to React.
```

kwaimind: Thanks for the suggestion. I guess we don't need to reference the old lifecycle methods since hooks is part of the new roadmap rather than a replacement if you see what I mean.

Happy to see what others say about this suggestion.

iamdarshshah:
```suggestion
Hooks are functions that `hook` into a functional component and allow us to use state and component lifecycle methods. Here are 3 of the most basic and most often used Hooks to React.
```
Okay, If we don't need to reference the old lifecycle methods, then we can update the content to the one mentioned above. This will make it grammatically correct and will help others to understand the concept clearly.

Would like to know your thoughts on this.

Himachallad: Personally I feel that the actual statement looks sleeker and on point.

rickhanlonii: I see what you're going for but since this is the Reference doc I don't think this is necessary. The hooks guide already covers this.

PR #3145, opened 2020-07-28
clarify hook language a bit

nyurik:
Minor doc cleanup

### DIFF
--- a/content/docs/hooks-reference.md
+++ b/content/docs/hooks-reference.md
@@ -47,7 +47,7 @@ During subsequent re-renders, the first value returned by `useState` will always
 
 >Note
 >
->React guarantees that `setState` function identity is stable and won't change on re-renders. This is why it's safe to omit from the `useEffect` or `useCallback` dependency list.
+>React guarantees that `setState` function identity is stable and won't change on re-renders. This is why it is safe to omit it from the `useEffect` or `useCallback` dependency list.
 
 #### Functional updates {#functional-updates}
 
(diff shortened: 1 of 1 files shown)


crypt096 (approved):

content/docs/hooks-reference.md
```diff
@@ -47,7 +47,7 @@ During subsequent re-renders, the first value returned by `useState` will always
 
 >Note
 >
->React guarantees that `setState` function identity is stable and won't change on re-renders. This is why it's safe to omit from the `useEffect` or `useCallback` dependency list.
+>React guarantees that `setState` function identity is stable and won't change on re-renders. This is why it is safe to omit it from the `useEffect` or `useCallback` dependency list.
```
eps1lon, line 50:
```suggestion
>React guarantees that `setState` function identity is stable and won't change on re-renders. This is why it is safe to omit the `setState` function from the `useEffect` or `useCallback` dependency list.
```

content/docs/hooks-reference.md
```diff
@@ -286,7 +286,7 @@ function Counter() {
 
 >Note
 >
->React guarantees that `dispatch` function identity is stable and won't change on re-renders. This is why it's safe to omit from the `useEffect` or `useCallback` dependency list.
+>React guarantees that `dispatch` function identity is stable and won't change on re-renders. This is why it is safe to omit it from the `useEffect` or `useCallback` dependency list.
```
eps1lon, line 289:
```suggestion
>React guarantees that `dispatch` function identity is stable and won't change on re-renders. This is why it is safe to omit the `setState` function from the `useEffect` or `useCallback` dependency list.
```

eps1lon (approved):
Let's be specific about what `it` refers to

PR #3147, opened 2020-07-29
Make og:type article for markdown pages

### LINKED ISSUE #3146: og:type is hardcoded as website, even for articles
On blog posts like https://reactjs.org/blog/2018/11/27/react-16-roadmap.html the `og:type` is hard coded as `website`.
https://github.com/reactjs/reactjs.org/blob/3e4f2687c6fd16dbeb075fee1a08c469093dac4e/src/components/TitleAndMetaTags/TitleAndMetaTags.js#L52

This is semantically incorrect and breaks @chickaree
https://chickar.ee/reactjs.org/YmxvZy8yMDE4LzExLzI3L3JlYWN0LTE2LXJvYWRtYXAuaHRtbA
here is an example from @nytimes of what it should look like:
https://chickar.ee/www.nytimes.com/MjAyMC8wNy8yOC9hcnRzL3RlbGV2aXNpb24vZW1teS1ub21pbmVlcy1saXN0Lmh0bWw

If this could be updated for blog posts to be `article` that would be awesome. :)

davidbarratt:
Fixes #3146

### DIFF
--- a/src/components/MarkdownPage/MarkdownPage.js
+++ b/src/components/MarkdownPage/MarkdownPage.js
@@ -74,6 +74,7 @@ const MarkdownPage = ({
       }}>
       <TitleAndMetaTags
         ogDescription={ogDescription}
+        ogType="article"
         canonicalUrl={createCanonicalUrl(markdownRemark.fields.slug)}
         title={`${titlePrefix}${titlePostfix}`}
       />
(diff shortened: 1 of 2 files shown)


threepointone:
This looks good to me, merging. Looks like we still don't have the previews working so I can't verify this, but I hope you've verified it on your end and made sure it doesn't appear for pages it shouldn't. Thank you very much!

PR #3157, opened 2020-07-30
Make example finished

Azatik1000:
In the edited example we created a method "handleSubmit" but didn't use it. I fixed it.


rickhanlonii:
Hey @Azatik1000, thanks for the PR.

The `handleSubmit` handler exists on the `form` element as:

```js
<form onSubmit={this.handleSubmit}>
```

When the input button is clicked, it directs the onSubmit event of the form element which calls handleSubmit.

I think this is working as expected so I'm going to close but let me know if you have any questions!

PR #3159, opened 2020-07-30
Updating the docs to reflect the fact that React.memo can be used wit…

ktajpuri:
Updating the docs to reflect the fact that React.memo can be used with both class and functional components

Currently it incorrectly conveys that React.memo can only be used with functional components.
https://reactjs.org/docs/react-api.html#reactmemo
details #13937

### DIFF
--- a/content/docs/reference-react.md
+++ b/content/docs/reference-react.md
@@ -124,9 +124,9 @@ const MyComponent = React.memo(function MyComponent(props) {
 });
 ```
 
-`React.memo` is a [higher order component](/docs/higher-order-components.html). It's similar to [`React.PureComponent`](#reactpurecomponent). It can be used with both functional components and classes.
+`React.memo` is a [higher order component](/docs/higher-order-components.html).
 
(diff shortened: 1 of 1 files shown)


content/docs/reference-react.md
```diff
@@ -124,7 +124,7 @@ const MyComponent = React.memo(function MyComponent(props) {
 });
 ```
 
-`React.memo` is a [higher order component](/docs/higher-order-components.html). It's similar to [`React.PureComponent`](#reactpurecomponent) but for function components instead of classes.
+`React.memo` is a [higher order component](/docs/higher-order-components.html). It's similar to [`React.PureComponent`](#reactpurecomponent). It can be used with both functional components and classes.
```
gaearon, line 127: Let's just drop this paragraph altogether?

ktajpuri: ok

ktajpuri: @gaearon does this look ok now?

ktajpuri:
@gaearon can you have a look please?

PR #3176, opened 2020-08-06
fix(docs): remove unnecessary import

rahulbhooteshwar:
Code snippet for Module Mocking guide under testing recipes has unnecessary import as Mocked module import is not required in test files.


rahulbhooteshwar:
I have signed the Facebook CLA now. How can I re run the checks?

PR #3182, opened 2020-08-07
Fix typos: Change <> to <div>

naxty:


gaearon:
This is not a typo, it's a syntax for Fragments.

PR #3188, opened 2020-08-10
add empty dependencies to useEffect in custom hook

justin-hackin:
The useEffect in https://reactjs.org/docs/hooks-custom.html#extracting-a-custom-hook is missing the empty dependency array


BulatDashiev:
Should the dependency be actually a 'friendID', so example where we selecting 'recipientID' works correctly?

PR #3190, opened 2020-08-10
Use modern syntax for a capture-phase event listener example: flags object rather than raw boolean

craigkovatch:
Using the flags object is still compatible with old browsers (evals to truthy), but makes the intent of the developer more clear, and allows for other flag additions (e.g. `passive`)


craigkovatch:
I have signed the CLA but not sure how to re-trigger the CLA check.

PR #1199, opened 2018-09-25
Update all Babel links to v7 in "Quickly Try JSX"

bodazhao:
**What's changed?**

Updated the Babel links to v7 in "Add React to a Website" section on the website

PR #3202, opened 2020-08-14
Upgrade Babel to v7 in Quick Try JSX

bodaso:
new pull request in place of [#1199](https://github.com/reactjs/reactjs.org/pull/1199)

### DIFF
--- a/content/docs/add-react-to-a-website.md
+++ b/content/docs/add-react-to-a-website.md
@@ -149,14 +149,14 @@ return (
 
 These two code snippets are equivalent. While **JSX is [completely optional](/docs/react-without-jsx.html)**, many people find it helpful for writing UI code -- both with React and with other libraries.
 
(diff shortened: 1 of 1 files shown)


content/docs/add-react-to-a-website.md
```diff
@@ -170,7 +170,7 @@ Adding JSX to a project doesn't require complicated tools like a bundler or a de
 Go to your project folder in the terminal, and paste these two commands:
 
 1. **Step 1:** Run `npm init -y` (if it fails, [here's a fix](https://gist.github.com/gaearon/246f6380610e262f8a648e3e51cad40d))
-2. **Step 2:** Run `npm install babel-cli@6 babel-preset-react-app@3`
```
bodazhao, line 173: Had troubles of using the latest version of `babel-preset-react-app`, mostly because the default compiled JS to ESM so users would have to add `script type="module"` and run index.html with local server like `serve`. It's just easier just to use `@babel/preset-react` instead and there's no issues.

crypt096 (approved):

content/docs/add-react-to-a-website.md
gaearon: Hmm, it was using the `app` preset was very intentional because it has a bunch of other plugins. I think there's a config option for the preset that disables the ESM?

PR #3207, opened 2020-08-17
Remove event pooling and SyntheticEvent#persist from documents

koba04:
As the blog mentioned, React has stopped event pooling, and the `SyntheticEvent#persist` mthod does nothing in v17, so I remove the description from the documentation.

https://reactjs.org/blog/2020/08/10/react-v17-rc.html#no-event-pooling

~**v17 is still in RC, so this PR might have to be merged after releasing v17 stable.**~

Does React have a plan to deprecate and remove the `SyntheticEvent#persist` method in the future??

### DIFF
--- a/content/docs/reference-events.md
+++ b/content/docs/reference-events.md
@@ -26,10 +26,15 @@ void preventDefault()
 boolean isDefaultPrevented()
 void stopPropagation()
 boolean isPropagationStopped()
+void persist()
 DOMEventTarget target
 number timeStamp
 string type
 ```
+
+> Note:
+>
+> As of v17, `e.persist()` doesn't do anything because the `SyntheticEvent` is no longer pooled.
 
 > Note:
 >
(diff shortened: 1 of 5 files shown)


gaearon:
There's a link in `_redirects` that links to this section. I think we need to add a separate page (unlisted) for this warning now.

koba04:
@gaearon 
Thank you. I've added a page for the legacy event pooling and updated the redict link.

image

content/docs/reference-events.md
```diff
@@ -26,7 +26,6 @@ void preventDefault()
 boolean isDefaultPrevented()
 void stopPropagation()
 boolean isPropagationStopped()
-void persist()
```
gaearon, line 29: Technically it's still there so it should be in the API reference. Just needs to note it doesn't do anything.

koba04: @gaearon 
Thank you. I've added the note as another note at https://github.com/reactjs/reactjs.org/pull/3207/commits/89a9c297799683b4bf159774f03c0c2178438732.

koba04:
React 17 has been released, so this PR is ready for review.

gaearon:
Can you please rebase and then also mirror changes in `vercel.json`?

koba04:
I've rebased and modified a redirect link in `vercel.json`

PR #3208, opened 2020-08-17
Update state-and-lifecycle.md

alinaf:
update punctuation for clarity, sounds like a dependent clause without this comma

### DIFF
--- a/content/docs/state-and-lifecycle.md
+++ b/content/docs/state-and-lifecycle.md
@@ -312,7 +312,7 @@ Let's quickly recap what's going on and the order in which the methods are calle
 
 3) When the `Clock` output is inserted in the DOM, React calls the `componentDidMount()` lifecycle method. Inside it, the `Clock` component asks the browser to set up a timer to call the component's `tick()` method once a second.
 
-4) Every second the browser calls the `tick()` method. Inside it, the `Clock` component schedules a UI update by calling `setState()` with an object containing the current time. Thanks to the `setState()` call, React knows the state has changed, and calls the `render()` method again to learn what should be on the screen. This time, `this.state.date` in the `render()` method will be different, and so the render output will include the updated time. React updates the DOM accordingly.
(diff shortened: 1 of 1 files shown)


content/docs/state-and-lifecycle.md
```diff
@@ -312,7 +312,7 @@ Let's quickly recap what's going on and the order in which the methods are calle
...
 
-4) Every second the browser calls the `tick()` method. Inside it, the `Clock` component schedules a UI update by calling `setState()` with an object containing the current time. Thanks to the `setState()` call, React knows the state has changed, and calls the `render()` method again to learn what should be on the screen. This time, `this.state.date` in the `render()` method will be different, and so the render output will include the updated time. React updates the DOM accordingly.
+4) Every second, the browser calls the `tick()` method. Inside it, the `Clock` component schedules a UI update by calling `setState()` with an object containing the current time. Thanks to the `setState()` call, React knows the state has changed, and calls the `render()` method again to learn what should be on the screen. This time, `this.state.date` in the `render()` method will be different, and so the render output will include the updated time. React updates the DOM accordingly.
```
rickhanlonii, line 315: What about 
```suggestion
4) The browser calls the `tick()` method every second. Inside it, the `Clock` component schedules a UI update by calling `setState()` with an object containing the current time. Thanks to the `setState()` call, React knows the state has changed, and calls the `render()` method again to learn what should be on the screen. This time, `this.state.date` in the `render()` method will be different, and so the render output will include the updated time. React updates the DOM accordingly.
```

alinaf: do you think in the second sentence, it now seems like it ("Inside it") refers to "the browser" instead of "the tick() method"? if not, LGTM :) thanks!

rickhanlonii: I don't think it refers to the browser any more than the previous version.

PR #3212, opened 2020-08-19
Update link for Glitch: React Starter Kit

ankitshaw42:
Remove a trailing slash in link which was causing redirect to a not found page.


hg-pyun (approved):
LGTM 👍

PR #3217, opened 2020-08-20
Mentioned that `value` is supported in `<select>`

phistuck:
Clarified the documentation.


eps1lon (approved):
Thanks!

eps1lon:
@phistuck Could you sign this please? Otherwise I'd have to revert this PR.

New to this repo and I wasn't aware this was a required check.

phistuck:
@eps1lon - hehe, slow bots are weird.
I signed it now. But I would not be surprised if they have an issue with my application.

phistuck:
Weird, it is already in production.

PR #3218, opened 2020-08-20
Fix wrong highlighted line numbers in nested Profiler example

dtinth:
**Before:** The Profiler component is not highlighted.

image

**After:**

No more deploy previews? 🥺


lex111 (approved):
Seems legit.

lex111:
@dtinth thanks!

PR #3219, opened 2020-08-20
Use implicit return in setState function example

sktguha:
Use implicit return in setState function example, instead of the extended return
In fact official eslint, https://eslint.org/docs/rules/arrow-body-style, auto-fixes the extended return to the implicit return by default.
I think it is a good thing to teach people who are reading the react docs best practises.

### DIFF
--- a/content/docs/faq-state.md
+++ b/content/docs/faq-state.md
@@ -59,10 +59,8 @@ Passing an update function allows you to access the current state value inside t
 
 ```jsx
 incrementCount() {
-  this.setState((state) => {
-    // Important: read `state` instead of `this.state` when updating.
-    return {count: state.count + 1}
-  });
+  // Important: read `state` instead of `this.state` when updating.
+  this.setState((state) => ({ count: state.count + 1 }));
 }
 
 handleSomething() {


gaearon:
Thanks, but:

1. https://eslint.org/docs/rules/arrow-body-style is merely a stylistic rule, which means ESLint has no opinion about it. You can tell ESLint to enforce it in either direction; the choice depends on your lint config.
2. We actually somewhat discourage people from using stylistic rules like this.
3. It is a lot more confusing if you forget the parenthesis since now it will return `undefined`. So we very intentionally want to keep the existing style here.

sktguha:
Thanks, Dan for the quick response.
"We actually somewhat discourage people from using stylistic rules like this."
curious to know why so?
Because I feel that with an implicit return, the reader immediately knows that there is only one statement in the function and that it is not doing anything else, whereas with an explicit return they have to read through the full body.

sktguha:
> hmm ok cool
> Also Dan @gaearon , I wanted your advice on a bit unrelated topic,
> so in react when debugging, renders are asynchronous. so it if we have
> Class Parent {
> render(){
> if condition 1
>  Child
> else if condition 2
>   Child
> }
> }
> 
> now if I have put a debugger in child's render method, it is a bit difficult to see from exactly where the render method came from.
> For example if the stack trace is
> somFunctionCallInParent which does setState of Parent > render method of Parent > conditon 2 >render method of child
> 
> but in chrome devtools it is
> some stack from react internals , as setstate calls are batched > render method of child
> 
> So it is difficult to trace where the render is happening from and who all changed the state.
> So only for debugging purpose, is it possible to temporary force react to make all setState calls synchronous(i.e doing a setState , synchronously calls the render method immediately instead of getting batched ) by changing temporarily some files in React internals? Or is there some other way to get the proper stack trace. I am only asking this to be temporary be enabled while debugging. Of course the developer has responsibility to check whether their fix/solution is working correctly in normal setstate batching mode also

hi dan @gaearon do you mind giving your opinions on the above?

PR #3228, opened 2020-08-27
Update lists-and-keys.md

cmcodes1:


content/docs/lists-and-keys.md
```diff
@@ -138,7 +138,7 @@ Here is an [in-depth explanation about why keys are necessary](/docs/reconciliat
 
 Keys only make sense in the context of the surrounding array.
 
-For example, if you [extract](/docs/components-and-props.html#extracting-components) a `ListItem` component, you should keep the key on the `<ListItem />` elements in the array rather than on the `<li>` element in the `ListItem` itself.
+For example, if you [extract](/docs/components-and-props.html#extracting-components) a `ListItem` component, you should keep the key of the `<ListItem />` elements in the array rather than on the `<li>` element in the `ListItem` itself.
```
eps1lon, line 141: I think this was correct. The idea is that you end up with

```diff
+function ListItem(props) {
+  return <li {...props} />;
+}

...
-  return <li key={index}>{value}</li>
+  return <ListItem key={index}>{value}</ListItem>
```

Which also matches the second "on the" in this sentence.

cmcodes1: Got it! 👍

PR #3234, opened 2020-08-31
Add :target styles to avoid headlines underneath the header

timkraut:
When opening a link like https://reactjs.org/docs/getting-started.html#try-react, the headline "Try React" is underneath the header and thus not visible:
grafik

This PR adds styles that only apply to the :target selector to avoid this issue. By using a negative top margin and a positive top padding of the same amount, this causes no visible layout changes.

With these styles, the page looks as follows depending on the break point:
grafik

grafik

grafik

grafik

grafik

grafik

### DIFF
--- a/gatsby-browser.js
+++ b/gatsby-browser.js
@@ -14,6 +14,7 @@ require('normalize.css');
 require('./src/css/reset.css');
 require('./src/prism-styles');
 require('./src/css/algolia.css');
+require('./src/css/target.css');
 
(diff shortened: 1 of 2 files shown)


BartoszKlonowski:
Hello @timkraut, thank you for taking the time to contribute!
With the react.dev website for the docs, I find this PR to be no longer required, so I will close this as outdated.
However, feel invited to create new PR for other open issues.

PR #3235, opened 2020-09-01
Fix Navbar covering heading while using anchor in docs

kavin25:
# This PR is a solution to the issue #3113 
I managed to solve the issue by using scroll margin top (equal to the navbar height which I found out was same for all devices=120)

Screenshot 2020-09-01 at 11 45 52 AM
^On laptop(large devices)

Screenshot 2020-09-01 at 11 46 49 AM
^On mobile

### DIFF
--- a/src/theme.js
+++ b/src/theme.js
@@ -12,469 +12,468 @@
 import hex2rgba from 'hex2rgba';
 
 const colors = {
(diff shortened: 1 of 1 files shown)


kavin25:
I just realised I repeated code

kavin25:
👍

PR #3236, opened 2020-09-01
Update hooks-overview.md

hyperoceanic:
If you're running in a sandbox, the changes to document.title are not apparent. Replacing the hard-coded string inside the rendered page with the documents title property makes it more plain what is going on.

### DIFF
--- a/content/docs/hooks-overview.md
+++ b/content/docs/hooks-overview.md
@@ -90,7 +90,7 @@ function Example() {
 
   return (
     <div>
-      <p>You clicked {count} times</p>
+      <p>{document.title}</p>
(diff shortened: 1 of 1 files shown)


eps1lon (changes requested):
This would always commit an outdated title though:
1. initial render `<p></p>`
2. set title to `You clicked 0 times`
3. click button sets `count` to `1`
4. render `<p>You clicked 0 times`
5. set title to `You clicked 1 times`

React won't automatically re-render if `document.title` changes.

hyperoceanic:
@eps1lon - you're right - thanks!

PR #3237, opened 2020-09-02
Explain where act comes from

koba04:
I found this page when I search how to use `act()`, but this page didn't tell me where `act()` come from, so I had to search further to find out.

https://reactjs.org/docs/testing-recipes.html#act

I think adding the import statement is helpful for readers.

### DIFF
--- a/content/docs/testing-recipes.md
+++ b/content/docs/testing-recipes.md
@@ -57,7 +57,7 @@ You may use a different pattern, but keep in mind that we want to execute the cl
 
 ### `act()` {#act}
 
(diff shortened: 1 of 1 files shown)


eps1lon (approved):
This is better than before. Technically it should mention that the `act` must be used from the corresponding renderer but if people copy this import and use the wrong `act` they get at least a descriptive error.

I'll follow-up with a note about different renderers.

eps1lon (changes requested):
Actually the following tests include where this comes from. In this particular code section it isn't that important. What do you think about

```diff
- React provides a helper called `act()`
+ `react-dom/test-utils` provides a helper called [`act()`](/docs/test-utils.html#act)

koba04:
@eps1lon 
> Actually the following tests include where this comes from. In this particular code section it isn't that important. What do you think about

That makes sense for me, so I've update my PR.

eps1lon (approved):
Thanks!

koba04:
Thanks!

PR #3238, opened 2020-09-02
Update hooks-effect.md

Arsikod:
added info about implementation of useEffect with async/await.
Because it is a common question that many react dev ask.


crypt096 (approved):

Arsikod:
Should I close this pr?

Arsikod:
Hey guys I know you are busy, but IMO this explanation is really needed in react docs.
Even I find myself coming back to this PR to check how I was fetching asynchronously from useEffect.

PR #3242, opened 2020-09-04
Phrase adjustment

myronrezzo:
Adjust the phrase about direct assignment to 'this.state' (line 339) to make it clearer.


myronrezzo:
The only place where you can directly assign to `this.state` is in the constructor.

PR #3243, opened 2020-09-04
Phrase adjustment

myronrezzo:
Adjust the phrase about direct assignment to 'this.state' to make it clearer.

### DIFF
--- a/content/docs/state-and-lifecycle.md
+++ b/content/docs/state-and-lifecycle.md
@@ -336,7 +336,7 @@ Instead, use `setState()`:
 this.setState({comment: 'Hello'});
 ```
 
-The only place where you can assign `this.state` is the constructor.
+The only place where you can directly assign to `this.state` is in the constructor.
 
 ### State Updates May Be Asynchronous {#state-updates-may-be-asynchronous}
 


crypt096 (approved):

content/docs/state-and-lifecycle.md
```diff
@@ -336,7 +336,7 @@ Instead, use `setState()`:
 this.setState({comment: 'Hello'});
 ```
 
-The only place where you can assign `this.state` is the constructor.
+The only place where you can directly assign to `this.state` is in the constructor.
```
eps1lon, line 339: Not sure this makes it clearer. Why is "directly" important here i.e. what is "indirectly assigning"?

eps1lon (changes requested):

content/docs/state-and-lifecycle.md
myronrezzo: Hello, @eps1lon.
**"directly"** is intended to emphasize that the assignment will actually be made to this.state and not via this.setState. It may seem redundant, but I see that it reinforces (especially for beginners) what really needs to be done.

But what really motivated this change request was the phrase **"assign` this.state`"** instead of **"assign to `this.state`"**. These are sentences with different semantics and I see that "assign to `this.state` " is the correct one in the context in which the sentence is found (something is assigned **to** ` this.state`, not `this.state` is assigned).

eps1lon: I'm undecided. Seems like we're adding words which could introduce additional questions. If somebody else wants to merge I won't argue against it.

PR #3245, opened 2020-09-06
Fix state update that might cause race condition

ugultopu:
In React, state updates might be asynchronous. Hence, when updating state based on old state or props, `setState` should receive a function as its argument instead of an object.

More information: https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous

### DIFF
--- a/content/docs/hooks-state.md
+++ b/content/docs/hooks-state.md
@@ -47,7 +47,7 @@ class Example extends React.Component {
     return (
       <div>
         <p>You clicked {this.state.count} times</p>
-        <button onClick={() => this.setState({ count: this.state.count + 1 })}>
+        <button onClick={() => this.setState( ({count}) => ({count: count + 1}) )}>
           Click me
         </button>
       </div>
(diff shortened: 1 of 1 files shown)


eps1lon (changes requested):
This is an introductory section that is about equivalence to the hooks example above. It was likely deliberate to not use functional state updates at this point.

ugultopu (commented):
Regardless, it should not encourage bad practices. Either the recommendation about using function parameter version `setState` when previous state is being referred should be changed, or this example needs to be fixed.

crypt096 (approved):

ugultopu:
Can this either be rejected or be merged?

eps1lon:
This should be adressed by https://github.com/reactjs/reactjs.org/issues/3308

eps1lon:
It's just that this is an introductory section so we need to be careful with the amount of React API we expose. But at the same time the example is problematic. 

The question is if we want to experiment with the introductory section when there's already work going on.

PR #3250, opened 2020-09-09
Reduced the height taken by BLM support message

Royal-lobster:
The BLM Support message is a bit big for people having wide screen monitors and for beginners who spend most of their time on the documentation. So, I reduced the height by very few pixels.

I donated to BLM initiative by the way 😅😅(Even though I live in India) and I spend most of my time with the documentation (still learning). 

I Don't know if the code I changed is enough... but please reduce the size a bit 😭


murashki:
It would not be bad if reactjs.org remained solely about technology.

lex111:
Not relevant anymore

PR #3252, opened 2020-09-10
Update hooks-reference.md: The update function should receive the "previous value"

othmanee:
Functional Update:
The update function should receive the "previous value":

  <button onClick={(prevCount) => setCount(prevCount => prevCount - 1)}>-</button>
  <button onClick={(prevCount) => setCount(prevCount => prevCount + 1)}>+</button>

### DIFF
--- a/content/docs/hooks-reference.md
+++ b/content/docs/hooks-reference.md
@@ -60,8 +60,8 @@ function Counter({initialCount}) {
     <>
       Count: {count}
       <button onClick={() => setCount(initialCount)}>Reset</button>
-      <button onClick={() => setCount(prevCount => prevCount - 1)}>-</button>
-      <button onClick={() => setCount(prevCount => prevCount + 1)}>+</button>
+      <button onClick={(prevCount) => setCount(prevCount => prevCount - 1)}>-</button>
+      <button onClick={(prevCount) => setCount(prevCount => prevCount + 1)}>+</button>
     </>
   );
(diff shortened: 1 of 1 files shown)


content/docs/hooks-reference.md
```diff
@@ -60,8 +60,8 @@ function Counter({initialCount}) {
     <>
       Count: {count}
       <button onClick={() => setCount(initialCount)}>Reset</button>
-      <button onClick={() => setCount(prevCount => prevCount - 1)}>-</button>
-      <button onClick={() => setCount(prevCount => prevCount + 1)}>+</button>
+      <button onClick={(prevCount) => setCount(prevCount => prevCount - 1)}>-</button>
+      <button onClick={(prevCount) => setCount(prevCount => prevCount + 1)}>+</button>
```
eps1lon, line 64: That wouldn't work. The first argument of event callbacks is the `event` itself. `prevCount` is passed by React itself when calling functional updates. That is we wrote `setCount(prevCount => prevCount + 1)` and not `setCount(prevCount + 1)`.

eps1lon (changes requested):

eps1lon:
Closing since the current code is working as intended.

content/docs/hooks-reference.md
othmanee: Yes, i didnt see it.

PR #3254, opened 2020-09-12
Clarify that functional components can access state with hooks

marcobarrera:
Clarify that functional components can access state without needed to refactor into ES6 class.
code

### DIFF
--- a/content/docs/state-and-lifecycle.md
+++ b/content/docs/state-and-lifecycle.md
@@ -70,7 +70,7 @@ ReactDOM.render(
 
 To implement this, we need to add "state" to the `Clock` component.
 
-State is similar to props, but it is private and fully controlled by the component.
+State is similar to props, but it is private and fully controlled by the component. It is not necessary to convert the functional component to a ES6 class in order to set and access state. State Hooks can be used to use state without writing a class component.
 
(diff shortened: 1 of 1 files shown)


content/docs/state-and-lifecycle.md
```diff
@@ -70,7 +70,8 @@ ReactDOM.render(
 
 To implement this, we need to add "state" to the `Clock` component.
 
-State is similar to props, but it is private and fully controlled by the component.
+State is similar to props, but it is private and fully controlled by the component. It is not necessary to convert the functional component to a ES6 class in order to set and access state. State Hooks can be used to use state without writing a class component.
```
eps1lon, line 73:
```suggestion
State is similar to props, but it is private and fully controlled by the component. It is not necessary to convert the function component to a ES6 class in order to set and access state. State Hooks can be used to use state without writing a class component.
```

See https://reactjs.org/docs/components-and-props.html#function-and-class-components

eps1lon (changes requested):
I think this would just add confusion. If it's not necessary why would the very next part talk about conversion to a class component?

A big picture approach is probably better suited. Let's revisit this after the big docs rewrite Dan and Rachel are doing.

PR #3262, opened 2020-09-16
mention onload when dom size calc

holylander:
When working with images and trying to get its size on react, you may need to wait until it is loaded by the browser, in order to get the correct size.

### DIFF
--- a/content/docs/hooks-faq.md
+++ b/content/docs/hooks-faq.md
@@ -490,9 +490,7 @@ We didn't choose `useRef` in this example because an object ref doesn't notify u
 
 Note that we pass `[]` as a dependency array to `useCallback`. This ensures that our ref callback doesn't change between the re-renders, and so React won't call it unnecessarily.
 
-Note that you may need to leverage the `onLoad` event, when working with certain type of content, as images. 
-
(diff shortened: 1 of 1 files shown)


eps1lon (changes requested):
I think this would be a perfect example for when a component resizes in the next paragraph. I don't think this section of the docs needs to document each possible workaround for observing element size changes.

holylander:
thanks @eps1lon. Im not 100% sure If I understood you well :) Where should I move this comment then? Or shall we just abandon the change? thanks !

eps1lon:
Check out the next paragraph that talks about the problem in general:

> In this example, the callback ref will be called only when the component mounts and unmounts, since the rendered `<h1>` component stays present throughout any rerenders. If you want to be notified any time a component resizes, you may want to use [`ResizeObserver`](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) or a third-party Hook built on it.

Your paragraph describes a more specific case. Talking about the specific and then general case is backwards in my opinion. 

I'm not familiar with the problem space itself so using ResizeObserve for layout shifts due to image loading may be overkill. Though I don't think this documentation is a good place to talk about all the possible approaches to detecting that an element's dimensions changed.

holylander:
Makes sense. I have move the `onload` question to the paragraph that contains the example usecase ;)

PR #3264, opened 2020-09-16
Correct prev/next links in the Advanced Guide

heychris:
Previous and next links in the Advanced Guide do not follow the order in the sidebar navigation. This pull request corrects that.

They currently follow the order the documents were in in 2018 when this pull request (#540) was submitted. Subsequent updates to the Advanced Guide have changed the order (it is alphabetical). 

From the comments on that pull request it seems there is some consideration about whether the Advanced Guide even needs prev/next links. FWIW I noticed the issue because I was using them, but I'm fine with submitting a separate pull request to delete them all together if there is a consensus that would be better.

### DIFF
--- a/content/docs/accessibility.md
+++ b/content/docs/accessibility.md
@@ -2,7 +2,6 @@
 id: accessibility
 title: Accessibility
 permalink: docs/accessibility.html
-prev: integrating-with-other-libraries.html
 next: code-splitting.html
 ---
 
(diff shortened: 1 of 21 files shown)


eps1lon:
Given that they weren't intended to have prev/next and already broke I think we might as well remove the links. Since they're alphabetical the amount of insert-sort once has to do when adding new articles is probably not worth it.

heychris:
Works for me, the links are removed in PR #3269

PR #3269, opened 2020-09-17
Remove prev/next links from the Advanced Guide

heychris:
See original discussion on PR #540 and more recently PR #3264. 

Summary is that since the Advanced Guide is alphabetical the chore of resorting new articles in the future may not be worth it.


eps1lon (approved):
Thanks. Ideally they'd be generated from the table of contents (if that one is sorted).

crypt096 (approved):

PR #3280, opened 2020-09-23
Link to explainer how to use new JSX transform with CRA 4

elanandkumar:
Updated the link for CRA release gist which mentions about the eslint configuration change to help others better.


eps1lon:
Thanks!

PR #3281, opened 2020-09-23
Rename state setter from useText to setText

SandroMiguel:
For consistency, I replaced the useState function name `useText` with `setText`.


eps1lon:
Thanks!

Issue #3253, opened 2020-09-11
Clarify that functional components can access state without needed to refactor into ES6 class


skeptek:
I think it might be useful to mention here: https://reactjs.org/docs/state-and-lifecycle.html that you don't have to convert the function to a ES6 class in order to set and access state.  Maybe add a short note with a link to here: https://reactjs.org/docs/hooks-state.html ?  Thanks!

PR #3282, opened 2020-09-24
Added a note linking the reader to effect and state hooks documentation

EC7495:
Made a note that functional components can behave like class components using React hooks. Response to issue #3253.


EC7495:
Signed CLA, waiting for approval.

PR #3283, opened 2020-09-25
Banner scheduling and snooze functionality

rachelnabors:
Expanding on the new banner widget for the site: https://github.com/reactjs/reactjs.org/pull/3274

Now the banner can be configured to run for a set period of time and can be "snoozed" for a period of days after dismissal.

### DIFF
--- a/src/components/Banner/BannerContextManager.js
+++ b/src/components/Banner/BannerContextManager.js
@@ -11,22 +11,22 @@ import BannerContext from './BannerContext';
 
 let activeBanner = null;
 let snooze_start_date = null;
-
 const today = new Date();
+
 function addTimes(date, days) {
   const time = new Date(date);
-  const newTime = time.setDate(time.getDate() + days);
-  return newTime;
+  time.setDate(time.getDate() + days);
+  return time;
 }
 
 // Example usage:
 // activeBanner = {
-//   storageId: 'react_banner_blm',
+//   storageId: 'react_banner',
 //   normalHeight: 60,
 //   smallHeight: 80,
 //   campaign_start_date: '4 Sep 2020',
 //   campaign_end_date: '28 Oct 2020',
-//   snooze_duration: 7,
+//   snooze_for_days: 7,
 //   content: dismiss => (
 //     <div>
 //       <a href="test">Test</a> <button onClick={dismiss}>close</button>
@@ -55,8 +55,7 @@ if (activeBanner) {
       // If we're in the campaign window, but the snooze has been set and it hasn't expired:
     } else if (
       snooze_start_date &&
-      addTimes(snooze_start_date, activeBanner.snooze_duration) >=
-        Date.parse(today.toString())
+      addTimes(snooze_start_date, activeBanner.snooze_for_days) >= today
     ) {
       activeBanner = null;
     }
--- a/src/components/Banner/BannerContextManager.js
+++ b/src/components/Banner/BannerContextManager.js
@@ -26,7 +26,7 @@ function addTimes(date, days) {
 //   smallHeight: 80,
 //   campaign_start_date: '4 Sep 2020',
 //   campaign_end_date: '28 Oct 2020',
-//   snooze_duration: 0,
+//   snooze_duration: 7,
 //   content: dismiss => (
 //     <div>
 //       <a href="test">Test</a> <button onClick={dismiss}>close</button>
@@ -38,7 +38,7 @@ if (activeBanner) {
   try {
     if (localStorage[activeBanner.storageId]) {
       snooze_start_date = new Date(
-        parseInt(localStorage.getItem(activeBanner.storageId)),
+        parseInt(localStorage.getItem(activeBanner.storageId), 10),
       );
     }
   } catch (err) {
@@ -56,7 +56,7 @@ if (activeBanner) {
     } else if (
       snooze_start_date &&
       addTimes(snooze_start_date, activeBanner.snooze_duration) >=
-        Date.parse(today)
+        Date.parse(today.toString())
     ) {
       activeBanner = null;
     }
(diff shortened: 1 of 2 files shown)


src/components/Banner/BannerContextManager.js
```diff
@@ -25,7 +36,28 @@ let activeBanner = null;
...
+      );
+    }
+  } catch (err) {
+    // Ignore.
+  }
+
+  try {
+    // If it's too early or long past the campaign, don't show the banner:
+    if (
+      today < new Date(activeBanner.campaign_start_date) ||
```
gaearon, line 51: Will this work for other locales? If we’re not sure maybe we can use ISO format?

I’m bad at dates so I have no idea

src/components/Banner/BannerContextManager.js
```diff
@@ -10,12 +10,23 @@ import React, {useState, useLayoutEffect} from 'react';
 import BannerContext from './BannerContext';
 
 let activeBanner = null;
+let snooze_start_date = null;
+
+const today = new Date();
+function addTimes(date, days) {
+  const time = new Date(date);
+  const newTime = time.setDate(time.getDate() + days);
```
gaearon, line 18: I think dates are mutable so you don’t need the return value. It’s `time` itself that gets modified by the call.

src/components/Banner/BannerContextManager.js
```diff
@@ -25,7 +36,28 @@ let activeBanner = null;
...
+    if (
+      today < new Date(activeBanner.campaign_start_date) ||
+      today > new Date(activeBanner.campaign_end_date)
+    ) {
+      activeBanner = null;
+      // If we're in the campaign window, but the snooze has been set and it hasn't expired:
+    } else if (
+      snooze_start_date &&
+      addTimes(snooze_start_date, activeBanner.snooze_duration) >=
+        Date.parse(today.toString())
```
gaearon, line 59: What’s the difference between this and just `today`?

gaearon (commented):
Have you tested the full flow? You’d want to clear the storage, see the banner, snooze, then verify it doesn’t show up on refresh. Then edit the place where you call new Date() to pretend we’re in the future and verify it reappears when expected. Verify you can snooze again.

src/components/Banner/BannerContextManager.js
```diff
@@ -10,12 +10,23 @@ import React, {useState, useLayoutEffect} from 'react';
...
 
 // Example usage:
 // activeBanner = {
-//   storageId: 'dismiss_banner_blm',
+//   storageId: 'react_banner_blm',
 //   normalHeight: 60,
 //   smallHeight: 80,
+//   campaign_start_date: '4 Sep 2020',
+//   campaign_end_date: '28 Oct 2020',
+//   snooze_duration: 7,
```
rickhanlonii, line 29:
```suggestion
//   snooze_duration: 7, // days
```

src/components/Banner/BannerContextManager.js
rickhanlonii: Yeah the dates are compared as seconds since epoch UTC so this should work fine.

src/components/Banner/BannerContextManager.js
```diff
@@ -25,7 +36,28 @@ let activeBanner = null;
...
+      today < new Date(activeBanner.campaign_start_date) ||
+      today > new Date(activeBanner.campaign_end_date)
+    ) {
+      activeBanner = null;
+      // If we're in the campaign window, but the snooze has been set and it hasn't expired:
+    } else if (
+      snooze_start_date &&
+      addTimes(snooze_start_date, activeBanner.snooze_duration) >=
+        Date.parse(today.toString())
+    ) {
```
rickhanlonii, line 60: Nit: I think `today` is already a date here:

```suggestion
    } else if (
      snooze_start_date &&
      addTimes(snooze_start_date, activeBanner.snooze_duration) >= today
    ) {
```

rickhanlonii (approved):
Looks great!

src/components/Banner/BannerContextManager.js
rickhanlonii: Or we could rename the setting to `snooze_days`!

rachelnabors: I like it! "Says what it does"

src/components/Banner/BannerContextManager.js
rachelnabors: My dwindling caffeine levels :p

src/components/Banner/BannerContextManager.js
gaearon: I wasn’t referring to the comparison — my question is about `new Date(string)` where the string is hardcoded above in a particular format. My question is whether we’re sure that parsing it works regardless of the locale etc.

For example MDN says (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date):

> Parsing of date strings with the Date constructor (and Date.parse(), which works the same way) is strongly discouraged due to browser differences and inconsistencies.

Can we manually pass all the values instead of relying on parsing here?

rachelnabors:
> Have you tested the full flow? You’d want to clear the storage, see the banner, snooze, then verify it doesn’t show up on refresh. Then edit the place where you call new Date() to pretend we’re in the future and verify it reappears when expected. Verify you can snooze again.

Oh, yes, I have! I'll mention that in future PRs :)

src/components/Banner/BannerContextManager.js
```diff
@@ -10,12 +10,23 @@ import React, {useState, useLayoutEffect} from 'react';
 import BannerContext from './BannerContext';
 
 let activeBanner = null;
+let snooze_start_date = null;
```
lex111, line 13: Why is the snake case used here? Shouldn't we follow camel case naming everywhere?

src/components/Banner/BannerContextManager.js
rachelnabors: How about using the UTC format exclusively: https://stackoverflow.com/a/5619588

rachelnabors: It is done

PR #3289, opened 2020-09-29
2020 Community survey banner

rachelnabors:
Adds a banner for our 2020 community survey. The banner opens and closes and won't be visible until Sept 30th. It will disappear and be ready for code removal after Oct 12

Screenshot 2020-09-29 at 11 16 36
Screenshot 2020-09-29 at 11 16 21


gaearon:
I'm seeing some console warnings.

Screenshot 2020-09-29 at 11 15 04

Screenshot 2020-09-29 at 11 15 08

The second one seems important. It says the DOM structure isn't valid (nested links) which causes unpredictable behavior in React.

src/components/Banner/BannerContextManager.js
```diff
@@ -20,19 +24,82 @@ function addTimes(date, days) {
...
+  snoozeForDays: 7,
+  content: dismiss => (
+    <div
+      css={{
+        display: 'flex',
+        justifyContent: 'center',
+        alignItems: 'flex-start',
+      }}>
+      <a
+        href="https://www.surveymonkey.co.uk/r/673TZ7T"
```
gaearon, line 42: Should this open in new tab by default? Seems like you'd almost never want to go to the docs and then _go away_ to the survey.

gaearon:
Doesn't look good on iPhone. I think we need fewer words.

Screenshot 2020-09-29 at 11 19 22

gaearon (changes requested):
Need to fix nested links (both from the technical standpoint and from the design one — I find it a bit confusing we have a blue link call to action but actually the whole thing except margins is also a link).

Need to fix display on small screens, probably by shortening the message.

src/components/Banner/BannerContextManager.js
```diff
@@ -20,19 +24,82 @@ function addTimes(date, days) {
...
+          transition: 'color 200ms ease-out',
+          ':hover': {
+            color: colors.white,
+          },
+        }}
+        xmlns="http://www.w3.org/2000/svg"
+        viewBox="0 0 5.8 5.8"
+        alt="close"
+        onClick={dismiss}>
+        <path
```
gaearon, line 95: I think the close button is too easy to miss. We need to make it more noticeable.

src/components/Banner/BannerContextManager.js
rachelnabors: Do we have a policy of opening things in new tabs? I tend to err on letting the user right click and open in a new tab if they like. Whereas forcing opening in a new tab gives them no choice. Whatever works for you, though

src/components/Banner/BannerContextManager.js
rachelnabors: I can make it a little larger

src/components/Banner/BannerContextManager.js
gaearon: Yeah we generally have external links on the homepage open in a new tab.

bvaughn:
Follow up PR #3295

gaearon (changes requested):
I need to do some work here to remove the content jump.

gaearon (approved):

espipj:
If you guys could deactivate the banner as the survey has already expired... I can open a PR :smiley:

bvaughn:
Oh, that doesn't seem good. The survey wasn't supposed to expire until December 13th.

gaearon:
No, it _was_ supposed to expire (the original date had a typo that's been fixed).

@espipj Are you saying you are _seeing_ the banner? It isn't supposed to be showing for several weeks by now.

espipj:
@gaearon still there image

Once you dismiss it keeps closed in the following visits at least :)

espipj:
Funny enough it seems that it only shows up on Firefox lol:
https://www.webpagetest.org/result/201112_DiE0_86b97fe63190449a9de8e7e3d29bd07b/
https://www.webpagetest.org/result/201112_Di32_6737936f5e5d1d91e3bca5d521b6b62e/

alexkrolick:
That appears to be because Firefox parses the date strictly and won't let you leave off the time in ISO format.

```
new Date('2020-12-13Z')
> Invalid Date
```

So these two lines are not valid:

https://github.com/reactjs/reactjs.org/pull/3289/files#diff-023eb4be0e5096f5a778ef1dfee421162b529c473ad2c2c7f3e12534eeaa44b7R68-R69

bvaughn:
Nice catch, @alexkrolick.

A PR that fixes this would be welcome.

alexkrolick:
https://github.com/reactjs/reactjs.org/pull/3384/files

PR #3293, opened 2020-09-30
Remove superfluous code in faq-functions.md example

ajbeaven:
The sections above this describe how to use the arrow function to bind functions to the component instance to avoid explicit binding in the constructor. This example though, uses both the arrow function and explicitly binds the function. I realise the point of this particular example isn't to describe the binding process, but considering the sections that immediately proceed it, I think it'd be good to avoid confusion here.


alexkrolick:
I think this is a sensible change given the context

PR #3296, opened 2020-10-01
Updated conditional rendering doc with a note on Falsy values

jeswinsimon:
Conditional rendering works if the expression returns a truthy/falsy value instead of true/false but the falsy value ends up being rendered. This can cause an unexpected value being rendered on screen. 

The discussion regarding this can be found [here](https://github.com/facebook/react-native/issues/29980#issuecomment-696437692)

### DIFF
--- a/content/docs/conditional-rendering.md
+++ b/content/docs/conditional-rendering.md
@@ -152,7 +152,7 @@ It works because in JavaScript, `true && expression` always evaluates to `expres
 
 Therefore, if the condition is `true`, the element right after `&&` will appear in the output. If it is `false`, React will ignore and skip it.
 
-Note that returning a falsy expression will still cause the element after `&&` to be skipped but will return the falsy expression. In the example below, 0 will be returned by the expression.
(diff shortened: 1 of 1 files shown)


content/docs/conditional-rendering.md
```diff
@@ -152,6 +152,19 @@ It works because in JavaScript, `true && expression` always evaluates to `expres
 
 Therefore, if the condition is `true`, the element right after `&&` will appear in the output. If it is `false`, React will ignore and skip it.
 
+Note that returning a falsy expression will still cause the element after `&&` to be skipped but will return the falsy expression. In the example below, 0 will be returned by the expression.
```
eps1lon, line 155: I think we can be a bit more specific what "the expression" is and instead explain directly what is returned from render. `<div>0</div>`

eps1lon (changes requested):
Explaining the drawback of shorthand-conditional rendering is indeed important, thanks! Just a suggestion on how the communicate the implications.

content/docs/conditional-rendering.md
jeswinsimon: Thanks! That makes more sense.

eps1lon (approved):

eps1lon:
@jeswinsimon Thanks!

PR #3299, opened 2020-10-01
Improved Docs

5HAD0W-P1R4T3:

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,20 +7,20 @@ This repo contains the source code and documentation powering [reactjs.org](http
 ### Prerequisites
 
 1. Git
-1. Node: any 12.x version starting with v12.0.0 or greater
-1. Yarn: See [Yarn website for installation instructions](https://yarnpkg.com/lang/en/docs/install/)
-1. A fork of the repo (for any contributions)
-1. A clone of the [reactjs.org repo](https://github.com/reactjs/reactjs.org) on your local machine
+2. Node: any 12.x version starting with v12.0.0 or greater.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -30,30 +30,30 @@ The documentation is divided into several sections with a different tone and pur
...
 ### Make the change
 
-1. Follow the ["Running locally"](#running-locally) instructions
-1. Save the files and check in the browser
-  1. Changes to React components in `src` will hot-reload
-  1. Changes to markdown files in `content` will hot-reload
-  1. If working with plugins, you may need to remove the `.cache` directory and restart the server
+1. Follow the ["Running locally"](#running-locally) instructions.
+2. Save the files and check in the browser.
+  2.1. Changes to React components in `src` will hot-reload.
```
eps1lon, line 41: An unordered list might make more sense here.

eps1lon (changes requested):
Having a single number means you don't have to change subsequent numbers if you insert an item.

The output is still the same: https://github.com/reactjs/reactjs.org/pull/3299/files?short_path=04c6e90#diff-04c6e90faac2675aa89e2176d2eec7d8

(apart from the formatting that is now broken).

5HAD0W-P1R4T3:
An ordered list tells the user it is done is it is in unordered list then the user may be confused to see only 1's and 1's

gaearon:
>An ordered list tells the user it is done if it is in unordered list then the user may be confused to see only 1's and 1's

If you [_read_](https://github.com/reactjs/reactjs.org) the README, you'll see that it's rendered as an ordered list.

PR #3320, opened 2020-10-09
Add missing word to latest blog post

davidblnc:


content/blog/2020-09-22-introducing-the-new-jsx-transform.md
```diff
@@ -54,7 +54,7 @@ function App() {
 
 However, this is not perfect:
 
-* Because JSX compiled into `React.createElement`, `React` needed to be in scope if you use JSX.
+* Because JSX was compiled into `React.createElement`, `React` needed to be in scope if you use JSX.
```
eps1lon, line 57:
```suggestion
* Because JSX was compiled into `React.createElement`, `React` needed to be in scope if you used JSX.
```

English is only my second language but shouldn't this sentence either be in past or present but not in a mixed state?

davidblnc: You're right! Just changed it :)

PR #3322, opened 2020-10-11
Fix ajax hooks example

KNaiskes:
Hello, I have been trying out  [AJAX and APIs](https://reactjs.org/docs/faq-ajax.html) examples and found out that 
```
setItems(result.items);
```
returns undefined while 
```
setItems(result);
```
returns the expected result (Array).


alexkrolick:
Thanks

PR #3329, opened 2020-10-12
fix(content/docs): Fix broken link

eziotedeschi:
Change broken link in documentation for Babel editors.

Closes #3329

### DIFF
--- a/content/docs/introducing-jsx.md
+++ b/content/docs/introducing-jsx.md
@@ -181,4 +181,4 @@ We will explore rendering React elements to the DOM in the [next section](/docs/
 
 >**Tip:**
 >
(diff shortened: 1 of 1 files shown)


eps1lon (commented):
Thanks for spotting it. Do you know what the `next` path segment stands for?

eziotedeschi:
Thanks! It stands for the master version of the documentation on Babel project, I believe we have more chances to have this link working for a long time if we target this specific branch (as we can see today for example).

eps1lon:
> These are the latest changes that have yet to be released.

-- https://babeljs.io/versions

We should revert this change once it's released. Until then a link is better than none.

PR #3339, opened 2020-10-17
Add a note for library authors to JSX blog post

gaearon:
Seems important to note.

### DIFF
--- a/content/blog/2020-09-22-introducing-the-new-jsx-transform.md
+++ b/content/blog/2020-09-22-introducing-the-new-jsx-transform.md
@@ -173,7 +173,7 @@ Starting from Babel 8, `"automatic"` will be the default runtime for both plugin
 >
 > If you use JSX with a library other than React, you can use [the `importSource` option](https://babeljs.io/docs/en/babel-preset-react#importsource) to import from that library instead -- as long as it provides the necessary entry points. Alternatively, you can keep using the classic transform which will continue to be supported.
 >
-> If you're a library author and you are implementing the `/jsx-runtime` entry point for your library, keep in mind that [there is a case](https://github.com/facebook/react/issues/20031#issuecomment-710346866) in which even the new transform has to fall back to `createElement` for backwards compatibility. In that case, it will auto-import `createElement` directly from the root entry point specified by `importSource`.
+> If you're a library author and you are implementing the `/jsx-runtime` entry point for your library, keep in mind that [there is a case](https://github.com/facebook/react/issues/20031#issuecomment-710346866) in which even the new transform has to fall back to `createElement` for backwards compatibility. In that case, it will auto-import `createElement` directly from the *root* entry point specified by `importSource`.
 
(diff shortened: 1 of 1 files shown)


rickhanlonii (approved):

content/blog/2020-09-22-introducing-the-new-jsx-transform.md
```diff
@@ -172,6 +172,8 @@ Starting from Babel 8, `"automatic"` will be the default runtime for both plugin
 > Note
 >
 > If you use JSX with a library other than React, you can use [the `importSource` option](https://babeljs.io/docs/en/babel-preset-react#importsource) to import from that library instead -- as long as it provides the necessary entry points. Alternatively, you can keep using the classic transform which will continue to be supported.
+>
+> If you're a library author and you are implementing the `/jsx-runtime` entry point for your library, keep in mind that [there is a case](https://github.com/facebook/react/issues/20031#issuecomment-710346866) in which even the new transform has to fall back to `createElement` for backwards compatibility. In that case, it will auto-import `createElement` directly from the *root* entry point specified by `importSource`.
```
eps1lon, line 176: @gaearon Quite the long post so I just want to ask if I understood it correctly: `key` after spread falls back to `createElement`? Does it make sense in your opinion to lint against this pattern and change it so that we always use a single runtime?

gaearon: Counter-intuitively (hence the post) it's the other way around. `<div {...obj} key="foo" />` is fine but it temporarily uses the old output (for Reasons™, see the post). `<div key="foo" {...obj} />` will eventually start warning if `obj` has a `key` in it, and we should probably discourage it statically too to avoid confusion. But it uses the `jsx` target.

eps1lon: `<div key="foo" {...obj} />` is what we use everywhere because we never treated `key` as a regular prop or something that could be spread. In typed languages we could even verify statically that both `<div key="foo" {...obj} />` and `<div {...obj} key="foo" />` are fine. 

I'm concerned about bundle size impact and would like to make sure I ship a single runtime usage even if  `createElement` will never be able to be stripped from the bundle.

Either way I'm not quite sure what I'm supposed to do with that notice. Can't use `<div {...obj} key="foo" />` because it uses the old runtime. But `<div key="foo" {...obj} />` is also not recommended. 

What negative impact would there be if we switch to `<div {...obj} key="foo" />` and force the new runtime with the assumption that we never spread `key`.

gaearon: Can we continue the discussion on the issue? I don't think this is a very good place for it.

eps1lon: Sure. Just wanted to check if I understood it correctly but didn't expect the result. Let me get an overview about the usage in our code base first and test some assumptions.

PR #3342, opened 2020-10-20
docs: fixing anti-pattern in one of the hooks state examples

doojin:
From `React.Component` `setState` function [documentation](https://reactjs.org/docs/react-component.html#setstate):

> Subsequent calls will override values from previous calls in the same cycle, so the quantity will only be incremented once. If the next state depends on the current state, we recommend using the updater function form.

In one of the Hooks [examples](https://reactjs.org/docs/hooks-state.html#equivalent-class-example) state is updated based on the `this.state` value which potentially may lead to incorrect result.

### DIFF
--- a/content/docs/hooks-state.md
+++ b/content/docs/hooks-state.md
@@ -47,7 +47,7 @@ class Example extends React.Component {
     return (
       <div>
         <p>You clicked {this.state.count} times</p>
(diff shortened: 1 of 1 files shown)


doojin:
@facebook-github-bot check

PR #3346, opened 2020-10-21
Correct `setState` signature typo in doc

prasook-jain:
Instead of `setState(updater, [callback])` it should be `setState(updater[, callback])` for optional callback function.

### DIFF
--- a/content/docs/reference-react-component.md
+++ b/content/docs/reference-react-component.md
@@ -501,7 +501,7 @@ There are just two of them: `setState()` and `forceUpdate()`.
 ### `setState()` {#setstate}
 
 ```javascript
-setState(updater, [callback])
+setState(updater[, callback])
 ```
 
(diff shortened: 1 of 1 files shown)


prasook-jain:
I am not here to participate in `Hacktoberfest` fest. I got confuse for a little while reading the doc, that callback function can be passed as arrays in second argument. I genuine feel that this need to be corrected.

eps1lon (commented):
I guess both variants are confusing since they're not actual JS syntax. The hooks API doesn't even mark optional parameters e.g. https://reactjs.org/docs/hooks-reference.html#usereducer

What do you think about just omitting the square brackets `[]`?

[MDN](https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle) and [php.net](https://www.php.net/manual/en/function.str-split) add an extra space between argument name and square bracket: `(arg1 [, arg2])`.

prasook-jain:
Thanks for the reply, I think this is stale now since the introduction of new doc and new react version, hence closing the PR.

PR #3349, opened 2020-10-23
updating Glitch example link

jennschiffer:
this simply updates the link to the Glitch remix of a CRA starter - this one is up to date and will be kept up to date!

### DIFF
--- a/content/docs/getting-started.md
+++ b/content/docs/getting-started.md
@@ -36,7 +36,7 @@ React has been designed from the start for gradual adoption, and **you can use a
 
 ### Online Playgrounds {#online-playgrounds}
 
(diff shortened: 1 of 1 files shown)


gaearon:
Oh wow I was just about to remove because it didn't load for me (https://github.com/reactjs/reactjs.org/pull/3350)

I'm still a bit concerned about this one.

Why does it take so long to start? I have to "show -> in sidebar" and then a minute later it says "Ready..." and nothing happens

Screenshot 2020-10-23 at 22 05 15

It took a few minutes to show up, compared to a few seconds in all other linked starters. It seems like it's running a real VM but that seems unnecessary given CRA's whole point is client-side apps. Is there a purely client-side template? 

Also, I was seeing this several times:

Screenshot 2020-10-23 at 22 04 32

I couldn't figure out what they mean or even how to see the errors. When I clicked on them they expanded into some menus and I couldn't find the actual errors and warnings.

crypt096 (approved):

jennschiffer:
Totally fair! I'll close this PR so you can act on the newer one you created removing the link. I'll make a new one once we get our act together! Thanks for the feedback and perspective, Dan.

PR #3350, opened 2020-10-23
Remove broken Glitch template

gaearon:
This is

1) using real CRA unnecessarily despite being in the browser so is slow to boot
2) shows "error" immediately to me and doesn't load
3) doesn't seem to get updated with newer React releases

We can readd it if it's fixed and doesn't have these problems (which none of the other templates do).


crypt096 (approved):

gaearon:
Gonna remove for now but @jennschiffer said something better might be coming in the future.

PR #3352, opened 2020-10-26
Fix a grammatical error in Context.md

sambigelow:
The sentence didn't quite make sense before. I think this makes the intent more clear.

(I realize this is just a markdown change, but this has nothing to do with Hacktoberfest and I feel like it is a [minimally] valuable contribution. Thanks for all that this team does!)


eps1lon (approved):
Thanks! I don't think it's an error. Your change is just more explicit what accepts the `value` which is always better in technical documentation.

sambigelow:
@eps1lon I think it was an error. It was missing a subject. Maybe it could have said "It accepts the `value` prop [...]" but it just said "Accepts [...]"

PR #3353, opened 2020-10-26
specify required option to use Flow with new JSX transform

FezVrasta:
Unfortunately the option is not documented on the Flow docs, and it's not very obvious how to get Flow to work with the new JSX transform.


gaearon:
Oh nice, thanks. Want to send a PR to Flow docs too?

FezVrasta:
https://github.com/facebook/flow/pull/8514/files

PR #3364, opened 2020-11-02
Update faq-ajax.md

toxicinx:
Key property of array should be item.id, not item.name

### DIFF
--- a/content/docs/faq-ajax.md
+++ b/content/docs/faq-ajax.md
@@ -72,7 +72,7 @@ class MyComponent extends React.Component {
       return (
         <ul>
           {items.map(item => (
-            <li key={item.name}>
+            <li key={item.id}>
               {item.name} {item.price}
             </li>
           ))}
@@ -120,7 +120,7 @@ function MyComponent() {
     return (
       <ul>
         {items.map(item => (
-          <li key={item.name}>
+          <li key={item.id}>
             {item.name} {item.price}
           </li>
         ))}


crypt096 (approved):

joekyo:
I found another mistake in this doc in the Hook example.
```
     .then(
        (result) => {
          setIsLoaded(true);
          setItems(result);
        },
```

It should be `setItems(result.items);`.

sijusamson:
> I found another mistake in this doc in the Hook example.
> 
> ```
>      .then(
>         (result) => {
>           setIsLoaded(true);
>           setItems(result);
>         },
> ```
> 
> It should be `setItems(result.items);`.

Don't you think this should be a new PR. ?

sijusamson:
> I found another mistake in this doc in the Hook example.
> 
> ```
>      .then(
>         (result) => {
>           setIsLoaded(true);
>           setItems(result);
>         },
> ```
> 
> It should be `setItems(result.items);`.

Well, I wonder why it should be **`setItems(result.items)`** ?

joekyo:
> > I found another mistake in this doc in the Hook example.
> > ```
> >      .then(
> >         (result) => {
> >           setIsLoaded(true);
> >           setItems(result);
> >         },
> > ```
> > 
> > 
> > It should be `setItems(result.items);`.
> 
> Well, I wonder why it should be **`setItems(result.items)`** ?

The `result` is an object and the `items` is an array, right?

Issue #3120, opened 2020-07-19
The prop value with an expression type of SequenceExpression could not be resolved. Please file issue to get this fixed immediately.


inblack67:
I was just contributing to the docs a few minutes ago and after running
```sh
yarn check
```
got this warning
 #### "The prop value with an expression type of SequenceExpression could not be resolved. Please file issue to get this fixed immediately." . 
Thought I'd let you know.

rickhanlonii:
Do you mean `yarn lint`?

I get this as well, seems to be an eslint issue. Do you want to upgrade our eslint version to see if it fixes it?

<img width="945" alt="Screen Shot 2020-07-19 at 3 13 39 PM" src="https://user-images.githubusercontent.com/2440089/87883037-80d55080-c9d2-11ea-8abf-ed34a2ad81e5.png">

inblack67:
I meant
```sh
yarn check-all
```
I tried upgrading eslint. No help!

nawazeverlane:
https://github.com/jsx-eslint/jsx-ast-utils/issues/95 could be the issue blocking.

nawazkhan:
https://github.com/jsx-eslint/jsx-ast-utils/issues/95 could be the issue blocking.

rrhg:
[jsx-eslint/jsx-ast-utils#101](https://github.com/jsx-eslint/jsx-ast-utils/pull/101) was merged & should fix this.

rickhanlonii:
@rrhg do you want to submit the PR to upgrade?

rrhg:
I can try. If someone could provide some guidance that would be great. Thanks.

rrhg:
Did some research & if I understand correctly, the 2 dependencies using it are eslint-plugin-react & eslint-plugin-jsx-a11y.

eslint-plugin-react has a newer release tag ([v7.21.4](https://github.com/yannickcr/eslint-plugin-react/tree/v7.21.4)) that updated jsx-ast-utils to 3.0.0 & should fix this.

But for eslint-plugin-jsx-a11y, we have to wait for an active branch ([v6.4.0](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/tree/tag-6.4.0)) to be merged. Apparently it should be ready soon.

rickhanlonii:
Thanks for looking into it @rrhg!

webrafaelso:
Please, fix this asap. GitLab ci/cd keep showing this error message when checking the project

[...]

PR #3370, opened 2020-11-04
Fix for #3120. Updated eslint-plugin-react to 7.21.5 & eslint-plugin-jsx-a11y to 6.4.1

rrhg:
These updates fix issue #3120. The real problem was in jsx-ast-utils
which is a dependency of both eslint-plugin-react &
eslint-plugin-jsx-a11y


crypt096 (approved):

rickhanlonii (approved):

rickhanlonii:
Thanks!

PR #3371, opened 2020-11-04
Fix: Collapsing Sidebar Sections

1209simran:
When we click on the sidebar sections of Community, Blog or Tutorials Page, the button doesn't collapse.


src/templates/components/Sidebar/Sidebar.js
```diff
@@ -52,7 +52,7 @@ class Sidebar extends Component {
         {sectionList.map((section, index) => (
           <SectionComponent
             createLink={createLink}
-            isActive={activeSection === section || sectionList.length === 1}
+            isActive={activeSection === section || sectionList.length === 0}
```
gaearon, line 55: This doesn't look right. If `sectionList.length === 0` then we wouldn't ever get here since we're inside `sectionList.map(...)`. I think this was actually intentional (what's the point of collapsing the only section?) although I agree this design is a bit confusing. It would make sense to me to just remove the arrow if the length is 1.

1209simran: Yes @gaearon, this design looks confusing to me for the single section that's why I replace the length. But I'll remove the arrow instead of replacing the length to 0.

1209simran: @gaearon, I've removed the arrow instead of changing the length. Please check the updated PR.

crypt096 (approved):

PR #3373, opened 2020-11-04
Fix for #3361: Redirected to the same page instead of Home Page

### LINKED ISSUE #3361: Can't stay in the same page after choosing the language
I've noticed that when you chose a language in the Reactjs doc, you're redirected to the home page instead of the place you were.

I think it'd be better to just stay in the same page.

1209simran:
Fixes #3361


BartoszKlonowski:
Closing as per comments in the related issues.

PR #3382, opened 2020-11-11
More explicit docs about uncontrolled forms.

VoytechG:
It is worth to point out `defaultValue` attribute behaviour explicitly. In a use case when a form is reused across different sub-routes of an app, where said routes are changed by a user, input fields to change their value on user's action. With uncontrolled forms, `defaultValue` change may be expected to trigger DOM changes. The docs only mention the behaviour on a component's mount, but don't describe what happens when the attribute changes.

Related:
https://github.com/reactjs/reactjs.org/issues/1126

### DIFF
--- a/content/docs/uncontrolled-components.md
+++ b/content/docs/uncontrolled-components.md
@@ -45,7 +45,7 @@ If it's still not clear which type of component you should use for a particular 
 
 ### Default Values {#default-values}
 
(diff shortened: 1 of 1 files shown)


crypt096 (approved):

eps1lon (approved):
It's somewhat implied by "uncontrolled" but doesn't hurt to emphasize how this translates to user-observable behavior.

PR #3384, opened 2020-11-12
fix date format for firefox

alexkrolick:


bvaughn (approved):

bvaughn:
Thanks for the fix!

We should probably update the banner parsing code to warn about invalid dates too so this doesn't sneak back in with a future survey :D

PR #3399, opened 2020-11-22

CyberMew:
Add links to the proper page for more information.


eps1lon (changes requested):
What you return from render is not the same as what you pass as `children`  to built-in components e.g. `<div />`.

So `<div>{undefined}</div>` is ok but `<Component>{undefined}</Component>` is not if that `Component` returns `props.children`.

Does that make sense?

PR #3404, opened 2020-11-24
Force Algolia search to redirect to window.origin URL

renatodex:
## Why did you do it?

While publishing a PR, I've noticed that search doesn't work on Vercel
Apps. This happens because Algolia results always default to production
URLs. This is not the expected behavior while testing, since you can
accidentally left the Vercel App without noticing, introducing all
kinds of confusions.

## What changed?

With this small change, we are always enforcing the origin URL to be
used as the host for the redirect.
This will work both for production an any other environment.

## Reproduction of the Issue:

https://share.getcloudapp.com/RBu9NyZb

### DIFF
--- a/src/components/LayoutHeader/DocSearch.js
+++ b/src/components/LayoutHeader/DocSearch.js
@@ -25,6 +25,10 @@ class DocSearch extends Component<{}, State> {
         apiKey: '36221914cce388c46d0420343e0bb32e',
         indexName: 'react',
         inputSelector: '#algolia-doc-search',
+        handleSelected: function(input, event, suggestion, datasetNumber, context) {
+          const suggestionUrl = new URL(suggestion.url)
+          window.location = `${window.origin}${suggestionUrl.pathname}`
+        },
       });
     } else {
       console.warn('Search has failed to load and now is being disabled');


src/components/LayoutHeader/DocSearch.js
```diff
@@ -25,6 +25,10 @@ class DocSearch extends Component<{}, State> {
         apiKey: '36221914cce388c46d0420343e0bb32e',
         indexName: 'react',
         inputSelector: '#algolia-doc-search',
+        handleSelected: function(input, event, suggestion, datasetNumber, context) {
+          const suggestionUrl = new URL(suggestion.url)
+          window.location = `${window.origin}${suggestionUrl.pathname}`
```
renatodex, line 30: All paths are coming with a slash from Algolia, that's why I haven't added slash between `window.origin` and `suggestionUrl.pathname`

More info about `handleSelected` behavior:
https://docsearch.algolia.com/docs/behavior/

plugins/gatsby-transformer-home-example-code/gatsby-node.js
```diff
@@ -3,7 +3,6 @@
  */
 
 const crypto = require('crypto');
-const path = require('path');
```
renatodex, line 6: This was just an unused var, reported on the lint test, so I've decided to remove it.
This was the error:
```6:7  warning  'path' is assigned a value but never used. Allowed unused vars must match /^_/  no-unused-vars```

**Comment by alexkrolick** [comment]
This is great, this has been a problem for a long time!

File: src/components/LayoutHeader/DocSearch.js [hunk 3]
```diff
@@ -25,6 +25,16 @@ class DocSearch extends Component<{}, State> {
...
         indexName: 'react',
         inputSelector: '#algolia-doc-search',
+        handleSelected: function(
+          input,
+          event,
+          suggestion,
+          datasetNumber,
+          context,
+        ) {
+          const suggestionUrl = new URL(suggestion.url);
```
alexkrolick, line 35: Does this get a corejs polyfill? Doesn't work under IE11. Not sure if that is supported still on this site or not.

renatodex: Thats a good point!
Can we find a confirmation of supported browsers? I would love to use a Polyfill, but don’t want to do that id IE11 compatibility is not required/expected.

PR #3409, opened 2020-11-26
Change tabIndex to match the types (string to number)

Ed-Ed:
Fixes https://github.com/reactjs/reactjs.org/issues/3408


lex111 (approved):

lex111:
@Ed-Ed this makes sense especially when using TypeScript, so thanks for the fix!

Cloud7050:
With these changes, part of [this introductory page](https://reactjs.org/docs/introducing-jsx.html) now does not make sense and left me confused till I figured out how to look into the history of the page to see the older content.

The affected part says "You may use quotes to specify string literals as attributes", yet the `tabIndex` in its example has been changed from quotes to use curly braces, which demonstrates the same curly brace concept as the part right below it ("You may also use curly braces ...").

PR #3415, opened 2020-11-30
Update error-boundaries.md

0000marcell:
I spend a lot of time trying to figure out why my code wasn't working, I only understood the problem after finding a stack overflow question with the same problem, https://bit.ly/2VhDpC9, the documentation should mention the behavior in development because most people will test in development first

### DIFF
--- a/content/docs/error-boundaries.md
+++ b/content/docs/error-boundaries.md
@@ -21,6 +21,7 @@ Error boundaries are React components that **catch JavaScript errors anywhere in
 > * Asynchronous code (e.g. `setTimeout` or `requestAnimationFrame` callbacks)
 > * Server side rendering
 > * Errors thrown in the error boundary itself (rather than its children)
+> * Error Bounderies only work in production mode, in development the app will be ummounted after showing the fallback ui
 
 A class component becomes an error boundary if it defines either (or both) of the lifecycle methods [`static getDerivedStateFromError()`](/docs/react-component.html#static-getderivedstatefromerror) or [`componentDidCatch()`](/docs/react-component.html#componentdidcatch). Use `static getDerivedStateFromError()` to render a fallback UI after an error has been thrown. Use `componentDidCatch()` to log error information.
 


content/docs/error-boundaries.md
```diff
@@ -21,6 +21,7 @@ Error boundaries are React components that **catch JavaScript errors anywhere in
 > * Asynchronous code (e.g. `setTimeout` or `requestAnimationFrame` callbacks)
 > * Server side rendering
 > * Errors thrown in the error boundary itself (rather than its children)
+> * Error Bounderies only work in production mode, in development the app will be ummounted after showing the fallback ui
```
eps1lon, line 24: The unmounting behavior stays the same. Maybe you're confusing the error overlay from create-react-app (which displays regardless of error boundaries) with the app being unmounted?

In https://codesandbox.io/s/error-boundary-vs-error-overlay-je86k?file=/src/App.js the app is still mounted. You just need to close the shown error overlay.

eps1lon (changes requested):

content/docs/error-boundaries.md
0000marcell: Hello @eps1lon, You're right, it's the error overlay, All these years I didn't know it was an overlay :D, but anyway I still think its a little confusing I'm sure a lot of people don't know that the thing is an overlay too, people in the stack overflow question that I've mentioned didn't know it was a overlay, I think we should at least mention the overlay then, what do you think ?

eps1lon: Definitely agree. This comes up every so often. https://github.com/facebook/create-react-app/issues/6530 tracks it already.

0000marcell: But the solution proposed is changing the way it behaves, I just want to include some information about it in the docs so other people don't have to suffer for now :smile: , I've created another PR, I'm going to change my PR now that I understand the problem better, thanks @eps1lon :smile:

PR #3416, opened 2020-11-30
Update error-boundaries.md

0000marcell:
I've spend a lot of time trying to figure out why my code wasn't working, I only understood the problem after finding a stack overflow question with the same problem, https://bit.ly/2VhDpC9, the documentation should mention the behavior in development because most people will test in development first

### DIFF
--- a/content/docs/error-boundaries.md
+++ b/content/docs/error-boundaries.md
@@ -21,6 +21,7 @@ Error boundaries are React components that **catch JavaScript errors anywhere in
 > * Asynchronous code (e.g. `setTimeout` or `requestAnimationFrame` callbacks)
 > * Server side rendering
 > * Errors thrown in the error boundary itself (rather than its children)
+> * Error Bounderies only work in production mode, in development the app will be ummounted after showing the fallback ui
 
 A class component becomes an error boundary if it defines either (or both) of the lifecycle methods [`static getDerivedStateFromError()`](/docs/react-component.html#static-getderivedstatefromerror) or [`componentDidCatch()`](/docs/react-component.html#componentdidcatch). Use `static getDerivedStateFromError()` to render a fallback UI after an error has been thrown. Use `componentDidCatch()` to log error information.
 


content/docs/error-boundaries.md
```diff
@@ -22,6 +22,8 @@ Error boundaries are React components that **catch JavaScript errors anywhere in
 > * Server side rendering
 > * Errors thrown in the error boundary itself (rather than its children)
 
+> In development an error overlay will be shown on top of the screen reguardless if you're using error boundaries or not, you can close the overlay by pressing esc
```
eps1lon, line 25: That depends on the framework you're using (e.g nextjs or create-react-app). It is not included in React itself. So we should probably include that an error overlay might be shown depending on the framework you're using. 

I would leave out instructions on how to close it since that depends on the specific implementation of the error overlay.

0000marcell: So something like this "In development an error overlay might be shown on top of the screen regardless if you're using error boundaries or not, usually you can close this overlay"

eps1lon: Yeah, and I would add that this depends on the framework you're using and give some examples (create-react-app, nextjs). 

Usually I would be careful with documenting 3rd party behavior but at this point it's somewhat convention that your React framework should implement some error overlay.

0000marcell: I've changed to "In development, an error overlay might be shown on top of the screen regardless if you're using error boundaries or not, create-react-app and nextjs implement this error overlay"

content/docs/error-boundaries.md
```diff
@@ -22,6 +22,8 @@ Error boundaries are React components that **catch JavaScript errors anywhere in
 > * Server side rendering
 > * Errors thrown in the error boundary itself (rather than its children)
 
+> In development an error overlay might be shown on top of the screen regardless if you're using error boundaries or not, create-react-app and nextjs implement this error overlay
```
eps1lon, line 25:
```suggestion
> In development an error overlay might be shown on top of the screen regardless if you're using error boundaries or not. For example, create-react-app and nextjs implement this error overlay.
```

eps1lon (approved):
Thanks!

PR #3424, opened 2020-12-01
Update introducing-jsx.md

joeldmyers:
Maybe I am wrong, but it seems that you meant to have quotes here instead of curly braces.


eps1lon (commented):
#3409 changed all of these.

I think we should just use a different example. A `<div>` with a `tabIndex` is a bad example anyway since you should use it sparringly.

How about we use `<a href="/"></a>` instead?

harish-sethuraman:
We actually updated it to string and now we have updated the example to. anchor tag since tabIndex example was pretty misleading. Thanks for sending PR.

PR #3426, opened 2020-12-03
Update strict-mode.md

infinnie:
Add documentation that explicitly states about modification to console behavior.

### DIFF
--- a/content/docs/strict-mode.md
+++ b/content/docs/strict-mode.md
@@ -118,7 +118,7 @@ By intentionally double-invoking methods like the component constructor, strict 
 
 > Note:
 >
-> Currently, React automatically [modifies](https://github.com/facebook/react/pull/18547) the console methods like `console.log()` to silence the logs in the second call to lifecycle functions. However, it may cause undesired behavior in certain cases, where [a workaround can be used](https://github.com/facebook/react/issues/20090#issuecomment-715927125).
+> Currently, React automatically modifies the console methods like `console.log()` to silence the logs in the second call to lifecycle functions. However, it may cause undesired behavior in certain cases where [a workaround can be used](https://github.com/facebook/react/issues/20090#issuecomment-715927125).
 
 ### Detecting legacy context API {#detecting-legacy-context-api}
 
--- a/content/docs/strict-mode.md
+++ b/content/docs/strict-mode.md
@@ -118,7 +118,7 @@ By intentionally double-invoking methods like the component constructor, strict 
 
 > Note:
 >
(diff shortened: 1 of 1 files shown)


content/docs/strict-mode.md
```diff
@@ -116,6 +116,10 @@ At first glance, this code might not seem problematic. But if `SharedApplication
 
 By intentionally double-invoking methods like the component constructor, strict mode makes patterns like this easier to spot.
 
+> Note:
+>
+> Currently, React automatically [modifies](https://github.com/facebook/react/pull/18547) the console methods like `console.log()` to silence the logs in the second call to lifecycle functions. However, it may cause undesired behavior in certain cases, where [a workaround can be used](https://github.com/facebook/react/issues/20090#issuecomment-715927125).
```
eps1lon, line 121:
```suggestion
> Currently, React automatically modifies the console methods like `console.log()` to silence the logs in the second call to lifecycle functions. However, it may cause undesired behavior in certain cases, where [a workaround can be used](https://github.com/facebook/react/issues/20090#issuecomment-715927125).
```

The implementation isn't important for documentation as far as I can tell. It may even be misleading in case code moves around.

```suggestion
> Currently, React automatically modifies the console methods like `console.log()` to silence the logs in the second call to lifecycle functions. However, it may cause undesired behavior in certain cases where [a workaround can be used](https://github.com/facebook/react/issues/20090#issuecomment-715927125).
```
Not sure about that comma. Seems off to me but if you're sure then keep it.

eps1lon (changes requested):
Much appreciated. Some editorial questions though!

content/docs/strict-mode.md
```diff
@@ -116,6 +116,10 @@ At first glance, this code might not seem problematic. But if `SharedApplication
 
 By intentionally double-invoking methods like the component constructor, strict mode makes patterns like this easier to spot.
 
+> Note:
+>
+> Currently, React automatically modifies the console methods like `console.log()` to silence the logs in the second call to lifecycle functions. However, it may cause undesired behavior in certain cases where [a workaround can be used](https://github.com/facebook/react/issues/20090#issuecomment-715927125).
```
eps1lon, line 121:
```suggestion
> Starting with React 17, React automatically modifies the console methods like `console.log()` to silence the logs in the second call to lifecycle functions. However, it may cause undesired behavior in certain cases where [a workaround can be used](https://github.com/facebook/react/issues/20090#issuecomment-715927125).
```

eps1lon (approved):
Thank you for sticking with it 👍

PR #3436, opened 2020-12-07
Add temporary note to various doc pages pointing to docs with functional components

gjunkie:
Addresses https://github.com/reactjs/reactjs.org/issues/3435

Adds a temporary link to the [netlify docs](https://reactwithhooks.netlify.app/docs/getting-started.html) where all tutorials and examples have been ported to use functional components with hooks.

This can be removed once the docs have been updated.

### DIFF
--- a/content/docs/components-and-props.md
+++ b/content/docs/components-and-props.md
@@ -20,7 +20,7 @@ Components let you split the UI into independent, reusable pieces, and think abo
 
 Conceptually, components are like JavaScript functions. They accept arbitrary inputs (called "props") and return React elements describing what should appear on the screen.
 
-**Note**: We're currently working on a rewrite of the React docs to teach functional components with hooks as the default approach. In the meantime, all of the tutorials found in these docs can be found at https://reactwithhooks.netlify.app/ using functional components and hooks.
+>**Note**: We're currently working on a rewrite of the React docs to teach functional components with hooks as the default approach. In the meantime, all of the tutorials found in these docs can be found at https://reactwithhooks.netlify.app/ using functional components and hooks.
 
 ## Function and Class Components {#function-and-class-components}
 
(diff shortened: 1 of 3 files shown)


content/docs/components-and-props.md
```diff
@@ -20,6 +20,8 @@ Components let you split the UI into independent, reusable pieces, and think abo
 
 Conceptually, components are like JavaScript functions. They accept arbitrary inputs (called "props") and return React elements describing what should appear on the screen.
 
+**Note**: We're currently working on a rewrite of the React docs to teach functional components with hooks as the default approach. In the meantime, all of the tutorials found in these docs can be found at https://reactwithhooks.netlify.app/ using functional components and hooks.
```
markerikson, line 23: Suggest formatting these as blockquote notes

markerikson (commented):
Can we also fit this onto the front page somewhere?  Perhaps right above the "A Simple Component" example?  That way it's visible, but not annoyingly in the middle of that page.

gjunkie:
@markerikson I'm not sure it's necessary that it go on the front page, to be honest. I'm assuming most people will be going to the docs directly, where these notices will be visible. I'm also not sure the team would want to link out directly from the front page, nor how they'd want to style this. Tho if they want to chime in here, I'm certainly willing to add it.

content/docs/components-and-props.md
gjunkie: 👍 done!

markerikson:
Fair enough :)

Looks like the blockquotes aren't kicking in - I think you need to add a space after the `>` characters?

gjunkie:
@markerikson That's odd. I'm following the same convention as the tips on those same pages. I did just push another update to follow the same formatting as the Tips.

markerikson:
Formatting looks correct now.

So, the question is whether @rachelnabors and @gaearon would like to accept this:

https://twitter.com/rachelnabors/status/1336007620675645440

rachelnabors:
Thanks so much, but we have some plans in the works for early 2021! Further explanation: https://github.com/reactjs/reactjs.org/issues/3435#issuecomment-740104729

PR #3447, opened 2020-12-11
Update rendering-elements.md

Whistler092:
Add a link to the next section

### DIFF
--- a/content/docs/rendering-elements.md
+++ b/content/docs/rendering-elements.md
@@ -73,3 +73,6 @@ You can verify by inspecting the [last example](codepen://rendering-elements/upd
 Even though we create an element describing the whole UI tree on every tick, only the text node whose contents have changed gets updated by React DOM.
 
 In our experience, thinking about how the UI should look at any given moment, rather than how to change it over time, eliminates a whole class of bugs.
+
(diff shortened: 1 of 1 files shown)


eps1lon:
The pages in "Main concepts" already have links to the next and previous section. Though these links are below the rating so maybe that's why you missed them?

Whistler092:
@eps1lon Yes ... it makes sense, but when you read all the chapters, you always see a small paragraph of what you will see later and the link, I did not see it in this and I thought it would be a good idea to add it. But you are right, this paragraph is unnecessary and there is no problem in removing my PR.

Thanks for taking the time to review it.

eps1lon:
> it makes sense, but when you read all the chapters, you always see a small paragraph of what you will see later and the link

I wasn't aware of that, thanks :+1: Now we're "stopping" at https://deploy-preview-3447--reactjs.netlify.app/docs/state-and-lifecycle.html.

Though reading through the paragraph you added: It's the first mention of props. Maybe I'm misunderstanding what "consolidated" means but I would expect that we learned something about props before we can consolidate that knowledge, no? The page even explicitly mentions that "We will introduce components in the next section"

PR #3460, opened 2020-12-23
Revert "Use vanity URL for Reactiflux"

nickmccurdy:
Reverts reactjs/reactjs.org#2372

We're allowing direct links to the #react-native channel again (though #help-react-native is a more popular channel, it's not in the React Native category)


nickmccurdy:
Sorry wrong repository, I thought these were the React Native docs

PR #3461, opened 2020-12-23
Put ThemedButton inside context provider

luckluster:
Fix a wrong example in the Context documentation: https://reactjs.org/docs/context.html

The ThemedButton used the context, so it must be inside the context provider! 
The whole page teaches us that the components that use the context has to be inside the context provider - but this was is not the case here - and as a result the button did not catch the context changes. So I placed the button INSIDE the context provider.

### DIFF
--- a/examples/context/theme-detailed-app.js
+++ b/examples/context/theme-detailed-app.js
@@ -37,10 +37,10 @@ class App extends React.Component {
       <Page>
         <ThemeContext.Provider value={this.state.theme}>
           <Toolbar changeTheme={this.toggleTheme} />
+          <Section>
+            <ThemedButton />
+          </Section>      
         </ThemeContext.Provider>
-        <Section>
-          <ThemedButton />
-        </Section>
       </Page>
     );
   }


luckluster:
Note that the build fails due to some configuration problem with **prettier**

image

eps1lon (changes requested):
One button is purposely outside of the context which has a default value:

> //highlight-range{1-3}
> // The ThemedButton button inside the ThemeProvider
> // uses the theme from state while the one outside uses
> // the default dark theme

-- https://github.com/luckluster/reactjs.org/blob/a95cfae6076d661de6c7b35c1656402c83dd1f9b/examples/context/theme-detailed-app.js#L31-L35

Does that comment clear things up for you?

luckluster:
> 
> 
> One button is purposely outside of the context which has a default value:
> 
> > //highlight-range{1-3}
> > // The ThemedButton button inside the ThemeProvider
> > // uses the theme from state while the one outside uses
> > // the default dark theme
> 
> -- https://github.com/luckluster/reactjs.org/blob/a95cfae6076d661de6c7b35c1656402c83dd1f9b/examples/context/theme-detailed-app.js#L31-L35
> 
> Does that comment clear things up for you?

Yes.. I shouldn't have missed that. Thank you.

PR #3475, opened 2021-01-04
remove componentWillMount/WillReceiveProps

amen-souissi:
componentWillMount and componentWillReceiveProps are no longer supported by React

### DIFF
--- a/content/docs/reconciliation.md
+++ b/content/docs/reconciliation.md
@@ -27,7 +27,7 @@ When diffing two trees, React first compares the two root elements. The behavior
 
 Whenever the root elements have different types, React will tear down the old tree and build the new tree from scratch. Going from `<a>` to `<img>`, or from `<Article>` to `<Comment>`, or from `<Button>` to `<div>` - any of those will lead to a full rebuild.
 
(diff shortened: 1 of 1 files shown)


eps1lon (changes requested):
That's a good point.

They're still supported though with an `UNSAFE_` prefix. I think they should be mentioned so that people don't wonder when these methods are called.

Maybe like https://reactjs.org/docs/react-component.html#mounting does?

amen-souissi:
Maybe something like this ?


When a component updates, the instance stays the same, so that state is maintained across renders. React updates the props of the underlying component instance to match the new element, and calls `UNSAFE_componentWillReceiveProps()`, `UNSAFE_componentWillUpdate()` and `componentDidUpdate()` on the underlying instance.

Next, the `render()` method is called and the diff algorithm recurses on the previous result and the new result.

>Note:
>
>These methods are considered legacy and you should [avoid them](/blog/2018/03/27/update-on-async-rendering.html) in new code:
>
>- `UNSAFE_componentWillUpdate()`
>- `UNSAFE_componentWillReceiveProps()`

eps1lon:
Seems good to me :+1:

eps1lon (approved):

eps1lon:
Thanks!

PR #3480, opened 2021-01-07
Remove reactarmory from the courses page

sontek:
Their SSL cert has been expired for 8 months.  I don't believe its a great idea to send new users to them at this point.


gaearon:
Seems like it would be nice to contact the author first? I'll reach out to James.

sontek:
Looks like at some point a different PR was merged to fix this

PR #3490, opened 2021-01-16
added hooks example to portal page

qpwo:


eps1lon:
Thanks for the suggestions.

The docs are being [rewritten with hooks in mind at the moment](https://github.com/reactjs/reactjs.org/issues/3308) so this will likely be dropped anyway.

PR #3493, opened 2021-01-19
fix: Remove duplicate conference links

iamdarshshah:
#### Changes:

- Removed duplicated social links for `React India 2021`
- Added missing bracket `)` for `React Native EU 2020`'s Instagram social link
- `ReactNL 2016` - added links in a proper format
- Removed iframe embed for `React.js Conf 2015` (Not needed, as we've already provided the YouTube link for the conf. and this embed video is already included [here](https://reactjs.org/community/videos.html#reactjs-conf-2015)). Otherwise, we should add an iframe after each conference, just to make it look symmetrical.

> Let me know if there is anything that needs to change.

### DIFF
--- a/content/community/conferences.md
+++ b/content/community/conferences.md
@@ -30,7 +30,7 @@ September 13-15, 2021. Atlanta, GA, USA
 ### React India 2021 {#react-india-2021}
 November 12-13, 2021 in Mumbai, India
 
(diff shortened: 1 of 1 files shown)


eps1lon:
@iamdarshshah Thanks!

PR #3495, opened 2021-01-23
Update React/Frontend armory course link

jamesknelson:
React Armory hasn't been updated for a while, and the SSL cert is expired (see #3480). This updates the link to point to the newer free course on Frontend Armory.


gaearon:
Thanks!

PR #3496, opened 2021-01-23
Update jsx-in-depth.md

Amr-Hamed:
- I believe there's a typing mistake in this section, It specifies that if a tag name is used in JSX as `<Foo />`, then `React` (NOT `Foo`) must be in scope to help understand the custom component.
- A confirmation to what I am pointing at in the first point, check the title of the section following the section with the proposed changes in this PR, it states `React Must Be in Scope {#react-must-be-in-scope}`


Amr-Hamed:
replaced with w new PR

PR #3497, opened 2021-01-23
Update jsx-in-depth.md

Amr-Hamed:
- I believe there's a typing mistake in this section, It specifies that if a tag name is used in JSX as <Foo />, then React (NOT Foo) must be in scope to help understand the custom component.
- A confirmation to what I am pointing at in the first point, check the title of the section following the section with the proposed changes in this PR, it states React Must Be in Scope {#react-must-be-in-scope}

### DIFF
--- a/content/docs/jsx-in-depth.md
+++ b/content/docs/jsx-in-depth.md
@@ -52,7 +52,7 @@ If you want to test out how some specific JSX is converted into JavaScript, you 
 
 The first part of a JSX tag determines the type of the React element.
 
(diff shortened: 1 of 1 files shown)


rickhanlonii:
Hey @Amr-Hamed, thanks for the suggestion!

The current text is correct, as `Foo` has to be in scope. As the next section states, React must also be in scope (though that's not necessarily true any more, thanks to https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html.

Going to close this, but thanks again!

PR #3501, opened 2021-01-28
Update Handling-Events

viren-vii:

### DIFF
--- a/content/docs/handling-events.md
+++ b/content/docs/handling-events.md
@@ -37,7 +37,7 @@ Another difference is that you cannot return `false` to prevent default behavior
 </a>
 ```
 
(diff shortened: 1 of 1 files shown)


sijusamson (commented):
I guess _**this could be**_  is more better as the same functionality can be achieved using different steps as well. 
So **_this could be_** is more appropriate IMO.
@rickhanlonii What is your say on it?

rickhanlonii:
Hey @viren-vii thanks for the suggestion. And thanks for qeighing in @sijusamson.

You don't always want to preventDefault, it depends on the use case. So you "could" do it but you don't necessarily need to.

Going to close, but thanks again!

PR #3502, opened 2021-01-28
Change the link to the NEW Yarn installation instructions

odeddugma:
Hello,

The README.md file direct users of React to use Yarn.
react-readme

The link directed to an old version of Yarn (Yarn 1. URL: https://classic.yarnpkg.com/en/docs/install/#windows-stable). I fix it so now it directs to the new version of Yarn (Yarn 2. URL: https://yarnpkg.com/getting-started/install).

Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ This repo contains the source code and documentation powering [reactjs.org](http
 
 1. Git
 1. Node: any 12.x version starting with v12.0.0 or greater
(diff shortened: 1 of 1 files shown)


rickhanlonii:
Hey @odeddugma, thanks for the suggestion. 

I believe all of the core contributors are using Yarn 1 so we should leave the instructions to match the workflow we're using.

Thanks again!

PR #3503, opened 2021-01-30
Update accessibility.md

PrinceRajRoy:
Updated sentence to be a bit more explicit, example code seemed to be outside of Parent class

### DIFF
--- a/content/docs/accessibility.md
+++ b/content/docs/accessibility.md
@@ -222,7 +222,7 @@ class Parent extends React.Component {
   }
 }
 
-// Now you can set focus when required.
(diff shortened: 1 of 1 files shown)


sijusamson (commented):
@PrinceRajRoy I guess the current statement is appropriate, and moreover it is already mentioned above this example, that if we need **_parent component can set focus to an element in a child component_**.
So IMO there's no need to change.

PrinceRajRoy:
@sijusamson it's still a bit ambiguous, given it's in the same snippet and not a separate one, could be confusing for some.

eps1lon:

PR #3509, opened 2021-02-04
Update thinking-in-react.md

ioilmio:
Links for step 4 and step 5 were inverted, now they reflect the correct implementations based on the instructions

### DIFF
--- a/content/docs/thinking-in-react.md
+++ b/content/docs/thinking-in-react.md
@@ -106,7 +106,7 @@ So finally, our state is:
 
 ## Step 4: Identify Where Your State Should Live {#step-4-identify-where-your-state-should-live}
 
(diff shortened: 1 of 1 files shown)


rickhanlonii:
Thanks for the suggestion, but I believe the order is currently correct:

Screen Shot 2021-02-04 at 12 05 36 PM
Screen Shot 2021-02-04 at 12 06 00 PM

ioilmio:
Hi Ricky, the order itself is not a problem,
even if I would prefer to have the "non-working input" section before, and
then the correct and working implementation after, but these are just
opinions.
------------------------
[image: image.png]
------------------------
The above link Thinking In React: Step 5
<https://codepen.io/gaearon/pen/LzWZvb> on CodePen <https://codepen.io>.
will lead to a pen where the actual input is working.


----------------------------
[image: image.png]

[image: image.png]

-----------------------------
The above link Thinking In React: Step 4
<https://codepen.io/gaearon/pen/qPrNQZ> on CodePen <https://codepen.io>.
The table can't be updated because the input is not working.
Just inverting the link should be fine.

PR #3513, opened 2021-02-09

DaniBedz:
Typo: "allow PropTypes to be proper applied" => "allow PropTypes to be properly applied"


eps1lon (approved):
Thanks!

We refer to "applied" so it's an adverb ("properly applied") not an adjective  ("the application is proper").

PR #3516, opened 2021-02-10
Fix redirects

gaearon:
Should fix https://github.com/reactjs/reactjs.org/issues/3138.

This method returns a Promise which wasn't awaited. I suspect this caused a race condition with another plugin to write to the same file, and the other plugin overwrote this one's result. Now they should write sequentially. (I hope.)


rickhanlonii (approved):
classic

gaearon:
that was it

PR #3518, opened 2021-02-10
Adding  "componentWillUnmount" in the Example function comment like in Tip section

AlaeddineMessadi:
Adding `componentWillUnmount` in the Example function in the comment like what is in the Tip section below


rickhanlonii:
Hey @AlaeddineMessadi, thanks for the suggestion. The effect that this comment is on doesn't return a cleanup function though, so it's not accurate to say it's like `componetWillUnmount`. The tip below is referring to `useEffect` in general, so it's accruate.

Thanks again!

AlaeddineMessadi:
Yeah, it makes sense :) thanks

PR #3522, opened 2021-02-12
Update node version of create-react-app

mottox2:
[create-react-app is now supported Node 10.16.0 or later version](https://github.com/facebook/create-react-app#creating-an-app). I updated it in docs.


lex111 (approved):

lex111:
@mottox2 thanks!

PR #3523, opened 2021-02-13
Updated paths to UMD builds for React and React-DOM in 'How to Contribute' page

tsairinius:
After building react (using either `yarn build` or `yarn build react/index,react-dom/index --type=UMD`), the resulting UMD builds no longer reside at `build/dist/react.development.js` and `build/dist/react-dom.development.js`, which is what is currently documented (see [How to Contribute](https://reactjs.org/docs/how-to-contribute.html): 

```
If you want to try your changes in your existing React project, you may copy build/dist/react.development.js, build/dist/react-dom.development.js, or any other build products into your app and use them instead of the stable version.
```

 The builds now appear to be located at `build/node_modules/react/umd/react.development.js` and `build/node_modules/react-dom/umd/react-dom.development.js`. This PR updates the paths in the documentation. 

Opened an issue for this: https://github.com/reactjs/reactjs.org/issues/3524


rickhanlonii (approved):
Awesome, thanks!

PR #3534, opened 2021-02-18
Update hooks-effect.md

lbfalvy:
Warn about variable-length dependency arrays

The question appeared under facebook/react#18229, where I also described other merits of this approach.

### DIFF
--- a/content/docs/hooks-effect.md
+++ b/content/docs/hooks-effect.md
@@ -470,7 +470,7 @@ useEffect(() => {
 
 In the future, the second argument might get added automatically by a build-time transformation.
 
->Note
+>Notes
 >
 >If you use this optimization, make sure the array includes **all values from the component scope (such as props and state) that change over time and that are used by the effect**. Otherwise, your code will reference stale values from previous renders. Learn more about [how to deal with functions](/docs/hooks-faq.html#is-it-safe-to-omit-functions-from-the-list-of-dependencies) and [what to do when the array changes too often](/docs/hooks-faq.html#what-can-i-do-if-my-effect-dependencies-change-too-often).
 >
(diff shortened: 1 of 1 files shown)


content/docs/hooks-effect.md
```diff
@@ -479,6 +479,8 @@ In the future, the second argument might get added automatically by a build-time
...
 >
 >We recommend using the [`exhaustive-deps`](https://github.com/facebook/react/issues/14920) rule as part of our [`eslint-plugin-react-hooks`](https://www.npmjs.com/package/eslint-plugin-react-hooks#installation) package. It warns when dependencies are specified incorrectly and suggests a fix.
+>
+>Also note that if you compose the dependencies array dynamically its length will not be checked, but the two arrays will be compared up to the length of the old one. When you pass variable-length arrays in second argument, always prepend the length of the array like so: `[myArray.length, ...myArray]`
```
eps1lon, line 483: Thanks for the work!

I would rather note that this is not supported. React will explicitly warn: 
> Warning: The final argument passed to useEffect changed size between renders. The order and size of this array must remain constant.

The behavior about what elements are compared is not mentioned and might just be an implementation detail.

To be sure, I would propose this suggestion as an addition to the warning if this is indeed the recommended fix. The relevant warning is issued here: https://github.com/facebook/react/blob/696e736be11b22e8d4fa40438d235b63db7fddde/packages/react-reconciler/src/ReactFiberHooks.old.js#L331-L339

If we want to recommend an "always" then the warning should include it as well. If there are cases where we don't want to include `myArray.length` we should describe them here.

eps1lon (changes requested):

content/docs/hooks-effect.md
lbfalvy: I checked the comparison logic in React and it appears that it only compares up to the shorter of the two lengths so while my suggestion works the explanation is factually incorrect.

In my rather limited experience with React, hooks being dependent on the indices of a variable length array is a pretty normal phenomenon and I can't think of any other way to handle them beside what I described here, so I think an "always" is warranted.

eps1lon: > I checked the comparison logic in React and it appears that it only compares up to the shorter of the two lengths so while my suggestion works the explanation is factually incorrect.

That's fine. But the source code of React does not imply a behavior we can rely on.

> so I think an "always" is warranted.

I'm not worried about whether this is warranted. If it is, then this should be part of the warning itself. So to confirm your proposal, please propose this to the React core as well and see what the core maintainers have to say. This will get a response faster.

PR #3541, opened 2021-02-24
Update hooks-state.md

nirasher1:
Change the count example to get a callback in setCount. It is necessary when the new value depends on the current value


eps1lon:
Thanks for the suggestion.

Since this is an introduction we need to strike a balance between simplisity and correctness. I'm closing this in favor of https://github.com/reactjs/reactjs.org/issues/3585 since this is already being worked on.

PR #3561, opened 2021-03-03
Remove CSS advice that is bad for accessibility

jamiebuilds:
From MDN: https://developer.mozilla.org/en-US/docs/Web/CSS/display

> `display: contents` -- Due to a bug in browsers this will currently remove the element from the accessibility tree — screen readers will not look at what's inside. See the Accessibility concerns section below for more details.

### DIFF
--- a/content/docs/strict-mode.md
+++ b/content/docs/strict-mode.md
@@ -73,10 +73,6 @@ class MyComponent extends React.Component {
 }
 ```
 
-> Note:
->
-> In CSS, the [`display: contents`](https://developer.mozilla.org/en-US/docs/Web/CSS/display#display_contents) attribute can be used if you don't want the node to be part of the layout.
-
 ### Detecting unexpected side effects {#detecting-unexpected-side-effects}
 
 Conceptually, React does work in two phases:


jamiebuilds:
More details: https://developer.mozilla.org/en-US/docs/Web/CSS/display#display_contents

gaearon:
I think we brought this up at some point with a browser vendor, and it looked like the widespread information about this lacked crucial nuance, which is _only that node disappears but not its children_. Is this something you can verify? If this is correct, then `display: contents` is fine to use on any wrapper that doesn't itself need to be announced. So we could edit to clarify that instead.

alice:
This issue is fixed in Chrome and Firefox, but not Safari, as of today.

gaearon:
I see, chrome tracker just wasn’t updated yet. Indeed marked as fixed now.

@alice do you know if my earlier assessment is correct? Does it only affect one element or the whole subtree?

alice:
That's also something you can check in developer tools:

Safari inspector inspecting an article with display: contents showing no accessibility information

Safari inspector inspecting a child of the article, showing accessibility information is computed

PR #3562, opened 2021-03-03
Fixing syntax error in #functions-as-children

cjang5:
Changing the call to the child prop, which is a function, to first access the child in the children list before invocation.

Found this while reading the docs and copy+pasting the snippet into a Codepen.

### DIFF
--- a/content/docs/jsx-in-depth.md
+++ b/content/docs/jsx-in-depth.md
@@ -388,7 +388,7 @@ Normally, JavaScript expressions inserted in JSX will evaluate to a string, a Re
 function Repeat(props) {
   let items = [];
   for (let i = 0; i < props.numTimes; i++) {
-    items.push(props.children(i));
+    items.push(props.children[0](i));
   }
   return <div>{items}</div>;
 }


eps1lon (changes requested):
The prior code works fine for me: https://codesandbox.io/s/listoftenthings-cm2su

Could you share a codesandbox where the current code passes?

cjang5:
> 
> 
> The prior code works fine for me: https://codesandbox.io/s/listoftenthings-cm2su
> 
> Could you share a codesandbox where the current code passes?

I remember doing this in a Codepen, just directly copy+pasted the inline code from the docs [here](https://reactjs.org/docs/jsx-in-depth.html#functions-as-children),

but I went and did it again just now and it looks like the only reason it was not working before was because the copy+paste incorrectly closed the `<Repeat>` tag on a separate line, i.e.

```
function ListOfTenThings() {
  return (
    <Repeat numTimes={10}>
      {(index) => <div key={index}>This is item {index} in the list</div>}    </Repeat>
  );
}
```

Once I moved the closing tag, `</Repeat>`, to its own line, the prior code ran fine.

I'm not a JS or React expert, so this is interesting behavior I cannot explain at this time. But looks like we can close this PR, my mistake!

cjang5:
And here's a pen if you want to see what I was seeing: https://codepen.io/cjang5/pen/rNWEWvw

PR #3568, opened 2021-03-08
Fix useReducer init param docs

leo-diehl:
Fix docs to correctly illustrator the use of the init param, passed to useReducer.
The last doc was giving the impression that the init funcion was comming from the upper scope, not the hook parameter

### DIFF
--- a/content/docs/hooks-reference.md
+++ b/content/docs/hooks-reference.md
@@ -314,7 +314,7 @@ function init(initialCount) {
   return {count: initialCount};
 }
 
-function reducer(state, action) {
+function reducer(state, action, init) {
   switch (action.type) {
     case 'increment':
       return {count: state.count + 1};


content/docs/hooks-reference.md
```diff
@@ -314,7 +314,7 @@ function init(initialCount) {
   return {count: initialCount};
 }
 
-function reducer(state, action) {
+function reducer(state, action, init) {
```
eps1lon, line 317: This param is not passed by React. The `init` referenced in `reducer` is declared  in https://github.com/reactjs/reactjs.org/pull/3568/files#diff-a1c8ec10dc40c86b4bbd55ad3f0126835bd45e9aa9e56c846a2e28ad6dd6acffR313.

Does this clear things up?

leo-diehl: Yes it does. Thank you

PR #3573, opened 2021-03-11
Fix hooks-faq.md, unnecessary dependency in example?

uid11:
If I understand correctly the intended API of the `useEventCallback` custom hook, the `dependencies` array should contain all the dependencies of the `fn` callback itself (as in the standard `useCallback` hook).

Thus, there is no need to additionally specify the `fn` callback itself in the dependencies of this effect -- you need to update the link to `fn` in the ref only when something from the `dependencies` array changes.

Moreover, now, due to the fact that `fn` is specified in the effect dependencies, this effect is called after every render of the `Form` -- since `fn`, as an arrow function literal, is created on every render.

### DIFF
--- a/content/docs/hooks-faq.md
+++ b/content/docs/hooks-faq.md
@@ -967,7 +967,7 @@ function useEventCallback(fn, dependencies) {
 
   useEffect(() => {
     ref.current = fn;
-  }, [fn, ...dependencies]);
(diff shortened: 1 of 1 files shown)


uid11:
Is there any other action I need to take to move this pull request forward (merge or close)?

PR #3576, opened 2021-03-12
Mention useReducer as a cause of rerender in React.memo

compulim:
When using `React.memo` HOC function, `useReducer` will also cause the component to be rendered despite their props didn't change.


eps1lon (approved):
Thanks :+1:

PR #3581, opened 2021-03-16
Update add-react-to-a-website.md

wansleynery:
Implemented the way to integrating React dependencies and its execution after page was loaded


w3bdesign (changes requested):
Spelling error here:

`mantain`

Should be `maintain`

wansleynery:
Thanks for the fix

PR #3589, opened 2021-03-22
remove unwanted vaiable from if statement

bishaludas:
Screenshot from 2021-03-23 00-24-12

The array for above tictactoe is :
```
let squares = [
"X", null, "O",
"X", null, "O",
"X", null, null
]
```

The `calculateWinner` function in tictactoe tutorial is satisfied for indexes [0, 3, 6] .  However the function contains an unwanted variable in if statement which is not required . Using the indexes , the check is done as follows : 
```
squares[0] && squares[0] === squares[3] && squares[0] === squares[6]
X && X === X && X === X
```
This can be done as below : 
```
squares[0] === squares[3] && squares[0] === squares[6]
X === X && X === X
```

### DIFF
--- a/content/tutorial/tutorial.md
+++ b/content/tutorial/tutorial.md
@@ -709,7 +709,7 @@ function calculateWinner(squares) {
   ];
   for (let i = 0; i < lines.length; i++) {
     const [a, b, c] = lines[i];
-    if (squares[a] && squares[a] === squares[b] && squares[a] === squares[c]) {
+    if (squares[a] === squares[b] && squares[a] === squares[c]) {
       return squares[a];
     }
   }


content/tutorial/tutorial.md
```diff
@@ -709,7 +709,7 @@ function calculateWinner(squares) {
   ];
   for (let i = 0; i < lines.length; i++) {
     const [a, b, c] = lines[i];
-    if (squares[a] && squares[a] === squares[b] && squares[a] === squares[c]) {
+    if (squares[a] === squares[b] && squares[a] === squares[c]) {
```
eps1lon, line 712: Given
```
null,null,null
X,X,X
O,O,null
```
This function would now return `null` when the winner is `X`

eps1lon (changes requested):
Thanks for the suggestion. But I do think this changes behavior.

content/tutorial/tutorial.md
bishaludas: Nice catch @eps1lon . This actually change the behavior of calculatewinner function. I had not tested for all win cases so its fault on my part.

Given:
```
null,null,null
X,X,X
O,O,null
```

Explanation:
when checking for indexes [0,1,2]
```
null === null && null === null
this return true
```
so calculatewinner function cannot check for indexes [3,4,5]

```
null && null === null && null === null
this returns null
```
so calculate winner can check for index [3,4,5]

Thanks for the review, I will close this PR

PR #3609, opened 2021-04-03
Update reference-caveats-solution.js

niklasp:
As referenced in https://reactjs.org/docs/context.html update the example to use the naming conventions used across the page.


eps1lon (approved):
Thanks :+1:

PR #3614, opened 2021-04-04
Be more concise about @types/react

rattrayalex:
The previous version implied that React bundles its own types, which is not the case. It also contained information which is better conveyed in TypeScript's docs (which it links to). It did not include a recommendation to include `@types/react-dom`, which most users will need.

### DIFF
--- a/content/docs/static-type-checking.md
+++ b/content/docs/static-type-checking.md
@@ -286,28 +286,16 @@ If you see no output, it means that it completed successfully.
 ### Type Definitions {#type-definitions}
 To be able to show errors and hints from other packages, the compiler relies on declaration files. A declaration file provides all the type information about a library. This enables us to use javascript libraries like those on npm in our project. 
 
-There are two main ways to get declarations for a library:
+React doesn't bundle its own declaration file. Instead we can get it from [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped), which is a huge repository of declarations crowd-sourced and managed by Microsoft and open source contributors. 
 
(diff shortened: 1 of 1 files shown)


content/docs/static-type-checking.md
```diff
@@ -286,28 +286,16 @@ If you see no output, it means that it completed successfully.
...
-```
-
-__Local Declarations__
-Sometimes the package that you want to use doesn't bundle declarations nor is it available on DefinitelyTyped. In that case, we can have a local declaration file. To do this, create a `declarations.d.ts` file in the root of your source directory. A simple declaration could look like this:
-
-```typescript
-declare module 'querystring' {
-  export function stringify(val: object): string
-  export function parse(val: string): object
-}
```
eps1lon, line 310: Why remove this section? It can still be helpful since you're probably not just using `react` and `react-dom`. It can also be helpful if `@types/react` becomes outdated.

rattrayalex: Good questions! The short answer to both is that it's no longer common for beginners to need to use this tool when working in the React ecosystem. 

If `@types/react` becomes outdated, the right thing to do would be to wait a day or so for the very large community of React TS users to update it, and only then update `react` and `@types/react` at the same time. Using `declare module` as a beginner would likely lead to problems, and may be a sign that they need to updated `@types/react` or are using the library incorrectly.

Using `declare module` for other situations is best documented in the TypeScript docs.

eps1lon: Even if you consider module augmentation of `@types/react` bad practice ([which it isn't](https://react-typescript-cheatsheet.netlify.app/docs/advanced/types_react_api/#adding-non-standard-attributes)), what about all the other libraries?

rattrayalex: The link you provide is a great place to learn about that – much more helpful than the `'querystring'` example that's here currently. Maybe we should add a link to that resource here instead? It also seems to have a wealth of other useful and relevant information.

PR #3616, opened 2021-04-05
Redirect to the previous page after language change

vkatatab:
> If user comes to an article on https://reactjs.org/ in English language (for example, he comes from google search), then he expects to be redirected to the same page when he switches the language.

There is an issue from [react repository](https://github.com/facebook/react/issues) that indicates this behavior, this pull request solve that [issue#21176](https://github.com/facebook/react/issues/21176)

### DIFF
--- a/src/components/LayoutHeader/Header.js
+++ b/src/components/LayoutHeader/Header.js
@@ -210,7 +210,8 @@ const Header = ({location}: {location: Location}) => (
                 borderRadius: 15,
               },
             }}
-            to="/languages">
+            to="/languages"
+            state={{prevPath: `${location.pathname}${location.hash}`}}>
             <LanguagesIcon />{' '}
             <span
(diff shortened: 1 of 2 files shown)


eps1lon (approved):
Looks good to me. I'm unfamiliar with the translation effort so another set of eyes would be nice. The only potential problem I can think of is when a given page only exists in the current language. But then we render a 404 error page which seems like a reasonable fallback?

@gaearon Any thoughts?

vkatatab:
In my opinion, that seems the expected behavior

BartoszKlonowski:
Hello @vkatatab!
The new react.dev manages translation more indirectly, by separate page mentioning separate translation repos out there. This makes this PR unfortunately no longer required, so I let me close this.
Still, thank you for working on it and for contributing!

PR #3619, opened 2021-04-08
Update hooks-reference.md

elanpang:
improve example code for useMemo that may cause misunderstanding

### DIFF
--- a/content/docs/hooks-reference.md
+++ b/content/docs/hooks-reference.md
@@ -364,7 +364,7 @@ Returns a [memoized](https://en.wikipedia.org/wiki/Memoization) callback.
 
 Pass an inline callback and an array of dependencies. `useCallback` will return a memoized version of the callback that only changes if one of the dependencies has changed. This is useful when passing callbacks to optimized child components that rely on reference equality to prevent unnecessary renders (e.g. `shouldComponentUpdate`).
 
-`useCallback(fn, deps)` is equivalent to `useMemo(() => fn, deps)`.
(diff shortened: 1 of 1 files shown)


content/docs/hooks-reference.md
```diff
@@ -364,7 +364,7 @@ Returns a [memoized](https://en.wikipedia.org/wiki/Memoization) callback.
 
 Pass an inline callback and an array of dependencies. `useCallback` will return a memoized version of the callback that only changes if one of the dependencies has changed. This is useful when passing callbacks to optimized child components that rely on reference equality to prevent unnecessary renders (e.g. `shouldComponentUpdate`).
 
-`useCallback(fn, deps)` is equivalent to `useMemo(() => fn, deps)`.
+`useCallback(fn, deps)` is equivalent to `useMemo(() => fn(), deps)`.
```
eps1lon, line 367: That would no longer be equivalent. This would mean that the memoized value is the returned value from calling `fn` not `fn` itself.

Does this clear things up?

elanpang: Thank you for your review, you are right, the returned value should be a function for both `left` and `right` if they are really equivalent.
I think maybe could add more description here for clarifying, as my understanding in most scenario `useMemo` will return a memoized value, not a function. This is a specific example for comparing with `useCallback.`

eps1lon: This section is specifically referring to `useCallback` to just the general use case.

PR #3626, opened 2021-04-13
add Glitch react starter to list of starters

jennschiffer:
This is our new "Hello World" for React, which will soon be one of our default starters on Glitch.com

### DIFF
--- a/content/docs/getting-started.md
+++ b/content/docs/getting-started.md
@@ -36,7 +36,7 @@ React has been designed from the start for gradual adoption, and **you can use a
 
 ### Online Playgrounds {#online-playgrounds}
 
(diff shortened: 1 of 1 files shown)


eps1lon (commented):
Any plans to make it a bit more minimal? The other starters are not that opinionated about file structure and include a lot less code. Right now the Glitch starter feels a bit intimidating.

jennschiffer:
@eps1lon We didn't have plans, no - but if what you mean to say is that a more minimal, less opinionated starter is more appropriate for this list, I think that's fair! I don't know what the rubric is, if any, for starters on this page. Do you find this minimal version of the starter less intimidating: https://glitch.com/edit/#!/glitch-mvp-react

PR #3628, opened 2021-04-14
docs: Add react learning path on frontend masters link

orama254:
updated the link to the react path on frontend masters

### DIFF
--- a/content/community/courses.md
+++ b/content/community/courses.md
@@ -28,7 +28,7 @@ permalink: community/courses.html
 
 - [Egghead.io](https://egghead.io/browse/frameworks/react) - Short instructional videos on React and many other topics.
 
-- [Frontend Masters](https://frontendmasters.com/learn/react/) - Video courses on React and other frontend frameworks.
+- [Frontend Masters](https://frontendmasters.com/learn/react/) - Video courses on React.
 
 - [Fullstack React](https://www.fullstackreact.com/) - The up-to-date, in-depth, complete guide to React and friends.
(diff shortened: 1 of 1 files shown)


content/community/courses.md
```diff
@@ -28,7 +28,7 @@ permalink: community/courses.html
 
 - [Egghead.io](https://egghead.io/browse/frameworks/react) - Short instructional videos on React and many other topics.
 
-- [Frontend Masters](https://frontendmasters.com/courses/) - Video courses on React and other frontend frameworks.
+- [Frontend Masters](https://frontendmasters.com/learn/react/) - Video courses on React and other frontend frameworks.
```
eps1lon, line 31: Considering the text after the link, it looks like the original link was intended. Maybe check when this was added to see if there was discussion about whether it should link to anything frontend related or just React.

orama254: Ok noted...I was trying to replicate the egghead.io link considering that maybe the exact path to react was not yet added to the Frontend Masters platform by then

eps1lon: Maybe back then they didn't have a react specific section so the description was adjusted accordingly. We should now update the link description to e.g 
```suggestion
- [Frontend Masters](https://frontendmasters.com/learn/react/) - Video courses on React.
```

orama254: Looks good...I've committed the changes as suggested

eps1lon (approved):
Thanks!

PR #3634, opened 2021-04-15
Small change in a statement in reconciliation.md

uchihamalolan:
Since underlying DOM node will have `class` attribute. I believe this change makes more sense

### DIFF
--- a/content/docs/reconciliation.md
+++ b/content/docs/reconciliation.md
@@ -59,7 +59,7 @@ When comparing two React DOM elements of the same type, React looks at the attri
 <div className="after" title="stuff" />
 ```
 
-By comparing these two elements, React knows to only modify the `className` on the underlying DOM node.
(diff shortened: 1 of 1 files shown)


content/docs/reconciliation.md
```diff
@@ -59,7 +59,7 @@ When comparing two React DOM elements of the same type, React looks at the attri
 <div className="after" title="stuff" />
 ```
 
-By comparing these two elements, React knows to only modify the `className` on the underlying DOM node.
+By comparing these two elements, React knows to only modify the `class` attribute on the underlying DOM node.
```
eps1lon, line 62: While React does set the attribute, that's only determined by the renderer not the reconciler. `react-dom` could very well decide to set the property instead, or even use `classList`. So I don't think this distinction is relevant to explaining how reconciliation works.

PR #3636, opened 2021-04-16
Update reference-react.md

ZackingIt:
The phrase "React.Component doesn't implement `shouldComponentUpdate`" is strictly incorrect:  React.Component *does* implement `shouldComponentUpdate`, it just always returns `true`.  Saying that it "doesn't implement" might suggest that the method is not available when we extend `React.Component`, which would be highly confusing and misleading.  Instead, we should emphasize that `shouldComponentUpdate` is "pre-implemented" for PureComponent, and we should also try to distinguish between a *change* in state, and an update in (i.e., setting a state slice's value to anything, even the same value)

### DIFF
--- a/content/docs/reference-react.md
+++ b/content/docs/reference-react.md
@@ -104,7 +104,7 @@ See the [React.Component API Reference](/docs/react-component.html) for a list o
 
 ### `React.PureComponent` {#reactpurecomponent}
 
-`React.PureComponent` is similar to [`React.Component`](#reactcomponent). The difference between them is that [`React.Component`](#reactcomponent) doesn't implement [`shouldComponentUpdate()`](/docs/react-component.html#shouldcomponentupdate), but `React.PureComponent` implements it with a shallow prop and state comparison.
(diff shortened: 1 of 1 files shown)


eps1lon (commented):
> saying that it "doesn't implement" might suggest that the method is not available when we extend React.Component, which would be highly confusing and misleading.

My class-knowledge is a bit rusty in JS but how did you access `shouldComponentUpdate` from within a class extending `React.Component`? 

As far as I can tell `shouldComponentUpdate` is not reachable in `React.Component` which is why the reconciler checks if it exists before calling it.

Technically "but `React.PureComponent` implements it with a shallow prop and state comparison." is also wrong since `React.PureComponent` doesn't implement it as in `class React.PureComponent { shouldComponentUpdate() {} }`. Again, the reconciler is the one actually implementing the described behavior.

[react-reconciler code responsible for `shouldComponentUpdate`](https://github.com/facebook/react/blob/782f689ca8380a715ab0c24e326fc293e5d25647/packages/react-reconciler/src/ReactFiberClassComponent.old.js#L315-L366)

ZackingIt:
Perhaps we're having a misunderstanding about what "reachable" means.  `shouldComponentUpdate` is reachable in React.Component, by that I mean:
```    shouldComponentUpdate() {
        return false
    }
```
in a component which extends `React.Component`

ZackingIt:
Perhaps we're having a misunderstanding about what "reachable" means. shouldComponentUpdate is reachable in React.Component, by that I mean:

shouldComponentUpdate() {
        return false
    }
in a component which extends React.Component actually does in fact prevent the component from being updated in a scenario where new state is being provided.  React.Component absolute has `shouldComponentUpdate`, in fact it's literally defined as part of the React.Component API :)

PR #3641, opened 2021-04-19
[RFC]: I added a usage about useState hooks

yklydxtt:
there is the issue :https://github.com/reactjs/reactjs.org/issues/3640

### DIFF
--- a/content/docs/hooks-state.md
+++ b/content/docs/hooks-state.md
@@ -187,6 +187,13 @@ In a function, we already have `setCount` and `count` as variables so we don't n
   </button>
 ```
 
+In a function, you can also change count like this: 
+```js
+  <button onClick={() => setCount(prevValue => prevValue + 1)}>
+    Click me
+  </button>
+```
+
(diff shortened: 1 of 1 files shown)


eps1lon:
This will be adressed in the new React docs. I'm not comfortable with this intermediate solution since it might just confuse begginers: it's unclear why both APIs exist for the (seemingly!) same thing (https://github.com/reactjs/reactjs.org/issues/3308#issuecomment-762823578 has more context).

yklydxtt:
> This will be adressed in the new React docs. I'm not comfortable with this intermediate solution since it might just confuse begginers: it's unclear why both APIs exist for the (seemingly!) same thing ([#3308 (comment)](https://github.com/reactjs/reactjs.org/issues/3308#issuecomment-762823578) has more context).

ok,thanks for your reply

BartoszKlonowski:
Closing as per discussion and issue closed.
Thank you, @yklydxtt for contributing!

PR #3642, opened 2021-04-19
Removed createBlockingRoot API docs and added a warning note

bvaughn:
Follow up to https://github.com/facebook/react/issues/21316

### DIFF
--- a/content/docs/concurrent-mode-adoption.md
+++ b/content/docs/concurrent-mode-adoption.md
@@ -21,6 +21,10 @@ next: concurrent-mode-reference.html
 >
 >This documentation is aimed at early adopters and people who are curious. **If you're new to React, don't worry about these features** -- you don't need to learn them right now.
 
+>Caution:
+>
+>All references below to "blocking mode" and `createBlockingRoot` are outdated and should be ignored.
+
 </div>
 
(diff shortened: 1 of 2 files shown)


eps1lon:
Want to remove `timeoutMS` as well while you're at it (https://github.com/facebook/react/pull/19703)? :smile:

bvaughn:
Nope :D

eps1lon:
> Nope  Sorry this was just a quick bandaid fix

Ok, I just looked for every mention of `timeoutMS` and it does look a bit more involved. I don't know enough about the current semantics to update sections such as

> If we make our API responses take 5 seconds, [we can confirm](https://codesandbox.io/s/relaxed-greider-suewh) that now React "gives up" and transitions anyway to the next screen after 3 seconds. This is because we passed `{timeoutMs: 3000}` to `useTransition()`. For example, if we passed `{timeoutMs: 60000}` instead, it would wait a whole minute.

gaearon (approved):
Seems fine but a lot of other things are outdated too so I don’t know that this one needs special treatment

bvaughn:
True! I only proposed this because it seemed to cause confusion but I don't feel strongly one way or another.

PR #3643, opened 2021-04-20
Update code-splitting.md (setup →  set up)

bryndyment:
(`setup` is the noun form and `set up` is the verb form.)


eps1lon (approved):
Thanks!

PR #3648, opened 2021-04-21
update: add link to Vite

padunk:
More and more browsers support ESM and Vite is also support creating new React app.


nickmccurdy (approved):

BartoszKlonowski:
Hello @padunk!
As Vite is already mentioned [here](https://react.dev/learn/start-a-new-react-project#can-i-use-react-without-a-framework) or [here](https://react.dev/learn/react-compiler#usage-with-vite), let me close this PR as outdated, moreover considering the fact that we mention it "inline" without special toolchains section.

PR #3650, opened 2021-04-21
`lifecycle` -> `lifecycle method`

danield770:


eps1lon (approved):
Makes sense. More explicit that way :+1:

PR #3652, opened 2021-04-22
Include where `setState` is comming from

m019m1:
Corrected mistake

### DIFF
--- a/content/docs/hooks-reference.md
+++ b/content/docs/hooks-reference.md
@@ -76,7 +76,7 @@ If your update function returns the exact same value as the current state, the s
 > Unlike the `setState` method found in class components, `useState` does not automatically merge update objects. You can replicate this behavior by combining the function updater form with object spread syntax:
 >
 > ```js
-> setState(prevState => {
+> useState(prevState => {
 >   // Object.assign would also work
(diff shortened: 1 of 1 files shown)


content/docs/hooks-reference.md
```diff
@@ -76,7 +76,7 @@ If your update function returns the exact same value as the current state, the s
 > Unlike the `setState` method found in class components, `useState` does not automatically merge update objects. You can replicate this behavior by combining the function updater form with object spread syntax:
 >
 > ```js
-> setState(prevState => {
+> useState(prevState => {
```
eps1lon, line 79: Appreciate the feedback!

I think it makese more sense to include where `setState` is comming from. We don't want `useState` here but the updater function i.e.

```suggestion
> const [state, setState] = useState({});
> setState(prevState => {
```

eps1lon (changes requested):

content/docs/hooks-reference.md
m019m1: yes, of course! I confused myself)

eps1lon (approved):

eps1lon:
@m019m1 Thanks!

PR #3662, opened 2021-04-28
fix(hook-reference): add missing props

neighborhood999:
Add missing props in `useImperativeHandle` section.

### DIFF
--- a/content/docs/hooks-reference.md
+++ b/content/docs/hooks-reference.md
@@ -448,7 +448,7 @@ function FancyInput(props, ref) {
       inputRef.current.focus();
     }
   }));
-  return <input ref={inputRef} ... />;
+  return <input ref={inputRef} {...props} />;
 }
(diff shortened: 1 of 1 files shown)


content/docs/hooks-reference.md
```diff
@@ -448,7 +448,7 @@ function FancyInput(props, ref) {
       inputRef.current.focus();
     }
   }));
-  return <input ref={inputRef} ... />;
+  return <input ref={inputRef} {...props} />;
```
eps1lon, line 451: Whether or not you should spread props has different opinions. Since this is not relevant to `useImperativeHandle` we could just omit it and avoid opinions?

neighborhood999: Agree. I thought someone will confuse `...` represent means in the here.

Feel free to close PR, if don't need it, thank you for reviewing.

PR #3665, opened 2021-05-01
Updated The React Podcast link and attribution

jonstuebe:
I noticed that the link and attribution for the React Podcast was out of date.


BartoszKlonowski:
Hello @jonstuebe!
As the podcasts section is no longer maintained and on the legacy site there's also a note about that, I believe it's OK to leave that link outdated, as it's not included in the react.dev docs.
Let me close this PR as outdated.
Thank you!

Issue #3656, opened 2021-04-24
alert() not working in codepen, consider alternative onclick functions for tutorial


user670:
https://github.com/reactjs/reactjs.org/blob/master/content/tutorial/tutorial.md

At some part of this tutorial (um, line 238 of that md file), I was asked to give the squares an onclick function of `alert("click")`. However, apparently Chrome has blocked this from running in codepen (`A different origin subframe tried to create a JavaScript dialog. This is no longer allowed and was blocked. See https://www.chromestatus.com/feature/5148698084376576 for more details.`), and can be confusing to people.

The next best alternative for `alert` that I can think of is `console.log`, and logged messages can be seen in both the browser's console and codepen's console.

denizbinay:
At first I had problems reproducing it. Cross Origin Iframe ban is just implemented in Chrome beta 91.0.4472.27. There are still some problems with Enterprise-Application, but it won't be long until it is in the stable version. It's really better to replace alert(). (https://groups.google.com/a/chromium.org/g/blink-dev/c/hTOXiBj3D6A/m/JtkdpDd1BAAJ?pli=1).

Is this the only place where alert() occurs?

user670:
> At first I had problems reproducing it. Cross Origin Iframe ban is just implemented in Chrome beta 91.0.4472.27. There are still some problems with Enterprise-Application, but it won't be long until it is in the stable version. It's really better to replace alert(). (https://groups.google.com/a/chromium.org/g/blink-dev/c/hTOXiBj3D6A/m/JtkdpDd1BAAJ?pli=1).
> 
> This should be replaced in the entire documentation.

No idea if `alert` is used anywhere else.

Oh, um, I think I forgot to mention that I'm on Chrome Beta, not Chrome Stable, and that's probably why I get this problem. Sorry about that.

PR #3666, opened 2021-05-03
Replace alert() with console.log() (#3656)

denizbinay:
The following Chromium feature is currently being tested.
"Feature: Remove alert(), confirm(), and prompt for cross origin iframes§
https://www.chromestatus.com/feature/5148698084376576

Therefore, alert() will no longer work on Codepen. console.log() still works. The feature is not yet in the stable version of Chromium, but in the current beta (tested with Chrome beta version 91.0.4472.27).


denizbinay:
I have forgotten the corresponding texts. I'll do it again tomorrow.

PR #3670, opened 2021-05-06
Update devtools screenshot to latest version

skaidra-bake:
Hi, I was working through the tutorial, and my devtools did not look like the ones provided in the image. It was a little bit confusing, so I think it could be updated. I think that the screenshot dimensions are the same but if there's anything that should be changed please let me know.


eps1lon (approved):
Very nice, thanks! I think it makes sense to include the  chrome devtools tablist to identify the react devtools tab

PR #3683, opened 2021-05-12
Added missing key attribute in list item in List and Key Example

novamaster-official:
Line Number 181
Hello, I added Key attribute in the list item which is missing


eps1lon:
Duplicate of https://github.com/reactjs/reactjs.org/pull/3684

PR #3684, opened 2021-05-12
Added Missing Key attribute to list item.

novamaster-official:
Hello, I added missing key attribute to the list item in the
in the example of **Correct Key Usage**
Please Review my changes.
Thank you :)

### DIFF
--- a/content/docs/lists-and-keys.md
+++ b/content/docs/lists-and-keys.md
@@ -178,7 +178,7 @@ ReactDOM.render(
 ```javascript{2,3,9,10}
 function ListItem(props) {
   // Correct! There is no need to specify the key here:
-  return <li>{props.value}</li>;
+  return <li key={props.key}>{props.value}</li>;
 }
 
 function NumberList(props) {


content/docs/lists-and-keys.md
```diff
@@ -178,7 +178,7 @@ ReactDOM.render(
 ```javascript{2,3,9,10}
 function ListItem(props) {
   // Correct! There is no need to specify the key here:
-  return <li>{props.value}</li>;
+  return <li key={props.key}>{props.value}</li>;
```
eps1lon, line 181: `key` is not reachable from the `props`. It is only handled by React. It's [in the docs](https://reactjs.org/docs/lists-and-keys.html#extracting-components-with-keys) and even has a specific comment in the code you're changing. 

Does this clear things up?

eps1lon (changes requested):

content/docs/lists-and-keys.md
novamaster-official: Yes, Thank you :)

PR #3691, opened 2021-05-20
Updated onclick method from props to prevent error

human-caching:
I was going through the  tutorial and got the following error because the onclick props were expressed with parentheses:
"Error: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops."

handle_click_react_tutorial

### DIFF
--- a/content/tutorial/tutorial.md
+++ b/content/tutorial/tutorial.md
@@ -440,7 +440,7 @@ class Square extends React.Component {
     return (
       <button
         className="square"
-        onClick={() => this.props.onClick()}
+        onClick={() => this.props.onClick}
       >
         {this.props.value}
       </button>


content/tutorial/tutorial.md
```diff
@@ -440,7 +440,7 @@ class Square extends React.Component {
     return (
       <button
         className="square"
-        onClick={() => this.props.onClick()}
+        onClick={() => this.props.onClick}
```
eps1lon, line 443: That wouldn't call the `onClick` prop now though which is what we wanted.

Could elaborate how you reproduce the "Error: Maximum update depth exceeded" with this particular code?

human-caching: Sorry for not seeing this before. I was going through the tutorial from scratch. When I get to this part of the tutorial, I get the error. However changing it to `onClick` fixed the issue.

eps1lon: > When I get to this part of the tutorial, I get the error.

Sorry, I'm not following what you mean by this. You you explain in detail what you did (from the start) to produce the error? Ideally include the full code that you produced e.g. in a [codesandbox](https://react.new)

PR #3697, opened 2021-05-25
Change example to use a <form>

gaearon:
Links with `#` are just bad. Also, you probably don't want to hijack a link unless you're writing a router and know what you're doing. Even there, you'd want to have a `href` and to determine whether to prevent navigation by checking whether the user has any special key pressed or not. Let's just change this to an example that doesn't have such significant caveats.

### DIFF
--- a/content/docs/handling-events.md
+++ b/content/docs/handling-events.md
@@ -29,27 +29,27 @@ is slightly different in React:
 </button>
 ```
 
-Another difference is that you cannot return `false` to prevent default behavior in React. You must call `preventDefault` explicitly. For example, with plain HTML, to prevent the default link behavior of opening a new page, you can write:
+Another difference is that you cannot return `false` to prevent default behavior in React. You must call `preventDefault` explicitly. For example, with plain HTML, to prevent the default form behavior of submitting, you can write:
 
 ```html
-<a href="#" onclick="console.log('The link was clicked.'); return false">
-  Click me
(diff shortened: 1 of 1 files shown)


content/docs/handling-events.md
```diff
@@ -29,27 +29,27 @@ is slightly different in React:
...
 
 In React, this could instead be:
 
-```js{2-5,8}
-function ActionLink() {
-  function handleClick(e) {
+```js{3}
+function Form() {
+  function handleSubmit(e) {
     e.preventDefault();
```
bvaughn, line 45: Always disliked the `e` naming convention.

bvaughn (approved):

content/docs/handling-events.md
gaearon: One benefit over `event` is that there's no built-in called `e`. So you're not going to reference `window.event` accidentally.

rickhanlonii (approved):

content/docs/handling-events.md
bvaughn: Maybe! Hopefully lint rules (or TypeScript/Flow) would be sufficient to catch that though.

gaearon: Unfortunately default lint configurations tend to allow it because it's a legit global. But we do forbid it in CRA preset for this reason.

PR #3701, opened 2021-05-28
Updating name of variable to which returned value of setState is assigned

CodeChari:
…ed for uniformity

in CodePen it is 'prevState' and in the tutorial, it is 'state', which creates confusion.


eps1lon (approved):
Good catch, thanks!

Issue #3659, opened 2021-04-26
Docs navigation styling is very confusing


KrimFandango:
![image](https://user-images.githubusercontent.com/27821441/116087749-78b4ea80-a6a1-11eb-921b-05640739eafa.png)

Problem: React docs' next/previous article links are not easy to spot

Current solution: describing the location of the link in the article, and also adding another link to the next page because the main link isn't visible enough

I suggest changing the color of the navigation section to have white background/dark text with a smaller margin from the top and bigger margin from the bottom, to look like ti belongs to the article, since at the moment it looks like it belongs to the footer.

denizbinay:
haha, I actually read all that and still never noticed that there was a navigation :D Sometimes I cursed because the link at the end of the article was missing.

I made two quick devtool mockups. There are several points.

- A light background works much better. What do you think? Grey or white? Grey blurs in the transition, but would fit from the idea. This background otherwise only appears in the navigation. Nevertheless, my feeling says white.
- Unfortunately, the Brandcolor no longer contrasts, so I guess both have to be black.
- Is it really necessary to reduce the distance? I think the colour changes alone do the trick.

![react-js-org-navigation](https://user-images.githubusercontent.com/69312106/116927479-01acc280-ac5c-11eb-9d48-5fe7138d837d.PNG)



![react-js-org-navigation-white](https://user-images.githubusercontent.com/69312106/116927525-11c4a200-ac5c-11eb-82ef-7959e339fb54.PNG)

reza-boroushaki:
cool

denizbinay:
Thank you, but which option is better?

makafsal:
I've fixed the issue, anyone please verify?

BartoszKlonowski:
Dears, on the latest of react.dev there is the updated styling of "next" and "previous" buttons, but additionally there's the side navigation bar on the right. As I find these being sufficient I will close this issue as not planned.
However, if you see any space for improvements feel free to create new issue.
Thank you!

PR #3712, opened 2021-05-30
#3659: Bottom navigation issue fixed

makafsal:
The fix for the below issue. The bottom navigation is detached from the footer and the background color changed to look like the part of the page.

#3659 Docs navigation styling is very confusing


BartoszKlonowski:
Closing as per comment in the issue.
Thank you, @makafsal for taking the time to contribute 👍

PR #3715, opened 2021-05-31
Fix broken link to devtools

tupini07:
The old link (https://github.com/facebook/react-devtools) pointed to an archived repo. I've updated it to point to the devtools package in the React repo (https://github.com/facebook/react/tree/master/packages/react-devtools)


eps1lon (approved):
Thanks!

PR #3728, opened 2021-06-07
Adds empty list as a valid child

mechamobau:
Empty lists are valid children and have the same behavior of `false`, `null`, `undefined`, and `true`.

### DIFF
--- a/content/docs/jsx-in-depth.md
+++ b/content/docs/jsx-in-depth.md
@@ -424,7 +424,7 @@ Children passed to a custom component can be anything, as long as that component
 <div>{true}</div>
 ```
 
-This can be useful to conditionally render React elements. This JSX renders the `<Header />` component only if `showHeader` is `true` or is an array with at least one item:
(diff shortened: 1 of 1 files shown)


content/docs/jsx-in-depth.md
```diff
@@ -404,9 +404,9 @@ function ListOfTenThings() {
 
 Children passed to a custom component can be anything, as long as that component transforms them into something React can understand before rendering. This usage is not common, but it works if you want to stretch what JSX is capable of.
 
-### Booleans, Null, and Undefined Are Ignored {#booleans-null-and-undefined-are-ignored}
+### Booleans, Null, Undefined and Empty Lists Are Ignored {#booleans-null-and-empty-lists-undefined-are-ignored}
```
eps1lon, line 407: It's not really ignoring empty lists. `[undefined]` is also not rendering. I'm not sure we need to mention lists here since React isn't rendering the list but rendering each item of the list.

mechamobau: Good point, so I'll close this PR :) Thanks for the advising!

PR #3731, opened 2021-06-10
Tutorial Revision

FellowshipOfThePing:
Revised tutorial to use function-based components instead of class-based, and hooks instead of lifecycle methods. I don't expect this PR to be accepted, but I do hope the React team will release something similar to this in the near future, as the current tutorial seems to be somewhat outdated, and not very user friendly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ The documentation is divided into several sections with a different tone and pur
 1. `git push my-fork-name the-name-of-my-branch`
 1. Go to the [reactjs.org repo](https://github.com/reactjs/reactjs.org) and you should see recently pushed branches.
 1. Follow GitHub's instructions.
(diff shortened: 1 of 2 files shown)


gaearon:
Hi! We're in the middle of rewriting the docs from scratch. It's not clear that we'll keep this tutorial there, at least in the initial revision. In the current old docs, the main content is all classes. So it's confusing either way. If the tutorial is Hooks, why does the main content use Classes? At least for now it's consistent: we teach Classes first, then Hooks. And in new docs it will be the reverse order. Hope that explains why we're not doing changes to docs now.

gaearon:
(But—really appreciate the effort. It pains me to see the old docs every day. I hope we can get at least the beta out as soon as possible.)

PR #3733, opened 2021-06-11
Update useTransition API in docs

justinhaaheim:
The order of these parameters has now changed from what's in the docs, so this updates the docs to match the current spec.

### DIFF
--- a/content/docs/concurrent-mode-patterns.md
+++ b/content/docs/concurrent-mode-patterns.md
@@ -79,7 +79,7 @@ Finally, we'll use it inside the `App` component:
 ```js{3-5}
 function App() {
   const [resource, setResource] = useState(initialResource);
-  const [startTransition, isPending] = useTransition({
+  const [isPending, startTransition] = useTransition({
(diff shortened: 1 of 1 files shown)


bvaughn (approved):

gaearon (changes requested):
I'm not sure this makes sense. The page is explicitly marked as outdated at the top. Maybe we need to make this marker more prominent. Or just delete these pages.

Currently, linked sandboxes match the code. If we update the code, sandboxes won't match it. This is confusing. If you update sandboxes, then their _behavior_ won't match the page because `timeoutMs` option is gone. It will also give a wrong impression that timeout option exists in the recent version because sandboxes presumably will use 18 alpha after being updated.

At least now it's all consistently outdated. Thoughts?

gaearon:
One alternative is a bigger reworking that would remove the mention of timeouts.

BartoszKlonowski:
Dears, I can see this already got updated in #5311, so let me close this PR as superseded.

PR #3739, opened 2021-06-16
remove data-reactroot reference

Gregoor:
It seems like it was removed in this PR https://github.com/facebook/react/pull/20996/files

Now I'm wondering if there are any internal attributes left, and if not, the whole sentence should be changed?


BartoszKlonowski:
Hello @Gregoor, thank you for taking the time to contribute!
I see this part is no longer present in the latest of react.dev, so I will close this PR as outdated.
Feel invited to create new PRs for other open issues.

PR #3741, opened 2021-06-16
Update courses.md

marcospereda:
you said the course at scrimba its free, but scrimba is not free at all


plondon:
I moved this to paid courses: https://github.com/marcospereda/reactjs.org/pull/1

PR #3749, opened 2021-06-24
Update portals.md

jmargolisvt:

### DIFF
--- a/content/docs/portals.md
+++ b/content/docs/portals.md
@@ -4,7 +4,7 @@ title: Portals
 permalink: docs/portals.html
 ---
 
-Portals provide a first-class way to render children into a DOM node that exists outside the DOM hierarchy of the parent component.
+Portals provide a first-class way to render children into a DOM node that exists outside the DOM hierarchy of the parent component.  They _must_ be returned from a component. The `createPortal` function is not a normal funciton that can be called from just anywhere.
 
 ```js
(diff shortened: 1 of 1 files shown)


content/docs/portals.md
```diff
@@ -4,7 +4,7 @@ title: Portals
 permalink: docs/portals.html
 ---
 
-Portals provide a first-class way to render children into a DOM node that exists outside the DOM hierarchy of the parent component.
+Portals provide a first-class way to render children into a DOM node that exists outside the DOM hierarchy of the parent component.  They _must_ be returned from a component. The `createPortal` function is not a normal funciton that can be called from just anywhere.
```
takeaways, line 7: What does mean "just anywhere"?

jmargolisvt: I'm stressing that this function will not work as expected if you call it from places other than as the `return` from a component.  One would think you could call this function from a helper, as a side-effect, etc. but it doesn't work when you call it from those other place (i.e. "just anywhere").  I'm submitting this addition (or similar if you think it can be said more clearly) because it's not obvious that this restriction exists.

takeaways (approved):

BartoszKlonowski:
Hello @jmargolisvt!
Unfortunately the sentence you update in this PR is no longer present in the latest of react.dev docs.
I will close this PR as outdated.
Thank you!

PR #3751, opened 2021-06-26
Fix incorrect order for input locking

tejasvi:
The input should be initially locked and show "hi" after a delay.


eps1lon (changes requested):
The previous code looks correct to me. When the value is controlled, but no onChange handler is declared, you can't change the value. But once you switch to uncontrolled, the input is editable:

https://codesandbox.io/s/input-controlled-null-sk678

tejasvi:
I see your point. My apologies for the PR, I am still learning.

PR #3753, opened 2021-06-27
setState gets undefined , so use arrow function

Js4Life:

### DIFF
--- a/content/docs/conditional-rendering.md
+++ b/content/docs/conditional-rendering.md
@@ -83,14 +83,15 @@ class LoginControl extends React.Component {
     this.state = {isLoggedIn: false};
   }
 
-  handleLoginClick() {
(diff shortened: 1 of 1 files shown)


Haroenv:
the existing code is actually correct, I think in your version you might have forgotten to bind in the constructor,

```
    this.handleLogoutClick = this.handleLogoutClick.bind(this);
```

(that's already done in the existing code)

harish-sethuraman:
https://codepen.io/gaearon/pen/QKzAgB?editors=0010

It seems to work fine in the codepen. I guess you missed to bind the function as mentioned in the previous comment.

```js
this.handleLoginClick = this.handleLoginClick.bind(this);
this.handleLogoutClick = this.handleLogoutClick.bind(this);
```

PR #3761, opened 2021-07-01
Fix broken link in the web-components.md file

gabrielkuhn:
The current URL is leading to a 404 page because webcomponents's polyfills repo default branch is master instead of main.


eps1lon (approved):
Thanks!

PR #3765, opened 2021-07-03
Fix broken Props vs State link

Laffini:
The Props vs State resource link is broken. This change updates the link so that it works.


eps1lon (approved):
Thanks!

PR #3777, opened 2021-07-08
Update tutorial.md

plondon:
I've included a third setup option, which is to navigate to https://codeamigo.dev/lessons/start/54 and complete the tutorial there, which is setup to be an interactive version of the docs.


plondon:
Going to move this to free courses instead

plondon:
https://github.com/reactjs/reactjs.org/pull/3780

PR #3778, opened 2021-07-08
how do I update nested state properties

lvndry:
Fixing https://github.com/reactjs/reactjs.org/issues/1101 by adding an example of how to update a nested state property.


BartoszKlonowski:
Closing as per discussion in the issue.
Thank you!

PR #3784, opened 2021-07-09
Making some changes in tutorial.md file of react.

SS-one-AND-two:


SS-one-AND-two:
This is an commit

eps1lon:
The previous heading structure looks intentional. 

Issue #3785, opened 2021-07-10
Typo Bug in React docs/react-dom.html#render


iemprashant:
![Screenshot (384)_LI](https://user-images.githubusercontent.com/53941491/125154621-7af2f500-e178-11eb-9a7f-cc4a233bddec.jpg)
This Brackets typo can be confusing in understanding the rendering function.

harish-sethuraman:
This is not a bug. We are changing the format on the new site. Thanks 😄 .

PR #3786, opened 2021-07-10
Fix Issue #3785

iemprashant:
I have fixed the basic typo error in the render  function doc as mentioned in #3785


harish-sethuraman:
We are changing the format on the new site. Sorry for the late turnaround time. Thanks 😄 .

PR #3787, opened 2021-07-10
fixed typo in code

ConaGo:
small typo fix


eps1lon:
`onsubmit` is working here since this is plain HTML not React.

PR #3788, opened 2021-07-12
Local state is updated without using this.setState

Nitisha-Srivastav:
In the first example of Error Boundaries the method getDerivedStateFromError() is setting the local state (this.state = { hasError: false };) of the component without using this.setState (return { hasError: true };).

### DIFF
--- a/content/docs/error-boundaries.md
+++ b/content/docs/error-boundaries.md
@@ -33,7 +33,12 @@ class ErrorBoundary extends React.Component {
 
   static getDerivedStateFromError(error) {
     // Update state so the next render will show the fallback UI.
-    return { hasError: true };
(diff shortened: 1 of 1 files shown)


eps1lon (changes requested):
The previous code is correct: https://reactjs.org/docs/react-component.html#static-getderivedstatefromerror

`this` would not refer to the current instance since `getDerivedStateFromError` is a `static` method.

Does this clear things up?

NitishaSrivastav:
Thanks for explaining. I got your point.

eps1lon:
> @eps1lon Thanks for explaining. I got your point.

Glad to hear it. Would you agree that we can close this PR then?

NitishaSrivastav:
Yes you can close it.

Nitisha-Srivastav:
@eps1lon  yes you can close it.

PR #3795, opened 2021-07-16
Updating strict-mode.md to mention double render

michal-ciechan:
I have spent last 4 hours trying to figure out why my component is rendering twice.

Stumbled upon an [article](https://deniapps.com/blog/why-the-react-component-renders-twice), that mentioned that `React.StrictMode`  may be the culprit. (turned out it was)

I am not sure of the exact wording to use, but I definitely think it i s worth adding various mentions of this "double render" regarding `StrictMode`.

### DIFF
--- a/content/docs/strict-mode.md
+++ b/content/docs/strict-mode.md
@@ -8,6 +8,8 @@ permalink: docs/strict-mode.html
 
 > Note:
 >
+> Strict mode may "_render_" your application twice, but only one will be visible to the DOM. The second render is necessary to carry out its checks.
+>
 > Strict mode checks are run in development mode only; _they do not impact the production build_.
 
 You can enable strict mode for any part of your application. For example:


content/docs/strict-mode.md
```diff
@@ -8,6 +8,8 @@ permalink: docs/strict-mode.html
 
 > Note:
 >
+> Strict mode may "_render_" your application twice, but only one will be visible to the DOM. The second render is necessary to carry out its checks.
```
eps1lon, line 11: The article already mentions this. This now gives the impression this is the only effect StrictMode has. But it has many more effects.

michal-ciechan: Could you point where it in this doc it already specifies that it renders twice? Clearly I am missing it....

eps1lon: https://reactjs.org/docs/strict-mode.html#detecting-unexpected-side-effects

PR #3796, opened 2021-07-17
Update concurrent-mode-suspense.md

AssafKr:
Link for `usePreloadedQuery` is broken


harish-sethuraman:
The link seems to be updated right now. Thanks for sending PR and sorry for the late turn around time.

PR #3807, opened 2021-07-24
Update isPending and startTransition

clive-h-townsend:
Per [20976](https://github.com/facebook/react/pull/20976), appears that isPending is now the first parameter.


BartoszKlonowski:
Hello @clive-h-townsend!
I checked the page of [useTransition](https://react.dev/reference/react/useTransition) in the react.dev docs and it appears that it's adjusted to the params order. So I will close this PR as no longer required.
Still, thank you for addressing it! 👍

PR #3810, opened 2021-07-25
Fix URL's anchor to "Callback Refs" page

tusba:
Replace "#the-ref-callback-attribute" with "#callback-refs"


content/docs/reference-react-dom.md
```diff
@@ -50,7 +50,7 @@ If the optional callback is provided, it will be executed after the component is
 >
 > `ReactDOM.render()` currently returns a reference to the root `ReactComponent` instance. However, using this return value is legacy
 > and should be avoided because future versions of React may render components asynchronously in some cases. If you need a reference to the root `ReactComponent` instance, the preferred solution is to attach a
-> [callback ref](/docs/more-about-refs.html#the-ref-callback-attribute) to the root element.
+> [callback ref](/docs/more-about-refs.html#callback-refs) to the root element.
```
eps1lon, line 53: The fragment is fixed but it's still not working (probably because of the redirect). 
```suggestion
> [callback ref](/docs/refs-and-the-dom.html#callback-refs) to the root element.
```

lex111:
@tusba @eps1lon thanks!

PR #3813, opened 2021-07-26
Update nav.yml

ShouryaSrivastava01:
removed isOrdered from Main Concepts and Hooks in side navigation bar for uniformity in the side bar
ordering is present:
image

ordering isn't present:
image

### DIFF
--- a/content/docs/nav.yml
+++ b/content/docs/nav.yml
@@ -11,7 +11,6 @@
     - id: release-channels
       title: Release Channels
 - title: Main Concepts
-  isOrdered: true
   items:
     - id: hello-world
       title: Hello World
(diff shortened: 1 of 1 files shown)


content/docs/nav.yml
```diff
@@ -11,7 +11,6 @@
     - id: release-channels
       title: Release Channels
 - title: Main Concepts
-  isOrdered: true
```
harish-sethuraman, line 14: This should be intentional since the documentation is sequential and refers to older pages in the next page.

BartoszKlonowski:
Hello @ShouryaSrivastava01,
with how the nav is done in the react.dev now, I don't see this PR any longer required.
Let me close this as outdated, but you can create new PRs for other open issues if you would still like to contribute.

PR #3816, opened 2021-07-28
"the browsers" -> "browsers"

Haroenv:
We are now talking about "the browsers", but I think just "browsers", "most browsers" or "web browsers" make more sense, what do you think?


eps1lon (approved):
Makes sense, thanks

Haroenv:
Thanks for looking at my suggestion @eps1lon !

PR #3823, opened 2021-08-01
Add explanation of step and move variables

mangkoran:
I (and [some people](https://stackoverflow.com/questions/54016624/react-tutorial-history-map-step-move) to be honest) found that `step` and `move` variable in `history.map` method is ambiguous and need to be explained more. I add brief explanation for these variables to avoid confusion. Let me know if it needs changes.


mangkoran:
I updated this to better explain the `step` and `move` variable. Hope this helps newcomers to better understands the tutorial!

eps1lon (approved):
Yeah that probably helps if you're not familiar with `Array.prototype.map`. Thanks!

mangkoran:
No worries! Glad to contribute to the community!

PR #3828, opened 2021-08-03
Update recommended React toolchains, add Vite.

leerob:
This PR was sparked by this comment in the CRA repo: https://github.com/facebook/create-react-app/issues/11180#issuecomment-874748552

> Starting with 2.0, it’s [CRA] mostly in maintenance mode and does not strive to be the best tool for production React apps. It is a tool to get started and get something running fast. Perhaps, it’s not even best at that anymore.

I've updated the docs to reflect this. I've also tried to remove words like "popular" (describing Next.js) and "best" (describing Gatsby) to be more unopinionated. I tried to be as unbiased as possible, so please suggest any changes that could make this more clear.


content/docs/create-a-new-react-app.md
```diff
@@ -30,43 +30,29 @@ This is also **the easiest way to integrate React into an existing website.** Yo
...
+- If you're building a **[single-page app](/docs/glossary.html#single-page-application)**, use [Create React App](#create-react-app).
 - If you're building a **component library** or **integrating with an existing codebase**, try [More Flexible Toolchains](#more-flexible-toolchains).
 
-### Create React App {#create-react-app}
-
-[Create React App](https://github.com/facebookincubator/create-react-app) is a comfortable environment for **learning React**, and is the best way to start building **a new [single-page](/docs/glossary.html#single-page-application) application** in React.
-
-It sets up your development environment so that you can use the latest JavaScript features, provides a nice developer experience, and optimizes your app for production. You’ll need to have [Node >= 10.16 and npm >= 5.6](https://nodejs.org/en/) on your machine. To create a project, run:
-
-```bash
```
leerob, line 44: I removed this in favor of linking out to the CRA site, similar to the other frameworks.

content/docs/create-a-new-react-app.md
```diff
@@ -30,43 +30,29 @@ This is also **the easiest way to integrate React into an existing website.** Yo
...
 ### Gatsby {#gatsby}
 
-[Gatsby](https://www.gatsbyjs.org/) is the best way to create **static websites** with React. It lets you use React components, but outputs pre-rendered HTML and CSS to guarantee the fastest load time.
+[Gatsby](https://www.gatsbyjs.org/) is a framework for creating **static websites** with React. It lets you use React components, but outputs pre-rendered HTML and CSS to guarantee the fastest load time.
 
 Learn Gatsby from [its official guide](https://www.gatsbyjs.org/docs/) and a [gallery of starter kits](https://www.gatsbyjs.org/docs/gatsby-starters/).
 
+### Create React App {#create-react-app}
+
+[Create React App](https://github.com/facebookincubator/create-react-app) is framework for **small, [single-page](/docs/glossary.html#single-page-application) applications** in React.
```
leerob, line 52: From https://github.com/facebook/create-react-app/issues/11180#issuecomment-874748552

> In the current client-only design, you could make an argument that CRA does not have the right defaults for the web ecosystem.

Should this say "client-only" instead of single-page?

content/docs/create-a-new-react-app.md
```diff
@@ -79,6 +65,8 @@ The following toolchains offer more flexibility and choice. We recommend them to
 
 - **[Razzle](https://github.com/jaredpalmer/razzle)** is a server-rendering framework that doesn't require any configuration, but offers more flexibility than Next.js.
 
+- **[Vite](https://vitejs.dev/)** is a build tool that aims to provide a faster and leaner development experience for modern web projects.
```
leerob, line 68: Description taken from their docs 👍

styfle (approved):

3imed-jaberi (approved):
LGTM +1

rachelnabors:
Thanks for your contribution—it's very close to the new page for the new site! We were considering updating the current content, but now we will have to consider just merging your contribution!

leerob:
@rachelnabors Anything else you need from me here?

content/docs/create-a-new-react-app.md
44px, line 52: I suggest to keep old description here, but remove "comfortable" and best". Something like this:
> [Create React App](https://github.com/facebookincubator/create-react-app) is a pre-configured environment which lets you start building [single-page](/docs/glossary.html#single-page-application) application or just learning React without configuring build tools like webpack or Babel.

My main concerns with proposed version are:

> is framework

It puts CRA on the same shelf with Next.js and Gatsby, but CRA provides almost no guidelines on app structure. Maybe it can serve as a difference here instead: if you want a framework, use Next.js or Gatsby.

> for small applications

I'd say size should not be differentiating point here. CRA can be a suitable choice for medium-sized or even large apps unless you need to configure something in underlying tools.

Seems like "small" here means "simple" or even "quite typical single-page app with predictable build tooling requirements" :)

styfle: @44px Would you prefer changing "small" to "non-production" to match [Dan's comment](https://github.com/facebook/create-react-app/issues/11180#issuecomment-874748552)?

> ...starting with 2.0, [CRA is] mostly in maintenance mode and does not strive to be the best tool for production React apps. It is a tool to get started and get something running fast. Perhaps, it's not even best at that anymore.

44px: I'm afraid it can nudge someone to build app with custom webpack config because they have "production" app :)

For the last year I saw at least two apps with config problems:
- One had broken caching because of improperly set chunk name
- Second had large production bundle because of inlined source maps

Both problems have simple fixes, but they might not happen at all if these apps were built with CRA. And they could, these were "simple" apps.

I still see the CRA as the only option if you need to build "classic SPA". It provides good DX (TypeScript support, Fast Refresh, babel-macro, using SVGs as React components, detection of CSS modules, REACT_APP_* env vars substitution) and produces optimized production bundles.

styfle: Should we add the "without configuring build tools like webpack or Babel" to the Next.js and Gastby descriptions too?

44px: Good point! We don't need to say this at all because all tools from "Recommended Toolchains" section are zero-config.

nickmccurdy:
I would like to see Vite recommended alongside `create-react-app`. It has a lot of overlap with CRA's use case so it could be a useful alternative (mainly because it's faster and has plugins), but CRA is still useful for more advanced use cases and older apps, so IMO we should give users both options.

leerob:
Hey @rachelnabors! Wanted to check in here. Anything more you need from me?

harish-sethuraman:
Can you rebase your changes please?

leerob:
@harish-sethuraman done, thank you!

content/docs/create-a-new-react-app.md
```diff
@@ -30,43 +30,29 @@ This is also **the easiest way to integrate React into an existing website.** Yo
 
 The React team primarily recommends these solutions:
 
-- If you're **learning React** or **creating a new [single-page](/docs/glossary.html#single-page-application) app,** use [Create React App](#create-react-app).
-- If you're building a **server-rendered website with Node.js,** try [Next.js](#nextjs).
-- If you're building a **static content-oriented website,** try [Gatsby](#gatsby).
+- If you're building a **static or server-rendered application,** use [Next.js](#nextjs).
+- If you're building a **static content-oriented website,** use [Gatsby](#gatsby).
+- If you're building a **[single-page app](/docs/glossary.html#single-page-application)**, use [Create React App](#create-react-app).
```
harish-sethuraman, line 35: In the [beta.reactjs.org](https://beta.reactjs.org/learn/start-a-new-react-project#getting-started-with-a-react-toolchain) we still recommend `create-react-app` if they are just getting started. Not sure how @gaearon feels about this? We might want to make it consistent in both sites.

gaearon:
How do you feel about the messaging on the new site? I think it's confusing that we have the newer messaging there, and now we have an intermediate version. Would this PR still be valuable after the beta website ships?

leerob:
From the new messaging:

> If you’re just getting started with React, we recommend [Create React App](https://create-react-app.dev/), the most popular way to try out React’s features and a great way to build a new single-page, client-side application. Create React App is an unopinionated toolchain configured just for React.

My only confusion is that this seems to conflict with [this comment](https://github.com/facebook/create-react-app/issues/11180#issuecomment-874748552).

> Starting with 2.0, it’s [CRA] mostly in maintenance mode and does not strive to be the best tool for production React apps. It is a tool to get started and get something running fast. Perhaps, it’s not even best at that anymore.

nickmccurdy (approved):

nickmccurdy:
@leerob Hi, was closing this intentional?

leerob:
Oh whoops, yeah I forgot that would happen. I was cleaning up my mess of forked repos.

I'll defer to Dan – happy to roll this forward to the news docs site.

PR #3831, opened 2021-08-04
Use console.log in tutorial instead of alert

bev-a-tron:
Chrome is deprecating alerts, so use console.log in the tutorial instead. Update text to match as well.

See discussion here: https://github.com/reactjs/reactjs.org/pull/3827

Updates to the CodePen with "full code" at various points of the tutorial are not needed because none of them include the alert().

cc @gaearon @sag1v


gaearon:
Thanks!

PR #3837, opened 2021-08-06
Change instances of "a pseudocode" to just "pseudocode"

zkldi:
I think (and I might be wrong?) that "a pseudocode" is grammatically incorrect, and should be just "pseudocode". I've also removed the hyphen from the second usage of the term, since that seems to be incorrect aswell.


zkldi:
Note: I'm closing this PR as it seems that the react documentation is being completely rewrote, and therefore this change no longer applies.

PR #3839, opened 2021-08-07
Fix code snippet typo in tutorial's lifting the state section

QualBean:
Removed parenthesis from `onClick` property in the square component button passed from the board component.

### DIFF
--- a/content/tutorial/tutorial.md
+++ b/content/tutorial/tutorial.md
@@ -429,7 +429,7 @@ Instead, we'll pass down a function from the Board to the Square, and we'll have
 Now we're passing down two props from Board to Square: `value` and `onClick`. The `onClick` prop is a function that Square can call when clicked. We'll make the following changes to Square:
 
 * Replace `this.state.value` with `this.props.value` in Square's `render` method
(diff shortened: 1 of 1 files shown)


content/tutorial/tutorial.md
```diff
@@ -440,7 +440,7 @@ class Square extends React.Component {
     return (
       <button
         className="square"
-        onClick={() => this.props.onClick()}
+        onClick={() => this.props.onClick}
```
eps1lon, line 443: This wouldn't call the `onClick` prop anymore when a user clicks on the `button`.

Does this clear things up?

eps1lon (changes requested):

content/tutorial/tutorial.md
QualBean: I closed the issue. Actually, I realized I had no callback function. It was actually confusing because I was passing down props from the board to the square so automatically I thought we'll access the function directly without a callback. Found it a couple of chapters later in the tutorial and decided to close the issue.
Thanks anyway @eps1lon

PR #3857, opened 2021-08-19
Else seems to be necessary

Nashev:


alexkrolick:
It's not necessary because the if-statement has an early return

PR #3859, opened 2021-08-20
in React.createElement() props go inside an object, not an array.

fseitun:


harish-sethuraman:
Square brackets were used for showing that the param was optional. We are changing this representation in the new site. So closing this PR. 

Thank you.

PR #3860, opened 2021-08-20
Update reference-javascript-environment-requirements.md

necto68:
React also needs `Symbol` type. So if you want to support older browsers, it will be great to include it as polyfill.


eps1lon:
> React also needs Symbol type.

I don't think it needs. It has a fallback specifically for missing `Symbol.for`. Do you have a repro where environments without `Symbol` crash?

necto68:
@eps1lon seems like you are right. So we don't need a fallback for `Symbol.for` as React has already had it.

PR #3864, opened 2021-08-23
Refactor for latest update - hook-rules.md update: "plugin:react-hooks/recommended" now preferred

cforsyth68:
May 26, 2020, React's documentation was updated to change how this is declared. See blame history here...

https://github.com/facebook/react/blame/main/packages/eslint-plugin-react-hooks/README.md


cforsyth68:
Can we merge?

harish-sethuraman:
There are two ways of using the `react-hooks` plugin. You can either make use of the recommended config of the react-hooks plugin like shown below by extending it.
```js
{
  "extends": [
    // ...
    "plugin:react-hooks/recommended"
  ]
}
```
Or you can  add `react-hooks` to the plugins and then configuring the rules 
```js
{
  "plugins": [
    // ...
    "react-hooks"
  ],
  "rules": {
    // ...
    "react-hooks/rules-of-hooks": "error",
    "react-hooks/exhaustive-deps": "warn"
  }
}
```


But I don't think you can actually add `plugin:react-hooks/recommended` to the plugins as per your code change.

cforsyth68:
> 

Then why does the ESLint documentation literally have that exact text?

https://github.com/facebook/react/blame/main/packages/eslint-plugin-react-hooks/README.md#L21-L30

content/docs/hooks-rules.md
```diff
@@ -38,7 +38,7 @@ npm install eslint-plugin-react-hooks --save-dev
 {
   "plugins": [
```
harish-sethuraman, line 39: The [URL](https://github.com/facebook/react/blame/main/packages/eslint-plugin-react-hooks/README.md#L21-L30 ) you sent has  `"extends"` but here you have modified in the `"plugins"` right ? 

Snippet in the URL:
```js
{
  "extends": [
    // ...
    "plugin:react-hooks/recommended"
  ]
}
```

cforsyth68: Good catch, thanks.

cforsyth68:
This solution was what worked for me, however I haven't had any time to respond to commit comments. I do not have time to prove anything either, so I'm just letting this go. I was only trying to help.

PR #3867, opened 2021-08-24
Add React Native EU to upcoming conferences

thymikee:
Hey! 

React Native EU is the biggest RN conference in Europe and it was already featured in this section in the past. We're launching another edition in a week (1-2 Sept), so I thought it would be cool to have it on the listing again :)

cc @hramos


lex111 (approved):

lex111:
@thymikee sorry for the delay, moved it to the list of past conferences.

PR #3883, opened 2021-08-31
docs: Improve "Thinking in React" for color-blind people

AElmoznino:
## What this PR does:
Updates an image and its alt-text in the "Thinking in React" section of the docs to make it accessible for people with color blindness. Kind of hard for color blind to see those colors - let's lead by example ☺️

**Before:**
before

**After:**
after


alexkrolick (approved):

alexkrolick:
lgtm - @eps1lon ?

eps1lon (approved):
I love everything about this 👍🏻

eps1lon:
Bonus points for anybody that converts this to actual HTML for even better a11y

AElmoznino:
@eps1lon Do you by chance happen to know other sections of the docs that need a bit of a11y care? 🙂

eps1lon:
> @eps1lon Do you by chance happen to know other sections of the docs that need a bit of a11y care?

I've never really looked into it for reactjs.org to be honest. Since a11y is a continuous process it's probably best to re-evaluate it occasionally. 

Though the docs are currently being rewritten so I would suggest you wait for https://github.com/reactjs/reactjs.org/issues/3308 to be completed.

PR #3899, opened 2021-09-09
Update hooks-reference.md

andzhr:
It seems that call useСallback is not equivalent to call useMemo. The first returns a function, and the second returns a value.


content/docs/hooks-reference.md
```diff
@@ -365,7 +365,7 @@ Returns a [memoized](https://en.wikipedia.org/wiki/Memoization) callback.
 
 Pass an inline callback and an array of dependencies. `useCallback` will return a memoized version of the callback that only changes if one of the dependencies has changed. This is useful when passing callbacks to optimized child components that rely on reference equality to prevent unnecessary renders (e.g. `shouldComponentUpdate`).
 
-`useCallback(fn, deps)` is equivalent to `useMemo(() => fn, deps)`.
+`useCallback(fn, deps)` is equivalent to `useCallback(() => fn(), deps)`.
```
cl-ansy, line 368: Why not remove the line if it's inaccurate? This change is even more confusing IMO

andzhr: I agree

PR #3900, opened 2021-09-09
Update node version of create-react-app

wasdJens:
Hi, 

according to the latest create-react-app documentations you now require node >= 14.0.0 in order to use create-react-app. 
(See https://github.com/facebook/create-react-app#creating-an-app).

This pull request simply updates the node version information inside the `Create React App` docs.


eps1lon (approved):
Thanks!

PR #3901, opened 2021-09-09
fix website to translations.reactjs.org

upupming:
It seems the old website is down: https://www.isitdownrightnow.com/isreacttranslatedyet.com.html

So I changed to the latest website, hope it helps.


gaearon:
thanks!

PR #3904, opened 2021-09-13
missing inputElement

kokoyroy:
there was missing the inputElement = null so i added it


kokoyroy (commented):
added a this and there was no need for that...

PR #3907, opened 2021-09-14
docs: remove cancelled conference

debs-obrien:
https://reactlive.nl/ has been cancelled completely so removing from the conference page.


gaearon:
thanks!

PR #3910, opened 2021-09-16
docs: add clarification on setState merging

Birdi7:
While I am learning the react from scratch, this point was not easy to understand. The provided statement provides a bit of clarification

### DIFF
--- a/content/docs/state-and-lifecycle.md
+++ b/content/docs/state-and-lifecycle.md
@@ -408,6 +408,7 @@ Then you can update them independently with separate `setState()` calls:
 ```
 
 The merging is shallow, so `this.setState({comments})` leaves `this.state.posts` intact, but completely replaces `this.state.comments`.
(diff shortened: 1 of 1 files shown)


BartoszKlonowski:
Hello @Birdi7!
I'm aware this is very old issue and comment now, but let me respond so it stays for those who will search for some info and encounter this.

So, the first sentence in the original part actually says exactly what you added:

> `this.setState({comments})` leaves `this.state.posts` intact, but completely replaces `this.state.comments`.

☝️ in other words, since dev added `comments` to the changing state param, only the `comments` will be replaced (shallowly merged) in the current `this.state`.

Also, referring to the new react.dev docs, the shallow merging is also described in the [setState(nextState, callback?)](https://react.dev/reference/react/Component#setstate-parameters)
> it will be shallowly merged into this.state.

If there are more questions, or doubts please let me know or create new issue if there's a space for improvements.

Thank you!

PR #3912, opened 2021-09-18
[Docs] Apply the code Markdown syntax for the value prop

SiarheiBobryk:
From my perspective here the code styling should be used, since "value" is used as the React attribute.


eps1lon (approved):
Thanks!

PR #3914, opened 2021-09-19
[Docs] Add the anchor for the first entry about "HOC"

SiarheiBobryk:
I'm reading the docs from cover to cover and I haven't seen any mentions about "HOC" before, so here I'd suggest navigating readers to the appropriate article.


eps1lon (approved):
Makes sense, thanks!

PR #3919, opened 2021-09-21
Fixed React syntax's in docs.

akgupta0777:
https://reactjs.org/docs/react-dom.html#render
https://reactjs.org/docs/react-dom.html#hydrate
https://reactjs.org/docs/react-component.html#setstate
has wrong syntaxes.
This PR fixes it.

### DIFF
--- a/content/docs/reference-react-component.md
+++ b/content/docs/reference-react-component.md
@@ -539,7 +539,7 @@ The second parameter to `setState()` is an optional callback function that will 
 You may optionally pass an object as the first argument to `setState()` instead of a function:
 
 ```javascript
-setState(stateChange[, callback])
+setState(stateChange, [callback])
 ```
 
 This performs a shallow merge of `stateChange` into the new state, e.g., to adjust a shopping cart item quantity:
--- a/content/docs/reference-react-dom.md
+++ b/content/docs/reference-react-dom.md
@@ -33,7 +33,7 @@ React supports all popular browsers, including Internet Explorer 9 and above, al
 ### `render()` {#render}
 
 ```javascript
-ReactDOM.render(element, container[, callback])
+ReactDOM.render(element, container, [callback])
 ```
 
 Render a React element into the DOM in the supplied `container` and return a [reference](/docs/more-about-refs.html) to the component (or returns `null` for [stateless components](/docs/components-and-props.html#function-and-class-components)).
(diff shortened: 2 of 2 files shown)


akgupta0777:
Please review @eps1lon @bvaughn.

akgupta0777:
image

Is this intentional because I found them a lot in documentation.

bvaughn:
I _think_ the former style _is_ intentional, to show that the 3rd argument itself (including the comma) is optional, but I agree it's confusing because it's not valid syntax. (Then again, even though the `[]` would be valid syntax, it's not a valid use either.)


I think maybe a better way for us to do this might be to follow [MDN's style](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#syntax), e.g.
> # Syntax
> ```js
> target.addEventListener(type, listener);
> target.addEventListener(type, listener, options);
> target.addEventListener(type, listener, useCapture);
> target.addEventListener(type, listener, useCapture, wantsUntrusted); // wantsUntrusted is Firefox only
> ```

akgupta0777:
Using MDN's style is a lot of work. I think we can use a **_(underscore)** to represent those parameters that are optional.
For Example :
```
ReactDOM.render(element, container, _callback)
setState(stateChange, _callback)
ReactDOM.hydrate(element, container, _callback)
```
What do you think?

bvaughn:
I think brackets are more intuitive (based on other API documentation I've seen). underscore means nothing to me.

akgupta0777:
So should we use MDN style or leave it as it is?

bvaughn:
No strong feelings. Maybe someone else has those.

eps1lon:
I think we already had one or two PRs for this syntax. It's clear by now that it's confusing to a considerable amount because it almost looks like valid JS. 

The MDN style has the benefit that it uses valid JS even though we don't actually have function overloading. But I find MDN API docs extraordinary good and value consistent styles of API documentation. So I'm all for changing it to the MDN style.

But we should check every signature first and change all of them where necessary. And it's unclear if that's needed once https://github.com/reactjs/reactjs.org/issues/3308 lands so it might be work that's short lived.

akgupta0777:
Thanks for checking this up. I think the best thing to do is to wait for the new reactjs site.

harish-sethuraman:
We are changing the format in the new site. Thank you 😄 .

PR #3921, opened 2021-09-23
Update jsx-in-depth.md

bukulele:
The example in section "Functions as Children" works only with "0" index in props.children, because props.children is an array-like object, where the function is the value of the first property.


eps1lon:
> The example in section "Functions as Children" works only with "0" index in props.children, because props.children is an array-like object, where the function is the value of the first property.

Not in this specific example. Are you maybe trying this example in `preact`?

PR #3922, opened 2021-09-23
Update tutorial.md

pphysch:
Clarified ambiguous use of `this`.


eps1lon (approved):
Sounds like a more useful description since the prior one just repeated literally what the code does.

PR #3930, opened 2021-09-28
Fix 404 page for "/link/switch-to-createroot"

ajaxlab:
Hello, I was testing react 18 alpha and I found a guide for using new api.

image

But the link https://reactjs.org/link/switch-to-createroot seems to be empty.

image

I guess the vercel config is missing the redirection url.
So I added the link to `vercel.json`.

Thanks.


eps1lon:
Not sure what happened between https://github.com/facebook/react/pull/21652 and now. @acdlite said it already redirects but maybe this broken in between?

PR #3936, opened 2021-09-30
Updated docs link.

playstore777:
The older link no longer exists, as it is now docs/getting-started.md instead of just docs.


harish-sethuraman (commented):
Looks like the link is now fixed. Thanks for the PR

PR #3938, opened 2021-10-01
enhanced README.md for better accesibility and user inclusiveness

Dtech-Dbug:
- just felt that the document could have been a little more acessible - by adding a few missing dynamic links

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,10 +7,10 @@ This repo contains the source code and documentation powering [reactjs.org](http
 ### Prerequisites
 
 1. Git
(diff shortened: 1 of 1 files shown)


BartoszKlonowski:
Hello @Dtech-Dbug!
I feel like this PR is not really required.
The main reason to avoid having tools linked is that the versions change, tools change, websites change and links get outdated. So to avoid confusion and maintenance effort we try to limit the number of links used.
If dev/reader wants to know more on how or where to get certain tool search can be used.
Let me close this PR as not required, but please create new PRs for other open issues if you would still like to contribute.
Thank you!

PR #3942, opened 2021-10-02
[Docs] Apply the code inline styles for "render"

SiarheiBobryk:


eps1lon:
Thanks for the suggestion.

"render method" not only refers to `render` of class components but also function components. So using the code formatting is not covering all cases.

SiarheiBobryk:
Got it. Sorry. I've missed this part.

PR #3946, opened 2021-10-04
Removing a confusion between mutating and owning the state

nirmalpatel:
"Next, we need to identify which component mutates, or owns, this state." In this guide, there are two different components. The search bar triggers mutation, and main component owns the state.


eps1lon (approved):
"trigger changes" over "mutates" is nice 👍🏻 

Please sign the CLA so that we can merge

PR #3948, opened 2021-10-05
Restyle logo width to make translations easier [issue #1817]

### LINKED ISSUE #1817: Restyle logo width to make translations easier

nigiiro-taro:
### About:
This PR fixes #1817.
Currently, when translated into Vietnamese, the nav link has two lines with a screen width between 1280px and 1290px.
To solve this problem,  I fix two points
- Unfix the width of the logo at calc(100% / 6)
- Centers the nav bar between the logo and search bar

This also helps to solve the problem of cut-offs in translation without hard coding of the logo width.

### Visuals:
**Before Changes:**
スクリーンショット 2021-10-05 18 11 12

**After Changes:**
スクリーンショット 2021-10-05 18 11 52

### DIFF
--- a/src/components/LayoutHeader/Header.js
+++ b/src/components/LayoutHeader/Header.js
@@ -77,9 +77,6 @@ const Header = ({location}: {location: Location}) => (
               color: colors.white,
             },
 
-            [media.greaterThan('small')]: {
-              width: 'calc(100% / 6)',
-            },
             [media.lessThan('small')]: {
               flex: '0 0 auto',
             },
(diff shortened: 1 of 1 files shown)


akgupta0777 (approved):
I saw the deployment works fine and looks good. Thanks 👍

PR #3951, opened 2021-10-07
Delete _redirects

gaearon:
This file isn't used by our current hosting provider.

**However** we can't actually delete it because this still uses it: https://github.com/reactjs/reactjs.org/blob/a4d4893bd3c56f4f7f821c46ed512eac6c3c21b9/plugins/gatsby-transformer-versions-yaml/gatsby-node.js. Which means our version redirects might be borked now (?)


harish-sethuraman:
I see that now we have a redirect file for our current hosting provider. https://github.com/reactjs/reactjs.org/blob/main/vercel.json

We might consider removing the redirects file and the plugin that generates it as well?

gaearon:
The problem is (I think?) that the version-based redirects are broken on the existing site. Can you fix that first?

harish-sethuraman:
/version/16.8  https://5d4b5feba32acd0008d0df98--reactjs.netlify.com/  302
/version/16.7  https://5c54aa429e16c80007af3cd2--reactjs.netlify.com/  302
/version/16.6  https://5c11762d4be4d10008916ab1--reactjs.netlify.com/  302
/version/16.5  https://5bcf5863c6aed64970d6de5b--reactjs.netlify.com/  302
/version/16.4  https://5b90c17ac9659241e7f4c938--reactjs.netlify.com  302
/version/16.3  https://5b05c94e0733d530fd1fafe0--reactjs.netlify.com  302
/version/16.2  https://5abc31d8be40f1556f06c4be--reactjs.netlify.com  302
/version/16.1  https://5a1dbcf14c4b93299e65b9a9--reactjs.netlify.com  302
/version/16.0  https://5a046bf5a6188f4b8fa4938a--reactjs.netlify.com  302
/version/15.6  https://react-legacy.netlify.com  302
Tried opening these links and none of the links seem to work except for the last one.

gaearon:
Yes, and the reason they won't work is because the existing script I mentioned writes to _redirects instead of vercel config. we need to fix it to write to the vercel config.

harish-sethuraman:
But I don't get it. Seems like we read the netlify URLs from https://github.com/reactjs/reactjs.org/blob/main/content/versions.yml
But these links do not work at all. These links are then added to `_redirects` in turn

gaearon:
Ohhh I see. You're saying that the _links themselves_ don't work. So it's a separate issue.

Let's say imagine they work. Can we at least get it writing to the correct file?

And then we'll think of how to fix the links.

harish-sethuraman:
Do I update the existing gatsby plugin and add script for the beta site too?  Or is it enough if i write it for the beta site?

harish-sethuraman:
We might want to merge this as well?

harish-sethuraman (approved):

harish-sethuraman:
We might want to get working links for older react versions: https://github.com/reactjs/reactjs.org/issues/4306

thiagodebastos:
Is there any other way to get access to old documentation other than the broken netlify instances?

harish-sethuraman:
> Is there any other way to get access to old documentation other than the broken netlify instances? 

Nope, AFAIK there is no other way until we find the right URL. (Or maybe we might want to revisit old commits in the repo and try local build for that version :( )

thiagodebastos:
Ah dang! Until then, I suppose users like me can rely on a web archive such as good old [Way Back Machine](http://web.archive.org/web/20180901000000*/reactjs.org) 😅

PR #3975, opened 2021-10-22
Style scrollbars in dark mode

TheOtterlord:
Chrome (and probably other webkit based browsers) doesn't automatically darken scrollbars.

Before:

image

After:

image


gaearon:
I was seeing it when switching between dark and light mode (on any page): the switched would jump because of the `width` style that only applied in one of the modes.

PR #4020, opened 2021-10-28
don't display shortcuts on mobile and use Ctrl on non-macOS-like devices

jackcmeyer:
- don't display shortcut keys on mobile devices (but retain shortcut keys for tablets)
- display "Ctrl" instead of ⌘ for non-macOS like devices
- use [react-device-detect](https://www.npmjs.com/package/react-device-detect) to detect devices 

macOS like devices
image

non-macOS like devices
image


beta/package.json
```diff
@@ -37,6 +37,7 @@
     "parse-numeric-range": "^1.2.0",
     "react": "18.0.0-alpha-930c9e7ee-20211015",
     "react-collapsed": "3.1.0",
+    "react-device-detect": "^2.1.2",
```
gaearon, line 40: This feels like an excessive dependency. What is the essence of the thing we're trying to check? What is the minimal way we can check it?

jackcmeyer: I've removed it and using a regex match against `navigator.userAgent` now

beta/src/components/Search.tsx
```diff
@@ -86,6 +90,13 @@ export const Search: React.FC<SearchProps> = ({
 
   useDocSearchKeyboardEvents({isOpen: isShowing, onOpen, onClose});
 
+  React.useEffect(() => {
+    if (typeof navigator !== 'undefined') {
+      setShowMacosShortcut(/(Mac|iPhone|iPod|iPad)/i.test(navigator.userAgent));
```
harish-sethuraman, line 95: Why are we testing for iPhone iPod? we do not show the shortcuts in those devices right ? 🤔 any specific reason ?

beta/src/components/Search.tsx
```diff
@@ -108,10 +119,12 @@ export const Search: React.FC<SearchProps> = ({
         onClick={onOpen}>
         <IconSearch className="mr-3 align-middle text-gray-30 flex-shrink-0 group-betterhover:hover:text-gray-70" />
         Search
-        <span className="ml-auto hidden sm:flex item-center">
-          <Kbd>⌘</Kbd>
-          <Kbd>K</Kbd>
-        </span>
+        {showKeyboardShortcuts && (
+          <span className="ml-auto hidden sm:flex item-center">
```
harish-sethuraman, line 123: By default it is hidden with CSS. `sm:flex` for devices other than small ones only we display the shortcuts. `showKeyboardShortcuts` we might not need this ?

beta/src/components/Search.tsx
jackcmeyer: When testing on my iPhone in landscape mode the shortcut key icons still display (happening on beta.reactjs.org and the preview deployment for this PR). 

If we don't test for iPhone/iPod, it would show the Ctrl key instead. Maybe a different breakpoint size is needed?

beta/src/components/Search.tsx
jackcmeyer: I'm using `showKeyboardShortcuts` as a way to prevent "flashing". On a Mac-like device with the current code, without this check, the shortcut key would first show "Ctrl" and then go to the command key after the browser has been checked (`showMacosShortcut` defaults to false). 

It is acting as conditional that represents if the user-agent has been checked or not.

beta/src/components/Search.tsx
harish-sethuraman: > When testing on my iPhone in landscape mode the shortcut key icons still display (happening on beta.reactjs.org and the preview deployment for this PR).
> 
> If we don't test for iPhone/iPod, it would show the Ctrl key instead. Maybe a different breakpoint size is needed?

For landscape mode we should't show command key or crtl rather can we add orientation in the tailwind and handle for smaller iphone and ipod's landscape mode?  `landscape: {raw: '(orientation: landscape)'}`

It should work I guess.
let me know your thoughts 😄

PR #4022, opened 2021-10-29
Take system's theme for the beta site

harish-sethuraman:
If the user has dark theme as preferred color scheme. apply dark mode.

It was a todo in the code 😄 
// TODO: respect system pref

### DIFF
--- a/beta/src/components/Layout/Nav/Nav.tsx
+++ b/beta/src/components/Layout/Nav/Nav.tsx
@@ -102,13 +102,10 @@ export default function Nav() {
   });
 
   React.useEffect(() => {
-    if (window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches) {
-      document.documentElement.classList.add('dark');
-      setIsDark(true);
-    } else {
-      setIsDark(false);
-    }
-  },[]);
+    window.__onThemeChange = () => {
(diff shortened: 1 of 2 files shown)


gaearon:
This won't work for the initial render. It will flash.

harish-sethuraman:
Shall I add an IIFE to the document then? Like you had done in over reacted site? https://github.com/gaearon/overreacted.io/blob/master/src/html.js#L21
and then preserve it in the local storage?

gaearon:
Seems like this doesn’t work for sandboxes?

harish-sethuraman:
Do you mean for the sandpack component in the site? 🤔

gaearon:
Oh I see. That's in https://github.com/reactjs/reactjs.org/pull/4019. We need to figure out what to get in first.

gaearon:
When I load a preview on this branch with initial Dark more, I see the toggle flash.

gaearon:
What's happening here is that the SSR content always assumes the theme is light. You should either hide the toggle or show the correct toggle icon with CSS.

harish-sethuraman:
Will work on it and push changes for toggle.

gaearon:
I think I'll take it over from here. Thanks!

gaearon:
thanks!

PR #4023, opened 2021-10-29
Improvement to dark-tweaks, presist dark mode, add transitions, etc

amitojsingh366:
Below mentioned things are added in this pr:
- persistent color mode (dark, light) 
- components using the new color mode system
- smooth transition between colorModes

this was achieved by creating a context provider using the [context api](https://reactjs.org/docs/context.html) called `ThemeProvider` it can be found at `beta/src/modules/ThemeProvider.tsx`


beta/src/components/Layout/Nav/MobileNav.tsx
```diff
@@ -4,9 +4,9 @@
 
 import * as React from 'react';
 import cn from 'classnames';
-import {RouteItem} from 'components/Layout/useRouteMeta';
-import {useRouter} from 'next/router';
-import {SidebarRouteTree} from '../Sidebar';
+import { RouteItem } from 'components/Layout/useRouteMeta';
```
gaearon, line 7: These changes make it harder to review what's actually changed in this PR. You can run `yarn prettier` in the `beta` folder to get rid of these.

gaearon:
- The formatting changes make it very difficult to see what's changed in the PR. When you send a PR, please exclude any formatting changes. We use Prettier. It seems like maybe you're also using Prettier but it doesn't respect the project settings? Not sure what happened here.
- I'm not sure I understands what value the transition brings. I don't think it's necessary.
- There is a significant flaw of this solution. That is — **the page flashes from white to dark on load**. That's not acceptable and is part of the reason why we haven't implemented this yet. I'll take another look but I think I'll need to implement this from scratch because we need to avoid this flaw.

beta/src/components/Layout/Nav/MobileNav.tsx
amitojsingh366: ok i ran the command

gaearon:
Hey! Appreciate the effort. We've had to do a bunch of things differently to make it work well with SSR. I already merged the corresponding PRs but thanks for trying!

PR #4024, opened 2021-10-30
Fix case-sensitive search term in example.

eastcoasting:
Converting both the name and search value to lowercase solves case sensitive searching.


gaearon:
thanks!

Issue #3977, opened 2021-10-22
[beta] Google translate breaks Sandboxes by translating their contents


rachelnabors:
When I use Google Translate to translate a page, it translates the source code and error messages in the sandboxes. Could we make it so the code in the sandbox and the code in the error messages are _not_ translated?

<img width="1024" alt="Screenshot 2021-10-22 at 15 14 15" src="https://user-images.githubusercontent.com/236306/138469529-a4a550bf-3f4f-492f-b8f8-059dc4e580fe.png">

gaearon:
Should be fixed.

PR #4034, opened 2021-11-03
[beta] fix(sandpack): do not translate code blocks

danilowoz:
Sandpack has introduced a new HTML attribute to avoid Google Translate translating the code editor and the preview layer, but still translating the action buttons. 

Solve #3977

Screenshot 2021-11-03 at 11 34 17
Screenshot 2021-11-03 at 11 34 25

### DIFF
--- a/beta/package.json
+++ b/beta/package.json
@@ -17,7 +17,7 @@
     "start": "next start"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


gaearon:
thanks!

PR #4036, opened 2021-11-03
[Beta] Add bundle analyzer

gaearon:
This adds `yarn analyze` command so that we can start looking at the bundle size and how to improve it.
If someone knows how to hook this up to GH actions to see a diff on each PR let me know!

### DIFF
--- a/beta/next.config.js
+++ b/beta/next.config.js
@@ -30,6 +30,18 @@ module.exports = {
     ];
   },
   webpack: (config, {dev, isServer, ...options}) => {
+    if (process.env.ANALYZE) {
+      const { BundleAnalyzerPlugin } = require('webpack-bundle-analyzer')
+      config.plugins.push(
+        new BundleAnalyzerPlugin({
+          analyzerMode: 'static',
+          reportFilename: options.isServer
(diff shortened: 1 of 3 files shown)


harish-sethuraman:
I came across this GH action that lists PR's impact on the bundle size. 
https://github.com/hashicorp/nextjs-bundle-analysis/  
shall we try this out? 

Or should we make use yarn analyze during PR builds and make use of the html inside the analyze directory generated in the build?

gaearon:
Hmm. I tried this but for some reason it ignores my `working-directory` setting and works from root.
https://github.com/reactjs/reactjs.org/runs/4126582048?check_suite_focus=true

harish-sethuraman:
Did you try running a cd beta and then commands ?

gaearon:
No, it looks like `cd` is isolated between steps.

harish-sethuraman:
I'll create a repo and raise the fix.

harish-sethuraman:
Nice work! BTW we should close and re open PR for the workflow to run right ?

gaearon:
you need to rebase if you have an older PR

PR #4037, opened 2021-11-04
Bump to Next 12

gaearon:
Upgrades to Next 12.

Tried enabling `concurrentFeatures: true`, but ran into https://github.com/vercel/next.js/issues/30714.


gaearon:
seems fine.

PR #4039, opened 2021-11-04
Add linefeed in the mdx and handle it

harish-sethuraman:
Problem:
Multiple commands are displayed in the same line with line feed missing. 
page: https://beta.reactjs.org/learn/start-a-new-react-project
Solution: 
- Add preserve whitespace to classes
- ~~We can use 2 trailing empty spaces or `<br/>` to add new line feed~~
- ~~when we have multiple lines `typeof children.props.children` is not string but an object. So use children.props.children directly when it is a string or an object (`typeof children.props.children === 'string'` )~~

Before:
Screenshot 2021-11-04 at 11 20 52 PM
After:
Screenshot 2021-11-04 at 11 21 04 PM

### DIFF
--- a/beta/src/components/MDX/TerminalBlock.tsx
+++ b/beta/src/components/MDX/TerminalBlock.tsx
@@ -28,7 +28,8 @@ function TerminalBlock({level = 'info', children}: TerminalBlockProps) {
   if (typeof children === 'string') {
     message = children;
   } else if (
-    React.isValidElement(children) 
+    React.isValidElement(children) &&
+    typeof children.props.children === 'string'
   ) {
     message = children.props.children;
   }
@@ -42,7 +43,7 @@ function TerminalBlock({level = 'info', children}: TerminalBlockProps) {
           <IconTerminal className="inline-flex mr-2 self-center" /> Terminal
         </div>
       </div>
-      <div className="px-8 pt-4 pb-6 text-primary-dark dark:text-primary-dark font-mono text-code">
+      <div className="px-8 pt-4 pb-6 text-primary-dark dark:text-primary-dark font-mono text-code whitespace-pre">
         <LevelText type={level} />
         {message}
       </div>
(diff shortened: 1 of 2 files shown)


gaearon:
>We can use 2 trailing empty spaces or <br/> to add new line feed

This doesn't make sense to me.

Can you make it work normally?

beta/src/components/MDX/TerminalBlock.tsx
```diff
@@ -28,8 +28,7 @@ function TerminalBlock({level = 'info', children}: TerminalBlockProps) {
   if (typeof children === 'string') {
     message = children;
   } else if (
-    React.isValidElement(children) &&
-    typeof children.props.children === 'string'
+    React.isValidElement(children)
```
gaearon, line 31: Do we understand why this transformation exists at all? What happens if you remove it altogether?

harish-sethuraman:
> > We can use 2 trailing empty spaces or  to add new line feed
> 
> This doesn't make sense to me.
> 
> Can you make it work normally?

we could make it work normally but we need to render the children on the whole instead of doing `children.props.children`. Since the children.props.children is considered as a string and not a react element new line is not added 🙁 . 

I was trying to look around to add empty spaces in markdowns and found this 

https://www.markdownguide.org/basic-syntax/

Screenshot 2021-11-05 at 12 13 02 AM

beta/src/components/MDX/TerminalBlock.tsx
harish-sethuraman: I hope the condition was added to render the strings directly and render the children if it has collection of elements together (Like the commands in two lines).

This check can be removed and we can render the children of terminalBlocks directly too. But when we render the children directly the texts are wrapped inside `<p>` and takes extra space like this.
Screenshot 2021-11-05 at 12 27 54 AM

gaearon:
How does this work inside triple back tick (code blocks)? Terminal blocks should be treated similarly.

gaearon:
Does setting `whiteSpace: 'pre'` in the style fix it?

harish-sethuraman:
Adding whitespace-pre fixed the issue 👀 . I've added that to the terminal block now. Thanks.

gaearon:
Thx!

PR #4043, opened 2021-11-05
Add git hook to generate headings ids

lex111:
Let's automate the generation of custom headings ids on per commit so we don't have to do it manually.


lex111:
Deployment of the beta site fails apparently because Vercel does not exposed `CI` env var, its value must be true for husky not to run.

gaearon:
Can you remind me why we need this? I'm forgetting why this system exists.

lex111:
In short: having stable headings ids help for translators, check out this issue for more details -> https://github.com/reactjs/reactjs.org/issues/1608

gaearon:
Oh ok.

>1) translating one heading would mean having to search in the entire docs and change multiple files simultaneously

This makes sense that it's annoying.

Can you fix CI pls?

lex111:
Without your help, I can't fix this. Need to set `CI` env var to `true` in the Vercel deployment settings for the beta site. In all likelihood, this env is not set now, otherwise `husky` would not have installed during the build step since already `is-ci` check is used. 

https://vercel.com/docs/concepts/projects/environment-variables#system-environment-variables

lex111:
Sure, just updated the PR branch. Have you added `CI` env var to Vercel deployment settings?

gaearon:
Hmm. Vercel project has this on:

Screenshot 2021-11-10 at 17 57 27

which supposedly includes CI:

https://vercel.com/docs/concepts/projects/environment-variables#system-environment-variables

Why does this not work?

lex111:
Sorry, that was my mistake in new npm script, just fixed it, and everything should be ok now.

gaearon:
Awesome. Thanks.

PR #4044, opened 2021-11-05
Hide navigation for one challenge and recipes

harish-sethuraman:
When we have only one challenge we can hide the navigation button ? 

Page with one challenge: https://beta.reactjs.org/learn/importing-and-exporting-components

Screenshot 2021-11-05 at 2 09 04 PM


gaearon:
For clarity, here is the "After" version.

Screenshot 2021-11-05 at 16 51 08

Not sure how I feel about this.

harish-sethuraman:
Was going through the site and found that the navigation was present even for 1 challenge and it was non interactive. Thought it would be nice if we didn't show the navigation rather than showing it and making it non interactive.
PS: we actually even show a next challenge button only when we have more than one challenge and current challenge is not the last challenge.

You can decide on whether it would be good or not to have this change.

PR #4045, opened 2021-11-05
[beta] sandpack: cmd+backspace to remove entire line

danilowoz:
@gaearon Sandpack originally has implemented the ability to remove the entire line in its editor, but for some reason, this behavior isn't working on the documentation. So I spent some time narrowing down what is blocking or preventing this feature from working there, but unfortunately, I couldn't find the original problem. So after a creating few reproduce repositories, it turned out that recreating the entire `yarn.lock` solved the problem. ¯\_(ツ)_/¯

Do you have any clue why did it happen?


danilowoz:
Plus, this also solved the bug reported [here](https://twitter.com/simonw/status/1451351038712188942), where the `return` button didn't work on Codemirror.

gaearon:
sounds good to me!

gaearon:
sounds good to me!

gaearon:
This broke decorations.

Screenshot 2021-11-05 at 16 28 46

PR #4047, opened 2021-11-05
Enable streaming SSR

gaearon:
This should enable the new streaming renderer, which will let use Suspense for code splitting. (https://github.com/reactwg/react-18/discussions/37)

https://nextjs.org/blog/next-12#server-side-streaming

Note this also opts us into using the Edge Runtime. (It's confusing that it's coupled in Next.js but the options will likely be decoupled later.)

Let's give it a spin! Might be nice if we find some SSR bugs with this.

### DIFF
--- a/beta/src/pages/_document.tsx
+++ b/beta/src/pages/_document.tsx
@@ -3,64 +3,60 @@
  */
 
 import * as React from 'react';
-import Document, {Html, Head, Main, NextScript} from 'next/document';
+import {Html, Head, Main, NextScript} from 'next/document';
 
-class MyDocument extends Document {
-  render() {
-    //  @todo specify language in HTML?
-    return (
-      <Html lang="en">
-        <Head />
-        <body className="font-sans antialiased text-lg bg-wash dark:bg-wash-dark text-secondary dark:text-secondary-dark leading-base">
-          <script
-            dangerouslySetInnerHTML={{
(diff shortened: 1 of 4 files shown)


leerob:
> Note this also opts us into using the Edge Runtime. (It's confusing that it's coupled in Next.js but the options will likely be decoupled later.)

Indeed, this is being renamed to the Web Runtime 👍 IMO it's more clear, since it's built on Web APIs, and works without being deployed at the Edge. Further, Next.js ships with a sandbox allowing you to run the Web runtime locally (e.g. `next dev`) which is completely unrelated to the Edge.

beta/src/pages/_document.tsx
```diff
@@ -3,64 +3,60 @@
...
+                    document.documentElement.classList.add('dark');
+                  } else if (newTheme === 'light') {
+                    document.documentElement.classList.remove('dark');
                   }
+                }
+
+                var preferredTheme;
+                try {
+                  preferredTheme = localStorage.getItem('theme');
+                } catch (err) { }
```
Parvezkhan0, line 30: Just spotted duplicate variable declarations in _document.tsx around lines 25–30 (two var preferredTheme; and two var initialTheme = preferredTheme;). Cleaning those up would tidy the code. Otherwise, LGTM!

PR #4049, opened 2021-11-05
Show sidebar scroll bar only when necessary

lex111:
Currently, the sidebar scroll bar is always displayed, even when it is not needed (this applies to Windows/Linux)

image

### DIFF
--- a/beta/src/components/Layout/Page.tsx
+++ b/beta/src/components/Layout/Page.tsx
@@ -18,7 +18,7 @@ export function Page({routeTree, children}: PageProps) {
     <MenuProvider>
       <SidebarContext.Provider value={routeTree}>
         <div className="h-auto lg:h-screen flex flex-row">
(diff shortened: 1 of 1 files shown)


gaearon:
I'm looking into this too... This solution isn't amazing because it can change the content width when you collapse/expand.

lex111:
Agreed, it's not very good but having a scrollbar always visible is not an option. Perhaps in the future this can be fixed via the [scrollbar-gutter](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-gutter) property. So for now it seems like a reasonable compromise to show the sidebar scrollbar only when necessary.

gaearon:
Can you please test my alternative? https://github.com/reactjs/reactjs.org/pull/4052

lex111:
Checked it out, it's pretty good for me, so I'm closing this PR.

PR #4052, opened 2021-11-05
Only show sidebar scrollbars when needed

gaearon:
Alternative to https://github.com/reactjs/reactjs.org/pull/4049.

Basically, this will hide the sidebar scrollbar _background_ but the actual scrollbar will still show up when needed. This way, expanding/collapsing sections won't jump width (which is an issue in https://github.com/reactjs/reactjs.org/pull/4049). To make background transparent, we need to style them, which disables macOS "hide the thumb unless scrolling" thing. This is annoying but we can more or less approximate the native behavior by only showing the sidebar trumb when the container is hovered or has focus within.

Note that unlike https://github.com/reactjs/reactjs.org/pull/4049, this is supposed to also make them look decent when they _are_ shown (i.e. when the sidebar content doesn’t fit). Especially in dark mode where the default was too jarring. 

Haven't tested this on Windows. Would appreciate someone having a look.

### DIFF
--- a/beta/src/components/Layout/Page.tsx
+++ b/beta/src/components/Layout/Page.tsx
@@ -18,7 +18,7 @@ export function Page({routeTree, children}: PageProps) {
     <MenuProvider>
       <SidebarContext.Provider value={routeTree}>
         <div className="h-auto lg:h-screen flex flex-row">
-          <div className="h-auto lg:h-full overflow-y-scroll fixed flex flex-row lg:flex-col py-0 top-0 left-0 right-0 lg:max-w-xs w-full shadow lg:shadow-none z-50">
+          <div className="no-bg-scrollbar h-auto lg:h-full lg:overflow-y-scroll fixed flex flex-row lg:flex-col py-0 top-0 left-0 right-0 lg:max-w-xs w-full shadow lg:shadow-none z-50">
             <Nav />
             <Sidebar />
           </div>
--- a/beta/src/styles/index.css
+++ b/beta/src/styles/index.css
@@ -105,6 +105,55 @@
     from { opacity: 0.5; }
     to   { opacity: 1; }
   }
+
+  /*
+   * Hopefully when scrollbar-color lands everywhere,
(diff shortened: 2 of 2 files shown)


lex111:
Can confirm, works well, the only thing I do not like is that in webkit-browsers scrollbar becomes visible only if you hover over the sidebar. This is quite unusual, and even a little confusing, so hopefully in the future this will be fixed.
 
And it's probably worth noting that scrollbars on the site after that change are different in styles, eg, sidebar vs code blocks (as shown below). Perhaps we should apply this custom scrollbar to all scrollable elements to get consistency in design?

| Linux (Chrome)   | Windows (Chrome)   |
| -------- | -------- |
| image | image |

Edge:

image

Btw, do you think it worth to reduce this padding a little bit? It seems to be slightly large right now:

image

gaearon:
>Can confirm, works well, the only thing I do not like is that in webkit-browsers scrollbar becomes visible only if you hover over the sidebar. This is quite unusual, and even a little confusing, so hopefully in the future this will be fixed.

Yeah.. Kind of a compromise since we can't hide completely and also don't want to always show them. I'm hoping this can go away as well.

>And it's probably worth noting that scrollbars on the site after that change are different in styles, eg, sidebar vs code blocks (as shown below). Perhaps we should apply this custom scrollbar to all scrollable elements to get consistency in design?

Maybe. Not sure. (I'll ask @lebo.) I'm hesitant to apply the "on hover" thing for everything because it is a bit unnatural. Doing it for the sidebar is a compromise. Maybe we can revisit this when `scrollbar-color` is widely supported and we can remove other hacks.

>Btw, do you think it worth to reduce this padding a little bit? It seems to be slightly large right now:

Hm yeah. I'll adjust.

gaearon:
>Hm yeah. I'll adjust.

I tried adjusting it and it's a bit complicated. We'd have to make all padding asymmetric and I'm worried I'll break a bunch of other things. I think I'll leave this as is for now, but we can iterate in small steps.

PR #4055, opened 2021-11-06
Adjust code line numbers background

lex111:
Currently because of the code line numbers background is transparent, it is difficult to read the long code line. And generally it  looks not good. 

Example page - https://beta.reactjs.org/learn/managing-state#sharing-state-between-components

| Before   | After    |
| -------- | -------- |
| image | image |

### DIFF
--- a/beta/src/components/MDX/Sandpack/Themes.tsx
+++ b/beta/src/components/MDX/Sandpack/Themes.tsx
@@ -10,7 +10,7 @@ export const CustomTheme = {
     defaultText: 'inherit',
     inactiveText: 'inherit',
     activeBackground: 'inherit',
-    defaultBackground: 'var(--sp-colors-bg-default)',
+    defaultBackground: 'inherit',
     inputBackground: 'inherit',
     accent: 'inherit',
     errorBackground: 'inherit',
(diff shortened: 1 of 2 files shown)


gaearon:
oh wow, I was sure I fixed it already...

beta/src/components/MDX/Sandpack/Themes.tsx
```diff
@@ -10,7 +10,7 @@ export const CustomTheme = {
     defaultText: 'inherit',
     inactiveText: 'inherit',
     activeBackground: 'inherit',
-    defaultBackground: 'inherit',
+    defaultBackground: 'var(--sp-colors-bg-default)',
```
gaearon, line 13: this doesn't look good to me that this one alone is different. can you override it in `sandpack.css` instead like the rest?

gaearon: the reason i did it this way is because the styles specified here unfortunately don't get applied on the SSR pass. so we need to apply them "earlier"

lex111: Done :heavy_check_mark:

PR #4056, opened 2021-11-06
Restore scroll after 'show less'

gaearon:
When you press "Show less" on an expanded sandbox, you shouldn't be left in the middle of the page. You should be scrolled up so that the "Show more" button (and the sandbox) is visible again.


gaearon:
voilà

PR #4057, opened 2021-11-06
Fix scroll to last scroll position after close modal search

lex111:
Currently, after closing Algolia search modal, the current page scrolls to top, so it is necessary to preserve the last scroll position, for this there is an option `initialScrollY`.

### DIFF
--- a/beta/src/components/Search.tsx
+++ b/beta/src/components/Search.tsx
@@ -42,7 +42,6 @@ const options = {
   appId: siteConfig.algolia.appId,
   apiKey: siteConfig.algolia.apiKey,
   indexName: siteConfig.algolia.indexName,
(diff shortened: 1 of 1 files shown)


beta/src/components/Search.tsx
```diff
@@ -42,7 +42,6 @@ const options = {
   appId: siteConfig.algolia.appId,
   apiKey: siteConfig.algolia.apiKey,
   indexName: siteConfig.algolia.indexName,
-  rednerModal: true,
```
lex111, line 45: This is invalid option, so removed it.

gaearon:
oh wow. and thanks for all these fixes! very appreciated and exciting

PR #4058, opened 2021-11-06
[Doesn't work yet] Use MDX 2

gaearon:
This.. doesn't really work. I don't know why. If someone knows what the missing steps are, would be appreciated!


beta/package.json
```diff
@@ -22,9 +22,9 @@
     "@docsearch/css": "3.0.0-alpha.41",
     "@docsearch/react": "3.0.0-alpha.41",
     "@headlessui/react": "^1.3.0",
-    "@mdx-js/mdx": "^1.6.16",
-    "@mdx-js/react": "^1.6.16",
-    "@mdx-js/tag": "^0.20.3",
+    "@mdx-js/mdx": "^2.0.0-rc.1",
+    "@mdx-js/react": "^2.0.0-rc.1",
+    "@mdx-js/tag": "^1.0.0-alpha.6",
```
wooorm, line 27: this is no longer a thing, it was removed in v1! `@mdx-js/react` was the replacement

beta/next.config.mjs
```diff
@@ -2,11 +2,15 @@
  * Copyright (c) Facebook, Inc. and its affiliates.
  */
 
-const path = require('path');
-const {remarkPlugins} = require('./plugins/markdownToHtml');
+import path from 'path';
+import {remarkPlugins} from './plugins/markdownToHtml.mjs';
+import { createRequire } from 'module';
```
wooorm, line 7:
```suggestion
import {createRequire } from 'module';
```

beta/plugins/markdownToHtml.mjs
```diff
@@ -6,18 +14,16 @@ const unrwapImages = require('remark-unwrap-images'); // Removes <p> wrapper aro
...
-    smartyPants,
-  ],
-  markdownToHtml,
-};
+export const remarkPlugins = [
+  remarkGfm,
+  externalLinks,
+  customHeaders,
+  images,
+  unrwapImages,
```
wooorm, line 22:
```suggestion
  unwrapImages,
```

Typo, should be fixed above as well

beta/plugins/markdownToHtml.mjs
```diff
@@ -6,18 +14,16 @@ const unrwapImages = require('remark-unwrap-images'); // Removes <p> wrapper aro
...
-    images,
-    unrwapImages,
-    smartyPants,
-  ],
-  markdownToHtml,
-};
+export const remarkPlugins = [
+  remarkGfm,
+  externalLinks,
+  customHeaders,
```
wooorm, line 20: This plugin is a really ancient copy/paste version — i don‘t think it’ll work anymore.
I’d recommend using https://github.com/rehypejs/rehype-autolink-headings instead, I think it’s a modern replacement.

I’d also recommend replacing `remark-external-links` with `rehype-external-links`.

Note that both plugins are called `rehype-*`. That’s because they work on an HTML ast.
`remark` is the markdown ecosystem — `rehype` is the HTML ecosystem.

beta/plugins/markdownToHtml.mjs
```diff
@@ -6,18 +14,16 @@ const unrwapImages = require('remark-unwrap-images'); // Removes <p> wrapper aro
...
+  images,
+  unrwapImages,
+  smartyPants,
+];
 
-async function markdownToHtml(markdown) {
+export async function markdownToHtml(markdown) {
   const result = await remark()
```
wooorm, line 27: What is this file doing? How is this function used?
Is the input MDX or markdown?
If it’s MDX, then there will be JSX components, right? Shouldn’t those components be evaluated?

wooorm: A recent version of this pipeline would be:

```js
import {unified} from 'unified'
import remarkParse from 'remark-parse'
import remarkImages from 'remark-images'
import remarkUnwrapImages from 'remark-unwrap-images'
import remarkSmartyPants from './your-custom-plugin.js'
import remarkRehype from 'remark-rehype'
import rehypeAutolinkHeading from 'rehype-autolink-headings'
import rehypeExternalLinks from 'rehype-external-links'
import rehypeStringify from 'rehype-stringify'

const result = await unified()
  .use(remarkParse) // Parse markdown
  .use(remarkImages) // Simpler image syntax (no `alt` support though!! Do you need this?)
  .use(remarkUnwrapImages) // Remove `<p>` wrapper around images
  .use(remarkSmartyPants) // !!If possible I’d recommend forcing this in a style guide rather than trying to fix it like this
  .use(remarkRehype) // Turn markdown AST into HTML AST
  .use(rehypeAutolinkHeading) // Add links to headings (might want to configure to prepend an icon)
  .use(rehypeExternalLinks) // Add `rel=nofollow` and such
  .use(rehypeStringify) // Serialize HTML
  .process('string')
  ```
  
But the big question remains is whether you’re dealing with ESM, expressions, and JSX that needs to be evaluated or not?

wooorm (commented):
some more

beta/plugins/markdownToHtml.mjs
gaearon: I think you can pretty much ignore this function, it's only used for some tangential thing (i think maybe on the blog) which isn't hooked up to the beta site yet

gaearon:
https://github.com/reactjs/reactjs.org/pull/5107

PR #4066, opened 2021-11-06
Fixed typing mdxc components

CatsAndDogs123:
This is my first time opening a pull request, so sorry if im doing something wrong.

I saw a type error in mdxc components, this error happens when you send children to copmponent like this:

```
<SimpleCallout title="You will learn">{children}</SimpleCallout>
```
And in the SimpleCallout component you distrcture the props {title, children} and specify their types.

When using the SimpleCallout it doesn't understand that this is a children prop and it says that you miss a a children prop.

We can get around this by defining the type of the component as React.FC which includes the type of the children inside and you don't have to specify the children type, and therefore you don't get that error.

Generally, its better to define every component with the type React.FC because it saves us defning some types that every functional component has like children or the returned type.

On the fly I added some types in that file, and refactored a small parts.

### DIFF
--- a/beta/src/components/MDX/MDXComponents.tsx
+++ b/beta/src/components/MDX/MDXComponents.tsx
@@ -44,14 +44,14 @@ const UL = (p: JSX.IntrinsicElements['ul']) => (
   <ul className="ml-6 my-3 list-disc" {...p} />
 );
 
-const Divider = () => (
+const Divider = (): React.FC => (
   <hr className="my-6 block border-b border-border dark:border-border-dark" />
 );
 
-const Gotcha = ({children}: {children: React.ReactNode}) => (
+const Gotcha = ({children}: React.FC) => (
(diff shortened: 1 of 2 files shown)


gaearon:
What’s the alternative solution? Is it to specify the children explicitly in types? I think I would prefer that over FC.

CatsAndDogs123:
@gaearon yeah, alterantive is to specify it explicitly. but i feel like using FC makes the code more readable, since we don't have to specify it, and also it is more correct to use it since at the end of  day it is not a regular function, it is a functional component and whenever we see that FC it makes it obvious.
but i guess it depends on personal preference.

gaearon:
I'd be in favor of not using it and being explicit. See also: https://github.com/facebook/create-react-app/pull/8177.

CatsAndDogs123:
@gaearon i see, thats fair.

PR #4068, opened 2021-11-06
Import HomepageHero component on index.md only

Charca:
# WIP~

### DIFF
--- a/beta/src/components/MDX/MDXComponents.tsx
+++ b/beta/src/components/MDX/MDXComponents.tsx
@@ -12,7 +12,6 @@ import Convention from './Convention';
 import ExpandableCallout from './ExpandableCallout';
 import ExpandableExample from './ExpandableExample';
 import {H1, H2, H3, H4} from './Heading';
-import HomepageHero from './HomepageHero';
 import InlineCode from './InlineCode';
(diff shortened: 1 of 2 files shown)


gaearon:
We can do this but it's not super sustainable. The problem is more general. What's a general solution to this?

gaearon:
I think the proper solution would be to move to Server Components. Seems like we’re not too far off from being able to try that.

Charca:
@gaearon I am not aware of a general solution that doesn't involve loading all components globally, unfortunately :/

Two options come to mind, but none of them are really general:
1. Dedicated sets of components for different layouts (i.e. `HomeLayout`, `APIRefenreceLayout`, `BlogLayout`, etc): this could be useful since there are a few components that are only used in the API reference pages, and others that are exclusive to the "learn" section, for example.
2. Use an mdx plugin that adds missing imports for JSX elements it finds in the document (I haven't explored this option too much, so this could be a terrible idea).

I'm not super familiar with Server Components, but if you think that's the way to go, that's really great to hear!

Charca:
I'll close this PR now, it wasn't a super big improvement anyways. Thanks for the input!

Issue #4027, opened 2021-11-01
[Beta] Alternative solution to challenge `Add and remove a CSS class` behaves inconsistently


beetcb:
> Hello 👋 react team! The new document is very well written, thank you for your efforts!

I dicovered that the alternative solution to [Challenge 1 of 3: Add and remove a CSS class](https://beta.reactjs.org/learn/reacting-to-input-with-state#thinking-about-ui-declaratively) in new docs #3308 behaved differently from the original solution.

This has caused me confusion, when i click on the kitten `img`, the background of outter `div` shouldn't be reset.

The issue can be fixed with a single line of code:
```diff
import { useState } from "react";

export default function Kitten() {
  const [isActive, setIsActive] = useState(false);
  if (isActive) {
    return (
      <div className="background" onClick={() => setIsActive(false)}>
        <img
          className="kitten kitten--active"
          alt="Kitten"
          src="https://placekitten.com/100"
          onClick={(e) => {
+           e.stopPropagation();
          }}
        />
      </div>
    );
  }
  return (
    <div className="background background--active">
      <img
        className="kitten"
        alt="Kitten"
        src="https://placekitten.com/100"
        onClick={() => setIsActive(true)}
      />
    </div>
  );
}
```
Failed to find source code for PR, so I hope this information is helpful!

gaearon:
Ah great catch. The file is here: https://github.com/reactjs/reactjs.org/blob/main/beta/src/pages/learn/reacting-to-input-with-state.md

Would you like to send a PR?

beetcb:
Thanks 😊, I'll create one and link to this issue

PR #4072, opened 2021-11-07
Fix alternative solution to add and remove a CSS class (#4027)

beetcb:
Fix provided for #4027 

There is no lint(or prettier) rules for markdown, hope this style meet your convention!

@gaearon


gaearon:
thanks!

PR #4074, opened 2021-11-07
Fixes wrong breadcrumb for 404 page

harish-sethuraman:
For 404 route we are showing wrong bread crumb (learn react) and showed home as active but showed options of learn. So removed breadcrumb for 404 route.

It seems like the when the layout was not specified we were taking Learn as the default subtree 
https://github.com/reactjs/reactjs.org/blob/7f995398f4cfffe9132341d4380b716d65d3a214/beta/plugins/md-layout-loader.js#L18
So changed the default sub tree to Home and then had to add layout to all the pages inside the learn directory.

Verified changes with both sidenav and mobile nav for both 404 page and valid page.

Before:
Screenshot 2021-11-07 at 3 46 41 PM
After:
Screenshot 2021-11-07 at 3 47 42 PM

Also the navigation highlights Home but shows options of learn.
Screenshot 2021-11-07 at 6 31 30 PM

### DIFF
--- a/beta/plugins/md-layout-loader.js
+++ b/beta/plugins/md-layout-loader.js
@@ -15,7 +15,7 @@ const fm = require('gray-matter');
 module.exports = async function (src) {
   const callback = this.async();
   const {content, data} = fm(src);
-  const layout = data.layout || 'Learn';
+  const layout = data.layout || 'Home';
   const code =
     `import withLayout from 'components/Layout/Layout${layout}';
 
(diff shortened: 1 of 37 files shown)


lex111:
@gaearon what if we automatically set a layout based on a content file path? Then we don't have to manually specify layout field in frontmatter of MD file.

gaearon:
Oh yeah that would be great.

PR #4075, opened 2021-11-08
fix(sandpack): upgrade dependencies

danilowoz:
@gaearon This fixes a couple of issues related to the Codemirror:
- [x] Command + backspace to remove entire line;
- [x] "Return" button to add a new line on mobile devices;
- [x] Decorations still working properly;


gaearon:
In general that `sandpack.css` file is extremely hacky and fragile, any help simplifying it and removing hacks is appreciated!

PR #4081, opened 2021-11-10
MDX 2 (attempt 2)

wooorm:
Related to: GH-4058.

Note this PR is giant. It changes the format for heading IDs. Please review by commit and skip most files in `` Use {/*id*/} for headings because MDX has expressions ``.

This PR gets stuff working-ish, but there’s still more to do.

Here are some notes:

A bunch of stuff on the site isn’t working.
My guess is that that has to do with `mdxType` and `isMDXComponent` no longer existing, and the rendered VDOM nodes being different, which are sniffed in several components here.

- The top nav bar isn’t showing (`This page was not added to one of the sidebar JSON files`), I don’t know why. Probably similar to the VDOM sniffing in `MarkdownPage`
- This currently completely breaks `LayoutPost`s ToC stuff.
  It’s sniffing the generated VDOM nodes that will be rendered. That changed.
  Can someone explain to me what the goal of the ToC is, and we can make a nice reusable version of it instead?
- There are a bunch of react warnings that’ll have to be investigated
- `inlineCode` was removed — need to figure out a way to enhance block code (`pre > code`) and differentiate with “inline” code (`code` not in `pre`)
- To do: Have to look at `MDX/APIAnatomy`, `MDX/Challanges/Challenges`, `MDX/CodeDiagram`, `MDX/Link`, `MDX/PackageImport`, `MDX/Sandpack`, `pages/_app`
- MDX provider is not used for a context based API — should be possible to pass `components` to `MDXContent` instead
- Some stuff is now in xdm, which will get released in the next MDX RC, swap that out last

---

My goal is to get some early feedback in. Will clean up later.


lex111:
Wonder if it is possible to keep the current headings IDs format? Will it work in MDX v2 or not?

wooorm:
> Wonder if it is possible to keep the current headings IDs format? Will it work in MDX v2 or not?

Not really. Braces are used as expression in JSX and thus they were added now in MDX. There could be ways around it. But I can see more expressions being added, to the point where it’d become more and more complex to differentiate between the two

gaearon:
>My guess is that that has to do with mdxType and isMDXComponent no longer existing, and the rendered VDOM nodes being different, which are sniffed in several components here.

What's the alternative to this? We do quite a bit of post-processing in a few places. Though I'm happy to simplify if we know how.

wooorm:
> What's the alternative to this? We do quite a bit of post-processing in a few places. Though I'm happy to simplify if we know how.

As I understand it, the code is currently “sniffing” React’s VDOM. It does some cool stuff! But I personally would say it’s error prone (i.e., the VDOM might change), it also “feels” sort of “impure” to me.
On the other hand, it’s very similar to ASTs, and ASTs are nice, so if it does what’s needed, :+1:.

That also leads to where I’d solve it: the AST. First two examples: I’ve previously shared [`remcohaszing/remark-mdx-frontmatter`](https://github.com/remcohaszing/remark-mdx-frontmatter), which takes YAML frontmatter and turns it into an `export const frontmatter = {...}`. A smaller, simpler, similar example is: [`remcohaszing/rehype-mdx-title`](https://github.com/remcohaszing/rehype-mdx-title). So what I’m thinking is to make a plugin, which I’m sure other people would be interested in as well, that adds an export:

```js
export const tableOfContents = {title: 'Hello, World!', url: '#hello-world', rank: 1, children: [/*...*/]}
```

I believe the “sniffing” is mostly done for the ToC, but I haven’t investigated everything.

gaearon:
There’s quite a few places where we do that but I agree we should get rid of that.

wooorm:
Externalized the heading braces stuff in a different PR — please review that separately!

gaearon:
Where are we with the state of this effort? What's needed to finish it?

wooorm:
Hey! Thanks for the ping and sorry I didn’t communicate that yet.
It’s a bit more involved than I initially expected due to the sniffing/rendered node inspection.
I do have the knowledge to help with MDX 2 but I don’t have knowledge of this repo to make all the decisions myself.
This PR is kinda a riff on your earlier PR, and I’m :+1: for going about it that way, someone else taking a stab at this idea here as it’s not really a priority for me.

- I did land the table of contents, but haven’t yet looked at anatomy, challenges, code diagrams, package imports, sandpack, or links.
- Then, I think the code components needs to be migrated (`inlineCode` isn’t a thing anymore, it’s just HTML tag names so `pre` and `code` or just `code`).
- Last, the nav stuff isn’t showing, which might be resolved by the other work?

gaearon:
https://github.com/reactjs/reactjs.org/pull/5107

PR #4082, opened 2021-11-10
Allow page zooming/scaling

lex111:
We need to allow page content scaling/zooming, since it is essential to users with low vision, for example.

### DIFF
--- a/beta/src/components/Seo.tsx
+++ b/beta/src/components/Seo.tsx
@@ -27,7 +27,7 @@ export const Seo = withRouter(
 
       <meta
         name="viewport"
-        content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=0"
+        content="width=device-width, initial-scale=1"
       />
 
       {title != null && <title key="title">{title}</title>}
--- a/beta/src/styles/sandpack.css
+++ b/beta/src/styles/sandpack.css
@@ -191,3 +191,12 @@ html.dark .sp-loading {
   padding-left: var(--sp-space-1) !important;
   padding-right: var(--sp-space-1) !important;
 }
+
+/* For iOS: prevent browser zoom when clicking on sandbox. */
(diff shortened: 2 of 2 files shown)


gaearon:
Can we do this but disable the iOS built-in “zoom into the sandbox on click” behavior somehow?

lex111:
Yes, this can be achieved, but then code in sandboxes will be increased to 16px (only in which case iOS will not scale a page), see https://stackoverflow.com/questions/11064237/prevent-iphone-from-zooming-form for more details (it's the old question, but mentioned solution still works). See also https://twitter.com/joshwcomeau/status/1379782931116351490?s=12

gaearon:
Btw what devices does this actually disable zooming on?

lex111:
At the very least, this applies to Android users. And in general, it is not recommended to disable page scaling, it is bad in terms of a11y.

gaearon:
> I have tested it on an iPhone 12, but anyway it is not recommended to disable page scaling, it is bad in terms of a11y.

I’m asking because I’m able to pinch the existing site (iPhone 11). So I want to be sure which interaction exactly we’re fixing.

lex111:
This solution should work well on iPhone 11 too. As mentioned earlier, it's possible to increase font size only for focus state to prevent browser zooming in iOS, then font size in sandboxes will keep same as it in the current production version if you want. (Probably not many people will edit code in sandbox.)

gaearon:
Sorry, I wasn't clear. I want to understand what exact scenario this PR is trying to fix by changing the meta tag. I know the meta tag says it's not scalable, but I'm able to zoom it on iOS just fine already. So I want to know what is the actual thing that's currently broken on the main branch.

lex111:
The problem is that on Android devices at the moment doesn't work zooming/scaling, an that's what I want to fix in the current PR.

PR #4083, opened 2021-11-10
Remove unwanted dependency

harish-sethuraman:
There are some packages that are not imported and aren't being used.
Are we planning to use these by any way later on? 
Or shall we remove them all ?


gaearon:
nice, thanks.

PR #4087, opened 2021-11-11
Fixing Eslint configuration 

harish-sethuraman:
What is the problem ? 
- The ESLint previously ran smoothly but seems like it didn't run in the beta dir 🤔 
- When ever I ran build it was asking to upgrade 
`error - Your project has an older version of ESLint installed (6.8.0). Please upgrade to ESLint version 7`
- Next.js comes with integrated ESLint and we aren't consuming that 🤔 

Solution: 
- Revamped Linting and now it works fine. (warnings are thrown and errors have been fixed)
- ~Also fixed errors for [anchor is valid](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md)~
- Added ESLint for pre commit too.

We might consider adding newer rules as well. 

Let me know your thoughts 😄


beta/src/components/Breadcrumbs.tsx
```diff
@@ -17,9 +17,9 @@ function Breadcrumbs() {
             <div className="flex mb-3 mt-0.5 items-center" key={i}>
               <React.Fragment key={crumb.path}>
                 <Link href={crumb.path}>
-                  <a className="text-link dark:text-link-dark text-sm tracking-wide font-bold uppercase mr-1 hover:underline">
+                  <button className="text-link dark:text-link-dark text-sm tracking-wide font-bold uppercase mr-1 hover:underline">
```
gaearon, line 20: This doesn't look right. This is a link, not a button, right? `<NextLink>` is the thing that's supposed to add a route, unless I'm missing something?

harish-sethuraman: https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md
So while I ran ESLint with old config it threw this error. We had no valid href for the anchor but used anchor tags. That's why I replaced it with button 🤔. Should I change everything to anchor tags and silence that rule like it was before ?

harish-sethuraman: BTW I changed those buttons to anchor tags and i guess that rule is not active by default in next ESlint config.

gaearon:
So does `lint` task run on CI and fail it if lint fails? Did it do that before this PR? I thought it did..

harish-sethuraman:
I haven't added the git workflow yet. I added to pre commit hook. So when someone tries to commit we try to fix auto fixable issues and then show error if any.

gaearon:
Doesn’t the existing Build / Lint action do this?

harish-sethuraman:
No it wasn't happening before.

gaearon:
Ah, the action does this for the old site.

beta/src/components/DocsFooter.tsx
```diff
@@ -69,12 +69,13 @@ function FooterLink({
...
     <NextLink href={href}>
       <a
-        className={cn(
-          'flex gap-x-4 md:gap-x-6 items-center w-full md:w-80 px-4 md:px-5 py-6 border-2 border-transparent text-base leading-base text-link dark:text-link-dark rounded-lg group focus:text-link dark:focus:text-link-dark focus:bg-highlight focus:border-link dark:focus:bg-highlight-dark dark:focus:border-link-dark focus:border-opacity-100 focus:border-2 focus:ring-1 focus:ring-offset-4 focus:ring-blue-40 active:ring-0 active:ring-offset-0 hover:bg-gray-5 dark:hover:bg-gray-80',
-          {
-            'flex-row-reverse justify-self-end text-right': type === 'Next',
-          }
-        )}>
+       className={cn(
+        'flex gap-x-4 md:gap-x-6 items-center w-full md:w-80 px-4 md:px-5 py-6 border-2 border-transparent text-base leading-base text-link dark:text-link-dark rounded-lg group focus:text-link dark:focus:text-link-dark focus:bg-highlight focus:border-link dark:focus:bg-highlight-dark dark:focus:border-link-dark focus:border-opacity-100 focus:border-2 focus:ring-1 focus:ring-offset-4 focus:ring-blue-40 active:ring-0 active:ring-offset-0 hover:bg-gray-5 dark:hover:bg-gray-80',
```
gaearon, line 73: What changed here and why?

beta/src/components/ExternalLink.tsx
```diff
@@ -7,12 +7,11 @@ import * as React from 'react';
 export function ExternalLink({
   href,
   target,
-  children,
   ...props
 }: JSX.IntrinsicElements['a']) {
   return (
     <a href={href} target={target ?? '_blank'} rel="noopener" {...props}>
-      {children}
+      {props.children}
```
gaearon, line 14: What is this meant to fix? Why?

beta/src/components/Layout/Nav/NavButtonLink.tsx
```diff
@@ -1,7 +1,7 @@
 /*
  * Copyright (c) Facebook, Inc. and its affiliates.
  */
-
+// Why do we have this component ? seems like this is not being used anywhere
```
gaearon, line 4: Then delete it :)

beta/src/components/MDX/CodeBlock/CodeBlock.tsx
```diff
@@ -109,7 +109,7 @@ const CodeBlock = React.forwardRef(
     );
   }
 );
-
+CodeBlock.displayName = 'CodeBlock'
```
gaearon, line 112: ?

beta/src/components/DocsFooter.tsx
harish-sethuraman: > came to the next line 👀 . Shall I put him back ?

beta/src/components/ExternalLink.tsx
harish-sethuraman: Screenshot 2021-11-12 at 10 13 14 PM

beta/src/components/Layout/Nav/NavButtonLink.tsx
harish-sethuraman: Will Do

beta/src/components/MDX/CodeBlock/CodeBlock.tsx
harish-sethuraman: Screenshot 2021-11-12 at 10 14 04 PM

gaearon: That seems like a bad rule to enable.

harish-sethuraman: That came in by default. Shall I disable it then?

beta/src/components/DocsFooter.tsx
harish-sethuraman: I've placed the old content as it is since there was minor changes in indentation.

beta/src/components/Layout/Nav/NavButtonLink.tsx
harish-sethuraman: Deleted this component

beta/src/components/MDX/CodeBlock/CodeBlock.tsx
gaearon: Where is it coming from? Which config?

harish-sethuraman: It is coming from `react/display-name` 🤔  I extended the `next` config and it seems to include `eslint-plugin-react`
https://nextjs.org/docs/basic-features/eslint

gaearon: I don't see `react-display-name` in the list of rules on that page.

harish-sethuraman: Screenshot 2021-11-12 at 11 51 21 PM
and `eslint-plugin-react` has display-name
https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/display-name.md

gaearon: Okay I see what's happening now. It's just because of forwardRef.

Try doing this:

```js
const Foo = forwardRef(function Foo() {
 ...
})
```

harish-sethuraman: Done updates the arrow function to function CodeBlock and removed the `CodeBlock.displayName`

beta/src/components/ExternalLink.tsx
gaearon: But the original code _is_ doing that. So why was the original code flagged?

gaearon: Maybe rename `props` to something like `rest` and it'll work.

harish-sethuraman: Sorry I missed it. Nice catch Thanks 🏅

beta/src/components/MDX/MDXComponents.tsx
```diff
@@ -234,8 +234,8 @@ function IllustrationBlock({
       ) : null}
       {sequential ? (
         <ol className="mdx-illustration-block flex">
-          {images.map((x: any) => (
-            <li className="flex-1">{x}</li>
+          {images.map((x: any, i :number) => (
```
gaearon, line 237: nit: `i: number`, not `i :number`. why didn't Prettier catch this?

harish-sethuraman: 🤔  Not sure how this was missed.

harish-sethuraman: Changed

gaearon: Yea but can we figure out why Prettier didn't fail CI if it mismatched.

gaearon: Maybe Prettier only runs ci for non-beta code? Then let's fix that.

harish-sethuraman: It seems like we have the `prettier` command but haven't run it 🤔 . Should we add the same to pre commit as well ?

gaearon: Yes. Also it should fail CI if it _doesn't_ pass. See `ci-check` task in package.json. We need to run this on CI if it doesn't already.

harish-sethuraman: So we already have a workflow for checking in the legacy site. Shall I add the ci-check to run there itself for the beta directory too ? Or shall I add a seperate workflow for it ?

harish-sethuraman: https://github.com/reactjs/reactjs.org/pull/4100 @gaearon  Since there were lot of file changes as prettier ran,  I've actually raised a seperate PR for that.Now we will run prettier before commits and have a workflow to check if any code is not prettified.

PR #4089, opened 2021-11-11
Enable dark mode in browser UI

lex111:
We can change a browser's color scheme in according with used dark mode on website. For example, then scrollbars will use dark colors.

Before:

image

After:

image


gaearon:
:OOOOOOOO

PR #4090, opened 2021-11-11
Set action key for search depending on OS

lex111:
Probably not everyone uses macOS, so they might be confused by the Command key instead of the Ctrl one in the search shortcut.

image

image

### DIFF
--- a/beta/src/components/Search.tsx
+++ b/beta/src/components/Search.tsx
@@ -32,7 +32,7 @@ function Kbd(props: {children?: React.ReactNode}) {
   return (
     <kbd
       className="border border-transparent mr-1 bg-wash dark:bg-wash-dark text-gray-30 align-middle p-0 inline-flex justify-center items-center  text-xs text-center rounded"
-      style={{width: '2.25em', height: '2.25em'}}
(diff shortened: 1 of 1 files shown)


beta/src/components/Search.tsx
```diff
@@ -51,6 +51,12 @@ export const Search: React.FC<SearchProps> = ({
 }) => {
   const [isLoaded] = React.useState(true);
   const [isShowing, setIsShowing] = React.useState(false);
+  const actionKey = React.useMemo<string | null>(() => {
+    if (typeof navigator !== 'undefined') {
+      return /(Mac|iPad)/i.test(navigator.platform) ? '⌘' : 'Ctrl';
+    }
+    return null;
+  }, []);
```
gaearon, line 59: This seems like it would create a mismatch between server and client, wouldn't it?

I'd suggest either using an effect to set it after first render, or to have an inline script in head that sets a CSS class, and then toggle based on CSS.

lex111: Yes, but why is that a bad thing in this case? Do you prefer to set the default action key for a prerendered page?
useEffect will cause a re-render every time after a route changed...

PR #4093, opened 2021-11-12
[fix]: add-listner is deprecated, switched those to add-event-listner

CatsAndDogs123:
According to https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/addListener `addListener` and `removeListener` are deprecated.

Switched them with `addEventListener` and `removeEventListener`.

Also i saw the whole app uses `addEventListener`  and `removeEventListener` so now it matches the standart.


harish-sethuraman (approved):

harish-sethuraman:
Thanks

PR #4097, opened 2021-11-12
Fix description to match code in referencing-values-with-refs.md

MarekBodinger:


beta/src/pages/learn/referencing-values-with-refs.md
```diff
@@ -656,4 +656,4 @@ export default function Chat() {
 
 </Solution>
 
-</Challenges>
\ No newline at end of file
+</Challenges>
```
harish-sethuraman, line 659: What is the change in this line?

gaearon: This is usually accidental change to newline at the end of the file. It can be ignored.

PR #4105, opened 2021-11-14
Use custom ID format that works better with MDX 2

wooorm:
This website currently uses a custom (non-standard) syntax extension to markdown to add custom IDs to headings.
“Custom” is not really a correct term: almost all of them (except 3) match exactly what GitHub would slug for them.
The reason for their existence is i18n.

The syntax extension to add IDs to heading like so:

```
## Some heading {#some-heading}
```

Does not interfere with markdown, so it works pretty nicely there (though it’s not supported in many places).
But in MDX 2, support for expressions was added similar to how expressions work in JSX:

```
# Hello, {props.name}
```

…so you can probably see the problem there. Braces signal (JavaScript) expressions.
To solve that problem, this PR switches to using MDX/JSX comments as the last “child” in headings instead:

```
## Some heading {/*some-heading*/}
```

This PR is part of GH-4081. Externalized as this might be able to go faster than the other PR, in which case that PR would become much easier to review.

For reviewers:

* PR is giant. Almost all of it is automated from two scripts, so you don’t have to look through every file
* See the first commit, which fixes two unicode whitespace cases which do not match how proper markdown works
* review `beta/scripts/generateHeadingIds.js` and `beta/plugins/remark-header-custom-ids`. The rest is docs automated by them.

/cc @lex111

### DIFF
--- a/beta/package.json
+++ b/beta/package.json
@@ -35,6 +35,7 @@
     "react": "18.0.0-alpha-930c9e7ee-20211015",
     "react-collapsed": "3.1.0",
     "react-dom": "18.0.0-alpha-930c9e7ee-20211015",
+    "remark-slug": "^7.0.0",
     "scroll-into-view-if-needed": "^2.2.25"
   },
   "devDependencies": {
--- a/beta/package.json
+++ b/beta/package.json
@@ -35,7 +35,6 @@
     "react": "18.0.0-alpha-930c9e7ee-20211015",
     "react-collapsed": "3.1.0",
     "react-dom": "18.0.0-alpha-930c9e7ee-20211015",
-    "remark-slug": "^7.0.0",
     "scroll-into-view-if-needed": "^2.2.25"
   },
   "devDependencies": {
@@ -77,6 +76,7 @@
     "remark-external-links": "^7.0.0",
     "remark-html": "^12.0.0",
     "remark-images": "^2.0.0",
+    "remark-slug": "^7.0.0",
     "remark-unwrap-images": "^2.0.0",
     "retext": "^7.0.1",
     "retext-smartypants": "^4.0.0",
--- a/beta/src/pages/community/meet-the-team.md
+++ b/beta/src/pages/community/meet-the-team.md
@@ -7,13 +7,13 @@ React development is led by a small dedicated team working full time at Facebook
 
 </Intro>
 
-## React Core {#react-core}
+## React Core {/*react-core*/}
 
 The React Core team members work full time on the core component APIs, the engine that powers React DOM and React Native, React DevTools, and the React documentation website.
 
 Current members of the React team are listed in alphabetical order below.
 
-### Andrew Clark {#andrew-clark}
+### Andrew Clark {/*andrew-clark*/}
 
 ![Andrew Clark](../images/team/acdlite.jpg)
 
@@ -21,7 +21,7 @@ Current members of the React team are listed in alphabetical order below.
 
 Andrew got started with web development by making sites with WordPress, and eventually tricked himself into doing JavaScript. His favorite pastime is karaoke. Andrew is either a Disney villain or a Disney princess, depending on the day.
 
-### Brian Vaughn {#brian-vaughn}
+### Brian Vaughn {/*brian-vaughn*/}
 
 ![Brian Vaughn](../images/team/bvaughn.jpg)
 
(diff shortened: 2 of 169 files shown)


beta/src/pages/community/meet-the-team.md
```diff
@@ -7,85 +7,85 @@ React development is led by a small dedicated team working full time at Facebook
...
+### Rick Hanlon {/*rick-hanlon*/}
 
 ![Ricky](../images/team/rickhanlonii.jpg)
 
 [@rickhanlonii on GitHub](https://github.com/rickhanlonii) &middot; [@rickhanlonii on Twitter](https://twitter.com/rickhanlonii)
 
 Ricky majored in theoretical math and somehow found himself on the React Native team for a couple years before joining the React team. When he's not programming you can find him snowboarding, biking, climbing, golfing, or closing GitHub issues that do not match the issue template.
 
-### Sebastian Markbåge {#sebastian-markbåge}
+### Sebastian Markbåge {#sebastian-markbåge} {/*sebastian-markbåge-sebastian-markbåge*/} {/*sebastian-markbåge-sebastian-markbåge-sebastian-markbåge-sebastian-markbåge*/} {/*sebastian-markbåge-sebastian-markbåge-sebastian-markbåge-sebastian-markbåge-sebastian-markbåge-sebastian-markbåge-sebastian-markbåge-sebastian-markbåge*/}
```
harish-sethuraman, line 72: Is something wrong here ? since we have special character ?

wooorm: nice catch! Will investigate!

harish-sethuraman: 👍🏻

beta/src/pages/learn/writing-markup-with-jsx.md
```diff
@@ -170,7 +170,7 @@ This is how Hedy Lamarr's image and list items look closed:
 </>
 ```
 
-### 3. camelCase <s>all</s> most of the things! {#3-camelcase-salls-most-of-the-things}
+### 3. camelCase <s>all</s> most of the things! {/*3-camelcase-salls-most-of-the-things*/}
```
harish-sethuraman, line 173: This is the only case where we have a tag in heading. shall we skip the tags? or leave it as such?

beta/src/pages/community/meet-the-team.md
wooorm: fixed! indeed, every generate was adding another “ID”

harish-sethuraman: Nice work! 😄

beta/src/pages/learn/writing-markup-with-jsx.md
wooorm: We’d need an actual HTML parser to parse HTML properly, and match exactly how GH does HTML, which gets complex fast (but I could make it work).
If possible, I’d recommend not putting the `<s>` in there.
Alternatively, MDX 2 would remove it from the slugs as well.

harish-sethuraman: https://beta.reactjs.org/learn/writing-markup-with-jsx#3-camelcase-salls-most-of-the-things

This heading might actually need a <s> I guess. 
Would it be good if we could remove the opening and closing tags using regex? We might not get complex HTML in heading and this might work?

wooorm: Trying to parse HTML with regexes is... not perfect. The alternative is using `remark-rehype`, `rehype-raw`, and then `rehype-slug`, which works! Just a bunch of code.

However, this exact ID was already there before though, right? I understand what you want, it’ll get complex but maybe it’s good. But I’d think that that could be another PR?

harish-sethuraman: Yep Sounds good 😄 !

beta/package.json
```diff
@@ -35,6 +35,7 @@
     "react": "18.0.0-alpha-930c9e7ee-20211015",
     "react-collapsed": "3.1.0",
     "react-dom": "18.0.0-alpha-930c9e7ee-20211015",
+    "remark-slug": "^7.0.0",
```
harish-sethuraman, line 38: Do we need it in dependencies or devDependencies?  🤔

wooorm: moved to dev-deps!

lex111 (approved):
Not really familiar with MDX v2, but these changes look good to me :+1:

PR #4106, opened 2021-11-15
Update keeping-components-pure.md

vikingviolinist:


gaearon:
that's intentional and is valid style. (maybe we could change it but that's a separate discussion)

Issue #4096, opened 2021-11-12
Docs beta: useRef Stopwatch doesn't stop if you click start twice


Aidurber:
I was just browsing the beta docs and playing around with the examples, I came across the useRef stop watch and spotted a bug. 

**Steps**
1. Go view the stop watch: https://beta.reactjs.org/learn/referencing-values-with-refs#example-building-a-stopwatch
2. Click start twice
3. Click stop
4. It doesn't stop

dev-tom-0108:
Thanks so much

Aidurber:
This seems to have been fixed by #4110 closing.

PR #4110, opened 2021-11-16
small fixes to stopwatch codesandbox

aayush-k:
noticed that the explanation for the first stopwatch codesandbox mentions "update the time every 10 milliseconds" so just updated the codesandbox to reflect that

also there's a small bug (#4096 ) in the stopwatch codesandboxes where each call to `handleStart()` sets a new interval without checking if there's already one ongoing. 

ie: If the user accidentally double clicks the start button, they set two intervals for updating `now` every 10ms and then intervalRef only retains the second interval ID. Thus, it's impossible to actually stop the timer because `handleStop()` will only clear the latest set interval while the original one will keep executing.

### DIFF
--- a/beta/src/pages/learn/referencing-values-with-refs.md
+++ b/beta/src/pages/learn/referencing-values-with-refs.md
@@ -99,9 +99,9 @@ export default function Stopwatch() {
     setNow(Date.now());
 
     setInterval(() => {
-      // Update the current time every 100ms.
+      // Update the current time every 10ms.
       setNow(Date.now());
-    }, 100);
+    }, 10);
   }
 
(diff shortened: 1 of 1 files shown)


import-brain (approved):

beta/src/pages/learn/referencing-values-with-refs.md
```diff
@@ -135,6 +135,10 @@ export default function Stopwatch() {
   const intervalRef = useRef(null);
 
   function handleStart() {
+    // Start counting, if not already.
```
gaearon, line 138: I feel like resetting would be more appropriate behavior.

PR #4114, opened 2021-11-18
Fixes unwanted scrolls in mobile

harish-sethuraman:
Fixes https://github.com/reactjs/reactjs.org/issues/4113


gaearon:
can we somehow make it prefer to break by word boundaries where possible?

harish-sethuraman:
`overflow-wrap: break-word;` used this property to break where ever possible

PR #4120, opened 2021-11-24
Add check-all script for beta site

harish-sethuraman:
We did not have a check-all script in the beta package.json so added that. also removed flow since we dont use it in beta.

`Run yarn check-all from the beta folder. (This will run Prettier, ESLint and flow)`

### DIFF
--- a/beta/README.md
+++ b/beta/README.md
@@ -48,7 +48,7 @@ The documentation is divided into several sections with a different tone and pur
 ### Test the change
 
 1. If possible, test any visual changes in all latest versions of common browsers, on both desktop and mobile.
-2. Run `yarn check-all` from the `beta` folder. (This will run Prettier and ESLint)
+2. Run `yarn check-all` from the `beta` folder. (This will run Prettier, ESLint and validate Types.)
 
 ### Push it
 
--- a/beta/README.md
+++ b/beta/README.md
@@ -48,7 +48,7 @@ The documentation is divided into several sections with a different tone and pur
 ### Test the change
 
 1. If possible, test any visual changes in all latest versions of common browsers, on both desktop and mobile.
-1. Run `yarn check-all` from the `beta` folder. (This will run Prettier, ESLint, and Flow.)
+2. Run `yarn check-all` from the `beta` folder. (This will run Prettier and ESLint)
 
 ### Push it
 
--- a/beta/README.md
+++ b/beta/README.md
@@ -48,7 +48,7 @@ The documentation is divided into several sections with a different tone and pur
 ### Test the change
 
 1. If possible, test any visual changes in all latest versions of common browsers, on both desktop and mobile.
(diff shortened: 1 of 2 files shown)


beta/package.json
```diff
@@ -17,7 +17,8 @@
     "ci-check": "npm-run-all prettier:diff --parallel lint tsc",
```
gaearon, line 17: We have `ci-check`. Let's just use that in README?

harish-sethuraman: We could do that but, The `ci-check` lists the prettier error and doesn't do auto fixing. But in the check-all we will prettify the code.

gaearon: We just need to have something that makes sense locally (so, autofix etc), and something that makes sense for CI (so, fail on mismatch) and have them be consistent.

harish-sethuraman: Then lets go with check-all itself. And we shall run prettier auto fixing and linting with auto fix too.

beta/package.json
```diff
@@ -17,7 +17,8 @@
...
     "tsc": "tsc --noEmit",
     "start": "next start",
-    "postinstall": "is-ci || (cd .. && husky install beta/.husky)"
+    "postinstall": "is-ci || (cd .. && husky install beta/.husky)",
+    "check-all": "npm-run-all prettier lint:fix"
```
gaearon, line 21: What about TS?

harish-sethuraman: added 👍🏻

beta/README.md
```diff
@@ -48,7 +48,7 @@ The documentation is divided into several sections with a different tone and pur
 ### Test the change
 
 1. If possible, test any visual changes in all latest versions of common browsers, on both desktop and mobile.
-1. Run `yarn check-all` from the `beta` folder. (This will run Prettier, ESLint, and Flow.)
+2. Run `yarn check-all` from the `beta` folder. (This will run Prettier and ESLint)
```
gaearon, line 51: and TS

harish-sethuraman: Shall I add `This will run Prettier, ESLint and validate Types.` ?

beta/package.json
```diff
@@ -17,7 +17,8 @@
     "ci-check": "npm-run-all prettier:diff --parallel lint tsc",
     "tsc": "tsc --noEmit",
     "start": "next start",
-    "postinstall": "is-ci || (cd .. && husky install beta/.husky)"
+    "postinstall": "is-ci || (cd .. && husky install beta/.husky)",
+    "check-all": "npm-run-all prettier lint:fix tsc"
```
gaearon, line 21: should we use the `--parallel` flag here too?

harish-sethuraman: Nope. We shall run it sequentially since we are auto fixing using prettier and linter. We shall first completely prettify and then run lint.

PR #4122, opened 2021-11-25
Fix array overflow in adding interactivity page

paKanhu:


gaearon:
Thanks but it’s okay — people like to fix this issue themselves. :) we also have a challenge for this on one of the pages.

PR #4123, opened 2021-11-25
fix: challenges navigation

ScytheDraven47:
> As brought up in [feedback by me](https://github.com/reactjs/reactjs.org/issues/3308#issuecomment-978689930)

Fixing a bug on the Challenges/Navigation component wherein the state `scrollPos` was not updated if the active challenge was changed from outside the navigation. (Possible via the "Next Challenge" button below the challenge).

The fix was to remove the state entirely, and refactor into a variable based on the `activeChallenge` prop, so it could be maintained on re-render.

Tested on pages with 1 challenge, and multiple challenges, spamming button clicks, and using all means of navigation (e.g. <, >, "Next Challenge", and challenge names)

### DIFF
--- a/beta/src/components/MDX/Challenges/Navigation.tsx
+++ b/beta/src/components/MDX/Challenges/Navigation.tsx
@@ -23,7 +23,9 @@ export function Navigation({
   const challengesNavRef = React.useRef(
     challenges.map(() => createRef<HTMLButtonElement>())
   );
-  const [scrollPos, setScrollPos] = React.useState(0);
+  const activeOrder =
+    challenges.find(({id}) => id === activeChallenge)?.order || 1;
+  const scrollPos = activeOrder - 1;
   const canScrollLeft = scrollPos > 0;
   const canScrollRight = scrollPos < challenges.length - 1;
 
@@ -37,7 +39,6 @@ export function Navigation({
         containerRef.current.scrollLeft = currentNavRef.offsetLeft;
       }
       handleChange(challenges[scrollPos + 1].id);
-      setScrollPos(scrollPos + 1);
     }
   };
 
@@ -51,7 +52,6 @@ export function Navigation({
         containerRef.current.scrollLeft = currentNavRef.offsetLeft;
       }
       handleChange(challenges[scrollPos - 1].id);
-      setScrollPos(scrollPos - 1);
     }
   };
 
@@ -64,7 +64,6 @@ export function Navigation({
       containerRef.current.scrollLeft = currentNavRef?.offsetLeft || 0;
     }
     handleChange(id);
-    setScrollPos(selectedChallenge);
   };
 
   const handleResize = React.useCallback(() => {
--- a/beta/src/components/MDX/Challenges/Navigation.tsx
+++ b/beta/src/components/MDX/Challenges/Navigation.tsx
@@ -11,21 +11,19 @@ const debounce = require('debounce');
 export function Navigation({
   challenges,
   handleChange,
-  activeChallenge,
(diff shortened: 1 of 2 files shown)


beta/src/components/MDX/Challenges/Navigation.tsx
```diff
@@ -23,7 +23,9 @@ export function Navigation({
   const challengesNavRef = React.useRef(
     challenges.map(() => createRef<HTMLButtonElement>())
   );
-  const [scrollPos, setScrollPos] = React.useState(0);
+  const activeOrder =
+    challenges.find(({id}) => id === activeChallenge)?.order || 1;
```
harish-sethuraman, line 27: We have the `currentChallenge` in its parent component. Would it be good if we use that instead of re computing it again? 
https://github.com/reactjs/reactjs.org/blob/c6353b9853116680ce51352872a3df1078eedee0/beta/src/components/MDX/Challenges/Challenges.tsx#L99

ScytheDraven47: Yes, we could. I was using `activeChallenge` as it was already being passed through as a prop.

However, that could be replaced with `currentChallenge.name` in the Navigation function, then use the `order` attribute as I was already doing.

I will recommit, thank you for the suggestion.

harish-sethuraman (approved):
Nice work thanks 😄

beta/src/components/MDX/Challenges/Navigation.tsx
```diff
@@ -11,19 +11,19 @@ const debounce = require('debounce');
...
   challenges,
   handleChange,
-  activeChallenge,
+  currentChallenge,
   isRecipes,
 }: {
   challenges: ChallengeContents[];
   handleChange: (id: string) => void;
-  activeChallenge: string;
+  currentChallenge?: ChallengeContents;
```
gaearon, line 19: Why is this nullable? Shouldn't there always be a current challenge?

ScytheDraven47: Yes there _**should**_ be a current challenge... However, because the `currentChallenge` variable comes from a `find()` TypeScript can't guarantee its existence. 
A fix/hack would be to place a `!` on the `find()` however, I wasn't sure that is allowed in the coding style guide (and haven't spent a lot of time with TypeScript).
If you're okay with that, I can make the appropriate changes.

gaearon: Seems ok.

gaearon: We can also throw if not found.

beta/src/components/MDX/Challenges/Navigation.tsx
```diff
@@ -16,14 +16,14 @@ export function Navigation({
...
-  currentChallenge?: ChallengeContents;
+  currentChallenge: ChallengeContents;
   isRecipes?: boolean;
 }) {
   const containerRef = React.useRef<HTMLDivElement>(null);
   const challengesNavRef = React.useRef(
     challenges.map(() => createRef<HTMLButtonElement>())
   );
-  const scrollPos = (currentChallenge?.order || 1) - 1;
+  const scrollPos = (currentChallenge.order || 1) - 1;
```
gaearon, line 26: you can drop `|| 1` now?

ScytheDraven47: Apologies. That's what happens when I try and fit a refactor into a 15min work break... Fixed now.

PR #4126, opened 2021-11-26
fix(sandpack): improve memory usage

danilowoz:
@gaearon This introduces a new prop to define the initialization strategy for the code editor and the preview components.

Current, there is no way to control how the code-editor and the preview components, which are the most expensive operations, are going to be initialized. As a result, this can be a problem for pages that contain multiple instances of Sandpack (eg: https://beta.reactjs.org/learn/describing-the-ui), which might lag the initialization of the page and even block the navigation between pages.

**Solution**
From now on, Sandpack supports a new prop called `initMode`, which expecting to receive one of the following values:

- `immediate`: it immediately mounts all components, such as the code-editor and the preview - this option might overload the memory usage and resource from the browser on a page with multiple instances;
- `lazy`: only initialize the components when the user is about to scroll them to the viewport and keep these components mounted until the user leaves the page - this is the default value;
- `user-visible`: only initialize the components when the user is about to scroll them to the viewport, but differently from `lazy`, this option unmounts those components once it's no longer in the viewport - destroying most of the Workers from the CodeSandbox bundler.

In this PR I set the Sandpack component to use the `user-visible` strategy, then we can measure the performance improvements (or not). 

**More improvements**
The code-editor should have been initialized in a more efficient way. Ideally, it could use the new `postTask scheduler` API, which unfortunately it's still not available. So, for while it uses a `setTimeout` to wrap the code-editor init, in order to postpone the most expensive operations.

### DIFF
--- a/beta/package.json
+++ b/beta/package.json
@@ -22,7 +22,7 @@
     "check-all": "npm-run-all prettier lint:fix tsc"
   },
   "dependencies": {
-    "@codesandbox/sandpack-react": "0.13.9-experimental.0",
+    "@codesandbox/sandpack-react": "0.13.9-experimental.5",
     "@docsearch/css": "3.0.0-alpha.41",
     "@docsearch/react": "3.0.0-alpha.41",
     "@headlessui/react": "^1.3.0",
--- a/beta/package.json
+++ b/beta/package.json
@@ -22,7 +22,7 @@
     "check-all": "npm-run-all prettier lint:fix tsc"
   },
   "dependencies": {
-    "@codesandbox/sandpack-react": "0.13.9-experimental.5",
+    "@codesandbox/sandpack-react": "0.13.9-experimental.6",
     "@docsearch/css": "3.0.0-alpha.41",
     "@docsearch/react": "3.0.0-alpha.41",
     "@headlessui/react": "^1.3.0",
--- a/beta/package.json
+++ b/beta/package.json
@@ -22,7 +22,7 @@
     "check-all": "npm-run-all prettier lint:fix tsc"
   },
   "dependencies": {
-    "@codesandbox/sandpack-react": "0.13.6-experimental.0",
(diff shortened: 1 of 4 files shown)


gaearon:
can you rebase this please?

gaearon:
This seems like a nice improvement but pretty flaky.

If I scroll up and down a lot, I'm able to trigger a situation where the in-viewport sandbox has no preview.

Screenshot 2022-01-25 at 18 21 12

gaearon (changes requested):
need to improve the viewport detection because it's not reliable.

danilowoz:
@gaearon It got stuck in an intermediate step between unmounting and remounting the client, and it needed a couple of tweaks on Sandpack side and yours in order to make it work correctly. But from now on, no matter how insanely you scroll, it always mounts the preview again.

Unfortunately, it still takes some time to load the sandbox dependencies once again between this transition (that isn't the nicest experience for the user). But as I mentioned before, we've been working in an internal refactor to skip the transpile step, which will likely improve the dependencies loading a lot.

gaearon:
this seems exceedingly aggressive to me — notice sometimes it seems to clean up something that's _already rendered_, only to render it again. what do you think about increasing the distance that's necessary for a sandbox to be unloaded? and ideally it would never destroy something that hasn't been destroyed anyway and is already in view again.

https://user-images.githubusercontent.com/810438/151606203-4a44b77a-fefb-469d-a513-38257a2330e3.mov

gaearon:
i think i see a bug.

1) scroll to any sandbox
2) scroll far down
3) scroll back up
4) introduce a syntax error

expected: the output shows a compile error message
actual: we see the last rendered output

danilowoz:
@gaearon fixed!

gaearon:
let's give this a go!

gaearon:
stray log?

Screenshot 2022-02-02 at 11 40 16

gaearon:
(it's not a blocker though, let's get this in)

PR #4136, opened 2021-11-30
Changed the version generation script to update vercel.json instead of _redirects

harish-sethuraman:
Trying to address this here https://github.com/reactjs/reactjs.org/pull/3951#issuecomment-981874581

For now I've fixed the existing plugin for the older site since it is broken too. 
Will add a separate node script for the beta site as well

So there were already 40 redirects in the vercel.json. There are 9 version re directs starting from `16.0` to `16.8` in the `versions.yml` so there are 49 redirects now.

### DIFF
--- a/beta/scripts/generateRedirects.js
+++ b/beta/scripts/generateRedirects.js
@@ -12,7 +12,6 @@ const file = readFileSync(versionsFile, 'utf8');
 const versions = safeLoad(file);
 const redirectsFilePath = path.join('vercel.json');
 
-
 function writeRedirectsFile(redirects, redirectsFilePath) {
   if (!redirects.length) {
     return null;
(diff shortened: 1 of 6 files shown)


beta/scripts/generateRedirects.js
```diff
@@ -0,0 +1,69 @@
...
+  /**
+   * We will first read the old config to validate if the redirect already exists in the json
+   */
+  const vercelConfigPath = resolve(__dirname, '../../vercel.json');
+  const vercelConfigFile = readFileSync(vercelConfigPath);
+  const oldConfigContent = JSON.parse(vercelConfigFile);
+  /**
+   * Map data as vercel expects it to be
+   */
+  redirects = redirects.map((redirect) => {
```
gaearon, line 30: Nit: instead of overwriting a variable of a conceptually different type, make a new one.

```js
const vercelRedirects = redirects.map(...)
```

beta/scripts/generateRedirects.js
```diff
@@ -0,0 +1,69 @@
...
+    const {fromPath, isPermanent, toPath} = redirect;
+
+    const pieces = {
```
gaearon, line 33: Nit: just return the object, `pieces` doesn't really explain anything as a name

beta/scripts/generateRedirects.js
```diff
@@ -0,0 +1,69 @@
...
+    };
+
+    return pieces;
+  });
+
+  /**
+   * Make sure we dont have the same redirect already
+   */
+  oldConfigContent.redirects.forEach((data) => {
+    redirects = redirects.filter(
```
gaearon, line 46: It's not efficient to have a loop over all items inside of a loop over all items. Can you think of some way to simplify this?

gaearon (approved):
seems ok. left some nits i'd prefer to see fixed. please check that redirects actually work before merging

beta/scripts/generateRedirects.js
harish-sethuraman: I've modified this logic

beta/scripts/generateRedirects.js
harish-sethuraman: Simplified by using an object :)

beta/vercel.json
```diff
@@ -1,46 +1,253 @@
 {
-  "github": { "silent": true },
+  "github": {
+    "silent": true
+  },
   "trailingSlash": false,
   "redirects": [
-    { "source": "/tips/controlled-input-null-value.html", "destination": "/docs/forms.html#controlled-input-null-value", "permanent": false },
```
harish-sethuraman, line 5: @gaearon  I dont think we need this beta site's vercel.json because we actually do not have any redirects related to the new beta site. Should we change destination to older site? or remove them?

gaearon: beta site will become main site someday so we're going to need them (but we'll revisit their content)

PR #4143, opened 2021-12-01
Added hook to validate if headings are present or not

harish-sethuraman:
Addressing https://github.com/reactjs/reactjs.org/issues/4135#issuecomment-983753299 and followup for https://github.com/reactjs/reactjs.org/pull/4139

Going forward we will validate if the headings are present or not in the beta site's markdown files. 
- If the headings are missing we will stop commit and show them an error message Screenshot 2021-12-02 at 12 32 43 AM
- They can then generate the ids and then commit it

We will let the contributor decide which files to add and with pre commit hook we can make sure that we don't miss out any headings.

EDIT:
 `yarn lint-heading-ids` --> Checks all files and causes an error if heading ID is missing
 `yarn lint-heading-ids --fix` --> Fixes all markdown file's heading IDs
 `yarn lint-heading-ids path/to/markdown.md` --> Checks that particular file for missing heading ID 
 (path can be a directory or particular file)
 `yarn lint-heading-ids --fix path/to/markdown.md` --> Fixes that particular file's markdown IDs 
 (path can be a directory or particular file)
 
 Making it a linter and added auto fixable option.


smikitky:
We could check only the staged files instead of the entire docs. I find this package useful (used in the Japanese fork to run additional linting for style problems):

https://www.npmjs.com/package/lint-staged

```js
// package.json
{
  "dependencies": {
    "lint-staged": "^12.1.2"
  },
  "lint-staged": {
    "*.md": [
      "scripts/generateId.js",
      "git add" // if auto-fix
    ]
  }
}
```

harish-sethuraman:
Yea we could do that but consider someone force merges the markdowns without heading ID. we will not be able to generate headings for that later on since it won't be detected in other's commits (until they work on same file). 

If we validate headings IDs only for staged files we should add a CI check for the PR to ensure we don't miss headings. (That will be added in all forked translation versions).

Should we validate headings only for staged files? and add a CI check ? Or rather would it make sense to validate heading IDs for all markdowns pre commit?

smikitky:
@harish-sethuraman IMHO, ideally, it's best to have both a CI that checks everything and a pre-commit hook that checks (and optionally fixes) only the staged files. On second thought, however, at the moment, even prettier and eslint run on the entire codebase before every commit, so it may not make much sense to introduce lint-staged only around this topic...

harish-sethuraman:
Yep we need to have checks consistent. We have been running ESlint and prettier for the whole repo even in the old site. 
If we want to introduce lint staged we can move all checks to lint staged at the same moment. Now shall we have checks run for all files as before?

PS: wanna try adding validate heading id check for CI ?

@gaearon your thoughts ? Shall we move all checks to lint staged ?

smikitky:
@harish-sethuraman I think we can just add `docs:validate-ids` to this line:

https://github.com/reactjs/reactjs.org/blob/f625b86a48309a430b90d92d233d93c010de62da/beta/package.json#L17

The hardest remaining task is getting `validateHeadingIDs.js` to accept file names as arguments.

harish-sethuraman:
>Perhaps the hardest remaining task is getting validateHeadingIDs.js to accept file names as arguments.

Yep we somehow have to figure out a way to validate headings only for staged files.


>I think we can just add the new docs:validate-ids command to this line:

Added to CI check

smikitky:
Whatever the policy will be, it would be very helpful to unify the scripts and having the API consistent with ESLint:

```bash
# checks all files and causes an error (process.exit(1)) if an ID is missing
# (currently validateHeadingIds)
lint-heading-ids.js

# checks all files and adds IDs if missing
# (currently generateHeadingIds)
lint-heading-ids.js --fix

# checks the specified files and causes an error if an ID is missing
# (used by hand or by lint-staged)
lint-heading-ids.js fileA.md path/to/fileB.md

# fix the specified files
# (used by hand or by lint-staged)
lint-heading-ids.js --fix fileA.md path/to/fileB.md
```

harish-sethuraman:
@smikitky  
 `yarn lint-heading-ids` --> Checks all files and causes an error if heading ID is missing
 `yarn lint-heading-ids --fix` --> Fixes all markdown file's heading IDs
 `yarn lint-heading-ids path/to/markdown.md` --> Checks that particular file for missing heading ID 
 (path can be a directory or particular file)
 `yarn lint-heading-ids --fix path/to/markdown.md` --> Fixes that particular file's markdown IDs 
 (path can be a directory or particular file)
 
 Is this okay? These changes might help when we try to use lint staged as well

smikitky:
@harish-sethuraman Yes the API looks perfect to me!

beta/.husky/pre-commit
```diff
@@ -2,7 +2,6 @@
 . "$(dirname "$0")/_/husky.sh"
 
 cd beta
-# yarn generate-ids
-# git add -u src/pages/**/*.md
+yarn lint-heading-ids
```
gaearon, line 5: Why not fix them instead? Other scripts in this file fix things.

beta/scripts/headingIDHelpers/validateHeadingIDs.js
```diff
@@ -0,0 +1,68 @@
...
+function validateHeaderId(line) {
+  if (!line.startsWith('#')) {
+    return;
+  }
+
+  const match = /\{\/\*(.*?)\*\/}/.exec(line);
+  const id = match;
+  if (!id) {
+    console.error(
+      'Uh Oh seems like the markdown misses headings. Can you run yarn fix-headings to generate headings.'
```
gaearon, line 21: Let's just do "Run yarn fix-headings to generate headings." :)

gaearon (approved):
seems fine, a few nits

beta/.husky/pre-commit
harish-sethuraman: - https://github.com/reactjs/reactjs.org/issues/4135#issuecomment-983729301

We did not want to fix the heading ids since other files without heading are modified while doing translations. Rather we wanted to just throw error so that we dont miss it in the main repo?

beta/scripts/headingIDHelpers/validateHeadingIDs.js
harish-sethuraman: Done

PR #4146, opened 2021-12-04
fix: correct the link to navigate to the team page

yyyang1996:
The original link is incorrect.
image


harish-sethuraman:
Thanks.

PR #4154, opened 2021-12-07
update reducers deep dive title for clarity

seanyusa:
Assuming I'm reading this section correctly, it's about where the name "reducer" comes from. This change makes that slightly clearer, as opposed to potentially referring to how a reducer function is called/invoked.

Before:
Screen Shot 2021-12-07 at 10 17 36 AM

After:
Screen Shot 2021-12-07 at 10 17 44 AM


harish-sethuraman:
I feel like the word `they` seems a bit vague here. 
If we name it `why are they called reducers` what does the word they refer to? (Consider if someone just skims by and see's this heading)
but the heading `why are reducers called this way?` seems complete? 💭

PR #4161, opened 2021-12-10
Sandpack: upgrade dependencies and adds ReactDevtools

danilowoz:
@bvaughn sorry you weren't able to upgrade Sandpack in https://github.com/reactjs/reactjs.org/pull/4159, so I created this PR to address those and a minor break change in the CodeSandbox fork button. So, this PR adds:
- Upgrade all codemirror dependencies, which the `Unrecognized extension value in extension set ([object Object])`, I think it's due to some inconsistency in the dependencies from codemirror;
- Refactor the Fork button: the previous implementation might cause an error on forking for long sandboxes, more details and migration guide here (https://github.com/codesandbox/sandpack/releases/tag/v0.9.12);
- Adds ReactDevTools, which looks stunning!

Screenshot 2021-12-10 at 10 06 19

**Blockers**
- It seems that it's not possible to render more than one devtool per page:
Screenshot 2021-12-10 at 09 58 00
 
- Something went wrong with the line height in the Code editor. I still need to investigate whether it's a Codemirror bug or a bug that we introduced in Sandpack (temp fixed ✅):
Screenshot 2021-12-10 at 10 09 42


harish-sethuraman:
Small observation. Previously in the sandpack `cmd + right arrow` or `cmd + left arrow` would take the cursor to the line starting and ending. It seems missing in the PR preview. Any idea why?

bvaughn:
Wowowow thank you! 🙇🏼 This is great! Let me pick it up and take a look.

bvaughn:
> It seems that it's not possible to render more than one DevTool per page.

This is something I can take a look at!

beta/src/components/MDX/Sandpack/CustomPreset.tsx
```diff
@@ -104,6 +106,8 @@ export function CustomPreset({
               </button>
             )}
           </div>
+
+          <SandpackReactDevTools />
```
bvaughn, line 110: I think we'll want to make this conditional. (Show DevTools for some sandboxes, but not every one.)

```diff
diff --git a/beta/src/components/MDX/Sandpack/CustomPreset.tsx b/beta/src/components/MDX/Sandpack/CustomPreset.tsx
index 7050dcb6..acf0cb77 100644
--- a/beta/src/components/MDX/Sandpack/CustomPreset.tsx
+++ b/beta/src/components/MDX/Sandpack/CustomPreset.tsx
@@ -22,9 +22,11 @@ import {CustomTheme} from './Themes';
 export function CustomPreset({
   isSingleFile,
   onReset,
+  showDevTools,
 }: {
   isSingleFile: boolean;
   onReset: () => void;
+  showDevTools: boolean;
 }) {
   const lineCountRef = React.useRef<{[key: string]: number}>({});
   const containerRef = React.useRef<HTMLDivElement>(null);
@@ -107,7 +109,7 @@ export function CustomPreset({
             )}
           </div>
 
-          <SandpackReactDevTools />
+          {showDevTools && <SandpackReactDevTools />}
         </SandpackThemeProvider>
       </div>
     </>
diff --git a/beta/src/components/MDX/Sandpack/index.tsx b/beta/src/components/MDX/Sandpack/index.tsx
index 3cfa4ebd..374a6d51 100644
--- a/beta/src/components/MDX/Sandpack/index.tsx
+++ b/beta/src/components/MDX/Sandpack/index.tsx
@@ -15,6 +15,7 @@ type SandpackProps = {
   children: React.ReactChildren;
   autorun?: boolean;
   setup?: SandpackSetup;
+  showDevTools?: boolean;
 };
 
 const sandboxStyle = `
@@ -64,7 +65,7 @@ ul {
 `.trim();
 
 function Sandpack(props: SandpackProps) {
-  let {children, setup, autorun = true} = props;
+  let {children, setup, autorun = true, showDevTools = false} = props;
   let [resetKey, setResetKey] = React.useState(0);
   let codeSnippets = React.Children.toArray(children) as React.ReactElement[];
   let isSingleFile = true;
@@ -144,6 +145,7 @@ function Sandpack(props: SandpackProps) {
           onReset={() => {
             setResetKey((k) => k + 1);
           }}
+          showDevTools={showDevTools}
         />
       </SandpackProvider>
     </div>
```

bvaughn: Tried to just push this small change myself but I don't have permissions to push to your fork :D

bvaughn:
>> It seems that it's not possible to render more than one DevTool per page.
> 
> This is something I can take a look at!

I think we'll have to solve this by passing through some params to DevTools to make sure it gets wired up to only its sibling sandbox. Going to dig a little deeper this afternoon and come up with a proposal.

This will likely require a small change to the `SandpackReactDevTools` component to allow pass-through props to the inner DevTools instance, but I'm not sure yet what the exact props will need to be.

bvaughn:
Status update: I need to update `react-devtools-inline` to support multiple DevTools on a single page- basically passing in custom "walls" that enable each DevTools to only communicate with its specific paired app (and not every app on the page). Once this is done, we'll need to make a small change to Sandpack and how it uses/configures the DevTools instances. (I can help with this.)

I got a good start on this and will try to wrap it up on Monday!

gaearon:
- Should this be on all sandboxes?
- Should it appear on touch devices / narrow viewports? Sandboxes already take a lot of vertical scroll space there.
- Cmd+left/right inside editor is broken as @harish-sethuraman noted.

Let's not merge until we've addressed this. (Would be good to chat with @lebo about the design)

bvaughn:
> * Should this be on all sandboxes?

No. It's off by default (after my commit).

> * Should it appear on touch devices / narrow viewports? Sandboxes already take a lot of vertical scroll space there.

Don't know, although DevTools works well on touch devices.

> * Cmd+left/right inside editor is broken as @harish-sethuraman noted.
Let's not merge until we've addressed this. (Would be good to chat with @lebo about the design)

Agreed. No hurry to merge this. Need to solve the multi-DevTools-per-page issue first too.

danilowoz:
> Should this be on all sandboxes?

danilowoz:
> Should this be on all sandboxes?
I would put this component behind an editor option (like a tick or something) because once the user has learned to use devtool, it should be available to use all the time. However, I would keep it hidden by default and only turn it on for specific sandboxes (those related to the devtool, for example). 

> Cmd+left/right inside editor is broken...
I'll address it. It seems something is no longer working on the Codemirror side.

bvaughn:
Once https://github.com/facebook/react/pull/22949 lands, we should be able to publish a new DevTools release and then update this PR to fix _this_ error:
image

bvaughn:
The `react-devtools-inline` 4.22.0 release fixes this issue:
image

_But_ we'll need to update the Code Sandbox client (backend) and Sandpack (frontend) to share a wall as shown in the README here: https://github.com/facebook/react/blob/main/packages/react-devtools-inline/README.md#advanced-integration-with-custom-wall

danilowoz:
All issues related to the Codemirror has been fixed:
- Cmd+left/right inside editor is broken...
- Gutter / line number height miscalculation; 

Could someone also confirm it?

harish-sethuraman:
> Cmd+left/right inside editor is broken...

This issues is  resolved. ✅

Small find here:

Seems like the scroll is hidden and appears when we scroll in the main branch. But in the branch preview the scroll always appears. (I don't think this should be an issue but let's make sure it is noted).

| In main branch  | In branch preview |
| ------------- | ------------- |
|  Screenshot 2021-12-15 at 9 16 19 PM | Screenshot 2021-12-15 at 9 15 58 PM |  


This is how it looks in mobile(small [320px]). You can see how the scroll changes make it visible

| In main branch  | In branch preview |
| ------------- | ------------- |
|  Screenshot 2021-12-15 at 9 24 44 PM  |  Screenshot 2021-12-15 at 9 25 07 PM  |

PS: I'll fix the fork button overflowing issue separately

bvaughn:
Circling back here. Anything I can do to help? :)

lebo:
makes sense to me, I just have a few css nits: 
- can we conditionally adjust the bottom left/bottom right border radii to be 0 when there is a ReactDevtools instance attached to the sandbox?
Screen Shot 2022-01-04 at 1 20 13 PM
- can we add a border radii to the bottom of the dev tools or add an `overflow:hidden` so it doesn't stack on top of the container border?
Screen Shot 2022-01-04 at 1 22 17 PM
- do we have control over the colors being used? we're now mixing a lot of different grays together, would be great to align the devtools unit with the existing sandbox colors if possible.

danilowoz:
| Dark | Light |
| - | - |
| Screenshot 2022-01-05 at 15 05 05 | Screenshot 2022-01-05 at 15 05 14 |

**Wrapping up:**
- Upgrade Sandpack to an experimental channel that uses the latest version of `react-devtools-inline`;
- Update React dependencies to use experimental version too, as this is a dependency from `react-devtools-inline` package - @bvaughn @gaearon please let me know if it's fine;
- Turn ReactDevTools component theme sensitive (dark/light);
- Small UI tweaks to accommodate the new devtool component (what @lebo mentioned above);
- Use the same color variables from syntax-highlight to the devtools keywords (for example, match the component name and the property name/value CSS colors);
- Refactor `reset` trigger to a non-destructive mode (shift to use the Sandpack API), meaning that it no longer re-render the entire sandbox as before, but reset only the files change to the original version;
- Took the freedom to introduce a new **Inspect** button, which render the React devtools component - of course, we can revert it as long as we came with a better approach;

Looking forward to merging it, let me know what is missing here! 🎉

beta/src/components/MDX/APIAnatomy.tsx
```diff
@@ -60,11 +59,7 @@ export function APIAnatomy({children}: APIAnatomyProps) {
           break;
         case 'pre':
           acc.code = (
-            <CodeBlock
-              ref={ref}
-              {...child.props.children.props}
-              noMargin={true}
-            />
+            <CodeBlock {...child.props.children.props} noMargin={true} />
```
danilowoz, line 62: Removed `ref` as this hasn't been used anywhere

beta/src/components/MDX/CodeBlock/CodeBlock.tsx
```diff
@@ -21,22 +20,19 @@ interface InlineHiglight {
...
-  }: {
-    children: string;
-    className?: string;
-    metastring: string;
-    noMargin?: boolean;
-    noMarkers?: boolean;
-  },
-  ref?: React.Ref<HTMLDivElement>
-) {
+const CodeBlock = function CodeBlock({
```
danilowoz, line 23: Only removed `forwardRef`

beta/src/components/MDX/Sandpack/NavigationBar.tsx
```diff
@@ -41,6 +48,11 @@ export function NavigationBar({
     return;
   }, [openPaths.length, resizeHandler]);
 
+  const onReset = () => {
```
danilowoz, line 51: Shift the approach to use the Sandpack API

beta/src/components/MDX/Sandpack/OpenInCodeSandboxButton.tsx
```diff
@@ -3,29 +3,21 @@
  */
 
 import * as React from 'react';
-import {useCodeSandboxLink} from '@codesandbox/sandpack-react';
+import {UnstyledOpenInCodeSandboxButton} from '@codesandbox/sandpack-react';
```
danilowoz, line 6: Just a minor Sandpack break change that fixed a bug for long sandboxes

beta/package.json
```diff
@@ -34,9 +34,9 @@
     "github-slugger": "^1.3.0",
     "next": "^12.0.3-canary.2",
     "parse-numeric-range": "^1.2.0",
-    "react": "18.0.0-alpha-930c9e7ee-20211015",
+    "react": "experimental",
     "react-collapsed": "3.1.0",
-    "react-dom": "18.0.0-alpha-930c9e7ee-20211015",
+    "react-dom": "experimental",
```
bvaughn, line 39: I think this is fine 👍🏼  at least until the caching APIs become stable.

bvaughn:
This is exciting! 🥳 

Noticed two small things while testing:

1. I'm not sure I like the theme override. It seems a tad buggy in some case (like the modal overlay background). For example, open the settings modal and change to light mode.
Screen Shot 2022-01-05 at 10 34 22 AM

2. There's a keyboard navigation bug 😃 but I bet we'll need to fix that on the DevTools side. Essentially the more DevTools you open on a page, the more times you have to hit the up or down arrow to change the selected element 😆 

I think updating the react/react-dom dependencies to be experimental makes sense and is okay though 👍🏼

harish-sethuraman:
The preview seems to load again when the inspect is closed and reopened 🙁  is this expected? 

https://user-images.githubusercontent.com/32865581/148254821-c0bb4d83-43cd-42b1-8f5c-c7626a7887bb.mov


previously the overflow didn't break the mobile experience but now it breaks. How do we expect the experience to be in mobile?  

Screenshot 2022-01-05 at 10 11 23 PM

danilowoz:
> I'm not sure I like the theme override. 

@bvaughn agreed, I can see your concern about hijacking the styles like that. However, I could find another decent way to turn this component theme sensitive (dark/light) and also match the Sandpack colors theme. I think that as long as we override all the necessary CSS variables, I don't think we will have many future problems keeping it updated. Can you see any problem with this approach? 


> How do we expect the experience to be in mobile?

Thanks, @harish-sethuraman, for pointing me out. @lebo what do you think about this suggestion:
Screenshot 2022-01-06 at 10 00 52

> The preview seems to load again when the inspect is closed and reopened

The first time, yes: because CSB bundler needs to load the `react-devtools-inline` dependencies in the background. But for the second time, it's totally unnecessary, as all the modules are already there, so I did a fix to improve this flow. Thanks for letting me know about it.

bvaughn:
> However, I could find another decent way to turn this component theme sensitive (dark/light) and also match the Sandpack colors theme.

I believe you meant to say "_couldn't find_"?

DevTools "auto" setting should pick up on the user's light/dark theme automatically. Does Sandpack not do this? (Is it always dark mode?)

I guess DevTools _could_ expose a parameter to force a certain theme (e.g. "dark") to be the only one users can choose if that would be useful. Is the concern that even the "dark" theme would just clash too badly with Sandpack?

(Okay. I just checked for myself and– you're right. It clashes.)

image

> I think that as long as we override all the necessary CSS variables, I don't think we will have many future problems keeping it updated. Can you see any problem with this approach?

My concern is partly about future updates. I don't consider style changes (new CSS vars, changed colors, etc) as part of the semver contract for `react-devtools-inline` because these aren't intended to be overridden (and we do add new ones somewhat often). Seems like the sort of coupling that might cause headaches in the future.

How easy would it be for us to update things on the Sandpack side when DevTools makes this sort of change? What would that process look like? 😄  Would it just be a matter of us updating `beta/src/styles/sandpack.css`? (That seems pretty reasonable.)

danilowoz:
> DevTools "auto" setting should pick up on the user's light/dark theme automatically. Does Sandpack not do this? (Is it always dark mode?)

Yes, `SandpackReactDevTools` has a prop to set a theme. However, when the user updates the website light/dark theme it needs to re-render the DevTools component and refresh the Sandpack preview in order to remount it with the latest theme. In my opinion, It's a pretty expensive operation just for a theme. Or is there a way to programmatically update the `react-devtools-inline` theme? 

But this still doesn't solve the problem that Dan mentioned about matching the colors between the DevTools and Sandpack. So I think we still have somehow to override the colors.

**Updating the Devtools**
So yes, I think it's doable to continue with this approach and keep in mind that any CSS change on DevTools might affect this current implementation, but IMO it's something we can deal with.

So here's the step to update the `react-devtools-inline`: 
- Backend: https://github.com/codesandbox/codesandbox-client/blob/master/packages/app/package.json#L198
- Frontend: https://github.com/codesandbox/sandpack/blob/experimental/sandpack-react/package.json#L53
- And the styles here on `beta/src/styles/sandpack.css`

bvaughn:
> So here's the step to update the `react-devtools-inline`: 
> - Backend: https://github.com/codesandbox/codesandbox-client/blob/master/packages/app/package.json#L198
> - Frontend: https://github.com/codesandbox/sandpack/blob/experimental/sandpack-react/package.json#L53
> - And the styles here on `beta/src/styles/sandpack.css`

Okay. I'm okay with this then, given the trade offs.

Whenever someone updates DevTools for Sandpack, they'll just need to step through the UI and do a visual check. (Or we could even add instructions somewhere to diff the DevTools CSS variables in GitHub to see if any have changed).

Then yeah it should be easy enough to update `beta/src/styles/sandpack.css` to add new ones.

beta/src/pages/learn/index.md
```diff
@@ -23,7 +23,7 @@ Welcome to the React documentation! Here is an overview of what you can find on
 
 This is a tiny React app. To get your first taste of React, **edit the code below** and make it display your name:
 
-<Sandpack>
+<Sandpack showDevTools>
```
bvaughn, line 26: This was good for demonstrating the new functionality but let's remove it before landing so that nothing changes as the result of this PR. Then we (@gaearon, @rachelnabors, and I) can add it where it makes sense in follow up PRs.

bvaughn (commented):
Is this PR okay to land (in terms of no user-facing changes) and then iterate further on? Or are there still open questions or things we need to address first?

danilowoz:
Hey @bvaughn, from my side, everything is done. I just pushed a new commit removing the demo, and it should be ready to go!

gaearon:
I still see an Inspect button on every sandbox which, when clicked, refreshes the sandbox.

danilowoz:
@gaearon just removed it too, let me know if anything else is missing.

bvaughn:
At this point, I defer to @gaearon to give the final stamp to land this. (Unless you want me to make that call, Dan.) I can't recall any blocking issues on the DevTools side in terms of features that we need to add before we can start making use of this in the new docs (@danilowoz– remind me or link me if I'm forgetting anything).

beta/src/components/MDX/Sandpack/index.tsx
```diff
@@ -121,29 +124,17 @@ function Sandpack(props: SandpackProps) {
     hidden: true,
   };
 
-  let key = String(resetKey);
```
gaearon, line 124: Why was this removed? Is this not needed anymore? We use it in development so that editing sandbox source immediately forces it to refresh. This is important for authoring content.

beta/src/components/MDX/Sandpack/NavigationBar.tsx
gaearon: I recall trying `resetAllFiles` and it wasn't enough for some reason. Maybe I was missing `refresh` though.

gaearon:
I'm not sure why bot comment doesn't work. But this seems to have increased the page weight a bit (+7 KB minified). Is this because of the DevTools feature specifically, or was it a general sandbox code increase?

https://github.com/reactjs/reactjs.org/runs/4839615866?check_suite_focus=true#step:10:187

I'd like, if possible, to hold the line since we're already not doing great on code size.

beta/src/components/MDX/Sandpack/index.tsx
danilowoz: Sorry if I'm missing something, but in my local tests, both situations worked fine (development and production mode). The main reason I removed the whole reset logic is that this key was causing an unnecessary render on production mode, and Sandpack already provides an API to reset all changes.

However, even in development mode, any changes on `files` shouldn't be enough to trigger a remount on the Sandpack component? At least, in the Sandpack implementation, we're taking this into account, but maybe I'm missing some specific case. 

I brought back the `key` logic anyway, but I added a new conditional not to remount it on production mode, let me know if it makes sense.

danilowoz:
@gaearon I tracked down and only the devtools dependency has increased the bundle in ~4.8k, the rest might be the CSS added in this PR and other minor changes in the Sandpack codebase. 

Although I agree that we need to keep in mind the final page weight, I'd like to highlight that the latest Sandpack version introduced many bug fixes and performance improvements (some of which we've already spoken about before). So given the tradeoff, I believe it's still worth merging and addressing the devtool dependency issue in a different PR (lazily loading it might help, as I'm doing here https://github.com/reactjs/reactjs.org/pull/4112).

beta/src/components/MDX/Sandpack/OpenInCodeSandboxButton.tsx
```diff
@@ -3,29 +3,21 @@
...
       title="Open in CodeSandbox">
-      <span className="hidden md:inline">
-        <IconNewPage className="inline mb-0.5 text-base" /> Fork
-      </span>
-      <span className="inline md:hidden">
-        <IconNewPage className="inline mb-0.5 text-base" /> Fork
-      </span>
-    </a>
+      <IconNewPage
+        className="inline mb-0.5 ml-1 mr-1 relative"
```
harish-sethuraman, line 15:
```suggestion
        className="inline mb-0.5 ml-1 mr-1 relative top-px"
```

beta/src/components/MDX/Sandpack/OpenInCodeSandboxButton.tsx
```diff
@@ -3,29 +3,21 @@
...
+        width=".8em"
+        height=".8em"
+        style={{top: 1}}
```
harish-sethuraman, line 18:
```suggestion
```

beta/src/components/Icon/IconNewPage.tsx
```diff
@@ -5,14 +5,15 @@
 import * as React from 'react';
 
 export const IconNewPage = React.memo<JSX.IntrinsicElements['svg']>(
-  function IconNewPage({className}) {
+  function IconNewPage({className, ...rest}) {
```
harish-sethuraman, line 8:
```suggestion
  function IconNewPage(props) {
```
Shall we spread the props completely? Instead of destructuring the className from props and then spreading the rest and then spreading the rest values in line 16 ?

beta/src/components/MDX/Sandpack/OpenInCodeSandboxButton.tsx
danilowoz: Wow, thanks @harish-sethuraman, I spent some time figuring out how I could do this.

harish-sethuraman: :)

beta/src/components/MDX/Sandpack/index.tsx
```diff
@@ -133,16 +135,25 @@ function Sandpack(props: SandpackProps) {
...
+        autorun={autorun}
+        key={key}>
         <CustomPreset
           isSingleFile={isSingleFile}
+          showDevTools={showDevTools}
+          onDevToolsLoad={() => setDevToolsLoaded(true)}
+          devToolsLoaded={devToolsLoaded}
           onReset={() => {
-            setResetKey((k) => k + 1);
+            /**
```
gaearon, line 150: Hmm actually this doesn't seem necessary anymore? If I remove this onReset and even the key generation logic, it still works in development and replaces the sandbox when MDX content is edited. So we can remove this after all?

gaearon:
This PR introduces a flash in the loading sequence:

https://user-images.githubusercontent.com/810438/150172175-43a0f0e0-3107-481d-8218-7cbcb2badde7.mov

For a moment, the entire code disappears:

Screenshot 2022-01-19 at 16 23 25

There's also a weird double border in the middle.

Compare it to the loading sequence on the live site:

https://user-images.githubusercontent.com/810438/150172588-1de79ca0-cbfd-4dcf-a0d7-a6eaef005fa5.mov

Let's make sure that we test this when doing future upgrades because we really don't want to regress like this.

gaearon (changes requested):
need to fix a regression in the initial loading sequence

beta/src/components/MDX/Sandpack/index.tsx
gaearon: > but in my local tests, both situations worked fine (development and production mode)

Yeah this looks like it didn't work out of the box in the beginning of 2021, but has since been fixed!

danilowoz:
- ✅ **For a moment, the entire code disappears:** indeed it was a regression due to some changes in the CodeMirror initialization. I shifted the approach of the placeholder component replacement and it looks as before;
- ✅ **Double border**: there was a custom scrollbar style that was causing this "border", but we no longer have any default styles for scrollbars customization - so it should be fixed;
- ✅/⚠️ **About the `onReset` & `key` thing:** I did many tests and all of them reset the entire Sandpack properly, because any changes on`files` should trigger a new render internally on Sandpack. However, I noted that `CodeBlock` wasn't updating when MDX content is edited (I couldn't understand why, as there is no significant difference between the `SandpackCodeEditor` and `SandpackCodeViewer`), so I added a `key` there to ensure it'll re-render on every content changes (will investigate why this's happening).

PR #4162, opened 2021-12-10
[Beta] Fix the post layout dynamically

justin-tse:
Fix the path: [ official React blog](https://beta.reactjs.org/blog)
We will see the page layout is wrong when the window width is not enough. 
image
Because the width is fixed by the container. 
So I change the style to something similar to the other page(LayoutLearn).


gaearon:
That page isn't really styled yet (that's why we don't link to it). I think we were planning to remove the sidebar altogether there. So I wouldn't touch it for now.

justin-tse:
Yeah, I know your meaning. I just found it is not style, so I want it to be more beautiful.  Thank you for your response.❤️

PR #4165, opened 2021-12-10
add crypto.randomUUID as source of keys

o-t-w:


gaearon:
seems premature until Safari adds support

o-t-w:
@gaearon Safari 15.4, which was released yesterday, has support for this feature.

harish-sethuraman (approved):

gaearon:
We should probably add an extra clarification that you can only use it when _creating_ items (like "add new todo"), not when _rendering_.

o-t-w:
@gaearon
It does already say "when creating items".
And then immediately below it already says "Don’t generate them while rendering." 
We could add it to this line of the pitfall section:
Similarly, do not generate keys on the fly, e.g. with `key={Math.random()}` or `key={crypto.randomUUID()}`  

But even so, for beginners it is not obvious enough. A code sample of showing in which scenario you _do_ use it would avoid any confusion.

PR #4169, opened 2021-12-13
Switch eslint-plugin-prettier link to eslint-config-prettier

eai04191:
Switch from eslint-plugin-prettier to eslint-config-prettier as a way to make ESLint and Prettier coexist.

According to the [Prettier documentation](https://prettier.io/docs/en/integrating-with-linters.html), the use of eslint-plugin-prettier is generally not recommended.

It should be fine to use eslint-config-prettier as a way to resolve the conflict between ESLint and Pretter.


import-brain (approved):

eai04191:
A change with the same purpose as this PR was already brought by #4636

PR #4171, opened 2021-12-13
Fix typo in Removing from an array part

ahmetfarukyilmaz:


harish-sethuraman (approved):

harish-sethuraman:
Changes:
Screenshot 2021-12-14 at 11 57 10 AM
Code and snippet have same value but the explanation has different value. With this PR value is same.

harish-sethuraman:
Thanks 😄

PR #4176, opened 2021-12-17
Fix: fixed out of bound of array

DaYePython:
🐛 In the "Chapter 2 overview: Adding interactivity", If you press "next"13 times, the index is out of bounds of "sculptureList".


harish-sethuraman:
Thanks for the PR. We had the same change in this PR too. 

>Thanks but it’s okay — people like to fix this issue themselves. :) we also have a challenge for this on one of the pages.


https://github.com/reactjs/reactjs.org/pull/4122#issuecomment-979090483

PR #4178, opened 2021-12-17
docs: fix formatting inconsistencies

stuymedova:


import-brain (approved):

gaearon:
Regarding the order, it's somewhat intentional. The goal is to show what's important in the initial viewport. So we try to use this depending on the example.

PR #4188, opened 2021-12-18
Update introducing-jsx.md

emrah-abdurahman:
This is a proposed change to the  element code. The original includes a closing </img> tag but the img HTML element is a self-closing empty element.

The original is: const element = </img>;
My proposed change is: const element = ;

P.S. Thank you so much for the awesome guides! I am learning a lot from it 👍🏻

### DIFF
--- a/content/docs/introducing-jsx.md
+++ b/content/docs/introducing-jsx.md
@@ -98,7 +98,7 @@ const element = <div tabIndex="0"></div>;
 You may also use curly braces to embed a JavaScript expression in an attribute:
 
 ```js
-const element = <img src={user.avatarUrl}></img>;
+const element = <img src={user.avatarUrl} />;
 ```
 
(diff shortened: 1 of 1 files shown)


harish-sethuraman:
Seems like a valid ask.
Even MDN specifies `Tag omission | Must have a start tag and must not have an end tag.` 
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#technical_summary

@gaearon your  thoughts?

import-brain (approved):
LGTM!

harish-sethuraman (approved):

harish-sethuraman:
The syntax is in JSX and not HTML. Seems like in HTML you might not need a closing tag and self-close it. But in JSX there is no technical difference between a self-closing tag and a closing tag.

PR #4189, opened 2021-12-19
Try concurrent features of nextjs 

harish-sethuraman:
Tried to see if it works in deployment.


harish-sethuraman:
Tried deleting the blogs folder since I got the following error
`Error: The Edge Function "blog/2015/02/18/react-conf-roundup-2015" is 1mb which exceeds the maximum size limit of 1mb. Learn More: https://vercel.link/edge-function-size` was thrown

Now not sure why this is throw 🤔 . But the build gets completed and the analyze bundle seems to have been completed.
`Error: Builder returned invalid middleware: should NOT be longer than 1000000 characters`

Will close this for now.

PR #4192, opened 2021-12-20
ci: Fix wrong workflow file name

smikitky:
<s>(This may be broken because I have no means to test this locally)</s>

I believe the current CI setup for the beta site is broken because the workflow file refers to an old file name when it tries to download `bundle_analysis.json` from its base branch (normally `main`). This PR fixes this.

The base file currently used by the CI comes from the same old artifact (RunID: 1429732265, Artifact: 111304536) that happens to exist only in the main repository.

image

Currently, all CI runs are failing on translation forks because there is no previous run associated with `bundle_analysis_upload.yml`. I suspect the same problem will surface soon also in the main repository when the avobe-mentioned artifact expires after 90 days.

CC @gaearon

### DIFF
--- a/.github/workflows/analyze.yml
+++ b/.github/workflows/analyze.yml
@@ -54,7 +54,7 @@ jobs:
         uses: dawidd6/action-download-artifact@v2
         if: success() && github.event.number
         with:
-          workflow: bundle_analysis_upload.yml
+          workflow: analyze.yml
           branch: ${{ github.event.pull_request.base.ref }}
           name: bundle_analysis.json
           path: beta/.next/analyze/base/bundle


smikitky:
Okay, I confirmed the CI for *this* PR uses a different artifact, so it seems to be working fine.

image

harish-sethuraman:
> The base file currently used by the CI comes from the same old artifact (RunID: 1429732265, Artifact: 111304536) that happens to exist only in the main repository.

Good find - Also I found that the artifact seems to be present in that run ID with this API (https://api.github.com/repos/reactjs/reactjs.org/actions/runs/1429732265/artifacts)

I guess it tries to take the latest artifact with that name which was created on 6.11.2021.


>  I suspect the same problem will surface soon also in the main repository when the avobe-mentioned artifact expires after 90 days. 

This would not have happened since the workflow will expire on `2022-02-04T18:52:20Z`

Workflow in this PR shows runId 1596411952 
(https://api.github.com/repos/reactjs/reactjs.org/actions/runs/1596411952/artifacts)
and the created_at is `2021-12-18T16:44:04Z`.
This was the last PR merge to main branch I guess. So now it pulls the latest artifact.

harish-sethuraman (approved):

harish-sethuraman:
Thanks 😄

PR #4195, opened 2021-12-21
missing class name added

abhay27chauhan:


BartoszKlonowski:
Hello @abhay27chauhan!
For the sake and simplicity of explanation sometimes it's not necessary to name all the classes or components, moreover sometimes it's better to avoid so there's no misleading of any kind.
I will close this PR, considering the fact that I couldn't find such part in the latest of the docs.
Thank you!

PR #4198, opened 2021-12-27
Checking the Square before (expensive) calculation

bolajiniy:
**Code optimization**

Checking that the square is filled/clicked first is less expensive than calculating the winner before checking the square. Since, this awesome document is read by young and senior engineers it would be better if we encourage best practice.

### DIFF
--- a/content/tutorial/tutorial.md
+++ b/content/tutorial/tutorial.md
@@ -740,7 +740,7 @@ We can now change the Board's `handleClick` function to return early by ignoring
 ```javascript{3-5}
   handleClick(i) {
     const squares = this.state.squares.slice();
-    if (calculateWinner(squares) || squares[i]) {
+    if (squares[i] || calculateWinner(squares)) {
       return;
     }
     squares[i] = this.state.xIsNext ? 'X' : 'O';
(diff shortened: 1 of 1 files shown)


import-brain (approved):

content/tutorial/tutorial.md
```diff
@@ -740,7 +740,7 @@ We can now change the Board's `handleClick` function to return early by ignoring
 ```javascript{3-5}
   handleClick(i) {
-    if (calculateWinner(squares) || squares[i]) {
+    if (squares[i] || calculateWinner(squares)) {
```
harish-sethuraman, line 743: It would be inconsistent if we change it in the docs alone. There are multiple codepen links where we might need to change as well?.

bolajiniy: Sure, should be changed in the `codepen` also. Just that I don't think that is managed in this repo. I would if I have access to the online codes, I guess the admins need to do that.

harish-sethuraman: Changing in codepen might require Dan's help. Let's wait for his thoughts on this change as well.

PR #4199, opened 2021-12-27
Update react-router examples to use v6 version

jonatasoc:
The exemple used to demonstrate Code-Splitting using react-router-dom is outdated. My suggestion is to use the new way of calling routes that was introduced in React Router v6


harish-sethuraman (approved):
LGTM

PR #4200, opened 2021-12-29
Improve example function for functional updates

Aman-Godara:
I failed to understand the necessity of functional updates (a feature of useState hook) in this [example](https://reactjs.org/docs/hooks-reference.html#functional-updates), therefore I decided to raise this PR.

If there is any gap in my understanding of the react docs' example, please enlighten me.

### DIFF
--- a/content/docs/hooks-reference.md
+++ b/content/docs/hooks-reference.md
@@ -61,32 +61,36 @@ function Counter({initialCount}) {
     setCount((prevCount) => prevCount + x);
   }
 
+  function multiplyBy(x) {
+    setCount((prevCount) => prevCount * x);
+  }
+
   return (
     <>
       Count: {count}
       <button onClick={() => setCount(initialCount)}>Reset</button>
       <button
         onClick={() => {
-          incrementBy(1);
-          setCount((prevCount) => prevCount / 2);
+          incrementBy(2);
+          multiplyBy(2);
         }}
       >
-        Increment by 1 and then divide by 2
+        update to (count + 2) * 2
       </button>
       <button
         onClick={() => {
-          incrementBy(2);
-          setCount((prevCount) => prevCount / 3);
+          multiplyBy(4);
+          incrementBy(1);
         }}
       >
-        Increment by 2 and then divide by 3
+        update to (count * 4) + 1
       </button>
     </>
   );
 }
 ```
 
-The "+" and "-" buttons use the functional form, because the updated value is based on the previous value. But the "Reset" button uses the normal form, because it always sets the count back to the initial value.
+The two functions use the functional form, because the updated value is based on the previous value. But the "Reset" button uses the normal form, because it always sets the count back to the initial value.
 
 If your update function returns the exact same value as the current state, the subsequent rerender will be skipped completely.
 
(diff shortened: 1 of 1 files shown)


content/docs/hooks-reference.md
```diff
@@ -56,18 +56,41 @@ If the new state is computed using the previous state, you can pass a function t
...
+
+  function multiplyBy(x) {
+    setCount((prevCount) => prevCount * x);
+  }
+
   return (
     <>
       Count: {count}
       <button onClick={() => setCount(initialCount)}>Reset</button>
-      <button onClick={() => setCount(prevCount => prevCount - 1)}>-</button>
```
harish-sethuraman, line 63: The `setCount` here takes an arrow function as param (`functional form`) since the state update depends on the previous state value. Whereas the reset action does not depend on the previous value and it is going to reset the state value so we do not use the functional form here.


`If the new state is computed using the previous state, you can pass a function to setState. The function will receive the previous value, and return an updated value. `
https://reactjs.org/docs/hooks-reference.html#functional-updates

harish-sethuraman:
The already present example seems relevant so closing this PR. Thanks :) .

Aman-Godara:
@harish-sethuraman if normal non-functional form is used instead of functional form in the example given in docs will it make any difference?

harish-sethuraman:
For that example I think you can still use the normal form and it should work fine. But when there are multiple state updates (for same state value) in the same onClick it might not work. It is better to write in functional form while the update depends on the previous value. 

There are some examples in this [link](https://stackoverflow.com/questions/57828368/why-react-usestate-with-functional-update-form-is-needed)

This should work fine
```js
export default function App({ initialCount = 0 }) {
  const [count, setCount] = useState(initialCount);
  return (
    <>
      Count: {count}
      <button onClick={() => setCount(initialCount)}>Reset</button>
      <button
        onClick={() => {
          setCount((prevCount) => prevCount + 1);
          console.log("updating state again");
          setCount((prevCount) => prevCount + 1);
        }}
      >
        +
      </button>
    </>
  );
}
```

And this wouldn't
```js
import { useState } from "react";

export default function App({ initialCount = 0 }) {
  const [count, setCount] = useState(initialCount);
  return (
    <>
      Count: {count}
      <button onClick={() => setCount(initialCount)}>Reset</button>
      <button
        onClick={() => {
          setCount(count + 1);
          console.log("updating state again");
          setCount(count + 1);
        }}
      >
        +
      </button>
    </>
  );
}

```

Aman-Godara:
Agreed, but this still doesn't make sense as to why functional update is necessary here.
I can directly do
````js
setCount(count + 2)
````
to get the desired result.
What are your thoughts on this?

Aim of the PR was to make examples more realistic while keeping them simple to understand (just like other examples of the docs).

harish-sethuraman:
> Agreed, but this still doesn't make sense as to why the functional update is necessary here.

I feel the same with the example you have shared. It adds the value to the state and then updates it and again multiplies and updates it while it can be done at the same time? 

Also, the example reflects the context of the section on when one should use functional form. showing a state update being dependent on the previous state value. I feel like it would be pretty overkill to add a bigger code snippet in that place.

PR #4207, opened 2022-01-04
Add Link to React Conf 2021 Playlist

sbaljeet13:

### DIFF
--- a/content/community/videos.md
+++ b/content/community/videos.md
@@ -10,6 +10,11 @@ redirect_from:
 
 Videos dedicated to the discussion of React and the React ecosystem.
 
+### React Conf 2021 {#react-conf-2021}
+
+A playlist of videos from React Conf 2021.
+<iframe title="React Conf 2021" width="650" height="366" src="https://www.youtube-nocookie.com/embed/videoseries?list=PLNG_1j3cPCaZZ7etkzWA7JfdmKWT0pMsa" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
+
 ### React Conf 2019 {#react-conf-2019}
 
 A playlist of videos from React Conf 2019.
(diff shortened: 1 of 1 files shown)


content/community/videos.md
```diff
@@ -10,6 +10,11 @@ redirect_from:
 
 Videos dedicated to the discussion of React and the React ecosystem.
 
+### React Conf 2021 {#react-conf-2021}
+
+A playlist of videos from React Conf 2021.
+<iframe title="React Conf 2021" width="650" height="366" src="https://www.youtube-nocookie.com/embed/videoseries?list=PLNG_1j3cPCaZZ7etkzWA7JfdmKWT0pMsa" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
```
harish-sethuraman, line 16: would we need so much access?(`allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture"`). I guess other videos on the same page have only `allowfullscreen`. Shall we have it consistent? What do you think?

sbaljeet13: Hi @harish-sethuraman, sorry I just copied the whole link as it is from YouTube to embed. I did notice that they were not present in other links but wasn't sure of their role. 

After reading more on their usage, I think we can definitely remove them but `picture-in-picture` could be nice to have. 

Thoughts?

harish-sethuraman: Yep we can add picture-in-picture alone then :)

sbaljeet13: Deleted the `allow` attribute from the `<iframe>` for React Conf 2021 YouTube playlist. It should be consistent with `<iframe>` tags for other videos. Thanks for reviewing and your input on this!

harish-sethuraman (approved):

harish-sethuraman:
Thanks 😄

PR #4209, opened 2022-01-05
docs: fix: spread syntax is not an operator. But react document saying that…

wndudqus:
"…" spread syntax is not an "operator" but document saying that is an "operator" so it can make developers confused

see more at [here](https://stackoverflow.com/questions/44934828/is-it-spread-syntax-or-the-spread-operator)

and [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax)

I just changed a few words "operator" to "syntax"

스크린샷 2022-01-05 오후 9 13 14

### DIFF
--- a/beta/src/pages/learn/updating-objects-in-state.md
+++ b/beta/src/pages/learn/updating-objects-in-state.md
@@ -1123,7 +1123,7 @@ select { margin-bottom: 10px; }
 
 The problem was in the mutation inside `handleMove`. It mutated `shape.position`, but that's the same object that `initialPosition` points at. This is why both the shape and the background move. (It's a mutation, so the change doesn't reflect on the screen until an unrelated update--the color change--triggers a re-render.)
 
-The fix is to remove the mutation from `handleMove`, and use the spread operator to copy the shape. Note that `+=` is a mutation, so you need to rewrite it to use a regular `+` operation.
+The fix is to remove the mutation from `handleMove`, and use the spread syntax to copy the shape. Note that `+=` is a mutation, so you need to rewrite it to use a regular `+` operation.
 
 <Sandpack>
 
--- a/content/docs/jsx-in-depth.md
+++ b/content/docs/jsx-in-depth.md
@@ -231,7 +231,7 @@ In general, we don't recommend *not* passing a value for a prop, because it can 
 
 ### Spread Attributes {#spread-attributes}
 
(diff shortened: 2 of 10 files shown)


content/docs/jsx-in-depth.md
```diff
@@ -244,7 +244,7 @@ function App2() {
 }
 ```
 
-You can also pick specific props that your component will consume while passing all other props using the spread operator.
+You can also pick specific props that your component will consume while passing all other props using the spread syntax.
```
harish-sethuraman, line 247: There are other places with same error like 

>The array spread operator also lets you prepend an item by placing it *before* the original `...artists`:

Link to the line:
https://github.com/reactjs/reactjs.org/blob/b41b1dc35679c01c3252e7d512ce28c5e100d0a4/beta/src/pages/learn/updating-arrays-in-state.md?plain=1#L146

Lets fix other places too and make it consistent?

beta/src/pages/learn/updating-objects-in-state.md
```diff
@@ -17,7 +17,7 @@ State can hold any kind of JavaScript value, including objects. But you shouldn'
 
 </YouWillLearn>
 
-## What's a mutation? {/*whats-a-mutation*/}
+## What's a mutation? {/_whats-a-mutation_/}
```
harish-sethuraman, line 20: How did these heading id's change? We do not generate ids on pre commit anymore 🤔. This is affecting multiple files. Can we restore it to older syntax of IDs ? 
```diff
+{/*whats-a-mutation*/}
-{/_whats-a-mutation_/}
```

beta/src/pages/learn/updating-objects-in-state.md
```diff
@@ -118,101 +125,108 @@ onPointerMove={e => {
 
 With `setPosition`, you're telling React:
 
-* Replace `position` with this new object
-* And render this component again
+- Replace `position` with this new object
```
harish-sethuraman, line 128: Seems like all `*` got replaced with `-` as well 😕.

content/docs/jsx-in-depth.md
wndudqus: thanks! I changed all "spread operator" to "spread syntax"

beta/src/pages/learn/updating-objects-in-state.md
wndudqus: Yes, I think my formeter is applied. I have restored all changes.

thanks!

harish-sethuraman (approved):

PR #4213, opened 2022-01-09
fix: 1st example from chapter 2 on the learn page

victordantasdev:
I've found this bug on the 1st example from chapter 2 on the learning page, when user press the next button more than 12 times this error happens: 

img1
img2


harish-sethuraman:
Thanks but it’s okay — people like to fix this issue themselves. We even have a challenge for this later.

https://beta.reactjs.org/learn/state-a-components-memory page's first challenge is based on this.

PR #4214, opened 2022-01-10
Fix bug in keyed fragment example

katiejenn:
I tried this out in my editor and without wrapping the Fragment in a return statement, the Fragment never renders in the DOM

### DIFF
--- a/content/docs/fragments.md
+++ b/content/docs/fragments.md
@@ -123,10 +123,12 @@ function Glossary(props) {
     <dl>
       {props.items.map(item => (
         // Without the `key`, React will fire a key warning
-        <React.Fragment key={item.id}>
-          <dt>{item.term}</dt>
-          <dd>{item.description}</dd>
-        </React.Fragment>
+        return (
+          <React.Fragment key={item.id}>
+            <dt>{item.term}</dt>
+            <dd>{item.description}</dd>
+          </React.Fragment>
+         )
       ))}
     </dl>
   );


content/docs/fragments.md
```diff
@@ -123,10 +123,12 @@ function Glossary(props) {
     <dl>
       {props.items.map(item => (
         // Without the `key`, React will fire a key warning
-        <React.Fragment key={item.id}>
-          <dt>{item.term}</dt>
-          <dd>{item.description}</dd>
-        </React.Fragment>
+        return (
```
harish-sethuraman, line 126: This should work without return 

```js
props.items.map(item => (
```
It is returned in the callback itself

wndudqus:
@katiejenn 
if you want to contribute in this repo you need to sign up for CLA agreement [check here](https://code.facebook.com/cla)

gaearon:
this won't work. you probably wrote `item => {` rather than `item => (` like in the example.

gaearon:
specifically, this works:

```js
      {props.items.map(item => ( // <--- parenthesis
        <React.Fragment key={item.id}>
```

this also works:

```js
      {props.items.map(item => { // <--- curly brace
        return (
          <React.Fragment key={item.id}>
```

but this (as in this PR) won't work:

```js
      {props.items.map(item => (
        return (
          <React.Fragment key={item.id}>
```

i hope this helps! sorry for the confusion.

PR #4224, opened 2022-01-20
Analytics for Beta

rachelnabors:
Originally we shipped without analytics for the beta docs, but we need to know how much traffic they're getting, from what sources, etc. Oversight of the original PR!


beta/.env.development
```diff
@@ -1,2 +1,3 @@
 NEXT_PUBLIC_HJ_SITE_ID = 2411683
-NEXT_PUBLIC_HJ_SITE_V = 6
\ No newline at end of file
+NEXT_PUBLIC_HJ_SITE_V = 6
+NEXT_PUBLIC_GA_TRACKING_ID = 'G-3918VS75Y6'
```
harish-sethuraman, line 3: Since we are going to use the same tracking ID in production and development we shall rather have it n the `next.config.js` alone?

gaearon:
This looks to me like it will only track page loads but not subsequent in-app navigations. How did you verify that it works?

I would expect that there needs to be an effect tracking page views. https://stackoverflow.com/a/62552263

gaearon (changes requested):
We need to add tracking for in-app navigations, I think

rachelnabors:
I thought so, too, but when I added them, GA's realtime dashboard started registering multiple clicks.

## What I did

* open GA's real time dashboard 
* click between navigation items and watch the page views tick up

## What next

We should also probably remove these from the local environment, but I left them in in case further testing is required!

beta/.env.development
rachelnabors: We should probably remove it from the local env after we're sure it's working as anticipated so we aren't tracking development work traffic

gaearon:
ok, so the reason this works is because GA has a new feature called "Enhanced measurement" which we have on.

Screenshot 2022-01-21 at 18 49 22

this feature overrides `history.pushState`:

Screenshot 2022-01-21 at 18 50 38

and this is how it tracks the pageviews:

Screenshot 2022-01-21 at 18 50 47

so seems fine

PR #4233, opened 2022-01-22
Removed google analytics for dev

harish-sethuraman:
We might not need to track development information? 

We shall merge after we confirm that GA is working properly 
- https://github.com/reactjs/reactjs.org/pull/4224#discussion_r789882998


beta/src/pages/_app.tsx
```diff
@@ -30,20 +30,24 @@ export default function MyApp({Component, pageProps}: AppProps) {
...
     <AppShell>
       <Component {...pageProps} />
-      <Script
-        strategy="lazyOnload"
-        src={`https://www.googletagmanager.com/gtag/js?id=${process.env.NEXT_PUBLIC_GA_TRACKING_ID}`}
-      />
-      <Script id="google-analytics" strategy="lazyOnload">
-        {`
+      {process.env.NODE_ENV === 'production' && (
+        <>
```
gaearon, line 34: no need for a fragment here

harish-sethuraman: There are two script tags? That is why I added fragment 🤔.

gaearon: ahh I missed it. Ok then!

harish-sethuraman: :)

PR #4236, opened 2022-01-25
Removed circle CI

harish-sethuraman:
Seems like this runs only `ci-check` which is already run by another git workflow. We might not need this now.


gaearon:
i don't see this workflow running on main. can we add it to main first?

harish-sethuraman:
> ci/circleci: Build Error — Your tests failed on CircleCI
This is the workflow name and you can see it fails in this PR since I removed it

PR #4240, opened 2022-01-26
Introduce lint staged 

harish-sethuraman:
Going forward we might consider linting only git staged files. Prettier was not configured for markdown and css. So added them in the pre commit lint staged. 
This PR includes:
- Adding `lint-staged` to lint staged files
- Including prettier configs for `CSS` and `MDX`
- We might generate headings only for staged files 🎉 
- Prettified CSS files
- Prettified MDX files - This actually breaks build not sure why...

### DIFF
--- a/beta/package.json
+++ b/beta/package.json
@@ -101,6 +101,6 @@
   "lint-staged": {
     "*.{js,ts,jsx,tsx}": "yarn prettier",
     "*.css": "prettier --write",
-    "*.md": "prettier --write"
+    "*.md": ["yarn fix-headings" , "prettier --write"]
   }
 }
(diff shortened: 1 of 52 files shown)


beta/package.json
```diff
@@ -96,5 +97,10 @@
     "budget": null,
     "budgetPercentIncreaseRed": 10,
     "showDetails": true
+  },
+  "lint-staged": {
+    "*.{js,ts,jsx,tsx}": "yarn prettier",
+    "*.css": "prettier --write",
```
gaearon, line 103: Why two different commands for prettier for CSS and everything else?

harish-sethuraman: Changed

PR #4245, opened 2022-01-27
[Beta] Rewrite Quick Start

gaearon:
The existing Quick Start is more of a "site overview". There was a bunch of negative feedback on it and it clearly didn't work out. This is a clean rewrite. The goal is to cover the basics but without distracting the reader with links or too many caveats. It needs to cover enough to prepare you to read Thinking in React, but be short enough that you wouldn't bail on it right away.

I hope this one will work.

**[Page preview: Quick Start](https://beta-reactjs-org-git-qs-fbopensource.vercel.app/learn)**

### DIFF
--- a/beta/src/pages/learn/index.md
+++ b/beta/src/pages/learn/index.md
@@ -4,38 +4,68 @@ title: Quick Start
 
 <Intro>
 
-Welcome to the React documentation! Here is an overview of what you can find on this site.
+Welcome to the React documentation! This page will give you an introduction to the 80% of React concepts that you will use on a daily basis.
 
 </Intro>
 
 <YouWillLearn>
 
-* [How to set up React](/learn/installation)
-* [How to think about building user interfaces with React](/learn/thinking-in-react)
-* [How to get something rendering on the screen](/learn/describing-the-ui)
-* [How to make your user interface respond to interactions](/learn/adding-interactivity)
-* [How to keep your logic maintainable as your app grows](/learn/managing-state)
-* [How to step outside React when necessary](/learn/escape-hatches)
+- How to define components
+- How to add markup and styles
+- How to display data
+- How to render conditions and lists
+- How to respond to events and update the screen
+- How to share data between components
 
 </YouWillLearn>
 
-## Introduction {/*introduction*/}
+## Creating and nesting components {/*components*/}
 
-This is a tiny React app. To get your first taste of React, **edit the code below** and make it display your name:
+React apps are made out of components. A component is a piece of the UI (user interface) that has its own logic and appearance. A component can be as small as a button, or as large as an entire page.
+
+React components are JavaScript functions that return markup:
+
+```js
+function MyButton() {
+  return (
+    <button>Click me</button>
+  );
+}
+```
+
+Now that you've declared `MyButton`, you can nest it into another component:
+
+```js {5}
+export default function MyApp() {
+  return (
+    <div>
+      <h1>Welcome to my app</h1>
+      <MyButton />
+    </div>
+  );
+}
+```
+
+Notice that `<MyButton />` starts with a capital letter. That's how you know it's a React component. React component names must always start with a capital letter, while HTML tags must be lowercase.
+
+Have a look at the result:
 
 <Sandpack>
 
 ```js
-function Greeting({ name }) {
-  return <h1>Hello, {name}!</h1>;
+function MyButton() {
+  return (
+    <button>
+      Click me
+    </button>
+  );
 }
 
-export default function App() {
+export default function MyApp() {
   return (
     <div>
-      <Greeting name="Divyesh" />
-      <Greeting name="Sarah" />
-      <Greeting name="Taylor" />
+      <h1>Welcome to my app</h1>
+      <MyButton />
     </div>
   );
 }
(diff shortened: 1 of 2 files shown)


beta/src/pages/learn/index.md
```diff
@@ -4,716 +4,486 @@ title: Quick Start
...
-When you're ready to start a project, there are several options. You can write React code online and save your work in an environment like [CodeSandbox](https://react.new/). You can add React as a `<script>` tag to any HTML file to try it for a small part of your existing webpage. Or you can create a whole new React app, with or without a framework. **Use the [Installation](/learn/installation) page to pick what works best for you--but you can skip it for now.**
-
-</DeepDive>
+The `export default` keywords specify the main component in the file. If you're not familiar with some piece of JavaScript syntax, [MDN](https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/export) and [javascript.info](https://javascript.info/import-export) have great references.
 
-## Learn React {/*learn-react*/}
+## Writing markup with JSX {/*writing-markup-with-jsx*/}
 
-There are a few ways to get started:
+The markup syntax you've seen above is called JSX. It is optional, but most React projects use JSX for its convenience. Browsers don't understand JSX natively, but you can [set up a tranform locally](/learn/installation) when you're ready.
```
markerikson, line 80: Typo:

```suggestion
The markup syntax you've seen above is called JSX. It is optional, but most React projects use JSX for its convenience. Browsers don't understand JSX natively, but you can [set up a transform locally](/learn/installation) when you're ready.
```

Also not sure if people reading this are going to know what "transform" means in this context. 

Would a different phrasing work here?

> , so JSX is [typically transformed into plain JS by a project build tool]() so that browsers can run that code.

beta/src/pages/learn/index.md
```diff
@@ -4,716 +4,486 @@ title: Quick Start
...
+```
 
-Read **[Describing the UI](/learn/describing-the-ui)** to learn how to make things appear on the screen, including declaring components, importing them, writing JSX with the curly braces, and writing conditions and lists.
+If you're unfamiliar with this JavaScript syntax, you can start by always using `if...else`.
 
-</LearnMore>
+## Rendering lists {/*rendering-lists*/}
 
-### Chapter 2 overview: Adding interactivity {/*chapter-2-overview-adding-interactivity*/}
+You will rely on JavaScript features like [`for` loop](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for) and the [array `map()` function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) to render lists of components.
```
markerikson, line 226:
```suggestion
You will rely on JavaScript features like [`for` loops](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for) and the [array `map()` function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) to render lists of components.
```

Even worth mentioning `for` loops here? I can't think of the last time I've used one for list rendering.

markerikson (changes requested):
Looks pretty good overall! Caught a couple typos, and a couple questions on phrasing.

beta/src/pages/learn/index.md
```diff
@@ -4,716 +4,486 @@ title: Quick Start
...
-* [How to keep your logic maintainable as your app grows](/learn/managing-state)
-* [How to step outside React when necessary](/learn/escape-hatches)
+- How to define components
+- How to add markup and styles
+- How to display data
+- How to render conditions and lists
+- How to respond to events and update the screen
+- How to share data between components
 
 </YouWillLearn>
```
taniarascia, line 21: It kind of feels like a brief one or two sentence overview is missing here, like how all these pieces (components, JSX, state) all fit together.

beta/src/pages/learn/index.md
```diff
@@ -4,716 +4,486 @@ title: Quick Start
...
+  </li>
+);
+
+return (
+  <ul>{listItems}</ul>
+);
 ```
 
-Notice how clicking the buttons updates the screen:
+Notice how `<li>` has a `key` attribute. A key helps React figure out what happened when list items are inserted, removed, or reordered. Usually you should use a key from your data, such as a database ID. Keys should be unique among items in the same parent tag. Although you can technically specify index as a key, your list won't work correctly when items are inserted, removed, or reordered. Always try to use ID as a key if you can.
```
taniarascia, line 252: > Although you can technically specify index as a key, your list won't work correctly when items are inserted, removed, or reordered. Always try to use ID as a key if you can.

Would probably rewrite this as "Do not use an index as a key, because..." since I don't think there's ever a good reason to do so, even if they technically can.

beta/src/pages/learn/index.md
```diff
@@ -4,716 +4,486 @@ title: Quick Start
...
 
-In the above example, all the data was written directly in markup. However, you'll often want to keep it separately. Here, the data is kept in an array. In React, you use JavaScript functions like [`map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) to [render lists](/learn/rendering-lists) of things.
+## Conditional rendering {/*conditional-rendering*/}
 
-<Sandpack>
-
-```js App.js
-import { people } from './data.js';
-import Profile from './Profile.js';
+In React, there is no special syntax for writing conditions. Instead, you'll use the same techniques as you use when writing regular JavaScript code. To render a component conditionally, you can use [`if`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/if...else) statement:
```
taniarascia, line 188: Might be worth mentioning in some way that `if` statements have to be used before a `return`? I guess it's mentioned earlier with the "dynamic expressions in JSX curlies" comment, but it's a common rookie confusion. I know it's one of those "trying not to give too many side details" types of things.

beta/src/pages/learn/index.md
```diff
@@ -4,716 +4,486 @@ title: Quick Start
...
 }
 ```
 
-```css
-label { display: block; }
-input { margin-left: 5px; margin-bottom: 5px; }
-```
-
-</Sandpack>
+Notice how `onClick={handleClick}` has no parentheses at the end. You don't need to _call_ your event handler, you need to _pass_ the event handler itself. React will call your handler when the user clicks the button.
```
taniarascia, line 301: Same here, one of the biggest rookie mistakes is people calling a handler in an `onClick` event, I think if there was some obvious way of highlighting that.

"Do not call the event handler", rather than "You don't need to", because actually you NEED to not do this.

taniarascia (commented):
Looks really nice and clean! Left a few comments. I think the overview can be expanded just a little bit to tie it all together.

beta/src/pages/learn/index.md
markerikson: I don't think it's possible to over-emphasize this point enough!

(does the `<blink>` tag work inside Markdown? :)  )

atomiks:
## Content

A big improvement overall in my opinion, much easier for someone new to read and understand. I think it's still missing something though: the "light bulb" moment that shows why React is so useful and how it all fits together. "Why" is everyone using this library, basically. This page sort of just lists all React's features and how to use them, but not why you should be doing this. Maybe this goes in a different section?

Others have given good feedback already for the rest. Though, I noticed in the "Displaying data" section, the Sandpack example after "You can put any dynamic expressions..." shows a bunch of code that doesn't seem to be used or relevant in the example, like the "if" statement. Perhaps show briefly how conditionals change the rendered output? 

## Styling

This isn't related to the content, so it'd be a separate PR. But I wanted to give high level feedback of "experiencing" the preview that's linked since the experience all fits together.

- Line highlighting is odd, it should span the entire code block. It looks like the `.sp-stack` forces the width to fit the content.
- The code editor/Sandpack examples have no proper indication they can be scrolled for devices without a scrollbar by default. I also think the max-height set on it could be higher (or adaptable to the screen height) so you can see more code at once. It feels limiting to be in a tiny window, e.g. to see two components at once, making the examples a bit harder to comprehend.

beta/src/pages/learn/index.md
```diff
@@ -4,716 +4,486 @@ title: Quick Start
 
 <Intro>
 
-Welcome to the React documentation! Here is an overview of what you can find on this site.
+Welcome to the React documentation! This page will give you an introduction to the 80% of React concepts that you will use on a daily basis.
```
harish-sethuraman, line 7:
```suggestion
Welcome to the React documentation! This page will give you an introduction to 80% of the React concepts that you will use on a daily basis.
```

NataliaTepluhina:
First of all: great work on updating the quick start. It's easy to read and it's easy to understand the main concepts. I like the section about `useState` a lot, especially the "evolution" of the count from isolated state to shared state. It's one of the common "real-life" tasks so it's nice to see it covered in the documentation.

A few little notes that came to mind while reading the guide:

- as @atomiks already mentioned, the snippet on `Displaying data` should be either simplified or we'd need to add some visual changes when `large` class is applied;
- for the complete outsider who never touched React, the example of `useState` feels a bit too magical. So, we call `useState`, we want to have a state property called `x`, and it provides us with a function to update a state property with `set-` prefix. It would be nice to mention somewhere that if `count` was `avocado`, we'd have `setAvocado` instead 😅

mungojam:
It could be neat if the first playground that the user sees was visible without scrolling down. Basically using it as a teaser somehow. Or otherwise just having a picture of the button next to the very first bit of code that you show, just to give real beginner people an instant feel of this is what you can do easily.

rickhanlonii (approved):
Looks great! Lots of good feedback here already so I don't have anything to add. I think we can ship and iterate from here!

beta/src/pages/learn/index.md
gaearon: I don't think it's that simple in practice. I do regularly bump into cases where index is the right key. If we push people towards "index always bad" this can be na issue, e.g. they can use random keys which are even worse. I'll try to rephrase this to be more concise though.

beta/src/pages/learn/index.md
gaearon: Yeah this is tricky because on some level we have to assume the reader knows JS. If they mess up the syntax, hopefully they can copy and/or use references.

beta/src/pages/learn/index.md
gaearon: Yeah I'll try to make this a bit clearer.

gaearon:
> I think it's still missing something though: the "light bulb" moment that shows why React is so useful and how it all fits together. "Why" is everyone using this library, basically. This page sort of just lists all React's features and how to use them, but not why you should be doing this. Maybe this goes in a different section?

Maybe I can add something short to the end. I'd like to have a page like this but I'm not sure it's this one.

>Others have given good feedback already for the rest. Though, I noticed in the "Displaying data" section, the Sandpack example after "You can put any dynamic expressions..." shows a bunch of code that doesn't seem to be used or relevant in the example, like the "if" statement. Perhaps show briefly how conditionals change the rendered output?

The code is there to make the `classes.join(' ')` call relevant (and also to answer one of the most common questions about how to do multiple CSS classes). But it's fair this isn't really clear if you don't already understand it. I'll try to move this.

>Line highlighting is odd, it should span the entire code block. It looks like the .sp-stack forces the width to fit the content.

As you mentioned, this isn't really relevant to this specific page. I remember fixing this and causing some other issue. But we should probably revisit this at some point. I'd like to have a fix for this.

> It feels limiting to be in a tiny window, e.g. to see two components at once, making the examples a bit harder to comprehend.

There's a "Show more" button on each sandbox that expands it. I realize it's not super noticeable but hopefully once the reader sees it's there, this is less of a problem.

> So, we call useState, we want to have a state property called x, and it provides us with a function to update a state property with set- prefix. It would be nice to mention somewhere that if count was avocado, we'd have setAvocado instead

👍 

>It could be neat if the first playground that the user sees was visible without scrolling down. Basically using it as a teaser somehow

That's what the existing quick start does. I've heard feedback that the example isn't helping and makes people feel overloaded. At this point I'm not sure which feedback to take into account, but I wanted to try the inverse, and then maybe we'll find some middle ground.

beta/src/pages/learn/index.md
gaearon: I don't disagree but not sure what to write. I'll get this in as is and revisit when we flesh out related pages and think about landing etc.

beta/src/pages/learn/index.md
gaearon: I think it's worth mentioning just to dispel the notion there's something special about map() here.

beta/src/pages/learn/index.md
```diff
@@ -4,716 +4,491 @@ title: Quick Start
...
 </YouWillLearn>
 
-## Introduction {/*introduction*/}
+## Creating and nesting components {/*components*/}
 
-This is a tiny React app. To get your first taste of React, **edit the code below** and make it display your name:
+React apps are made out of components. A component is a piece of the UI (user interface) that has its own logic and appearance. A component can be as small as a button, or as large as an entire page.
 
-<Sandpack>
+React components are JavaScript functions that return markup:
```
sylwiavargas, line 26: **problem**: "Markup" is not a term commonly used in tech curricula. Also, if you google "markup programming", you get an umbrella-term explanation. 

**possible solutions**:
- explain what markup is,
- link to a definition,
- find a synonym

sylwiavargas:
I love the new QuickStart! I also agree with the feedback already given. I'd like to add a few nits from a potential beginner's perspective ✨

beta/src/pages/learn/index.md
```diff
@@ -4,716 +4,491 @@ title: Quick Start
...
+```js {3}
+return (
+  <h1>
+    {user.name}
+  </h1>
+);
 ```
 
-</Sandpack>
+You can also "escape into JavaScript" from JSX attributes, but you have to use curlies *instead of* quotes. For example, `className="avatar"` passes the `"avatar"` string as the CSS class, but `src={user.imageUrl}` reads the JavaScript `user.imageUrl` variable value, and then passes that value as the `src` attribute:
```
sylwiavargas, line 128: **problem**: "curlies" is cute but may be confusing for non-English speakers

**possible solution**: 
- add the curly symbols, like so: "but you have to use curlies (`{` and `}`)" or: "but you have to use curlies (`{}`)"

beta/src/pages/learn/index.md
```diff
@@ -4,716 +4,491 @@ title: Quick Start
...
+return (
+  <img
+    className="avatar"
+    src={user.imageUrl}
+  />
+);
+```
 
-Just like you can pass some information to the browser `<img>` tag, you can also pass information to your own components like `<Profile>`. Such information is called [_props_](/learn/passing-props-to-a-component). Here, three `<Profile>`s receive different props:
+You can put any dynamic expressions inside JSX curlies, including string concatenation and object literals:
```
sylwiavargas, line 139: **problem**: "string concatenation" and "object literals" are not straightforward beginner-friendly or ESL-friendly terms.

**solution**: let's link them up to a resource, maybe?
- [string concatenation](https://javascript.info/operators#string-concatenation-with-binary)
- [object literals](https://javascript.info/object#literals-and-properties)

beta/src/pages/learn/index.md
gaearon: One thing I've been trying to avoid here is having too many links so that people don't get distracted with rabbit holes. "Markup" seems commonly known as in we assume the knowledge of HTML, which is a "markup language", and HTML tags are commonly referred to as "markup".

It _is_ a bit intentionally vague since the "markup" we refer to is actually JSX and not HTML. (But we haven't introduced JSX yet.) This is why we don't say "functions that return HTML".

Ultimately I don't think this one is an issue because even if you don't know what markup is, the code example serves to define it. ("Aha, `<button>Click me</button>` is markup"). But if you have alternative wordings (I prefer not to introduce new links here) I'm open to it.

beta/src/pages/learn/index.md
```diff
@@ -4,716 +4,491 @@ title: Quick Start
...
-  id: 3,
-  name: 'Percy Lavon Julian',
-  imageId: 'IOjWm71'
-}, {
-  id: 4,
-  name: 'Subrahmanyan Chandrasekhar',
-  imageId: 'lrWQx8l'
-}];
-```
+If you prefer terser code, you can use the [conditional `?` operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator). Unlike `if`, it works inside JSX:
```
sylwiavargas, line 200: **problem**: "terser" is not Simple English

**solution**: try "shorter", "more compact"?

beta/src/pages/learn/index.md
```diff
@@ -4,716 +4,491 @@ title: Quick Start
...
 
-<LearnMore path="/learn/describing-the-ui">
+```js
+<div>
+  {isLoggedIn && <AdminPanel />}
+</div>
+```
 
-Read **[Describing the UI](/learn/describing-the-ui)** to learn how to make things appear on the screen, including declaring components, importing them, writing JSX with the curly braces, and writing conditions and lists.
+All of the same approaches work for conditionally specifying attributes. If you're unfamiliar with some of this JavaScript syntax, you can start by always using `if...else`.
```
sylwiavargas, line 220: **suggestion**: "all of these approaches" instead of "all of the same approaches"

beta/src/pages/learn/index.md
```diff
@@ -4,716 +4,491 @@ title: Quick Start
...
+button {
+  display: block;
+  margin-bottom: 5px;
+}
 ```
 
 </Sandpack>
 
-You can remove it and simplify the code by calculating `fullName` while the component is rendering:
+Functions starting with `use` are called Hooks. `useState` is a built-in Hook provided by React. You can find other built-in Hooks in the [React API reference](/reference). You can also write your own Hooks by combining the existing ones.
```
sylwiavargas, line 382: **problem**: It feels like this paragraph lacks context. 

**possible solution**: add a header ("About hooks") or place this paragraph a visually-different section (for example some "NOTE" section)

beta/src/pages/learn/index.md
```diff
@@ -4,716 +4,491 @@ title: Quick Start
...
+Hooks are more restrictive than regular functions. You can only call Hooks *at the top level* of your components (or other Hooks). If you want to `useState` in a condition or a loop, extract a new component and put it there.
 
-```js
-import { useState } from 'react';
+## Sharing data between components {/*sharing-data-between-components*/}
 
-export default function Form() {
-  const [firstName, setFirstName] = useState('');
-  const [lastName, setLastName] = useState('');
+In the above example, each counter is independent. If you want them to display the same value and update together, *move the state up* to their closest common ancestor. First, move state from `MyButton` into `MyApp`:
```
sylwiavargas, line 388: **problem**: Since it's a new section, it feels like the "above example" is missing

**possible solution**: Replace "In the above example" with "In the previous section's example" or "In the previous example"

PR #4256, opened 2022-01-30
Splitting sandpack from main bundle

harish-sethuraman:
Tried to split the sandpack and added a fallback with suspense

Adding suspense to `Sandpack` reduced `~9.55KB`
Adding a wrapper to `codeBlock` and `Sandpack` reduced `~150KB`

Fallback preview here: https://beta-reactjs-org-namamgdvj-fbopensource.vercel.app/learn


harish-sethuraman:
Sandpack (without highlighting and output) renders, Then the fallback and then the complete sandpack... 

https://user-images.githubusercontent.com/32865581/151690964-75ac2121-0817-4ff6-946e-4e2542910ec7.mov

beta/src/components/MDX/MDXComponents.tsx
```diff
@@ -3,7 +3,7 @@
  */
 
 import * as React from 'react';
-
+import {SandpackSetup} from '@codesandbox/sandpack-react';
```
gaearon, line 6: this should just be `import type`?

harish-sethuraman: Yes will change

gaearon:
Prerequisite fixes for this to work:

- [ ] https://github.com/vercel/next.js/pull/33861
- [ ] https://github.com/mdx-js/mdx/pull/1924

beta/src/components/MDX/CodeBlock/index.tsx
```diff
@@ -2,6 +2,30 @@
...
+import * as React from 'react';
+const CodeBlock = React.lazy(
+  () =>
+    import('./CodeBlock').then(
+      (x) =>
+        new Promise((resolve) => {
+          setTimeout(() => resolve(x), 3000);
+        })
+    ),
+  {
```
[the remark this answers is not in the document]
gaearon: yeah this isn't needed

gaearon:
Next thing I'm seeing is that with production builds, HTML always includes fallback rather than content. I suspect this might just be some Next.js optimization or because we haven't enabled the `concurrentFeatures` flag. Maybe it's due to static generation.

gaearon:
OK, i can confirm this is because SSG skips content and only shows fallbacks. Not sure what Next.js canonical solution is. I'll ask them.

gaearon:
I guess that doesn't really matter though because we should just make really good fallbacks with pre tags. Then this is what we want to show on SSR anyway. So that's the next step.

harish-sethuraman:
I've added fallback but actually at some point when the syntax highlighting is getting applied there is minimal shift.. 
This appears in the main branch too...

Screenshot 2022-02-01 at 7 45 38 PM

danilowoz:
Hey guys, FYI, Sandpack will be able to provide a syntax highlight component on SSR soon, which will avoid any flash between server and client, more details in this [PR](https://github.com/codesandbox/sandpack/pull/276). Otherwise, I think it makes sense to only Suspense the code-editor component (which contains the heaviest dependencies), instead of the whole Sandpack package, like https://github.com/reactjs/reactjs.org/pull/4112.

Let me know if there is anything we can do from our side to make it easier.

gaearon:
I think @danilowoz means we'd still want to lazy-load `SandpackCodeViewer`, not sure though

harish-sethuraman:
> I think @danilowoz means we'd still want to lazy-load `SandpackCodeViewer`, not sure though

We are anyway splitting `codeBlock` component that consumes `SandpackCodeViewer` right?

danilowoz:
yeah, I'd narrow it down and lazily load the following components:
- CodeBlock: https://github.com/reactjs/reactjs.org/blob/main/beta/src/components/MDX/CodeBlock/CodeBlock.tsx#L93-L97
- Sandpack: https://github.com/reactjs/reactjs.org/blob/main/beta/src/components/MDX/Sandpack/CustomPreset.tsx#L73-L81

Both contain the same dependencies and most of the heaviest parts from CodeMirror.

harish-sethuraman:
We have lazily loaded them both 😄

gaearon:
filed https://github.com/vercel/next.js/issues/33879 for why fallbacks appear in SSG

gaearon:
so the issue here was that we really shouldn't be doing this without the flag: https://github.com/vercel/next.js/issues/33879#issuecomment-1027324625

so we need to wait for https://github.com/vercel/next.js/issues/31317

gaearon:
actually https://github.com/vercel/next.js/pull/33886 would be enough

gaearon:
The padding on code blocks jumps a bit.

harish-sethuraman:
Pretty weird before the preview before I reverse merged with main has perfect fallback without any jumps in the UI. 

Before: https://beta-reactjs-org-namamgdvj-fbopensource.vercel.app/reference/usestate

Now: https://beta-reactjs-org-o28nb0eep-fbopensource.vercel.app/reference/usestate

gaearon:
maybe because we merged a sandpack bump so it's a new version

harish-sethuraman:
> maybe because we merged a sandpack bump so it's a new version

Yes, It is caused because of the sandpack bump. Verified it in local by checking out before that change

harish-sethuraman:
`v12.0.11-canary.9` with the changes we require broke the build 🤔.

harish-sethuraman:
`v12.0.11-canary.10` even this doesn't work I even tried latest canary. Only the stable tag worked. Should we report this ?

harish-sethuraman:
@danilowoz during the initial render the sandpack content is loaded but without the code formatting (it is loaded as string). Later we get the fallback and then the formatted code. Need help in finding why code is rendered as string during initial render.

PS: Build fails you can try the same change in local

wYuRiv-W

danilowoz:
@harish-sethuraman currently, it's not possible, due to some CodeMirror limitations. But Sandpack is going to support it very soon, as we made some important progress in SSR support (https://github.com/codesandbox/sandpack/pull/276).

If you want to give it a try, take a look at this PR https://github.com/danilowoz/reactjs.org/pull/1 - but, please don't use it for production, there are a few breaking change comming in this implementation.

harish-sethuraman:
This change splits sandpack and codeblock into separate chunks. So there are lot of components to test. 

- [x] Codeblocks (ex: https://beta-reactjs-org-ec9ekfllb-fbopensource.vercel.app/learn)
- [x] Code blocks with terminals  (ex: https://beta-reactjs-org-ec9ekfllb-fbopensource.vercel.app/apis)
- [x] Sandpack (ex: https://beta-reactjs-org-ec9ekfllb-fbopensource.vercel.app/learn/installation)
- [x] API anatomy (ex: https://beta-reactjs-org-ec9ekfllb-fbopensource.vercel.app/apis/usestate)

gaearon:
looks great. wanna press merge? if something breaks we'll roll back

harish-sethuraman:
🚀

PR #4259, opened 2022-01-31
using JSX syntax instead of js expressions

Ahmed-Hakeem:
using JSX syntax is more readable than js expressions, as we don't have to write js expressions because the render property just accepts a pure functional component that we can call with JSX syntax.

### DIFF
--- a/content/docs/render-props.md
+++ b/content/docs/render-props.md
@@ -182,6 +182,7 @@ class Mouse extends React.Component {
   }
 
   render() {
+    const {render : Slot} = this.props;
     return (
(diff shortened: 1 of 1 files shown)


BartoszKlonowski:
Hello @Ahmed-Hakeem!
As the part you modify in this PR is no longer in the react.dev docs I will close this PR as no longer required.
Thank you for taking the time to contribute!

Ahmed-Hakeem:
> Hello @Ahmed-Hakeem!
> 
> As the part you modify in this PR is no longer in the react.dev docs I will close this PR as no longer required.
> 
> Thank you for taking the time to contribute!

It's my pleasure. Thank you for the update.

PR #4262, opened 2022-01-31
using JSX syntax instead of js expressions

Ahmed-Hakeem:
using JSX syntax is more readable than js expressions, as we don't have to write js expressions because the render property just accepts a pure functional component that we can call with JSX syntax.


harish-sethuraman:
Duplicate of
- https://github.com/reactjs/reactjs.org/pull/4259

PR #4263, opened 2022-01-31
Update useState API Reference (beta)

sylwiavargas:
**tl;dr** This PR offers a reordered useState API reference page (beta). 

## motivation

Previously, we first learned how to use `useState` and only afterwards, what it is. Now, we start with the "what" and then we move to the "how".

## changes implemented
- The first section of the page is now the "Declaring a state variable", which explains the anatomy of `useState`
- Within that section, the anatomy items start with what we pass (the right side) and then only the ones we get (the left side)
- The anatomy items link to an updater and setter function sections of the page.

---
**SCREENSHOT**
Screen Shot 2022-01-31 at 4 13 56 PM

### DIFF
--- a/beta/src/pages/reference/usestate.md
+++ b/beta/src/pages/reference/usestate.md
@@ -20,7 +20,7 @@ You must declare state variables **at the top level of your component,** outside
 
 <AnatomyStep title="Initial state">
 
-You pass the value you want the state to be initially.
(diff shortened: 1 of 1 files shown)


gaearon (changes requested):
we're going to take a larger pass at this since the ordering of the examples doesn't quite make sense now

gaearon:
thanks for the start. we did a bigger reorder in https://github.com/reactjs/reactjs.org/pull/4265.

PR #4264, opened 2022-01-31
[Beta] Address Quick Start review nits

gaearon:
This is meant to address https://github.com/reactjs/reactjs.org/pull/4245#issuecomment-1025931701.

### DIFF
--- a/beta/src/pages/learn/index.md
+++ b/beta/src/pages/learn/index.md
@@ -217,7 +217,7 @@ When you don't need the `else` branch, you can also use a shorter [logical `&&` 
 </div>
 ```
 
-All of these approaches work for conditionally specifying attributes too. If you're unfamiliar with some of this JavaScript syntax, you can start by always using `if...else`.
+All of these approaches also work for conditionally specifying attributes. If you're unfamiliar with some of this JavaScript syntax, you can start by always using `if...else`.
 
 ## Rendering lists {/*rendering-lists*/}
 
--- a/beta/src/pages/learn/index.md
+++ b/beta/src/pages/learn/index.md
@@ -125,7 +125,7 @@ return (
 );
 ```
 
-You can also "escape into JavaScript" from JSX attributes, but you have to use curlies *instead of* quotes. For example, `className="avatar"` passes the `"avatar"` string as the CSS class, but `src={user.imageUrl}` reads the JavaScript `user.imageUrl` variable value, and then passes that value as the `src` attribute:
+You can also "escape into JavaScript" from JSX attributes, but you have to use curly braces *instead of* quotes. For example, `className="avatar"` passes the `"avatar"` string as the CSS class, but `src={user.imageUrl}` reads the JavaScript `user.imageUrl` variable value, and then passes that value as the `src` attribute:
 
 ```js {3,4}
 return (
@@ -136,7 +136,7 @@ return (
 );
 ```
 
-You can put any dynamic expressions inside JSX curlies, including string concatenation and object literals:
+You can put more complex expressions inside the JSX curly braces too, for example, [string concatenation](https://javascript.info/operators#string-concatenation-with-binary):
 
 <Sandpack>
 
@@ -177,7 +177,7 @@ export default function Profile() {
 
 </Sandpack>
 
-In the above example, `style={{ }}` is not a special syntax, but a regular object inside the JSX curlies. You can use the `style` attribute when your styles depend on JavaScript variables.
+In the above example, `style={{}}` is not a special syntax, but a regular `{}` object inside the `style={ }` JSX curly braces. You can use the `style` attribute when your styles depend on JavaScript variables.
 
 ## Conditional rendering {/*conditional-rendering*/}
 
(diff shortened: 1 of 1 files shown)


beta/src/pages/learn/index.md
```diff
@@ -379,13 +379,37 @@ button {
...
 Functions starting with `use` are called Hooks. `useState` is a built-in Hook provided by React. You can find other built-in Hooks in the [React API reference](/reference). You can also write your own Hooks by combining the existing ones.
 
 Hooks are more restrictive than regular functions. You can only call Hooks *at the top level* of your components (or other Hooks). If you want to `useState` in a condition or a loop, extract a new component and put it there.
 
 ## Sharing data between components {/*sharing-data-between-components*/}
 
-In the above example, each counter is independent. If you want them to display the same value and update together, *move the state up* to their closest common ancestor. First, move state from `MyButton` into `MyApp`:
+In the previous example, each button had its own counter. Imagine the structure of that example:
+
+```js {2-4}
```
gaearon, line 394: We'll replace this with a diagram later

harish-sethuraman (approved):

beta/src/pages/learn/index.md
```diff
@@ -217,7 +217,7 @@ When you don't need the `else` branch, you can also use a shorter [logical `&&`
 </div>
 ```
 
-All of the same approaches work for conditionally specifying attributes. If you're unfamiliar with some of this JavaScript syntax, you can start by always using `if...else`.
+All of these approaches work for conditionally specifying attributes too. If you're unfamiliar with some of this JavaScript syntax, you can start by always using `if...else`.
```
rickhanlonii, line 220:
```suggestion
All of these approaches also work for conditionally specifying attributes. If you're unfamiliar with some of this JavaScript syntax, you can start by always using `if...else`.
```

beta/src/pages/learn/index.md
```diff
@@ -379,13 +379,37 @@ button {
...
+## Using Hooks {/*using-hooks*/}
+
 Functions starting with `use` are called Hooks. `useState` is a built-in Hook provided by React. You can find other built-in Hooks in the [React API reference](/reference). You can also write your own Hooks by combining the existing ones.
 
 Hooks are more restrictive than regular functions. You can only call Hooks *at the top level* of your components (or other Hooks). If you want to `useState` in a condition or a loop, extract a new component and put it there.
 
 ## Sharing data between components {/*sharing-data-between-components*/}
 
-In the above example, each counter is independent. If you want them to display the same value and update together, *move the state up* to their closest common ancestor. First, move state from `MyButton` into `MyApp`:
+In the previous example, each button had its own counter. Imagine the structure of that example:
```
rickhanlonii, line 392: Maybe

```suggestion
In the previous example, each button had its own counter. Consider the structure of that example:
```

beta/src/pages/learn/index.md
```diff
@@ -410,7 +434,7 @@ export default function MyApp() {
 }
 ```
 
-Then, *pass down* the state and the event handler that's now inside `MyApp` to `MyButton`. You can pass information to `MyButton` using the JSX curly braces, just like you previously did with built-in tags like `<img>`:
+Then, *pass the state down* from `MyApp` to each `MyButton`, together with the shared click handler. You can pass information to `MyButton` using the JSX curly braces, just like you previously did with built-in tags like `<img>`:
```
rickhanlonii, line 437: Maybe a good opportunity to say "as props" to show how state becomes props when passed down?

```suggestion
Then, *pass the state down* from `MyApp` to each `MyButton`, together with the shared click handler, as props. You can pass information to `MyButton` using the JSX curly braces, just like you previously did with built-in tags like `<img>`:
```

rickhanlonii (approved):
LGTM ship it!

beta/src/pages/learn/index.md
rickhanlonii: Or: "These approaches also work for conditionally specifying attributes."

beta/src/pages/learn/index.md
sylwiavargas: This already looks good!

beta/src/pages/learn/index.md
```diff
@@ -379,13 +379,37 @@ button {
...
+In the previous example, each button had its own counter. Imagine the structure of that example:
+
+```js {2-4}
+- MyApp
+  - MyButton (count: 3)
+  - MyButton (count: 1)
+  - MyButton (count: 2)
+```
+
+If you want all buttons to display the same `count` and *update together* on click, you need to move the state from the individual buttons "upwards" to the closest component containing all of them. In this example, it is `MyApp`:
```
sylwiavargas, line 401: I really like this paragraph because it provides a further explanation.

beta/src/pages/learn/index.md
gaearon: We haven't introduced the words "props" yet. I wanted to wait with introducing it until we've already intuitively explained the concept. But I'm going to slightly rearrange this though.

beta/src/pages/learn/index.md
gaearon: >Because it feels more as a "btw" given that later we build the opposite.

Not sure I'm following. That _is_ what we build later.

sylwiavargas: riiiight.
( jetlag brain 🤦‍♀️ )

gaearon: I pushed a hopefully clearer explanation. The ordering was correct — the thing we're building next is the second case. should be a bit clearer now.

PR #4265, opened 2022-01-31
[Beta] Reorder useState page

gaearon:
This is inspired by https://github.com/reactjs/reactjs.org/pull/4263, but is a more expansive reordering.

The idea is that we got a bunch of feedback about the API page feeling a bit too "guide-y". This is because we start with the higher-level explanation instead of the API syntax explanation.

This flips it around. Since you're on the API page, you probably want to look up the syntax. So we start with the syntax, then "zoom out" to the usage, then "zoom out" to the recipes. I didn't change much about the special cases at the end.

This also adds clarity about whether an API "gets" or "returns" something. And we link to the function overloads from the main API "pills" so that you can jump to them if that's what you're looking for. (However, they're intentionally small and styled like footnotes.)

**[Preview page](https://beta-reactjs-org-d8to37jk8-fbopensource.vercel.app/reference/usestate)**

### DIFF
--- a/beta/src/pages/reference/usestate.md
+++ b/beta/src/pages/reference/usestate.md
@@ -4,7 +4,7 @@ title: useState()
 
 <Intro>
 
-`useState` is a React Hook that lets a component "remember" some information (called [state](/learn/state-a-components-memory)). It returns two values: the current state, and the function you can use to update it.
+The `useState` Hook lets your component ["remember" information that changes over time](/learn/state-a-components-memory) (called state). It returns two values: the current state, and the function that lets you update it.
 
 ```js
 const [state, setState] = useState(initialState);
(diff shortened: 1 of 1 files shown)


rickhanlonii (approved):
This is really good IMHO

sylwiavargas:
I agree with Rick - I love the reorder. 
However, I worry that folks won't scroll after the recipes -- I myself almost didn't. This might be because elsewhere in the docs the challenges (of which the recipes remind me) are usually located at the very bottom of the page. 

I wonder if there could be some way of giving more visibility to the content that comes after the recipes?

gaearon:
> I worry that folks won't scroll after the recipes -- I myself almost didn't.

This is partly intentional, since that deals with edge case-y things that don't come up that often. And so it's a good cutoff point for most readers. So that it doesn't feel like you have to read the entire page just to start using it.

However, I do want to try yet another reorder, which I might submit separately.

gaearon:
let's get this in and iterate.

PR #4269, opened 2022-02-01
Prefer clsx over classnames

nguyenyou:
Prefer [clsx](https://github.com/lukeed/clsx) as

> a faster & smaller drop-in replacement for the `classnames` module.

### DIFF
--- a/beta/package.json
+++ b/beta/package.json
@@ -28,7 +28,7 @@
     "@headlessui/react": "^1.3.0",
     "@mdx-js/react": "^1.6.16",
     "body-scroll-lock": "^3.1.3",
-    "classnames": "^2.2.6",
+    "clsx": "^1.1.1",
     "date-fns": "^2.16.1",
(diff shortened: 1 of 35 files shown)


harish-sethuraman:
Not much of a difference in the page sizes? I could see that  -1.76 KB was reduced in `/reference/usestate` page alone 🤔

nguyenyou:
The bundle size is reduced a little bit (not much). Beside that, the [benchmark](https://measurethat.net/benchmarks/Show/12817/0/classnames-vs-clsx) show that performance of `clsx` is a bit better than `classnames`. The implementation of `clsx` is also simpler (just two short functions in a single file) than `classnames`. 

| Test case name  | Result |
| ------------- | ------------- |
| classnames  | classnames x 2,390,894 ops/sec ±0.17% (68 runs sampled)  |
| clsx  | clsx x 3,348,060 ops/sec ±0.36% (69 runs sampled)  |

PR #4275, opened 2022-02-02
[Beta] Replace Hotjar with GA buttons

gaearon:
**This isn't styled yet.**

Replaces the feedback button at the bottom with two buttons that are hooked up to GA.

TODO:

- [x] Do the same for mobile site
- [x] Add styles
- [ ] Add footer with a link to survey (will do in a follow-up)


beta/src/components/Layout/Nav/Nav.tsx
```diff
@@ -3,6 +3,7 @@
  */
 
 import * as React from 'react';
+import {createPortal} from 'react-dom';
```
harish-sethuraman, line 6: Are we using `createPortal` here?

PR #4276, opened 2022-02-02
[Beta] Change useState structure, add examples

gaearon:
## [Page preview](https://beta-reactjs-org-1de9ee6va-fbopensource.vercel.app/reference/usestate)

This introduces a new API page structure I'm experimenting with. The idea is to avoid "solo" sandboxes in favor of "example galleries" (I'm renaming Recipes to Examples -- but there's more than one on the page now). This lets us keep each subsection dealing with a particular API shorter because we can cover a batch of them, and then do a batch of examples.

So, the new structure is:

- Common cases (things you can't live without)
  - API maps
  - Main example
  - Example gallery
- Special cases
  - API maps
  - Example gallery (this is new)
 
Then this will let us add another section like Troubleshooting that goes in a more Q&A style and then _also_ has a gallery of examples. So all of the page is broken down like Theory - Practice - Theory - Practice - Theory - Practice. If you only care to reference theory, you can quickly scan over chunks of it without getting interrupted. And giving examples more space to breathe lets us add more concrete information there and deeplink into the guides where appropriate.

The "special cases" section feels less heavy now. This is good because there's more stuff we need to add there. Notably, the "setState in render" pattern. I will be adding that separately in a follow-up.

### DIFF
--- a/beta/src/components/Layout/MarkdownPage.tsx
+++ b/beta/src/components/Layout/MarkdownPage.tsx
@@ -50,9 +50,9 @@ export function MarkdownPage<
       }
       if (child.props.mdxType === 'Recipes') {
         return {
-          url: '#recipes',
+          url: '#' + (child.props.titleId ?? 'examples'),
           depth: 0,
-          text: 'Recipes',
+          text: child.props.titleText ?? 'Examples',
         };
       }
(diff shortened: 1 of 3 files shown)


rickhanlonii (commented):
LGTM

PR #4278, opened 2022-02-03
fix(sandpack): disable init-mode

danilowoz:
The current Sandpack `initMode` implementation is leading to a race conditional in the sandpack-client, as a result, some sandpacks might break during the transpile step. So this PR removes temporarily the `initMode` configuration, while we investigate a way to fix it.

### DIFF
--- a/beta/package.json
+++ b/beta/package.json
@@ -22,7 +22,7 @@
     "check-all": "npm-run-all prettier lint:fix tsc"
   },
   "dependencies": {
-    "@codesandbox/sandpack-react": "0.13.9-experimental.6",
+    "@codesandbox/sandpack-react": "0.13.11-experimental.0",
     "@docsearch/css": "3.0.0-alpha.41",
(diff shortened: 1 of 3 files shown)


harish-sethuraman (approved):

harish-sethuraman:
>some sandpacks might break during the transpile step

Does this result in page breakage as well? 

I saw someone report page breakage not sure if these are related... 

https://twitter.com/YuriSulyma/status/1489280731616460800?s=20&t=xBdQtfq08nz3TC4BKQcYcg

danilowoz:
It's very unlikely that the transpilation process could break the main page, as it's running in another process inside iframe - but the interface between the iframe and the react application could break, but I can't see now any reason for this happening. 

Let's keep an eye on it - if it happened once, might happen again.

danilowoz:
@harish-sethuraman turned out, it was a bug on iOS 12 - addressed it [here](https://github.com/reactjs/reactjs.org/pull/4289)

PR #4281, opened 2022-02-03
feature: inform about synchronous value updates

DCzajkowski:
It is necessary for controlled component's value updates to be synchronous; otherwise, it can cause unexpected behavior like cursor "jumping" to the end.

Added a note about the need of using synchronous updates.


DCzajkowski:
Fixed in the new docs: https://react.dev/reference/react-dom/components/input#my-input-caret-jumps-to-the-beginning-on-every-keystroke

PR #4283, opened 2022-02-03
[Beta] Dry version of useState API doc

gaearon:
I'm not very into it but curious what people think. Maybe there's some middle ground between this and that?

### DIFF
--- a/beta/src/pages/reference/usestate.md
+++ b/beta/src/pages/reference/usestate.md
@@ -1,763 +1,84 @@
 ---
-title: useState()
+title: useState
 ---
 
 <Intro>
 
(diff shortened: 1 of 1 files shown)


beta/src/pages/reference/usestate.md
```diff
@@ -1,763 +1,84 @@
...
 ---
 
 <Intro>
 
-The `useState` Hook lets your component ["remember" information that changes over time](/learn/state-a-components-memory) (called state). It returns two values: the current state, and the function that lets you update it.
+`useState` is a React Hook that lets you declare a state variable.
 
 ```js
-const [state, setState] = useState(initialState);
+let [state, setState] = useState(initialState);
```
harish-sethuraman, line 10: but why let?

gaearon:
https://github.com/reactjs/reactjs.org/pull/4284

PR #4288, opened 2022-02-04
fix(sandpack): make sure iframe has always the same height and position

danilowoz:
@gaearon In order to avoid layouts shift (especially on mobile devices), I added set o measures:
- Adds a new state to make sure the current Sandpack has the latest updated height from the iframe protocol (on unmount it's out to date);
- Removes the absolute position, which was the main reason for the shift (why was it necessary?);

### DIFF
--- a/beta/src/components/MDX/Sandpack/Preview.tsx
+++ b/beta/src/components/MDX/Sandpack/Preview.tsx
@@ -35,6 +35,7 @@ export function Preview({
 }: CustomPreviewProps) {
   const {sandpack, listen} = useSandpack();
   const [isReady, setIsReady] = React.useState(false);
+  const [heightIsUpdated, setHeightIsUpdated] = React.useState(false);
   const [iframeComputedHeight, setComputedAutoHeight] = React.useState<
     number | null
   >(null);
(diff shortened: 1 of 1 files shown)


harish-sethuraman:
What if the output turns out to be smaller than the reserved area? 
Seems like the output window is resized (height decreases) and the content below that moves up.

danilowoz:
Hey, @harish-sethuraman I'm not sure I got about the problem. Could you make a screen recording? 
I shouldn't cause any problems related to the height, as the purpose of my changes is to make the loading state more reliable and avoid the preview going away off the container (causing a resizing).

gaearon:
The absolute position is for this:

https://user-images.githubusercontent.com/810438/153027467-e44cd8dd-6f2c-4de9-a656-43dc0a8d9bac.mov

gaearon (changes requested):
error screen jumps down

harish-sethuraman:
https://user-images.githubusercontent.com/32865581/153028025-e9e10806-e6db-4962-abe1-da0fffe14bd9.mov

These outputs are smaller than the initial width? and height reduces?

gaearon:
>These outputs are smaller than the initial width? and height reduces?

At least that shouldn't be an issue on main after #4322.

danilowoz:
I'm going to close this, as the layout shift has been fixed in another PR

PR #4289, opened 2022-02-04
fix(useMediaQuery): adds fallback to matchMedia

danilowoz:
It addresses a bug reported on iOS12, which breaks the entire page due to an incompatible API. So, I added a fallback (deprecated API) for those devices that are still supporting the previous API. 

Bug report


harish-sethuraman (approved):

harish-sethuraman:
Thanks :)

PR #4295, opened 2022-02-04
Removes readme and contributing.md from generate heading lint

harish-sethuraman:
Removes readme and contributing.md from generate heading

### DIFF
--- a/beta/package.json
+++ b/beta/package.json
@@ -100,6 +100,6 @@
   },
   "lint-staged": {
     "*.{js,ts,jsx,tsx,css}": "yarn prettier",
-    "!(*README|CONTRIBUTING).md": "yarn fix-headings"
+    "src/**/*.md": "yarn fix-headings"
   }
 }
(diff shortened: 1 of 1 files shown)


beta/package.json
```diff
@@ -100,6 +100,6 @@
   },
   "lint-staged": {
     "*.{js,ts,jsx,tsx,css}": "yarn prettier",
-    "*.md": "yarn fix-headings"
+    "!(*README|CONTRIBUTING).md": "yarn fix-headings"
```
gaearon, line 103: Why * before README?

gaearon: Can we just scope it down to src/**/*.md?

harish-sethuraman: We can do that :)

gaearon (approved):
have you verified the commit hook works as expected?

harish-sethuraman:
Yep I did

PR #4301, opened 2022-02-05
[beta] Don't call it "derived state"

sophiebits:
I propose that "storing information from previous renders" sounds less like a _cool thing_ and also more clearly describes the rare cases this pattern is for.


gaearon:
Love it

PR #4304, opened 2022-02-06
Generate Ids when there are none in local development

harish-sethuraman:
Seems like we weren't generating new ids at runtime so while developing in local it was generating link as undefined?
So added code to generate fallback ids when there are none.

### DIFF
--- a/beta/plugins/remark-header-custom-ids.js
+++ b/beta/plugins/remark-header-custom-ids.js
@@ -31,49 +31,25 @@ module.exports = ({
   return function transformer(tree) {
     visit(tree, 'heading', (node) => {
       const children = node.children;
-      const rawHeader = toString(node);
-      const match = /^.+(\s*\{.*\}\s*)$/u.exec(rawHeader);
-      /**
-       * we check if there is a heading and custom id with regex.
-       * If there is a match then we set the id to be the match[1] which is the custom id
-       * else we have to generate id
-       */
-      let id =
-        match && match.length === 2
-          ? match[1]
-          : slugs.slug(rawHeader, maintainCase);
-
       let tail = children[children.length - 1];
-
-      /**
-       * If there was a match then we might want to return when they dont have proper heading ids
-       * Else we generate ids
-       */
-      if (match) {
-        if (!tail || tail.type !== 'text' || tail.value !== '/}') {
-          return;
-        }
-
+      // Generate slugs on the fly (even if not specified in markdown)
+      // so that it's possible to copy anchor links in newly written content.
+      let id = slugs.slug(toString(node), maintainCase);
+      // However, for committed docs, we'll extract slug from the headers.
+      if (tail && tail.type === 'text' && tail.value === '/}') {
         tail = children[children.length - 2];
-
-        if (!tail && tail.type !== 'emphasis') {
-          return;
+        if (tail && tail.type === 'emphasis') {
+          // Use custom ID instead.
+          id = toString(tail);
+          // Until we're on MDX 2, we need to "cut off" the comment syntax.
+          tail = children[children.length - 3];
+          if (tail && tail.type === 'text' && tail.value.endsWith('{/')) {
+            // Remove the emphasis and trailing `/}`
+            children.splice(children.length - 2, 2);
+            // Remove the `{/`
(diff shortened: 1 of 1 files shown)


beta/plugins/remark-header-custom-ids.js
```diff
@@ -31,33 +31,51 @@ module.exports = ({
...
-      if (!tail || tail.type !== 'text' || tail.value !== '/}') {
-        return;
-      }
+      /**
+       * If there was a match then we might want to return when they dont have proper heading ids
+       * Else we generate ids
+       */
+      if (match) {
+        if (!tail || tail.type !== 'text' || tail.value !== '/}') {
+          return;
```
gaearon: These early exits didn't seem right. We want to assign ID from a slug regardless of whether we succeeded to modify the tail.

beta/plugins/remark-header-custom-ids.js
```diff
@@ -31,33 +31,51 @@ module.exports = ({
   return function transformer(tree) {
     visit(tree, 'heading', (node) => {
       const children = node.children;
+      const rawHeader = toString(node);
+      const match = /^.+(\s*\{.*\}\s*)$/u.exec(rawHeader);
+      /**
+       * we check if there is a heading and custom id with regex.
+       * If there is a match then we set the id to be the match[1] which is the custom id
```
gaearon: These comments are repeating the code. It's better if we use the code for things that aren't clear (like why does the code do things we want it to do).

beta/plugins/remark-header-custom-ids.js
```diff
@@ -31,33 +31,51 @@ module.exports = ({
...
+       * else we have to generate id
+       */
+      let id =
+        match && match.length === 2
+          ? match[1]
+          : slugs.slug(rawHeader, maintainCase);
+
       let tail = children[children.length - 1];
 
-      // A bit weird: this is to support MDX 2 comments in expressions,
```
gaearon, line 36: This was a valuable comment so we shouldn't nix it.

beta/plugins/remark-header-custom-ids.js
```diff
@@ -32,32 +32,26 @@ module.exports = ({
...
+            children.splice(children.length - 2, 2);
+            // Remove the `{/`
+            tail.value = tail.value.replace(/[ \t]*\{\/$/, '');
+          }
+        }
       }
 
-      tail = children[children.length - 2];
-
-      if (!tail && tail.type !== 'emphasis') {
```
gaearon, line 44: This was a mistake in the original code btw. It should've used `||` because otherwise the second check would crash whenever `tail` is null or undefined.

beta/plugins/remark-header-custom-ids.js
```diff
@@ -32,32 +32,26 @@ module.exports = ({
...
-
-      // A bit weird: this is to support MDX 2 comments in expressions,
-      // while we’re still on MDX 1, which doesn’t support them.
-      if (!tail || tail.type !== 'text' || tail.value !== '/}') {
-        return;
+      // Generate slugs on the fly (even if not specified in markdown)
+      // so that it's possible to copy anchor links in newly written content.
+      let id = slugs.slug(toString(node), maintainCase);
+      // However, for committed docs, we'll extract slug from the headers.
+      if (tail && tail.type === 'text' && tail.value === '/}') {
```
harish-sethuraman, line 39:
```suggestion
      if (tail?.type === 'text' && tail?.value === '/}') {
```
may be?

beta/plugins/remark-header-custom-ids.js
```diff
@@ -32,32 +32,26 @@ module.exports = ({
...
+        tail = children[children.length - 2];
+        if (tail && tail.type === 'emphasis') {
```
harish-sethuraman, line 41:
```suggestion
        if (tail?.type === 'emphasis') {
```

beta/plugins/remark-header-custom-ids.js
gaearon: i don't think this is helping clarity tbh

beta/plugins/remark-header-custom-ids.js
gaearon: i think the presence of the tail itself is kind of an important thing to underscore there so i like it being explicit

PR #4309, opened 2022-02-07
[Beta] useState Troubleshooting

gaearon:
## [Preview page](https://beta-reactjs-org-dyvj1mf29-fbopensource.vercel.app/apis/usestate)

Adds the most common mistakes and common solutions. I tried to keep it light and to-the-point.

Also added a bit of extra content to the earlier example gallery.

### DIFF
--- a/beta/src/pages/apis/usestate.md
+++ b/beta/src/pages/apis/usestate.md
@@ -1082,12 +1082,12 @@ console.log(nextCount); // 1
 
 React will **ignore your update if the next state is equal to the previous state,** as determined by an [`Object.is`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) comparison. This usually happens when you change an object or an array in state directly:
 
-```js
+```js {2}
 obj.x = 10;
-setObj(obj);
+setObj(obj); // Has no effect
 ```
 
-You called `setObj` with the same `obj` object, so React bailed out of rendering. To fix this, you need to ensure that you're always [_replacing_ objects arrays in state instead of _mutating_ them](#updating-objects-and-arrays-in-state):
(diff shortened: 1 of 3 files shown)


beta/src/pages/apis/usestate.md
```diff
@@ -856,3 +1044,72 @@ button { margin-bottom: 10px; }
...
+  setCount(count + 1); // Request a re-render with 1
+  console.log(count);  // Still 0!
+
+  setTimeout(() => {
+    console.log(count); // Also 0!
+  }, 5000);
+}
+```
+
+This is because [states behaves like a snapshot](/learn/state-as-a-snapshot). Updating state requests another render with the new state value, but does not affect the `count` JavaScript variable in your already running event handler.
```
sylwiavargas, line 1069:
```suggestion
This is because [states behaves like a snapshot](/learn/state-as-a-snapshot). Updating state requests another render with the new state value, but does not affect the `count` JavaScript variable in your already-running event handler.
```

beta/src/pages/apis/usestate.md
```diff
@@ -856,3 +1044,72 @@ button { margin-bottom: 10px; }
...
+return <button onClick={handleClick()}>Click me</button>
+
+// ✅ Correct: passes down the event handler
+return <button onClick={handleClick}>Click me</button>
+
+// ✅ Correct: passes down an inline function
+return <button onClick={(e) => handleClick(e)}>Click me</button>
+```
+
+If you can't find the cause of this error, click on the arrow next to the error in the console, and look through the JavaScript stack to find the specific `set` function call responsible for the error.
```
sylwiavargas, line 1114: I love this tip. Do you think we could add an example of this (a screenshot, for instance)?

Also, an unnecessary comma:
```suggestion
If you can't find the cause of this error, click on the arrow next to the error in the console and look through the JavaScript stack to find the specific `set` function call responsible for the error.
```

PR #4312, opened 2022-02-08
Cache the font files

shamwela:
I cached the font files for a year to improve performance.


beta/vercel.json
```diff
@@ -254,5 +254,43 @@
       "destination": "https://5a046bf5a6188f4b8fa4938a--reactjs.netlify.com",
       "permanent": false
     }
+  ],
+  "headers": [
+    {
+      "source": "/fonts/SourceCodePro-Regular.ttf",
```
harish-sethuraman, line 260: There should probably be a way to add all font files inside the `fonts` like `/fonts/(.*).ttf` ?

shamwela: Thanks. I've improved the syntax.

beta/vercel.json
```diff
@@ -254,5 +254,16 @@
       "destination": "https://5a046bf5a6188f4b8fa4938a--reactjs.netlify.com",
       "permanent": false
     }
+  ],
+  "headers": [
+    {
+      "source": "/fonts(.*)",
```
harish-sethuraman, line 260:
```suggestion
      "source": "/fonts/.*",
```
Also we might want to check if this works or we need to do `\/fonts\/.*` cc: @gaearon

gaearon:
I don’t think we want to hold them in the same repo since they’re not provided by the same license.

shamwela:
I didn't know about the licenses. Should I only cache the "Source Code Pro Regular" font file?

gaearon:
Let me chat to some folks on our side and then report back

shamwela:
Any update on this, please?

gaearon:
Let's do this just for SourceCode for now?

shamwela:
I'll do that in a new pull request. Thanks.

PR #4313, opened 2022-02-09
Add prevState-only example

LWChris:
Add an example where only the previous state is used to update the state; this prepares the reader for the toggle button example in the Handling Events chapter

### DIFF
--- a/content/docs/state-and-lifecycle.md
+++ b/content/docs/state-and-lifecycle.md
@@ -373,6 +373,15 @@ this.setState(function(state, props) {
 });
 ```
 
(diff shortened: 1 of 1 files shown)


LWChris:
Please note: I do not have a Facebook account, I do not want a Facebook account, and I will not create a Facebook account.

I can tell you that you can use this commit and the content in any way you like. If that doesn't do it, please ignore my request.

shamwela:
Contributor License Agreement is just an agreement. You don't need a Facebook account.

LWChris:
Okay I think I signed it.

PR #4317, opened 2022-02-09
Update dangerouslySetInnerHTML example

B0und:
Ive used this code in my application and it caused endless rerenders...


BartoszKlonowski:
Hello @B0und,
I can see that the `createMarkup` is no longer in the docs, so I will close this PR as no longer required.
Thanks!

PR #4318, opened 2022-02-10
[Beta] Mention batching in useState

gaearon:
Seems like we should document this if we try to be exhaustive and useful.

https://beta-reactjs-org-qfyn6vxmo-fbopensource.vercel.app/apis/usestate#setstate-caveats


beta/src/pages/apis/usestate.md
```diff
@@ -87,12 +87,16 @@ function handleClick() {
 
 `set` functions do not have a return value.
 
-#### Caveats {/*caveats*/}
+#### Caveats {/*setstate-caveats*/}
 
 * The `set` function only updates the state variable for the next render. If you read the state variable after calling the `set` function, you will still get the value that was on the screen before your call.
```
sylwiavargas, line 92: How do you feel about adding an emphasis?

```suggestion
* The `set` function only updates the state variable for the _next_ render. If you read the state variable after calling the `set` function, you will still get the value that was on the screen before your call.
```

Maybe we could link the last subsentence to the learn section where we explain it with an example? Con: kinda makes the text lengthier. Pro: helps people see it themselves.

beta/src/pages/apis/usestate.md
```diff
@@ -87,12 +87,16 @@ function handleClick() {
...
 
 * If the new value you provide is identical to the current `state`, as determined by an [`Object.is`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) comparison, React will bail out of re-rendering the component and its children. However, in some cases React may still need to call your component before discarding the result.
```
sylwiavargas, line 94: "However, in some cases React may still need to call your component before discarding the result."

Could you provide an example? Otherwise it may induce a low-key anxiety similar to "I don't know what I don't know."

Also, I'm proposing to change the "bail out" because it's not simple English:

```suggestion
If the new value you provide is identical to the current `state`, as determined by an [`Object.is`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) comparison, React will not re-render the component and its children. However, in some cases React may still need to call your component before discarding the result.
```

beta/src/pages/apis/usestate.md
```diff
@@ -87,12 +87,16 @@ function handleClick() {
...
 
+* React batches state updates: it does not update the screen until right after all the event handlers have called their `set` functions. This prevents multiple re-renders during a single event. In the rare case that you need to force React to update the screen earlier so that you can access the DOM, you can use [`flushSync`](/apis/flushsync).
```
sylwiavargas, line 96:
```suggestion
* React batches state updates: it does not update the screen until right after all the event handlers have called their `set` functions. This prevents multiple re-renders during a single event. In the rare case that you need to force React to update the screen earlier, for example to access the DOM, you can use [`flushSync`](/apis/flushsync).
```

beta/src/pages/apis/usestate.md
```diff
@@ -87,12 +87,16 @@ function handleClick() {
...
+  - *In React 17 and earlier,* batching is only enabled for the `set` function calls during React-managed event handlers, like the ones you put in your JSX. React updates the screen at the end of the browser event. Updates from two distinct events never get batched together. 
+  - *In React 18 and later,* batching is enabled for all `set` function calls. This includes both React event listeners and listeners you've added manually with `addEventListener`. React updates the screen in the next [microtask](https://developer.mozilla.org/en-US/docs/Web/API/HTML_DOM_API/Microtask_guide). Updates from two distinct *intentional* events like clicks or typing never get batched together. 
+
 * Calling the `set` function *during rendering* is only allowed from within the currently rendering component. React will discard its output and immediately attempt to render it again with the new state. React will do this in a loop until you either stop setting state, or do it too many times. This pattern is called "derived state". It is uncommon, but you can use it to store information from previous renders. [See an example below.](#storing-information-from-previous-renders)
```
sylwiavargas, line 100: A rogue comma.

```suggestion
* Calling the `set` function *during rendering* is only allowed from within the currently rendering component. React will discard its output and immediately attempt to render it again with the new state. React will do this in a loop until you either stop setting state or do it too many times. This pattern is called "derived state". It is uncommon, but you can use it to store information from previous renders. [See an example below.](#storing-information-from-previous-renders)
```

gaearon:
Btw my main concern is with the length of this section. Any ideas on what to do with it?

beta/src/pages/apis/usestate.md
gaearon: >Could you provide an example?

This is tricky — *the whole point* is that we're not giving any guarantees and when that might happen, and it *should not matter* for your code. I don't know how to communicate this.

beta/src/pages/apis/usestate.md
```diff
@@ -87,12 +87,16 @@ function handleClick() {
...
 
 * In Strict Mode, React will call your updater function twice in order to help you find accidental impurities. This is development-only behavior and does not affect production. If your updater function is pure (as it should be), this should not affect the logic of your component. The result from one of the calls will be ignored.
```
sylwiavargas, line 102: Do you think we could link the "accidental impurities" to the learn section on pure functions? I just worry that some readers may not know about what constitutes impurity.

beta/src/pages/apis/usestate.md
sylwiavargas: How about:

```suggestion
If the new value you provide is identical to the current `state`, as determined by an [`Object.is`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) comparison, React will not re-render the component and its children. Although in some cases React may still need to call your component before discarding the result, that bears no consequence for your code.
```

sylwiavargas:
Ideas for shortening the section:
- We could make the React 17 and React 18 specifics (subpoints of the third bullet point) be a `<details>` kinda thing that readers can expand to learn more, which would save a length of two bullets;
- We could have each of the caveats be just one-sentence references to a page of the Learn section
- We could try rewriting the points to make them sound like "you don't need more, you can't have less" explanation

Does any of these feel right?

PR #4319, opened 2022-02-10
[Beta] Change render() API page format

gaearon:
## [Rendered preview](https://beta-reactjs-org-pz36xb7a4-fbopensource.vercel.app/apis/render)

To match the one we use in the useState page.

I had to get a bit creative with the overload because I don't want to use the `[, optional]` stuff or similar which gets confused for arrays or `?` stuff which requires understanding types.


beta/src/pages/apis/render.md
```diff
@@ -4,16 +4,74 @@ title: render()
...
 ```
 
-</Intro>
+React will take over managing the DOM inside of the `container`.
+
+An app fully built with React will usually only have one `render` call with its root component.  A page that uses "sprinkles" of React for parts of the page may have as many `render` calls as needed. [See examples below.](#usage)
+
+
+* `reactNode`: A *React node* that you want to display. This will usually be a piece of JSX like `<App />`, but you can also pass a React elmenent constructed with [`createElement()`](/TODO), a string, a number, `null`, or `undefined`.
```
harish-sethuraman, line 38:
```suggestion
* `reactNode`: A *React node* that you want to display. This will usually be a piece of JSX like `<App />`, but you can also pass a React element constructed with [`createElement()`](/TODO), a string, a number, `null`, or `undefined`. 
```

beta/src/pages/apis/render.md
```diff
@@ -4,16 +4,74 @@ title: render()
...
-</Intro>
+React will take over managing the DOM inside of the `container`.
+
+An app fully built with React will usually only have one `render` call with its root component.  A page that uses "sprinkles" of React for parts of the page may have as many `render` calls as needed. [See examples below.](#usage)
+
+#### Parameters {/*parameters*/}
+
+* `reactNode`: A *React node* that you want to display. This will usually be a piece of JSX like `<App />`, but you can also pass a React element constructed with [`createElement()`](/TODO), a string, a number, `null`, or `undefined`. 
+
+* `container`: A [DOM element](https://developer.mozilla.org/en-US/docs/Web/API/Element). React will change its content to the rendering output of the first argument.
```
harish-sethuraman, line 40:
```suggestion
* `container`: A [DOM element](https://developer.mozilla.org/en-US/docs/Web/API/Element) to render the `reactNode`
```

I feel that sentence to be quite confusing 🤔 we can rather have it simple?

beta/src/pages/apis/render.md
```diff
@@ -4,16 +4,74 @@ title: render()
...
+#### Caveats {/*caveats*/}
+
+* The first time you call `render` with some `container` node, React will clear that node's existing HTML content before rendering the React component into it. If your `container` contains HTML generated by React on the server or during the build, use [`hydrate()`](/TODO) instead, which attaches the event handlers to the existing HTML.
+
+* If you call `render` on the same `container` more than once, React will update the DOM as necessary to reflect the latest JSX you passed. React will decide which parts of the DOM can be reused and which need to be recreated by ["matching it up"](/learn/preserving-and-resetting-state) with the previously rendered tree. Calling `render` on the same `container` again is similar to calling the [`set` function](/apis/usestate#setstate) on the root component: React avoids unnecessary DOM updates.
+
+* If your app is fully built with React, you shouldn't need to use `render` more than in a single place in your app. (You might even not use it at all if a framework does this for you.) If you want to render something in a different part of the DOM tree (for example, a modal or a tooltip), use [`createPortal`](TODO) instead.
```
harish-sethuraman, line 52:
```suggestion
* If your app is fully built with React, you shouldn't use `render` more than in a single place in your app. (You might even not use it at all if a framework does this for you.) If you want to render something in a different part of the DOM tree (for example, a modal or a tooltip), use [`createPortal`](TODO) instead.
```

beta/src/pages/apis/render.md
```diff
@@ -4,16 +4,74 @@ title: render()
...
+
+---
 
-## Rendering the root component {/*rendering-the-root-component*/}
+### `render(element, container, callback)` {/*render-callback*/}
+
+Same as [`render(element, container)`](#renderreactnode-container).
```
harish-sethuraman, line 58:
```suggestion
Same as [`render(element, container)`](#render).
```

beta/src/pages/apis/render.md
```diff
@@ -4,16 +4,74 @@ title: render()
...
+
+#### Parameters {/*render-callback-parameters*/}
+
+* `element`: Same as above.
```
harish-sethuraman, line 62: Shall we rather include the same text here too? If someone lands at this part from the TOC they should again try to find use of these in other block?

beta/src/pages/apis/render.md
```diff
@@ -4,16 +4,74 @@ title: render()
...
+* `container`: Same as above.
```
harish-sethuraman, line 63: In the TOC:
Screenshot 2022-02-10 at 9 32 19 AM

In the right pane:
Screenshot 2022-02-10 at 9 32 27 AM

beta/src/pages/apis/render.md
harish-sethuraman: Not sure how this got missed 😶‍🌫️

PR #4322, opened 2022-02-11
[Beta] Fix layout shift from previews

gaearon:
## [Preview](https://beta-reactjs-org-n4126yxxu-fbopensource.vercel.app/learn)

This should fix the layout shift from previews on layouts where we overflow to the "vertically stacked" code / preview blocks.  In this case we don't know the preview height, so we're going to always render 40% of the viewport. This is not ideal but it's better than jumpy unpredictable scroll.

### DIFF
--- a/beta/src/components/MDX/Sandpack/CustomPreset.tsx
+++ b/beta/src/components/MDX/Sandpack/CustomPreset.tsx
@@ -73,8 +73,7 @@ export function CustomPreset({
             <SandpackCodeEditor
               customStyle={{
                 height: getHeight(),
-                maxHeight: isExpanded ? '' : '[40vh]',
-                minHeight: isExpanded ? '' : 'min-h-[40vh] xl:min-h-0',
+                maxHeight: isExpanded ? '' : '40vh',
               }}
               showLineNumbers
               showInlineErrors
--- a/beta/src/components/MDX/Sandpack/CustomPreset.tsx
+++ b/beta/src/components/MDX/Sandpack/CustomPreset.tsx
@@ -73,8 +73,8 @@ export function CustomPreset({
             <SandpackCodeEditor
               customStyle={{
                 height: getHeight(),
-                maxHeight: isExpanded ? '' : '40vh',
-                minHeight: isExpanded ? '' : 'min-h-40vh xl:min-h-0',
+                maxHeight: isExpanded ? '' : '[40vh]',
+                minHeight: isExpanded ? '' : 'min-h-[40vh] xl:min-h-0',
               }}
               showLineNumbers
               showInlineErrors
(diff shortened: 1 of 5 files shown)


beta/src/components/MDX/Sandpack/CustomPreset.tsx
```diff
@@ -73,15 +73,16 @@ export function CustomPreset({
...
+                minHeight: isExpanded ? '' : 'min-h-40vh xl:min-h-0',
               }}
               showLineNumbers
               showInlineErrors
               showTabs={false}
             />
             <Preview
               isExpanded={isExpanded}
-              className="order-last xl:order-2"
+              className="order-last min-h-40vh xl:order-2 xl:min-h-0"
```
harish-sethuraman, line 85:
```suggestion
              className="order-last min-h-[40vh] xl:order-2 xl:min-h-0"
```
May be we could do this instead of config level changes?

gaearon: What’s that? Some new syntax?

harish-sethuraman: Arbitrary values in tailwind https://tailwindcss.com/docs/min-height#arbitrary-values

gaearon:
Hmm. Looks like long sandboxes like in Thinking in React now get expanded in mobile mode. Need to fix this somehow.

beta/src/components/MDX/Sandpack/CustomPreset.tsx
```diff
@@ -73,15 +73,16 @@ export function CustomPreset({
             <SandpackCodeEditor
               customStyle={{
                 height: getHeight(),
-                maxHeight: isExpanded ? '' : 406,
+                maxHeight: isExpanded ? '' : '[40vh]',
```
harish-sethuraman, line 76: Why [] here ?

gaearon: yeah that's a mistake

gaearon:
Still seems a bit wrong because on desktop we don't want vertical resize to "squeeze" the sandbox. We probably need to somehow make the vh max-height conditional on the media breakpoint.

harish-sethuraman:
Are you looking for this https://github.com/reactjs/reactjs.org/blob/52393daffe7baec8af53bc2359d74077ddd1e2bb/beta/src/components/Layout/useMediaQuery.tsx#L7 ? 👀

gaearon:
I'd prefer if we did this in CSS without any JS code to track the media query.

gaearon:
So I think we just need to turn these styles into tailwind classes.

gaearon:
I don't think `SandpackCodeEditor` accepts a `className`, but maybe we could put it in a div and style that.

gaearon:
If you wanna give it a go, feel free to :)

harish-sethuraman:
This change adds scroll for larger previews. I hope that should be fine 😄
Screenshot 2022-02-12 at 11 04 57 AM
In smaller previews we have too much of empty space (bigger than the code area)? Is this okay? Or shall we rather arrange the output in the center?

gaearon:
>This change adds scroll for larger previews.

Yeah, that's the tradeoff. One thing we could do is just to tweak the code examples to assume they need to fill in a bit more space.

>Or shall we rather arrange the output in the center?

I'd be curious to see how this looks, if you wanna try! Not sure it'll work.

PR #4327, opened 2022-02-12
Fix broken link in render-and-commit.md

mayank99:
`/reference/setState` -> `/apis/usestate#setstate`

For the actual text, I also considered being more verbose (e.g. "the `set` function from `useState`") but decided to go with the shorter version.


harish-sethuraman (approved):
LGTM

Issue #4287, opened 2022-02-04
[Beta] Code editor component: the error icon is overlapping the code 


danilowoz:
How to reproduce:
- Got to `learn/javascript-in-jsx-with-curly-braces`;
- Last challenge (Write an expression inside JSX curly braces);
- Get a inline error in long line;
<img width="473" alt="Screenshot 2022-02-04 at 09 39 07" src="https://user-images.githubusercontent.com/4838076/152506560-1e56331d-3b4c-4bb7-8c5c-c0ebad255e46.png">

amareshsm:
@harish-sethuraman  Can I pick this?

harish-sethuraman:
Feel free to send a PR :)

gaearon:
I reverted the fix because it broke the more common case.
So this is still unfixed.

imanpalsingh:
@gaearon  I have raised a PR for this https://github.com/reactjs/reactjs.org/pull/4332 please review.

PR #4332, opened 2022-02-13
Fix - Sandpack error icon overlapping issue 

imanpalsingh:
Issue: Sandpack error icon overlapping code #4287

### Before 
Screenshot 2022-02-13 at 12 00 11 PM

### After
Screenshot 2022-02-13 at 11 58 30 AM

### DIFF
--- a/beta/src/styles/sandpack.css
+++ b/beta/src/styles/sandpack.css
@@ -90,10 +90,7 @@ html.dark .sp-tabs .sp-tab-button[data-active='true'] {
 .sp-code-editor .cm-errorLine {
   background-color: rgba(255, 107, 99, 0.1);
   position: relative;
-}
-
-.sp-code-editor .cm-content .cm-errorLine {
-  padding-right: 26px;
+  padding-right: 26px !important;
 }
 
 .sp-code-editor .cm-errorLine:after {
--- a/beta/src/styles/sandpack.css
+++ b/beta/src/styles/sandpack.css
@@ -92,6 +92,10 @@ html.dark .sp-tabs .sp-tab-button[data-active='true'] {
   position: relative;
 }
 
+.sp-code-editor .cm-content .cm-errorLine {
+  padding-right: 26px;
+}
+
 .sp-code-editor .cm-errorLine:after {
   position: absolute;
   right: 8px;


harish-sethuraman:
Seems like adding padding for the icon introduces an additional scroll just to see the error icon 
| PR change | Current prod |
|----- | ------ | 
|Screenshot 2022-02-13 at 12 28 44 PM | Screenshot 2022-02-13 at 12 28 56 PM


`________</h1>< alt={'Photo of ' + user.name} className=``"avtar"`

In case you want to test this out ☝🏻  is the line that I've added to the sandpack

imanpalsingh:
@harish-sethuraman yes, the scrollbar will appear if line takes up the entire width. Where else should the icon go? Should it go to the next line?

harish-sethuraman:
>  Where else should the icon go?

I'm not exactly sure of this. I'll wait for Dan's thought on where the icon should go.

PS: I think it would be nice if the icon appears on top of code without showing the overlap (with appropriate background color)

imanpalsingh:
@harish-sethuraman  I agree but if the user has to edit the character that is directly below the icon then the user will not be able to see it properly.

harish-sethuraman:
> I agree but if the user has to edit the character that is directly below the icon then the user will not be able to see it properly.

That is how the current experience already is. Adding a scroll just for showing the error icon might not be right either...

gaearon:
I'm still seeing overlap with this PR.

Screenshot 2022-02-13 at 23 49 46

imanpalsingh:
@gaearon can you please mention the steps to reproduce this? I am testing myself and I don't see it overlapping

imanpalsingh:
@gaearon  @harish-sethuraman  Please review again.

beta/src/styles/sandpack.css
```diff
@@ -90,6 +90,9 @@ html.dark .sp-tabs .sp-tab-button[data-active='true'] {
 .sp-code-editor .cm-errorLine {
   background-color: rgba(255, 107, 99, 0.1);
   position: relative;
+  padding-right: 26px !important;
+  display: inline-block;
```
danilowoz, line 94: I don't think this display is necessary, this element is already `block` and the following property (min-width) is a consequence of the `display: inline-block`. So, the padding should be more than enough to have the same result.

amareshsm:
image

In the above case, Error icon is not visible. It appears only at the end (after the longest text). 

image

gaearon:
thanks.

PR #4335, opened 2022-02-14
[Beta] Add ARIA labels

shamwela:
This will improve accessibility for screen readers. I've checked with Lighthouse locally.

### DIFF
--- a/beta/src/components/Layout/Footer.tsx
+++ b/beta/src/components/Layout/Footer.tsx
@@ -126,11 +126,13 @@ export function Footer() {
               </FooterLink>
               <div className="flex flex-row mt-8 gap-x-2">
                 <ExternalLink
+                  aria-label="React on Facebook"
                   href="https://www.facebook.com/react"
                   className={socialLinkClasses}>
                   <IconFacebookCircle />
                 </ExternalLink>
                 <ExternalLink
+                  aria-label="React on Twitter"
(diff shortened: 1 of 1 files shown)


beta/src/components/Search.tsx
```diff
@@ -95,13 +95,15 @@ export const Search: React.FC<SearchProps> = ({
...
       <button
+        aria-label="Search"
         type="button"
         className="inline-flex md:hidden items-center text-lg p-1 ml-4 lg:ml-6"
         onClick={onOpen}>
         <IconSearch className="align-middle" />
       </button>
 
       <button
+        aria-label="Search"
```
gaearon, line 106: This one shouldn't be necessary because there's text inside?

harish-sethuraman:
Would you like look into these as well? 
Screenshot 2022-02-14 at 10 23 00 PM
Screenshot 2022-02-14 at 10 23 10 PM

beta/src/components/Search.tsx
shamwela: I've removed it now. Thanks.

shamwela:
I'll take a look.

shamwela:
I've solved ARIA label issues. I'll solve other accessibility issues in another pull request.

harish-sethuraman (approved):
LGTM

PR #31, opened 2017-10-06
Docs for Error Boundaries

Krizzu:
Hey,

Moved this PR here, as said :)

PR #4339, opened 2022-02-15
[beta] PoC: replace Sandpack with react-runner

nihgwu:
PoC for https://github.com/reactjs/reactjs.org/issues/4338

[Preview](https://beta-reactjs-org-d27vse7l5-fbopensource.vercel.app/learn/scaling-up-with-reducer-and-context)

@gaearon It's quite rough but should work already, haven't checked why theme is not applied for CodeMirror but it works [here](https://react-runner-codemirror.vercel.app/)

### DIFF
--- a/beta/src/components/MDX/LiveRunner/LivePreview.tsx
+++ b/beta/src/components/MDX/LiveRunner/LivePreview.tsx
@@ -1,0 +1,170 @@
+import React, {useState, useMemo} from 'react';
+import {Runner, importCode, Scope} from 'react-runner';
+import * as UseImmer from 'use-immer';
+
+import {Error as LiveError} from './Error';
+import {ShadowRoot} from './ShadowRoot';
+import {RunnerFile} from './types';
+
+const defaultStyle = `
+* {
+  box-sizing: border-box;
+}
+
+:host {
+  font-family: sans-serif;
+  margin: 20px;
+  padding: 0;
+}
+
+h1 {
+  margin-top: 0;
+  font-size: 22px;
+}
+
+h2 {
+  margin-top: 0;
+  font-size: 20px;
+}
+
+h3 {
+  margin-top: 0;
+  font-size: 18px;
+}
+
+h4 {
+  margin-top: 0;
+  font-size: 16px;
+}
+
+h5 {
+  margin-top: 0;
+  font-size: 14px;
+}
+
+h6 {
+  margin-top: 0;
+  font-size: 12px;
+}
+
+ul {
+  padding-left: 20px;
+}
+`.trim();
+
+const withFiles = (scope: Scope, files: Record<string, string>) => {
+  const imports: Scope = {...scope.import};
+  const lookup = new Set<string>();
+  const importsProxy: Scope = new Proxy(imports, {
+    getOwnPropertyDescriptor(target, prop) {
+      if (target.hasOwnProperty(prop)) {
+        return Object.getOwnPropertyDescriptor(target, prop);
+      }
+      if (files.hasOwnProperty(prop)) {
+        return {writable: true, enumerable: true, configurable: true};
+      }
+      return undefined;
+    },
+    get(target, prop: string) {
+      if (prop in target) return target[prop];
+      if (files.hasOwnProperty(prop)) {
+        if (lookup.has(prop)) {
+          throw new Error(
+            `Circular dependency detected: ${[...Array.from(lookup), prop].join(
+              ' -> '
+            )}`
+          );
+        }
+        lookup.add(prop);
+        return (target[prop] = importCode(files[prop], {
+          ...scope,
+          import: importsProxy,
+        }));
+      }
+    },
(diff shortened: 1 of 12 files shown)


gaearon:
I think this looks very promising. How much work is it to get this styled so that the editor works and we can compare 1:1? I see some sandboxes on https://beta-reactjs-org-n2pucbumf-fbopensource.vercel.app/learn/updating-objects-in-state are a bit borked.

beta/src/components/MDX/LiveRunner/LivePreview.tsx
```diff
@@ -0,0 +1,117 @@
+import React, {useState, useMemo} from 'react';
+import {Runner, importCode, Scope} from 'react-runner';
+import * as UseImmer from 'use-immer';
```
gaearon, line 3: It's not super great that we have to load everything we use in every sandbox. We might want to have "heavier" examples in the future with third party libs. They shouldn't bloat the initial editor. I guess having a list of them here is probably fine though. Maybe as dynamic imports and somehow lazy-loaded?

nihgwu: Yes lazy loading should definitely work, and we can still keep the `package.json` file and analyse the dependency, right now I just ignored that file

nihgwu:
@gaearon There are some visual regressions, as previously we target the to root document but now we use a shadow DOM instead, I think it should be super easy to fix, if we decided to make the switch

gaearon:
I don't think we can decide whether to make a switch until we have a good idea of how it works end-to-end, including all the styling, having the code highlighting hooked up, etc. I know it's a bunch of work but it's hard to compare otherwise.

nihgwu:
@gaearon I totally understand, I'll try to fix the styling issues first, it works well in my demo though, and the benefit is obvious, the demo is part of our app, so it's super fast, and can reflect user's input with no delay, no bundling, only compiling the JSX syntax

nihgwu:
@gaearon tweaked the styling a bit, though there are still a lot of work to do

gaearon:
A few quick thoughts.

### Overall approach

I think it's pretty clear this is the right direction for the tradeoffs we have today. I was hoping we can introduce a mode like this to Sandpack in longer term. The instant feedback is hard to beat. The network waterfalls and other inefficiencies in Sandpack today are unfortunate because our use case is very limited. I think it's plausible the Sandpack package will evolve to support this model as opt-in. I'd like to hear more from their perspective and I'm not saying we should necessarily switch packages to yours. But it seems clear an approach _like this_ is promising, whether through this package or something else.

### Production vs Development

One thing that seems unfortunate with this approach is that errors are minified.
E.g. try adding `{{}}` into some JSX. You'll get:

>Minified React error #31; visit https://reactjs.org/docs/error-decoder.html?invariant=31&args[]=object%20with%20keys%20%7B%7D for the full message or use the non-minified dev environment for full errors and additional helpful warnings.

But since we're on the docs site, it seems confusing. It also means we'd miss `console.error`s. (I don't think Sandpack currently displays them, but ideally, I think we'd like to do that for key warning etc.)

One option would be to make the docs site use a development build of React. Seems bad.

Another option would be to load development build of React, but just for sandboxes. Seems okay-ish. Though also feels a bit silly if we just want to render a few examples. I'm not sure about the best approach.

There could be some middle ground. Like use prod build until dev builds loads lazily. Though might be confusing if a sandbox has an error initially.

### Fast Refresh

It's not great that sandboxes lose state on edits. Sandpack implements a simple Fast Refresh integration to avoid that. I think you can't do this very easily because you don't have Babel though and can't run the Fast Refresh transform. Maybe some simplified version could be written. It's thorny with multi-file edits though.

Maybe it's ok that it doesn't work. But feels like a regression.

nihgwu:
@gaearon Thanks for your thorough feedbacks, here are my rough answers:

### Production vs Development

One possible solution is map error message back to decoded one, since the data will be stored in the same app, I think it should not be that hard

### Fast Refresh

I haven't tested yet but the transpiler React Runner uses is `sucrase`, which supports `react-hot-loader`, not sure if that would work in this case, but again I think that's not a big problem as we can also use Babel to compile the code, then we can use the plugin ecosystem of Babel. The only reason I chose `sucrase` is the much smaller bundle size, but if we want provide more versatile functionalists, we can use `Babel` in this case and lazy load it in a worker like some other solutions do

### Server Components

This website it built with Next.js which also supports Server Components, probably we can just use it? Or for this case we can just other solutions like Sandpack or CodeSandbox or StackBlitz, we are not bonded to use only one tool for all scenarios

### Third Party Libs/Transforms

If we do need babel plugins, we can use `Babel` as the transplier instead of `sucrase`, BTW it's works well with `styled-components` and `emotion` without any babel plugins

BartoszKlonowski:
Closing as per comment in the issue, but thank you very much, @nihgwu for taking the time to contribute! 👍

PR #4340, opened 2022-02-15
[18] ReactDOM reference to createRoot/hydrateRoot

rickhanlonii:
## Overview

*Note: PR against the react-18 branch*

This PR updates the `ReactDOM` reference to add `createRoot` and `hydrateRoot`. Will followup with all the other references in the repo.

I'm sure I missed some thing in the explanation, but want to open this up to start the discussion.

### DIFF
--- a/content/docs/reference-react-dom.md
+++ b/content/docs/reference-react-dom.md
@@ -12,11 +12,17 @@ If you load React from a `<script>` tag, these top-level APIs are available on t
 
 The `react-dom` package provides DOM-specific methods that can be used at the top level of your app and as an escape hatch to get outside of the React model if you need to. Most of your components should not need to use this module.
 
-- [`render()`](#render)
-- [`hydrate()`](#hydrate)
-- [`unmountComponentAtNode()`](#unmountcomponentatnode)
-- [`findDOMNode()`](#finddomnode)
+- [`createRoot()`](#createroot)
+- [`hydrateRoot()`](#hydrateroot)
 - [`createPortal()`](#createportal)
+
+> Note
+> 
+> These methods are considered legacy. Both `render` and `hydrate` will warn that your app is running in React 17:
+>- [`render()`](#render)
+>- [`hydrate()`](#hydrate)
+>- [`findDOMNode()`](#finddomnode)
+>- [`unmountComponentAtNode()`](#unmountcomponentatnode)
 
 ### Browser Support {#browser-support}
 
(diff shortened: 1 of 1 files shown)


content/docs/reference-react-dom.md
```diff
@@ -26,12 +32,65 @@ React supports all popular browsers, including Internet Explorer 9 and above, al
...
 
+## Reference {#reference}
+
+### `createRoot()` {#createroot}
+
+```javascript
+ReactDOM.createRoot(container[, options]);
+```
+
+Create a React root for the supplied `container` and return the root. The root can be used to Render a React element into the DOM with `render`:
```
acdlite, line 43: Nit: "Render" -> "render"

content/docs/reference-react-dom.md
```diff
@@ -12,12 +12,18 @@ If you load React from a `<script>` tag, these top-level APIs are available on t
...
-- [`hydrate()`](#hydrate)
-- [`unmountComponentAtNode()`](#unmountcomponentatnode)
-- [`findDOMNode()`](#finddomnode)
+- [`createRoot()`](#createroot)
+- [`hydrateRoot()`](#hydrateroot)
 - [`createPortal()`](#createportal)
 
+> Note
+> 
+> These methods are considered legacy. Both `render` and `hydrate` will warn that your app is running in React 17:
```
eps1lon, line 21: I think what this means is that React is only rendering synchronously? "running in React 17" is not that clear to me.

gaearon: They both render synchronously (createRoot by itself doesn't opt into concurrency), so I think this is primarily about disabling automatic batching. I agree "in React 17" is confusing. We should make sure that, whichever way we word it, your upgrade to 18 is not considered "completed" until you've switched away from render.

rickhanlonii: @gaearon are you saying that we should include upgrade instructions here? My goal in including this line was to give  heads up about the warning you'll see in the console which is:

> ReactDOM.hydrate is no longer supported in React 18. Use hydrateRoot instead. Until you switch to the new API, your app will behave as if it's running React 17. Learn more: https://reactjs.org/link/switch-to-createroot

Updated to clarify a bit better, let me know if that works now.

content/docs/reference-react-dom.md
```diff
@@ -26,12 +32,65 @@ React supports all popular browsers, including Internet Explorer 9 and above, al
...
 
-### `render()` {#render}
+```javascript
+ReactDOM.hydrateRoot(element, container[, options])
+```
+
+Same as [`createRoot()`](#createroot), but is used to hydrate a container whose HTML contents were rendered by [`ReactDOMServer`](/docs/react-dom-server.html). React will attempt to attach event listeners to the existing markup.
+
+`hydrateRoot` accepts two optional callbacks as options:
+- `onDeleted`: callback called when content is deleted.
```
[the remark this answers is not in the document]
rickhanlonii: Good callout, I don't know what the canonical information/links for this should be. @gaearon?

content/docs/reference-react-dom.md
eps1lon: Sweet!

PR #4344, opened 2022-02-16
[Beta] Add aria-label and title to permalinks

shamwela:
<s>The heading anchors shouldn't be focusable. So, I added `tabIndex={-1}`. This improves accessibility.</s>

Adds `aria-label` and `title` to permalinks

### DIFF
--- a/beta/src/components/MDX/Heading.tsx
+++ b/beta/src/components/MDX/Heading.tsx
@@ -20,6 +20,11 @@ const Heading = forwardRefWithAs<HeadingProps, 'div'>(function Heading(
   {as: Comp = 'div', className, children, id, isPageAnchor = true, ...props},
   ref
 ) {
+  let label = 'Link for this heading';
+  if (typeof children === 'string') {
+    label = 'Link for ' + children;
+  }
+
   return (
     <>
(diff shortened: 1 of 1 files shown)


gaearon:
Where is the source for this being better?

gaearon:
Isn't the issue here that they have `aria-hidden="true"`? Should they have it or not? I'd like to understand the best practice around that. 

Making anchors non-tabbable seems strictly worse for accessibility. As someone who doesn't use the mouse, how would you select the header to copy a link to it?

Using the lighthouse audit is not enough, we should also use human judgement and try to figure out the right way to do things.

gaearon:
No problem, I’m just saying we could’ve easily made a mistake here, so it’s better to research this from the first principles. I think this is an interesting article with some pros and cons of different approaches: https://amberwilson.co.uk/blog/are-your-anchor-links-accessible/. I don't feel like I understand this space well enough to propose a concrete solution but maybe you could compare them?

shamwela:
image
https://www.w3.org/TR/wai-aria-1.1/#introduction
w3.org does something similar. I also tested with the screen reader and Lighthouse.

gaearon:
thank you!

PR #4350, opened 2022-02-17
Cache a font file

shamwela:
Replaced `ttf` file with `woff2` because it's smaller and faster. Cached this font file for a year. Caching is good for performance.


gaearon:
Looks like the pattern is wrong

beta/vercel.json
```diff
@@ -254,5 +254,16 @@
       "destination": "https://5a046bf5a6188f4b8fa4938a--reactjs.netlify.com",
       "permanent": false
     }
+  ],
+  "headers": [
+    {
+      "source": "/fonts(.*)",
```
harish-sethuraman, line 260: Shall we do something like `/fonts/(.*).woff2`? instead of opening up entire fonts ?

shamwela:
Authorization required to deploy.

gaearon:
Where did you get this version? The ones I see on https://github.com/adobe-fonts/source-code-pro are larger. Why is this one so small?

gaearon:
thanks, this is really good.

gaearon:
ok, i did this for the other font in https://github.com/reactjs/reactjs.org/commit/8def74995a416377a366d882d20d5bc111ccf9d0.

shamwela:
> Where did you get this version? The ones I see on https://github.com/adobe-fonts/source-code-pro are larger. Why is this one so small?

I got it [here](https://google-webfonts-helper.herokuapp.com/fonts/source-code-pro?subsets=latin) but I don't know why it's so small.

PR #4353, opened 2022-02-18
Revert "Cache a font file"

gaearon:
Reverts reactjs/reactjs.org#4350

Bizarrely this appears to break syntax highlighting.


gaearon:
actually looks like it's only broken if the font exists locally

gaearon:
fixed via 45ac43a9f

PR #4357, opened 2022-02-18
Fixes initial jump in sandpack 

harish-sethuraman:
`sp-pristine` class is changed to `sp-dirty` during code change so this CSS is not applied.


gaearon:
thanks!

PR #4361, opened 2022-02-18
Fix incorrect writing of custom documents

decadez:
Fix incorrect writing of the document in https://github.com/reactjs/reactjs.org/pull/4256.

It made a mistake :
- error - unhandledRejection: TypeError: readable.getReader is not a function.

This makes it impossible to debug the project.


harish-sethuraman:
> error - unhandledRejection: TypeError: readable.getReader is not a function.


Trying to understand why this change is required? 

> This makes it impossible to debug the project.

what are you not able to debug?

PR #4362, opened 2022-02-18
Add Strict Effects to Strict Mode

rickhanlonii:
## Overview

Adds a sorta brief overview of Strict Effects to the Strict Mode docs for React 18. I tried to match the level of depth that Detecting Side Effects for double rendering above. Feedback appreciated!

## [Preview](https://reactjs-org-fqccjvzmq-fbopensource.vercel.app/docs/strict-mode.html#detecting-unsafe-effects)

### DIFF
--- a/content/docs/strict-mode.md
+++ b/content/docs/strict-mode.md
@@ -21,6 +21,7 @@ In the above example, strict mode checks will *not* be run against the `Header` 
 * [Warning about deprecated findDOMNode usage](#warning-about-deprecated-finddomnode-usage)
 * [Detecting unexpected side effects](#detecting-unexpected-side-effects)
 * [Detecting legacy context API](#detecting-legacy-context-api)
+* [Detecting unsafe effects](#detecting-unsafe-effects)
 
 Additional functionality will be added with future releases of React.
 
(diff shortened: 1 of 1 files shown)


content/docs/strict-mode.md
```diff
@@ -127,3 +128,65 @@ The legacy context API is error-prone, and will be removed in a future major ver
...
+* React unmounts the component.
+  * Layout effects are destroyed.
+  * Effect effects are destroyed.
+```
+
+> Note:
+>
+> It's common to hear effects being "mounted" and "unmounted". This is because effects have historically been created when a component was mounted, and destroyed when a component was unmounted. This can cause confusion, so we refer to them as "created" and "destroyed" here instead.
+
+In the future, we'd like to add features to React which would allow a component to mount without immediately creating effects (such as pre-rendering), or to destroy effects in already-mounted components (such as when a component isn't visible). These features will add better performance and resource management out-of-the-box to React, but require effects to be decoupled from the component lifecycle and resilient to being created and destroyed multiple times in a component.
```
eps1lon, line 160: I went back and forth whether this should mean 
```suggestion
In the future, we'd like to add features to React which would allow a component to mount without immediately creating effects (such as pre-rendering), or to destroy effects in already-mounted components (such as when a component becomes invisible). These features will add better performance and resource management out-of-the-box to React, but require effects to be decoupled from the component lifecycle and resilient to being created and destroyed multiple times in a component.
```
or not.

But I guess it makes sense to mean "is invisible" considering the "pre-rendering" scenario where we would run effects (knowing that something cannot be visible), committing it, checking visibility, noticing that it isn't, and therefore destroying the effect (just guessing here about the semantics of the Offscreen API).
It's just such a new concept (especially considering focus managment and different definitions of what is considered "visible") that I felt like pointing out that this paragraph might be confusing. 

Though you do say "pre-rendering **or** destroy in already-mounted components" so it sounds like completely separate scenarios in which case I'm wondering why we would ever create effects in a component that never was visible.

content/docs/strict-mode.md
```diff
@@ -127,3 +128,65 @@ The legacy context API is error-prone, and will be removed in a future major ver
...
 
 Read the [new context API documentation](/docs/context.html) to help migrate to the new version.
+
+
+### Detecting unsafe effects {#detecting-unsafe-effects}
+
+Conceptually, React effects include three phases:
+* The **create** phase is called for all effects when the effect is created.
+* The **update** phase is called when the dependencies change, if applicable.
+* The **destroy** phase is called for all effects when the effect is destroyed.
```
eps1lon, line 138: Love this clarification and tying it back to the historical behavior with mounting and unmounting 👍🏻 

Though is there a difference for React developers between update and create?

```suggestion
Conceptually, React effects include three phases:
* The **create** phase is called for all effects when the effect is created i.e. when dependencies change.
* The **destroy** phase is called for all effects when the effect is destroyed.
```
The wording can be polished. It's more about reducing it to two phases: create and destroy.

content/docs/strict-mode.md
rickhanlonii: Would you say something is invisible if it was covered by a modal or dialog? How about "such as when a component is no longer visible on screen"?

rickhanlonii: Resolving, re-wrote this section so no longer relevant.

content/docs/strict-mode.md
rickhanlonii: Removed this section

rickhanlonii:
Going to merge to get this into the 18 branch, but will address any feedback left here as a followup.

PR #4368, opened 2022-02-21
Describing setState function parameter

tarikpnr:
I realized there is no explanation regarding setState can take function parameters, and state can be updated by using setState's function parameters in a more safe way.

### DIFF
--- a/content/docs/hooks-state.md
+++ b/content/docs/hooks-state.md
@@ -187,6 +187,14 @@ In a function, we already have `setCount` and `count` as variables so we don't n
   </button>
 ```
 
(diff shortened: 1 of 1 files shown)


content/docs/hooks-state.md
```diff
@@ -187,6 +187,14 @@ In a function, we already have `setCount` and `count` as variables so we don't n
   </button>
 ```
 
+Alternatively, we could use another way of incrementing count variable by using `setCount` function parameter. This is in fact more safe way to update count variable, because we will make sure to receive the latest count variable from our state.
```
harish-sethuraman, line 190: https://reactjs.org/docs/hooks-reference.html#functional-updates we have it here?

gaearon: Yeah I think it’s reasonable to keep it in the API reference. In new docs it has its own page in the guides.

PR #4371, opened 2022-02-22
Upgrade sandpack to fix code cutting issue in view only editor

harish-sethuraman:
Addresses 
- https://github.com/reactjs/reactjs.org/issues/3308#issuecomment-1046023982


harish-sethuraman:
I've upgraded `0.13.16-experimental.0` to `0.14.1` and build fails @danilowoz Is there any recommended version to upgrade?

beta/package.json
```diff
@@ -22,7 +22,7 @@
     "check-all": "npm-run-all prettier lint:fix tsc"
   },
   "dependencies": {
-    "@codesandbox/sandpack-react": "0.13.16-experimental.0",
+    "@codesandbox/sandpack-react": "0.14.1",
```
danilowoz, line 25: React docs use a different Sanpack release channel (`experimental`), as this application has some hard constraints on peer-dependency. To be clearer, React docs use a specific React version (https://github.com/reactjs/reactjs.org/blob/main/beta/package.json#L38) that we still can't release for most of the Sandpack users. 

I'm still waiting to address one more fix on Sandpack and then I'll release a new `experimental` that will fit here.

danilowoz: Hey @harish-sethuraman, feel free to try `v0.14.3-experimental.0`.
I would suggest following this checklist: https://github.com/reactjs/reactjs.org/pull/4337#issue-1138571647

harish-sethuraman: Verified the checklists and works fine

gaearon:
seems worse after the fix because now it's focusable and has a cursor in it. also, you can "tab" inside.

PR #4372, opened 2022-02-22
Improve ESLint

shamwela:
I replaced `next` with `next/core-web-vitals` because it's stricter and better in my opinion. (https://nextjs.org/docs/basic-features/eslint)
I also added a rule that warns about unused variables. I think this is essential. When I ran `yarn lint`, I saw many unused variables. I'll fix them later.

### DIFF
--- a/beta/.eslintrc
+++ b/beta/.eslintrc
@@ -1,6 +1,9 @@
 {
   "root": true,
-  "extends": "next",
+  "extends": "next/core-web-vitals",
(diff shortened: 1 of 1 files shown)


harish-sethuraman (approved):
We don't have any other rules for this config? Only `no-unused-vars`?

shamwela:
Yes, only `no-unused-vars`.

Issue #4370, opened 2022-02-21
[Beta] Layout bug on Blog page


redhaam:
Hello 👋 
Nice work on the new documentation, I really like the new look.

I found a small bug when visiting the blog page on the website where we can see that:
- A portion of the main section is hidden behind tha sidebar.
- The text color is undreable because it doesn't adapt for the dark theme.
- There's a horizontal scroll on the mobile view.

Link to the page containing the bug : https://beta.reactjs.org/blog

I tried to figure out the source of the bugs and I think I'm able to fix it so please let me know if I can work on it I'd be happy to contribute.

![image](https://user-images.githubusercontent.com/44533239/155036789-e5084d5e-0ca7-4d41-a1ec-c5f0d2b51c22.png)
![image](https://user-images.githubusercontent.com/44533239/155036853-02e05d59-b6cc-40db-a204-d170e27dc9fa.png)

gaearon:
the blog page doesn't really "exist" yet. (did you find it by manually typing the url?)

the fix will be to use a separate `LayoutBlog` which wouldn't have the left sidebar, i think.

redhaam:
Hello Dan thank you for your reply.
Actually I accessed the blog page through a link within the `community` page. Anyway you can fix this later once you're done with  the content I guess.

gaearon:
i wouldn't mind merging a fix! if you'd like to work on it, try the approach i suggested — look at how `Layout*` files are mapped in `md-layout-loader.js` and try creating a separate layout alongside `LayoutLearn.tsx` and such, which should have no sidebar.

BartoszKlonowski:
I can see this issue is no longer valid, hence closing.
Thanks everyone!

PR #4373, opened 2022-02-22
[Beta] Fix Layout on Blog page

redhaam:
Hello 
This PR adresses a fix for the bug mentionned on #4370 
I've made a small change on the `Page` Layout to pass a prop about whether to display or not the sidebar since it's the base layout used in all pages. Also I prefered to not add the Navbar component for now since it isn't adaptable for desktop viewport and I don't have an idea about how it would look like in the design. 
What still need to be fixed :
- Link to `feed.xml` is broken.
- Reference links for other blogs in the docs are broken since the links on the old website end with `.html`  
- There's an error message on the console when accessing a not found page about the routes on the sidebar.

Looking forward for your review.

Screenshots : 
image
image

### DIFF
--- a/beta/src/pages/blog/all.tsx
+++ b/beta/src/pages/blog/all.tsx
@@ -19,12 +19,12 @@ export default function Archive() {
           <h1 className="text-5xl font-bold">Blog Archive</h1>
           <a
             href="/feed.xml"
-            className="p-2 betterhover:hover:bg-gray-20 transition duration-150 ease-in-out rounded-lg inline-flex items-center">
+            className="p-2 bg-card betterhover:hover:bg-secondary-button dark:bg-secondary-button-dark betterhover:hover:dark:bg-card-dark transition duration-150 ease-in-out rounded-lg inline-flex items-center">
             <IconRss className="w-5 h-5 mr-2" />
             RSS
           </a>
         </div>
-        <p className="text-gray-70 text-2xl">
+        <p className="text-gray-70 dark:text-gray-20 text-2xl">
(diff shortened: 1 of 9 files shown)


beta/src/pages/blog/all.tsx
```diff
@@ -67,5 +68,11 @@ export default function Archive() {
 Archive.displayName = 'Index';
 
 Archive.appShell = function AppShell(props: {children: React.ReactNode}) {
-  return <Page routeTree={blogIndexRecentRouteTree} {...props} />;
+  return (
+    <Page
+      displaySidebar={false}
```
harish-sethuraman, line 73: instead of adding `displaySidebar` as true to all layout pages except blog would it be good if we add `hideSidebar` to `Layoutblog` alone?

redhaam: Thank you for your review, So I created a new layout for blog pages now.

harish-sethuraman (commented):
This change looks amazing and works fine as far as I've checked. Also, I have one small suggestion we have so much space on the left since we do not have a navbar on the left. May be can we take up equal space from the left and the right TOC? Thoughts? 


Screenshot 2022-04-05 at 2 32 54 PM
 

Also please make sure we arent affecting any other page's styles since we have some changes in other layout pages. Other than that we are GTG.

redhaam:
Thank you @harish-sethuraman for you feedback, I can continue working on the side table of content. Perhaps, I need to know if there's already pre-made components or I can start working on it from scratch.

BartoszKlonowski:
Closing as per issue.
Thank you!

PR #4376, opened 2022-02-23
Fix bad link

rmschindler:


gaearon:
what's bad about it?

rmschindler:
On having a deeper look, the link I thought should replace the one at issue does not seem right either. Sorry for wasting time everyone.

PR #4377, opened 2022-02-23
Add diagrams to Quick Start

rickhanlonii:
## Overview

Adds new `<Diagram>` and `<DiagramGroup>` for embedding diagrams into the docs, and adds new diagrams to the Quick Start page as the initial diagrams on the beta site.

### Mobile (light)

Screen Shot 2022-02-23 at 4 06 53 PM

### Mobile (dark)

Screen Shot 2022-02-24 at 11 14 10 AM

### Desktop (light)

Screen Shot 2022-02-23 at 4 07 29 PM

### Desktop (dark)

Screen Shot 2022-02-24 at 11 13 53 AM


gaearon:
Any particular reason for PNG rather than SVG export?

rickhanlonii:
No reason, switched to svg.

gaearon:
nice, this looks a lot crispier

rickhanlonii:
Updated with better dark mode diagrams

gaearon:
WOAH

beta/src/pages/learn/index.md
```diff
@@ -389,27 +388,33 @@ Hooks are more restrictive than regular functions. You can only call Hooks *at t
...
-- MyApp (count: 3)
-  - MyButton
-  - MyButton
-  - MyButton
-```
+In this example, it is `MyApp`:
+
+<DiagramGroup>
+    <Diagram name="sharing_data_parent.svg" alt="TODO: Sharing data" height={770} width={820}>
+        Before clicking, count is stored in MyApp and passed down to both children as props.
```
gaearon, line 410: I think you should be able to write this without indentation. We generally prefer that in MDX. But you need newlines. So,

```js
<DiagramGroup>

<Diagram>

Before ...

</Diagram>

...

```

beta/src/pages/learn/index.md
```diff
@@ -389,27 +388,33 @@ Hooks are more restrictive than regular functions. You can only call Hooks *at t
...
-```js {1}
-- MyApp (count: 3)
-  - MyButton
-  - MyButton
-  - MyButton
-```
+In this example, it is `MyApp`:
+
+<DiagramGroup>
+    <Diagram name="sharing_data_parent.svg" alt="TODO: Sharing data" height={770} width={820}>
```
gaearon, line 409: Does this need an alt? The caption conveys what alt would've conveyed. Maybe this should be an empty alt.

beta/src/pages/learn/index.md
gaearon: Should we backtick technical names like `count` in the captions?

beta/src/pages/learn/index.md
```diff
@@ -389,27 +388,33 @@ Hooks are more restrictive than regular functions. You can only call Hooks *at t
...
+        Before clicking, count is stored in MyApp and passed down to both children as props.
+    </Diagram>
+    <Diagram name="sharing_data_parent_clicked.svg" alt="TODO: Sharing data" height={770} width={820}>
+        After clicking, count updates in MyApp and the update is passed to both children as props.
```
gaearon, line 413: "the update is passed" is a bit confusing. What's being passed is the new value of `count`. Can we slightly reword this to avoid the implication that an update can be passed?

gaearon (approved):
with nits

beta/src/pages/learn/index.md
rickhanlonii: Do you have a tip for how to make that work? Since it's a prop string and not markdown the back ticks just show as back ticks.

beta/src/pages/learn/index.md
rickhanlonii: I think they're subtly different. The alt should describe what the image contains for accessibility, where the subtitle is more of an explanation which you need to consume the diagram (either by viewing or reading the alt) to understand.

beta/src/pages/learn/index.md
rickhanlonii: Yeah, I was trying to avoid "state" since that seems to be avoided in this section. I'll clarify!

beta/src/pages/learn/index.md
gaearon: Right, it doesn't work because there are no newlines before and after. (MDX 1 is a bit confusing in that way.) If you format it like I wrote above, you should be able to use backticks inside.

rickhanlonii:
Feeling good about the latest state, will ship and iterate if needed!

PR #4390, opened 2022-02-26
Add diagrams for "Putting markup into JavaScript"

rickhanlonii:
## Overview

Adds diagrams for the "Putting markup into JavaScript" section of "Writing Markup with JSX".

### Light
Screen Shot 2022-02-26 at 4 54 42 PM

### Dark
Screen Shot 2022-02-26 at 4 54 48 PM


eps1lon (commented):
TL;DR: I think you meant `isLoggedIn` in the JavaScript diagram not `onLogin`

From the handlers mentioned in the JavaScript diagram only `onClick` and `onSubmit` are used in the component files. `onLogin` is only used in JavaScript, but not in any component file while Sidebar.js contains `isLoggedIn` which isn't mentioned in the JavaScript diagram.

rickhanlonii:
Nice catch, fixed!

PR #4393, opened 2022-02-28
updated addons-animation.md

Bakon:
Hello!

The docs/animation page is very dated and I hope to fix some issues of that page with this PR.
I'm aware that this add-on has become a package.

But I do think that this page should contain the same information as the [package](https://github.com/reactjs/react-transition-group/tree/v1-stable).

I can imagine novice / junior developers scratching their heads over this for longer than they have to be, because `ReactCSSTransitionGroup` is not a default export from `react-transition-group` (anymore).

In the first code example it uses the index of loop as a `key` for the component which was inconsistent with the other examples, while `item` is always unique anyway.

### DIFF
--- a/content/docs/addons-animation.md
+++ b/content/docs/addons-animation.md
@@ -12,19 +12,19 @@ redirect_from:
 
 > Note:
 >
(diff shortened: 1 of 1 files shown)


BartoszKlonowski:
Hello @Bakon, thanks for contributing!
Unfortunately the latest of react.dev documentation no longer mentiones the `TransitionGroup` so I will close this PR as no longer required.

Issue #4429, opened 2022-03-04
Add Social Banner with "Support for DONBASS. US stop delivering missiles to the battlefield for the sake of humanity."


nyepi-3055:
Donbass was part of Ukraine in south-eastern,where has been shelled by Ukraine since 2014.
A French journalist released a documentary called Donbass in 2017: https://www.youtube.com/watch?v=6Oh-IE2zmJc

So why not add Donbass?selective vocalization?Double standard?
I will give you one slogan:The people in Donbass lives matter!

Before delete this issue.Please [tell me why](https://www.youtube.com/watch?v=F_N6s4u3uvk) the world is so dirty?Why so-called open source community is so ridiculous?And which rule indicates that my issue is in violation?

hmmhmmhm:
Why do you think it will be deleted? Go ahead find an official account sponsoring Donbas and create a PR.

Currently, Russia is bombing Ukrainian civilians, and economic fundraising related to Russia is blocked by U.S. economic sanctions, so it is questionable whether we can legally find a channel that only sponsors civilians.

Wouldn't it be better to proceed to create PR not just arguing if you think this is humanitarian aid and you can prove it...?

nyepi-3055:
>Why do you think it will be deleted? Go ahead find an official account sponsoring Donbas and create a PR.

A pr about 'no politics in open source project' is deleted. It could be closed. Why delete it?

>so it is questionable whether we can legally find a channel that only sponsors civilians.

Russian civilians(students,musicians,athletes,scientists,journalists and even cats) are sanctioned by US and European countries, they are innocent too. Do you think Meta will aids these people? No,it is not political correctness. Instead, facebook mutes Russia civilians.

>Wouldn't it be better to proceed to create PR not just arguing if you think this is humanitarian aid and you can prove it...?

If somebody tell me 'support Donbass' pr will be merged.I would love to pr.

I think the react.js project just becomes a tool to help Meta do 'political correctness'. Many Open source projects not controlled by companies choose to be neutral. A react coreteam member said:
>**It is true that we have not responded to every humanitarian disaster. There is no single objective threshold of scale, and we’ve only started doing these banners recently. However, this is not a zero-sum activity.** First, we are only a single project, and there are other projects which may choose different causes (or not support any causes at all). Second, getting the message out about one crisis does not hurt people affected by a different crisis. This is why the “whataboutism” is mostly a **distraction**, and we will not spend time seriously entertaining it.'

[...]

PR #4438, opened 2022-03-07
docs: Change to a more neutral message of ukraine support. #4429

hmmhmmhm:
https://github.com/reactjs/reactjs.org/issues/4429#issuecomment-1060297476

Some people have asked for a change to a neutral message, which seems acceptable. As the request, can be expect the dispute to end by making it clear that the campaign aims to help innocent people.

I think it will help alleviate the some dispute if accept this.

### DIFF
--- a/src/components/SocialBanner/SocialBanner.js
+++ b/src/components/SocialBanner/SocialBanner.js
@@ -16,8 +16,8 @@ const linkProps = {
   rel: 'noopener',
 };
 
-const bannerText = 'Support Ukraine 🇺🇦 ';
-const bannerLink = 'Help Provide Humanitarian Aid to Ukraine.';
+const bannerText = 'Support Ukraine people 🇺🇦 ';
+const bannerLink = 'Help Provide Humanitarian Aid to Ukraine people.';
 
 export default function SocialBanner() {
   return (


hmmhmmhm:
I just signed the CLA.

hellotaotao:
@hmmhmmhm  I have read the above listed discussions and appreciate your responses in those.
However, imaging when I see a murder is happening, I would stand up and say "STOP!", rather than murmuring something like "let's work together towards a better world where people can live longer blah blah......"

And what this PR is doing, is the latter.

We already have a clear message in response to the invasion. I don't see a good reason to change it.

gaearon:
To expand on this a little, "humanitarian aid" already includes "supporting civilians".

Wikipedia definition of "humanitarian aid":

>Humanitarian aid is material and logistic assistance to people who need help. It is usually short-term help until the long-term help by the government and other institutions replaces it. Among the people in need are the homeless, refugees, and victims of natural disasters, wars, and famines. Humanitarian relief efforts are provided for humanitarian purposes and include natural disasters and man-made disasters. The primary objective of humanitarian aid is to save lives, alleviate suffering, and maintain human dignity. 

This seems more precise than "supporting civilians".

Similarly, it seems unnecessary to hide the name of the country being invaded and whose civilians are becoming refugees, if we all know which country it is. Therefore, we're happy with the existing message.

gaearon:
See also my response in https://github.com/reactjs/reactjs.org/issues/4433#issuecomment-1060759889.

PR #4448, opened 2022-03-10
fix: use `globalThis` over `global`

SimenB:
`global` only exists in node (and `jest-environment-jsdom` [fakes it](https://github.com/facebook/jest/blob/54420eb51baecaba7dff293770f8cb2579825731/packages/jest-environment-jsdom/src/index.ts#L71-L72) for backwards compat). The specified name is `globalThis`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis.

This works in Node 12 and up (current lowest supported version of Node), and not IE (which you're dropping anyways), so should be fine.

---

As an aside, happy to see you removing `typeof jest` check - that does not exist in native ESM, so would have been broken for those use cases anyways 🙂

### DIFF
--- a/content/blog/2022-03-08-react-18-upgrade-guide.md
+++ b/content/blog/2022-03-08-react-18-upgrade-guide.md
@@ -227,11 +227,11 @@ When you first update your tests to use `createRoot`, you may see this warning i
 
 > The current testing environment is not configured to support act(...)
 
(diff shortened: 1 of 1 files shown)


rickhanlonii:
Thanks!

PR #4454, opened 2022-03-12
Add Xuan to team page

Huxpro:
Whola!


acdlite:
Btw when you visit New York we should play tennis, either on or off a table

Huxpro:
Oh I just realized that the Beta site has its own copy of the page: https://github.com/reactjs/reactjs.org/blob/main/beta/src/pages/community/meet-the-team.md

One minute let me amend

PR #4455, opened 2022-03-12
Beta: Fix Feedback being shifted down to invisible

Huxpro:
`<SocialBanner/>` shifted `<Feedback />` down to 40px on `>lg` screen.

Screen Shot 2022-03-11 at 6 01 56 PM
 
This diff fixed it by using `calc`. Not sure if there is a better way but it works
Screen Shot 2022-03-11 at 6 03 41 PM
:


harish-sethuraman:
https://github.com/reactjs/reactjs.org/pull/4451 I raised a pr for the same with same fix. 😅

harish-sethuraman:
I've merged the PR that had the same fix that was raised yesterday. So closing this PR. 😅

PR #4457, opened 2022-03-14
Update invalid-aria-prop.md

DeltaZN:
Actually, they are rendered to DOM


eps1lon (approved):
Good catch thanks. Could you sign the CLA so that we can merge this?

PR #4468, opened 2022-03-17
[18] Update ReactDOMClient docs

rickhanlonii:
## Overview

Some changes/mistakes since the last RC.


salazarm (approved):
good catch by seb

content/docs/reference-react-dom.md
```diff
@@ -85,14 +85,14 @@ Same as [`createRoot()`](#createroot), but is used to hydrate a container whose
 ### `createPortal()` {#createportal}
 
 ```javascript
-ReactDOM.createPortal(child, container)
+ReactDOMClient.createPortal(child, container)
```
sebmarkbage, line 88: All of these are *not* on ReactDOMClient. Only hydrateRoot and createRoot are.

content/docs/reference-react-dom.md
```diff
@@ -55,25 +55,25 @@ root.unmount();
...
+ReactDOMClient.hydrateRoot(element, container[, options])
 ```
 
 Same as [`createRoot()`](#createroot), but is used to hydrate a container whose HTML contents were rendered by [`ReactDOMServer`](/docs/react-dom-server.html). React will attempt to attach event listeners to the existing markup.
 
-`hydrateRoot` accepts two optional callbacks as options:
-- `onDeleted`: callback called when content is deleted.
-- `onHydrated`: callback called after hydration completes.
+`hydrateRoot` accepts two options:
+- `onRecoverableError`: optional callback called when React automatically recovers from errors during hydration.
```
sebmarkbage, line 75: This is not just during hydration. It's also for concurrency issues.

content/docs/reference-react-dom.md
```diff
@@ -55,25 +55,25 @@ root.unmount();
...
-ReactDOM.hydrateRoot(element, container[, options])
+ReactDOMClient.hydrateRoot(element, container[, options])
 ```
 
 Same as [`createRoot()`](#createroot), but is used to hydrate a container whose HTML contents were rendered by [`ReactDOMServer`](/docs/react-dom-server.html). React will attempt to attach event listeners to the existing markup.
 
-`hydrateRoot` accepts two optional callbacks as options:
-- `onDeleted`: callback called when content is deleted.
-- `onHydrated`: callback called after hydration completes.
+`hydrateRoot` accepts two options:
```
sebmarkbage, line 74: These options can also be passed createRoot.

PR #4471, opened 2022-03-18

krishnasaga:
It is not a mistake in the documentation at all but It is getting misinterpreted by few developers as "you should be doing only in rare case" and not knowing what the rare case is :) If you think that will help to make it more **precise** in humble sense  :) you can accept the change


harish-sethuraman:
I guess it is better that we leave it as such and it makes sense to not change `rare` to `a few`. Thanks for sending a PR.

PR #4474, opened 2022-03-19
[18] Add docs for flushSync

rickhanlonii:
## Overview

I'm sure there's a lot more to say here. Opening as an MVP, feel free to suggest more content.

### DIFF
--- a/content/docs/reference-react-dom.md
+++ b/content/docs/reference-react-dom.md
@@ -68,9 +68,11 @@ Force React to flush any updates inside the provided callback synchronously. Thi
 > 
 > `flushSync` can have a significant impact on performance. Use sparingly.
 > 
-> `flushSync` will force Suspense boundaries to show their `fallback` state.
+> `flushSync` may force pending Suspense boundaries to show their `fallback` state.
 > 
 > `flushSync` may also run pending effects and synchronously apply any updates they contain before returning.
+> 
+> `flushSync` may also flush updates outside the callback when necessary to flush the updates inside the callback. For example, if there are pending updates from a click, React may flush those before flushing the updates inside the callback.
 
 ## Legacy Reference
 ### `render()` {#render}
--- a/content/docs/reference-react-dom.md
+++ b/content/docs/reference-react-dom.md
@@ -26,7 +26,7 @@ The `react-dom` package also provides modules specific to client and server apps
 
 The `react-dom` package exports these methods:
 - [`createPortal()`](#createportal)
-
+- [`flushSync()`](#flushsync)
 
 These `react-dom` methods are also exported, but are considered legacy:
 - [`render()`](#render)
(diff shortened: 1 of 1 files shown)


content/docs/reference-react-dom.md
```diff
@@ -55,6 +55,21 @@ createPortal(child, container)
...
 
 Creates a portal. Portals provide a way to [render children into a DOM node that exists outside the hierarchy of the DOM component](/docs/portals.html).
+
+### `flushSync()` {#flushsync}
+
+```javascript
+flushSync(callback)
+```
+
+Force React to flush updates synchronously. This method is useful for being able to read the value of `state` immediately.
```
sebmarkbage, line 65: This a bit misleading since the easiest way to read the value of state immediately is just read what you set it to. I don't think we should encourage this as a workaround for that.

It's more about reading the outcomes. Almost always from the DOM.

It could use an explanation what the purpose of the callback is since callbacks are typically associated with async operation completing, which this by definition is not.

It's meant to form a grouping of any setStates that you need to flush synchronously.

content/docs/reference-react-dom.md
```diff
@@ -55,6 +55,21 @@ createPortal(child, container)
...
+
+> Note:
+> 
+> `flushSync` can have a significant impact on performance. Use sparingly.
+> 
+> `flushSync` may also run pending effects and synchronously apply any updates they contain before returning.
```
sebmarkbage, line 71: It can also worse Suspense boundaries to show a "fallback" state when they otherwise wouldn't have.

content/docs/reference-react-dom.md
rickhanlonii: Makes sense, I was being a little value about the callback because it's not necessarily only the updates passed in that are flushed, correct? I.e. just calling `flushSync()` without a callback will flush work?

sebmarkbage: Yea, if you happen to know that they're also scheduled in the sync lane. You can force any pending discrete ones to flush.

However, it's a little sketchy to assume that the context you're executing within is already discrete so adding more setStates before it might not be a good idea. Since you could've been wrapped in a startTransition.

```js
// Reasonable if that's your use case
flushSync();
```

```js
// Bad
setState(true);
flushSync();
```

```js
// Good
flushSync(() => setState(true));
```

rickhanlonii: I'm trying to avoid explaining discrete updates since we haven't really explained it elsewhere, how about a note like:

> `flushSync` may also flush updates outside of the callback when necessary to flush the updates provided.

content/docs/reference-react-dom.md
```diff
@@ -55,6 +55,23 @@ createPortal(child, container)
...
+flushSync(callback)
+```
+
+Force React to flush any updates inside the provided callback synchronously. This method is useful for being able to read the result of those updates immediately.
+
+> Note:
+> 
+> `flushSync` can have a significant impact on performance. Use sparingly.
+> 
+> `flushSync` will force Suspense boundaries to show their `fallback` state.
```
sebmarkbage, line 71: *can ... if they're still loading.

content/docs/reference-react-dom.md
sebmarkbage: Sounds good but that part might even need explanation.

The important part is that you put the updates you want to do inside the closure.

eps1lon (approved):

PR #4476, opened 2022-03-20
[18] Update browser requirements

rickhanlonii:
## Overview

Adds requirement for browsers using microtask-based Promise implementation.


eps1lon (commented):
`Symbol` polyfill is now also required (https://github.com/facebook/react/pull/23348) as well as `Object.assign` (https://github.com/facebook/react/pull/23351).

- Just with IE 11 as a transpilation target: https://vercel.com/eps1lon/csb-6dp95j/n77L71izekRj7hJJkbn3wYgapaY1
- +Symbol polyfill (crashing): https://csb-6dp95j-ro3etzcoc-eps1lon.vercel.app/
- +Object.assign polyfill (not crashing): https://csb-6dp95j-aa31xubxu-eps1lon.vercel.app/ 

The CodeSandbox didn't include Map, Set and Promise. Seems like they're either not required for just a render or indirectly polyfilled.

content/docs/reference-javascript-environment-requirements.md
```diff
@@ -6,9 +6,11 @@ category: Reference
...
+React 18 depends on modern browser features including [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise), [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) and [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set).
 
-A polyfilled environment for React 16 using core-js to support older browsers might look like:
+If you support older browsers and devices which do not provide a native microtask-based Promise implementation such as Internet Explorer, we recommend you stay with React 17, because microtasks cannot be polyfilled without impacting performance in React 18.
```
sebmarkbage, line 11: The main issue isn't performance but the semantics of controlled/sequential things. For example, you might get the second click before the first click flushes. Most controlled text inputs probably work because it's built in to React's event system.

So it's really likely not actually that big of a deal.

content/docs/reference-javascript-environment-requirements.md
```diff
@@ -6,9 +6,11 @@ category: Reference
...
+React 18 depends on modern browser features including [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise), [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) and [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set).
 
-A polyfilled environment for React 16 using core-js to support older browsers might look like:
+If you support older browsers and devices which do not provide a native microtask-based Promise implementation such as Internet Explorer, we recommend you stay with React 17, because microtasks cannot be polyfilled without impacting performance in React 18.  
+
+To support older browsers which do not include `Map` or `Set`, consider using [core-js](https://github.com/zloirock/core-js):
```
sebmarkbage, line 13: Do we actually recommend core-js for anything anymore?

`Map` and `Set` were already in IE11, at least for how we used them.

So I think it's only `Symbol` and `Object.assign` that we now require polyfilling. Is `core-js` what we'd recommend for that?

content/docs/reference-javascript-environment-requirements.md
rickhanlonii: Ok cool, I removed the caveat.

content/docs/reference-javascript-environment-requirements.md
rickhanlonii: Updated! I think the best practice now is to configure Browserlist if available, otherwise provide the polyfill directly. So I removed the core-js example and explained the current state.

rickhanlonii:
Thanks @eps1lon, added `Symbol` and `Object.assign, removed `Map` and `Set`, and rewrote the content.

eps1lon (approved):

PR #4477, opened 2022-03-20
[18] Add stubs for new API references

rickhanlonii:
## Overview

Add stubs to fill in for the new APIs in React 18.

### DIFF
--- a/content/docs/hooks-reference.md
+++ b/content/docs/hooks-reference.md
@@ -24,6 +24,12 @@ If you're new to Hooks, you might want to check out [the overview](/docs/hooks-o
   - [`useImperativeHandle`](#useimperativehandle)
   - [`useLayoutEffect`](#uselayouteffect)
   - [`useDebugValue`](#usedebugvalue)
+  - [`useDeferredValue`](#usedeferredvalue)
+  - [`useTransition`](#usetransition)
+- [Library Hooks](#library-hooks)
+  - [`useId`](#useid)
+  - [`useInsertionEffect`](#useinsertioneffect)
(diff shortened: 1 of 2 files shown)


content/docs/hooks-reference.md
```diff
@@ -508,3 +514,59 @@ For example a custom Hook that returned a `Date` value could avoid calling the `
...
+const [isPending, startTransition] = useTransition();
+```
+
+TODO: description
+
+## Library Hooks {#library-hooks}
+
+The following Hooks are provided for library authors to integrate libraries deeply into the React model, and are not typically used in application code.
+
+### `useId` {#useid}
```
sebmarkbage, line 538: This one is something you might need to use in application code a bit more often when you have disconnected labels/forms for example. Libraries might be able to hide it but not always. So probably worth promoting this higher and mention the label/form field case as an example explicitly so it's not confused for other use cases.

content/docs/hooks-reference.md
```diff
@@ -508,3 +514,59 @@ For example a custom Hook that returned a `Date` value could avoid calling the `
...
+const id = useId(value);
+```
+
+TODO: description
+
+> Note:
+> 
+> TODO: identifierPrefix
+
+### `useInsertionEffect` {#useinsertioneffect}
```
sebmarkbage, line 550: Maybe we put this last since it's less common than useSES.

rickhanlonii: Yeah, and @acdlite made a good point that `useId` isn't strictly 'library' so I moved that up as well.

PR #4482, opened 2022-03-22
[18] Update versions page

rickhanlonii:
## Overview

Updates the versions page to call out the link to React 17.

Also removes the old broken documentation links until we figure out what to do with them.

## Screen
Screen Shot 2022-03-22 at 1 57 43 PM

### DIFF
--- a/content/versions.yml
+++ b/content/versions.yml
@@ -2,6 +2,7 @@
   changelog: https://github.com/facebook/react/blob/main/CHANGELOG.md
 - title: '17.0.2'
   changelog: https://github.com/facebook/react/blob/main/CHANGELOG.md#1702-march-22-2021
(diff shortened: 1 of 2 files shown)


slorber:
@rickhanlonii did these links go offline without notice? I'm facing this problem on the Docusaurus versioning page: https://github.com/facebook/docusaurus/issues/8634
It seems we wrongly assumed those links will stay online forever? 😅 

Didn't Netlify provide any solution to restore them? 
Did users complain of the inability to access older docs versions?

Looking for an eventual new solution because these deployment permalinks were the suggestion I made to a few Facebook sites (including Jest and React-Native) to "archive" a docs version forever, but apparently it's not very reliable 😓

PR #4484, opened 2022-03-23
[18] Updated Suspense doc with behavior during SSR and Hydration

salazarm:
Not sure if this is the right amount of detail


content/docs/reference-react.md
```diff
@@ -370,9 +370,19 @@ It is documented in our [code splitting guide](/docs/code-splitting.html#reactla
...
+#### `React.Suspense` in SSR
+During SSR trees wrapped in Suspense are rendered synchronously as deeply as possible. If the boundary never suspends 
+then the tree renders as though the boundary were never there. If a component does suspend then we schedule a low 
+priority task to render the fallback instead. If the component unsuspends before we flush the fallback then we send 
+down the actual content and throw away the fallback.
+
+#### `React.Suspense` during hydration
+Suspense boundaries depend on their parent boundaries being hydrated before they can hydrate, but they can hydrate independently from sibling boundaries. Events on a boundary before its hydrated will cause the boundary to hydrate at 
+a higher priority than neighboring boundaries but discrete events (eg. click) will not be replayed upon hydration. Only 
+the last continuous event (`focusin`, `mousein`, `dragenter`, `mouseover`, `pointerover`, `gotpointercapture`, `hover`)
```
eps1lon, line 382: I believe you meant `mouseenter` not `mousein`?

Is `focusin` a continious event?

I think it was brought up earlier whether discrete vs continous events needs a separate section. If this is something devs should be aware of, then it's probably better to link to a list of discrete and continous events.

Because intuitively a `gotpointercapture` is a discrete thing because a pointer being captured is a discrete state from my understanding (pointer is captured or not). Same for `dragenter` I guess. For `mouseover` it makes sense to consider it continous. 

Either way, I think this illustrates that it isn't clear what events are considered "discrete" and which are "continious".

gaearon: This is why I don't like "discrete" as a word. Our usage of "discrete" is "user _meant_ to do this, and there could be a mistake if we batch it with another same event". Our usage of "continuous" is "user _did not_ intentionally mean to do _this particular_ event, and so we can skip/batch it up with next event of the same type". At least to my understanding.

sebmarkbage: It doesn't really have anything to do with the user intent. That's just something we use as an explanation.

Btw, I don't think we should use the term replay. Because actually, we probably shouldn't actually replay the same event. It's more about reissuing an observation of the current state.

This is like if something is animating below your cursor. Even if you don't move the mouse it should ideally trigger mouseenter/leave anyway as the thing below moves in and out to trigger hover effects. CSS hovers do this.

Focus is a little special since when it's accompanied by other events like key events, it effectively becomes discrete as a result of how it's controlled.

sebmarkbage: For purposes of the hydration, these are not the same thing as the "discrete" and "continuous" hints that we use to inform priorities and batching.

I think we shouldn't couple them and probably not use the same terminology.

These events are things where you can issue new events to inform what the state of the world is. If the user twitched a little, it's the same enter/leave happening on a moving target. The goal in the end is to inform what the current state of the world is.

`focusin` is like saying "activeElement is now on this". So restating that "activeElement is now on this" is fine.

Similarly, `gotpointercapture` is like restating that you currently has the pointer captured.

sebmarkbage: There's no `hover` event is there?

salazarm: We emulate it off `mousenter`

eps1lon: "persistent events" makes sense to me now. Thanks for clarifying!

sebmarkbage: @salazarm Is that an FB-only thing?

salazarm: @sebmarkbage oops it was a mistake. I updated the events

content/docs/reference-react.md
```diff
@@ -370,9 +370,17 @@ It is documented in our [code splitting guide](/docs/code-splitting.html#reactla
...
->
->`React.lazy()` and `<React.Suspense>` are not yet supported by `ReactDOMServer`. This is a known limitation that will be resolved in the future.
+#### `React.Suspense` in SSR
+During SSR trees wrapped in Suspense are rendered synchronously as deeply as possible. If the boundary never suspends 
+then the tree renders as though the boundary were never there. If a component does suspend then we schedule a low 
+priority task to render the fallback instead. If the component unsuspends before we flush the fallback then we send 
+down the actual content and throw away the fallback.
+
+#### `React.Suspense` during hydration
+Suspense boundaries depend on their parent boundaries being hydrated before they can hydrate, but they can hydrate independently from sibling boundaries. Events on a boundary before its hydrated will cause the boundary to hydrate at
```
eps1lon, line 380: I was struggling with the "they" references. Does it sound clearer to split this sentence up into two? 
```suggestion
Suspense boundaries depend on their parent boundaries being hydrated before they can hydrate. Suspense boundaries can hydrate independently from sibling boundaries. Events on a boundary before it's hydrated will cause the boundary to hydrate at 
```

eps1lon (approved):

PR #4485, opened 2022-03-23
[18] renderToPipeableStream doc

salazarm:


rickhanlonii:
Just curious, what about `renderToReadableStream`, and the changes to the other hooks?

content/docs/reference-react-dom-server.md
```diff
@@ -54,8 +56,41 @@ Similar to [`renderToString`](#rendertostring), except this doesn't create extra
...
 
 * * *
+### `renderToPipeableStream()` {#rendertopipeablestream}
 
-### `renderToNodeStream()` {#rendertonodestream}
+```javascript
+ReactDOMServer.renderToPipeableStream(element, options)
+```
+
+Render a React element to its initial HTML. Returns a [Control object](https://github.com/facebook/react/blob/3f8990898309c61c817fbf663f5221d9a00d0eaa/packages/react-dom/src/server/ReactDOMFizzServerNode.js#L49-L54) that allows you to pipe the output or abort the request. Fully supports Suspense and streaming of HTML with "delayed" content blocks "popping in" later through javascript execution. [Read more](https://github.com/reactwg/react-18/discussions/37)
```
sebmarkbage, line 65: We should rename that type. I think of it as a "PipeableStream" object. It's kind of like Node's "ReadableStream" except limited to the pipe function and with the addition of the `abort()`.

content/docs/reference-react-dom-server.md
```diff
@@ -54,8 +56,41 @@ Similar to [`renderToString`](#rendertostring), except this doesn't create extra
...
+
+* * *
+
+### `renderToReadableStream()` {#rendertoreadablestream}
+
+```javascript
+    ReactDOMServer.renderToReadableStream(element, options);
+```
+
+Render a React element to its initial HTML. Returns a [Readable Stream](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream). Fully supports Suspense but not streaming of HTML. [Read more](https://github.com/reactwg/react-18/discussions/127)
```
sebmarkbage, line 82: This is the API that canonical for streaming HTML. The one that doesn't stream HTML is `renderToNodeStream`.

content/docs/reference-react-dom-server.md
```diff
@@ -54,8 +56,41 @@ Similar to [`renderToString`](#rendertostring), except this doesn't create extra
...
+
+If you call [`ReactDOM.hydrateRoot()`](/docs/react-dom-client.html#hydrateroot) on a node that already has this server-rendered markup, React will preserve it and only attach event handlers, allowing you to have a very performant first-load experience.
+
+> Note:
+>
+> Server-only. This API is not available in the browser.
```
sebmarkbage, line 88: This actually not server-only. https://github.com/facebook/react/blob/main/packages/react-dom/server.browser.js#L40

It's more intended for Service Workers specifically though.

content/docs/reference-react-dom-server.md
```diff
@@ -54,8 +56,41 @@ Similar to [`renderToString`](#rendertostring), except this doesn't create extra
...
+ReactDOMServer.renderToPipeableStream(element, options)
+```
+
+Render a React element to its initial HTML. Returns a [Control object](https://github.com/facebook/react/blob/3f8990898309c61c817fbf663f5221d9a00d0eaa/packages/react-dom/src/server/ReactDOMFizzServerNode.js#L49-L54) that allows you to pipe the output or abort the request. Fully supports Suspense and streaming of HTML with "delayed" content blocks "popping in" later through javascript execution. [Read more](https://github.com/reactwg/react-18/discussions/37)
+
+If you call [`ReactDOM.hydrateRoot()`](/docs/react-dom-client.html#hydrateroot) on a node that already has this server-rendered markup, React will preserve it and only attach event handlers, allowing you to have a very performant first-load experience.
+
+> Note:
+>
+> Server-only. This API is not available in the browser.
```
sebmarkbage, line 71: It's worth highlighting that this is a Node.js specific API and that "modern server environments should use `renderToReadableStream` instead.

sebmarkbage:
It's probably worth inlining the examples from https://github.com/reactwg/react-18/discussions/127 it has a lot of the details in just that code.

content/docs/reference-react-dom-server.md
gaearon: Wait.. do we not consider Node.js modern anymore? 😄

PR #4486, opened 2022-03-23
[18] Add docs for useInsertionEffect

rickhanlonii:


content/docs/hooks-reference.md
```diff
@@ -599,8 +599,8 @@ const selectedField = useSyncExternalStore(
 useInsertionEffect(didUpdate);
 ```
 
-TODO: description
+The signature is identical to `useEffect`, but it fires synchronously _before_ all DOM mutations. Use this to inject styles into the DOM before reading layout in [`useLayoutEffect`](#uselayouteffect). Since this hook is limited in scope, this hook does not have access to refs and cannot schedule updates.
```
eps1lon, line 602: Does it currently warn if you update state in `useInsertionEffect` or are updates during this phase just dropped?

acdlite: It should. Want to submit a PR?

PR #4487, opened 2022-03-23
[18] Add docs for useSyncExternalStore

rickhanlonii:


content/docs/hooks-reference.md
```diff
@@ -550,14 +550,48 @@ The following Hooks are provided for library authors to integrate libraries deep
...
+const state = useSyncExternalStore(subscribe, getSnapshot[, getServerSnapshot]);
 ```
 
-TODO: description
+`useSyncExternalStore` is a hook recommended for reading and subscribing from external data sources in a way that's compatible with concurrent rendering features like selective hydration and time slicing.
+
+This method returns the value of the store and accepts three arguments:
+- `subscribe`: function to register a callback that is called whenever the store changes.
+- `getSnapshot`: function that returns the current value of the store.
+- `getServerSnapshot` (optional): function that returns the snapshot used during server rendering.
```
acdlite, line 561: It's optional but the consequence of omitting it is that it will de-opt the nearest Suspense boundary to client rendering. So it's optional, but not if you want to do server rendering.

acdlite (approved):

content/docs/hooks-reference.md
rickhanlonii: Removed!

PR #4488, opened 2022-03-24
[18] Add docs for useId

rickhanlonii:


content/docs/hooks-reference.md
```diff
@@ -534,14 +534,50 @@ TODO: description
...
+  return (
+    <>
+      <label htmlFor={id}>Do you like React?</label>
+      <input id={id} type="checkbox" name="react"/>
+    </>
+  );
+};
+```
+
+For multiple IDs in the same component, `useId` supports appending a suffix:
```
acdlite, line 556: If I were reading this, I would infer that useId might not return a normal string, because why else would docs explain that it supports appending?

acdlite (approved):

rickhanlonii:
Thanks @acdlite, fixed.

PR #4500, opened 2022-03-27
[18] Update suspense docs for unexpected fallbacks

rickhanlonii:
From https://github.com/reactjs/rfcs/pull/213

### DIFF
--- a/content/docs/code-splitting.md
+++ b/content/docs/code-splitting.md
@@ -139,6 +139,52 @@ function MyComponent() {
 }
 ```
 
+### Unexpected fallbacks
+Any component may suspend as a result of rendering, even components that were already shown to the user. In order for screen content to always be consistent, if an already shown component suspends, React has to hide its tree up to the closest <Suspense> boundary. However, from the user's perspective, this can be disorienting.
+
+Consider this tab switcher:
+
+```js
+import React, { Suspense } from 'react';
+import Tabs from './Tabs';
+import Glimmer from './Glimmer';
+
+const Comments = React.lazy(() => import('./Comments'));
+const Photos = React.lazy(() => import('./Photos'));
+
+function MyComponent() {
+  const [tab, setTab] = React.useState('photos');
+  
+  function handleTabSelect(tab) {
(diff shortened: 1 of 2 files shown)


content/docs/code-splitting.md
```diff
@@ -139,6 +139,52 @@ function MyComponent() {
...
+  const [tab, setTab] = React.useState('photos');
+  
+  function handleTabSelect(tab) {
+    setTab(tab);
+  };
+
+  return (
+    <div>
+      <Tabs onTabSelect={handleTabSelect} />
+      <Suspense fallback={<div>Loading...</div>}>
```
hamlim, line 165: Should `fallback` here render the `Glimmer` component? It's referenced below and imported in this code snippet, but never used

content/docs/code-splitting.md
```diff
@@ -139,6 +139,52 @@ function MyComponent() {
...
+        {tab === 'photos' ? <Photos /> : <Comments />}
+      </Suspense>
+    </div>
+  );
+}
+
+```
+
+In this example, if tab gets set to from `'photos'` to `'comments'`, but `Comments` suspends, the user will see a glimmer. This makes sense because the user no longer wants to see `Photos`, the `Comments` component is not ready to render anything, and React needs to keep the user experience consistent, so it has no choice but to show the `Glimmer` above.
```
hamlim, line 174:
```suggestion
In this example, if tab gets set from `'photos'` to `'comments'`, but `Comments` suspends, the user will see a glimmer. This makes sense because the user no longer wants to see `Photos`, the `Comments` component is not ready to render anything, and React needs to keep the user experience consistent, so it has no choice but to show the `Glimmer` above.
```

Seemed like an additional `to` snuck in there

rickhanlonii: Thanks!

content/docs/code-splitting.md
rickhanlonii: Nice catch, fixed!

acdlite (approved):

PR #4507, opened 2022-03-29
changed copyright header facebook to meta

Vishal-beep136:
changed copyright header facebook to meta


harish-sethuraman:
@gaearon This seems invalid anyways since we do not want to change copyright? or if we want to do so we can do it in another PR? like 👇🏻 
- https://github.com/reactjs/reactjs.org/pull/4543

PR #4518, opened 2022-03-29
Fix useTransition document link

t73liu:
Fixes the "See docs here" link. Currently routes to Not Found page


gaearon:
already merged via another PR, thanks

PR #4520, opened 2022-03-30
Fix usage of useDeferredValue

SukkaW:
Since the very beginning of the `useDeferredValue` being introduced from Facebook into react back in https://github.com/facebook/react/pull/17058, the usage of `useDeferredValue` is always `const deferredValue = useDeferredValue(...args);`, not `const [deferredValue] = useDeferredValue(...args);`.

The typo is introduced by @rickhanlonii in https://github.com/reactjs/reactjs.org/pull/4499


gaearon:
thanks :)

PR #4523, opened 2022-03-30
Update CDN links for React 18

Chalarangelo:
The CDN links referred to React v17.x, so I updated them to v18.x.


harish-sethuraman (approved):
LGTM

harish-sethuraman (commented):
We also have other pages with these CDN links. Can you change that as well? 

Like https://reactjs.org/docs/optimizing-performance.html#single-file-builds

and in the pages in the beta directory as well?

PR #4528, opened 2022-03-30
Update CDN links for React 18 on a few more pages

Chalarangelo:
Update the CDN links to v18 on another page.


harish-sethuraman (commented):
We also have other pages with these CDN links. Can you change that as well?

Like https://reactjs.org/docs/optimizing-performance.html#single-file-builds

and in the pages in the beta directory as well?

Chalarangelo:
@harish-sethuraman I think this is all of them. Can you take another look?

harish-sethuraman:
Thanks for the PR :)

Issue #4526, opened 2022-03-30
add support for PWA


dewanshDT:
The reactjs docs website does not have support for PWA, It would be better if is added as it can be then installed on any device

dewanshDT:
can I work on this feature

harish-sethuraman:
Feel free to send PRs. We can take it from there 😄

dewanshDT:
please check this out #4530

PR #4530, opened 2022-03-30
PWA support added successfully #4526

dewanshDT:
- [x] service worker added
- [x] manifest file added
gatsby


dewanshDT:
@harish-sethuraman please can you help me with unsuccessful checks

harish-sethuraman:
```
error gatsby-plugin-manifest@4.11.0: The engine "node" is incompatible with this module. Expected version ">=14.[15](https://github.com/reactjs/reactjs.org/runs/5758858121?check_suite_focus=true#step:4:15).0". Got "12.22.10"
```

Seems like the node version of the lint analyzer is lower(`12.22.10`) than the actual expected 14 for the library

dewanshDT:
What should I do?

gaearon (changes requested):
I'm not sure we actually want to do this.

In the past I've seen very bad experiences with service workers, including cases where things were getting cached forever, and it was impossible to get them updated properly. I don't think we want to merge this in unless we understand very clearly what the tradeoffs are.

dewanshDT:
@gaearon can you suggest me something else to work on

PR #4535, opened 2022-04-01
Linted files according to the .prettierrc

Harry-Hopkinson:
Linted files according to the .prettierrc for code clarity.


gaearon:
Hi, thanks but we don't take purely stylistic changes. Prettier is meant to be automated, if something in our automation is broken then that part is what needs to be fixed.

Harry-Hopkinson:
Okay thanks

PR #4541, opened 2022-04-03
Updated typo of createRoot and hydrateRoot syntax

nitinajaydixit:
According to their implementations :

export function createRoot(container: Element | DocumentFragment, options?: RootOptions): Root;

export function hydrateRoot(
    container: Element | Document,
    initialChildren: React.ReactChild | Iterable<React.ReactNode>,
    options?: HydrationOptions,
): Root;

### DIFF
--- a/content/docs/reference-react-dom-client.md
+++ b/content/docs/reference-react-dom-client.md
@@ -38,7 +38,7 @@ React supports all modern browsers, although [some polyfills are required](/docs
 ### `createRoot()` {#createroot}
 
 ```javascript
-createRoot(container[, options]);
+createRoot(container, [ options ]);
 ```
 
 Create a React root for the supplied `container` and return the root. The root can be used to render a React element into the DOM with `render`:
@@ -71,7 +71,7 @@ root.unmount();
 ### `hydrateRoot()` {#hydrateroot}
 
 ```javascript
-hydrateRoot(element, container[, options])
+hydrateRoot(element, container, [ options ])
 ```
 
 Same as [`createRoot()`](#createroot), but is used to hydrate a container whose HTML contents were rendered by [`ReactDOMServer`](/docs/react-dom-server.html). React will attempt to attach event listeners to the existing markup.


content/docs/reference-react-dom-client.md
```diff
@@ -38,7 +38,7 @@ React supports all modern browsers, although [some polyfills are required](/docs
 ### `createRoot()` {#createroot}
 
 ```javascript
-createRoot(container[, options]);
+createRoot(container, [ options ]);
```
eps1lon, line 41: I think this is mostly an issue with where you got introduced first to this syntax for optional parameters. I'm mostly familiar with `[, options]` but I've also seen `, [options]` e.g. https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/named-and-optional-arguments)

I've seen plenty of people struggling with this syntax (also because it looks like array syntax).

MDN does not have special syntax and just states in prose that a parameter is optional (e.g.`eventInitDict` in https://developer.mozilla.org/en-US/docs/Web/API/SubmitEvent/SubmitEvent)

@gaearon Is this syntax for optional parameters still in use in the Beta docs or do you know if you settled on a specific syntax for optional parameters?

gaearon: In beta we just write multiple signatures after newlines, I think.

harish-sethuraman: I believe we aren’t currently using the same syntax in beta docs. Also the syntax in the older documentation looks fine. We use  [, and not ,[ to denote optional params just like mdn docs

eps1lon: > We use [, and not ,[ to denote optional params just like mdn docs

Do you have an example for that? The articles I linked don't have special syntax e.g. https://developer.mozilla.org/en-US/docs/Web/API/SubmitEvent/SubmitEvent.

The important part here is that the chose syntax should not be ambigious. Not using any special syntax would fit that requirement IMO. Just like using multiple signatures.

harish-sethuraman: I've seen the same in multiple places as well. 
like: 
https://developer.mozilla.org/en-US/docs/Web/API/console/log#syntax
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/onUpdated#syntax 

> The important part here is that the chose syntax should not be ambigious. Not using any special syntax would fit that requirement IMO. Just like using multiple signatures.

Agreed

gaearon:
I don't think we want to make this change.

However, I'll happily take a change that removes the `[` and `]` optional syntax from _all_ API references and instead writes the same signature multiple times. E.g. instead of

```
ReactDOM.render(element, container[, callback])
```

we could write

```
ReactDOM.render(element, container)
ReactDOM.render(element, container, callback)
```

content/docs/reference-react-dom-client.md
nitinajaydixit: Got it and thanks @harish-sethuraman for mentioning #3919 . I got confused with that [, options ] syntax.

nitinajaydixit:
> we could write
> 
> ```
> ReactDOM.render(element, container)
> ReactDOM.render(element, container, callback)
> ```

Yeah, overloading signature is less confusing.

PR #4543, opened 2022-04-04
docs: update branding in docs

asciidiego:
Super simple change.

### DIFF
--- a/content/docs/composition-vs-inheritance.md
+++ b/content/docs/composition-vs-inheritance.md
@@ -165,7 +165,7 @@ class SignUpDialog extends React.Component {
 
 ## So What About Inheritance? {#so-what-about-inheritance}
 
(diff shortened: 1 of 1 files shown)


diegonyc:
@facebook-github-bot CLA signed.

gaearon:
Seems like we'd want to do this consistently across other pages as well

diegonyc:
> Seems like we'd want to do this consistently across other pages as well

Any way to find what pages are there? Some of them have a link with a "Facebook" label but those are links to, say, the React Facebook page. In those cases, I am assuming you would not want to change those.

Is there a way to find places where it makes sense to change it?

gaearon:
Probably stuff in the `content` folder.

BartoszKlonowski:
Dears, I went through the latest of react.dev searching for any mentions of "Facebook" instead of "Meta" and couldn't find any that wouldn't be valid URLs or mentions at that time.
So I think that we can safely close this PR as no longer required, but thanks @asciidiego for bringing this!

PR #4502, opened 2022-03-27
Fix grammar

jeongwoopark0514:
Closes #4501.
Fix small grammar issues in the description.

PR #4545, opened 2022-04-04
Small grammar issue solved

AakashTyagi354:
Small grammatical update to the reactjs.org docs.
The changes that where made

1. Instead of " The only difference " it was changed to " The only difference ~~ is "
2.  Instead of " many of the same component " it was changed to " many of the same components "

issue number : #4502

### DIFF
--- a/beta/src/pages/learn/rendering-lists.md
+++ b/beta/src/pages/learn/rendering-lists.md
@@ -30,7 +30,7 @@ Say that you have a list of content.
 </ul>
 ```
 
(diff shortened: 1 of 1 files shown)


AakashTyagi354:
I have successfully signed the license

gaearon:
The first change seems like a misunderstanding. (~~ doesn't mean anything, I think you got confused by the comment from the person who filed the issue).

The second change is also confusing ("many of the same components" is ambiguous) so I merged https://github.com/reactjs/reactjs.org/pull/4502 with some rewording.

PR #4548, opened 2022-04-04
Fix version in babel url

DavidKessel:
The babel compiler produced an error with this url. I updated the version to fix it.


gaearon:
Something's wrong with the commits?

DavidKessel:
sorry - just saw. Let me try again.

PR #4549, opened 2022-04-04
fix babel url version

DavidKessel:
The babel compiler produced an error with this url. I updated the version to fix it.

### DIFF
--- a/beta/src/pages/learn/add-react-to-a-website.md
+++ b/beta/src/pages/learn/add-react-to-a-website.md
@@ -136,7 +136,7 @@ return <button onClick={() => setLiked(true)}>Like</button>;
 
 These two code snippets are equivalent. JSX is popular syntax for describing markup in JavaScript. Many people find it familiar and helpful for writing UI code--both with React and with other libraries. You might see "markup sprinkled throughout your JavaScript" in other projects!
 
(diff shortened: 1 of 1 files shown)


beta/src/pages/learn/add-react-to-a-website.md
```diff
@@ -136,7 +136,7 @@ return <button onClick={() => setLiked(true)}>Like</button>;
...
 
-> You can play with transforming HTML markup into JSX using [this online converter](https://babeljs.io/en/repl#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=DwIwrgLhD2B2AEcDCAbAlgYwNYF4DeAFAJTw4B88EAFmgM4B0tAphAMoQCGETBe86WJgBMAXJQBOYJvAC-RGWQBQ8FfAAyaQYuAB6cFDhkgA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=es2015%2Creact%2Cstage-2&prettier=false&targets=&version=7.4.3).
+> You can play with transforming HTML markup into JSX using [this online converter](https://babeljs.io/en/repl#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=DwIwrgLhD2B2AEcDCAbAlgYwNYF4DeAFAJTw4B88EAFmgM4B0tAphAMoQCGETBe86WJgBMAXJQBOYJvAC-RGWQBQ8FfAAyaQYuAB6cFDhkgA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=es2015%2Creact%2Cstage-2&prettier=false&targets=&version=7.17).
```
harish-sethuraman, line 139: A quick doubt here? we could have rather used https://babeljs.io/en/repl right? since this generates the later part when we visit the page and would never break?

gaearon: How do you pre-fill it with a piece of JSX? That’s what this url is doing.

harish-sethuraman: 😮 Got it. Also, I noticed that when we don't give a version it takes the latest one

PR #4550, opened 2022-04-05
Update higher-order-components.md

princeibs:
Fixed typo in comment of code example.
Changed "any" -> "an"

### DIFF
--- a/content/docs/higher-order-components.md
+++ b/content/docs/higher-order-components.md
@@ -156,7 +156,7 @@ function withSubscription(WrappedComponent, selectData) {
 
     render() {
       // ... and renders the wrapped component with the fresh data!
-      // Notice that we pass through any additional props
+      // Notice that we pass through an additional props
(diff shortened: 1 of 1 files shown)


content/docs/higher-order-components.md
```diff
@@ -156,7 +156,7 @@ function withSubscription(WrappedComponent, selectData) {
 
     render() {
       // ... and renders the wrapped component with the fresh data!
-      // Notice that we pass through any additional props
+      // Notice that we pass through an additional props
```
eps1lon, line 159: there are mutliple props. it's like "we pass through any additional content" or "any additonal stuff". not a singular "props". Hope this makes sense

eps1lon (changes requested):

content/docs/higher-order-components.md
princeibs: It makes more sense now. 
Thanks for the reply @eps1lon

PR #4552, opened 2022-04-06
Fix params order for hydrateRoot reference

tonyghiani:
After double-checking the types for the createRoot and hydrateRoot methods, I found a discrepancy in how the arguments were named in those methods.
In my understanding, the hydrateRoot() method gets as the first argument the DOM container, and the second one would be the initial React element we want to hydrate.

I just switch the order of those two arguments to make it consistent with its definition in the createRoot method.

### DIFF
--- a/content/docs/reference-react-dom-client.md
+++ b/content/docs/reference-react-dom-client.md
@@ -71,7 +71,7 @@ root.unmount();
 ### `hydrateRoot()` {#hydrateroot}
 
 ```javascript
-hydrateRoot(element, container[, options])
+hydrateRoot(container, element[, options])
 ```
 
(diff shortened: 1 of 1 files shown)


eps1lon (approved):
Thanks!

tonyghiani:
> Thanks! Please sign the CLA so that we can merge this.

I signed it about 20 mins ago, I guess the check should re-run automatically to get the approval.

harish-sethuraman:
> I signed it about 20 mins ago, I guess the check should re-run automatically to get the approval.

We will get an update from bot after sometime. :)

harish-sethuraman (commented):
Thanks!

PR #4557, opened 2022-04-06
Document useId is not for keys

gaearon:
Self-explanatory.


rickhanlonii (approved):
Is it worth adding this note to the List and Keys page too? Maybe at the bottom of the section [here](https://reactjs.org/docs/lists-and-keys.html#keys)

gaearon:
I'm not sure. Don't want to distract people learning about the concept of keys (since presumably they haven't learned about `useId` yet anyway). Let's wait and see how common this is?

sophiebits:
AFAICT the only way it could even seem useful for this purpose is if you called it in a loop, so I don't think this is a huge risk.

gaearon:
It's unfortunate we don't currently flag `users.map(useId)` in the lint rule. https://github.com/facebook/react/pull/22044

But yeah I think this is just because of naming. ("Ah, an ID! I need those.") But we can probably remove this now or in a few weeks after content creators write blog posts etc.

PR #4558, opened 2022-04-06
Edits for ReactDOMServer API

gaearon:
- Move renderToString + renderToStaticMarkup to the bottom
- Clarify which methods are supported in which environments
- Add links to full options
- Tweak usage examples

### DIFF
--- a/content/docs/reference-react-dom-server.md
+++ b/content/docs/reference-react-dom-server.md
@@ -205,5 +205,3 @@ ReactDOMServer.renderToStaticMarkup(element)
 Similar to [`renderToString`](#rendertostring), except this doesn't create extra DOM attributes that React uses internally, such as `data-reactroot`. This is useful if you want to use React as a simple static page generator, as stripping away the extra attributes can save some bytes.
 
 If you plan to use React on the client to make the markup interactive, do not use this method. Instead, use [`renderToString`](#rendertostring) on the server and [`ReactDOM.hydrateRoot()`](/docs/react-dom-client.html#hydrateroot) on the client.
-
-* * *
--- a/content/docs/reference-react-dom-server.md
+++ b/content/docs/reference-react-dom-server.md
@@ -10,7 +10,7 @@ The `ReactDOMServer` object enables you to render components to static markup. T
 
 ```js
 // ES modules
-import ReactDOMServer from 'react-dom/server';
+import * as ReactDOMServer from 'react-dom/server';
 // CommonJS
 var ReactDOMServer = require('react-dom/server');
 ```
@@ -27,7 +27,7 @@ These methods are only available in the **environments with [Web Streams](https:
 
 - [`renderToReadableStream()`](#rendertoreadablestream)
 
-The following methods can be used in both the server and browser environments:
+The following methods can be used in both the server and Web API environments:
 
 - [`renderToString()`](#rendertostring)
 - [`renderToStaticMarkup()`](#rendertostaticmarkup)
--- a/content/docs/reference-react-dom-server.md
+++ b/content/docs/reference-react-dom-server.md
@@ -72,9 +72,9 @@ const stream = renderToPipeableStream(
       // res.setHeader('Content-type', 'text/html');
       // stream.pipe(res);
     },
-    onError(x) {
+    onError(err) {
       didError = true;
-      console.error(x);
+      console.error(err);
     },
   }
 );
@@ -101,6 +101,7 @@ If you call [`ReactDOM.hydrateRoot()`](/docs/react-dom-client.html#hydrateroot) 
 
 ```javascript
 let controller = new AbortController();
+let didError = false;
 try {
   let stream = await renderToReadableStream(
     <html>
@@ -108,6 +109,10 @@ try {
     </html>,
     {
       signal: controller.signal,
+      onError(e) {
+        didError = true;
+        console.error(err);
+      }
     }
   );
   
@@ -118,6 +123,7 @@ try {
   // await stream.allReady;
 
   return new Response(stream, {
+    status: didError ? 500 : 200,
     headers: {'Content-Type': 'text/html'},
   });
 } catch (error) {
--- a/content/docs/reference-react-dom-server.md
+++ b/content/docs/reference-react-dom-server.md
@@ -66,7 +66,7 @@ const stream = renderToPipeableStream(
     onAllReady() {
       // If you don't want streaming, use this instead of onShellReady.
       // This will fire after the entire page content is ready.
-      // You can use this for crawlers (SEO) or static generation.
+      // You can use this for crawlers or static generation.
 
       // res.statusCode = didError ? 500 : 200;
       // res.setHeader('Content-type', 'text/html');
@@ -118,7 +118,7 @@ try {
   
   // This is to wait for all Suspense boundaries to be ready. You can uncomment
   // this line if you want to buffer the entire HTML instead of streaming it.
-  // You can use this for crawlers (SEO) or static generation:
+  // You can use this for crawlers or static generation:
 
   // await stream.allReady;
 
--- a/content/docs/reference-react-dom-server.md
+++ b/content/docs/reference-react-dom-server.md
@@ -17,45 +17,22 @@ var ReactDOMServer = require('react-dom/server');
 
 ## Overview {#overview}
 
+These methods are only available in the **environments with [Node.js Streams](https://nodejs.dev/learn/nodejs-streams):**
+
+- [`renderToPipeableStream()`](#rendertopipeablestream)
+- [`renderToNodeStream()`](#rendertonodestream) (Deprecated)
+- [`renderToStaticNodeStream()`](#rendertostaticnodestream)
+
+These methods are only available in the **environments with [Web Streams](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API)** (this includes browsers, Deno, and some modern edge runtimes):
+
+- [`renderToReadableStream()`](#rendertoreadablestream)
+
 The following methods can be used in both the server and browser environments:
 
 - [`renderToString()`](#rendertostring)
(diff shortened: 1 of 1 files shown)


content/docs/reference-react-dom-server.md
```diff
@@ -17,93 +17,89 @@ var ReactDOMServer = require('react-dom/server');
...
+> This is a Node.js-specific API. Environments with [Web Streams](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API), like Deno and modern edge runtimes, should use [`renderToReadableStream`](#rendertoreadablestream) instead.
 >
 
-```
+```javascript
+let didError = false;
 const {pipe, abort} = renderToPipeableStream(
   <App />,
   {
-    onAllReady() {
```
gaearon, line 79: This seems to have suggested to always buffer. I think it's because it was taken from https://github.com/reactwg/react-18/discussions/127 which uses buffering as example.

content/docs/reference-react-dom-server.md
```diff
@@ -17,93 +17,89 @@ var ReactDOMServer = require('react-dom/server');
...
+## Reference {#reference}
 
 ### `renderToPipeableStream()` {#rendertopipeablestream}
 
 ```javascript
 ReactDOMServer.renderToPipeableStream(element, options)
 ```
 
-Render a React element to its initial HTML. Returns a [Control object](https://github.com/facebook/react/blob/3f8990898309c61c817fbf663f5221d9a00d0eaa/packages/react-dom/src/server/ReactDOMFizzServerNode.js#L49-L54) that allows you to pipe the output or abort the request. Fully supports Suspense and streaming of HTML with "delayed" content blocks "popping in" later through javascript execution. [Read more](https://github.com/reactwg/react-18/discussions/37)
+Render a React element to its initial HTML. Returns the `{ pipe, abort }` methods that allows you to pipe the output or abort the request. Fully supports Suspense and streaming of HTML with "delayed" content blocks "popping in" via inline `<script>` tags later. [Read more](https://github.com/reactwg/react-18/discussions/37)
```
sebmarkbage, line 43: It should really be treated as a "class" and not a destructured object.

content/docs/reference-react-dom-server.md
```diff
@@ -17,93 +17,89 @@ var ReactDOMServer = require('react-dom/server');
...
 > Note:
 >
-> This is a Node.js specific API and modern server environments should use renderToReadableStream instead.
+> This is a Node.js-specific API. Environments with [Web Streams](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API), like Deno and modern edge runtimes, should use [`renderToReadableStream`](#rendertoreadablestream) instead.
 >
 
-```
+```javascript
+let didError = false;
 const {pipe, abort} = renderToPipeableStream(
```
sebmarkbage, line 49: We should probably change this to be `const stream = ` and not destructure it. It's meant to mimic a node ReadableStream class instance.

gaearon: already

content/docs/reference-react-dom-server.md
```diff
@@ -115,8 +111,10 @@ try {
     }
```
sebmarkbage, line 111: This could use the `onError` part of the other example too.

content/docs/reference-react-dom-server.md
```diff
@@ -17,93 +17,89 @@ var ReactDOMServer = require('react-dom/server');
...
+      // The content above all Suspense boundaries is ready.
+      // If something errored before we started streaming, we set the error code appropriately.
+      res.statusCode = didError ? 500 : 200;
       res.setHeader('Content-type', 'text/html');
-      pipe(res);
+      stream.pipe(res);
     },
-    onShellError(x) {
+    onShellError(error) {
+      // Something errored before we could complete the shell so we emit an alternative shell.
```
gaearon, line 60: I find this comment counter-intuitive. If this is where errors "before the shell" go, then why does "on complete shell" check errors from "on error"?

sebmarkbage: It doesn't have to but technically you could have had an error inside a suspense boundary before the shell completed and you can use that to tag the status code. However, that's not guaranteed to be sufficient.

content/docs/reference-react-dom-server.md
gaearon: Oh. I thought try/catch is enough.

content/docs/reference-react-dom-server.md
sebmarkbage: I'm actually not sure it's ever best-practice to set a 500 status code. You'd think that maybe it would be good for SEO bots but it's actually worse. Because they stop reindexing so aggressively. So instead of trying again when it might be fixed, it tries less.

If there is already something in the index maybe it prevents it from being overridden but I'm not sure and if it's the first time it's not good.

Also if the error is small, maybe it's still better to index what you successfully did.

content/docs/reference-react-dom-server.md
```diff
@@ -17,93 +17,89 @@ var ReactDOMServer = require('react-dom/server');
...
       res.statusCode = 500;
       res.send(
         '<!doctype html><p>Loading...</p><script src="clientrender.js"></script>'
       );
-    }
+    },
+    onAllReady() {
+      // If you don't want streaming, use this instead of onShellReady.
+      // This will fire after the entire page content is ready.
+      // You can use this for crawlers (SEO) or static generation.
```
sebmarkbage, line 69: I'm not sure we should mention SEO here because these days, it's not actually mainly for SEO bots since they support JS. It's actually mainly for other types of dumb bots like social network's OpenGraph parsers etc.

content/docs/reference-react-dom-server.md
sebmarkbage: The try/catch is for onShellError only. onError can fire multiple times (which is why it's an event and not a promise) and still recover by client rendering.

content/docs/reference-react-dom-server.md
gaearon: We should probably show what people to do today since we're just trying to explain the API

content/docs/reference-react-dom-server.md
sebmarkbage: `onError` is kind of like `onRecoverableError` on the client. Maybe we should've just called it that.

content/docs/reference-react-dom-server.md
sebmarkbage: Yea, people seem to come to the conclusion that streaming is impossible/bad by following that logic though.

gaearon: ok just tell me what to put here :)

gaearon: or we could just ship this for the other changes since this part was already there

sebmarkbage: Yea we can follow up

sebmarkbage (approved):

content/docs/reference-react-dom-server.md
```diff
@@ -17,109 +17,113 @@ var ReactDOMServer = require('react-dom/server');
...
-### `renderToStaticMarkup()` {#rendertostaticmarkup}
+These methods are only available in the **environments with [Web Streams](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API)** (this includes browsers, Deno, and some modern edge runtimes):
 
-```javascript
-ReactDOMServer.renderToStaticMarkup(element)
-```
+- [`renderToReadableStream()`](#rendertoreadablestream)
 
-Similar to [`renderToString`](#rendertostring), except this doesn't create extra DOM attributes that React uses internally, such as `data-reactroot`. This is useful if you want to use React as a simple static page generator, as stripping away the extra attributes can save some bytes.
+The following methods can be used in both the server and browser environments:
```
sebmarkbage, line 30: Using the term “browser” environment is misleading because we *never* recommend using these in the browser except maybe in a Service Worker for dev.

They’re only for Web-like environments that is not actually the Browser.

content/docs/reference-react-dom-server.md
```diff
@@ -17,109 +17,113 @@ var ReactDOMServer = require('react-dom/server');
```
sebmarkbage, line 17: The ESM import is incorrect here. It should be `import * as `.

We shouldn't recommend `import *` neither since it breaks DCE but I guess we do that every where in old docs. It matters a bit for the server since we pull in multiple copies of the entire bundles through this one entry point.

content/docs/reference-react-dom-server.md
```diff
@@ -10,116 +10,120 @@ The `ReactDOMServer` object enables you to render components to static markup. T
...
-### `renderToStaticMarkup()` {#rendertostaticmarkup}
+These methods are only available in the **environments with [Web Streams](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API)** (this includes browsers, Deno, and some modern edge runtimes):
 
-```javascript
-ReactDOMServer.renderToStaticMarkup(element)
-```
+- [`renderToReadableStream()`](#rendertoreadablestream)
 
-Similar to [`renderToString`](#rendertostring), except this doesn't create extra DOM attributes that React uses internally, such as `data-reactroot`. This is useful if you want to use React as a simple static page generator, as stripping away the extra attributes can save some bytes.
+The following methods can be used in both the server and Web API environments:
```
sebmarkbage, line 30: I guess for these two it's actually for any *other* JS environment that doesn't support any of the above. Their existence is basically just because they don't have any complicated dependencies.

gaearon: https://github.com/reactjs/reactjs.org/blob/main/content/docs/reference-react-dom-server.md

gaearon: https://github.com/reactjs/reactjs.org/commit/1015a73e1abe829e61612474485549e034596df8

PR #4559, opened 2022-04-07
Add more details on batching

gaearon:
- It wasn't clear how to use flushSync.
- We should link from setState's to flushSync.
- Add some minimal info about batching to reference.

### DIFF
--- a/content/docs/reference-react-dom.md
+++ b/content/docs/reference-react-dom.md
@@ -62,7 +62,19 @@ Creates a portal. Portals provide a way to [render children into a DOM node that
 flushSync(callback)
 ```
 
-Force React to flush any updates inside the provided callback synchronously. This method is useful for being able to read the result of those updates immediately.
+Force React to flush any updates inside the provided callback synchronously. This ensures that the DOM is updated immediately so that you can do something with the result.
+
+```javascript
+function handleCarouselNextClick() {
+  // Force this state update to be synchronous.
+  flushSync(() => {
+    setCarouselIndex(carouselIndex + 1);
+  });
+
+  // By this point, DOM has been updated.
+  carouselRef.current.lastChild.scrollIntoView();
+}
+```
 
 > Note:
 > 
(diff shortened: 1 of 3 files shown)


rickhanlonii (approved):

content/docs/reference-react-dom.md
```diff
@@ -62,7 +62,19 @@ Creates a portal. Portals provide a way to [render children into a DOM node that
...
+```javascript
+function handleCarouselNextClick() {
+  // Force this state update to be synchronous.
+  flushSync(() => {
+    setCarouselIndex(carouselIndex + 1);
+  });
+
+  // By this point, DOM has been updated.
+  carouselRef.current.lastChild.scrollIntoView();
+}
```
sebmarkbage, line 76: This is an anti-pattern example, no?

It's tricky because some use cases are just bad but it's the easiest to get things done.

Others are rare but more legit.

gaearon: What's more canonical? Reading some layout?

Realistically this is what I use in some new docs too. I don't know of an easier way to do this without either effects or setState callback.

sebmarkbage: I think effects is probably correct. Especially if you imagine this might need an animation and then flushSync doesn't work.

sebmarkbage: My canonical go-to is a rAF loop although you probably shouldn't use React for that. If we could maybe we would just make that a discrete event anyway. Some things like dispatchEvent are arguably bugs in React. So those are not great examples neither.

The most canonical would probably be interop with a third party library that assumes the DOM to be updated when something is invoked on it.

gaearon: >interop with a third party library that assumes the DOM to be updated when something is invoked on it.

Do you have an example (even made up)? I'm struggling to follow the scenario

gaearon:
Let's use a generic example and follow-up if we have a better one.

PR #4575, opened 2022-04-12
Make note about React.PropTypes explicitly address its users (fixes #3632)

### LINKED ISSUE #3632: Note on PropTypes docs that directs users to prop-types can be confusing for new users
I was looking for the docs on how to use `defaultProps` and found [the PropTypes docs](https://reactjs.org/docs/typechecking-with-proptypes.html). The first thing I saw there was the yellow box at the top that says:

> Note:
> React.PropTypes has moved into a different package since React v15.5. Please use [the `prop-types` library instead](https://www.npmjs.com/package/prop-types).

I read this as meaning "this page is out of date, please see those docs", especially because the yellow box looks so important, and went to that package, which has much less comprehensive documentation. (Including no mention of `defaultProps` currently, although that's [going to be fixed](https://github.com/facebook/prop-types/issues/343).) I didn't understand until re-reading it the next day that it was intended for people using PropTypes under its old name and I was supposed to carry on reading the page.

I suspect that I may not be the only learner prone to this kind of misreading, so may I please suggest that you move the note to a little further down the page and possibly tweak the wording? Thank you.

scottdotjs:
Minor tweak of wording so new users know the note isn't for them.

### DIFF
--- a/content/docs/typechecking-with-proptypes.md
+++ b/content/docs/typechecking-with-proptypes.md
@@ -6,9 +6,9 @@ redirect_from:
   - "docs/react-api.html#typechecking-with-proptypes"
 ---
 
-> Note:
+> Note for `React.PropTypes` users:
 >
-> `React.PropTypes` has moved into a different package since React v15.5. Please use [the `prop-types` library instead](https://www.npmjs.com/package/prop-types).
+> In React v15.5 it moved into a different package, [`prop-types`](https://www.npmjs.com/package/prop-types). Please use that instead.
 >
 >We provide [a codemod script](/blog/2017/04/07/react-v15.5.0.html#migrating-from-reactproptypes) to automate the conversion.
 


gaearon:
I think the confusion in #3632 is because the page _is not about defaultProps_. It's about PropTypes. So maybe this could be clarified somehow instead? The defaultProps documentation is in https://reactjs.org/docs/react-component.html#defaultprops. So maybe the defaultProps section of this page could link to it first.

scottdotjs:
Hi Dan, thanks for the observation. It was definitely a bit confusing for me at the time to end up on this page rather than the defaultProps one itself. I think it would make sense for the paragraph beginning "If you are using a Babel transform...", the following example, and paragraph to move to that page, and be replaced on this one with a "for more details" link.

This particular PR is just to address the potential misreading of the note about React.PropTypes [like I had](https://github.com/reactjs/reactjs.org/issues/3632#issuecomment-820316807) at first glance.

gaearon:
I'm not sure I understand the confusion in the original comment.

>Please use the prop-types library instead. Oh, okay. I guess I have to follow this link. This must be legacy documentation.

This seems correct to me. A user _should_ use `prop-types` library instead and read the documentation there. PropTypes is essentially gone from React. The page only exists for historical users so people using PropTypes earlier could find it.

scottdotjs:
🤦🏻‍♂️ Oh my god. Okay. So my initial reading was actually correct. By the time I'd found what I was actually looking for (the part about defaultProps) I'd convinced myself otherwise. Total brain fart, sorry.

Happy to change this PR to just be the defaultProps stuff if that'll make up for it.

gaearon:
No prob, let's do that!

PR #4576, opened 2022-04-12
fix the code block for createRoot typescript error

folt:
Improving the documentation will improve the developer experience and provide a solution for issues:
[React 18 - typescript gives an error when reproducing official Upgrade Guide](https://github.com/facebook/react/issues/24208)
[[Bug]: createRoot type error - TS2345](https://github.com/facebook/react/issues/24196)

### DIFF
--- a/content/blog/2022-03-08-react-18-upgrade-guide.md
+++ b/content/blog/2022-03-08-react-18-upgrade-guide.md
@@ -39,8 +39,20 @@ render(<App tab="home" />, container);
 
 // After
 import { createRoot } from 'react-dom/client';
-const container = document.getElementById('app') as HTMLElement;
+const container = document.getElementById('app');
 const root = createRoot(container);
+root.render(<App tab="home" />);
+```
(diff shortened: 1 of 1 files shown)


gaearon:
This is not a typescript example. We can add a separate typescript example below but we shouldn’t have invalid JS code as the only example.

gaearon (changes requested):
The blog post is written with JS in mind. This is not valid JS. Let’s think of another way.

folt:
> The blog post is written with JS in mind. This is not valid JS. Let’s think of another way.

I propose to add a block with an example for typescript. What do you think?

gaearon:
This still seems confusing to me. It breaks the flow of the page (all other examples have Before/After, but not this one), and it's inconsistent with the previous one (different ID).

gaearon:
I fixed it slightly differently. This doesn't fix all examples but should be enough for people to figure out how to do this.

PR #4579, opened 2022-04-13
Update code-splitting.md

Shivampurbia:


gaearon:
thanks, but this is not necessary.

Issue #4581, opened 2022-04-15
Confusing Variable Naming: "step" & "move"


n795113:

n795113:
I am not sure how to edit the code on Codepen together so please let me know how if the above commit makes sense.

tahseenio:
Hi I can take a look at this

PR #4582, opened 2022-04-15
Change naming of variable in tutorial.md

n795113:
Makes variable naming consistent ( #4581 )


gaearon:
We should just fix it in the new WIP tutorial that's being ported in https://github.com/reactjs/reactjs.org/pull/4320. Let's not touch the existing one because updating all CodePens is also a pain.

n795113:
great! I really enjoyed this tutorial.
Looking forward to the HOOK version.

PR #4583, opened 2022-04-16
Update forms.md

AndryBethpalko:
When select is multiple it does not return array in its value property. To select all options you should check them manually.
Possible solution is:
handleChange(event) {
        this.setState(
            {
                value: Array.from(event.target.children).filter(
                    children => children.selected
                ).map(
                    children => children.value
                )
            }
        );
    }


BartoszKlonowski:
Hello @AndryBethpalko!
As the examples in the latest of react.dev, included [here](https://react.dev/reference/react-dom/components/select#reading-the-select-box-value-when-submitting-a-form) contain the advise on how to read and format the received values I will close this PR as outdated.
Thank you!

PR #4584, opened 2022-04-16
Update lifting-state-up.md

AndryBethpalko:


gaearon:
thanks, but this is not necessary — we can't explain the same syntax on every page

PR #4585, opened 2022-04-17
Fix "Integrating with Other View Libraries"

smikitky:
This seems to be a simple mistake made during the docs update for React 18.

What you need to call to make multiple roots is `createRoot()`, not `root.render()`.

### DIFF
--- a/content/docs/integrating-with-other-libraries.md
+++ b/content/docs/integrating-with-other-libraries.md
@@ -192,7 +192,7 @@ class Chosen extends React.Component {
 
 React can be embedded into other applications thanks to the flexibility of [`createRoot()`](/docs/react-dom-client.html#createRoot).
 
-Although React is commonly used at startup to load a single root React component into the DOM, `root.render()` can also be called multiple times for independent parts of the UI which can be as small as a button, or as large as an app.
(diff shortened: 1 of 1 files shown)


content/docs/integrating-with-other-libraries.md
```diff
@@ -192,7 +192,7 @@ class Chosen extends React.Component {
 
 React can be embedded into other applications thanks to the flexibility of [`createRoot()`](/docs/react-dom-client.html#createRoot).
 
-Although React is commonly used at startup to load a single root React component into the DOM, `root.render()` can also be called multiple times for independent parts of the UI which can be as small as a button, or as large as an app.
+Although React is commonly used at startup to load a single root React component into the DOM, `createRoot()` can also be called multiple times for independent parts of the UI which can be as small as a button, or as large as an app.
```
harish-sethuraman, line 195: We createRoot first and then use the root.render() to render it right? 
```js
const element = <h1>Hello, world</h1>;
const root = ReactDOM.createRoot(
  document.getElementById('root')
);
root.render(element);
```

smikitky: The original sentence looks to me like one `root` object can be shared across multiple root DOMs. The message here is that we can have multiple `root`s by calling `createRoot()` multiple times.

gaearon:
thanks

Issue #4572, opened 2022-04-11
Proposal to change state hook documentation


matowang:
The[ current documentation for setting state hooks](https://reactjs.org/docs/hooks-intro.html) is shown as:
```js
<button onClick={() => setCount(count + 1)}>
```
However wouldn't it be much better to use the previous state?
```js
<button onClick={() => setCount(prevCount => prevCount+ 1)}>
```
This would be more consistent with the [original setState documentation](https://reactjs.org/docs/state-and-lifecycle.html#state-updates-are-merged).

I have also personally had issues with states being handled incorrectly handled due to using the documented method of setting state with the rendered state.

Wouldn't this prevent a lot of logical errors people may face due to the documentation?

PR #4588, opened 2022-04-18
docs: update setState hook in docs

jeffascript:
Updated the doc to reflect the change for prevState callback on the [useState hook](https://reactjs.org/docs/hooks-state.html)  based on #4572

Screen Shot 2022-04-18 at 18 28 07

### DIFF
--- a/content/docs/hooks-state.md
+++ b/content/docs/hooks-state.md
@@ -187,6 +187,14 @@ In a function, we already have `setCount` and `count` as variables so we don't n
   </button>
 ```
 
+Alternatively, we can use the second form of setState() that accepts a function. That function will receive the previous state as the first argument `prevState`, which will always contain the modified state in the recent `setCount` call:
+
+```js{1}
+  <button onClick={() => setCount((prevState) => prevState + 1)}>
+    Click me
+  </button>
+```
+
 ## Recap {#recap}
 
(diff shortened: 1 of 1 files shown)


jeffascript:
@facebook-github-bot CLA signed.

content/docs/hooks-state.md
```diff
@@ -187,6 +187,14 @@ In a function, we already have `setCount` and `count` as variables so we don't n
   </button>
 ```
 
+Alternatively, we can use the second form of setState() that accepts a function. That function will receive the previous state as the first argument `prevState`, which will always contain the modified state in the recent `setCount` call:
```
harish-sethuraman, line 190: I believe we have something similar here https://reactjs.org/docs/hooks-reference.html#functional-updates

jeffascript: Good catch @harish-sethuraman. What about adding the link you provided as a hyperlink in the PR for reference purposes? Would that suffice?

jeffascript:
@facebook-github-bot CLA signed

content/docs/hooks-state.md
harish-sethuraman: The page flow of the one you have edited seems to explain about states in class components vs state in functional components. we can better leave it as such in the hooks reference where it already is.

PR #4600, opened 2022-04-26
Fix behavior of renderToStaticMarkup

smikitky:
If I understand correctly, output from `renderToStaticMarkup` cannot be hydrated, so it's impossible to retry the rendering of its contents on the client.


gaearon:
makes sense

PR #4602, opened 2022-04-26
Corrected link in strict-mode.md

r3m0t:


eps1lon:
While the new link also describes the same contents, the original link directly talks about `display: contents` not just display-box in general.

PR #4605, opened 2022-04-27
React Tutorial: Disallow a gamer from editing current game while viewing historical moves

fullreboot:
Currently the final code version of the tutorial allows you to edit historical games moves. Added a small section and applicable code which would stop users from being able to edit a game while viewing history.

### DIFF
--- a/content/tutorial/tutorial.md
+++ b/content/tutorial/tutorial.md
@@ -1177,7 +1177,7 @@ We will also replace reading `this.state.history` with `this.state.history.slice
 
 Finally, we will modify the Game component's `render` method from always rendering the last move to rendering the currently selected move according to `stepNumber`:
 
-```javascript{3}
+```javascript
   render() {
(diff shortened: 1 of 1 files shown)


smikitky:
The current behavior is intentional. You must have overlooked the following paragraph in the tutorial:

> We will also replace reading `this.state.history` with `this.state.history.slice(0, this.state.stepNumber + 1)`. **This ensures that if we “go back in time” and then make a new move from that point, we throw away all the “future” history that would now be incorrect**.

Clearly, this game allows the user to overwrite the past history.

fullreboot:
@smikitky 

Ah, you're absolutely right. I was able to finish the tutorial. I didn't originally have that line of code in my version either. I must have missed the bottom part of the tutorial or something. Thanks for pointing that out.

It's a little strange to be intentional, but you're absolutely right that it was intentional based on that last paragraph.

PR #4610, opened 2022-04-28
Fix grammar

OnkarRuikar:
In sentence:
> Previously, we only encountered React elements that represent DOM tags

emphasis is on "React elements" and not the action "encountered". So, "only" needs to be placed before "React elements".

For examples,
```
- We only played.        // we did nothing else apart from playing
- We played only chess.  // we played no other games but chess
```


gaearon:
Thanks, but I don't think this is necessary or this distinction is important in this context

PR #4616, opened 2022-04-30
Update Docs

mike667:
The old version of documentation (about class-based components) doesn't seem to have updated the new changes.
Can you check if these fixes are correct?

### DIFF
--- a/content/docs/jsx-in-depth.md
+++ b/content/docs/jsx-in-depth.md
@@ -72,6 +72,10 @@ function WarningButton() {
 
 If you don't use a JavaScript bundler and loaded React from a `<script>` tag, it is already in scope as the `React` global.
 
+>**Note:**
+>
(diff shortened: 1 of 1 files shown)


content/docs/jsx-in-depth.md
```diff
@@ -408,6 +412,10 @@ Children passed to a custom component can be anything, as long as that component
 
 `false`, `null`, `undefined`, and `true` are valid children. They simply don't render. These JSX expressions will all render to the same thing:
 
+>**Note:**
+>
+>Since version 18, React rendering `undefined`.
```
gaearon, line 417: I don’t think this is relevant to this page.

gaearon:
The other change makes sense but it should instead be rewritten so that the new behavior is described by default. I.e. the docs always describe the latest behavior.

PR #4620, opened 2022-05-01
refactor(docs): use functional components

mendelgordon:
Class components require the user to use "this", and we can now use functional components with hooks instead.


mendelgordon:
I didn't see https://beta.reactjs.org/ until after making this pull request 😆

PR #4622, opened 2022-05-02
Fixing broken code example in tutorial

tomchapin:
Curly brackets were missing - example code was non-functional if copied and pasted


tomchapin:
I have signed the CLA, but it seems to be taking a while to update the CI checks for some reason

PR #4623, opened 2022-05-03
Expand renderToPipeableStream()’s documentation

RoyalIcing:
The documentation to `renderToPipeableStream()` is a sparse. For example, the current documentation links out the React source code for a full list of options. It also doesn’t explain that you can opt-out of the inline `<script>` approach by using the `onAllReady` callback.

This is a draft that I hope can be iterated on to clarify and expand the documentation. I won’t be offended if you don’t like my long headings or how I’ve structure it. I hope it provides a good first step to explain the core behaviour.

See: https://twitter.com/dan_abramov/status/1521297849593614339

Note that some terms seem interchangeable. I’ve tried using “ready” as that what the callbacks are named. It perhaps would be helpful to settle on canonical terms for:

- Suspended content blocks vs suspended content boundaries
- Ready vs Resolved vs Loaded vs Unsuspended
- Initial HTML vs Initial HTML shell

### DIFF
--- a/content/docs/reference-react-dom-server.md
+++ b/content/docs/reference-react-dom-server.md
@@ -42,13 +42,12 @@ ReactDOMServer.renderToPipeableStream(element, options)
 
 Render a React element to its initial HTML without interaction. Returns a [Node.js stream](https://nodejs.dev/learn/nodejs-streams) with a `pipe(res)` method to pipe the output and `abort()` to abort the request.
 
-This method fully supports Suspense, and offers two approaches for handling suspended content. The first lets you stream content blocks as they become ready. The second waits until **all** suspended content blocks are ready before sending a response.
+This method fully supports Suspense, allowing you to stream in suspended content blocks as they become ready.
 
-#### Suspense approach one: Initial shell with streamed content blocks
 
-The first allows you to respond with an initial HTML shell and then "pop in" content blocks when they become ready via inline `<script>` tags later. [Read more on how this works](https://github.com/reactwg/react-18/discussions/37)
+An initial HTML shell with your suspense fallbacks is first outputted, and then content blocks are "popped in" when ready via inline `<script>` tags later. [Read more on how this works](https://github.com/reactwg/react-18/discussions/37)
 
-To get this behaviour, implement the `onShellReady` callback and start streaming your response here. The `onShellError` option will be called if an error occurs before the shell was ready. The `onError` option will be called if an error occurs after the shell is ready.
+To get this behaviour, implement the `onShellReady` callback and start piping the stream into your response here. The `onShellError` option will be called if an error occurs before the shell was ready. The `onError` option will be called if an error occurs after the shell is ready.
 
 ```javascript
 let didError = false;
@@ -80,11 +79,11 @@ const stream = renderToPipeableStream(
 );
 ```
 
-#### Suspense approach two: Wait until all suspended blocks are ready
+#### Opting-out of inline scripts
 
-If you’d like to support clients that can’t run client-side JavaScript like crawlers, you can wait until all suspended blocks are ready and then stream the HTML in one go.
+If you’d like to support clients that can’t run client-side JavaScript such as crawlers, you can wait until **all** suspended blocks are ready and then stream the HTML in one go.
 
-To do so, implement the `onAllReady` callback and start streaming your response here. The `onError` option will be called if an error occurs during rendering.
+To do so, implement the `onAllReady` callback and start piping the stream into your response here. The `onError` option will be called if an error occurs during rendering, while may mean you HTML will only be partially outputted up to the point the error occurred.
 
 ```javascript
 let didError = false;
@@ -110,21 +109,21 @@ const stream = renderToPipeableStream(
 
 #### Hydrating to add interactivity {#rendertopipeablestream-hydrate}
 
-If you call [`ReactDOM.hydrateRoot()`](/docs/react-dom-client.html#hydrateroot) on a DOM node that already has this server-rendered markup, React will preserve it and only attach event handlers, allowing you to have a very performant first-load experience. This works with either approach to handling suspense.
+If you call [`ReactDOM.hydrateRoot()`](/docs/react-dom-client.html#hydrateroot) on a DOM node that already has this server-rendered markup, React will preserve it and only attach event handlers, allowing you to have a very performant first-load experience.
 
 #### Options
 
+- `onShellReady?: () => void` — use to stream a initial shell and then 
+- `onShellError?: () => void` — called if an error occurs while rendering the initial shell.
+- `onAllReady?: () => void` — use instead of `onShellReady` to wait for all suspense boundaries to be ready before streaming.
+- `onError?: (error: mixed) => void` — (called when?)
 - `identifierPrefix?: string`
-- `namespaceURI?: string`
+- `namespaceURI?: string` — supports `http://www.w3.org/2000/svg` for SVGs or `http://www.w3.org/1998/Math/MathML` for MathML
 - `nonce?: string`
 - `bootstrapScriptContent?: string`
 - `bootstrapScripts?: Array<string>`
 - `bootstrapModules?: Array<string>`
 - `progressiveChunkSize?: number`
-- `onShellReady?: () => void`
-- `onShellError?: () => void`
-- `onAllReady?: () => void`
-- `onError?: (error: mixed) => void`
 
 > Note:
 >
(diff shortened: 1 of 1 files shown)


content/docs/reference-react-dom-server.md
```diff
@@ -40,47 +40,91 @@ The following methods can be used in the environments that don't support streams
 ReactDOMServer.renderToPipeableStream(element, options)
 ```
 
-Render a React element to its initial HTML. Returns a stream with a `pipe(res)` method to pipe the output and `abort()` to abort the request. Fully supports Suspense and streaming of HTML with "delayed" content blocks "popping in" via inline `<script>` tags later. [Read more](https://github.com/reactwg/react-18/discussions/37)
+Render a React element to its initial HTML without interaction. Returns a [Node.js stream](https://nodejs.dev/learn/nodejs-streams) with a `pipe(res)` method to pipe the output and `abort()` to abort the request.
 
-If you call [`ReactDOM.hydrateRoot()`](/docs/react-dom-client.html#hydrateroot) on a node that already has this server-rendered markup, React will preserve it and only attach event handlers, allowing you to have a very performant first-load experience.
+This method fully supports Suspense, and offers two approaches for handling suspended content. The first lets you stream content blocks as they become ready. The second waits until **all** suspended content blocks are ready before sending a response.
+
+#### Suspense approach one: Initial shell with streamed content blocks
```
gaearon, line 47: Describing it as two approaches sounds like you can pick either one. In practice you should pick the first one for everything except static generation and maybe crawlers (although some crawlers can handle JS too). So I think we'd want to describe this based on the use case rather than either/or. Also, describing them separately makes it hard to explain how to switch the approach based on condition like `isCrawler`.

BurntCaramel:
BTW in a quick search for usage of this function I didn’t find many examples of people “doing it right” with regards to error handling so I feel that documentation might help here: https://github.com/search?q=renderToPipeableStream&type=code

content/docs/reference-react-dom-server.md
BurntCaramel: Yeah I agree. I’m trying to find a term of art for what this is. “When you can’t support client-side JavaScript”

content/docs/reference-react-dom-server.md
```diff
@@ -40,47 +40,91 @@ The following methods can be used in the environments that don't support streams
...
 ```
 
-See the [full list of options](https://github.com/facebook/react/blob/14c2be8dac2d5482fda8a0906a31d239df8551fc/packages/react-dom/src/server/ReactDOMFizzServerNode.js#L36-L46).
+#### Hydrating to add interactivity {#rendertopipeablestream-hydrate}
+
+If you call [`ReactDOM.hydrateRoot()`](/docs/react-dom-client.html#hydrateroot) on a DOM node that already has this server-rendered markup, React will preserve it and only attach event handlers, allowing you to have a very performant first-load experience. This works with either approach to handling suspense.
+
+#### Options
+
+- `identifierPrefix?: string`
```
BurntCaramel, line 117: I would be great to document what at least some of these are for, or to link to somewhere that shows how to implements them correctly. But I think at least putting them here is a good first step?

content/docs/reference-react-dom-server.md
```diff
@@ -40,47 +40,91 @@ The following methods can be used in the environments that don't support streams
...
       stream.pipe(res);
     },
-    onShellError(error) {
-      // Something errored before we could complete the shell so we emit an alternative shell.
-      res.statusCode = 500;
-      res.send(
-        '<!doctype html><p>Loading...</p><script src="clientrender.js"></script>'
-      );
+    onError(err) {
+      // TODO: when is this called?
```
BurntCaramel, line 74: I don’t quite understand when this is called — is it after the shell is ready? So if a suspended content blocks produce an error?

content/docs/reference-react-dom-server.md
```diff
@@ -40,47 +40,91 @@ The following methods can be used in the environments that don't support streams
...
+
+- `identifierPrefix?: string`
+- `namespaceURI?: string`
+- `nonce?: string`
+- `bootstrapScriptContent?: string`
+- `bootstrapScripts?: Array<string>`
+- `bootstrapModules?: Array<string>`
+- `progressiveChunkSize?: number`
+- `onShellReady?: () => void`
+- `onShellError?: () => void`
```
BurntCaramel, line 125: Should we mention somewhere that

An html doctype `<!DOCTYPE html>` is outputted if you render a root `<html>` element.

content/docs/reference-react-dom-server.md
```diff
@@ -40,47 +40,89 @@ The following methods can be used in the environments that don't support streams
...
+#### Hydrating to add interactivity {#rendertopipeablestream-hydrate}
+
+If you call [`ReactDOM.hydrateRoot()`](/docs/react-dom-client.html#hydrateroot) on a DOM node that already has this server-rendered markup, React will preserve it and only attach event handlers, allowing you to have a very performant first-load experience.
+
+#### Options
+
+- `onShellReady?: () => void` — called when an initial shell is ready to be streamed. Start piping into your response here and suspended content blocks will "pop in" via inline <script> tags.
+- `onShellError?: () => void` — called if an error occurs while rendering the initial shell.
+- `onAllReady?: () => void` — use instead of `onShellReady` to wait for all suspense boundaries to be ready before streaming. Start piping into your response here and the entire HTML tree will be outputted in one go.
+- `onError?: (error: mixed) => void` — (called when?)
```
BurntCaramel, line 118: I’m not sure when this is called and how it behaves different using `onShellReady` vs `onAllReady`

content/docs/reference-react-dom-server.md
```diff
@@ -40,47 +40,91 @@ The following methods can be used in the environments that don't support streams
...
+Render a React element to its initial HTML without interaction. Returns a [Node.js stream](https://nodejs.dev/learn/nodejs-streams) with a `pipe(res)` method to pipe the output and `abort()` to abort the request.
 
-If you call [`ReactDOM.hydrateRoot()`](/docs/react-dom-client.html#hydrateroot) on a node that already has this server-rendered markup, React will preserve it and only attach event handlers, allowing you to have a very performant first-load experience.
+Rendering a root `<html>` element will include a doctype (`<!DOCTYPE html>`) for you.
+
+This method fully supports Suspense, allowing you to stream in suspended content blocks as they become ready.
+
+An initial HTML shell with your suspense fallbacks is first outputted, and then content blocks are "popped in" when ready via inline `<script>` tags later. [Read more on how this works](https://github.com/reactwg/react-18/discussions/37)
+
+To get this behaviour, implement the `onShellReady` callback and start piping the stream into your response here. The `onShellError` option will be called if an error occurs before the shell was ready. The `onError` option will be called if an error occurs after the shell is ready.
```
BurntCaramel, line 51: Note that I’m saying “The `onShellError` option will be called if an error occurs before the shell was ready. The `onError` option will be called if an error occurs after the shell is ready.” — please verify if this is true?

BurntCaramel:
I have created a little project to test `renderToPipeableStream()` in different scenarios: https://github.com/BurntCaramel/renderToPipeableStreamPlayground

content/docs/reference-react-dom-server.md
BurntCaramel: @gaearon I’ve updated the documentation to follow more how you say, where `onShellReady` is the default and can be opted-into `onAllReady` based on a `isCrawler` condition. Not sure if it’s clear how I’ve done it, please let me know :)

content/docs/reference-react-dom-server.md
```diff
@@ -40,47 +40,98 @@ The following methods can be used in the environments that don't support streams
...
+An initial HTML shell with your suspense fallbacks is first outputted, and then content blocks are "popped in" when ready via inline `<script>` tags later. [Read more on how this works](https://github.com/reactwg/react-18/discussions/37)
+
+To get this behaviour, implement the `onShellReady` callback and start piping the stream into your response here.
 
 ```javascript
 let didError = false;
 const stream = renderToPipeableStream(
   <App />,
   {
+    onError(err) {
```
BurntCaramel, line 58: I put the error callbacks first to reflect the order that they would get called in.

BurntCaramel:
There might be some new stuff to document from the 18.2.0 release: https://github.com/facebook/react/releases/tag/v18.2.0

gaearon:
Addressed this via https://github.com/reactjs/reactjs.org/pull/5339. Thank you for starting this.

Issue #4569, opened 2022-04-10
update the codepen code "index.js" in tutorial for react18


awxiaoxian2020:
I'm a React beginner. when I use the given examples in [tutorial](https://reactjs.org/tutorial/tutorial.html) , it warns "ReactDOM.render is no longer supported in React 18. Use createRoot instead. ". I think we can update "index.js" to give a better experience for beginners.

==================================================================================
![image](https://user-images.githubusercontent.com/69554233/162738033-383e6e8b-8fa3-4663-b4cd-2bad03f21161.png)

==================================================================================

![image](https://user-images.githubusercontent.com/69554233/162738196-bce3bdd4-75a2-47f4-b8fa-c234d5247ef4.png)

==================================================================================

![react-issue](https://user-images.githubusercontent.com/69554233/162601324-7bf8687d-3e36-4998-b65a-913e29ef2a90.png)
.

harish-sethuraman:
Thanks for reporting. Can you please share where you found this codepen link? (the page's actual link) ?

awxiaoxian2020:
> Thanks for reporting. Can you please share where you found this codepen link? (the page's actual link) ?

[JS code here](https://codepen.io/gaearon/pen/oWWQNa?editors=0010) in [react tutorial](https://reactjs.org/tutorial/tutorial.html).

harish-sethuraman:
I'm not getting `ReactDOM.render is no longer supported in React 18. Use createRoot instead.` for the codepen you have provided? Can you share a screenshot please?

awxiaoxian2020:
> I'm not getting `ReactDOM.render is no longer supported in React 18. Use createRoot instead.` for the codepen you have provided? Can you share a screenshot please?

I use it in "Local Development Environment" , following the tutourial. It shows in the browser console. I'm busy now.  I'll give a screenshot later.

However, I think we shouldn't use the "React.render" as it said. it could not depend on where it repeats.

awxiaoxian2020:
Hi, you can see, when I use the js file, it warns in the browser console.

![react](https://user-images.githubusercontent.com/69554233/162733791-dcf5b43e-533e-4cdc-8d64-464f6765070c.png)

I have given the original source file in [my repo](https://github.com/awxiaoxian2020/react-issue). @harish-sethuraman

harish-sethuraman:
Did you use Create react app for local development? If so we might want to fix this cra’s repo for it to use newer syntax.

[...]

PR #4630, opened 2022-05-05
Tutorial markdown changes for React18

sathvikrijo:
Since ReactDOM.render has been deprecated from React-18, we have to have start using createRoot which is part of react-dom/client.

The [starter code](https://codepen.io/gaearon/pen/oWWQNa?editors=0010) having the ReactDOM.createRoot but we documented like we have to import ReactDOM from 'react-dom'. That's making confusion to the new react developers and also facing this warning in the console.

`Warning: You are importing createRoot from "react-dom" which is not supported. You should instead import it from "react-dom/client".`

Changes👇

Before:
Before

After:
After

### DIFF
--- a/content/tutorial/tutorial.md
+++ b/content/tutorial/tutorial.md
@@ -106,7 +106,7 @@ cd ..
 
 ```js
 import React from 'react';
-import ReactDOM from 'react-dom';
+import ReactDOM from 'react-dom/client';
 import './index.css';
 ```
 


content/tutorial/tutorial.md
```diff
@@ -106,9 +106,10 @@ cd ..
 
 ```js
 import React from 'react';
-import ReactDOM from 'react-dom';
+import ReactDOM from 'react-dom/client';
 import './index.css';
 ```
+Copy the JS and CSS from the **[Starter Code](https://codepen.io/gaearon/pen/oWWQNa?editors=0010)** and add it in your local in `index.js` and `index.css` respectively.
```
harish-sethuraman, line 112: I believe we already have this documented right? 
Screenshot 2022-05-05 at 8 50 40 PM

sathvikrijo: Missed to check the highlighted text. Removing those

harish-sethuraman (approved):

harish-sethuraman:
Thanks :)

awxiaoxian2020:
This closes #4569. Thank you very much indeed!

PR #4633, opened 2022-05-06
Rename confusing variable in useImperativeHandle example

dbismut:
Unless I've missed something, the example uses an internal ref called `inputRef` to bind the ref passed to the `FancyInput` component to the input dom node. The fact that the ref passed to the `<FancyInput />` component instance has the same name as the ref used internally in the component declaration is confusing.

### DIFF
--- a/content/docs/hooks-reference.md
+++ b/content/docs/hooks-reference.md
@@ -473,7 +473,7 @@ function FancyInput(props, ref) {
 FancyInput = forwardRef(FancyInput);
 ```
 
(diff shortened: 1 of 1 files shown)


BartoszKlonowski:
Hello @dbismut!
The reason to name it that way is that in that example this ref is eventually passed to be the `<input />`'s ref.
However, I checked whether this example is still in the react.dev and couldn't find it, so I will close this PR as it's outdated now.
Thank you!

PR #4640, opened 2022-05-07
Fixes layout shift

harish-sethuraman:
Fixes 
- https://github.com/reactjs/reactjs.org/issues/4638

Seems like the pitfall sandpack had output at bottom and the output screen didn't have a min height so it was reducing when the iframe inside it was killed. So added a default min height when it is inside expandable callouts

### DIFF
--- a/beta/src/components/MDX/ExpandableCallout.tsx
+++ b/beta/src/components/MDX/ExpandableCallout.tsx
@@ -41,6 +41,7 @@ function ExpandableCallout({children, type}: ExpandableCalloutProps) {
   return (
     <div
       className={cn(
+        'expandable-callout',
         'pt-8 pb-4 px-5 sm:px-8 my-8 relative rounded-none shadow-inner -mx-5 sm:mx-auto sm:rounded-lg',
         variant.containerClasses
       )}>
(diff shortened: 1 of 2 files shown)


gaearon:
What makes expandable callout special? Is it because it shifts the sandbox into mobile mode even on desktop?

harish-sethuraman:
Yes it basically takes tablet's width
Screenshot 2022-05-07 at 11 38 34 PM

gaearon:
Why don't we have the same problem on mobile for _all_ previews then? Or do we?

harish-sethuraman:
Why don't we have the same problem on mobile for all previews then?

We already set min height for all devices [1]. When the desktop view kicks in we unset the min-height [2]
1. https://github.com/reactjs/reactjs.org/blob/d1f2a31321d804fd4c5b4a48b0ace0507a124873/beta/src/styles/sandpack.css#L267-L274
2. https://github.com/reactjs/reactjs.org/blob/d1f2a31321d804fd4c5b4a48b0ace0507a124873/beta/src/styles/sandpack.css#L292-L297

So now when we have this expandable callout (pitfall component) we set the min height because the width cant accommodate the sandpack's output in the right

gaearon:
Heh. Okay.

PR #4648, opened 2022-05-09
Small correction

OnkarRuikar:
1. In the section:
> `this.props` contains the props that were defined by the caller of this component. See Components and Props for an introduction to props.
In particular, `this.props.children` is a special prop, typically defined by the child tags in the JSX expression rather than in the tag itself.

 The statement “`this.props.children` represents the child tags”  is more clear than “`this.props.children` is defined by the child tags”. Because in the statement above we are saying “The caller component defines the props contained in `this.props`”.

2. These two statements have different meaning: 
- React will still only update the DOM...
- React will still update the DOM only if the markup changes.

I think we mean the later statment.

### DIFF
--- a/content/docs/reference-react-component.md
+++ b/content/docs/reference-react-component.md
@@ -583,7 +583,7 @@ component.forceUpdate(callback)
 
 By default, when your component's state or props change, your component will re-render. If your `render()` method depends on some other data, you can tell React that the component needs re-rendering by calling `forceUpdate()`.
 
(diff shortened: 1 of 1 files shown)


content/docs/reference-react-component.md
```diff
@@ -635,7 +635,7 @@ The `displayName` string is used in debugging messages. Usually, you don't need
 
 `this.props` contains the props that were defined by the caller of this component. See [Components and Props](/docs/components-and-props.html) for an introduction to props.
 
-In particular, `this.props.children` is a special prop, typically defined by the child tags in the JSX expression rather than in the tag itself.
```
eps1lon, line 638: It's still important to include the "rather than the tag itself" because `<div children={<span>Hello, Dave!</span>} />` is also possible (typically when spreading props). We shouldn't omit that possibility.

eps1lon (changes requested):

PR #4650, opened 2022-05-09
Missing info about changes in StrictMode related to useEffect

### LINKED ISSUE #4649: Missing info about changes in StrictMode discoverable through useEffect docs
This ticket was originally reported in the React issue as a potential bug: https://github.com/facebook/react/issues/24502

After a short discussion with @gaearon I volunteered to update the docs so that once someone learns about empty dependencies array in `useEffect` simulating `componentDidMount` they won't be surprised to see effect callback run twice on mount.

cytrowski:
closes #4649

I found only 2 places where `useEffect` is mentioned as a replacement for `componentDidMount`. I feel this will make the new StrictMode behavior more discoverable for newcomers.


cytrowski:
I force-pushed this commit to trigger the CLA Check but apparently there is some propagation time and I just need to wait.

PR #4652, opened 2022-05-10

Olabisim:
i was not able to understand pure and impure functions from the doc.

the datatype of account should be added appropriately,
stating how it changes.


Olabisim:
by inline code i don't understand what you mean

HyunseungLee-Travis:
# from

account is an object that has a property of total, altering account.total means altering account 

# to

`account` is an object that has a property of total, altering `account.total` means altering account.

PR #4663, opened 2022-05-17
Relegate Reactathon 2022 to the annals of history

erikras:
Just moved it to the "past" section. 😄


harish-sethuraman (approved):

gaearon:
so shall it be

PR #4665, opened 2022-05-17
[Beta] Adds eslint integration on Sandpack

danilowoz:
This introduces the integration between eslint and CodeMirror lint on Sandpack.

**How to test**
1. Enable feature flag `localStorage['SANDPACK_ESLINT'] = true`
2. Access the page example `/eslint` to see some example errors;

**Todo**
- [x] What are the eslint rules we want to implement? All of them or only react-hooks?
- [x] Proper place and design for list of errors;
- [x] Lazily load the eslint modules, but how?;
- [x] Should it work on all sandboxes or only specific ones? We could implement something like `showDevTools`;
- [x] Remove feature flag

Screenshot 2022-05-17 at 21 25 47


beta/src/components/MDX/Sandpack/eslint-integration.tsx
```diff
@@ -0,0 +1,93 @@
...
+const options = {
+  parserOptions: {
+    ecmaVersion: 12,
+    sourceType: 'module',
+    ecmaFeatures: {jsx: true},
+  },
+  rules: {
+    'react-hooks/rules-of-hooks': 'error',
+    'react-hooks/exhaustive-deps': 'warn',
+    ...Array.from(rules).reduce((result, [ruleId, rule]) => {
```
danilowoz, line 41: Do we really want to lint all kinds of errors?

beta/src/components/MDX/Sandpack/utils.ts
```diff
@@ -95,3 +99,25 @@ export const createFileMap = (codeSnippets: any) => {
...
+  line: number;
+  column: number;
+  severity: 'warning' | 'error';
+  message: string;
+}[];
+
+export const useSandpackLint = () => {
+  const [diagnostic, setDiagnostic] = useState<LintDiagnostic>([]);
+
+  const onLint = linter((props: any) => {
```
danilowoz, line 113: Fix type annotation

gaearon:
Only the two official React Hooks rules. Both should fail the build (at least by default) so no need for multiple view. Just show the first lint error as a compilation error.

gaearon:
So basically, same behavior as having multiple build errors. We show the first one on the right but we can (in principle) highlight multiple on the left.

beta/src/components/MDX/Sandpack/eslint-integration.tsx
harish-sethuraman: I've removed this since since we wanted only the two react hooks rules

beta/src/components/MDX/Sandpack/utils.ts
harish-sethuraman: Done

harish-sethuraman:
Splitting the `eslint-integration` file alone had some good reduction. need to split `@codemirror/lint` as well!
Before: 
| Page |  Size (compressed) | First Load
| -- | -- | -- |
| / | 436.24 KB (🔴 +381.35 KB) | 531.48 KB |



After:
| Page |  Size (compressed) | First Load
| -- | -- | -- |
| / | 116.99 KB (🔴 +62.1 KB) | 212.29 KB |

gaearon (changes requested):
Amazing, and thanks @harish-sethuraman for first set of edits.
I'd like to:

- Show just the first error in the output box instead of all
- Add codeframe to lint errors if possible so they look more similar to build errors
- Fix styling of the popup on the light theme (currently it has light color). Also let's remove the red/orange color accent there. At the very least it should match in color to our colors.

harish-sethuraman:
> Add codeframe to lint errors if possible so they look more similar to build errors


I guess we have to construct the code frame with the existing code to show it for lint errors. The build errors are coming from sandpack with those code frames.

gaearon:
When there's both syntax and lint error, syntax error should be displayed alone. Also they need to have the same paddings.

Screenshot 2022-05-18 at 18 32 08

gaearon (approved):
great

harish-sethuraman (approved):
Thanks @danilowoz

PR #4672, opened 2022-05-22
added code for sandpack console

harish-sethuraman:
Adds console to sandpack component

UI :
- [x] Remove the "no action" cursor over the console bar, it's confusing IMO.
- [x] Change "console" to "Console". Align text to the left.
- [x] Add a bit more padding. Use the same horizontal padding as in the top bar. Basically, on mobile, try to make "Console" label align with "Show more" label, and make arrows the same size/position.
- [x] When the console is empty, expanding it should still work. It should just expand some area with a min height of one line. So that you know this is where console messages will appear.

Behaviour: 
- [x] Let's limit this to `console.log` only. No errors or warnings.
- [x] JSON.stringify call needs to be in try/catch. If it fails, try `Object.prototype.toString.call(value)`. If that throws too, just print `"[" + typeof(value) + "]"`.
- [x]  When you print objects like `console.log("hi", {}, {})` the vertical space between objects is too large. Put them closer. The excessive indentation is also strange, let's remove it.
- [x]  The sandbox loading indicator should not overlap the console. Position it so that it is above it (when the console is present)? (ignoring this one)
- [x]  When there's an error, console logs should not disappear or change their count. (using `message.firstLoad` from sandpack)
- [x] Sometimes putting a `console.log` into component body and then edit it causes to console to always clear before next render. I don't know why. Seems like a bug.
- [x]  we should probably remove the 2px border and just do a bottom border on all items except for :last.
- [x] use the monospace font


harish-sethuraman:
@danilowoz Seems like the new bundler doesn't emit `console` type messages. But the old one does emit so I've reverted that in this PR.

beta/src/components/MDX/Sandpack/SandpackRoot.tsx
```diff
@@ -64,7 +64,13 @@ ul {
 `.trim();
 
 function SandpackRoot(props: SandpackProps) {
-  let {children, setup, autorun = true, showDevTools = false} = props;
+  let {
+    children,
+    setup,
+    autorun = true,
+    showDevTools = false,
+    showConsole = true,
```
harish-sethuraman, line 72: We can disable this when we are done testing

danilowoz:
Hey @harish-sethuraman, I noted that after a few consoles, the list is overlapping the preview: 

Screenshot 2022-05-23 at 17 42 23

danilowoz:
Hey @harish-sethuraman, the following sandbox is breaking the whole page. I think there is something wrong with the render of each console message
Screenshot 2022-05-24 at 09 29 57

gaearon (changes requested):
It's a good start. 

- Let's limit this to `console.log` only. No errors or warnings.
- JSON.stringify call needs to be in try/catch. If it fails, try `Object.prototype.toString.call(value)`. If that throws too, just print `"[" + typeof(value) + "]"`.
- Remove the "no action" cursor over the console bar, it's confusing IMO.
- Change "console" to "Console". Align text to the left.
- Add a bit more padding. Use the same horizontal padding as in the top bar. Basically, on mobile, try to make "Console" label align with "Show more" label, and make arrows the same size/position.
- When you print objects like `console.log("hi", {}, {})` the vertical space between objects is too large. Put them closer. The excessive indentation is also strange, let's remove it.
- When the console is empty, expanding it should still work. It should just expand some area with a min height of one line. So that you know this is where console messages will appear.
- The sandbox loading indicator should not overlap the console. Position it so that it is above it (when the console is present)?
- When there's an error, console logs should not disappear or change their count.
- Sometimes putting a `console.log` into component body and then edit it causes to console to always clear before next render. I don't know why. Seems like a bug.

harish-sethuraman:
> When there's an error, console logs should not disappear or change their count.

When there is a sandpack error there were two consoles emmited. One is `Error in the sandbox` message which is of type `log` and another `error`. How can we go about blocking this `Error in the sandbox` message?

harish-sethuraman:
> Hey @harish-sethuraman, the following sandbox is breaking the whole page. I think there is something wrong with the render of each console message

Seems to be because the event's object cannot be mapped? Ill have a look into it :)

gaearon:
The error with event object is likely that it can’t be JSON.stringify’ed because it’s circular. This is why I suggested a try/catch.

gaearon:
@harish-sethuraman Do you expect to be able to finish it this week?

harish-sethuraman:
Will try to wrap up kinda caught up with other works.

harish-sethuraman:
@gaearon  Can you check now? Also I'm wondering if this works in local for you? I'm getting CORS error

harish-sethuraman:
- When you print objects like console.log("hi", {}, {}) the vertical space between objects is too large. Put them closer. The excessive indentation is also strange, let's remove it.

This is because the message `{}` is not considered to be string and it is rendered inside `SandpackCodeViewer`. We can reduce the vertical space but the same will apply for other objects too? 

How it looks 
Screenshot 2022-05-31 at 10 48 07 PM

How it will look
Screenshot 2022-05-31 at 10 48 30 PM

-  The sandbox loading indicator should not overlap the console. Position it so that it is above it (when the console is present)?

This one is not reproducible for me? The console appears only after the sandbox loader has gone away? 

- When there's an error, console logs should not disappear or change their count. Sometimes putting a console.log into component body and then edit it causes to console to always clear before next render. I don't know why. Seems like a bug.

This is because we are listening to the `start` message from Sandpack. I use start because when we move away from sandpack it is killed (I dont see any event specific to sandpack getting killed) then when we scroll up again the consoles appear again. @danilowoz  which event is recommended for clearing console?

gaearon:
>How it will look

seems fine to me

>This one is not reproducible for me? The console appears only after the sandbox loader has gone away?

hmm i don't remember how i got this. ok let's ignore for now

gaearon:
@lebo any thoughts on the console design? can we keep it like this or need to redo it? add `console.log('foo')` to any sandbox on https://beta-reactjs-b2bt5iob0-fbopensource.vercel.app/learn to try it. we'd hide it until first message by default so it'll only show up when you use console

lebo:
Looks cool! UX seems right (having it as a collapsible section that appears once there's something logged). 

Design wise, I think it could match the code section a bit more. Any opposition to using the monospace font and aligning the margins, line height, and spacing? Also we should probably remove the 2px border and just do a bottom border on all items except for :last.

gaearon:
We need to keep the normal font for the "Console" header itself. Monospace is just for the content.

gaearon:
And let's make the console hidden until the first `console.log` statement runs.

beta/src/components/MDX/Sandpack/CustomPreset.tsx
```diff
@@ -70,11 +73,14 @@ export function CustomPreset({
               extensions={lintExtensions}
               extensionsKeymap={emptyArray}
             />
-            <Preview
-              className="order-last xl:order-2"
-              isExpanded={isExpanded}
-              lintErrors={lintErrors}
-            />
+            <div className="sp-stack h-full">
```
gaearon, line 76: This breaks the narrow view.

Screenshot 2022-06-10 at 20 00 32

PR #4680, opened 2022-05-23
[Beta] Update Next and React

gaearon:
Bumps React and Next.
Adds some fixes for issues I noticed.

### DIFF
--- a/beta/next.config.js
+++ b/beta/next.config.js
@@ -17,14 +17,16 @@ module.exports = {
   async redirects() {
     return redirects.redirects;
   },
-  rewrites() {
-    return [
-      {
-        source: '/feed.xml',
-        destination: '/_next/static/feed.xml',
-      },
-    ];
-  },
+  // TODO: this causes extra router.replace() on every page.
+  // Let's disable until we figure out what's going on.
+  // rewrites() {
+  //   return [
+  //     {
+  //       source: '/feed.xml',
+  //       destination: '/_next/static/feed.xml',
+  //     },
+  //   ];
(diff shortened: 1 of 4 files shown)


beta/next.config.js
```diff
@@ -17,14 +17,16 @@ module.exports = {
...
   },
-  rewrites() {
-    return [
-      {
-        source: '/feed.xml',
-        destination: '/_next/static/feed.xml',
-      },
-    ];
-  },
+  // TODO: this causes extra router.replace() on every page.
```
gaearon, line 20: See here https://github.com/vercel/next.js/pull/33861#issuecomment-1135071203.

Originally I thought the `startTransition` call worked around it, but I was still seeing the issue. I ended up just ripping out this feature for now to avoid the false positive router navigations.

beta/src/components/MDX/Sandpack/DownloadButton.tsx
```diff
@@ -7,23 +7,33 @@ import {useSandpack} from '@codesandbox/sandpack-react';
...
+        (HTMLScriptElement as any).supports &&
+        (HTMLScriptElement as any).supports('importmap');
     }
-  }, [supported]);
+    return supportsImportMap;
+  }
+  function getServerSnapshot() {
+    return false;
+  }
+  return React.useSyncExternalStore(
```
gaearon, line 27: useSES seems like the way to go here since the value may change between sandboxes mounting. So it's not legit to cache it and then read the value that could be inconsistent from the server into useState.

PR #4685, opened 2022-05-25
Clarify conditional rendering using &&

vwkd:
Tried to make it clearer, that for any falsy value other than`false`, conditional rendering using `&&` might not produce what you expect and how to fix that.

Also added the name of the operator, a link to MDN, and removed a redundant mention of code in JSX, since it is already extensively used since chapter 2 „Introducing JSX“. Also mentioned the same caveat applies to the ternary operator.

### DIFF
--- a/content/docs/conditional-rendering.md
+++ b/content/docs/conditional-rendering.md
@@ -164,6 +164,8 @@ render() {
 }
 ```
 
+In these cases a boolean expression like `count !== undefined` is needed.
+
 ### Inline If-Else with Conditional Operator {#inline-if-else-with-conditional-operator}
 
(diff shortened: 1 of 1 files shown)


content/docs/conditional-rendering.md
```diff
@@ -143,11 +143,15 @@ root.render(<Mailbox unreadMessages={messages} />);
 
 [**Try it on CodePen**](https://codepen.io/gaearon/pen/ozJddz?editors=0010)
 
-It works because in JavaScript, `true && expression` always evaluates to `expression`, and `false && expression` always evaluates to `false`.
+The `&&` operator returns the left operand if it is falsy, otherwise if the left operand is truthy it returns the right operand.
 
-Therefore, if the condition is `true`, the element right after `&&` will appear in the output. If it is `false`, React will ignore and skip it.
+React will render any value that is not `false`.
```
vwkd, line 148: Need to fix this to say exactly for which which falsy values it doesn’t render, like `null` and `undefined`.

PR #4688, opened 2022-05-25
[Beta] Slightly reduce bundle size

gaearon:
- Don't pull in stuff from the Algolia package on the initial load
- Don't bundle the useSyncExternalStore shim https://github.com/vercel/next.js/issues/37197

### DIFF
--- a/beta/next.config.js
+++ b/beta/next.config.js
@@ -40,6 +40,9 @@ module.exports = {
       );
     }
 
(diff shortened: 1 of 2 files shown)


gaearon:
Hmm maybe the Algolia stuff was getting tree-shaken well? Still kinda risky so good to remove the import.

Issue #4691, opened 2022-05-26
[Beta] Reflect pending navigation in the sidebar


gaearon:
Turn on "slow 3g" in Network emulation. Notice that navigation feels slow in part because there's no visual feedback.

https://user-images.githubusercontent.com/810438/170408691-8486215f-ebfc-4bfe-b2a9-124e92b41be2.mov

I think what should happen instead is that the active sidebar component should reflect the currently pending route. Even if it's still loading. Alternatively, maybe there should be a separate "pending" state that feels in-between hover and selected.

I don't know what's the best way to do this with the current Next.js conventions. Maybe the pending route is available somehow. Keep in mind that it's not the same as "last clicked" because you could also press Back or click a link afterwards.

michaelaflores:
One low-hanging fruit thing I've done before in a NextJS site was using `nprogress` to provide some immediate visual feedback about the pending route change. The UX you describe is definitely better though, and more effort.

mmocny:
Hi!  I work on Interaction to Next Paint (INP) and I'm especially keen to learn how a great modern React site can do well on this metric.  I very much appreciate you sharing your [perf improvements thread](https://twitter.com/dan_abramov/status/1529677207869825024) publically, and I'd like to help here however needed.

I know you folks are insightful enough to understand the metric, but for any bystanders and passers by... I hope you don't mind if I share a small overview and a few general INP principles since this is a new metric we're all learning to optimize collectively?

---

* [Interaction to Next Paint (INP) blog post](https://web.dev/inp) is a good start to learn about the definition and goals
* You may prefer to watch a [recent 3 minute lightning video](https://youtu.be/o1pJJuQiCmQ?t=3291) from BlinkOn (or the [longer I/O video](https://www.youtube.com/watch?v=Mizzbsvv8Os))
* If you have a deployed site, you can quickly check your INP field metrics thanks to [Chrome User Experience (CrUX) report](https://developers.google.com/web/tools/chrome-user-experience-report)
  * The easiest way to get a simple score quickly is probably to use the [pagespeed insights tool](https://pagespeed.web.dev/)
  * You may get more detailed breakdowns by using BigQuery, setting up a CrUX dashboard, or using the API directly.  This can be useful to get insights when vitals issues affect only a specific subset of your overall user population (for example).
[...]

PR #4693, opened 2022-05-26
Add loading indicator to sidebar links

yousefelgoharyx:
Add feature explained in #4691.
The style of the loading indicator needs to aligned with the design system of the docs. I am just literally adding "Loading..." as a string to the pending route.
I didn't test thoroughly tbh because I have exams lol so excuse me if something does not work well.
Wish me luck.

### DIFF
--- a/beta/src/hooks/usePendingRoute.ts
+++ b/beta/src/hooks/usePendingRoute.ts
@@ -1,3 +1,7 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ */
+
 import {useRouter} from 'next/router';
 import React from 'react';
 
--- a/beta/src/hooks/usePendingRoute.ts
+++ b/beta/src/hooks/usePendingRoute.ts
@@ -15,15 +15,18 @@ const usePendingRoute = () => {
         setPendingRoute(url);
       }, 100);
     };
-    const handleRouteChangeComplete = () => {
+    const cleanupPendingState = () => {
       setPendingRoute(null);
       clearTimeout(routeTransitionTimer);
     };
     events.on('routeChangeStart', handleRouteChange);
-    events.on('routeChangeComplete', handleRouteChangeComplete);
+    events.on('routeChangeComplete', cleanupPendingState);
+    events.on('routeChangeError', cleanupPendingState);
     return () => {
       events.off('routeChangeStart', handleRouteChange);
-      events.off('routeChangeComplete', handleRouteChangeComplete);
+      events.off('routeChangeComplete', cleanupPendingState);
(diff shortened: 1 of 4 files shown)


yousefelgoharyx:
When route transition is pretty fast I think the loading indicator would be useless and it actually looks bad or maybe when the loading indicator is a style (not text) it wouldn't be as obvious. I don't know

gaearon:
Good luck with exams!

yousefelgoharyx:
Should I continue improving on this or is the method am using do not cover all use cases?

gaearon:
Let's try to make it similar to the "hover" state? Maybe slightly darker.

yousefelgoharyx:
I made the pending style like hover but a bit lighter. I also added a `settimeout` to route transition so that if the route transition takes less than 100ms it will not show any pending indicator since it would be useless. I also extracted the pending route logic into a separate hook

beta/src/hooks/usePendingRoute.ts
```diff
@@ -0,0 +1,33 @@
...
+        currentRoute.current = url;
+        setPendingRoute(url);
+      }, 100);
+    };
+    const handleRouteChangeComplete = () => {
+      setPendingRoute(null);
+      clearTimeout(routeTransitionTimer);
+    };
+    events.on('routeChangeStart', handleRouteChange);
+    events.on('routeChangeComplete', handleRouteChangeComplete);
```
xiel, line 23: I think we need to handle routeChangeError (abort) as well, in case the user clicks another link or the back button, while the route is still loading. Or clear the timeout before setting up a new one in line 12?

beta/src/hooks/usePendingRoute.ts
```diff
@@ -0,0 +1,33 @@
...
+    return () => {
+      events.off('routeChangeStart', handleRouteChange);
+      events.off('routeChangeComplete', handleRouteChangeComplete);
```
xiel, line 26: We should also clear the timeout in cleanup.

yousefelgoharyx:
I did the routeChangeError approach and I cleared timeout in cleanup fn but could you tell me why is it important?
I mean one of the events listener will eventually work clearing up the timeout right?
so I can think that's just best practice to clear up the timeout in the cleanup fn

yousefelgoharyx:
I removed the routeChangeError event handler because for some reason if I click on the link twice, Next will assume this as an error or abort. I concluded that when I tried clicking on a link twice and the pending style would just disappear. You can try logging anything when routeChangeError fires and it will log whenever you click on a link twice

gaearon:
Looks great. Thank you.

PR #4694, opened 2022-05-26
[Beta] Upgrade React Types & Fix resulting type issues

xiel:
- make new APIs available in types, e.g. useId, useTransition etc.
- fixed some type errors


gaearon:
Thanks!

PR #4695, opened 2022-05-26
[Beta] Drop next/dynamic usage

SukkaW:
Ref: https://github.com/vercel/next.js/issues/37197#issuecomment-1138524590

Replace the usage of `next/dynamic` with `React.lazy` (`next/dynamic` with `{ suspense: true }` use `React.lazy` directly under the hood).

By dropping `next/dynamic`, we can also remove `react-lodable` client-side implementation (which is for `{ suspense: false }`) from emitted bundle (see https://github.com/vercel/next.js/issues/37197#issuecomment-1138496911).

### DIFF
--- a/beta/src/components/MDX/Sandpack/index.tsx
+++ b/beta/src/components/MDX/Sandpack/index.tsx
@@ -3,10 +3,9 @@
  */
 
 import * as React from 'react';
-import dynamic from 'next/dynamic';
 import {createFileMap} from './createFileMap';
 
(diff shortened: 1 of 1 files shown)


harish-sethuraman (approved):

SukkaW:
I only realize that after `Next.js Bundle Analysis` action commented: this one line of change reduces the bundle size by `1.34 KB`!

harish-sethuraman:
Yep! Interesting

PR #4702, opened 2022-05-27
[beta] fix(eslint): messages severity

danilowoz:
**This is more a question than a proposal:**

Originally, the `react-hooks/exhaustive-deps` eslint rule throws a warning in the React Docs implementations, and actually whatever you use this rule (vscode, cli, etc). However, currently, this rule is leading to a blocking state in the preview component, even showing a `Lint error` title there. 

So, taking into account the educational purposes of the documentation should we treat all rules as an error or leave the exhaustive-deps rule as a warning and update the preview error? 

**Current**
Screenshot 2022-05-27 at 13 55 38

**This PR**
Screenshot 2022-05-27 at 13 55 00

### DIFF
--- a/beta/src/components/MDX/Sandpack/useSandpackLint.tsx
+++ b/beta/src/components/MDX/Sandpack/useSandpackLint.tsx
@@ -29,7 +29,9 @@ export const useSandpackLint = () => {
         // Ignore parsing or internal linter errors.
         const isReactRuleError = (error: any) => error.ruleId != null;
         setLintErrors(errors.filter(isReactRuleError));
-        return codeMirrorErrors.filter(isReactRuleError);
+        return codeMirrorErrors
+          .filter(isReactRuleError)
+          .map((error) => ({...error, severity: 'error'}));
       });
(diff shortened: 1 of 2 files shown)


beta/src/components/MDX/Sandpack/useSandpackLint.tsx
```diff
@@ -29,7 +29,9 @@ export const useSandpackLint = () => {
         // Ignore parsing or internal linter errors.
         const isReactRuleError = (error: any) => error.ruleId != null;
         setLintErrors(errors.filter(isReactRuleError));
-        return codeMirrorErrors.filter(isReactRuleError);
+        return codeMirrorErrors
+          .filter(isReactRuleError)
+          .map((error) => ({...error, severity: 'error'}));
```
danilowoz, line 34: I could update it directly in the `runEslint` file, but I'd do all the customizations in the UI layer and leave the eslint integration as pure as possible.

gaearon:
Let's just change the lint configuration to treat it as an error.

PR #4710, opened 2022-05-30
fix typo in higher-order-components.md

awxiaoxian2020:
In my perception, this "to" should not be omitted. Adding it would make it clearer.

### DIFF
--- a/content/docs/higher-order-components.md
+++ b/content/docs/higher-order-components.md
@@ -124,7 +124,7 @@ const BlogPostWithSubscription = withSubscription(
 
 The first parameter is the wrapped component. The second parameter retrieves the data we're interested in, given a `DataSource` and the current props.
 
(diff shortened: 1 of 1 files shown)


eps1lon (changes requested):
they're being "passed a data prop" i.e. `<CommentList data />` and not "to a data prop" i.e. `<SomeComponent data={CommentList } />`

awxiaoxian2020:
> they're being "passed a data prop" i.e. `<CommentList data />` and not "to a data prop" i.e. `<SomeComponent data={CommentList } />`

I think we should change the expressions. The grammar is not correct. The passive shouldn't have objects. You mean that "they are being `as` passed a `as` data prop"?

eps1lon:
>  You mean that "they are being passed as a data prop"?

I mean that a `data` prop is being passed to them.

awxiaoxian2020:
> > You mean that "they are being passed as a data prop"?
> 
> I mean that a `data` prop is being passed to them.

So why don't we change the expression like what you said?
The current doc is saying the opposite meanings. I think.

Issue #4619, opened 2022-05-01
The "Support Ukraine 🇺🇦" banner should disappear when scrolling


HyunseungLee-Travis:
# The point of this banner
The reason the banner exists is to help people notify that there is a way to help Ukraine people.
However, making them permanent is not helping as it can intefere with the reading.

# The problem
1. Scrolling through contents is very hard to do because of the banner
2. Reading the contents are harder as the banner takes up to 10% to 50% of the existing mobile page
3. If people first look at the banner, they will either help or not help and they won't be likely to change their mind as they scroll through the contents. 

# How to fix
I believe it is right to make the banner disappear when scrolling through contents like [React Native Website](https://reactnative.dev). Comparing two websites, the website looks much better without the banner when scrolling.

# Clarification
I am not stating that the banner should be removed, only suggesting it should disappear when scrolled.
It is much more better to learn when the banner is not present every time in computers. phones, and ipads.

# Screenshots comparison
![react](https://user-images.githubusercontent.com/78584173/166144378-ea336606-1c78-41f6-8875-bed9d8f9b652.png)
![reactfixed](https://user-images.githubusercontent.com/78584173/166144415-51c9a773-2e15-4cd6-bdf0-30d73d0adcfd.png)

nurbashanghai:
I would like to work on this issue, let me raise a PR

HyunseungLee-Travis:
> I would like to work on this issue, let me raise a PR 

If you do, please mention it in this issue.

nurbashanghai:
yes, im currently working on it, will raise pr later today

nurbashanghai:
https://github.com/reactjs/reactjs.org/pull/4713 this is my pull request

HyunseungLee-Travis:
Works perfectly.

Rayanysilva12777:
![Uploading SAVE_20220529_155236.jpg…]()

Rayanysilva12777:
Eu

Rayanysilva12777:
Quero ver meus arquivos messa foto

HyunseungLee-Travis:
Please upload the picture again and use english, please for the majority.

Rayanysilva12777:
I need to see . behold I need to see my files for this photo

aniket1010:
is someone working on this issue?

harish-sethuraman:
Yes there are already two PRs for this issue.

harish-sethuraman:
[...]

PR #4713, opened 2022-05-31
hide social banner on scroll

nurbashanghai:
Addressing this issues "The "Support Ukraine 🇺🇦" banner should disappear when scrolling" #4619 

I simply added event listener to the SocialBanner component. It behaves the same way as a https://reactnative.dev React Native's website's Social banner, it disappears when we scroll down on the page.

### DIFF
--- a/beta/src/components/Layout/Page.tsx
+++ b/beta/src/components/Layout/Page.tsx
@@ -21,7 +21,7 @@ export function Page({routeTree, children}: PageProps) {
       <MenuProvider>
         <SidebarContext.Provider value={routeTree}>
           <div className="h-auto lg:h-screen flex flex-row">
-            <div className="no-bg-scrollbar h-auto lg:h-[calc(100%-40px)] lg:overflow-y-scroll fixed flex flex-row lg:flex-col py-0 top-16 sm:top-10 left-0 right-0 lg:max-w-xs w-full shadow lg:shadow-none z-50">
+            <div className="no-bg-scrollbar h-auto lg:h-[calc(100%-40px)] lg:overflow-y-scroll fixed flex flex-row lg:flex-col py-0 left-0 right-0 lg:max-w-xs w-full shadow lg:shadow-none z-50">
               <Nav />
               <Sidebar />
             </div>
--- a/beta/src/components/Layout/Page.tsx
+++ b/beta/src/components/Layout/Page.tsx
@@ -15,13 +15,25 @@ interface PageProps {
 }
 
 export function Page({routeTree, children}: PageProps) {
+  const [showBanner, setShowBanner] = React.useState(true);
+  const updateVisibility = () => setShowBanner(window.scrollY <= 6);
+  React.useEffect(() => {
+    setShowBanner(window.scrollY <= 6);
+    window.addEventListener('scroll', updateVisibility);
+    return () => window.removeEventListener('scroll', updateVisibility);
+  }, []);
   return (
     <>
-      <SocialBanner />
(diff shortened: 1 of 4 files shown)


nurbashanghai:
https://user-images.githubusercontent.com/63287548/171123668-17773601-1157-478b-ab4b-bc6592503208.mov

gaearon:
Let's also fix the `scroll-margin-top` value for headers to exclude the banner height.

gaearon:
It would be great to also figure out something similar for https://beta.reactjs.org/ (in the `beta` directory).

nurbashanghai:
Thanks for feedback! I will work on that today

nurbashanghai:
https://user-images.githubusercontent.com/63287548/171781248-ce66f654-2037-446a-8514-5d34661d6195.mov

nurbashanghai:
> Screen.Recording.2022-06-03.at.09.35.33.mov

This is how it works in beta folder

nurbashanghai:
> It would be great to also figure out something similar for https://beta.reactjs.org/ (in the `beta` directory).

Done

harish-sethuraman:
https://user-images.githubusercontent.com/32865581/171781795-285d5636-b11f-4430-a8dd-cbb90490b1a4.mov

I think it would be nice if we don't do the same for desktop? It kind of shifts the entire page and doesn't go with the flow? Thoughts?

nurbashanghai:
> Screen.Recording.2022-06-03.at.9.12.44.AM.mov 
> 
> I think it would be nice if we don't do the same for desktop? It kind of shifts the entire page and doesn't go with the flow? Thoughts?

Let me make it smoother :)

nurbashanghai:
> Screen.Recording.2022-06-03.at.9.12.44.AM.mov 
> 
> I think it would be nice if we don't do the same for desktop? It kind of shifts the entire page and doesn't go with the flow? Thoughts?

Yeah, I think we need to leave banner on the desktop, but turn it off on mobile screens

harish-sethuraman:
Yep we can make it smoother like in the https://reactnative.dev/ or we can leave the banner as such in desktop site since it doesn't take much height anyways

nurbashanghai:
Screen Shot 2022-06-03 at 16 10 45
I updated a branch, now social banner is always present in desktop version. But on small screens it will disappear on scroll down

nurbashanghai:
@harish-sethuraman @gaearon PR is ready guys) waiting for feedback :)

nurbashanghai:
Hey guys! Since I made all requested changes, maybe we can proceed to merge this pull request? 
Or I can make any other needed changes to the PR if needed.
I fear that this PR will just forever stuck here :)

harish-sethuraman:
Can we slightly soften the hiding of banner in mobile alone? Other than that it LGTM

nurbashanghai:
Yes we definitely can!) I will work on that today

beta/src/components/Layout/Page.tsx
```diff
@@ -15,13 +15,25 @@ interface PageProps {
 }
 
 export function Page({routeTree, children}: PageProps) {
+  const [showBanner, setShowBanner] = React.useState(true);
+  const updateVisibility = () => setShowBanner(window.scrollY <= 6);
```
gaearon, line 19: It would be nice to think of some way to not run `set*` function on every single scroll event (even if it bails out). Could we do that?

nurbashanghai: Yes, let me change it

harish-sethuraman:
closed by https://github.com/reactjs/reactjs.org/pull/5002

PR #4723, opened 2022-06-04
Update the URL of Yarn's installation instructions

xfq:


README.md
```diff
@@ -8,7 +8,7 @@ This repo contains the source code and documentation powering [reactjs.org](http
 
 1. Git
 1. Node: any 12.x version starting with v12.0.0 or greater
-1. Yarn: See [Yarn website for installation instructions](https://yarnpkg.com/lang/en/docs/install/)
+1. Yarn: See [Yarn website for installation instructions](https://yarnpkg.com/getting-started/install)
```
eps1lon, line 11: I think we should rather change it to 
```suggestion
1. Yarn: See [Yarn website for installation instructions](https://classic.yarnpkg.com/en/docs/getting-started)
```
since this repo needs Yarn v1 not v2+

xfq: Sorry, I thought this repo needs v2+ because I installed the npm dependencies with yarn 3.2.1 and it worked fine.

Is there a plan to upgrade the yarn version in the future (in the [New React Docs](https://github.com/reactjs/reactjs.org/issues/3308), maybe)?

eps1lon: Right now everybody is busy landing the beta docs. facebook/react is also still on yarn v1 so I doubt this'll change anytime soon. 

But that's not really relevant here.

xfq: I fully understand that this will not change anytime soon.

I have updated the PR to make it clear that this repo needs Yarn v1. Is that OK?

eps1lon (approved):
Thanks!

PR #4732, opened 2022-06-09
docs: fix optional argument representation

kbrandwijk:
This change was introduced in https://github.com/reactjs/reactjs.org/pull/2164, but is mildly inconsistent with the existing syntax in line 542, that follows the original intent from https://github.com/reactjs/reactjs.org/commit/0560d6be8b343b63c4412d35204d5b24847a0eac so I'd like to see it reverted.


wodin:
Makes sense to me :)

harish-sethuraman (approved):
Thanks!

PR #4758, opened 2022-06-18
Change plural sentence to singular

peterhauke:

### DIFF
--- a/content/docs/reconciliation.md
+++ b/content/docs/reconciliation.md
@@ -45,7 +45,7 @@ This will destroy the old `Counter` and remount a new one.
 
 >Note:
 >
->These methods are considered legacy and you should [avoid them](/blog/2018/03/27/update-on-async-rendering.html) in new code:
+>This method is considered legacy and you should [avoid it](/blog/2018/03/27/update-on-async-rendering.html) in new code:
 >
 >- `UNSAFE_componentWillMount()`
 


content/docs/reconciliation.md
```diff
@@ -45,7 +45,7 @@ This will destroy the old `Counter` and remount a new one.
 
 >Note:
 >
->These methods are considered legacy and you should [avoid them](/blog/2018/03/27/update-on-async-rendering.html) in new code:
+>This method is considered legacy and you should [avoid it](/blog/2018/03/27/update-on-async-rendering.html) in new code:
```
harish-sethuraman, line 48: There are two methods mentioned below 
- UNSAFE_componentWillUpdate()
- UNSAFE_componentWillReceiveProps()
And the already existing version seems to make sense.

peterhauke: Yes, I saw the other note below. There is only one method listed in the first note, so the sentence does not read correctly when it is in the plural, but I agree, it does seem to make sense.

Maybe, the following would be more clear and better:

Note:
Methods prefixed with UNSAFE_ are considered to be legacy and you should avoid them in new code, for example:
* UNSAFE_componentWillMount()

and later

Note:
Methods prefixed with UNSAFE_ are considered to be legacy and you should avoid them in new code, for example:
* UNSAFE_componentWillUpdate()
* UNSAFE_componentWillReceiveProps()

If you agree, then I will make another PR with this proposal.

harish-sethuraman: Hey really sorry. Now it makes sense to me I missed out the first note.

harish-sethuraman:
Before:
Screenshot 2022-06-19 at 9 49 23 PM

After:
Screenshot 2022-06-19 at 9 49 06 PM

peterhauke:
Thanks, @harish-sethuraman

I understand. It happens. I would go with the original PR then.

Do I need to make another PR or is this one OK now?

content/docs/reconciliation.md
peterhauke: Hello @harish-sethuraman , can you merge my original commit if it is OK with you?

harish-sethuraman:
Thanks!

PR #4760, opened 2022-06-19
[Removed] console debuggings

rautNitesh:
from the createPages.js


gaearon:
if it's there, it's probably intentional.

PR #4777, opened 2022-06-27
Events deep dive

yousefelgoharyx:
I think it is vital for readers to at least get a sneak peak into what synthetic events are. If some learner for some reason decides to log the received synthetic event, confusion will arise and it is not beautiful to see an instance of SyntheticEvent since the learner has an assumption that it is a regular Event object.


yousefelgoharyx:
I just noticed the formatter ruined the pull request lol

PR #4784, opened 2022-06-30
[Beta] Fix empty message styling in console

harish-sethuraman:
Should we probably stop empty consoles getting added inside? When the user is entering console.log it adds an empty console statement inside the console. which might be confusing? Or we might want to increase the debounce timing for updating the preview once the code is updated.

https://user-images.githubusercontent.com/32865581/176703214-87cdbafc-6dd2-4050-826b-5d062e388710.mov


harish-sethuraman:
Updated :) to use min height of the padding + text’s height.

PR #4787, opened 2022-07-01
[Beta] Remove prefetch for navlinks in the beta site

harish-sethuraman:


gaearon:
Removing them everywhere might be a bit much?

harish-sethuraman:
Main branch:
Screenshot 2022-07-01 at 9 37 11 PM
This branch: 
Screenshot 2022-07-01 at 9 38 09 PM



> Removing them everywhere might be a bit much? 
Yea kinda I wanted to check with the coverage for diff components. This is after removing the links inside the markdowns + navigations (verified in page - https://beta.reactjs.org/learn/describing-the-ui)

harish-sethuraman:
One reason why I removed from markdown anchors as well is because we have this in this chapter section where all of the open nav items are displayed in the markdown landing as well. So when we click on that page it results in prefetching the markdown although we have disabled it in the nav bar. 

So where do you think we need to disable prefetching? would it make sense to disable prefetching the route's script?

PR #4792, opened 2022-07-05
feat: hide social banner on scroll

akulsr0:
This hides social banner when scrolled on mobile/tablet.

**Demo**
ezgif com-gif-maker

### DIFF
--- a/beta/src/components/Layout/Sidebar/Sidebar.tsx
+++ b/beta/src/components/Layout/Sidebar/Sidebar.tsx
@@ -13,6 +13,7 @@ import {MobileNav} from '../Nav/MobileNav';
 import {Feedback} from '../Feedback';
 
 const SIDEBAR_BREAKPOINT = 1023;
+const SM_BREAKPOINT = 640;
 
 export function Sidebar() {
   const {menuRef, isOpen} = React.useContext(MenuContext);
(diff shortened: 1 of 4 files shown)


akulsr0:
@harish-sethuraman @gaearon any feedback/suggestions here?

harish-sethuraman:
Hey 👋🏻  There is one another PR that fixes the same thing 
- https://github.com/reactjs/reactjs.org/pull/4713

akulsr0:
@harish-sethuraman Yeah, I see there were some changes requested and that pr was not active ~1 month, so I tried a different approach here to fix the issue.
Since on smaller devices, the social banner is taking much space, it would be good to fix this with either approach :)

akulsr0:
Bump?

beta/src/components/Layout/Sidebar/Sidebar.tsx
```diff
@@ -13,32 +13,44 @@ import {MobileNav} from '../Nav/MobileNav';
...
 export function Sidebar() {
   const {menuRef, isOpen} = React.useContext(MenuContext);
   const isMobileSidebar = useMediaQuery(SIDEBAR_BREAKPOINT);
   let routeTree = React.useContext(SidebarContext);
   const isHidden = isMobileSidebar ? !isOpen : false;
 
+  const bannerHeight = useMediaQuery(SM_BREAKPOINT) ? 72 : 40;
+  let navMarginTop = bannerHeight;
+  if (typeof window !== 'undefined') {
+    const scrollAmount = document.documentElement.scrollTop;
```
gaearon, line 27: We shouldn't be reading global mutable values during render. This usually indicates a bug (even when it's not obvious how to trigger it).

beta/src/components/Layout/Sidebar/Sidebar.tsx
```diff
@@ -13,32 +13,44 @@ import {MobileNav} from '../Nav/MobileNav';
...
 
 export function Sidebar() {
   const {menuRef, isOpen} = React.useContext(MenuContext);
   const isMobileSidebar = useMediaQuery(SIDEBAR_BREAKPOINT);
   let routeTree = React.useContext(SidebarContext);
   const isHidden = isMobileSidebar ? !isOpen : false;
 
+  const bannerHeight = useMediaQuery(SM_BREAKPOINT) ? 72 : 40;
+  let navMarginTop = bannerHeight;
+  if (typeof window !== 'undefined') {
```
gaearon, line 26: Different render outputs for server and initial client output means hydration can produce mismatches. This doesn't seem right.

gaearon (changes requested):
From the coding perspective, this solution doesn't seem to follow React's rules. There should be the same render output for server and initial client, and render shouldn't read mutable values. So I don't think we can merge this.

PR #4795, opened 2022-07-06
fix typo in passing-data-deeply-with-context.md

sunil-sharma-999:
Fixes https://github.com/facebook/react/issues/24858


harish-sethuraman (changes requested):
I see there are more references for the word `header` inside the page. can you fix them as well?

harish-sethuraman:
LGTM thanks :)

PR #4801, opened 2022-07-07
Remove Singapore from the list

zanechua:
React Knowledgeable website seems to go to some spam website now and the community facebook page has been updated since 2019. Probably no longer active

https://www.facebook.com/reactknowledgeable/


harish-sethuraman (approved):
Thanks for the PR :)

PR #4806, opened 2022-07-09
fix: api links in content

akulsr0:


harish-sethuraman (changes requested):
There is one another link that throws 404 at page https://beta-reactjs-k05yeaedb-fbopensource.vercel.app/apis/reactdom would you like to fix that as well?

akulsr0:
@harish-sethuraman done!

harish-sethuraman (approved):
Thanks :)

PR #4807, opened 2022-07-10
fix: sandbox preview height

akulsr0:
### Before 
before

### After
after

### DIFF
--- a/beta/src/styles/sandpack.css
+++ b/beta/src/styles/sandpack.css
@@ -290,7 +290,7 @@ html.dark .sp-devtools > div {
   min-width: 431px;
   /* Keep preview a fixed size on mobile to avoid jumps. */
   /* This is because we don't know its content in advance. */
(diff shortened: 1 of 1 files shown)


harish-sethuraman (commented):
If we reduce the height here some sandpack output screens might get a scroll to view the output. like the last one in the thinking in react page.

akulsr0:
@harish-sethuraman The last sandbox (in thinking in react) is already having scroll in sandbox preview in current beta website? What is the expected output in this case?

harish-sethuraman:
We have explicit height here because the sandpack's output window are killed when they move out of the screen. so the height of iframe becomes 0 how ever when we use fit content the height is shrinking and then the content move upwards. But we dont want this to happen. So we are explicitly setting the height to be 40vh.

lunaleaps:
I'll close this PR for now.

PR #4817, opened 2022-07-13
closed #4817

lxmarinkovic:


harish-sethuraman:
Hey, Thanks for sending an PR how ever we are not taking PRs for adding and removing punctuations from older docs and we'd like to leave it as such.

lxmarinkovic:
Thanks for the info @harish-sethuraman  ; it was just an idea since I noticed some errors.
Anyway, can you tell me how I can level up my repo role here ✍️

PR #4831, opened 2022-07-21
[Beta] Adjusting Effect Dependencies

gaearon:
## TODO

- [x] Writing
- [ ] Team review
- [ ] Edit/simplify pass
- [ ] Recap
- [ ] Challenges

### DIFF
--- a/beta/src/pages/learn/adjusting-effect-dependencies.md
+++ b/beta/src/pages/learn/adjusting-effect-dependencies.md
@@ -569,10 +569,10 @@ function ChatRoom({ roomId }) {
     }
 
     const connection = createConnection(roomId);
-    connection.on('message', handleMessage);
+    connection.addListener('message', handleMessage);
     connection.connect();
     return () => {
-      connection.on('message', handleMessage);
+      connection.removeListener('message', handleMessage);
       connection.disconnect();
     };
   }, [roomId, messages]); // ✅ All dependencies declared
@@ -595,10 +595,10 @@ function ChatRoom({ roomId }) {
     }
 
     const connection = createConnection(roomId);
-    connection.on('message', handleMessage);
+    connection.addListener('message', handleMessage);
     connection.connect();
     return () => {
-      connection.on('message', handleMessage);
+      connection.removeListener('message', handleMessage);
       connection.disconnect();
     };
   }, [roomId]); // ✅ All dependencies declared
@@ -629,10 +629,10 @@ function ChatRoom({ roomId }) {
     }
 
     const connection = createConnection(roomId);
-    connection.on('message', handleMessage);
+    connection.addListener('message', handleMessage);
     connection.connect();
     return () => {
-      connection.on('message', handleMessage);
+      connection.removeListener('message', handleMessage);
       connection.disconnect();
     };
   }, [roomId]); // 🔴 React Hook useEffect has a missing dependency: 'isMuted'
@@ -664,10 +664,10 @@ function ChatRoom({ roomId }) {
 
   useEffect(() => {
     const connection = createConnection(roomId);
-    connection.on('message', handleMessage);
+    connection.addListener('message', handleMessage);
     connection.connect();
     return () => {
-      connection.on('message', handleMessage);
+      connection.removeListener('message', handleMessage);
       connection.disconnect();
     };
   }, [roomId]); // ✅ All dependencies declared
@@ -693,10 +693,10 @@ function ChatRoom({ roomId, onMessage }) {
 
   useEffect(() => {
     const connection = createConnection(roomId);
-    connection.on('message', handleMessage);
+    connection.addListener('message', handleMessage);
     connection.connect();
     return () => {
-      connection.on('message', handleMessage);
+      connection.removeListener('message', handleMessage);
       connection.disconnect();
     };
   }, [roomId]); // ✅ All dependencies declared
@@ -1214,10 +1214,10 @@ function ChatRoom({ roomId }) {
 
   useEffect(() => {
     const connection = createConnection(roomId);
-    connection.on('message', handleMessage);
+    connection.addListener('message', handleMessage);
     connection.connect();
     return () => {
-      connection.on('message', handleMessage);
+      connection.removeListener('message', handleMessage);
       connection.disconnect();
     };
   }, [roomId, handleMessage]); // ✅ All dependencies declared
@@ -1248,10 +1248,10 @@ function ChatRoom({ roomId }) {
 
   useEffect(() => {
     const connection = createConnection(roomId);
-    connection.on('message', handleMessage);
+    connection.addListener('message', handleMessage);
     connection.connect();
     return () => {
-      connection.on('message', handleMessage);
+      connection.removeListener('message', handleMessage);
       connection.disconnect();
     };
   }, [roomId]); // ✅ All dependencies declared
--- a/beta/src/pages/learn/adjusting-effect-dependencies.md
+++ b/beta/src/pages/learn/adjusting-effect-dependencies.md
@@ -1,0 +1,1345 @@
+---
+title: 'Adjusting Effect Dependencies'
+---
+
+<Intro>
+
+An Effect re-runs after any change to the values that it depends on. This is why writing Effects requires a different mindset than writing event handlers. You'll write the code inside your Effect first. Then you'll specify its dependencies according to the code you already wrote. Finally, if a dependency changes too often and causes your Effect to re-run more than needed, you'll have to adjust the code so that it *does not need* that dependency, and follow these steps again. It's not always obvious how to do this, so this page will help you learn common patterns and idioms.
+
+</Intro>
+
+<YouWillLearn>
+
+- How writing Effects is different from writing event handlers
+- What causes Effects to re-run
+- How to choose the Effect dependencies
+- How the fix the dependency linter errors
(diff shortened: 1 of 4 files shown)


beta/src/pages/learn/adjusting-effect-dependencies.md
```diff
@@ -0,0 +1,1345 @@
...
+    function handleMessage(receivedMessage) {
+      // 🔴 Problem: makes the Effect re-run
+      setMessages([...messages, receivedMessage]);
+    }
+
+    const connection = createConnection(roomId);
+    connection.on('message', handleMessage);
+    connection.connect();
+    return () => {
+      connection.on('message', handleMessage);
```
sophiebits, line 575: should this be `off` or something? (several times in this file)

beta/src/pages/learn/adjusting-effect-dependencies.md
```diff
@@ -0,0 +1,1345 @@
...
+  }, [roomId]); // 🔴 React Hook useEffect has a missing dependency: 'isMuted'
+
+  function handleMuteClick() {
+    setIsMuted(!isMuted);
+  }
+
+  // ...
+```
+
+There is a linter error in the above example. Since you read `isMuted` inside your Effect, you must add `isMuted` as a dependency. However, if you add `isMuted` to the dependency array, every Mute button press will re-run the Effect (because its dependency `isMuted` has changed) and reconnect to the chat server. It's bad that pressing the Mute button reconnects to the server. **You want the `handleMessage` function to read the latest value of the `isMuted` state variable, but you don't want to your Effect to re-run when the `isMuted` state variable changes.**
```
sophiebits, line 647: mention that without adding the dep, isMuted will be stale?

beta/src/pages/learn/adjusting-effect-dependencies.md
```diff
@@ -0,0 +1,1345 @@
...
+  }, [roomId]); // ✅ All dependencies declared
+
+  function handleMuteClick() {
+    setIsMuted(!isMuted);
+  }
+
+  // ...
+```
+
+Code inside Effects is reactive. It re-runs when the values inside change. If you don't want a part of your Effect's code to be reactive, extract it to an Event function like `handleMessage` above, and call that Event function from your Effect. **Event functions are omitted from dependencies. Editing them won't make any code run more often.**
```
sophiebits, line 682: > are omitted

explain why this is "allowed", unlike every other value?

beta/src/pages/learn/adjusting-effect-dependencies.md
```diff
@@ -0,0 +1,1345 @@
...
+    return () => {
+      connection.on('message', handleMessage);
+      connection.disconnect();
+    };
+  }, [roomId]); // ✅ All dependencies declared
+
+  // ...
+```
+
+If you tried to call `onMessage` directly from your Effect, you would have to include `onMessage` into your dependencies. This would cause your Effect to re-run whenever the parent component re-renders. However, `handleMessage` that you got from the `useEvent(onMessage)` declaration is an Event function. This means that you don't need to specify it in the Effect dependencies. Now it can never make your Effect re-run accidentally.
```
sophiebits, line 707: add something about even though the effect doesn't rerun, it'll call the latest value of onMessage? (latest committed value)

beta/src/pages/learn/adjusting-effect-dependencies.md
```diff
@@ -0,0 +1,1345 @@
...
+      eventName: 'visit_chat',
+      roomId: roomId,
+      notificationCount: notificationCount
+    });
+  }, [roomId]); // 🔴 React Hook useEffect has a missing dependency: 'notificationCount'
+  // ...
+}
+```
+
+You're using `notificationCount` inside your Effect, so you must add it to the dependency list. However, if you do that, your Effect will re-run whenever the `notificationCount` changes. This doesn't make sense: a change to the number of notification doesn't mean that the user visited the chat again.
```
sophiebits, line 755:
```suggestion
You're using `notificationCount` inside your Effect, so you must add it to the dependency list. However, if you do that, your Effect will re-run whenever the `notificationCount` changes. This doesn't make sense: a change to the number of notifications doesn't mean that the user visited the chat again.
```

beta/src/pages/learn/adjusting-effect-dependencies.md
```diff
@@ -0,0 +1,1345 @@
...
+  });
+
+  useEffect(() => {
+    handleVisit(roomId);
+  }, [roomId]); // ✅ All dependencies declared
+  // ...
+}
+```
+
+With this code, the first analytics log will include the initial `roomId` and the initial `notificationCount`. Then, if the `roomId` changes, your Effect will call `handleVisit` again and send a new analytics log with that `roomId` and the current `notificationCount`. However, if the `notificationCount` count changes alone, nothing will happen. **Code inside Event functions is not reactive, so editing `handleVisit` won't make your Effect run more often.**
```
sophiebits, line 776: although you can kind of infer it, I think this needs some explicit commentary on why roomId is an argument instead of captured

beta/src/pages/learn/adjusting-effect-dependencies.md
```diff
@@ -0,0 +1,1345 @@
...
+  useEffect(() => {
+    const connection = createConnection(roomId, options);
+    connection.connect();
+    return () => connection.disconnect();
+  }, [roomId, options]); // ✅ All dependencies declared
+  // ...
+}
+```
+
+You can't know whether the parent component will pass an `options` object that's "stable" (i.e. does not change between re-renders) or if it's going to create a completely different object every time that it re-renders.
```
sophiebits, line 937: this example rewrite is fine but I feel like it's also legit to have stable identity as part of the contract for the component, especially if the options type is from a third-party (here, websocket) library – saying "connectionOptions prop; each time it changes the connection will be reestablished; please useMemo" – wdyt?

beta/src/pages/learn/adjusting-effect-dependencies.md
```diff
@@ -0,0 +1,1345 @@
...
+}
+```
+
+When your component mounts, your Effect will *run*, connecting to the room with the initial `roomId`. When it receives a different `roomId`, your Effect will *[clean up](/learn/synchronizing-with-effects#step-3-add-cleanup-if-needed)* (disconnecting from the previous room) and *run* again (connecting to the next room). When your component unmounts, the Effect will *clean up* one last time.
+
+Let's recap what has happened:
+
+1. You edited the code, potentially introducing a bug.
+2. The linter found the bug and suggested a change *to make the dependencies match the Effect code*.
+3. You accepted the linter's suggestion, which (correctly) *made your Effect run more often than before*.
```
sophiebits, line 137: I appreciate these well-placed parentheticals

beta/src/pages/learn/adjusting-effect-dependencies.md
```diff
@@ -0,0 +1,1345 @@
...
+function Form() {
+  const [submitted, setSubmitted] = useState(false);
+
+  useEffect(() => {
+    if (submitted) {
+      // 🔴 Avoid: Event-specific logic inside an Effect
+      post('/api/register');
+      showToast('Successfully registered!');
+    }
+  }, [submitted]); // ✅ All dependencies declared
```
sophiebits, line 264: I feel like it is a tiny bit confusing to have the green checkmark in the same code sample as the red Avoid. almost feels like it is contradicting it. ("yes there was an error above, but this makes up for it.") wdyt of just "// All dependencies declared" for these?

sophiebits (commented):
a few notes

beta/src/pages/learn/adjusting-effect-dependencies.md
```diff
@@ -0,0 +1,1345 @@
...
+
+  function handleSubmit() {
+    setSubmitted(true);
+  }  
+
+  // ...
+}
+```
+
+Later, you want to style the notification toast according to the current theme, so you read the current theme. Since `theme` is declared in the component body, it is a reactive value, and you must declare it as a depedency:
```
harish-sethuraman, line 274:
```suggestion
Later, you want to style the notification toast according to the current theme, so you read the current theme. Since `theme` is declared in the component body, it is a reactive value, and you must declare it as a dependency:
```

beta/src/pages/learn/you-might-not-need-an-effect.md
```diff
@@ -358,6 +358,98 @@ function Form() {
...
+      }
+    }
+  }
+
+  // ...
+```
+
+This is a lot more efficient. Also, if you implement a way to view game history, now you will be able to set each state variable to a move from the past without triggering the Effect chain that adjusts every other value. If you need to reuse logic between several event handlers, you can [extract a function](#sharing-logic-between-event-handlers) and call it from those handlers.
+
+Remember that inside event handlers, [state behaves like a snapshot](/learn/state-as-a-snapshot). For example, even after you call `setRound(rount + 1)`, the `round` variable will reflect the value at the time the user clicked the button. If you need to use the next value for calculations, define it manually like `const nextRound = round + 1`.
```
harish-sethuraman, line 449:
```suggestion
Remember that inside event handlers, [state behaves like a snapshot](/learn/state-as-a-snapshot). For example, even after you call `setRound(round + 1)`, the `round` variable will reflect the value at the time the user clicked the button. If you need to use the next value for calculations, define it manually like `const nextRound = round + 1`.
```

PR #4835, opened 2022-07-23
Update constant name in useImperative example

Nightrider0098:
User Imperative example has to same ref name in child and in the parent. Causing confusion while understanding the concept. 

`function FancyInput(props, ref) {
  const **inputRef** = useRef();
  useImperativeHandle(ref, () => ({
    focus: () => {
      inputRef.current.focus();
    }
  }));
  return <input ref={**inputRef**} ... />;
}
FancyInput = forwardRef(FancyInput);`

`In this example, a parent component that renders <FancyInput ref={**inputRef**} /> would be able to call inputRef.current.focus().'


Nightrider0098:
@sunny  minor pr, can you please look into it?

sunny:
Wrong @sunny I think :)

Nightrider0098:
> Wrong @sunny I think :)

updated my comment. can you pls review again?

Nightrider0098:
@sunny shall I close the pr if the change suggested are wrong?

sunny:
What I meant by “Wrong @sunny” is that you mentioned me, but I am not a reactjs.org maintainer.

Nightrider0098:
@gaearon can you pls review the pr?

PR #4841, opened 2022-07-27
Update how-to-contribute.md

awells111:
`yarn debug-test` is now deprecated in React package.json. It has been replaced by `yarn test --debug`.

### DIFF
--- a/content/docs/how-to-contribute.md
+++ b/content/docs/how-to-contribute.md
@@ -89,7 +89,7 @@ The core team is monitoring for pull requests. We will review your pull request 
 3. If you've fixed a bug or added code that should be tested, add tests!
 4. Ensure the test suite passes (`yarn test`). Tip: `yarn test --watch TestName` is helpful in development.
 5. Run `yarn test --prod` to test in the production environment.
-6. If you need a debugger, run `yarn debug-test --watch TestName`, open `chrome://inspect`, and press "Inspect".
+6. If you need a debugger, run `yarn test --debug --watch TestName`, open `chrome://inspect`, and press "Inspect".
 7. Format your code with [prettier](https://github.com/prettier/prettier) (`yarn prettier`).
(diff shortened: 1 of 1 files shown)


content/docs/how-to-contribute.md
```diff
@@ -89,7 +89,7 @@ The core team is monitoring for pull requests. We will review your pull request
 3. If you've fixed a bug or added code that should be tested, add tests!
 4. Ensure the test suite passes (`yarn test`). Tip: `yarn test --watch TestName` is helpful in development.
 5. Run `yarn test --prod` to test in the production environment.
-6. If you need a debugger, run `yarn debug-test --watch TestName`, open `chrome://inspect`, and press "Inspect".
+6. If you need a debugger, run `yarn test --debug --watch TestName`, open `chrome://inspect`, and press "Inspect".
```
harish-sethuraman, line 92: https://github.com/facebook/react/blob/2c7dea73659da283fa433f5fd657a5a29797fd67/package.json#L119

Just wondering? `debug-test` in turn runs `yarn test --debug` only right?

awells111: I've tried in both a bash console and in powershell and the result that I get from `yarn debug-test` is 
> Please run: \`'yarn\` instead.

From searching `please run:` in the repo, I found that on line 352 of `scripts\jest\jest-cli.js` there is the below code
```
  if (argv.deprecated) {
    console.log(chalk.red(`\nPlease run: \`${argv.deprecated}\` instead.\n`));
    return;
  }
```

I think that it's supposed to output `Please run: 'yarn test --debug'` but it's getting truncated, which is another problem that I can open an issue for pending the merge of this issue. To answer your question, it doesn't run `yarn test --debug`. I think it's only supposed to output a message to console of the command that should be run.

When I run `yarn test --debug` there are no issues and I can debug the application as expected.

PR #4842, opened 2022-07-27
[Beta] Update copy to clipboard

shanpriyan:
Set `copied` state to `true` only after clipboard's contents have been updated.


harish-sethuraman (approved):
LGTM but will leave for other reviewers to give their suggestions as well

PR #4848, opened 2022-07-29
[beta] refactor: remove unused state

shanpriyan:
- Removed `isLoaded` state, since it was always `true` and never changed
- Removed `Promise.all` and used `import()` directly, since `import()` was the only promise used


harish-sethuraman (approved):
LGTM Thanks :)

PR #4858, opened 2022-08-01
Adding Script component for inline script :)

sanjaiyan-dev:
Using `<Script />` component from `next/script` for inline colour mode script.


package.json
```diff
@@ -95,7 +95,5 @@
     "recursive-readdir": "^2.2.1",
     "unist-util-map": "^1.0.3"
   },
-  "resolutions": {
-    "**/sharp": "^0.30.3"
-  }
+  "packageManager": "yarn@3.2.1"
```
eps1lon, line 98: This is creating a large diff in the lockfile. This project requires yarn v1. If you use that for install you should not have any changes to the lockfile.

eps1lon:
> Using <Script /> component from next/script for inline colour mode script.

Can you elaborate what behavior you're seeing without this change, what behavior the change results in, and why we should change it?

eps1lon:
> Sorry if I made any mistakes raised_handsblush

I didn't mean to imply that. I was wondering what this change is doing in the first place or what it's intended to do.

package.json
sanjaiyan-dev: > This is creating a large diff in the lockfile. This project requires yarn v1. If you use that for install you should not have any changes to the lockfile.

Sorry, I think this error might happened while running `next:lint` :(

sanjaiyan-dev:
> > Sorry if I made any mistakes raised_handsblush
> 
> I didn't mean to imply that. I was wondering what this change is doing in the first place or what it's intended to do.

And i thought it is a best practise to use `next/script` :)

PR #4859, opened 2022-08-01
Update handling-events.md

suulola:
e argument was passed as the first argument in the second button example. Fix to make it the second so it matches what the docs describe

### DIFF
--- a/content/docs/handling-events.md
+++ b/content/docs/handling-events.md
@@ -140,7 +140,7 @@ Inside a loop, it is common to want to pass an extra parameter to an event handl
 
 ```js
 <button onClick={(e) => this.deleteRow(id, e)}>Delete Row</button>
-<button onClick={this.deleteRow.bind(this, id)}>Delete Row</button>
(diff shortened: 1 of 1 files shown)


content/docs/handling-events.md
```diff
@@ -140,7 +140,7 @@ Inside a loop, it is common to want to pass an extra parameter to an event handl
 
 ```js
 <button onClick={(e) => this.deleteRow(id, e)}>Delete Row</button>
-<button onClick={this.deleteRow.bind(this, id)}>Delete Row</button>
+<button onClick={this.deleteRow.bind(id, this)}>Delete Row</button>
```
eps1lon, line 143: This would mean that `this` points to `id` in `deleteRow` and the `id` argument would point to `this`.

That's not what this code is intended to do. Check the [documentation for `Function.prototype.bind()
`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Function/bind) for more information.

eps1lon (changes requested):

PR #4863, opened 2022-08-02
Fix event links for React Brussels 2022

AymenBenAmor:
Just fixing the event links to React Brussels conference.


eps1lon (approved):
Links check out, thanks!

PR #4864, opened 2022-08-02
Added punctuation mark to documentation.

hermitex:
This PR adds a punctuation mark to the documentation.

### Before

> Typically, in React constructors are only used for two purposes:

### After

> Typically, in React, constructors are only used for two purposes:


harish-sethuraman (commented):
Thanks for the PR. But I dont think this will add value....

eps1lon (approved):
I think this is correct. To check, replace "React constructor" with any other (keyword-like) word and the sentence is broken.

PR #4867, opened 2022-08-03
Fix Robin Pokorny’s article link

abdulrahmanhatem:
Fix Robin Pokorny’s Index as a key is an anti-pattern article link
Notes:
Link is not working neither in my PC nor my phone.
I tried to open it in incognito browser and didn't work.
I don't know the website works in all countries or not but here in Cairo, Egypt doesn't work.
And this is not the first link I encountered doesn't work


harish-sethuraman (approved):
LGTM

PR #4872, opened 2022-08-04
fixed having duplicate tailwind classes in pages/blog

Rec0iL99:
Fixes a duplication of tailwind classes found in `src/pages/blog/index.tsx`


eps1lon (approved):
Thanks!

PR #4884, opened 2022-08-07
update second block on mobile (ui)

Arhell:
resolve https://github.com/reactjs/reactjs.org/issues/4883


Arhell:
cc @eps1lon

Arhell:
before
before
after
after

PR #4885, opened 2022-08-08
fix: update all immutable js broken link

Zhhanou:
Immutable.js link https://facebook.github.io/immutable-js/ is now 404, change them to the correct address https://immutable-js.com/.


harish-sethuraman (approved):
Thanks :)

PR #4903, opened 2022-08-15
Update Refs and the DOM: Use HTML element or DOM element instead of node

rounakcodes:
Unless text nodes or other non-node HTML elements can be assigned a ref, using the word *node* leads to confusion.

### DIFF
--- a/content/docs/refs-and-the-dom.md
+++ b/content/docs/refs-and-the-dom.md
@@ -11,7 +11,7 @@ redirect_from:
   - "tips/children-undefined.html"
 ---
 
-Refs provide a way to access DOM nodes or React elements created in the render method.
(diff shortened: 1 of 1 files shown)


content/docs/refs-and-the-dom.md
```diff
@@ -11,7 +11,7 @@ redirect_from:
   - "tips/children-undefined.html"
 ---
 
-Refs provide a way to access DOM nodes or React elements created in the render method.
+Refs provide a way to access HTML elements or React elements created in the render method.
```
eps1lon, line 14: But now it's confusing since HTML elements does not include SVG elements. I'm not even sure what the "React elements" refers to, React component instances?

PR #4904, opened 2022-08-15
Update faq-ajax.md

amin-da:
change isLoaded state vlaue for Simplicity and better code readability

### DIFF
--- a/content/docs/faq-ajax.md
+++ b/content/docs/faq-ajax.md
@@ -35,7 +35,7 @@ class MyComponent extends React.Component {
     super(props);
     this.state = {
       error: null,
-      isLoaded: false,
+      isLoaded: true,
       items: []
     };
   }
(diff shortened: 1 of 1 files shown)


content/docs/faq-ajax.md
```diff
@@ -88,7 +88,7 @@ Here is the equivalent with [Hooks](https://reactjs.org/docs/hooks-intro.html):
 ```js
 function MyComponent() {
   const [error, setError] = useState(null);
-  const [isLoaded, setIsLoaded] = useState(false);
+  const [isLoaded, setIsLoaded] = useState(true);
```
eps1lon, line 91: It can't start as "is loaded" since we need to load first. The variable name should be read as a question: "is the data loaded?". Hope this clarifies why this change would worsen readability

eps1lon (changes requested):

PR #4906, opened 2022-08-16
[Beta] Fix theme toggle logic

holazz:


harish-sethuraman (commented):
We change theme when system's theme changes and user hasn't selected any theme in the beta site. that is why we have that check.

holazz:
> We change theme when system's theme changes and user hasn't selected any theme in the beta site. that is why we have that check.

All right, but once the theme toggle button is clicked, the theme can't follow the system theme change unless I manually clear the LocalStorage.

PR #4917, opened 2022-08-20
[beta] Accessibility update text colors

Arhell:
resolve https://github.com/reactjs/reactjs.org/issues/4916

### DIFF
--- a/beta/colors.js
+++ b/beta/colors.js
@@ -21,6 +21,8 @@ module.exports = {
   'border-dark': '#343A46', // gray-80
   'secondary-button': '#EBECF0', // gray-10
   'secondary-button-dark': '#404756', // gray-70
+  'text-link': '#07779c' /* Accessibility home page link color */,
+  'text-link-dark': '#2ABAEA',
 
   // Gray
   'gray-95': '#16181D',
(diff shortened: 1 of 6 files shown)


gaearon:
Can you please provide screenshots before/after for specific parts you changed? We'd also want to avoid hardcoding colors. If we need a new shade of blue, I think we should add it to the config. Thank you!

Arhell:
> Can you please provide screenshots before/after for specific parts you changed? We'd also want to avoid hardcoding colors. If we need a new shade of blue, I think we should add it to the config. Thank you!

After the update, only 1 error remained

before



after
Screenshot_1

there it is not noticeable to the eye, but the Accessibility Insights for Web application no longer produces errors

Arhell:
before
https://beta.reactjs.org/

after
https://beta-reactjs-gfgg4z7ad-fbopensource.vercel.app/

gaearon:
we keep the colors in `colors.js` which is referenced from the tailwind config. would you mind adding a proposed color there and matching the existing naming scheme? thanks!

Arhell:
> we keep the colors in `colors.js` which is referenced from the tailwind config. would you mind adding a proposed color there and matching the existing naming scheme? thanks!

Done, but locally for some reason the colors that are added do not work

Arhell:
> we keep the colors in `colors.js` which is referenced from the tailwind config. would you mind adding a proposed color there and matching the existing naming scheme? thanks!

fix & update

PR #4928, opened 2022-08-24
[beta] Added documentation for the useMemo() API reference.

arati-1:
Added documentation for the useMemo() API.

## [Preview](https://beta-reactjs-org-git-fork-arati-1-apiusememo-fbopensource.vercel.app/)


beta/src/pages/apis/react/useMemo.md
```diff
@@ -2,17 +2,328 @@
...
+import {useMemo} from 'react';
+  // ...
 ```
 
-</Intro>
+
+* `computeExpensiveValue`: The parameter is the expensive computational function that requires memoization.
+
+* `[a,b]`: This value represents the dependency on which the recompute of the `computeExpensiveValue` rests. There can be an array of dependencies or a single value.
```
rickhanlonii, line 297: Should this be `deps`? `[a,b]` is confusing because there's not always two deps.

beta/src/pages/apis/react/useMemo.md
```diff
@@ -2,17 +2,328 @@
...
+
+#### Returns {/*returns*/}
+
+`useMemo` returns a value or object depending on the computed value:
```
rickhanlonii, line 301:
```suggestion
`useMemo` returns a memoized value for the result of `computeExpensiveValue`.
```

beta/src/pages/apis/react/useMemo.md
```diff
@@ -2,17 +2,328 @@
...
+  });
+}
+
+//...
+
+```
+
+### Skip re-render of components {/*skip-re-render-of-components*/}
+
+Often you would want to skip re-render of components when not necessary.
```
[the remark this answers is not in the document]
arati-1: We can add a note, stating the recommended way to use `useMemo`.

beta/src/content/apis/react/useMemo.md
```diff
@@ -2,19 +2,342 @@
...
+  - [Skip expensive recalculations](#skip-expensive-recalculations)
+  - [Skip re-render](#skip-re-render)
+- [Reference](#reference)
+  - [`const memoizedValue = useMemo(callback, [...dependencies])`](#usememo)
+- [Troubleshooting](#troubleshooting)
+  - [Every time my component renders `useMemo` is triggered](#every-time-my-component-renders-useMemo-is-triggered)
+
+---
+
+## Usage {/*usage*/}
```
rickhanlonii, line 33: Heads up: this is handled already with <InlineToc /> above now, so it's duplicated:

Screen Shot 2022-09-06 at 1 37 13 PM

arati-1: Thanks, will remove it.

PR #4935, opened 2022-08-25
Update CONTRIBUTING.md

NShravanReddy:
Complex word "cater to" is removed and replaced for better understanding


eps1lon:

PR #4936, opened 2022-08-25
[Beta] Removing Effect Dependencies

gaearon:
- [x] Content
- [x] Recap
- [ ] Challenges

## [Preview](https://beta-reactjs-org-git-fx-deps-fbopensource.vercel.app/learn/removing-effect-dependencies)

### DIFF
--- a/beta/src/pages/learn/removing-effect-dependencies.md
+++ b/beta/src/pages/learn/removing-effect-dependencies.md
@@ -1278,6 +1278,7 @@ This only works for [pure](/learn/keeping-components-pure) functions because the
 - To remove a dependency, you need to "prove" to the linter that it's not necessary.
 - If the code in your Effect should run in response to a specific interaction, move that code to an event handler.
 - If different parts of your Effect should re-run for different reasons, split it into several Effects.
+- If you want to update some state based on the previous state, pass an updater function.
 - If you want to read the latest value without "reacting" it, extract an Event function from your Effect.
 - In JavaScript, objects and functions are considered different if they were created at different times.
 - Try to avoid object and function dependencies. Move them outside the component or inside the Effect.
--- a/beta/src/pages/learn/removing-effect-dependencies.md
+++ b/beta/src/pages/learn/removing-effect-dependencies.md
@@ -1236,7 +1236,7 @@ The logic gets a little repetitive (you read some values from an object outside 
 
 #### Calculate primitive values from functions
 
-The same approach can work functions. For example, suppose the parent component passes a function:
+The same approach can work for functions. For example, suppose the parent component passes a function:
 
 ```js {3-8}
 <ChatRoom
--- a/beta/src/pages/learn/removing-effect-dependencies.md
+++ b/beta/src/pages/learn/removing-effect-dependencies.md
@@ -1,0 +1,716 @@
+---
+title: 'Removing Effect Dependencies'
+---
+
+<Intro>
+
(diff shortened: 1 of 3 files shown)


beta/src/pages/learn/removing-effect-dependencies.md
```diff
@@ -0,0 +1,1286 @@
...
+```css
+input { display: block; margin-bottom: 20px; }
+button { margin-left: 10px; }
+```
+
+</Sandpack>
+
+This is why you could now specify an [empty (`[]`) dependency list.](/learn/lifecycle-of-reactive-effects#what-an-effect-with-empty-dependencies-means) Your Effect *really doesn't* depend on any reactive value anymore, so it *really doesn't* need to re-run when any of the component's props or state change.
+
+### To change the dependencies, change the code {/*to-change-the-dependencies-change-the-code*/}
```
rickhanlonii, line 268: I like this as a thesis, maybe even to the point of titling the page "Changing effect dependencies".

beta/src/pages/learn/removing-effect-dependencies.md
```diff
@@ -0,0 +1,1286 @@
...
+
+You might have noticed a pattern in your workflow:
+
+1. First, you **change the code** of your Effect or how your reactive values are declared.
+2. Then, you follow the linter and adjust the dependencies to **match the code you have changed.**
+3. If you're not happy with the list of dependencies, you **go back to the first step** (and change the code again).
+
+The last part is important. **If you want to change the dependencies, change the surrounding code first.** You can think of the dependency list as [a list of all the reactive values used by your Effect's code.](/learn/lifecycle-of-reactive-effects#react-verifies-that-you-specified-every-reactive-value-as-a-dependency) You don't intentionally *choose* what to put on that list. The list *describes* your code. To change the dependency list, change the code.
```
rickhanlonii, line 276: > If you want to change the dependencies, change the surrounding code first.

Maybe this could be emphasized higher, maybe even in the introduction.

gaearon: yeah i've tried that before, and in practice people don't believe it even if it's the first thing on the page. i feel like it works better if you first show a few examples of the workflow, and *then* summarize the workflow.

beta/src/pages/learn/removing-effect-dependencies.md
```diff
@@ -0,0 +1,1286 @@
...
+```
+
+However, since the Effect now uses the `city` state variable, you've had to add `city` to the list of dependencies. That, in turn, has introduced a problem. Now, whenever the user selects a different city, the Effect will re-run and call `fetchCities(country)`. As a result, you will be unnecessarily refetching the list of cities many times.
+
+**The problem with this code is that you're synchronizing two different unrelated things:**
+
+1. You want to synchronize the `cities` state to the network based on the `country` prop.
+1. You want to synchronize the `areas` state to the network based on the `city` state.
+
+Split the logic into two Effects, each of which reacts to the prop that it needs to synchronize with:
```
rickhanlonii, line 506: What if you can use `Promise.all` in a single effect? Is that preferred?

gaearon: i don't think this solves the problem in this section? the problem is that one process's dependency triggers another process's logic

gaearon:
let's get this in and i'll do challenges later

PR #4941, opened 2022-08-27
[beta] Replace ReactDOM.render with createRoot in docs

SukkaW:
So I just stumbled upon the page [Render and Commit](https://beta.reactjs.org/learn/render-and-commit#step-1-trigger-a-render) at React beta docs, and I notice that playgrounds' console has logged some warnings:

image

It turns out that those playgrounds are still using the legacy `ReactDOM.render` method. The PR updates them.


SukkaW:
Close per https://github.com/reactjs/reactjs.org/commit/681a93527644a3e0d838883a6a7c37f475af9104

PR #4946, opened 2022-08-28
Copy button for code snippets

simicoder:
issue: https://github.com/reactjs/reactjs.org/issues/4920

and hiding button text in code snippets for small screen

image


harish-sethuraman (commented):
Nice change. LGTM would like others to validate if we want to have this or not 👍🏻

PR #4953, opened 2022-09-01
Fixed typo in passing-data-deeply-with-context.md

kaushalyap:
Typo fix: "far removed" to " far above"


eps1lon:
It might be below. So removed covers all directions.

kaushalyap:
@eps1lon Oh I see, where can I help? 

Is it okay to add new challenges where there are few challenges or non? Or only typos, error fixes are approved by external contributors.

PR #4954, opened 2022-09-01
Mention about recursion in choosing-the-state-structure.md

kaushalyap:
> You can nest state as much as you like, but making it “flat” can solve numerous problems.

It can be confusing for users without proper background to understand how infinite nesting is achieved. So I suggest mentioning recursion and linking into MDN glossary might be worth it.

### DIFF
--- a/beta/src/pages/learn/choosing-the-state-structure.md
+++ b/beta/src/pages/learn/choosing-the-state-structure.md
@@ -1474,7 +1474,7 @@ button { margin: 10px; }
 
 </Sandpack>
 
-You can nest state as much as you like, but making it "flat" can solve numerous problems. It makes state easier to update, and it helps ensure you don't have duplication in different parts of a nested object.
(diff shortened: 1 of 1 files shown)


gaearon:
I’m not sure this article really helps explain how it works.

kaushalyap:
@gaearon What do you think about https://javascript.info/recursion? I referenced MDN since it is the most part of the docs refers to.

gaearon:
I feel like it’s hard to make a connection from reading about JS recursion to recursion in rendering / JSX. We could be more explicit about it but I think a link alone isn’t enough to help.

kaushalyap:
@gaearon IMHO I do not think adding something like “React components are also functions” will fit the context there, and it will also drive the reader away from the current topic at hand. I think it is best to put in a glossary page [like MDN](https://developer.mozilla.org/en-US/docs/Glossary)?

PR #4955, opened 2022-09-01
[Beta] Initial API doc for Suspense

acdlite:
This doesn't include everything but I think it's good enough for a start.

I didn't add any interactive examples yet because I want to use the `use` API, so I'll wait until that's documented.

### DIFF
--- a/beta/src/content/apis/react/Suspense.md
+++ b/beta/src/content/apis/react/Suspense.md
@@ -2,19 +2,126 @@
 title: Suspense
 ---
 
-<Wip>
-
-This section is incomplete, please see the old docs for [Suspense.](https://reactjs.org/docs/react-api.html#reactsuspense)
-
-</Wip>
-
-
 <Intro>
 
+`Suspense` is a React Component that displays a fallback until its children have finished loading.
+
+
 ```js
-<React.Suspense fallback={<Spinner />}>{...}</React.Suspense>
+<Suspense fallback={<Placeholder />}>
+  <ComponentThatLoadsData />
+</Suspense>
 ```
 
 </Intro>
 
-<InlineToc />
+- [Usage](#usage)
+  - [Displaying a fallback while data is loading](#displaying-a-fallback-while-data-is-loading)
+  - [Revealing nested content as it loads](#revealing-nested-content-as-it-loads)
+  - [Lazy loading components with Suspense](#suspense-for-code-splitting)
+  - [Fetching data with Suspense](#suspense-for-data-fetching)
+- [Reference](#reference)
+  - [Props](#props)
+- [Troubleshooting](#troubleshooting)
+  - [How do I prevent the UI from being replaced by a fallback during an update?](#preventing-unwanted-fallbacks)
+
+---
+
+## Usage {/*usage*/}
+
+### Displaying a fallback while data is loading {/*displaying-a-fallback-while-data-is-loading*/}
+
+You can wrap any part of your application with a Suspense component. If the data in <CodeStep step={2}>its children</CodeStep> hasn't loaded yet, React will switch to rendering the <CodeStep step={1}>`fallback`</CodeStep> prop instead. For example:
+
+```js [[1, 3, "<LoadingSpinner />"], [2, 4, "<Comments />"]]
+<>
+  <Post />
+  <Suspense fallback={<LoadingSpinner />}>
+    <Comments />
+  </Suspense>
+</>
+```
+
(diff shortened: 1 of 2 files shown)


gaearon:
If we're comfortable adding a dependency on launch, I feel like you could make the sandboxes use the `experimental` build and have them use it. WDYT?

beta/src/pages/apis/react/Suspense.md
```diff
@@ -2,17 +2,126 @@
...
+```js [[1, 2, "<Post />"], [2, 3, "<LoadingSpinner />"], [3, 4, "<Comments />"]]
+<>
+  <Post />
+  <Suspense fallback={<LoadingSpinner />}>
+    <Comments />
+  </Suspense>
+</>
+```
+
+Suppose that <CodeStep step={3}>`<Comments />`</CodeStep> takes longer to load than <CodeStep step={1}>`<Post />`</CodeStep>. Without a Suspense boundary, React wouldn't be able to show either component until both have loaded — <CodeStep step={1}>`<Post />`</CodeStep> would be blocked by <CodeStep step={3}>`<Comments />`</CodeStep>.
```
gaearon, line 45: I think the "code step" usage here is a bit more distracting than helpful. I'd prefer if we keep it limited and instead of highlighting pieces of JSX in code, we use it for terms like "content" and "loading indicator" and try to use at most three (ideally two) colors at the time for any given explanation. We also don't have to highlight all matches. E.g. I think it would actually be better to remove highlights here but add them in the sentence above the snippet for "its children" and "fallback prop" instead.

Ditto for sections below.

gaearon (changes requested):
let's tweak the highlight thing.

ideally we need _some_ example. e.g. at least lazy(). i think it should work in the sandbox?

use() with experimental build could also be nice. you can override which react build is used by specifying a package.json file in the sandbox. you can search markdown for examples of `package.json` file, we use this for adding Immer to some pages

beta/src/pages/apis/react/Suspense.md
```diff
@@ -2,17 +2,126 @@
...
+
+## Reference {/*reference*/}
+
+### Props {/*props*/}
+* `children`: The actual UI you intend to render. If `children` suspends while rendering, the Suspense boundary will switch to rendering `fallback`.
+* `fallback`: An alternate UI to render in place of the actual UI if it has not finished loading. Any valid React node is accepted, though in practice, a fallback is a lightweight placeholder view, such as a loading spinner or skeleton. Suspense will automatically switch to `fallback` when `children` suspends, and back to `children` when the data is ready. If `fallback` suspends while rendering, it will activate the next parent Suspense boundary.
+
+## Troubleshooting {/*troubleshooting*/}
+
+### How do I prevent the UI from being replaced by a fallback during an update? {/*preventing-unwanted-fallbacks*/}
```
sophiebits, line 118: Mention useTransition for a loading indicator? Folks may think they don't want startTransition because there's no visual feedback.

acdlite: Yeah I can mention `useTransition`. Though I don't want to get into a full explanation of transitions here. I believe the plan is to have a comprehensive guide to transitions in the "Learn" section similar to what we have for Effects. When I chatted to @gaearon about it, the idea was to sort of assume that doc already exists when writing this API doc. The other reason I didn't mention `isPending` is that a lack of a loading indicator is still preferable to UI disappearing.

gaearon: I don't think a full explanation is necessary but I think mentioning how to solve a problem is useful. Like, assume that the doc exists so you can refer to the concept.

acdlite: I guess I just disagree it’s relevant to this particular troubleshooting section (how do I prevent UI from switching to a fallback?). Seems more appropriate for the useTransition or startTransition docs, which aren’t written yet but this section can link to.

acdlite: Maybe in a different section like “How do I show loading indicator while a transition is suspended?” that can link out to the useTransition docs.

gaearon: Well, I'm thinking it's a part of *using Suspense*. Maybe this shouldn't go in Troubleshooting but in Usage instead. What I'm trying to do with Usage is to surface related APIs that solve problems with the API you're using. Cause you don't know that you need to use `useTransition` in the first place. This section helps cross-API discovery.

gaearon: I can add this if it's not super clear what I mean, but I do think it should be here

gaearon:
any appetite for adding the sandbox with lazy? i was thinking maybe add a large markdown library as a dep of a component and trigger it on click. might be enough to show the lazy loading if codesandbox actually supports dynamic imports (i think it does?)

PR #4964, opened 2022-09-02
Very important note for the main example to work correctly

ahmadbsse:
It should be added and highlighted that the `FancyButton` component has to be converted to class based implementation to be able to get the ref forwarding


gaearon:
This definitely isn't right, `forwardRef` is primarily _for_ functions, and the example works. What do you mean?

PR #4965, opened 2022-09-03
[Beta] Clarify grammar

zqran:
In `Purity: Components as formulas` section:

> * **Minds its own business.** It does not change any objects or variables that existed before it was called.

In `Recap` section:

>  * **Mind its own business.** It should not change any objects or variables that existed before rendering.

Keep the two contents consistent.


gaearon:
The change you proposed is a bit gramatically incorrect, so I did it a bit differently

zqran:
Thanks ~

PR #4966, opened 2022-09-03
fix: add css variables of header and fix layout collapses caused by headers, etc.

k4a-l:
Some css variables such as banner were defined, but header css variables were not, and layout margins were hard-coded without css variables.The layout was visually normal, but the markdown section was encroaching into the header. It was forced to look correct.

* Add css variables of header.
* Change hard-coded  to css variable (header,layout).
* Change css variables to hard-coded or javascript variables, that are no longer needed due to the above change.


gaearon:
thanks!

PR #4967, opened 2022-09-03
[Beta] Link to Getting Started guide in Parcel docs

kidonng:
Parcel docs has a dedicated page for React, it would probably be better linking to that instead.


gaearon:
I feel like that page isn't great because it assumes you already have Parcel set up.

kidonng:
That's fair. How about pointing to https://parceljs.org/getting-started/webapp/ instead, in the same vein as the link to Vite?

PR #4969, opened 2022-09-03
[Beta] add a link to using `AbortSignal`

kidonng:
I appreciate the docs is not using lots of APIs outside of React scope which may confuse the reader, but I also think in this particular example, a link to using `AbortSignal` would be beneficial. Compared to the relatively simple boolean check, using `AbortSignal` is superior as it can _cancel_ the request all together.

#1142 was contributing the same effort (to current docs, not beta), but the situation has changed a lot. This API is now [available in major browsers](https://caniuse.com/mdn-api_fetch_init_signal_parameter) for more than 4 years, almost as old as `<script type="module">`, so this is not too out of place.

### DIFF
--- a/beta/src/pages/learn/you-might-not-need-an-effect.md
+++ b/beta/src/pages/learn/you-might-not-need-an-effect.md
@@ -739,7 +739,7 @@ function SearchResults({ query }) {
 }
 ```
 
-This ensures that when your Effect fetches data, all responses except the last requested one will be ignored.
+This ensures that when your Effect fetches data, all responses except the last requested one will be ignored. Note that in this example, you can also use a [`AbortSignal`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal#aborting_a_fetch_operation_using_an_explicit_signal) instead.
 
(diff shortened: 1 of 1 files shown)


gaearon:
This is pretty subtle. I believe that a naïve usage of `AbortSignal` can, in fact, be *worse* because the only thing it helps with is the original request. You *still* need a boolean check before you set state because you could have some async operation in between (e.g. an extra `then()` call) and the original `AbortSignal` would not help there at all. One way to avoid an extra boolean is to check `signal.aborted` just before setting state but I don't think that if we just link to `AbortSignal`, people will figure it out. A deep dive on this topic might be appropriate though.

kidonng:
Thanks for the response. I chose to add a link as it doesn't seem appropriate to rewrite the existing example. I do agree a simple link is not enough, given how other sections provide concrete examples.

> A deep dive on this topic might be appropriate though.

The existing content already covers this topic very well, the difference is, as you have mentioned, pretty subtle. I feel a deep dive might become a little off-topic. We can revisit this in the future though.

PR #4973, opened 2022-09-04
Added examples to `hydrateRoot()`

adriancuadrado:


adriancuadrado:
I should consider applying this PR to the beta, as @gaearon says in this comment: https://github.com/reactjs/reactjs.org/pull/4972#issuecomment-1236240786

PR #4975, opened 2022-09-04
[Beta] Refactor Layouts

gaearon:
There's more indirection here than I'd like. It's making it difficult to see what's being passed to what. Since we don't actually need radically different layouts, I'm simplifying it to hardcode the structure. This will make it easier to consolidate our MDX preprocessing, later add MDX 2, and eventually move to Server Components.

This commit breaks the `/blog` index page. I could port it over. However, this page doesn't really make sense in the current design anyway because the titles are repeated in the sidebar. So I think for the first release we'll just make the index page redirect to the newest blog entry, just like it used to be on the old site.


gaearon:
Seems like global size went up but page sizes went down. Must be some factoring thing. I'll look again later.

PR #4978, opened 2022-09-05
Reduce sandbox button text sizes,show dropdown button by default for …

mdmathewdc:
### **Issue:** 

If a sandbox example in a page had multiple tabs, it was causing an overflow when viewed in mobile devices which was breaking the nav bar and the announcement. eg: https://beta.reactjs.org/learn/reusing-logic-with-custom-hooks, https://beta.reactjs.org/learn/scaling-up-with-reducer-and-context

https://user-images.githubusercontent.com/49617008/188349045-88d52fe2-41f7-41ee-a35a-16184ad8cee2.mp4

### **Fixes:** 
- Reduce the font size of _Download_, _Reset_ and _Open in CodeSandbox_ buttons so that the file tabs are highlighted more which results in a better user experience
- Set the file dropdown state for Sandbox files as true by default  which helps to maintain the responsiveness while the Sandbox loads

### **Commits**:
- Reduce sandbox button text sizes, show dropdown button by default for sandbox in mobile devices

### DIFF
--- a/beta/src/components/MDX/Sandpack/DownloadButton.tsx
+++ b/beta/src/components/MDX/Sandpack/DownloadButton.tsx
@@ -88,7 +88,7 @@ ${css}
 
   return (
     <button
-      className="text-sm text-primary dark:text-primary-dark inline-flex items-center hover:text-link duration-100 ease-in transition mx-1"
+      className="text-xs text-primary dark:text-primary-dark inline-flex items-center hover:text-link duration-100 ease-in transition mx-1"
       onClick={downloadHTML}
       title="Download Sandbox"
       type="button">
--- a/beta/src/components/MDX/Sandpack/FilesDropdown.tsx
+++ b/beta/src/components/MDX/Sandpack/FilesDropdown.tsx
@@ -22,9 +22,9 @@ export function FilesDropdown() {
         {({open}) => (
           <span
             className={cn(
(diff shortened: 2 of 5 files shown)


gaearon:
Is there any way we can change the logic so that it doesn't rely on an Effect at all, and instead shows the correct visual state depending on a CSS media query or something? So that it even works before HTML loads if you resize the page.

mdmathewdc:
@gaearon Using an effect in this use case has a much cleaner logic as our end goal is to toggle the dropdown for the files based on the number of file tabs and window width rather than using a media query alone. The effect is needed as we also have to handle the responsiveness when the user drags and resizes their browser window which is taken care by the resize window listener. One additional advantage is that it offers better responsiveness and doesn't result in a broken page when the page loads which is especially visible in slower connections.

mdmathewdc:
@gaearon Any other clarification needed from my end?

gaearon:
The goal is to minimize visual jumps from HTML to hydrated state. Both on mobile and desktop. I want to better understand what happens today in these cases, and how your solution changes what happens:

- Hydrating sandbox with one file (mobile)
- Hydrating sandbox with one file (desktop)
- Hydrating sandbox with many files, first is active (desktop)
- Hydrating sandbox with many files, first is active (mobile)
- Hydrating sandbox with many files, second is active (mobile)
- Hydrating sandbox with many files, second is active (desktop)

gaearon:
Thanks for the PR. I took a slightly different solution in https://github.com/reactjs/reactjs.org/pull/5014 but really appreciate pointing out the source of the issue.

PR #4981, opened 2022-09-05
[Beta] Remount Page Content

gaearon:
I broke this in https://github.com/reactjs/reactjs.org/pull/4975. Since it no longer remounts _accidentally_, the page content is reconciled between navigations. But we don't want that. So, we want a key.

### DIFF
--- a/beta/src/pages/_app.tsx
+++ b/beta/src/pages/_app.tsx
@@ -60,5 +60,9 @@ export default function MyApp({Component, pageProps}: AppProps) {
     );
   }
 
-  return <Page routeTree={routeTree}>{content}</Page>;
+  return (
+    <Page routeTree={routeTree}>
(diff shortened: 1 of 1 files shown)


beta/src/pages/_app.tsx
```diff
@@ -60,5 +60,9 @@ export default function MyApp({Component, pageProps}: AppProps) {
     );
   }
 
-  return <Page routeTree={routeTree}>{content}</Page>;
+  return (
+    <Page routeTree={routeTree}>
+      <React.Fragment key={router.pathname}>{content}</React.Fragment>
```
gaearon, line 65: This doesn't seem to show up in DevTools. Not good?

gaearon: Filed as https://github.com/facebook/react/issues/25187

PR #4982, opened 2022-09-05
[Beta] Pregenerate TOC

gaearon:
Replaces runtime generation of TOC with an early Markdown AST pass. Similar approach to https://github.com/reactjs/reactjs.org/pull/4081/. Unfortunately we can't remove the MaxWidth runtime wrapping the same way because the actual final tree structure matters, and we don't have this info at the Markdown AST time.

### DIFF
--- a/beta/plugins/rehype-toc.js
+++ b/beta/plugins/rehype-toc.js
@@ -1,0 +1,62 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ */
+
+const toHtml = require('hast-util-to-html');
+const visit = require('unist-util-visit');
+
+module.exports = function generateToc(options) {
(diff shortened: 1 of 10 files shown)


beta/plugins/rehype-toc.js
```diff
@@ -0,0 +1,62 @@
...
+    tree.children.push({
+      type: 'export',
+      value:
+        'export const toc = [' +
+        toc
+          .map(
+            (item) => `{
+          url: ${JSON.stringify(item.url)},
+          depth: ${JSON.stringify(item.depth)},
+          jsx: <>${item.html}</>
```
gaearon, line 53: This seems super dodgy. I need to think more about this. I do want to keep the inline code working, but this looks weird and also doesn't always work.

PR #4993, opened 2022-09-05
[Beta] update quote content

zqran:

### DIFF
--- a/beta/src/content/learn/conditional-rendering.md
+++ b/beta/src/content/learn/conditional-rendering.md
@@ -181,7 +181,7 @@ While this duplication isn't harmful, it could make your code harder to maintain
 
 ### Conditional (ternary) operator (`? :`) {/*conditional-ternary-operator--*/}
 
-JavaScript has a compact syntax for writing a conditional expression -- the [conditional operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator) or "ternary operator."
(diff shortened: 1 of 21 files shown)


gaearon:
So, I think it's a stylistic choice (I think there's a style in which periods occur inside quotes.. or maybe I'm making it up). But tbh I'd like to change this to be consistent with how it's usually written in other docs, i.e. plainly. There are many places where this happens so can you consolidate?

zqran:
Yeah, I'd consolidate it. The React Doc helps me a lot. Which stylistic should we use?

gaearon:
I think wikipedia gets it right with "logical quotation". https://en.wikipedia.org/wiki/Wikipedia:Manual_of_Style#Quotation_marks

So let's follow that approach here.

zqran:
Got it. I'll check the Doc content and consolidate theam.

zqran:
@gaearon 
I checked all Learn Doc and made the following changes:
1. change `."` to `".`(in most cases)
2. change `,"` to `",`

Some cases like this, keep its stylistic:
> To paraphrase Albert Einstein, "Make your state as simple as it can be—but no simpler."

---

When I checked the doc, I also found other cases of stylistic consistency, like this(The difference is whether the period is included in the link) in Installation Doc:

> `[download this HTML page](https://raw.githubusercontent.com/reactjs/reactjs.org/main/static/html/single-file-example.html).`

> `[add React with a script tag.](https://beta.reactjs.org/learn/add-react-to-a-website)`

I think we should use the first style.

gaearon:
Great, thank you!!

For links, I prefer to keep the period inside so that there isn't a color clash between a blue link and a mark right after it.

zqran:
It's a good idea! Got it. Should we keep all these links stylistic consistency?

gaearon:
Ideally yes

PR #5000, opened 2022-09-06
Updating `<script>` to `next/script` component 🛩️ 

sanjaiyan-dev:
Using `<Script />` component from `next/script` for inline colour mode script.
Which help for better streaming support and might have behind the scene optimisations.

_Sorry if I made any mistakes :(_


sanjaiyan-dev:
Converting this to draft until next js version is updated to 12.1.6 or later :)

PR #5007, opened 2022-09-07
[Beta] Change the position of the comma

zqran:


gaearon:
when punctuation is next to some range (bold, link, etc) i’d to keep it inside because it looks less jarring

zqran:
Got it. I will change it in this style.

gaearon:
great, thanks.

PR #5019, opened 2022-09-09
[Beta] Auto-expand on inline search

gaearon:
Chrome can do this with built-in `<details>` tag now.

Screenshot 2022-09-09 at 03 38 13

Also it's better semantically anyway.

### DIFF
--- a/beta/src/components/MDX/ExpandableExample.tsx
+++ b/beta/src/components/MDX/ExpandableExample.tsx
@@ -22,17 +22,29 @@ function ExpandableExample({
   excerpt,
   type,
 }: ExpandableExampleProps) {
+  const ref = React.useRef<null | HTMLDetailsElement>(null);
   const [isExpanded, setIsExpanded] = React.useState(false);
   const isDeepDive = type === 'DeepDive';
   const isExample = type === 'Example';
 
   return (
-    <div
+    <details
+      open={isExpanded}
+      onToggle={(e: any) => {
+        setIsExpanded(e.target!.open);
+      }}
+      onClick={(e) => {
+        // We toggle using a button instead of this whole area.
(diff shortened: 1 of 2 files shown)


beta/src/components/MDX/ExpandableExample.tsx
```diff
@@ -22,17 +22,29 @@ function ExpandableExample({
...
   const isExample = type === 'Example';
 
   return (
-    <div
+    <details
+      open={isExpanded}
+      onToggle={(e: any) => {
+        setIsExpanded(e.target!.open);
+      }}
+      onClick={(e) => {
```
sophiebits, line 36: Should this be on the summary? Won't this mean that any links in the children, say, won't work?

beta/src/components/MDX/ExpandableExample.tsx
```diff
@@ -22,17 +22,29 @@ function ExpandableExample({
...
   const [isExpanded, setIsExpanded] = React.useState(false);
   const isDeepDive = type === 'DeepDive';
   const isExample = type === 'Example';
 
   return (
-    <div
+    <details
+      open={isExpanded}
+      onToggle={(e: any) => {
+        setIsExpanded(e.target!.open);
```
sophiebits, line 34: (nit: shouldn't this be currentTarget for the hypothetical case of nested details tags?)

gaearon: hmm i guess so

gaearon: sophienits

gaearon:
thanks, fixed in 51ad3c6b5d49744a4ef032049d2d923cde2218e1

PR #5020, opened 2022-09-09
[Beta] Update the component element format

zqran:
There are 3 different ways to describe components in the Doc:

1. `Chat`
2. `<Chat>`
3. `<Chat />`

Which way should be used in which case? @gaearon


harish-sethuraman:
I guess we might have things like this throughout almost all pages in the doc?

zqran:
Yeah, the 3 different ways are a little confusing, especially `<Chat>` and `<Chat />`. So, I want to unify these descriptions as much as possible.

PR #5022, opened 2022-09-09
[Beta] Add periods to list items

zqran:


gaearon:
we should only do this when items in the list are complete sentences

zqran:
oh, got it. I will close this.

PR #5029, opened 2022-09-10
[Beta] Avoid grey screen on Safari iOS swipe

gaearon:
Alternative to https://github.com/reactjs/reactjs.org/pull/5026.
Let's see if this works.


gaearon:
amazing.

PR #5031, opened 2022-09-10
[Beta] Fix some typo

zqran:


gaearon:
i think this is intentional and matches the logs — "connecting..." is meant to signify it's potentially async (which in a later implementation in one of the pages, it is)

zqran:
Oh, I see. There is no period in the log output. Do we need to keep the period here?

PR #5032, opened 2022-09-10
[Beta] Fix navigation sticky failure

zqran:
In the current Doc, the sticky of the navigation bar on both sides is invalid. 
Associated with #5028


gaearon:
Oh interesting. I guess this would work too. I did https://github.com/reactjs/reactjs.org/commit/8f8bcb93e07e7d34894bc38428494c3a1e4f46d2. Not sure I feel strongly about which fix to take.

gaearon:
Hm. My fix is actually breaking other stuff.

zqran:
My fix is according to your description in #5028, it may not be considered very comprehensive.

beta/src/styles/index.css
```diff
@@ -76,6 +76,12 @@
     margin: 0;
   }
 
+  @media screen and (max-width: 768px) {
```
gaearon, line 79: should this be 1024?

zqran: yes, it should be.  I will fix it

gaearon:
Hm, sticky in sandboxes is still broken by https://github.com/reactjs/reactjs.org/pull/5033. Not sure how.

gaearon:
ok i fixed them in https://github.com/reactjs/reactjs.org/commit/acc8dcf9a3593ba070aead2154cb61fe98717231

PR #5035, opened 2022-09-10
[Beta] Remove `useEvent` link

zqran:
Remove the link to `useEvent` because it is 404 when opened.


gaearon:
we should add a stub page instead.

zqran:
Can I try this to add a stub page?

PR #5036, opened 2022-09-10
Paths on windows

yousefelgoharyx:
Generating paths was broken on Windows since Windows does not use slashes but rather backslashes.


yousefelgoharyx:
Also does anyone know why every pull request all my commits appear?

gaearon:
Maybe you haven’t reset your main branch to what’s in the repo.

PR #5044, opened 2022-09-11
[Beta]: fix typo

rohanmathur91:

### DIFF
--- a/beta/src/content/learn/preserving-and-resetting-state.md
+++ b/beta/src/content/learn/preserving-and-resetting-state.md
@@ -494,7 +494,7 @@ label {
 
 You might expect the state to reset when you tick checkbox, but it doesn't! This is because **both of these `<Counter />` tags are rendered at the same position.** React doesn't know where you place the conditions in your function. All it "sees" is the tree you return. In both cases, the `App` component returns a `<div>` with `<Counter />` as a first child. This is why React considers them as _the same_ `<Counter />`.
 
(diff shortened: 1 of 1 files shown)


gaearon:
It’s not a typo. Counter the first child of the first child (div) of the root (App).

rohanmathur91:
> It’s not a typo. Counter is the first child of the first child (div) of the root (App). Would you like to edit to this clarification?

Okay, it was a bit confusing, editing that.

rohanmathur91:
> I feel like it's clearer to just delete this paragraph? It repeats what's said above, kinda

I think this repeats the above `You can think of them as having the same “address”: the first child of the first child of the root. `

beta/src/content/learn/preserving-and-resetting-state.md
```diff
@@ -494,7 +494,7 @@ label {
 
 You might expect the state to reset when you tick checkbox, but it doesn't! This is because **both of these `<Counter />` tags are rendered at the same position.** React doesn't know where you place the conditions in your function. All it "sees" is the tree you return. In both cases, the `App` component returns a `<div>` with `<Counter />` as a first child. This is why React considers them as _the same_ `<Counter />`.
 
-You can think of them as having the same "address": the first child of the first child of the root. This is how React matches them up between the previous and next renders, regardless of how you structure your logic.
+You can think of them as having the same "address": the first child of the first child (`div`) of the root (`App`). This is how React matches them up between the previous and next renders, regardless of how you structure your logic.
```
rohanmathur91, line 497:
```suggestion
And this is how React matches them up between the previous and next renders, regardless of how you structure your logic.
```

PR #5045, opened 2022-09-11
[Beta] Fix the blank page displayed when the page navigate

zqran:
Steps to reproduce:

1. Jump to a page that Not Found. For example,
   1. Jump to page [State: A Component's Memory](https://beta.reactjs.org/learn/state-a-components-memory#challenges)
   2. Click `4. Remove unnecessary state` in challenges
   3. Click on the [why this didn’t work](https://beta.reactjs.org/learn/troubleshooting-state-updates#setting-state-does-not-update-variables) link
2. Click on the Learn or API Menu on the left bar

The page will be blank.

### DIFF
--- a/beta/src/components/SocialBanner.tsx
+++ b/beta/src/components/SocialBanner.tsx
@@ -18,7 +18,7 @@ export default function SocialBanner() {
       if (y === 0) {
         // We're trying to reset scroll.
         // If we already scrolled past the banner, consider it as y = 0.
-        const bannerHeight = ref.current!.offsetHeight; // Could be zero (e.g. mobile)
+        const bannerHeight = ref.current?.offsetHeight ?? 0; // Could be zero (e.g. mobile)
         y = Math.min(window.scrollY, bannerHeight);
       }
       return realScrollTo(x, y);


gaearon:
Why does this happen precisely?

zqran:
In development environment, when accessing a link that does not exist, it will cause the page to report an error: "Text content does not match server-rendered HTML."

Then, after closing the error prompt pop-up window, click Learn or API at this time, the `ref.current` value in SocialBanner is null.

zqran:
I'm not sure what's causing the problem, I think it might be related to hydration.

But what I'm sure about is that we should try to avoid `non-null assertions`(`!`) as it's just type-wise to keep the code from getting type errors, but if its value is null at runtime it will cause a runtime error.

harish-sethuraman:
`
Warning: Prop className did not match. Server: "text-link border-link dark:text-link-dark dark:border-link-dark font-bold inline-flex w-full items-center border-b-2 justify-center text-base leading-9 px-3 py-0.5 hover:text-link dark:hover:text-link-dark whitespace-nowrap" Client: "border-transparent inline-flex w-full items-center border-b-2 justify-center text-base leading-9 px-3 py-0.5 hover:text-link dark:hover:text-link-dark whitespace-nowrap"
`

and it comes from https://github.com/reactjs/reactjs.org/blob/c7d858947f832d1ba4e78caebc391fd964ff6de6/beta/src/components/Layout/Nav/NavLink.tsx#L17-L24

gaearon:
The actual cause here is that the component gets remounted when we navigate between different Next.js Pages (and `404` and `[...markdownPath]` are two different Pages). The `scrollTo` call happens during the commit and squeezes in after the ref is detached but before the next one is attached. Not sure I fully understand why. But let's take this fix.

PR #5051, opened 2022-09-12
[Beta] Expand useMemo documentation

gaearon:
Added:

- Deep dive ("Memoizing individual JSX nodes")
- Usage: Memoizing a function with a link to `useCallback`
- Deep dive ("Does every value need to be memoized?")
- Implied caching guarantees in the Reference
- Troubleshooting: A component re-renders every time
- Troubleshooting: Calling `useMemo` in a loop


beta/src/content/apis/react/useMemo.md
```diff
@@ -1178,7 +1277,144 @@ You can then right-click on the arrays from different re-renders in the console
...
+      {items.map(item =>
+        <Report key={item.id} item={item} />
+      )}
+    </article>
+  );
+}
+
+function Report({ item }) {
+  // ✅ Call useMemo at the top level:
+  const data = calculateReport(item);
```
sophiebits, line 1409: missing useMemo?

gaearon: fixed in main!

PR #5054, opened 2022-09-13
[Beta] Update links and add semicolons in API/react Doc

zqran:


beta/src/content/apis/react/index.md
```diff
@@ -85,7 +85,7 @@ function MyComponent() {
 Creates a context that components can provide or read.
 
 ```js
-const ThemeContext = createContext('light');
+const ThemeContext = createContext('light')
```
zqran, line 88: @gaearon 

Is the semicolon here also removed?

zqran: > intro snippet

Can I understand that as `<Intro> ... </Intro>` content?

PR #5056, opened 2022-09-13
[Beta] Update links and add fix MD syntax error in react-dom/client Doc

zqran:

### DIFF
--- a/beta/src/content/apis/react-dom/client/createRoot.md
+++ b/beta/src/content/apis/react-dom/client/createRoot.md
@@ -7,7 +7,7 @@ title: createRoot
 `createRoot` lets you create a root to display React components inside a browser DOM node.
 
 ```js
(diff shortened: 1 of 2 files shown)


beta/src/content/apis/react-dom/client/createRoot.md
```diff
@@ -7,7 +7,7 @@ title: createRoot
 `createRoot` lets you create a root to display React components inside a browser DOM node.
 
 ```js
-const root = createRoot(domNode, options?)
+const root = createRoot(domNode, options?);
```
gaearon, line 10: I think we intentionally don't use semicolons on these intro lines because they're more like definitions and aren't sypposed to be valid JS

zqran: oh, got it. thanks for your explanation.

PR #5059, opened 2022-09-14
[beta] Added memo API reference page

davidmccabe:
Adds a page for `memo`.

Not sure how to get rid of the extra commits from previously in my fork. However, they don't affect the files changes.


beta/src/content/apis/react/memo.md
```diff
@@ -2,21 +2,314 @@
...
+});
+```
+
+A React component should always render the same output if its props, state, and context haven't changed. By using `memo`, you are telling React that your component complies with this requirement, so React doesn't need to re-render as long as none of those things have changed. When you use `memo`, your component will still re-render if its own state changes or if a context that it's using changes.
+
+<Gotcha>
+
+Use `memo` only as a performance optimization, not to guarantee that a re-render
+will not happen. React may choose to re-render anyway in some situations, such as to save
+memory when a component is off-screen.
```
gaearon, line 50: I think the key idea we want to convey here is that if a component *breaks* due to a re-render, there's a bug you need to fix in the component. Usually it would be that component is impure, so we can link to `/learn/keeping-components-pure`.

beta/src/content/apis/react/memo.md
```diff
@@ -2,21 +2,314 @@
...
+        <input value={address} onChange={e => setAddress(e.target.value)} />
+      </label>
+      <Greeting name={name} />
+    </>
+  );
+}
+
+const Greeting = memo(function Greeting({name}) {
+  useEffect(() => {
+    console.log("Greeting was rendered");
```
gaearon, line 82: This is a bit misleading since it really shows the timing of effects. I think it would make sense to place the log in the body directly. It might be good to do `rendered at [...]` with a timestamp or something so that it's easier to distinguish repeated renders

beta/src/content/apis/react/memo.md
```diff
@@ -2,21 +2,314 @@
...
+## Usage {/*usage*/}
+
+### Avoiding re-rendering when props are unchanged {/*avoiding-rerendering-when-props-are-unchanged*/}
+
+React normally re-renders a component whenever its parent re-renders. With `memo`, you can create a component that React will not re-render when its parent re-renders so long as its new props are the same as the old props. Such a component is said to be *memoized*.
+
+`memo` is a function: you use it by wrapping your component in a call to `memo` and using the value that it returns in place of your original component:
+
+```
+export default memo(function Greeting({name}) {
```
gaearon, line 39: there are several common styles

```js
function Foo() { }
export default memo(Foo)
```


```js
let Foo = memo(() => { })

export default Foo
```


```js
function Foo() { }
Foo = memo(Foo)
export default Foo
```

and there also different things you can do with named ones.

should we stick with some single style consistently? which one and why?

gaearon: ideally the syntax we'd use would work consistently whether you render a memo'd component in the same file or export it

davidmccabe: The big advantage of your second example is that the memoization is co-located with the component definition and not snuck in someplace else where it's hard to see.

Does having it be an anonymous function mess up debugging?

beta/src/content/apis/react/memo.md
davidmccabe: I'm not too happy about this because I'm concerned that we're showing an example of a side-effect in render. Also, either way you get the initial logs on page load plus there are lots of renders (4) per change, which I could see could confuse the issue.

Can you think of a better way to indicate when the render is happening or not happening?

davidmccabe:
Updated from your comments.

beta/src/content/apis/react/memo.md
gaearon: Yeah it's not ideal for debugging. I guess

```js
let Foo = memo(function Foo() {

})
```

would also work even if a bit clunky

gaearon:
One thing I'm wary of is that people might take away an impression that memoizing tiny components like Button etc is very useful. Or that they should strive to always avoid re-renders. This is why I've tried to use "higher-level" examples like wrapping a form or a list. What do you think about this? I do think there's value in simplicity though.

If you read through https://beta.reactjs.org/apis/react/useCallback and https://beta.reactjs.org/apis/react/useMemo, how do you feel about the difference in examples and some other stylistic choices in the structure? Would you change anything on this page to match those better, or do you feel it would be good to change some things on those pages to match this one better? It seems like it would be nice if they felt like they were written cohesively but I don't want to push strongly into some specific direction.

beta/src/content/apis/react/memo.md
```diff
@@ -2,21 +2,300 @@
...
+#### Returns {/*returns*/}
+
+`memo` returns a new React component. It behaves the same as the component provided to `memo` except that React will not always re-render it when its parent is being re-rendered unless its props have changed.
+
+---
+
+## Troubleshooting {/*troubleshooting*/}
+### My component re-renders when a prop is an object, array, or function {/*my-component-rerenders-when-a-prop-is-an-object-or-array*/}
+
+React compares old and new props by shallow equality: that is, it considers whether each new prop is reference-equal to the old prop. If you create a new object or array each time the parent is re-rendered, even if the individual elements are each the same, React will still consider it to be changed. Similarly, if you create a new function when rendering the parent component, React will consider it to changed even if the function has the same definition.  Avoid this by [simplifying props or memoizing props in the parent component](#minimizng-props-changes).
```
harish-sethuraman, line 301:
```suggestion
React compares old and new props by shallow equality: that is, it considers whether each new prop is reference-equal to the old prop. If you create a new object or array each time the parent is re-rendered, even if the individual elements are each the same, React will still consider it to be changed. Similarly, if you create a new function when rendering the parent component, React will consider it to have changed even if the function has the same definition.  Avoid this by [simplifying props or memoizing props in the parent component](#minimizng-props-changes).
```

beta/src/content/apis/react/memo.md
davidmccabe: Could we have a transform that assigns a name to a function if it's passed to memo based on what the memo gets assigned to?

davidmccabe:
I had another look at `useMemo` and `useCallback`. I wonder how much it's possible to marry exposition and concision. I think my instincts are more for the later, and there's value in getting the facts across quickly so that they aren't lost. But it's also important to expand on them enough that they're put in context and understood. The issue of not encouraging memoizing every little component is a good example of that, which is handled well in the `useMemo` page.

The `useCallback` page I feel goes on a long buildup — focused on `memo`, a related but different feature — before saying what `useCallback` does. I think it would be better to plainly state what it does, and then immediately after that fill in the context of where it's most important to use.

I think the artificial slowdown is a better way to illustrate than the console.log that I'm doing. I'll update with that, and generally try to reconcile the styles a bit.

beta/src/content/apis/react/memo.md
gaearon: I think for practical reasons we don't consider console logs a side effect -- it's a debugging mechanism and we do expect people to log during render

beta/src/content/apis/react/memo.md
gaearon: In theory yes, but in practice we haven't made/encouraged one

gaearon:
>The useCallback page I feel goes on a long buildup — focused on memo, a related but different feature — before saying what useCallback does. I think it would be better to plainly state what it does, and then immediately after that fill in the context of where it's most important to use.

Yea I kind of did that in https://github.com/reactjs/reactjs.org/pull/5064. Is that closer to what you had in mind?

PR #5063, opened 2022-09-15
[beta] Remove 'each next' and 'every next'

davidmccabe:
I noticed a few places where "each next" or "every next" is used. This is not standard usage, as the word "next" can only apply to the single immediate item in a sequence, not every item. In some places it was clear without the word "next" at all, while in others I replaced it with "subsequent". If this is felt to be too formal then "following" or "later" or "after the first" could be used instead.

### DIFF
--- a/beta/src/content/apis/react/useCallback.md
+++ b/beta/src/content/apis/react/useCallback.md
@@ -80,7 +80,7 @@ You need to pass two things to `useCallback`:
 
 On the initial render, the <CodeStep step={3}>returned function</CodeStep> you'll get from `useCallback` will be the function you passed.
 
-On every next render, React will compare the <CodeStep step={2}>dependencies</CodeStep> with the dependencies you passed during the last render. If none of the dependencies have changed (compared with [`Object.is`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is)), `useCallback` will return the function you passed on the *last* render. Otherwise, React will return the function you passed on *this* render.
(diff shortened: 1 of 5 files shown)


beta/src/content/apis/react/useCallback.md
```diff
@@ -80,7 +80,7 @@ You need to pass two things to `useCallback`:
 
 On the initial render, the <CodeStep step={3}>returned function</CodeStep> you'll get from `useCallback` will be the function you passed.
 
-On every next render, React will compare the <CodeStep step={2}>dependencies</CodeStep> with the dependencies you passed during the last render. If none of the dependencies have changed (compared with [`Object.is`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is)), `useCallback` will return the function you passed on the *last* render. Otherwise, React will return the function you passed on *this* render.
+On every subsequent render, React will compare the <CodeStep step={2}>dependencies</CodeStep> with the dependencies you passed during the last render. If none of the dependencies have changed (compared with [`Object.is`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is)), `useCallback` will return the function you passed on the *last* render. Otherwise, React will return the function you passed on *this* render.
```
rickhanlonii, line 83: Just a though, maybe this can be worded as:

```suggestion
On every render, React will compare the <CodeStep step={2}>dependencies</CodeStep> with the dependencies you passed during the previous render. If this is the first render, or any of the dependencies have changed (compared with [`Object.is`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is)), `useCallback` will return the function you passed on *this* render. Otherwise, `useCallback` will return the function you passed on the *previous* render.
```

rickhanlonii (commented):
Makes sense to me!

davidmccabe:
Looks like my editor removes trailing whitespace, is that ok?

rickhanlonii:
Seems fine to me, does prettier not already do that?

davidmccabe:
@rickhanlonii  Nice suggestion. Please go ahead and merge or give me write access.

PR #5065, opened 2022-09-15
[Beta] Give `ImageSizeContext` an appropriate default value

liby:


gaearon:
The default value is intentionally large so that it's very noticeable it's wrong. It needs to be clear what breaks when you remove a provider.

PR #5067, opened 2022-09-15
[Beta] Fragment edits

gaearon:
Nit:

- `Fragment` rather than `React.Fragment` for consistency
- Add syntax to sidebar and title
- Avoid `Notification` as an example since you'd expect notifications to have a border around them. (Specifically, if you show a list of notifications, it doesn't make sense for them to have no wrappers. But for showing a list of posts it kinda makes sense. So it lets us demonstate the use case more convincingly.)
- Add a couple of sandboxes
- Add some wild trivia


gaearon:
cc @davidmccabe

PR #5071, opened 2022-09-16
[Beta] Fix you-might-not-need-an-effect.md challenge 4 expected output

kaushalyap:
In [you-might-not-need-an-effect](https://beta.reactjs.org/learn/you-might-not-need-an-effect) challenge 4 text

> Change the `showForm` state variable to initialize to `false` instead of `true`. As soon as you make that change, the console will show that an empty message was sent **twice**. Something in this logic is wrong!

But according to the given code empty message only appears once, since there is only one re-render and `useEffect` runs only once after the re-render.

```
export default function Form() {
  const [showForm, setShowForm] = useState(false);
  const [message, setMessage] = useState('');

  useEffect(() => {
    if (!showForm) {
      sendMessage(message);
    }
  }, [showForm, message]);

  function handleSubmit(e) {
    e.preventDefault();
    setShowForm(false);
  }

  if (!showForm) {
    return (
      <>
        <h1>Thanks for using our services!</h1>
        <button onClick={() => {
          setMessage('');
          setShowForm(true);
        }}>
          Open chat
        </button>
      </>
    );
  }

  return (
    <form onSubmit={handleSubmit}>
      <textarea
        placeholder="Message"
        value={message}
        onChange={e => setMessage(e.target.value)}
      />
      <button type="submit" disabled={message === ''}>
        Send
      </button>
    </form>
  );
}

function sendMessage(message) {
  console.log('Sending message: ' + message);
}


```

So fixed the solution text as below, telling that empty message only appear once.

> Change the `showForm` state variable to initialize to `false` instead of `true`. As soon as you make that change, the console will show that **an empty message was sent**

### DIFF
--- a/beta/src/content/learn/you-might-not-need-an-effect.md
+++ b/beta/src/content/learn/you-might-not-need-an-effect.md
@@ -1594,7 +1594,7 @@ button {
 
 This `Form` component lets you send a message to a friend. When you submit the form, the `showForm` state variable is set to `false`. This triggers an Effect calling `sendMessage(message)`, which sends the message (you can see it in the console). After the message is sent, you see a "Thank you" dialog with an "Open chat" button that lets you get back to the form.
 
-Your app's users are sending way too many messages. To make chatting a little bit more difficult, you've decided to show the "Thank you" dialog *first* rather than the form. Change the `showForm` state variable to initialize to `false` instead of `true`. As soon as you make that change, the console will show that an empty message was sent twice. Something in this logic is wrong!
+Your app's users are sending way too many messages. To make chatting a little bit more difficult, you've decided to show the "Thank you" dialog *first* rather than the form. Change the `showForm` state variable to initialize to `false` instead of `true`. As soon as you make that change, the console will show that an empty message was sent. Something in this logic is wrong!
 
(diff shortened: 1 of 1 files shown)


gaearon:
hm yeah this is confusing... it will be twice if the sandbox is force-reloaded but i'm not sure why this doesn't happen since editing initial values is supposed to at least remount the component. this might be a bug in how Fast Refresh works in CodeSandbox, not sure. also i noticed the console log gets lost if there was an error earlier which seems more concerning

gaearon:
there's something wrong with how console works with this example. it seems like if you make a syntax error and later fix it, console stops working until next file edit or something.. cc @danilowoz

kaushalyap:
@gaearon you are right it should be twice, sorry my bad, jumped to the conclusion too early. It clearly seems to be an issue with sandbox, I tried running the example locally empty message is shown twice. Actually, empty message appears twice if you fork it too, they just group the same console message like in Chrome dev tools.(see the screenshot below), but sadly the console.log on the website does not reflect it.

image


It seems to me that CodeSandbox reuse resources between multiple CodeSandbox for the same session, this explains clearing of console.

PR #5085, opened 2022-09-21
Changed word compiler to transcompiler in line 90

vinitparekh17:
For more info about transcompiler

check out https://en.wikipedia.org/wiki/Source-to-source_compiler#:~:text=A%20source%2Dto%2Dsource%20translator,or%20a%20different%20programming%20language.


gaearon:
Thanks, but this is not needed. It’s still a kind of a compiler, and we want to use familiar terminology.

PR #5094, opened 2022-09-23
[Beta] separating-events-from-effects Fixed challenge 2 typo

kaushalyap:
Corrected following challenge 2 text from separating-events-from-effects

> For example, try pressing the plus button **nine** times, and notice that the count now increases each second by ten rather than by one.

to

> For example, try pressing the plus button **ten** times, and notice that the count now increases each second by ten rather than by one.


gaearon:
It's not a typo: ten minus one equals nine. You need to press nine times.

kaushalyap:
Sorry, my bad.

PR #5097, opened 2022-09-23
[Beta] Suggest/referencing values with refs

gsong:
Is there a reason we need the input to be controlled?


gaearon:
thanks, but the Prettier changes are not needed. can you please undo them so I can see what you're changing?

gsong:
@gaearon Formatting the code according to the `.prettierrc` in the repo. The changes are in a separate commit: https://github.com/reactjs/reactjs.org/pull/5097/commits/3c127505ecf7b0738873064a246acc0ec2992596

gaearon:
sorry, I don't know how to configure prettier correctly, but we don't run it on Markdown

gsong:
@gaearon No worries, I'll configure it to ignore markdown with another PR, and update this one accordingly.

gsong:
> sorry, I don't know how to configure prettier correctly, but we don't run it on Markdown

Content markdown ignored in #5098

gsong:
> thanks, but the Prettier changes are not needed. can you please undo them so I can see what you're changing?
> 
> we don't run Prettier on Markdown

I've forced pushed up just a single commit with the proposed changes.

beta/src/content/learn/referencing-values-with-refs.md
```diff
@@ -641,7 +639,7 @@ export default function Chat() {
   return (
     <>
       <input
-        value={text}
+        defaultValue={textRef.current}
```
eps1lon, line 642: The problem here is that this example reads from a ref during render. Even if that specific prop is only relevant once, the overall pattern is still problematic. React might even warn if `defaultValue` changes. But the biggest problem is reading from a ref during render which we don't want to encourage.

eps1lon (changes requested):

PR #5099, opened 2022-09-23
[beta] feat: alternative solution for scrolling image carousel

gsong:
Image carousel alternate solution using a ref on the list itself.


beta/src/content/learn/manipulating-the-dom-with-refs.md
```diff
@@ -1078,6 +1078,102 @@ img {
...
+  const listRef = useRef(null);
+  const [index, setIndex] = useState(0);
+
+  return (
+    <>
+      <nav>
+        <button onClick={() => {
+          const nextIndex = index < catList.length - 1 ? index + 1 : 0;
+          setIndex(nextIndex);
+          listRef.current.children.item(nextIndex).scrollIntoView({
```
eps1lon, line 1099: The upside of the prior solution is that it even works if the component implementation (or a children) adds intermediate DOM nodes (e.g. separators).
The proposed, alternative implementation relies on a specific DOM structure.

Generally, if we show alternatives we should spell out what pros and cons of those are so that people can make an informed decision which to choose. Right now it's just personal preference which I feel like is not appropriate for a tutorial. Curious what @gaearon thinks about this.

@gsong  What would you say are the benefits of your implementation?

PR #5107, opened 2022-09-24
[Beta] Upgrade to MDX 2

gaearon:
MDX 2 has vastly better error messages because it uses AST instead of string manipulation.
It also has fewer confusing syntax pitfalls.

I don't expect any perf improvement because we were already skipping MDX runtime in favor of lean output.

Supersedes https://github.com/reactjs/reactjs.org/pull/4058 and https://github.com/reactjs/reactjs.org/pull/4081.

This was relatively easy to upgrade thanks to the work in https://github.com/reactjs/reactjs.org/pull/4105, https://github.com/reactjs/reactjs.org/pull/4990 and https://github.com/reactjs/reactjs.org/pull/4991.


wooorm:
Awesome!

PR #5110, opened 2022-09-25
[Beta] Fully SSR CodeBlock

gaearon:
Our CodeBlock uses the same mechanism for highlighting as regular Sandpack instance. This isn't great for two reasons:

- We have to run the highlighting parser during hydration.
- The CodeMirror "decorations" like line and inline highlight only attach after hydration, causing a visible jump.

This rewrites the logic to be 100% inline without dependency on Sandpack. Instead, we roughly do what the Sandpack implementation does under the hood. We also remove a bunch of logic we don't need (like actually initializing CodeMirror). I've reimplemented the logic for placing highlights and decorations by hand.

Ideally, this would run *only* on the server. If we had Server Components, this would be one, but since we don't yet, I'll send a follow-up PR to do some ad-hoc preprocessing so this never has to run on the client.

### DIFF
--- a/beta/src/components/MDX/CodeBlock/CodeBlock.tsx
+++ b/beta/src/components/MDX/CodeBlock/CodeBlock.tsx
@@ -3,10 +3,9 @@
  */
 
 import cn from 'classnames';
-import {
-  SandpackCodeViewer,
-  SandpackProvider,
-} from '@codesandbox/sandpack-react';
+import {highlightTree} from '@codemirror/highlight';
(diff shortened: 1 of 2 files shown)


danilowoz:
Interesting, I think we can implement this on Sandpack itself because the CodeBlock component on Sandpack is also static and read-only. The only dependency from CodeMirror on this component was the decorator, but you just decoupled the logic, which seems a very good plan. 

Hope you're fine if I use this implementation on Sandpack :D https://github.com/codesandbox/sandpack/issues/577

PR #5112, opened 2022-09-25
[Beta] Highlight code blocks at build

gaearon:
Builds on https://github.com/reactjs/reactjs.org/pull/5110.

This moves the syntax highlighting for static code blocks entirely out of the rendering path. In the future this could be a Server Component instead, but for now I'm moving it to MDX preprocessing stage.


beta/src/components/MDX/CodeBlock/index.tsx
```diff
@@ -25,12 +25,11 @@ const CodeBlock = function CodeBlock({
...
-      {/* These classes are fragile and depend on Sandpack. TODO: some better way. */}
-      <div className="sp-wrapper sp-121717251 sp-c-fVPbOs sp-c-fVPbOs-LrWkf-variant-dark">
-        <div className="sp-stack sp-c-kLppIp">
-          <div className="sp-code-editor sp-c-bNbSGz">
-            <pre className="sp-cm sp-pristine sp-javascript sp-c-jcgexo sp-c-jkvvao flex align-start">
-              <code className="sp-pre-placeholder sp-c-fWymNx grow-[2]">
+      <div className="sp-wrapper">
+        <div className="sp-stack">
+          <div className="sp-code-editor">
+            <pre className="sp-cm sp-pristine sp-javascript flex align-start">
```
harish-sethuraman, line 31: I once remember we had this class `sp-pristine` getting removed when we type something. we better not hardcode it?

gaearon: It's all static here, so this doesn't matter. There's no CodeMirror anymore

gaearon:
Hm. This blows up the payload size quite a bit.

PR #5121, opened 2022-09-27
Fix escaped Docsearch title

albertkol:
Manually un-escaping the tags from the highlighted results from DocSearch.

DocSearch seems to not have a way to prevent escaping the string.

https://github.com/reactjs/reactjs.org/issues/5116


shortcuts:
hey, would you mind providing the fix to [DocSearch](https://github.com/algolia/docsearch) instead? I've put more context on the issue in [a reply to Dan's tweet](https://twitter.com/sh0rtcts/status/1590084900450799616) :)

PR #5124, opened 2022-09-27
[beta] Sandpack: provide bare components and move styles

danilowoz:
It aims to remove the hack to avoid runtime CSS and introduces the possibility of moving all styles from Sandpack to React Docs. By doing this, we can eliminate Stitches runtime execution and add only the necessary styles instead of injecting everything. 

Revert & solves: https://github.com/reactjs/reactjs.org/pull/5122

| Server | Client |
| - | - |
| localhost_3000_test | localhost_3000_test (1) |


danilowoz:
@gaearon it's ready to go!

gaearon (changes requested):
Need to fix the code block font size regression

gaearon:
awesome, thanks!

gaearon:
this regressed a couple of things:

there's a weird space to the left of the highlight

Screenshot 2022-10-07 at 03 04 59

the focus ring is messed up

Screenshot 2022-10-07 at 03 05 06

gaearon:
fixed in https://github.com/reactjs/reactjs.org/pull/5162

PR #5129, opened 2022-09-28
Proofread Passing Props to a Component

tuukka:


beta/src/content/learn/passing-props-to-a-component.md
```diff
@@ -11,16 +11,16 @@ React components use *props* to communicate with each other. Every parent compon
...
 * How to pass some JSX to a component
 * How props change over time
 
 </YouWillLearn>
 
-## Familiar props {/*familiar-props*/}
+## Familiar-looking props {/*familiar-props*/}
 
-Props are the information that you pass to a JSX tag. For example, `className`, `src`, `alt`, `width`, and `height` are some of the props you can pass to an `<img>`:
+Props are the information that you pass to a React element. For example, `className`, `src`, `alt`, `width`, and `height` are some of the props you can pass to an `img` element:
```
gaearon, line 23: We're deemphasizing the concept of "elements" in the new docs. They're kind of fading into background a bit as an implementation detail. E.g. JSX transform doesn't even call `createElement`. We want to discourage their introspection too. So I'm not sure we'd like to rely on this terminology.

tuukka: Can I see a draft of the new terminology somewhere? As element, tag, and attribute are HTML/XML terminology, I'd expect them to be used with JSX and React as well, regardless of the implementation choices. I currently think of tags and attributes as markup syntax and of elements as component instances. Then, props (encompassing `children` and lacking `key`) are what the component instance receives. Am I far off?

beta/src/content/learn/passing-props-to-a-component.md
```diff
@@ -343,7 +343,7 @@ export function getImageUrl(person, size = 's') {
 
 </Sandpack>
 
-Try replacing the `<Avatar>` inside `<Card>` with some text to see how the `Card` component can wrap any nested content. It doesn't need to "know" what's being rendered inside of it. You will see this flexible pattern in many places.
+Try replacing the `Avatar` element inside the `Card` element with some text to see how the `Card` component can wrap any nested content. It doesn't need to "know" what's being rendered inside of it. You will see this flexible pattern in many places.
 
 You can think of a component with a `children` prop as having a "hole" that can be "filled in" by its parent components with arbitrary JSX. You will often use the `children` prop for visual wrappers: panels, grids, and so on.
```
ishwar00, line 348: New to React and going through docs to learn, came across this mistake(I believe). 
```suggestion
You can think of a component with a `children` prop as having a "hole" that can be "filled in" by its children components with arbitrary JSX. You will often use the `children` prop for visual wrappers: panels, grids, and so on.
```

PR #5131, opened 2022-09-29
Unify .js filename style to success the minification command

neko314:
I always thank you all to maintain docs or codes.
I recently started to learn React with beta docs.

## What's problem
I found the command in [gaearon](https://gist.github.com/gaearon)/[minification.md](https://gist.github.com/gaearon/42a2ffa41b8319948f9be4076286e1f3)  fails. This page is referenced from React docs.

```
$ npx terser -c -m -o like_button.min.js -- like_button.js
ERROR: ENOENT: no such file or directory, open 'like_button.js'
    at Object.openSync (node:fs:599:3)
    at Object.readFileSync (node:fs:467:35)
    at read_file (/Users/keiko/playground/hello-react/node_modules/terser/dist/bundle.min.js:30065:23)
    at /Users/keiko/playground/hello-react/node_modules/terser/dist/bundle.min.js:29872:41
    at Array.forEach (<anonymous>)
    at run_cli (/Users/keiko/playground/hello-react/node_modules/terser/dist/bundle.min.js:29871:32)
```

## The reason of this problem

The reason is why the difference of filename `like_button.js` or like-button.js.

The gist uses `like_button.js` but current [React Docs BETA](https://github.com/neko314/reactjs.org/blob/37eb77407d1375a2ca76a39b5647cafa8c891b09/beta/src/content/learn/add-react-to-a-website.md) users `like-button.js`

## What I confirmed 

With actual filename, the command success.

## What I did in this PR

In this time,  I updated filename in React docs, not gist and made this PR.

## What I want you to help

Honestly, I can't judge which is better,`like_button.js` or`like-button.js` by my self. 
I found the [style-guide by Google](https://google.github.io/styleguide/jsguide.html#file-name) on filename. It says both is OK. It makes me feel more difficult to make a choice by me alone. That means, this PR is appropriate or not. What do you think?


gaearon:
let's just use a different link, I made https://gist.github.com/gaearon/ee0201910608f15df3f8cd66aa83f98e instead

neko314:
@gaearon 
oh, that's a simple and better idea!
I'll commit to update link later.

neko314:
I remade the commit and update description. As the movie I attached to description, I could access new gist page you created from doc on my development env.

neko314:
oops！
I took a mistake, that I added extra file.
I fixed the commit and push again.

harish-sethuraman (approved):

PR #5137, opened 2022-09-29
[Beta] Patch broken fragment links from Docsearch

eps1lon:
I hope this isn't the correct fix. Mainly filed this to highlight the current issue.

Docsearch currently provides broken fragment links. For example, typing "comparing useState" in the docsearch will include https://beta.reactjs.org/learn/extracting-state-logic-into-a-reducer#comparing-usestate-and-usereducer--comparing-usestate-and-usereducer- in the search results. Note the broken fragment that is both duplicated and has an extraneous trailing `-`.

We can patch these up (this PR) but I'd hope there's something else not configured correctly.

### DIFF
--- a/beta/src/components/Search.tsx
+++ b/beta/src/components/Search.tsx
@@ -180,9 +180,32 @@ export function Search({
             transformItems={(items: any[]) => {
               return items.map((item) => {
                 const url = new URL(item.url);
+                // patch duplicated fragment links e.g. `some-heading` will be picked up by Algolia as `#some-heading-some-heading`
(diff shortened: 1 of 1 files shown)


gaearon:
i think we need to trigger a re-index. we just had a bug for about a week where the anchors were actually wrong.

gaearon:
just forced a recrawl. seems like it worked?

eps1lon:
> just forced a recrawl. seems like it worked?

Tested with the "comparing useState" example and now it works 👍🏻

PR #5146, opened 2022-10-03
Remove fixed width and justify content to end

lakshmi930:
Fixes: #1817 

Justify content to flex-end so it is lang response (handles RTL as well)


gaearon:
This seems to change the layout (I see the topbar content move), so something is wrong.

Please note that we're going to replace this site with https://beta.reactjs.org soon so we don't really plan to make changes to the old one.

lakshmi930:
Thanks @gaearon . Will close this PR then.

PR #5149, opened 2022-10-04
[Beta] Add Alternatives to createFactory API

gaearon:
Rewrite the page a bit to introduce a new section: Alternatives. We'll use this pattern for all legacy and deprecated APIs.

### DIFF
--- a/beta/src/content/apis/react/createFactory.md
+++ b/beta/src/content/apis/react/createFactory.md
@@ -4,7 +4,7 @@ title: createFactory
 
 <Deprecated>
 
-This API will be removed in a future major version of React. Use [JSX](/learn/writing-markup-with-jsx) or [`createElement`](/api/react/createElement) instead.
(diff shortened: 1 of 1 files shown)


beta/src/content/apis/react/createFactory.md
```diff
@@ -4,7 +4,7 @@ title: createFactory
 
 <Deprecated>
 
-This API will be removed in a future major version of React. Use [JSX](/learn/writing-markup-with-jsx) or [`createElement`](/api/react/createElement) instead.
+This API will be removed in a future major version of React. [See the alternatives.](#alternatives)
```
harish-sethuraman, line 7:
```suggestion
This API will be removed in a future major version of React. [See the alternatives](#alternatives).
```
nit 😄

gaearon: nah, i prefer to keep punctuation inside the links

PR #5164, opened 2022-10-07
[Beta] Sandpack bundler improvements

danilowoz:
Set of measures to improve the Sandpack bundler performance:
- [x] Cache babel assets;
- [x] Use a minimal babel config instead of babel/standalone;
- [x] Improve spinner component transition;
- [ ] ~Explorer how to decrease the number of dependency files;~ I don't it's still relevant because it only affects the very first load, and then the following loading will be cached;


beta/src/components/MDX/Sandpack/LoadingOverlay.tsx
```diff
@@ -0,0 +1,144 @@
...
+        setState((prev) => {
+          return prev === 'LOADING' ? 'PRE_FADING' : 'HIDDEN';
+        });
+      }
+    }, clientId);
+
+    return (): void => {
+      unsubscribe();
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
gaearon, line 112: pls let’s not?

beta/src/components/MDX/Sandpack/Preview.tsx
```diff
@@ -106,43 +97,51 @@ export function Preview({
...
   );
 
-  const overrideStyle = error
-    ? {
-        // Don't collapse errors
-        maxHeight: undefined,
-      }
-    : null;
-  const hideContent = !isReady || error;
+  // const hideContent = !bundlerIsReady || error;
```
gaearon, line 144: ?

danilowoz: Yeah, forgot to remove it

beta/src/components/MDX/Sandpack/LoadingOverlay.tsx
```diff
@@ -0,0 +1,143 @@
...
+        setState((prev) => {
+          return prev === 'LOADING' ? 'PRE_FADING' : 'HIDDEN';
+        });
+      }
+    }, clientId);
+
+    return (): void => {
+      unsubscribe();
+    };
+  }, [clientId, sandpack.status === 'idle']);
```
gaearon, line 112: this won’t pass the linter right? can we fix the code itself? the fix should be to make this a regular variable.

beta/src/components/MDX/Sandpack/LoadingOverlay.tsx
```diff
@@ -0,0 +1,143 @@
...
+    const unsubscribe = listen((message) => {
+      if (message.type === 'done') {
+        setState((prev) => {
+          return prev === 'LOADING' ? 'PRE_FADING' : 'HIDDEN';
+        });
+      }
+    }, clientId);
+
+    return (): void => {
+      unsubscribe();
```
gaearon, line 110: should this unset the ref you set in setup function?

beta/src/components/MDX/Sandpack/LoadingOverlay.tsx
danilowoz: It was just a lint warning, but I fixed it anyway. Thanks for pointing this out

beta/src/components/MDX/Sandpack/LoadingOverlay.tsx
danilowoz: Humm I'm not sure I got what you mean. This should only unsubscribe this listener and nothing else, so all other listeners and initial iframe ref should stay untouched.

beta/src/components/MDX/Sandpack/LoadingOverlay.tsx
gaearon: we’re moving closer to considering these equivalent in severity to compiler errors so we’d like to avoid violations on our own site. thanks for fixing!

beta/src/components/MDX/Sandpack/LoadingOverlay.tsx
gaearon: i mean that the cleanup function should usually be fully symmetric to the setup function, so i’m curious what the ref you’re setting on the first line represents, and whether that should be “undone” too.

the idea is that we’ll add a mode in the which effects get unmounted but state/refs get preserved. like unmounting that preserves state/DOM (just hiding the nodes visually). in that world, would it make sense to reset the ref to false? that’s what i mean by it being symmetric.

danilowoz: Ah got it! I removed this ref on the previous commit, as this is a piece from the legacy CodeSandbox bundler and doesn't do anything on the new bundler. 

Thanks for the explanation. So, answering your question and considering the scenario you described, it'd make sense to preserve the ref value on unmounted. Because this ref passes a variable to the bundler that stays mounted regardless of the usage of this hook.

gaearon:
Looks like the sticky behavior broke when "show more" is expanded.

Screenshot 2022-10-16 at 15 05 43

Also, compiler overlay seems a bit borked:

Screenshot 2022-10-16 at 15 06 07

gaearon:
line numbers are gone!

Screenshot 2022-10-16 at 15 43 11

gaearon:
we need a screenshot regression suite or something. :P

beta/src/components/MDX/Sandpack/LoadingOverlay.tsx
```diff
@@ -0,0 +1,144 @@
...
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [clientId, sandpack.status === 'idle']);
+
+  /**
+   * Fading transient state
+   */
+  useEffect(() => {
+    let fadeTimeout: NodeJS.Timer;
```
shanpriyan, line 117:
```suggestion
    let fadeTimeout: ReturnType<typeof setTimeout>;
```

danilowoz:
@gaearon, did it happen on a specific page? I can't reproduce it, even on different devices
Screenshot 2022-10-16 at 21 38 29

danilowoz:
Actually, there is something wrong with this page because these changes didn't introduce it, but it's also happening in production 

https://beta.reactjs.org/learn/thinking-in-react#step-1-break-the-ui-into-a-component-hierarchy

gaearon:
Noticed there's a jarring flash of white rounded rect on every navigation. Can we make sure the initial state looks clean?

https://user-images.githubusercontent.com/810438/196060853-e8190a59-fcde-4a81-bb0f-8c32d9a60de6.mov

Another thing I'm wondering about is what is it doing for over a second after each navigation if both my cache is on, and there's no throttling. Is there something else taking time?

gaearon:
Hmm. The way I was imagining it is that the entire white pill should be hidden until the first time the bundler completes. The user should never see an empty white pill. The white pill itself signifies completion. It should be fading it, not the content. If we can’t show the white pill *with* the content fast enough then we do need to show the spinner. We should never be showing an empty white pill.

danilowoz:
Yeah, it looks like a good idea. However, the problem is always the images because they take a bit to load, causing another resize on the white pill. I wish we could predict or set an initial body height (via sandbox style.css?) for those sandboxes that contain images and appear on the first fold, but it's not ideal either. What do you think? 

https://user-images.githubusercontent.com/4838076/196558827-17470727-bafd-4f2e-9842-c828d76a6d6d.mov

gaearon:
Hm. Why does it shift layout when images have hardcoded heights in JSX?

danilowoz:
I was investigating why this layout shift happens even with images with fixed height, and then I realized the bundler was sending the resize event too early. It didn't use to be a problem because it used to have the spinner component over everything, but now without it, it's kind of noticeable. 

The event used to be sent in between the evaluation step. From now on, it's sent after everything (bundling and evaluation), which might add some extra ms to the white pill be done.

gaearon:
(And btw, this looks _great._ Massive kudos for this work.)

gaearon:
Would be nice to reduce the flashing when runtime erroring, if possible.

https://user-images.githubusercontent.com/810438/196790930-8b29f862-d181-4404-8ac2-fd1ab21fac0f.mov

danilowoz:
@gaearon I think we're ready to go 😄

gaearon:
let's give this a try! thank you ❤️

PR #5166, opened 2022-10-07
[beta] fix: initial render in order to reflect initial state properly

gsong:
The initial state isn't rendered, so we need to update the DOM to reflect that.


eps1lon (changes requested):
Can't seem to spot what changed:
[Before](https://beta.reactjs.or
Screenshot from 2022-10-08 10-58-50
g/learn/reacting-to-input-with-state#challenges):



[Your PR](https://beta-reactjs-l7orld5sr-fbopensource.vercel.app/learn/reacting-to-input-with-state#challenges):

Screenshot from 2022-10-08 10-59-11

Similar, prior sandboxes also don't call `updateDOM` initially. Could you point to the specific part of the page that changed? Ideally with screenshots or recording if necessary.

PR #5173, opened 2022-10-09
[Beta] Add Object.is Pitfall to custom hooks

alvarlagerlof:
Adding a Pitfall section under `Passing reactive values between Hooks` exploring what happens if an object that is re-created on each render is used as a `useEffect` dependency.


alvarlagerlof:
Will update this to the Pitfall component

alvarlagerlof:
Initially I was going to write more in depth in the pitfall I added, but I noticed there was already another section about the same thing on the useEffect api ref. That page explains it, but it does not do it in the same context as the guide. I don't know if it's clear enough. Additionally there may be an opportunity to make some kind of page or section explaining `Object.js `next to how useState, useMemo, useRef... keeps their data around between renders.

PR #5174, opened 2022-10-11
Offer double negation as a solution to the problem with falsy expressions

ADTC:


content/docs/conditional-rendering.md
```diff
@@ -160,6 +160,8 @@ render() {
 }
 ```
 
+To avoid the problem, you can use double negation: `{!!count && <h1>Messages: {count}</h1>}`. This will return `<div></div>`, because double negation always returns the boolean `false` for falsy expressions.
```
harish-sethuraman, line 163: We can use !! and && for conditionally rendering but rather than doing these two operations we can rather go ahead with ternary. I once found this article to be nice which talks about `&&` and `ternary` https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx#conclusion

ADTC: Hmm, yeah. I guess it's less readable for those who aren't well-versed in Javascript. Do you suggest **not** merging? Or a simpler suggestion?

PR #5170, opened 2022-10-09
[Beta] fix: gutter overlap with sandpack code on mobile

shanpriyan:
Fix #5169 

**After fix:**

https://user-images.githubusercontent.com/42532987/194761342-9fda6702-a0cd-43ec-9b74-ac1219bc2a46.mp4

PR #5176, opened 2022-10-11
[beta] Sandpack: fix UI regression on the editor on horizontal scroll

danilowoz:
For some reason, we removed the background color from the CodeMirror gutter, which caused a bug where the line numbers overlapped the code:
Screenshot 2022-10-11 at 21 28 14

----

Edit: @gaearon, I could see you introduced this change on https://github.com/reactjs/reactjs.org/pull/5162, do you remember what you were trying to fix?

### DIFF
--- a/beta/src/styles/sandpack.css
+++ b/beta/src/styles/sandpack.css
@@ -199,6 +199,10 @@ html.dark .sp-wrapper {
   background: var(--sp-colors-surface1);
 }
 
+.sandpack .sp-code-editor .cm-editor .cm-gutters {
+  margin-left: 1px;
+}
+
 .sandpack .sp-code-editor .cm-editor {
   background-color: transparent;
 }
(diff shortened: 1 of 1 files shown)


gaearon:
the focus ring was getting obscured. does this regress on focus ring? i’m on mobile so can’t check.

harish-sethuraman:
Seems like the focus ring is still broken? 
Screenshot 2022-10-12 at 1 44 34 PM

danilowoz:
Oh, I can see now why you removed the background from the gutter element. It's a bit challenging to fix both at the same time, I'll need to spend more time to come up with a fix. 

Screenshot 2022-10-12 at 09 19 14

gaearon:
Can we make the ring an entirely separate element and give it a high z-index?

harish-sethuraman:
Why is the focus ring a box shadow and not outline 😅

gaearon:
From what I remember, the native outline was being obscured by something so it wasn't fully visible anyway

harish-sethuraman:
I tried setting outline offset to be negative 4px and outline to be 4px and it worked fine. We can consider using outline instead of box-shadow

harish-sethuraman:
I guess this is fixed with #5170

PR #5180, opened 2022-10-13
[Beta] fix navigation to the blog page of the React docs

### LINKED ISSUE #5125: [Beta] Issue in the navigation of the React Docs Page
There is an issue with the navigation buttons in the footer as while navigating between the pages in the `DocsPageFooter` with `Next` and `Previous` Button, I found a broken link for the `Blog` in the navigation of [Meet the team](https://beta.reactjs.org/community/meet-the-team) page

image

jnadroj:
A validation is made in the sending of the url to the component FooterLink. This validation is in charge of corroborating if the url is external, in case if it is, it sends the url in a normal way and if not, it applies a function that removes the final part of a url with parameters.

Also the validation is made inside FooterLink in case the url is external it applies a target to the element <a>.

Closes #5125

### DIFF
--- a/beta/src/components/DocsFooter.tsx
+++ b/beta/src/components/DocsFooter.tsx
@@ -38,12 +38,15 @@ export const DocsPageFooter = memo<DocsPageFooterProps>(
               ) : (
                 <div />
               )}
-
               {nextRoute?.path ? (
                 <FooterLink
                   type="Next"
                   title={nextRoute.title}
-                  href={removeFromLast(nextRoute.path, '.')}
+                  href={
+                    nextRoute.path.startsWith('https://')
+                      ? nextRoute.path
+                      : removeFromLast(nextRoute.path, '.')
+                  }
                 />
               ) : (
                 <div />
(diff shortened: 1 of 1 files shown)


beta/src/components/DocsFooter.tsx
```diff
@@ -66,9 +69,15 @@ function FooterLink({
...
 }) {
+  let target = '';
+  if (href.startsWith('https://')) {
+    target = '_blank';
+  }
+
   return (
     <NextLink href={href}>
       <a
+        target={target}
```
harish-sethuraman, line 80: Shouldn't target be added only when we have href to be url? if else we wont need target attribute at all. Lets not have target to be empty string.
Screenshot 2022-10-13 at 3 31 14 PM

jnadroj: I have corrected the target problem in the FooterLink component and also in the SidebarLink component that also had the same error. 

image

beta/src/components/Layout/Sidebar/SidebarLink.tsx
```diff
@@ -48,7 +48,7 @@ export function SidebarLink({
     }
   }, [ref, selected]);
 
-  let target = '';
+  let target = undefined;
```
shanpriyan, line 51:
```suggestion
  let target;
```

beta/src/components/DocsFooter.tsx
```diff
@@ -38,12 +38,15 @@ export const DocsPageFooter = memo<DocsPageFooterProps>(
...
-
               {nextRoute?.path ? (
                 <FooterLink
                   type="Next"
                   title={nextRoute.title}
-                  href={removeFromLast(nextRoute.path, '.')}
+                  href={
+                    nextRoute.path.startsWith('https://')
+                      ? nextRoute.path
+                      : removeFromLast(nextRoute.path, '.')
```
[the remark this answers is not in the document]
jnadroj: I have also added the validation for the previous button

harish-sethuraman (approved):
LGTM thanks

shanpriyan (approved):

PR #5183, opened 2022-10-17
[beta] Show `AuthorCredit` on hover and focus

mayank99:
Reused the `AuthorCredit` component that was already there, except added @rachelnabors as default values. Now uses `sr-only` class by default and `not-sr-only` when illustration is hovered or focused.

### Pending (?)

- Haven't gone through every illustration to validate which ones should be credited to who.
- Havne't touched the `Diagram` component (yet)

### Screenshots

demo of the feature showing text appearing on hover and focus

demo of the feature showing text appearing when tapping on mobile

### DIFF
--- a/beta/src/components/MDX/MDXComponents.tsx
+++ b/beta/src/components/MDX/MDXComponents.tsx
@@ -181,7 +181,7 @@ function AuthorCredit({
 }) {
   return (
     <div className="sr-only group-hover:not-sr-only group-focus-within:not-sr-only hover:sr-only">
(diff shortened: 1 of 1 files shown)


gaearon:
nice, thank you for kicking this off!

a few suggestions:

- let's move it to the top of the image rather than the bottom
- can you style it a bit more like a tooltip pill? i'll ask @lebo to actually "tailwind it up" so it's nbd if you'd rather not

mayank99:
@gaearon I've moved it to the top, but I need to think a bit more about the tooltip suggestion. For a11y, tooltips typically require the "trigger" element to be focusable. We could make the illustrations focusable, but then focusable elements require a role and an accessible name. 🤔

gaearon:
merged via https://github.com/reactjs/reactjs.org/pull/5188. thank you!

PR #5184, opened 2022-10-18
[docs-beta] fix: broken link for render-and-commt in keeping-components-pure

jukrb0x:
> You could think of your components as recipes: if you follow them and don’t introduce new ingredients during the cooking process, you will get the same dish every time. That “dish” is the JSX that the component serves to React to [render.](https://beta.reactjs.org/render-and-commit)

This PR fixes that in the article _Keeping Components Pure_, the link referring to article _Render and Commit_ missed out the `/learn/` which will lead to a 404 page.


harish-sethuraman (approved):

harish-sethuraman:
Thanks :)

PR #5187, opened 2022-10-18
[beta] Fix 'queueing a series of state changes->updates' typo in beta docs

melissapthai:
In the [Queueing a Series of State Changes](https://beta.reactjs.org/learn/adding-interactivity#queueing-a-series-of-state-changes) section in the beta docs, when you click the "Read More" button, it leads to a broken link (see screenshot).

Before:
image

This is happening because the page is actually called "Queueing a Series of State _Updates_" instead of "... State _Changes_". This fix updates these instances, and fixes the broken link.

After:
image
image


gaearon:
thank you!

PR #5190, opened 2022-10-19
[beta] broken links in updating-objects-in-state

jukrb0x:
This PR fixes the broken link referring to https://beta.reactjs.org/apis/react/memo

> Optimizations: Common React [optimization strategies](https://beta.reactjs.org/api/react/memo) rely on skipping ...


jukrb0x:
maybe there are some ci bot that can help to check the internal links in the scope of react docs, just thoughts.

gaearon:
if you have ideas on how to do that, would love to see! i don’t know if something like this exists but i’m sure it would be valuable

jukrb0x:
> if you have ideas on how to do that, would love to see! i don’t know if something like this exists but i’m sure it would be valuable

This may help with, https://github.com/tcort/markdown-link-check

gaearon:
PRs welcome :)

PR #5191, opened 2022-10-19
[beta] Fix typo on apis/react-dom/render

eps1lon:


kassens (approved):
👍

gaearon:
thanks! we do leave punctuation inside so that the color doesn't jump though i don't feel super strong about it

eps1lon:
> thanks! we do leave punctuation inside so that the color doesn't jump though i don't feel super strong about it

Gotcha. I always left it out but I couldn't tell you why 😆

PR #5193, opened 2022-10-19
[beta] Port testing docs with React 18 in mind

eps1lon:
Copied recipes and environment from the old docs. recipes got a polish with React 18 in mind.

- [Testing overview](/learn/testing)
- [Testing Environments](/learn/testing-recipes)
- [Testing Recipes](/learn/testing-environments)

Links to test-utils point to old docs for now though I tried to remove them where it made sense.

TODO:
- [ ] check links
- [ ] Read though it


gaearon:
can you spruce up the formatting a bit to match other pages? e.g. Intro, What you'll learn, etc

gaearon:
maybe "Testing Components"?

sophiebits:
Is big-picture feedback on content helpful or are you just trying to get something out?

eps1lon:
> Is big-picture feedback on content helpful or are you just trying to get something out?

The initial motivation was just to get it updated for React 18. Neither I nor Dan had special ideas. So it would be super helpful if you have feedback for the big picture!

sophiebits:
I guess there are two audiences:
* Folks looking for prescriptive guidance about what to do (mostly beginners)
* Folks who are familiar with testing in general (mocking, etc) who want to know about any React-specific things

Basically my feedback is I don't think this serves either audience very well.
* For the latter use case, I think it spends too much time discussing things that are not specific to React. Advice like "here is a basic pointer on how to mock out fetch or timers" as written is probably (a) overkill for people who already know that they want that (and there is a good chance those folks are better served by the docs for jest/etc itself) and (b) not detailed enough for folks who don't.
* For folks who feel they "should have tests" but don't know what to do, I think we could be more prescriptive and also present a more scaffolded tutorial. For this we should just lean heavily into a specific solution (e2e tests, RTL, etc) and try to help people understand _what_ to test and why, not just how to. From what I've seen, most people without experience who think they need frontend tests try to go overboard and write a lot of hard-to-maintain tests that are low quality.

gaearon:
FWIW I'd be ok with simply not porting these versions and instead I'd love to see a more focused tutorial like Sophie is describing. _Not sure you were signing up for that..._

eps1lon:
Yeah that wasn't really what we talked about. I just open a PR that fixes the existing docs instead so that we at least don't document the wrong thing.

sophiebits:
Sorry, not trying to con you into more/different work than you signed up for!

gaearon:
Well, we’d need to make a decision about what to go with for the launch. One is porting existing docs like you did. At least for the first time being. Another option is to not port this content at all. And maybe later to add something separate. But then new docs won’t have anything on testing. What do you think would be best?

PR #5195, opened 2022-10-19
Add key prop to list item

saurabhp75:
Add key prop to list item as per React's best practices.

### DIFF
--- a/beta/src/content/learn/keeping-components-pure.md
+++ b/beta/src/content/learn/keeping-components-pure.md
@@ -688,7 +688,7 @@ export default function StoryTray({ stories }) {
           {story.label}
         </li>
       ))}
-      <li>Create Story</li>
+      <li key={'create'}>Create Story</li>
     </ul>
   );
 }


beta/src/content/learn/keeping-components-pure.md
```diff
@@ -688,7 +688,7 @@ export default function StoryTray({ stories }) {
           {story.label}
         </li>
       ))}
-      <li>Create Story</li>
+      <li key={'create'}>Create Story</li>
```
harish-sethuraman, line 691: We will need key only when we use map or other loop to render elements dynamically. I see that this `li` is outside the map so we won't need a key here.

gaearon: that’s correct

saurabhp75: I will close the PR then, thanx.

saurabhp75:
Close the PR as key prop is not required for manually added individual list item.

PR #5196, opened 2022-10-20
[Beta] React.Component API

gaearon:
## [Rendered](https://beta-reactjs-org-git-component-api-fbopensource.vercel.app/apis/react/Component)

- [x] Usage
- [x] Alternatives
- [x] Reference

### DIFF
--- a/beta/src/content/apis/react/Component.md
+++ b/beta/src/content/apis/react/Component.md
@@ -104,7 +104,7 @@ export default class Counter extends Component {
   handleAgeChange = () => {
     this.setState({
       age: this.state.age + 1 
-    })
+    });
   };
 
   render() {
@@ -230,7 +230,7 @@ export default class ChatRoom extends Component {
             onChange={e => {
               this.setState({
                 serverUrl: e.target.value
-              })
+              });
             }}
           />
         </label>
@@ -358,7 +358,7 @@ export default class Counter extends Component {
   handleAgeChange = (e) => {
     this.setState({
       age: this.state.age + 1 
-    })
+    });
   };
 
   render() {
@@ -542,7 +542,7 @@ export default class ChatRoom extends Component {
             onChange={e => {
               this.setState({
                 serverUrl: e.target.value
-              })
+              });
             }}
           />
         </label>
@@ -973,7 +973,7 @@ class Counter extends Component {
   handleAgeChange = () => {
     this.setState({
       age: this.state.age + 1 
-    })
+    });
   };
 
   render() {
@@ -988,8 +988,6 @@ class Counter extends Component {
   }
 }
 ```
-
-[See more examples.](#defining-a-class-component)
 
 <Note>
 
@@ -1151,6 +1149,8 @@ TODO
 
 This API has been renamed from `componentWillMount` to [`UNSAFE_componentWillMount`.](#unsafe_componentwillmount) The old name has been deprecated. In a future major version of React, only the new name will work.
 
+Run the [`rename-unsafe-lifecycles` codemod](https://github.com/reactjs/react-codemod#rename-unsafe-lifecycles) to automatically update your components.
+
 </Deprecated>
 
 ---
@@ -1161,6 +1161,8 @@ This API has been renamed from `componentWillMount` to [`UNSAFE_componentWillMou
 
 This API has been renamed from `componentWillReceiveProps` to [`UNSAFE_componentWillReceiveProps`.](#unsafe_componentwillreceiveProps) The old name has been deprecated. In a future major version of React, only the new name will work.
 
+Run the [`rename-unsafe-lifecycles` codemod](https://github.com/reactjs/react-codemod#rename-unsafe-lifecycles) to automatically update your components.
+
 </Deprecated>
 
 ---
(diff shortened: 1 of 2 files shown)


beta/src/content/apis/react/Component.md
```diff
@@ -1,28 +1,725 @@
...
+
+</Pitfall>
+
+---
+
+### Adding lifecycle methods to a class component {/*adding-lifecycle-methods-to-a-class-component*/}
+
+There are a few special methods you can define on your class.
+
+If you define the [`componentDidMount`](#componentdidmount) method, React will call it when your component is first added *(mounted)* to the screen. React will call [`componentDidUpdate`](#componentdidupdate) after your component re-renders due to changed props or state. React will call [`componentWillUnmount`](#componentwillunmount) after your component has been removed *(unmounted)* from the screen.
```
checkmatez, line 143: > React will call componentWillUnmount after your component has been removed

Shouldn't this be:
React will call componentWillUnmount before your component has been removed
?

gaearon: Hm I guess you're right.

beta/src/content/apis/react/Component.md
```diff
@@ -1,28 +1,1713 @@
...
+
+Run the [`rename-unsafe-lifecycles` codemod](https://github.com/reactjs/react-codemod#rename-unsafe-lifecycles) to automatically update your components.
+
+</Deprecated>
+
+---
+
+### `componentWillUnmount()` {/*componentwillunmount*/}
+
+If you define the `componentWillUnmount` method, React will call it when your component is removed *(unmounted)* from the screen. This is a common place to cancel data fetching or remove subscriptions.
```
shanpriyan, line 1298: Can we ?
```suggestion
If you define the `componentWillUnmount` method, React will call it before your component is removed *(unmounted)* from the screen. This is a common place to cancel data fetching or remove subscriptions.
```

beta/src/content/apis/react/Component.md
```diff
@@ -1,28 +1,1797 @@
...
+class Greeting extends Component {
+  render() {
+    return <h1>Hello, {this.props.name}!</h1>;
+  }
+}
+```
+
+React will call your [`render`](#render) method whenever it needs to figure out what to display on the screen. Usually, you will return some [JSX](/learn/writing-markup-with-jsx) from it. Your `render` method should be a [pure function:](https://en.wikipedia.org/wiki/Pure_function) it should only calculate the JSX.
+
+Similarly to [function components,](/learn/your-first-component#defining-a-component) a class component can [receive information by props](/learn/your-first-component#defining-a-component) from its parent component. However, the syntax for reading props is different. For example, if the parent component renders `<Greeting name="Taylor" />`, then you can read the `name` prop from [`this.props`](#props), like `this.props.name`:
```
harish-sethuraman, line 47:
```suggestion
Similar to [function components,](/learn/your-first-component#defining-a-component) a class component can [receive information by props](/learn/your-first-component#defining-a-component) from its parent component. However, the syntax for reading props is different. For example, if the parent component renders `<Greeting name="Taylor" />`, then you can read the `name` prop from [`this.props`](#props), like `this.props.name`:
```

beta/src/content/apis/react/Component.md
```diff
@@ -1,28 +1,1797 @@
...
+}
+```
+
+</Sandpack>
+
+Note that Hooks (functions starting with `use`, like [`useState`](/apis/react/useState)) are not supported inside class components.
+
+<Pitfall>
+
+We recommend to define components as functions instead of classes. [See how to migrate.](#migrating-a-simple-component-from-a-class-to-a-function)
```
harish-sethuraman, line 77: Screenshot 2022-11-15 at 11 31 35 PM
This text is almost 5 times in this page. Should we impose it to this extent? I feel we could do it once on top alone?

beta/src/content/apis/react/Component.md
```diff
@@ -1,28 +1,1797 @@
...
+      <Greeting name="Cahal" />
+      <Greeting name="Edite" />
+    </>
+  );
+}
+```
+
+</Sandpack>
+
+Note that Hooks (functions starting with `use`, like [`useState`](/apis/react/useState)) are not supported inside class components.
```
harish-sethuraman, line 73: Should we wrap this text inside `<Note>`?
```suggestion
<Note>
Hooks (functions starting with `use`, like [`useState`](/apis/react/useState)) are not supported inside class components.
</Note>
```

beta/src/content/apis/react/Component.md
gaearon: no, I want it to be in-your-face

beta/src/content/apis/react/Component.md
gaearon: i believe similarly is correct here

beta/src/content/apis/react/Component.md
gaearon: meh, not necessary. want to avoid too many boxes one after another

beta/src/content/apis/react/Component.md
harish-sethuraman: 🖖🏻

PR #5197, opened 2022-10-20
Make the usage of the `props` parameter explicit

jmacqueen:
As the expected audience for this document are first-time users of React, it is more clear what is happening if you don't destructure in the function definition the first time you demonstrate passing `props` into a component.

### DIFF
--- a/beta/src/content/learn/index.md
+++ b/beta/src/content/learn/index.md
@@ -479,10 +479,10 @@ The information you pass down like this is called _props_. Now the `MyApp` compo
 Finally, change `MyButton` to *read* the props you have passed from its parent component:
 
 ```js {1,3}
-function MyButton({ count, onClick }) {
+function MyButton(props) {
   return (
-    <button onClick={onClick}>
-      Clicked {count} times
+    <button onClick={props.onClick}>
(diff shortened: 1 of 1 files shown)


gaearon:
I disagree. We consistently use destructuring throughout the docs. Why does the intro page need to be different? I'm open to something more subtle, but I don't want the introductory page to use different syntax from the entirety of the docs.

jmacqueen:
I thought I saw other places where `props` was used as a parameter in the function definition. Maybe that needs an audit? (e.g. [the definition of Clock on this page](https://beta.reactjs.org/learn/choosing-the-state-structure) )

Regardless, based on my time coaching junior devs over the years, I think it will be clearer to more people if the initial exposure to passing `props` clearly marks the provenance of the variable in use (even if it's just the once) and then maybe add some text to the effect of "it's a common pattern to destructure `props` in the function definition"

To be fair, I personally don't favor destructuring in the definition for larger components. Having `props.` in the code can help a lot in the debugging process. But I recognize that this is very much a style preference that could be debated at length and I promise this thought is not about pushing that particular agenda!

PR #5202, opened 2022-10-23
Display ctrl key in search bar for non-mac browsers

jirkavrba:
In the new documentation, there is an option key displayed for all users in the search bar.
This PR provides a check for the navigator platform and changes the displayed key to `Ctrl` when displayed on Windows or Linux.

**Search bar displayed on Mac**
image

**Search bar displayed on Windows/Linux**
image

### DIFF
--- a/beta/src/components/Search.tsx
+++ b/beta/src/components/Search.tsx
@@ -29,12 +29,13 @@ function Hit({hit, children}: any) {
 }
 
 function Kbd(props: {children?: React.ReactNode; wide?: boolean}) {
-  const width = props.wide ? 'w-12' : 'w-6';
+  const {wide, ...rest} = props;
+  const width = wide ? 'w-12' : 'w-6';
 
   return (
     <kbd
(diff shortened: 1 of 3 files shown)


arafatiqbal:
I like this, nice!

gaearon (changes requested):
This doesn't work for initially rendered HTML. Can you instead make it so that we know the platform in advance? E.g. by detecting it in an inline script that runs before the rest of the page, and setting an appropriate HTML class name on the document.

jirkavrba:
I've added the platform detection via an inline script, but I had some troubles with integrating it within the component (as `document` is not available during the initial render)

Can anybody please point me at how that reading from document could be accomplished during the initial render?

jirkavrba:
Thanks a lot, I'll implement it this way

jirkavrba:
Because the search bar component has changed quite a lot since I opened this PR I will close it and I will open a new PR with changes to the current version.

The original issue that this PR is trying to solve seems to be still present on the website.

PR #5211, opened 2022-10-27
fix #5210

### LINKED ISSUE #5210: [Beta] Make the description of Reference in memo more meaningful
I feel it would be more meaningful, if the word `parent` was included at the marked place in the image.
So it would be
```
This memoized component will usually not be re-rendered when 

its parent component is re-rendered as long as its props have not changed.
```

Screenshot 2022-10-27 at 11 56 43 AM

kavinkuma6:
Closes #5210


hexiaokang (approved):

eps1lon (approved):
make sense, thanks!

PR #5220, opened 2022-10-29
[Beta] Add missing dependencies

holazz:
Before

before

After

after

### DIFF
--- a/beta/src/content/learn/separating-events-from-effects.md
+++ b/beta/src/content/learn/separating-events-from-effects.md
@@ -796,6 +796,22 @@ With `useEvent`, there is no need to "lie" to the linter, and the code works as 
 
 <Sandpack>
 
+```json package.json hidden
+{
+  "dependencies": {
+    "react": "experimental",
+    "react-dom": "experimental",
+    "react-scripts": "latest"
+  },
+  "scripts": {
+    "start": "react-scripts start",
+    "build": "react-scripts build",
+    "test": "react-scripts test --env=jsdom",
+    "eject": "react-scripts eject"
+  }
+}
+```
+
 ```js
 import { useState, useEffect } from 'react';
 import { experimental_useEvent as useEvent } from 'react';


beta/src/content/learn/separating-events-from-effects.md
```diff
@@ -796,6 +796,22 @@ With `useEvent`, there is no need to "lie" to the linter, and the code works as
 
 <Sandpack>
 
+```json package.json hidden
+{
+  "dependencies": {
+    "react": "experimental",
+    "react-dom": "experimental",
```
eps1lon, line 803: These should probably use exact versions. Are other Sandpacks using these unstable dist tags as well?

beta/src/content/learn/separating-events-from-effects.md
```diff
@@ -796,6 +796,22 @@ With `useEvent`, there is no need to "lie" to the linter, and the code works as
...
+    "react-scripts": "latest"
```
eps1lon, line 804: This should be at least pinned to SemVer major e.g. `^4.0.0`. Otherwise this Sandpack may break once a new major version is released.

eps1lon (changes requested):

beta/src/content/learn/separating-events-from-effects.md
holazz: Yes, I just did it the same way as other Sandpacks.

eps1lon: Gotcha. Then this can stay and we'll discuss usage of this separately.

beta/src/content/learn/separating-events-from-effects.md
eps1lon: `latest` is used in other sandpacks as well?

holazz: > `latest` is used in other sandpacks as well?

Yes!

eps1lon (approved):

PR #5232, opened 2022-11-02
Update reference-react-dom-client.md (import)

reloaded22:
I think this should be clarified, specially for newbies like myself


hexiaokang (commented):
I don't think it's necessary.

PR #5234, opened 2022-11-02
[Beta] Add missing styles to search box in modal

zqianem:
The style when focused now matches that of the search box in the sidebar.

Existing style in sidebar:
image

Style in modal before this PR:
image

Style in modal after this PR:
image

This was most noticeable in Firefox because of the default browser focus styling, which was not visible in Chromium.

### DIFF
--- a/beta/src/styles/algolia.css
+++ b/beta/src/styles/algolia.css
@@ -39,6 +39,7 @@ html.dark {
   @apply bg-gray-10;
   @apply outline-none;
   @apply h-auto;
+  @apply focus-within:ring;
 }
 html.dark .DocSearch-Form {
   @apply bg-gray-80;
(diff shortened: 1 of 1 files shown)


harish-sethuraman:
I don't see a focus ring for modal search input in chrome? May be we should remove it in firefox as well? or we should consider adding at both places? Lets keep it consistent

zqianem:
Assuming you mean before this PR, then yes, Chrome did not have a visible focus ring on the modal input. I believe this was unintentional, as the `focus:outline-none` class was not used here when it was used for the sidebar input.

If you mean that there is no focus ring on Chrome after the PR, then that is strange — the focus ring is working for me on Chromium.

I think we should add the ring to the modal, since it already exists in the sidebar and makes it clear that the inputs are related as the modal opens. This is more apparent when the sidebar input moves to the topbar when window size is `~900px`: before, clicking on the input would cause a flash of the `ring` style before it disappeared; now, it stays styled.

harish-sethuraman:
Ah I got it 😅 . It works fine my bad :(

zqianem:
No worries, all good. Thank you for the review!

PR #5253, opened 2022-11-10
Update 2022-03-08-react-18-upgrade-guide.md

AnatuGreen:
This is because simply using npm install react react-dom did not do it for me. It just ran but no update was made to react. This suggested method for upgrade is from: https://bobbyhadz.com/blog/react-update-version


harish-sethuraman (commented):
`npm install` installs the latest version of the package? We would not need `@latest` at the end

PR #5259, opened 2022-11-12
Added my modern looking stopwatch and timer app project repo in examples.md

Vasu7389:
Added modern looking stopwatch and timer app built with ReactJs.


BartoszKlonowski:
Hello @Vasu7389!
Unfortunately the examples section is no longer kept nor maintained in the react.dev, so I will close this PR as outdated.
Thank you!

Issue #5218, opened 2022-10-28
[Beta] [Bug] State: A Component's Memory - Challenge 1 of 4: Complete the gallery


dwolrdcojp:
```JavaScript
  function handleNextClick() {
    setIndex(index + 1000);
  }
```

Should be

```JavaScript
  function handleNextClick() {
    setIndex(index + 1);
  }
```

![Screenshot 2022-10-28 at 10 35 21 AM](https://user-images.githubusercontent.com/6629474/198649345-6090ee63-ee85-42e1-9fd7-8f67d3d18c9b.jpg)

arafatiqbal:
I see `index + 1` not 1000 
![image](https://user-images.githubusercontent.com/70502771/198846132-7f8d5bff-bc06-4b00-bcaf-689cc34d4817.png)

dwolrdcojp:
> I see `index + 1` not 1000 
> 
> ![image](https://user-images.githubusercontent.com/70502771/198846132-7f8d5bff-bc06-4b00-bcaf-689cc34d4817.png)
> 
> 

Looks like that's an example higher up. I'm referring to the challenge problem at the bottom specifically.

arafatiqbal:
I see what you mean, yeah. The challenge description states that it should crash when pressing "next" on the last picture. But it doesn't ever go to last picture. If nobody has taken this on can I create a PR?

PR #5266, opened 2022-11-15
Update with correct index incrementation

arafatiqbal:
Issue: #5218


gaearon:
not sure what happened there. thanks

PR #5269, opened 2022-11-16
Change in codeblock

FelipeAlvesLeao:
Changed the codeblocks that used <Fragment> component, which needed importing, to the commonly used <React.Fragment> syntax


gaearon:
Thanks, but this is not necessary, and in most cases the global isn't available so showing importing is correct.

PR #5272, opened 2022-11-16
[Beta] useDebugValue API

gaearon:
## [Preview](https://beta-reactjs-org-git-usedebug-fbopensource.vercel.app/apis/react/useDebugValue)


beta/src/content/apis/react/useDebugValue.md
```diff
@@ -1,22 +1,125 @@
 ---
-title: useDebugValue
+title: useContext
```
harish-sethuraman, line 2: We are not using the title anywhere? and this should be `useDebugValue`

gaearon: i think i fixed that after merge

PR #5278, opened 2022-11-18
[Beta] Split API into Hooks and APIs

gaearon:
## [Preview](https://beta-reactjs-org-git-new-nav-fbopensource.vercel.app/learn)

---

This removes the "Home" tab which wasn't really useful and was taking up space.

The "API" tab has been split in two:

- Hooks
- APIs

This emphasizes modern APIs (they're no longer shifted to the bottom of the sidebar, as it used to be with a single API tab). We'll also use the Hooks tab for referencey material like Rules of Hooks which previously didn't quite fit anywhere. The APIs tab now starts with components (like `<Fragment>`) which are placed before other APIs. We'll do the same for HTML tags in `react-dom` entry point.

## Before

Screenshot 2022-11-18 at 16 59 27

## After

Screenshot 2022-11-18 at 21 20 03
Screenshot 2022-11-18 at 21 20 07

## TODO

- [x] Initial move
- [ ] Rehome content previously at "Home" 
  - [ ] Where does the blog go?
- [ ] Write index pages for Hooks
- [ ] Decide on titles for index pages
- [ ] Legacy index page


gaearon:
https://github.com/reactjs/reactjs.org/pull/5287

PR #5285, opened 2022-11-21
This document wrongly states that you cannot use ref in a functional …

sonic1981:
…component


eps1lon (changes requested):
The documentation is about using refs **on** function components e.g. `<FunctionComponent ref={ref} />`. Not refs **in** function components. This is highlighted in the example code.

PR #5287, opened 2022-11-21
[Beta] Revamp navigation

gaearon:
### tldr

- Two tabs (Learn and APIs), not three. "Home" is removed.
- React APIs are split into Hooks, Other APIs, and Legacy APIs.
- Makes the API index pages a lot more dense.

The goal is to emphasize modern APIs.

---

Alternative to https://github.com/reactjs/reactjs.org/pull/5278.

There's two tabs: Learn and API.

Screenshot 2022-11-21 at 17 07 00

Inside API, we have several groups. This gives us more flexibility as we can introduce and merge groups as needed. This also keeps the same URL format for all APIs so tools can assume `/apis/<entry>/<export>` always goes somewhere.

Legacy stuff is moved to the very bottom beneath a separator.

Screenshot 2022-11-21 at 17 07 40


eps1lon:
Tooling consideration wrt to URLs is nice 👍

PR #5290, opened 2022-11-22
Added about arrow functions to docs

naokiuma:
Hello!

I'm a beginner in React.
When I saw Document page,I can't tell whether it is going to use "Arrow function" to make component  in React.
I think, some people know about arrow function but some don't know that you can use it in React.
 
 
So I thought This Document page( https://reactjs.org/docs/components-and-props.html ) should contains a description about it.
※like this page. https://reactjs.org/tutorial/tutorial.html
 
thanks :-)

### DIFF
--- a/content/docs/components-and-props.md
+++ b/content/docs/components-and-props.md
@@ -46,6 +46,15 @@ The above two components are equivalent from React's point of view.
 
 Function and Class components both have some additional features that we will discuss in the [next sections](/docs/state-and-lifecycle.html).
 
+>**Note:** Of course, you can use arrow functions too.
+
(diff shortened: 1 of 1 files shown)


BartoszKlonowski:
Hello @naokiuma!
Function component are deeply described in details [here](https://react.dev/learn/your-first-component#defining-a-component), while the arrow notation is a common syntax for JS, not reserved to React, so it doesn't need to be described anywhere in React docs (can be linked when mentioned though).
However, as the part you modify is no longer maintained in the docs I will close this PR as outdated.
Thank you!

PR #5292, opened 2022-11-23
Chore: its -> it is, to 1 line

marcelrsoub:
Quick correction and setting it to one line as in the other paragraphs.


eps1lon (approved):
Thanks!

PR #5293, opened 2022-11-24
Chore: fix code example

YongwoonJang:
Because "const EnhancedComponent = withRouter(connect(commentSelector)(WrappedComponent))"(line 282) is not the same as "const enhance = compose(withRouter, connect(commentSelector)); const EnhancedComponent = enhance(WrappedComponent)"(line 286 ~ 291).

### DIFF
--- a/content/docs/higher-order-components.md
+++ b/content/docs/higher-order-components.md
@@ -279,7 +279,7 @@ This form may seem confusing or unnecessary, but it has a useful property. Singl
 
 ```js
 // Instead of doing this...
-const EnhancedComponent = withRouter(connect(commentSelector)(WrappedComponent))
+const EnhancedComponent = withRouter(connect(commentSelector))(WrappedComponent)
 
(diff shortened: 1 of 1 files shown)


content/docs/higher-order-components.md
```diff
@@ -279,7 +279,7 @@ This form may seem confusing or unnecessary, but it has a useful property. Singl
 
 ```js
 // Instead of doing this...
-const EnhancedComponent = withRouter(connect(commentSelector)(WrappedComponent))
+const EnhancedComponent = withRouter(connect(commentSelector))(WrappedComponent)
```
eps1lon, line 282: Previous code was correct. `withRouter` takes a single argument and returns a component: https://v5.reactrouter.com/web/api/withRouter. It doesn't return a function as your code change suggests.

eps1lon (changes requested):

PR #5298, opened 2022-11-26
Update the link to the Node.js stream module

### LINKED ISSUE #5297: The link of node.js streams is outdated
On the page https://reactjs.org/docs/react-dom-server.html the link the link on node.js streams is outdated and doesn't work https://reactjs.org/docs/react-dom-server.html#:~:text=the%20environments%20with-,Node.js%20Streams%3A,-renderToPipeableStream() . I believe the link should link to https://nodejs.org/api/stream.html

xfq:
Fix #5297.

### DIFF
--- a/content/docs/reference-react-dom-server.md
+++ b/content/docs/reference-react-dom-server.md
@@ -17,7 +17,7 @@ var ReactDOMServer = require('react-dom/server');
 
 ## Overview {#overview}
 
-These methods are only available in the **environments with [Node.js Streams](https://nodejs.dev/learn/nodejs-streams):**
+These methods are only available in the **environments with [Node.js Streams](https://nodejs.org/api/stream.html):**
 
 - [`renderToPipeableStream()`](#rendertopipeablestream)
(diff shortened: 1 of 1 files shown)


content/docs/reference-react-dom-server.md
```diff
@@ -17,7 +17,7 @@ var ReactDOMServer = require('react-dom/server');
 
 ## Overview {#overview}
 
-These methods are only available in the **environments with [Node.js Streams](https://nodejs.dev/learn/nodejs-streams):**
+These methods are only available in the **environments with [Node.js Streams](https://nodejs.org/api/stream.html):**
```
harish-sethuraman, line 20: The previous link seems to be `/learn/` ? Also the main `nodejs.org` site is being re written at `nodejs.dev` so it should be better that we link to nodejs.dev? WDYT?

harish-sethuraman: Yea I think we can use `nodejs.dev` link itself.

gaearon:
let's get this in for now. thanks

PR #5307, opened 2022-11-30
fix: correct order of future conference dates content

techieeliot:
- Update markdown word placement
- Move June 2023 conference date behind May 2023 dates


techieeliot:
✍️ Signed the CLA

Screen Shot 2022-11-30 at 11 55 19 AM

eps1lon (changes requested):
Conferences are sorted by date descending so the previous order makes sense to me. Thanks for the suggestion though.

techieeliot:
> Conferences are sorted by date descending so the previous order makes sense to me. Thanks for the suggestion though.

I should have included this screenshot earlier. 

The dates for upcoming conferences is chronological order. June 27 is wrongly placed before May, and chronologically it is last in the list @eps1lon

techieeliot:
Thank you, @eps1lon!

PR #5310, opened 2022-12-01
[Beta] Add link anchor for ExpandableExample (Deep Dive)

ibarapascal:
The issue is here: https://github.com/reactjs/reactjs.org/issues/5304

### What's included

Added an anchor for the `ExpandableExample` component in Beta.

- src/components/MDX/ExpandableExample.tsx

### What's achieved

We can now share the link of the content like 'deep dive', once we open that link, it can scroll to the exact place where the content is, just like other markdown headers (which have the anchor link already).

try it locally: http://localhost:3000/apis/react/isValidElement#react-elements-vs-react-nodes

### Does this affect the UI

No, all the styles maintain the same as before, only an existing anchor link component was added. It appears when the user hovers over the title.

---

~~### Does this need adjustment for the existing markdown documents?~~

~~No, currently just using the title as the URL name after the hash mark.~~

~~### Remark points~~

~~- Block title should be unique inside the page, if not, we may want to add another prop for the anchor (require edition on the existing markdowns), which is not actually necessary I guess.~~

~~- Later on, we can make the anchor name better formatted, although currently, it's already good at the function itself. Keeping the space or trying another formatting could cause a malfunction in some cases.~~

### DIFF
--- a/beta/src/components/MDX/ExpandableExample.tsx
+++ b/beta/src/components/MDX/ExpandableExample.tsx
@@ -22,7 +22,7 @@ function ExpandableExample({children, excerpt, type}: ExpandableExampleProps) {
   const isExample = type === 'Example';
   const hasTitle =
     Array.isArray(children) &&
-    ['h2', 'h3', 'h4'].includes(children[0].type.mdxName);
+    ['h2', 'h3', 'h4'].includes(children[0]?.type?.mdxName);
 
   return (
     <details
--- a/beta/src/components/MDX/ExpandableExample.tsx
+++ b/beta/src/components/MDX/ExpandableExample.tsx
@@ -8,28 +8,24 @@ import {IconChevron} from '../Icon/IconChevron';
 import {IconDeepDive} from '../Icon/IconDeepDive';
 import {IconCodeBlock} from '../Icon/IconCodeBlock';
 import {Button} from '../Button';
+import {H4} from './Heading';
 
 interface ExpandableExampleProps {
   children: React.ReactNode;
-  title: string;
   excerpt?: string;
   type: 'DeepDive' | 'Example';
 }
 
-function ExpandableExample({
-  children,
-  title,
-  excerpt,
-  type,
-}: ExpandableExampleProps) {
+function ExpandableExample({children, excerpt, type}: ExpandableExampleProps) {
   const [isExpanded, setIsExpanded] = React.useState(false);
   const isDeepDive = type === 'DeepDive';
   const isExample = type === 'Example';
-  const id = title.split(' ').join('');
+  const hasTitle =
+    Array.isArray(children) &&
+    ['h2', 'h3', 'h4'].includes(children[0].type.mdxName);
 
   return (
     <details
-      id={id}
       open={isExpanded}
       onToggle={(e: any) => {
         setIsExpanded(e.currentTarget!.open);
@@ -42,7 +38,8 @@ function ExpandableExample({
         className="list-none p-8"
         tabIndex={-1 /* there's a button instead */}
         onClick={(e) => {
-          if (e.target instanceof HTMLAnchorElement) return;
+          // Escape case for the header anchor link
+          if (e.target instanceof SVGElement) return;
           // We toggle using a button instead of this whole area.
           e.preventDefault();
         }}>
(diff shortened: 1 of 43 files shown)


ibarapascal:
Yeah I didn't have thought that far, will try more later via that approach

ibarapascal:
>- Changing the DeepDive component to deal with that format
>- Changing all deep dives in MDX to be in this format

```
<DeepDive>

#### This is a title {/*this-is-a-title*/}

...

</DeepDive>
```

@gaearon  The above two points are done, and all the docs with `DeepDive` are adjusted.

From the current implementation

- If merged, the `title` props in the `DeepDive` element won't show any longer.  
- The title/link could be optional.  
- If there is a title, it needs to be the first child inside `DeepDive`.  
- The title could be `h2`, `h3`, or `h4` for any different level of usage.  

Thought it would be a better fit now compared to the last try xD

beta/src/components/MDX/ExpandableExample.tsx
```diff
@@ -8,23 +8,21 @@ import {IconChevron} from '../Icon/IconChevron';
...
-  excerpt,
-  type,
-}: ExpandableExampleProps) {
+function ExpandableExample({children, excerpt, type}: ExpandableExampleProps) {
   const [isExpanded, setIsExpanded] = React.useState(false);
   const isDeepDive = type === 'DeepDive';
   const isExample = type === 'Example';
+  const hasTitle =
+    Array.isArray(children) &&
+    ['h2', 'h3', 'h4'].includes(children[0].type.mdxName);
```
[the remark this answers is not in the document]
ibarapascal: > nit: Does type always exists in the `children[0]`? If not we could optional chain else chill 👍🏻

Yeah, I think it always exists, according to its type definition. No matter how, since the first child should always be the title, that should be ok, and the case with an optional title was also been tested.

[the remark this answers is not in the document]
ibarapascal: @harish-sethuraman  thx! updated!

gaearon:
Navigating to a Deep Dive anchor should auto-expand it.

beta/src/components/MDX/ExpandableExample.tsx
```diff
@@ -40,6 +38,8 @@ function ExpandableExample({
         className="list-none p-8"
         tabIndex={-1 /* there's a button instead */}
         onClick={(e) => {
+          // Escape case for the header anchor link
+          if (e.target instanceof SVGElement) return;
```
gaearon, line 42: Nit: please use multiline style for if statements

beta/src/components/MDX/ExpandableExample.tsx
```diff
@@ -8,23 +8,21 @@ import {IconChevron} from '../Icon/IconChevron';
...
-  children,
-  title,
-  excerpt,
-  type,
-}: ExpandableExampleProps) {
+function ExpandableExample({children, excerpt, type}: ExpandableExampleProps) {
   const [isExpanded, setIsExpanded] = React.useState(false);
   const isDeepDive = type === 'DeepDive';
   const isExample = type === 'Example';
+  const hasTitle =
```
gaearon, line 23: I would like to raise an error if the title is not found. Also let's force it to always be `h4`. Basically we should be strict about the format so that it's impossible to write a DeepDive without a proper header.

Another way you can do this, btw, is to preprocess it in `prepareMDX` and generate a regular `title` prop there. But I guess for now this is fine.

gaearon (commented):
looks great. let's make it a bit stricter and then we're good to go. thank you so much for working on this, people have been asking for it for a long time.

beta/src/components/MDX/ExpandableExample.tsx
ibarapascal: Updated

ibarapascal:
emmm not sure what let the CI fail, local build went all green though, anyone who could enlighten me?

gaearon:
Looks like this:

```

Error occurred prerendering page "/apis/react/useDeferredValue". Read more: https://nextjs.org/docs/messages/prerender-error
Error: Expandable content DeepDive is missing a corresponding title at the beginning
    at ExpandableExample (/home/runner/work/reactjs.org/reactjs.org/beta/.next/server/chunks/915.js:2904:15)
```

Try rebasing and then fix the newly merged page.

gaearon:
This is great, thanks.

A few follow-up ideas if you're interested:

- Navigating to a deep dive link should auto-expand it. Not sure what's the best way to implement this, so ideas are welcome.
- Individual Challenges should similarly be linkable.

PR #5311, opened 2022-12-01
[Beta] useTransition API

gaearon:
## [Preview](https://beta-reactjs-org-git-usetransition-fbopensource.vercel.app/apis/react/useTransition)

### DIFF
--- a/beta/src/content/apis/react/useTransition.md
+++ b/beta/src/content/apis/react/useTransition.md
@@ -2,19 +2,1393 @@
 title: useTransition
 ---
 
-<Wip>
-
-This section is incomplete, please see the old docs for [useTransition.](https://reactjs.org/docs/hooks-reference.html#usetransition)
-
-</Wip>
-
-
 <Intro>
 
+`useTransition` is a React Hook that lets you mark a state transition as non-urgent.
+
 ```js
-const [isPending, startTransition] = useTransition();
+const [isPending, startTransition] = useTransition()
 ```
 
 </Intro>
 
 <InlineToc />
+
+---
+
+## Usage {/*usage*/}
+
+<Note>
+
+`useTransition` is mostly useful if you're building a [Suspense-enabled](/apis/react/Suspense) router or a framework.
+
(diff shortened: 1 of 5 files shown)


beta/src/content/apis/react/useTransition.md
```diff
@@ -2,19 +2,1393 @@
...
 
-<Wip>
+<Intro>
 
-This section is incomplete, please see the old docs for [useTransition.](https://reactjs.org/docs/hooks-reference.html#usetransition)
+`useTransition` is a React Hook that lets you mark a state transition as non-urgent.
 
-</Wip>
+```js
+const [isPending, startTransition] = useTransition()
```
jbonta, line 10: I'm sure this is intentional and it's been covered in the past, but can you talk about the decision on whether or not to include a semicolon in these intro code blocks?

On one hand, when it's a statement (and especially a jsx tag), a semicolon seems awkward or distracting. On the other, especially when it's an assignment, it sort of seems incomplete or like it could be a mistake.

damianstasik:
@gaearon In [this example](https://beta-reactjs-org-git-usetransition-fbopensource.vercel.app/apis/react/useTransition#indicating-that-a-transition-is-happening-inline) you are passing `isPending` to `Layout` inside `Router` component in order to change opacity of the header section. By "rewrapping" `navigate` with `startTransition` in the `IndexPage` component it seems like `isPending` inside `Router` is always `false` and does not affect the header's opacity. Is that expected behavior? If so I would imagine it could break router logic that depends on the `isPending` flag.

gaearon:
Hmm interesting. Maybe this is a bug. I'll need to follow-up on that.

beta/src/content/apis/react/useTransition.md
gaearon: I don't care strongly either way, the idea was just that these intro blocks aren't always complete statements and they're more about showing the signature. E.g. some aren't even valid JS syntax, like using `?` for optional args. If someone feels like these should have semis I'm happy to take a PR adding them.

PR #5318, opened 2022-12-05
[Beta] Anchor for individual challenges & deepdive

ibarapascal:
prev PR here: https://github.com/reactjs/reactjs.org/pull/5310#issuecomment-1335866959

### What's included

- Made DeepDive's content auto-expand once accessed via a direct link.  
- Made individual challenges linkable.  

### Does this affect the UI

Expected to be no, Individual Challenges' styles have been updated to fit the previous layout.  

~~### Test results~~

~~All the included functions work well, but perhaps it needs further caution to watch closely when we access the page via the browser `goBack` button, and check if there are unexpected/unnecessary renders, breaks during scroll, etc.~~

---

<details>

<summary>Discussion points</summary>

#### 1. To auto-expand the content

As well known, we can directly set the initial state rather than reset it inside `useEffect` later.  
However, to check the initial expanded status for DeepDive, I faced an issue using the nextjs `asPath` to get the current generated URL, as shown below

>Using the `asPath` field may lead to a mismatch between client and server if the page is rendered using server-side rendering or [automatic static optimization](https://nextjs.org/docs/advanced-features/automatic-static-optimization). Avoid using `asPath` until the `isReady` field is `true`.  

ref: [nextjs router docs](https://nextjs.org/docs/api-reference/next/router#router-object)

there are some workarounds I've tried

- dynamic import without SSR for part of the component which uses that initial status -> seems won't work cause we are using that state at the very beginning in `<details />`.  
- use [suppressHydrationWarning](https://reactjs.org/docs/dom-elements.html#suppresshydrationwarning) for all the elements using that state -> seems not to be a good practice at this moment.  
- use Effect to reset the state from default `false` to `true` if we are on a DeepDive anchor, this could cause page flash/re-rendering.  

After those attempts, I just realized that why don't we use a default `true` for that status to make that workaround better, because

- If we are on a DeepDive, expanded at the beginning so no more broken user experience.  
- If not, the expanded content would change its status to `false` without being seen.  

Now they work as expected, but after all, I guess it's still a workaround so maybe it's not the best way to do it and could be changed later if possible.

#### 2. For the browser's `goBack`

We definitely want to maintain our DeepDive content expanded even if we route to another place and then go back again (if we are on the DeepDive already), that's the reason I use mounted state rather than an independent variable outside the component to make the Effect only run once (as recommended in docs), and set the state inside Effect without further conditions.  

#### 3. About the multiple linkable individual challenges

Each challenge is now having its own anchor, but except for the default (first) one, others won't anchor in the first place cause they are not been rendered.

Assuming we are not gonna step into the _render all, visible one_ solution, lacking further tricks known, the only way that came to my mind is to use an Effect again.  

In this way, I used the previous `scrollIntoVIew` without thinking, I must admit this may not be the best solution, and if it really won't match the expected behavior, closing this would be fine. Please let me know if there are better ways to do that, thanks for reading!

</details>

### DIFF
--- a/beta/src/components/MDX/Challenges/Challenges.tsx
+++ b/beta/src/components/MDX/Challenges/Challenges.tsx
@@ -2,7 +2,7 @@
  * Copyright (c) Facebook, Inc. and its affiliates.
  */
 
-import {Children, useRef, useEffect, useState, useCallback} from 'react';
+import {Children, useRef, useEffect, useState} from 'react';
 import * as React from 'react';
 import cn from 'classnames';
 import {H2} from 'components/MDX/Heading';
@@ -83,20 +83,15 @@ export function Challenges({
   const {asPath} = useRouter();
   const [isMounted, setMounted] = useState(false);
 
-  const scrollIntoView = useCallback(() => {
-    scrollAnchorRef.current!.scrollIntoView({
-      block: 'start',
-      behavior: 'smooth',
-    });
-  }, []);
-
   useEffect(() => {
     const challengeIndex = challenges.findIndex(
       (challenge) => challenge.id === asPath.split('#')[1]
     );
     if (challengeIndex !== -1) {
       if (isMounted) {
-        scrollIntoView();
+        scrollAnchorRef.current!.scrollIntoView({
+          block: 'start',
+        });
       } else {
         setActiveIndex(challengeIndex);
       }
@@ -104,12 +99,15 @@ export function Challenges({
     if (!isMounted) {
       setMounted(true);
     }
-  }, [asPath, challenges, isMounted, scrollIntoView]);
+  }, [asPath, challenges, isMounted]);
 
   useEffect(() => {
     if (queuedScrollRef.current === true) {
       queuedScrollRef.current = false;
-      scrollIntoView();
+      scrollAnchorRef.current!.scrollIntoView({
+        block: 'start',
+        behavior: 'smooth',
+      });
     }
   });
 
--- a/beta/src/components/MDX/Challenges/Challenges.tsx
+++ b/beta/src/components/MDX/Challenges/Challenges.tsx
@@ -68,6 +68,11 @@ const parseChallengeContents = (
   return contents;
 };
 
+enum QueuedScroll {
+  INIT = 'init',
+  NEXT = 'next',
+}
+
 export function Challenges({
   children,
   isRecipes,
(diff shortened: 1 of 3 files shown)


gaearon:
This looks good. When navigating Back to a Challenge, can we make it scroll into view too? Currently this only seems to work for Deep Dives.

ibarapascal:
> This looks good. When navigating Back to a Challenge, can we make it scroll into view too? Currently this only seems to work for Deep Dives.

Thanks! Move away the `smooth` inside `scrollIntoVIew` fixed it.

<details>

<summary>More</summary>

I know the implementation here seems redundant. If we are not using SSR for this part, the init state could be set directly without further re-rendering. I will keep an eye on this part, later on, to see if any changes have been made to improve the user experience. Thanks again for your review!

</details>

beta/src/components/MDX/Challenges/Challenges.tsx
```diff
@@ -79,6 +80,26 @@ export function Challenges({
...
+      if (isMounted) {
+        scrollAnchorRef.current!.scrollIntoView({
+          block: 'start',
+        });
+      } else {
+        setActiveIndex(challengeIndex);
+      }
+    }
+    if (!isMounted) {
+      setMounted(true);
```
gaearon, line 100: This seems unnecessary since `isMounted` is never used for rendering. We should avoid triggering a re-render since there's no actual visual change. Let's keep this boolean in a ref instead.

beta/src/components/MDX/ExpandableExample.tsx
```diff
@@ -17,15 +19,28 @@ interface ExpandableExampleProps {
...
+  // use mounted state here incase of browser goBack
+  const [isMounted, setMounted] = useState(false);
+  // init as expanded to prevent flash
+  const [isExpanded, setIsExpanded] = useState(true);
+
+  // asPath would mismatch between server and client, reset here instead of put it into init state
+  useEffect(() => {
+    if (!isMounted) {
+      setMounted(true);
+      setIsExpanded(id === asPath.split('#')[1]);
```
gaearon, line 41: We should only do these state updates if we the condition actually matched. Otherwise, don't do anything. This prevents unnecessary work.

beta/src/components/MDX/Challenges/Challenges.tsx
ibarapascal: Updated! That's totally my bad, I remember the recommendation of the ref usage in beta docs but failed to react as the first option to use.

gaearon:
Let's get this in. Thank you!

PR #5332, opened 2022-12-08
Functional updates, use term current state

toreerstad:
I find it a bit confusing using term previous state for functional updates. Suggest changing to current state.


harish-sethuraman (commented):
Hey 👋🏻  thanks for sending an PR but however newer documentation is being written at beta.reactjs.org Please make your suggestions for that documentation for the newer one.

toreerstad:
Ah, the new documentation explains this much better. Thanks.

PR #5335, opened 2022-12-09
createPortal docs

davidmccabe:
Tried to balance being fussy and technically correct about React nodes vs elements etc. with understandability and avoiding confusion between React nodes and DOM nodes.


gaearon:
rebased as https://github.com/reactjs/reactjs.org/pull/5335

PR #5339, opened 2022-12-09
[Beta] renderToPipeableStream API

gaearon:
Closes https://github.com/reactjs/reactjs.org/pull/4623. (Thanks, @RoyalIcing — this was really helpful.)

## [Preview](https://beta-reactjs-org-git-server-2-fbopensource.vercel.app/apis/react-dom/server/renderToPipeableStream)


beta/src/content/apis/react-dom/server/renderToPipeableStream.md
```diff
@@ -2,22 +2,576 @@
...
+It determines the earliest loading state that the user may see:
+
+```js {3-5,13
+<ProfileLayout>
+  <ProfileCover />
+  <BigSpinner />
+</ProfileLayout>
+```
+
+If you wrap the whole app into a `<Suspense>` boundary at the root, the shell will only contain that spinner. However, that's not a pleasant user experience because seeing a big spinner on the screen can feel slower and more annoying than waiting a bit more and seeing the real layout. This is why usually you'll want to place the `<Suspense>` boundaries so that the shell feels *minimal but complete*--like a skeleton of the entire page layout.
```
gnoff, line 271: @gaearon @sebmarkbage 

I don't think this is supported actually. You can't really put a Suspense boundary around the <html> tag because the placeholder html will force the HTML parser into body mode before the actual <html> tag is parsed. and when you swap the real content streams in the <html> and <head> tags it contains will get swallowed.

We probably need to indicate that the highest you can put a Suspense Boundary is as a child of `<head>` or `<body>`

gaearon: Hmm yeah by “root” here I meant inside body but I can see how that’s confusing

PR #5345, opened 2022-12-10
Expected header ID to be like: {/*some-header*/}.

domscript:
Expected header ID to be like: {/*some-header*/}.

        throw Error(
          'Cannot have a duplicate header with id "' +
            id +
            '" on the page. ' +
            'Rename the section or give it an explicit unique ID. ' +
            'For example: #### Arguments {/*setstate-arguments*/}'
        );

in "/blog/*.md" 4 or 5  documents have the same ;)

### DIFF
--- a/beta/src/components/Layout/Nav/Nav.tsx
+++ b/beta/src/components/Layout/Nav/Nav.tsx
@@ -241,6 +241,11 @@ export default function Nav() {
             </button>
           </div>
         </div>
+        {!isOpen && (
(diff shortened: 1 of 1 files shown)


gaearon:
what did you run to get the error? not sure why i can't reproduce locally

domscript:
[screen_shot with error](https://drive.google.com/file/d/1HxkTPC_YorhZ3hgqoscx1YHfi3cTB_ok/view)

PR #5348, opened 2022-12-11

gunayanil:
The statement about returning the falsy expression in the conditional rendering (`&&`) is not 100% correct. React only returns the falsy expression if it is `0`, `-0` or `NaN`. So, all other falsy expressions or values (e.g. `undefined`, `null`, `false`, `''`, etc) will not be returned by the render method.


gunayanil:
Any update?

PR #5363, opened 2022-12-14
[Beta] Split "Components" and "APIs"

gaearon:
Feels nicer.

Components like `<Suspense>` feel pretty fundamental so it's weird to group them under "_other_ APIs".

This also mirrors a similar upcoming split for ReactDOM.

### DIFF
--- a/beta/src/sidebarAPIs.json
+++ b/beta/src/sidebarAPIs.json
@@ -76,8 +76,8 @@
           ]
         },
         {
-          "title": "react: Other APIs",
-          "path": "/apis/react/other",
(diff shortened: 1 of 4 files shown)


beta/src/sidebarAPIs.json
```diff
@@ -94,7 +94,13 @@
             {
               "title": "<Suspense>",
               "path": "/apis/react/Suspense"
-            },
+            }
+          ]
+        },
+        {
+          "title": "react: APIs",
+          "path": "/apis/react/apis",
```
gaearon, line 102: lol

Issue #5321, opened 2022-12-05
[Beta][Bug] Command+Click Search Results Opens New Tab


yungsters:
When I command+click a search result, the result opens in a new tab. However, the search results "blur" / disappear (and search input resets).

I think that instead, the search results should continue being visible. My use case for command+clicking a result is to open multiple results in separate tabs so I can look at all of them.

![CleanShot 2022-12-05 at 16 33 30](https://user-images.githubusercontent.com/55161/205747365-d33a6bc5-097a-4389-a90b-2250b47f5f2c.gif)

abhinandanwadwa:
Hey @harish-sethuraman , I've fixed this issue

abhinandanwadwa:
@gaearon

PR #5370, opened 2022-12-16
[Beta][Bug] Fix: Command+Click Search Results Opens New Tab #5321

abhinandanwadwa:
I've fixed this bug ( #5321 ): When command+clicked on the search results, it opens in a new tab but the sidebar closes and the search field resets. I've fixed this bug by adding an onClick listener to the parent div of the search results component.

https://user-images.githubusercontent.com/74089487/208090245-d3cdd07b-37b1-4b7c-8ee5-e269e8ba5277.mov


shanpriyan (changes requested):
Modal is closing when we click on the search bar

[react-search.webm](https://user-images.githubusercontent.com/42532987/208149247-05fe9924-603d-4797-a81a-16c384f2084d.webm)

beta/src/components/Search.tsx
```diff
@@ -128,8 +128,23 @@ export function Search({
...
 
   const onClose = useCallback(
-    function onClose() {
-      setIsShowing(false);
+    function onClose(event: any) {
+      if (event.ctrlKey || event.metaKey) {
+      } else {
+        const targetElement = event.target;
+        const clickedOnLink =
+          targetElement.classList.contains('DocSearch-Hit') ||
```
gaearon, line 136: These hardcoded class checks aren't great and look fragile. What are you trying to solve here? Can you think of another approach?

abhinandanwadwa: I'm trying to think of better approach. Here, first i added a parent div to the around the `DocSearchModal` to add an `onClick` listener, to check if the person did cmd+click or normal click so as to not close the search bar if it was cmd+click. Then, as @shanpriyan mentioned that the search bar is getting closed whenever clicked anywhere inside it. So, i started searching for the child classes so as to **close the sidebar only if that was a normal click on the search result and not cmd+click**.

PR #5373, opened 2022-12-16
[Beta] useEvent -> useEffectEvent

gaearon:
To mirror https://github.com/facebook/react/pull/25881.

- `useEvent` -> `useEffectEvent`
- "Event function" -> "Effect Event"
- Not stable functions anymore, instead they have to be omitted from deps


karlhorky:
Thanks for this @gaearon ! 

Since @sebmarkbage [mentioned that the RFC is "defunct"](https://github.com/facebook/react/pull/25881#issuecomment-1355157375), should one or both of the following things also happen?

1. Remove the link to the RFC from the docs
2. Remove or rework the RFC at https://github.com/reactjs/rfcs/blob/useevent/text/0000-useevent.md (I can create a separate issue about this over there if necessary)

gaearon:
We need to post a new RFC but it's effectively the same content that's in the docs. I guess removing the link would also work.

PR #5375, opened 2022-12-17
[Beta] Fix scroll jumps on navigating to fragments

gaearon:
Fixes a regression introduced in https://github.com/reactjs/reactjs.org/pull/5310.

Previously, we would expand all deep dives in SSR and then collapse them in an Effect except the one that has the current fragment hash. This doesn't make sense and causes layout jumps on hydration. Particularly I've observed jumps on mobile Safari when navigating to a fragment anchor that comes *after* a deep dive because the deep dive expansion shifts layout. What we should be doing instead is keep them all collapsed, and then auto-expand the one that's in the hash. It's fine if it shifts layout because the fragment navigation case is not the common one.

### DIFF
--- a/beta/src/components/MDX/ExpandableExample.tsx
+++ b/beta/src/components/MDX/ExpandableExample.tsx
@@ -28,20 +28,17 @@ function ExpandableExample({children, excerpt, type}: ExpandableExampleProps) {
   const isExample = type === 'Example';
   const id = children[0].props.id;
 
-  const queuedExpandRef = useRef<boolean>(true);
   const {asPath} = useRouter();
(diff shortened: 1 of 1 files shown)


AndrewAntoniuk:
Hi! @gaearon 
Could you let me know if you are sure it is fixed?
I found that smth is forcing the page to jump still 
[Screencast 2023-05-04 17-41-30.webm](https://user-images.githubusercontent.com/83472735/236285668-afd33c41-300a-44a9-85bc-0424f74d014d.webm)

gaearon:
Please file a new issue — we'll lose this otherwise :)

PR #5385, opened 2022-12-20
[Beta] make code some simpler.

zhangenming:

### DIFF
--- a/beta/src/content/learn/choosing-the-state-structure.md
+++ b/beta/src/content/learn/choosing-the-state-structure.md
@@ -2754,15 +2754,13 @@ export default function MailClient() {
 
   const selectedCount = selectedIds.size;
 
-  function handleToggle(toggledId) {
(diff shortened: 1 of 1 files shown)


beta/src/content/learn/choosing-the-state-structure.md
```diff
@@ -2754,15 +2754,13 @@ export default function MailClient() {
...
   const selectedCount = selectedIds.size;
 
-  function handleToggle(toggledId) {
-    // Create a copy (to avoid mutation).
-    const nextIds = new Set(selectedIds);
-    if (nextIds.has(toggledId)) {
-      nextIds.delete(toggledId);
+  function handleToggle(id) {
+    if (selectedIds.has(id)) {
+      selectedIds.delete(id);
```
gaearon, line 2759: This mutates the previous state which is the thing we wanted to avoid.

PR #5390, opened 2022-12-21
Update preserving-and-resetting-state.md

zhangenming:
inspired by https://github.com/reactjs/reactjs.org/blob/966b7b849f7e103d580be5cb5abf2884aca05b6d/beta/src/content/learn/choosing-the-state-structure.md?plain=1#L2758
```js
 function handleToggle(toggledId) {
    // Create a copy (to avoid mutation).
    const nextIds = new Set(selectedIds);
    if (nextIds.has(toggledId)) {
      nextIds.delete(toggledId);
    } else {
      nextIds.add(toggledId);
    }
    setSelectedIds(nextIds);
  }
```

### DIFF
--- a/beta/src/content/learn/preserving-and-resetting-state.md
+++ b/beta/src/content/learn/preserving-and-resetting-state.md
@@ -1572,13 +1572,8 @@ export default function ContactManager() {
   );
 
   function handleSave(updatedData) {
-    const nextContacts = contacts.map(c => {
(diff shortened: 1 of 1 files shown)


harish-sethuraman (changes requested):
I feel like this introduces too much complexity around just a simple map method 😅 . Let's leave it as such?

PR #5391, opened 2022-12-21
404 is shown for 1 second before matching client paths

Asad9988:
Hi React Team,

This is my first time open source contribution. I have followed the React Guidelines too. 
Issue: When we hit the blogs Link from [Beta] docs, the 404 page is shown for a second and then actual content is got through graphQl Query and is displayed. It's not naive to say that it creates a bad UX. I myself ( while i was on slow internet ), thought that docs server is down and moved away from website twice. I have been working on this and found out that this thing was fixed in Gastby new versions, and there is an implemenation/work around to fix this into previous versions ( that react docs are using). I have managed to fix it and am creating this PR. Please do let me know if there are further changes required in this PR. I have tested it and it's working fine. :)

### DIFF
--- a/src/pages/404.js
+++ b/src/pages/404.js
@@ -15,25 +15,31 @@ import {sharedStyles} from 'theme';
 type Props = {
   location: Location,
 };
-
-const PageNotFound = ({location}: Props) => (
-  <Layout location={location}>
-    <Container>
-      <div css={sharedStyles.articleLayout.container}>
-        <div css={sharedStyles.articleLayout.content}>
-          <Header>Page Not Found</Header>
-          <TitleAndMetaTags title="React - Page Not Found" />
-          <div css={sharedStyles.markdown}>
-            <p>We couldn't find what you were looking for.</p>
-            <p>
-              Please contact the owner of the site that linked you to the
-              original URL and let them know their link is broken.
-            </p>
-          </div>
-        </div>
-      </div>
-    </Container>
-  </Layout>
-);
+const PageNotFound = ({location}: Props) => {
+  const browser = typeof window !== 'undefined' && window;
+  return (
+    <div>
+      {browser && (
+        <Layout location={location}>
+          <Container>
+            <div css={sharedStyles.articleLayout.container}>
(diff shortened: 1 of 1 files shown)


gaearon:
This doesn't look like a good fix to me. It tries to work around the problem by showing different content on the client on the server (which is not supported by React). What is the real canonical way to solve this?

Assad-Ashraf:
This can be fixed by updating the  Gatsby version, canonically. But updating the version may cause some other packages conflicts, that's why i came up with this idea to check the window object.

Assad-Ashraf:
@gaearon Another work around is to fetch the actual URL while re-directing from the Beta docs and redirect to the actual one rather then going for /blog which then fetches the latest blog.

Assad-Ashraf:
In 2018, Gatsby team fixed this problem in their package and they started supporting SSR stuff.

Assad-Ashraf:
@gaearon  In 2018, Gatsby team fixed this problem in their package and they started supporting SSR stuff in V3. You are using a version lesser then v3. The real way is to have a separate page for Beta version itself, but as you are re-directing them to the previous docs, this is happening. There is no actual way to resolve it other then updating the package itself. I had it updated and i think change of this level will not be accepted in this regard. So i cam up with this solution that i have implemented. I can also create a PR after updating the version of Gatsby to v3 if you want. ATM this is the only possibility to fix it. @gaearon  What do you think about this?

Assad-Ashraf:
@gaearon I have seen there is a route to redirect /blog to /blog/all.html with all the Blogs listing, this can also be done here. Furthermore i have looked into updating the Gatsby to v3, the amount of work on it is drastic and has many conflicts with other packages. I think we should redirect the /blog to /blog/all.html or accept the work around done here because it is also recommended by Gatsby team if there is drastic change in codebase for version upgrade from v2 to v3. etc. What do you think ?
Screenshot from 2023-01-05 17-39-33
ml ? Please let me know your thoughts.

Assad-Ashraf:
@gaearon Thanks for implementing the idea i gave here. Now blogs are working fine. Should this be closed now?

PR #5392, opened 2022-12-22
Making serverUrl a reactive dependency for the provided example.

rahulrao0209:
The serverUrl variable should be moved inside the component and can be stored in the state to indicate that it's a reactive value and will be highlighted as such by the linter. Previously the serverUrl variable was stored outside the component function body thus making it a non-reactive value and so it shouldn't have been highlighted by the linter as a useEffect dependency as shown in the example.


gaearon:
nice find. thanks.

PR #5399, opened 2022-12-25
[Beta] Move Reference before Usage

gaearon:
The last time we tried it, I didn't like it. But, from what I recall, we didn't have a TOC at the time, so it was easy to miss that there's less dry content below. Now we have a TOC so it's more in your face. And we have a *lot* of Usage content unlike in the beginning. So it's become annoying to get to the Reference (presumably something you'd do more than once), even when you know it exists in the TOC. And Reference is a tldr of the whole page anyway. So it makes sense for it to come first because if you're lucky, your question is answered right there.

Let's try this again.

---

Done via regex:

```
(## Usage {.*\n(.*\n)+)(## Reference.*\n(.*\n)*)(## Troubleshooting.*\n(.*\n))
\3\1\5
```

Then

```
(## Usage {.*\n(.*\n)+)(## Reference.*\n(.*\n)*)
\3\1
```

The second one messed up some spacing and horizontal lines which I corrected manually.

---

A few other things to consider in the future:

- TOC is sometimes too large and pushes content too far out. E.g. below fold. Like on useEffect page. Maybe we should break it apart? E.g. TOC with Reference/Usage/Troubleshooting, and then a mini-TOC in each of those sections.
- For pages with one API, avoid the intermediate header to decrease repetition
- Maybe rename /apis and "API" tab to Reference now that reference comes first. Then we have two tabs that are both English words (Learn/Reference) which looks nicer.


gaearon:
lets go

PR #5400, opened 2022-12-25
[Beta] APIs -> Reference

gaearon:
The "Learn" section is structured like a course. It works the best with sequential access, even though in practice people likely will land on pages randomly.

The "APIs" section is structured like a reference. It's random-access: you start with some API as an entry point and then dig deeper. You can stop at examples or follow the "Learn" links that explain topics thoroughly.

I don't like "APIs" much as the top-level section title. Beginners don't know what an "API" is and they get confused whether this has something to do with AJAX APIs they're calling, or something else. The difference between "Learn" and "APIs" is a bit confusing (am I not "learning" the APIs?) The `/apis/` path is awkward (Next.js won't let us use `/api/` or idk how to do it). We now have a page like `/apis/react/apis` due to the "react: APIs". "APIs -> react: APIs" is weird.

Early on, we used Reference but changed it because the Usage section came first, and if the section is named Reference, then people would complain "this doesn't look like a reference". But we changed that in https://github.com/reactjs/reactjs.org/pull/5399 so we can just call the whole thing a reference. And then all the examples and explanations below are a nice bonus and act as entry points to the rest of the site. Learn/Reference tab names make sense to me based on purpose. Did you come here to learn the thing, or to look something up.

I updated static redirects for beta site but dropped redirects for an even earlier url scheme.


gaearon:
Hmm. Another option is "Docs | API". Makes more sense if we want to put more content in the left tab (which we will need to do).

PR #5401, opened 2022-12-26
fix(typos): some small typos

pfongkye:
use name instead of title as in the data provided


pfongkye:
Hello,
When doing this [challenge](https://github.com/reactjs/reactjs.org/blob/main/beta/src/content/learn/rendering-lists.md#list-with-a-separator-list-with-a-separator) and using `Fragment` is returning me `undefined` in the code sandbox:
image

PR #5408, opened 2022-12-27
[Bate] Make it clear that you can return here

zhangenming:
The rest don't need to be executed


gaearon:
I'd like to keep it without a return because a return wouldn't be valid if you had Hooks after that. So we need to show it's okay to _not_ return too.

PR #5426, opened 2022-12-30
MVP: Support Sandpack snippets in TypeScript

eps1lon:
Preview:
- [single-file snippet](https://beta-reactjs-org-git-fork-eps1lon-typescript-demos-fbopensource.vercel.app/learn/keeping-components-pure)
- [multi-file snippet](https://beta-reactjs-org-git-fork-eps1lon-typescript-demos-fbopensource.vercel.app/learn/describing-the-ui)

The selected "target language" (JS or TS) is selected globally and defaults to TS. It's not persisted at the moment but should be (cookie or localStorage?).

Right now we require duplicating codeblocks in TS and JS (Approach A). This was mainly motivated by having code decorations that may not be easy to convert from TS decorations to JS decorations. 

Alternatively (approach B), we would either have the codeblocks in TypeScript (and transpile them to JS automatically) or just JavaScript (i.e. not have a TS version (yet)).

IIRC @gaearon was against formatting the snippets which makes both approaches a bit more tricky since
A. We can't easily determine if TS and JS would be out-of-sync (by checking the transpiled+formatted TS version against the JS version).
B. Just having a TS version would go against formatting since now the JS version is auto-formatted.

TODO (this PR):
- [ ] don't allow switching target language if only JS is available 
- [ ] how to keep in sync (depends on implementation approach A or B)

TODO (not necessarily this PR):
- [ ] persistence
- [ ] track progress (to drive community effort (should start before or after release?))


beta/src/components/MDX/Sandpack/NavigationBar.tsx
```diff
@@ -39,6 +42,43 @@ const getFileName = (filePath: string): string => {
...
+  } = useContext(SnippetTargetLanguageContext);
+
+  const isCurrent = currentSnippetTargetLanguage === snippetTargetLanguage;
+
+  return (
+    <button
+      aria-pressed={isCurrent ? true : undefined}
+      className={classNames(
+        'text-sm text-primary dark:text-primary-dark inline-flex mx-1 border-black',
+        isCurrent && 'border-b-4'
```
eps1lon, line 59: Tried `aria-pressed:border-b-4` but this didn't work. Either I misunderstood that API or it would require a Tailwind bump which seems excessive.

beta/src/components/MDX/Sandpack/SandpackRoot.tsx
```diff
@@ -67,23 +69,32 @@ ul {
...
-  const files = createFileMap(codeSnippets);
+  const {snippetTargetLanguage} = React.useContext(
+    SnippetTargetLanguageContext
+  );
+  const files = createFileMap(codeSnippets, snippetTargetLanguage);
 
   files['/styles.css'] = {
     code: [sandboxStyle, files['/styles.css']?.code ?? ''].join('\n\n'),
-    hidden: !files['/styles.css']?.visible,
+    hidden: true,
```
eps1lon, line 87: Will extract this change which was revealed by less `any` typing.

beta/src/components/MDX/Sandpack/createFileMap.ts
```diff
@@ -40,14 +50,33 @@ export const createFileMap = (codeSnippets: any) => {
...
     {}
   );
+
+  for (const filePath in fileMap) {
+    // Only remove JS files if we have a TS version available.
+    // If no TS version is available we continue to display JS files.
+    // Assuming that if one file is available as TS, every file is.
+    // Assuming a JS version is available all the time.
+    if (snippetTargetLanguage === 'ts' && isJSFile(filePath) && hasTSVersion) {
+      delete fileMap[filePath];
```
eps1lon, line 75: Deleting instead of `hidden: true` since hidden files would still be part of the created CodeSandbox.

harish-sethuraman:
https://github.com/reactjs/reactjs.org/pull/4720 @eps1lon Are you implementing something similar to this?

eps1lon:
> #4720 @eps1lon Are you implementing something similar to this?

I don't think so. My PR just about having the Sandpack snippets in TypeScript. From the PR description in #4720, it doesn't sound like we have any overlap. 

Though #4720 will eventually benefit from this PR since the snippets will have more accurate completions   for snippets written in TypeScript.

gaearon:
It would be interesting to explore a version that compiles from just TS source. For line or character highlights we can use inline syntax like special comments so that they are automatically translated. I guess auto-formatting is the way to go in that case.

eps1lon:
> For line or character highlights we can use inline syntax like special comments so that they are automatically translated.

Do you have an example where we currently use line or character highlights? Just so I can test different implementations.

eps1lon:
@gaearon Now compiles from just the TS version. Produces inconsistent formatting though: 

> Right now we transpile+format the TS version when parsing MDX to produce the JS version. This is nice from a snippet author perspective (just need to write TS) but results in inconsistent formatting (e.g. bracket spacing). Either we accept that, or revisit https://github.com/reactjs/reactjs.org/pull/5192.

beta/plugins/markdownToHtml.js
```diff
@@ -24,6 +26,7 @@ async function markdownToHtml(markdown) {
     .use(images)
     .use(unrwapImages)
     .use(smartyPants)
+    .use(sandpackTargetLanguages)
```
eps1lon, line 29: I don't understand why the plugins are listed in `remarkPlugins` and composed via `.use`. Just `.use` didn't produce any change.

beta/src/pages/[[...markdownPath]].js
```diff
@@ -55,7 +55,7 @@ function reviveNodeOnClient(key, val) {
 
 // ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 // ~~~~ IMPORTANT: BUMP THIS IF YOU CHANGE ANY CODE BELOW ~~~
-const DISK_CACHE_BREAKER = 7;
+const DISK_CACHE_BREAKER = 8;
```
Comment (eps1lon) [inline beta/src/pages/[[...markdownPath]].js:58]: Bumped this for safe measure since I wasn't sure if "ANY CODE BELOW" also includes changes to imported code.

beta/src/content/learn/keeping-components-pure.md
```diff
@@ -49,8 +49,8 @@ React is designed around this concept. **React assumes that every component you
 
 <Sandpack>
 
-```js App.js
-function Recipe({ drinkers }) {
+```ts App.tsx
```
eps1lon, line 52:
```suggestion
```tsx App.tsx
```

beta/src/components/MDX/Sandpack/index.tsx
```diff
@@ -46,25 +59,37 @@ const SandpackGlimmer = ({code}: {code: string}) => (
 );
 
 export default memo(function SandpackWrapper(props: any): any {
-  const codeSnippet = createFileMap(Children.toArray(props.children));
+  const {snippetTargetLanguage} = useContext(SnippetTargetLanguageContext);
+
+  const codeSnippets = Children.toArray(props.children) as React.ReactElement[];
```
eps1lon, line 64: Using same pattern as `beta/src/components/MDX/Sandpack/SandpackRoot.tsx`. Should we pass `files` as a prop instead of recomputing in `beta/src/components/MDX/Sandpack/SandpackRoot.tsx`?

beta/src/components/MDX/Sandpack/NavigationBar.tsx
harish-sethuraman: Yea this is available with `3.2.0` https://github.com/tailwindlabs/tailwindcss/releases/tag/v3.2.0 may be we can bump the package? its a minor bump only?

beta/src/components/MDX/Sandpack/SandpackRoot.tsx
eps1lon: https://github.com/reactjs/reactjs.org/pull/5428

gaearon:
Are there any other sites that do this with good UX? I find it a bit confusing that toggling resets the edits although I get why. It's also confusing I don't see actual type errors if I make a mistake though I assume that's fixable. I do agree that we should probably just enable formatting everywhere and use a consistent Prettier setting.

eps1lon:
> Are there any other sites that do this with good UX? I find it a bit confusing that toggling resets the edits although I get why.

Yeah I haven't thought about it. I think it's mostly an issue if you accidentally change the language. I think this is going to be quite the effort to save edits so that you can restore them when you change back the language.

> It's also confusing I don't see actual type errors if I make a mistake though I assume that's fixable. 

Hopefully just a setting for the Sandbox I need to enable. Will do in this PR. Do we lint the snippets?

gaearon:
> Do we lint the snippets?

Yeah, we load ESLint with code splitting.

>I think this is going to be quite the effort to save edits so that you can restore them when you change back the language.

Yeah I don't think we should save them, but I do think the way it works is a bit confusing. It's not clear if it "translates" from TS to JS or if it resets. I know it resets but I initially thought it would translate what I wrote lol.

eps1lon:
>  I know it resets but I initially thought it would translate what I wrote lol.

I think magically "upcasting" the JS+edits to TS+edits is out-of-scope (everybody switching to TS needs that 😉 ). For pristine snippets we don't have to do anything though.

When there are edits, we should probably ask if the user wants to keep the edits (potentially causing type-errors) or start fresh with the TypeScript code from the snippet. Not sure if that's easily doable (I remember something about prompts from iframes being deprecated?).

Though what do we do when the user resets a snippet that started as JS, had edits and then switched to TypeScripts? Reset to JS or reset to TS?

gaearon:
I think we'd need a bit of designer help on this. We're busy with the landing but remind me to ask for help after.

PR #5428, opened 2022-12-31
Improve type-safety when using `createFileMap`

eps1lon:
Extracted from https://github.com/reactjs/reactjs.org/pull/5426

### DIFF
--- a/beta/src/components/MDX/Sandpack/SandpackRoot.tsx
+++ b/beta/src/components/MDX/Sandpack/SandpackRoot.tsx
@@ -74,7 +74,7 @@ function SandpackRoot(props: SandpackProps) {
 
   files['/styles.css'] = {
     code: [sandboxStyle, files['/styles.css']?.code ?? ''].join('\n\n'),
-    hidden: !files['/styles.css']?.visible,
+    hidden: !files['/styles.css']?.hidden,
   };
 
   return (
--- a/beta/src/components/MDX/Sandpack/index.tsx
+++ b/beta/src/components/MDX/Sandpack/index.tsx
@@ -46,20 +46,20 @@ const SandpackGlimmer = ({code}: {code: string}) => (
 );
 
 export default memo(function SandpackWrapper(props: any): any {
-  const codeSnippet = createFileMap(Children.toArray(props.children));
+  const codeSnippets = Children.toArray(props.children) as React.ReactElement[];
+  const files = createFileMap(codeSnippets);
 
   // To set the active file in the fallback we have to find the active file first.
   // If there are no active files we fallback to App.js as default.
-  let activeCodeSnippet = Object.keys(codeSnippet).filter(
+  let activeCodeSnippet = Object.keys(files).filter(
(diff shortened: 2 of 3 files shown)


beta/src/components/MDX/Sandpack/index.tsx
```diff
@@ -46,20 +46,20 @@ const SandpackGlimmer = ({code}: {code: string}) => (
 );
 
 export default memo(function SandpackWrapper(props: any): any {
-  const codeSnippet = createFileMap(Children.toArray(props.children));
+  const codeSnippets = Children.toArray(props.children) as React.ReactElement[];
```
eps1lon, line 49: Using same pattern as `beta/src/components/MDX/Sandpack/SandpackRoot.tsx`. Should we pass `files` as a prop instead of recomputing in `beta/src/components/MDX/Sandpack/SandpackRoot.tsx`?

beta/src/components/MDX/Sandpack/SandpackRoot.tsx
```diff
@@ -74,7 +74,7 @@ function SandpackRoot(props: SandpackProps) {
 
   files['/styles.css'] = {
     code: [sandboxStyle, files['/styles.css']?.code ?? ''].join('\n\n'),
-    hidden: !files['/styles.css']?.visible,
+    hidden: !files['/styles.css']?.hidden,
```
eps1lon, line 77: `files` used to be `any` since `createFileMap` returned `any`. Now that we type the input of `createFileMap`, TypeScript is able to infer a more accurate return type of `createFileMap`.

eps1lon, line 77: Couldn't find neither `css styles.css hidden` nor `css styles.css visible` so maybe this should be 
```suggestion
    hidden: true,
```
which would match the old behavior?

eps1lon, line 77:
```suggestion
    hidden: files['/styles.css']?.hidden,
```

beta/src/components/MDX/Sandpack/SandpackRoot.tsx
```diff
@@ -74,7 +74,7 @@ function SandpackRoot(props: SandpackProps) {
 
   files['/styles.css'] = {
     code: [sandboxStyle, files['/styles.css']?.code ?? ''].join('\n\n'),
-    hidden: !files['/styles.css']?.visible,
+    hidden: files['/styles.css']?.hidden,
```
eps1lon, line 77: `files` used to be `any` since `createFileMap` returned `any`. Now that we type the input of `createFileMap`, TypeScript is able to infer a more accurate return type of `createFileMap`.

eps1lon, line 77: Couldn't find neither `css styles.css hidden` nor `css styles.css visible` so maybe this should be 
```suggestion
    hidden: true,
```
instead which would match the old behavior?

PR #5435, opened 2023-01-03
Updating what I think is a minor typo

rahulrao0209:
Replacing the word pending with previous on the useState API reference page. I think the author meant to use previous as the content is regarding the updater function for setState which receives the previous state value as an argument.


harish-sethuraman (commented):
I think this is correct 🤔 I see the same in this subsection too https://beta.reactjs.org/reference/react/useState#updating-state-based-on-the-previous-state

gaearon:
Pending is more precise.

rahulrao0209:
Deleting branch.

PR #5437, opened 2023-01-05
Fix for ReactDOM.createRoot() example

chalder:
Fixing the code for ReactDOM.createRoot() example

### DIFF
--- a/beta/src/content/reference/react-dom/client/createRoot.md
+++ b/beta/src/content/reference/react-dom/client/createRoot.md
@@ -24,7 +24,7 @@ Call `createRoot` to create a React root for displaying content inside a browser
 
 ```js
 const domNode = document.getElementById('root');
-const root = createRoot(domNode);
+const root = ReactDOM.createRoot(domNode);
 ```
 
 React will create a root for the `domNode`, and take over managing the DOM inside it. After you've created a root, you need to call [`root.render`](#root-render) to display a React component inside of it:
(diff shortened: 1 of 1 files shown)


beta/src/content/reference/react-dom/client/createRoot.md
```diff
@@ -24,7 +24,7 @@ Call `createRoot` to create a React root for displaying content inside a browser
 
 ```js
 const domNode = document.getElementById('root');
-const root = createRoot(domNode);
+const root = ReactDOM.createRoot(domNode);
```
eps1lon, line 27: Definitely agree that we should be specific where `createRoot` is coming from.

We're not using that access in any other code examples on this page though.

Instead, like for [`createRef`](https://beta.reactjs.org/reference/react/createRef), let's show how import it by adding `import { createRoot } from 'react-dom/client'` at the beginning of that snippet.

There are other API reference pages where this is missing as well. So it'd be awesome if you could apply similar changes to other pages as well. Not required to get this merged though.

eps1lon (changes requested):

beta/src/content/reference/react-dom/client/createRoot.md
chalder: Using it as  `ReactDOM.createRoot(domNode)` or having import statement `import { createRoot } from 'react-dom/client'` both works. 
I was reading [old doc]( https://reactjs.org/docs/rendering-elements.html ) and [new doc](https://beta.reactjs.org/reference/react-dom/client/createRoot) in parallel . Added this change similar to old doc.

Thanks for you comment! Will raise PRs as I see issues while reading.

gaearon:
I did the rest in https://github.com/reactjs/reactjs.org/pull/5442.

PR #5440, opened 2023-01-06
Fixing error in the sample code

siddtheone:
`e` is not defined.


gaearon:
I checked the gist, and it's defined in that file. What do you mean? I agree it's unnecessary though and we should just show createElement but that's a separate topic.

siddtheone:
I can't find it in any of the 3 code snippets
`const e = React.createElement;`

image

siddtheone:
I saw that it's already part of `the starter code`
Closing the request.

PR #5449, opened 2023-01-08
fix link

domscript:
I think it should be like this!
to fix link error:
[video to show error](https://youtu.be/mUnit2IzELk)


harish-sethuraman (approved):
Thanks

PR #5452, opened 2023-01-09
Update docs with info about onError

votemike:
Closes: https://github.com/facebook/react/issues/25961

onError for the picture element works differently in React to Vanilla HTML

### DIFF
--- a/content/docs/reference-events.md
+++ b/content/docs/reference-events.md
@@ -406,6 +406,8 @@ Event names:
 onLoad onError
 ```
 
+Note: When using onError with the `<picture>` element, onError can be used on either `<picture>` or `<img>` (not `<source>`).
+
 * * *
 
 ### Animation Events {#animation-events}


eps1lon (commented):
From the linked issue it sounds like the issue is only about bubbling. But this talks about `source` not being supported.

Could you clarify what the take-away from the issue was? I also left a note in the issue since I had trouble reproducing.

votemike:
> From the linked issue it sounds like the issue is only about bubbling. But this talks about `source` not being supported.
> 
> Could you clarify what the take-away from the issue was? I also left a note in the issue since I had trouble reproducing.

If following vanilla specs while creating a React app, a user will not successfully trigger the onerror of a `<picture>`.
If the difference between vanilla HTML/JS and React is desired due to the way React bubbles these events, the "React way" to trigger onerror should be documented.

PR #5453, opened 2023-01-09
Update meetups.md

bmbaron:
add Reactplus in Hanoi, Vietnam to meetups list


bnzone (approved):

rickhanlonii:
Sorry I missed this, looks like the meetup group is missing now. If there's an active link let me know and we can add it.

PR #5455, opened 2023-01-11
fix: index out of bounds of sculptureList

regchiu:
Add check in `handleNextClick` function to avoid errors.


harish-sethuraman:
> https://github.com/reactjs/reactjs.org/issues/3308#issuecomment-1038641042

It is intentional since first challenge on the page state a component's memory is based on fixing this. :)

https://beta.reactjs.org/learn/state-a-components-memory#challenges

regchiu:
At first I also thought it was intentional, but I didn’t see the context, it turned out that the teaching on another page used this example.

Thank you.

PR #5463, opened 2023-01-15
Update how-to-contribute.md

ggold7046:
Fixed typo.


eps1lon (changes requested):
[releasable is fine](https://en.wiktionary.org/wiki/releasable#:~:text=Capable%20of%2C%20or%20suitable%20for%2C%20release.)

PR #5479, opened 2023-01-21
corrected the faulty code of React documentation

KarkiAnmol:
image

1. I changed the way the Avatar component receives its props from being passed as an object props to destructuring the user object {user}. This change makes the component more readable and easy to understand. It also allows us to directly access the properties of the user object, such as avatarUrl and name, without having to use the props object. This improves the component's performance and maintainability, making it easier to understand and update in the future. Additionally, it makes the component code more consistent with the rest of the codebase, where similar destructuring is used throughout.

image
2. I also made a similar change to the UserInfo component, where I changed the way the component receives its props from being passed as an object props to destructuring the user object {user}. This change makes the component more readable and easy to understand, allowing us to directly access the properties of the user object, such as name, without having to use the props object. This improves the component's performance and maintainability, making it easier to understand and update in the future. Additionally, it makes the component code more consistent with the rest of the codebase, where similar destructuring is used throughout. This change also aligns the UserInfo component with the Avatar component, making it easier to understand the relationship between the two components.

### DIFF
--- a/content/docs/components-and-props.md
+++ b/content/docs/components-and-props.md
@@ -154,11 +154,11 @@ This component can be tricky to change because of all the nesting, and it is als
 First, we will extract `Avatar`:
 
 ```js{3-6}
-function Avatar(props) {
+function Avatar({user}) {
   return (
     <img className="Avatar"
-      src={props.user.avatarUrl}
-      alt={props.user.name}
+      src={user.avatarUrl}
+      alt={user.name}
     />
   );
 }
@@ -194,12 +194,12 @@ function Comment(props) {
 Next, we will extract a `UserInfo` component that renders an `Avatar` next to the user's name:
 
 ```js{3-8}
-function UserInfo(props) {
+function UserInfo({user}) {
   return (
     <div className="UserInfo">
-      <Avatar user={props.user} />
+      <Avatar user={props.author} />
       <div className="UserInfo-name">
-        {props.user.name}
+        {user.name}
       </div>
     </div>
   );


KarkiAnmol:
I have completed the Contributor License Agreement (CLA) check and have been approved by Vercel.

KarkiAnmol:
i have already authorized Vercel to deploy but why is the check still failing?

KarkiAnmol:
image
My repository shows that the vercel deployment has been completed,but,somehow there is still 1 failing check of Vercel Authorization.

content/docs/components-and-props.md
```diff
@@ -154,11 +154,11 @@ This component can be tricky to change because of all the nesting, and it is als
 First, we will extract `Avatar`:
 
 ```js{3-6}
-function Avatar(props) {
+function Avatar({user}) {
```
harish-sethuraman, line 157: We have not used de structuring in the page already? There are other places where we usually don't use destructuring as well.

content/docs/components-and-props.md
```diff
@@ -194,12 +194,12 @@ function Comment(props) {
 Next, we will extract a `UserInfo` component that renders an `Avatar` next to the user's name:
 
 ```js{3-8}
-function UserInfo(props) {
+function UserInfo({user}) {
   return (
     <div className="UserInfo">
-      <Avatar user={props.user} />
+      <Avatar user={props.author} />
```
harish-sethuraman, line 200: Since you have already removed props from params and destructured it we won't have `props`? 
PS: currently we are building beta.reactjs.org you can also use that.

PR #5480, opened 2023-01-21
corrected the faulty code of React documentaion

KarkiAnmol:
image

1. I changed the way the Avatar component receives its props from being passed as an object props to destructuring the user object {user}. This change makes the component more readable and easy to understand. It also allows us to directly access the properties of the user object, such as avatarUrl and name, without having to use the props object. This improves the component's performance and maintainability, making it easier to understand and update in the future. Additionally, it makes the component code more consistent with the rest of the codebase, where similar destructuring is used throughout.

image

2. I also made a similar change to the UserInfo component, where I changed the way the component receives its props from being passed as an object props to destructuring the user object {user}. This change makes the component more readable and easy to understand, allowing us to directly access the properties of the user object, such as name, without having to use the props object. This improves the component's performance and maintainability, making it easier to understand and update in the future. Additionally, it makes the component code more consistent with the rest of the codebase, where similar destructuring is used throughout. This change also aligns the UserInfo component with the Avatar component, making it easier to understand the relationship between the two components.

### DIFF
--- a/content/docs/components-and-props.md
+++ b/content/docs/components-and-props.md
@@ -154,11 +154,11 @@ This component can be tricky to change because of all the nesting, and it is als
 First, we will extract `Avatar`:
 
 ```js{3-6}
-function Avatar(props) {
+function Avatar({user}) {
   return (
     <img className="Avatar"
-      src={props.user.avatarUrl}
-      alt={props.user.name}
+      src={user.avatarUrl}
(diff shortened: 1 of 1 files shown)


KarkiAnmol:
I approved Vercel but it's change was not refleceted on the list of unsuccessful check.Hence i am closing this request.

PR #5485, opened 2023-01-21
update last challenge to be be a better use of ref

Spoof14:


src/content/learn/referencing-values-with-refs.md
```diff
@@ -616,33 +616,26 @@ export default function Chat() {
...
+      alert('Sending: ' + textRef.current.value);
     }, 3000);
   }
 
   return (
     <>
       <input
-        value={text}
-        onChange={handleChange}
+        value={textRef}
```
Spoof14, line 638:
```suggestion
        ref={textRef}
```

PR #5511, opened 2023-01-27
Upgrade react-collapsed to @collapsed/react

roginfarrer:
This fork was created while react-collapsed didn't support React 18's `useId`. I've since updated that library to `@collapsed/react` and it now uses `useId`!

### DIFF
--- a/beta/.npmrc
+++ b/beta/.npmrc
@@ -1,0 +1,1 @@
+registry=https://registry.yarnpkg.com
--- a/beta/package.json
+++ b/beta/package.json
@@ -23,6 +23,7 @@
   },
   "dependencies": {
     "@codesandbox/sandpack-react": "1.15.5",
+    "@collapsed/react": "^4.0.1",
     "@docsearch/css": "3.0.0-alpha.41",
     "@docsearch/react": "3.0.0-alpha.41",
     "@headlessui/react": "^1.7.0",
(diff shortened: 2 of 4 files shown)


gaearon:
I've wanted to update but I noticed there was a significant rewrite, and I'm not sure the approach quite makes sense to me. In particular I found it difficult to tell if React's rules are being followed there. For example, [this](https://github.com/roginfarrer/collapsed/blob/15af3d556cee9391dd1a9412cc4dcb047321bafe/packages/react/src/useCollapse.ts#L43) is a mutation during render. That's not supported in React. Similarly, all `this.get*` functions read from mutable fields, which is not supported during rendering. This doesn't mean something in particular is broken — but more that it's *hard* to say whether this will work or not.

For a more concrete blocker, I'm seeing a layout jump on the initial load. The SSR'd version of the page appears to default to an expanded (rather than a collapsed) state. This is likely a bug.

roginfarrer:
Appreciate the feedback, Dan. I see the bugs you raised. I may have missed the forest for the trees when trying to rewrite Collapse to be compatible with more libraries. I'll see if I can address the issues, but will reconsider this rewrite if I can't make it a first-class citizen for React.

gaearon:
For this type of code, it's valuable to try to separate pure rendering logic from the parts that deal with DOM directly. Ideally the part that's pure should be driven by React state. Things like "calculate style", "get props for X", etc, should be straightforward pure transformations based on state variables. This should not require reading any mutable fields. Then, you'd want to update that state when needed, and/or keep it in sync with the DOM. For that part, it makes sense to me that you might want the imperative logic outside the component to group it. That logic should only be called from event handlers or Effects. There are some examples of different ways you can approach splitting imperative from pure logic [in this section](https://beta.reactjs.org/learn/reusing-logic-with-custom-hooks#there-is-more-than-one-way-to-do-it).

PR #5535, opened 2023-02-01
Fix Challenge Solutions

chalder:
1. choosing-the-state-structure.md > added text " should not" . That's what makes sense.
2. state-a-components-memory.md > removing import of useState since we are using regular variable here


chalder:
@gaearon please review.

PR #5544, opened 2023-02-04
`Conferences`: Move past conferences to past section

NickFoden:
### Desc:
- move existing conferences that have passed down to past section

### DIFF
--- a/content/community/conferences.md
+++ b/content/community/conferences.md
@@ -12,20 +12,6 @@ Do you know of a local React.js conference? Add it here! (Please keep the list c
 
 ## Upcoming Conferences {#upcoming-conferences}
 
-### React Global Online Summit 22.2 by Geekle {#react-global-v2-2022}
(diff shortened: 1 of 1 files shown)


content/community/conferences.md
```diff
@@ -64,6 +50,21 @@ Oct 5 - 7, 2023. In-person in Goa, India (hybrid event) + Oct 3 2023 - remote da
 
 ## Past Conferences {#past-conferences}
 
+### React Global Online Summit 22.2 by Geekle {#react-global-v2-2022}
```
harish-sethuraman, line 53: Can you re arrange the same in the reverse order? so that the last conference is the first in past conferences? older ones are that way?

NickFoden: Hi @harish-sethuraman good shout, updated now. 

Maybe will raise another pr where code can handle the ordering so this doesn't fall out of date on its own ?

harish-sethuraman (approved):

PR #5547, opened 2023-02-05
Fix error in multiple inputs section of forms.md

Chriserus:
This PR fixes an error in Handling Multiple Inputs section, where previous state wasn't taken into consideration on state change resulting in undefining every other field.

### DIFF
--- a/content/docs/forms.md
+++ b/content/docs/forms.md
@@ -228,6 +228,7 @@ class Reservation extends React.Component {
     const name = target.name;
 
     this.setState({
+      ...state,
       [name]: value
     });
   }


content/docs/forms.md
```diff
@@ -228,6 +228,7 @@ class Reservation extends React.Component {
     const name = target.name;
 
     this.setState({
+      ...state,
```
lunaleaps, line 231: This isn't necessary. `setState` will take an object will merge that change to the underlying state. This is different with the `setState` that takes in a reducer function. See docs: https://react.dev/reference/react/Component#setstate-parameters

Chriserus: With all due respect, did you check it by running current version of `setState` following this doc? The origin of this PR is a confusing moment when following the documentation with a fresh react dev. As described in the PR, it undefined every other field. If it's possible for you to verify, I would be grateful. Thank you in advance.

PR #5565, opened 2023-02-08
Adding WebdriverIO as recommended tool for component testing

christian-bromann:
The current list of recommend tools is a bit behind of what is currently possible in the testing ecosystem. Many browser automation frameworks, including [WebdriverIO](https://webdriver.io/) now offer component testing capabilities which might be a better alternative for many users given it comes much closer to the real world with just minimal speed detriments. For pure component tests I personally would not advice to use JSDOM at all given that:

- it lacks in certain features (e.g. triggering pseudo states) and has some [known caveats](https://github.com/jsdom/jsdom#caveats)
- it completely ignores the interactivity state of elements, e.g. if my component hides a button or an input is disabled, many user interactions are still possible within JSDOM and non-organic interactions via JS clicks
- mounting and rendering components in an actual browser allows the use of common developer debugging tools like DevTools rather than using e.g. `screen.debug()` to print the DOM tree providing a much better DX
- speed advancements are just minimal, starting a headless browser is almost as fast

I know that a lot of folks are working on the new React docs and I asked a question whether there will be a testing section for it as well, I am happy to contribute to that too.


christian-bromann:
Closing in favor of #5675. I forgot I had this open already but like the wording of the new PR.

Issue #5588, opened 2023-02-11
[Beta Docs]: overflow of the sidebar nav


TAKANOME-DEV:
When visiting the react beta docs on Firefox, I noticed that the sidebar overflows and hides some texts of the main layout

![react_beta_docs](https://user-images.githubusercontent.com/79809121/218256071-b6096ce1-fa9b-4b6f-9ba2-fb42041ff32f.png)

hugostanley:
Can be replicated in 1366px width screen size.

mkotik:
unable to replicate this problem

hugostanley:
> unable to replicate this problem

Ways to replicate:

1. Open beta docs specifically on Firefox
2. Change screen size to >= 1024px

PR #5591, opened 2023-02-14
fix(styles): change sidebar width in Firefox to avoid overlapping on

hugostanley:
#5588 

In beta docs, the sidebar overlaps the content when min-width of screen is 1024px. This happens on Firefox only. 

Added a conditional media query depending on browser.

### DIFF
--- a/beta/src/components/Layout/Nav/Nav.tsx
+++ b/beta/src/components/Layout/Nav/Nav.tsx
@@ -101,10 +101,17 @@ export default function Nav({
 }) {
   const [isOpen, setIsOpen] = useState(false);
   const [showFeedback, setShowFeedback] = useState(false);
+  const [isFirefox, setIsFirefox] = useState(false);
   const scrollParentRef = useRef<HTMLDivElement>(null);
   const feedbackAutohideRef = useRef<any>(null);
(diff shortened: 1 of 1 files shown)


gaearon:
we should figure out the actual cause i think and fix that instead.

beta/src/components/Layout/Nav/Nav.tsx
```diff
@@ -101,11 +101,18 @@ export default function Nav({
...
   const [showFeedback, setShowFeedback] = useState(false);
+  const [isFirefox, setIsFirefox] = useState(false);
   const scrollParentRef = useRef<HTMLDivElement>(null);
   const feedbackAutohideRef = useRef<any>(null);
   const {asPath} = useRouter();
   const feedbackPopupRef = useRef<null | HTMLDivElement>(null);
 
+  useEffect(() => {
+    if (typeof window !== undefined) {
+      setIsFirefox(window.navigator.userAgent.indexOf('Firefox') > -1);
```
gaearon, line 112: if you do this in an effect, the issue would still exist until hydration. then, by setting state, it would cause an extra re-render of the nav. seems not ideal. we should fix the CSS ideally.

PR #5601, opened 2023-02-17
Forgotten import statement for `createElement`

MarkE16:
Added `createElement` in the import statement since it was used within the code snippet but was not imported in the API Reference for `isValidElement`.


eps1lon (approved):
thanks

Issue #5606, opened 2023-02-19
Bug: React Docs Beta - Keeping Components Pure


PradipKafleDev:
Inside Keeping Components Pure you will find a topic Side Effects: (un)intended consequences, in the first example where it shows component that break this rule , the output produce by given example is not correct it return 
Tea cup for guest #1 
Tea cup for guest #2
Tea cup for guest #3 not
Tea cup for guest #2
Tea cup for guest #4
Tea cup for guest #6

yousefelgoharyx:
Could you explain a bit more
Provide some screenshots maybe

PR #5607, opened 2023-02-19
Bug: Function return different output as compared to the output that is shown in React Docs Beta

PradipKafleDev:
React doc beta--> keeping components pure--> side effects:(un)intended consequences : Previously the value that was added with guest was 1(guest=guest+1)  and which results different output as compare to the output image thats  shows in react beta doc  , changing the guest value to 2 , (guest=guest+2) will return the desired output as shown in the output image in react beta doc

#5606

### DIFF
--- a/beta/src/content/learn/keeping-components-pure.md
+++ b/beta/src/content/learn/keeping-components-pure.md
@@ -98,7 +98,7 @@ let guest = 0;
 
 function Cup() {
   // Bad: changing a preexisting variable!
-  guest = guest + 1;
+  guest = guest + 2;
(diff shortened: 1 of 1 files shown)


PradipKafleDev:
The output produce by the TeaSet() shows in react beta doc==>keeping components pure==>Side Effects: (un)intended consequences  is :
Tea cup for guest #2
Tea cup for guest #4
Tea cup for guest #6 , 
But which is not correct  because the guest variable is incremented buy +1 so in order to to get 2,4,6 we should add +2  with guest variable.

PR #5608, opened 2023-02-20
docs(cloneElement.md): fix a wrong result in a comment

pobch:


harish-sethuraman (approved):

harish-sethuraman (commented):
@gaearon Just wanted to confirm if this is the case? I see `<Row title="Cabbage" isHighlighted={true}>Goodbye</Row>` in the console too. The [old doc says](https://reactjs.org/docs/react-api.html#cloneelement) second param is config

PR #5614, opened 2023-02-21
Update conferences.md - React Rally

mikelnorth:
Added React Rally 2023 conference to upcoming conferences list both current and beta docs


mikelnorth:
Closed due to newer PR

PR #5617, opened 2023-02-22
fix typo

lonr:
Added a missing parameter of `hydrateRoot`.


lonr:
Close as it has already been resolved.

PR #5620, opened 2023-02-23
Update create-a-new-react-app.md

Adarshkumar76:


Adarshkumar76:
It is a very importent for biggners

Issue #5597, opened 2023-02-16
Using the structuredClone() for updating a nested object


IvanMilosav:
https://beta.reactjs.org/learn/updating-objects-in-state#updating-a-nested-object

I think that deep cloning object with structuredClone() and putting the clone into setPerson() is better solution for the problem than using the spread operator.

The the example that was provided can easily be changed this way:
```js
personCopy = structuredClone(person);
personCopy.artwork.city = "New Delhi";
setPerson(personCopy);
 ```

This can help avoiding the mistake when spread operator is used to deep clone objects. See example below:
```js
const original = { a: { b: 1 } };
const falseCopy = { ...original };

falseCopy.a.b = 2;

console.log(falseCopy); // logs {a: {b: 2}}
console.log(original); // also logs {a: {b: 2}}
 ```

I am sorry if I have failed to follow contributing guidelines. I am aware that the second example is not the best representation of my point.

yousefelgoharyx:
Using a library like Immer would be better since structuredClone creates a deep copy of the whole object unlike Immer which does structural sharing to increase performance.

IvanMilosav:
I agree, but it requires additional library.

PR #5642, opened 2023-03-02
Add the structuredClone method example on the Updating Objects in State page.

paterson1720:
In addition of using the spread operator to update state objects, we can use the global structuredClone method of JavaScript to make a copy of the state, mutate the copy, and set the state with the new object.
In this PR I added a section to show an example using the structuredClone method.

Screenshot of added section:
Screen Shot 2023-03-02 at 17 33 02


IvanMilosav:
I suggested this too, you can check it out here https://github.com/reactjs/react.dev/issues/5597

PR #5657, opened 2023-03-07
Re-wording static-type-checking.md

behrangsa:
With the assistance of ChatGPT:

:person_bald: Transfer "Static type checkers like Flow and TypeScript identify certain types of problems before you even run your code. " to the style of Bertrand Russel

> :robot: Static type checkers such as Flow and TypeScript possess the capability to apprehend specific types of quandaries prior to code execution, as they are equipped with the faculty of detecting irregularities in the typing system.

### DIFF
--- a/content/docs/static-type-checking.md
+++ b/content/docs/static-type-checking.md
@@ -4,7 +4,7 @@ title: Static Type Checking
 permalink: docs/static-type-checking.html
 ---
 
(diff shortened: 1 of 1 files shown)


eps1lon:

behrangsa:
@eps1lon It's not a big deal but "before you even run" has a redundant _even_, however "prior to code execution" does not have any unnecessary words.

PR #5660, opened 2023-03-08
Fix a misleading style in the example

kkshinkai:
The document said:

> To check that your fix worked, try changing the value of `imageSize` to `'b'`. The image should resize after your edit.

If the height is restricted, readers may be confused if they do not see any changes after adjusting the `imageSize`.

See https://beta.reactjs.org/learn/javascript-in-jsx-with-curly-braces#write-an-expression-inside-jsx-curly-braces

### DIFF
--- a/beta/src/content/learn/javascript-in-jsx-with-curly-braces.md
+++ b/beta/src/content/learn/javascript-in-jsx-with-curly-braces.md
@@ -469,7 +469,7 @@ export default function TodoList() {
 ```css
 body { padding: 0; margin: 0 }
 body > div > div { padding: 20px; }
-.avatar { border-radius: 50%; height: 90px; }
+.avatar { border-radius: 50%; }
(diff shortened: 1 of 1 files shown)


kkshinkai:
@facebook-github-bot Check CLA again (?

harish-sethuraman (approved):
Makes sense. Thanks

gaearon:
Please don't change styles in the *question* without also changing them in the *solutions*. It's inconsistent now.

gaearon:
Fixed in https://github.com/reactjs/reactjs.org/commit/19aa5b4852c3905757edb16dd62f7e7506231210.

(Appreciate the fix! This is more of a note to @harish-sethuraman — please be careful :)

harish-sethuraman:
Totally a miss on my side. Will make sure next time :)

PR #5664, opened 2023-03-10
docs: add missing import statement

skoyah:
The current Beta documentation shows a code snippet a little different from the one available in the Sandbox.

It seems that the `React` import is missing from the documentation.

Steps to reproduce:

- Fork the first code sandbox that appears on the tutorial on the "Tutorial Setup" section (https://beta.reactjs.org/learn/tutorial-tic-tac-toe#setup-for-the-tutorial)

Screenshot 2023-03-10 at 18 47 17

Screenshot 2023-03-10 at 20 05 06

This PR fixes the documentation to be in sync with the sandbox example.

### DIFF
--- a/beta/src/content/learn/tutorial-tic-tac-toe.md
+++ b/beta/src/content/learn/tutorial-tic-tac-toe.md
@@ -330,7 +330,7 @@ Click on the file labeled `styles.css` in the _Files_ section of CodeSandbox. Th
 Click on the file labeled `index.js` in the _Files_ section of CodeSandbox. You won't be editing this file during the tutorial but it is the bridge between the component you created in the `App.js` file and the web browser.
 
 ```jsx
-import {StrictMode} from 'react';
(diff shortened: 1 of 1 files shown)


harish-sethuraman:
These react imports [might not be needed](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) 🤔. constantly throughout the documentation we don't have react imported I guess.

skoyah:
I come from Vueland and is my first contact with React. Started following the tutorial and encountered this "error".

harish-sethuraman:
It should work even if we remove the import of `React` in the codesandbox and it should be fine.

shaonkabir8 (commented):
It's great and working fine with several tests ... 
keep it up mate @skoyah

skoyah:
So perhaps we should remove it from the sandbox itself?

Having that import statement does not add anything.

Beginners might even wonder why is there in the first place.

PR #5681, opened 2023-03-15
docs: fix typo in memo.md

Akatroj:
Fix typo - `React.memo` is an optimization **hint**.


sophiebits:
Thanks for sending but I think this is correct as is.

PR #5703, opened 2023-03-16
Update conferences.md - React Miami

michellebakels:


vjeux:
Thank you! Super excited about attending React Miami myself <3

Issue #5710, opened 2023-03-17
useEffect example could use something different like scrollY or scrollX.


avsssai:
The useEffect example in the "Synchronizing with Effects" could use a different example, maybe like a window.scrollX or a window.scrollY instead of e.clientX and e.clientY as they don't exist on a scroll event and cause it to show undefined on scroll event.

`useEffect(() => {
  function handleScroll(e) {
    console.log(e.clientX, e.clientY);
  }
  window.addEventListener('scroll', handleScroll);
  return () => window.removeEventListener('scroll', handleScroll);
}, []);
`

![image](https://user-images.githubusercontent.com/34810623/225854498-24a0afb6-5a4e-4188-b7c9-3170cdc5cefd.png)

PR #5711, opened 2023-03-17
docs: changed useEffect example

avsssai:
The useEffect example in "Synchronizing with Effects" gives undefined as a MouseEvent is being logged for a scroll event trigger .
This is in the section [Subscribing to events](https://react.dev/learn/synchronizing-with-effects#subscribing-to-events)

I have changed it to window.scrollX and window.scrollY, so they output some value instead of undefined.

This is for the issue - [#5710 ](https://github.com/reactjs/reactjs.org/issues/5710)

image


sophiebits:
Thank you!

PR #5721, opened 2023-03-17
Improve anchor link

oliviertassinari:
The current value is broken, the header has the `.h-16` class name which equals to 4rem.

https://react.dev/reference/react/useDeferredValue#showing-stale-content-while-fresh-content-is-loading

**Before**
Screenshot 2023-03-18 at 00 46 41

**After**
Screenshot 2023-03-18 at 00 46 53

---

For this test, I used:

Screenshot 2023-03-18 at 00 48 03

but it equally improve the experience with the default value (16px for the default font size)

### DIFF
--- a/beta/src/styles/index.css
+++ b/beta/src/styles/index.css
@@ -239,7 +239,7 @@ html.dark .code-step * {
 }
 
 .mdx-heading {
-  scroll-margin-top: 3em;
+  scroll-margin-top: calc(4rem + 20px);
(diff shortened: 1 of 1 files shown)


gaearon:
Thanks! Don't know how this happened. I thought I fixed this before. :P

oliviertassinari:
I was working on updating all the links in https://mui.com/ this weekend from the old to the new React docs, happy to have solved this small annoyance :)

PR #5728, opened 2023-03-18
fix: spacing in responsive site navigation

safiranugroho:
This fixes the following issue: https://github.com/reactjs/reactjs.org/issues/5725
`::-webkit-scrollbar` offsets its containing element's width in browsers that support it.
This PR overlays the scrollbar so the containing element's width is not affected (and adds the appropriate padding/margin to even out the spacing in mobile).

Chrome (before):
Screenshot 2023-03-19 at 12 33 04 am

Chrome (after):
Screenshot 2023-03-19 at 12 33 15 am

Firefox (before):
Screenshot 2023-03-19 at 12 32 24 am

Firefox (after):
Screenshot 2023-03-19 at 12 32 27 am

### DIFF
--- a/beta/src/components/Layout/Sidebar/SidebarRouteTree.tsx
+++ b/beta/src/components/Layout/Sidebar/SidebarRouteTree.tsx
@@ -148,12 +148,12 @@ export function SidebarRouteTree({
                 {index !== 0 && (
                   <li
                     role="separator"
-                    className="mt-4 mb-2 ml-5 xs:mx-5 border-b border-border dark:border-border-dark"
+                    className="mt-4 mb-2 mx-5 sm:mr-0 border-b border-border dark:border-border-dark"
                   />
                 )}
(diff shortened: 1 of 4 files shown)


harish-sethuraman:
In safari Im still getting the scrollbar with this branch preview? 

Im curious why you did not do you might want to do ` -webkit-appearance: none;` for `-webkit-scrollbar`?

safiranugroho:
> 

i didn’t write the impl. for the scrollbar, but i assume the idea is to set a transparent background color instead of hiding it completely. 

in the source code `scrollbar-color` is defined, but since only Firefox supports that property,  `-webkit-scrollbar` is used as fallback to achieve the same behavior (set transparent background color).

my PR is to fix the fact that `-webkit-scrollbar` offsets the width of its containing div, bc i didn’t want to disturb the original implementation.

PR #5744, opened 2023-03-19
chore: delete duplicate classname

li-jia-nan:


li-jia-nan:
@gaearon review pls~

li-jia-nan:
@gaearon review pls~

li-jia-nan:
@sophiebits review pls~

Issue #5727, opened 2023-03-18
Scrolling issue on https://react.dev/reference/react


emAyush56:
The scroll bar on the left sidebar hides on hovering the cursor or on clicking on it. Following is an attached GIF of the same issue.

![reactdev-scrollissue](https://user-images.githubusercontent.com/29714782/226108170-1251cdce-1cec-4bcc-bb2f-7c62225acac4.gif)

OS: Windows 10
Browser: Chrome. Same issue on Edge. Similar issue on Firefox Developer Edition

PR #5750, opened 2023-03-19
fix scrollbar hiding on hover/click

getusha:
Fixes: https://github.com/reactjs/react.dev/issues/5727

the main layout is overlapping the SideBarNav which is preventing scroll using drag.

before:

https://user-images.githubusercontent.com/75031127/226160054-84dbcedc-b889-4723-af0b-6bbb22e7f4e6.mov

after:

Screenshot 2023-03-19 at 12 18 28 AM


gaearon (changes requested):
This changes the layout. Instead, we should change other things (where needed) but keep the same layout.

pwbriggs:
Any updates on this?

See also: #5898

rickhanlonii:
Fixed by https://github.com/reactjs/react.dev/pull/6586 which doesn't change layout.

PR #2378, opened 2019-10-01
Move 'Handling Events' example from Codepen to CodeSandbox

washingtonsoares:
The main goal of this PR is to move the '**Handling Events**' examples page currently in Codepen to Codesandbox! 🎉 

Preview changes: https://deploy-preview-2378--reactjs.netlify.com/docs/handling-events.html 
image

As can be seen from this PR, there is only one JS file that is imported into `handling-events.md` and is the same file that will be sent to CodeSandbox. So far so good, and the idea sounds great!

However there is a problem: Comments that are used to highlight/hide lines are also sent to CodeSandbox as shown below.:
image*Highlight/hideline on CodeSandbox*.

That seems a little strange to me. "Messing" the code in the examples doesn't look cool.
In that case, would it be interesting for me to adapt the `gatsby-remark-code-repls` plugin (or some other plugin) to remove comments that highlight/hide the lines before sending to CodeSandbox?

Despite what I did working, there is another approach. Like @cyan33 did in #913, there is a [specific file](https://github.com/reactjs/reactjs.org/pull/913/files#diff-a3b8b2661de6bd21ba4e614fbec5f329) for the markdown, and [other specific](https://github.com/reactjs/reactjs.org/pull/913/files#diff-07a5cdb642e870e6bd008f46d8f956ed) to be sent to CodeSandbox.
This approach does not have the problem of "messing up" CodeSandbox examples with highlight/hideline comments, however you will need to maintain two files.

This PR brings the approach I found most interesting and would like to hear some opinions on what makes the most sense.

cc @cyan33 @bvaughn @gaearon

PR #5758, opened 2023-03-19
fix: ui elements shifting to the right

phanendraguptha:
The main layout elements are shifting to the right when the search bar is open

before:

https://user-images.githubusercontent.com/47347106/226173941-f18d6faa-9565-466b-bab1-151e15017623.mp4

after:

https://user-images.githubusercontent.com/47347106/226173953-6c83189b-fe85-464f-8b08-4714a17702e3.mp4


gaearon:
Is there any more natural way to do this?

Why does the scrollbar disappear underneath the model? Can we fix that instead?

phanendraguptha:
If the scrollbar appears, it means that the main content below the modal can be scrolled, which is not desirable since we disable scrolling when the popup modal opens. This causes the scrollbar to disappear, similar to using `overflow: hidden`.

phanendraguptha:
On further investigation, I found that `@docsearch/react` is adding a class `DocSearch--active` to the body element. Which is making the scrollbar hide and they are using `overflow: hidden` to do that. When we remove that class, the main content below the modal can be scrolled.

phanendraguptha:
Hi,

The issue causing layout shifts due to the scrollbar disappearing has been fixed in the latest version of the Algolia DocSearch package (fix: [#2378](https://github.com/algolia/docsearch/issues/2378)).

To resolve the problem in this repository, please consider bumping the DocSearch version to 3.8.0 (or the latest available). This will ensure the layout remains stable when the search bar modal is opened.

Thank you!

Issue #5738, opened 2023-03-18
Off-by-one when clicking in right menu bar


nh2:
See video:

https://user-images.githubusercontent.com/399535/226140687-48621dc4-ddbe-4730-9e2a-e777a27e6063.mp4

Firefox 110.0

PR #5765, opened 2023-03-19
fix toc to highlight the correct header on click

alexandruvisan19:
Hi, 

If the user clicks one of the topics from the "On this Page" list, it will always highlight the one topic before. By increasing the TOP_OFFSET, this will solve the issue.

Thanks!


alexandruvisan19:
#5738 issue reference

gaearon:
thanks.

PR #5767, opened 2023-03-19
FIX: Use `typos` tool to detect various (`~20`) typos across `react.dev` codebase ✨ :)

Pranav-yadav:
## Summary

This diff fixes various (`~20`) **typos** across `react.dev` codebase :)

## How did you test this change?

This diff fixes various typos across react including all of the packages and the remaining codebase.

I came across an excellent tool named [**typos**](https://github.com/crate-ci/typos) and used it for this codebase, _yet I fixed all the typos manually_ to take into account the context of the words (e.g. React uses `seeked` and `onSeeked`, which are lexically incorrect, but in react's context they are correct, so basically it's kept as it is).

For all of the changes, I assumed the `en-us` locale.

>**NOTE**: I did **not** use `--write-changes` flag while running the tool; hence, all changes are done manually.

Finally ran:
- `yarn check-all` --> `#00ff00`

### DIFF
--- a/src/components/MDX/CodeBlock/CodeBlock.tsx
+++ b/src/components/MDX/CodeBlock/CodeBlock.tsx
@@ -10,7 +10,7 @@ import {css} from '@codemirror/lang-css';
 import rangeParser from 'parse-numeric-range';
 import {CustomTheme} from '../Sandpack/Themes';
 
-interface InlineHiglight {
+interface InlineHighlight {
   step: number;
   line: number;
   startColumn: number;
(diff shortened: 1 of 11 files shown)


Pranav-yadav:
Also, @gaearon  can we add this tool ([typos](https://github.com/crate-ci/typos)) to the **CI** or as a **pre-commit hook**?
This should help avoid typos.

We can set up the _typos_ [workflow/job/gh-action](https://github.com/crate-ci/typos#integrations) by keeping the `_typos.toml` config that allows us to skip dirs, files, files w/ specific extensions, and repo-specific spell rules like (react uses `seeked` and `onSeeked` etc. that lexically are typos yet they can be skipped), etc...

e.g. `_typos.toml` file:

```toml
[default.extend-words]
seeked = "seeked"
onSeeked = "onSeeked"

[files]
extend-exclude = ["*.patch"]
```

gaearon:
sure, thanks. i don't mind trying this on CI if you wanna add it

Issue #5770, opened 2023-03-20
Wrong link for `useDeferredValue`


LoyalPotato:
In the [useTransition](https://react.dev/reference/react/useTransition#starttransition-caveats) page there's a reference to `useDeferredValue` that isn't linked correctly to [this](https://react.dev/reference/react/useDeferredValue).

PR #5771, opened 2023-03-20
Wrong link fix - #5770

LoyalPotato:


LoyalPotato:
Do I close the issue associated?

PR #5776, opened 2023-03-20
Replace deprecated command with environment file

### LINKED ISSUE #5775: Replace deprecated `set-output` command with environment file
**Describe the issue**
In workflow, `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information, see: [https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)

jongwooo:
## Description

Resolve #5775

Update `.github/workflows/analyze_comment.yml` to use environment file instead of deprecated `set-output` command. 

For more information, see: [https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)

I found the workflow files that use `set-output` command through the following command:

```bash
$ find .github/workflows -name '*.yml' -o -name '*.yaml' | xargs egrep '\bset-output\b'
```

**AS-IS**

```yml
echo ::set-output name=body::$body
echo ::set-output name=pr-number::$pr_number
```

**TO-BE**

```yml
echo "body=$body" >> $GITHUB_OUTPUT
echo "pr-number=$pr_number" >> $GITHUB_OUTPUT
```

### DIFF
--- a/.github/workflows/analyze_comment.yml
+++ b/.github/workflows/analyze_comment.yml
@@ -44,8 +44,8 @@ jobs:
           body="${body//'%'/'%25'}"
           body="${body//$'\n'/'%0A'}"
           body="${body//$'\r'/'%0D'}"
-          echo ::set-output name=body::$body
-          echo ::set-output name=pr-number::$pr_number
+          echo "body=$body" >> $GITHUB_OUTPUT
+          echo "pr-number=$pr_number" >> $GITHUB_OUTPUT
 
       - name: Find Comment
         uses: peter-evans/find-comment@v1


eps1lon:
Works: https://github.com/reactjs/react.dev/actions/runs/4471334178

There are still set-output usages (probably in actions we use). @jongwooo Would be awesome if you could update the usage of set-output in the used actions as well and then update the versions of those actions we use so that we're ready once set-output gets removed.

gaearon:
The comments appear jumbled. I'll try to figure out why

gaearon:
alright i fixed that

Issue #5783, opened 2023-03-21
Modal doesn't disappear when using search bar shortcut on initially rendered page


bb8dd:
https://user-images.githubusercontent.com/89984740/226610066-1c1bbff1-e93c-449e-b1b0-0c11d106cffe.mp4

PR #5784, opened 2023-03-21
Fix: Modal doesn't disappear when using search bar shortcut on initially rendered page

bb8dd:
Fixes: #5783 

- Add condition `(event.key === 'Enter' && isOpen)`
- `(event.keyCode === 27 && isOpen)` to `(event.key === 'Escape' && isOpen)` [KeyboardEvent.keyCode is deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode)

> press Enter without entering a search because of the conditions i added, the search modal disappears.

### DIFF
--- a/src/components/Layout/TopNav/TopNav.tsx
+++ b/src/components/Layout/TopNav/TopNav.tsx
@@ -247,9 +247,6 @@ export default function TopNav({
               </div>
               <div className="flex w-full md:hidden"></div>
               <div className="flex items-center -space-x-2.5 xs:space-x-0 ">
-                <div className="flex md:hidden">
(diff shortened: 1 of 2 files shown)


gaearon:
I don't think this is the root of the issue. The root seems to be the `<Search>` component (and related logic) being rendered twice in the tree. We should get rid of that.

gaearon:
That's what I meant, but just removing it is not the right fix. It was there for a reason :)

gaearon:
Thanks for looking into this. I merged a fix in https://github.com/reactjs/react.dev/pull/5794.

PR #5790, opened 2023-03-21
Beginner friendly method: Create react app added

### LINKED ISSUE #5781: Add create-react-app in New Project
Getting started with Production-grade React frameworks is certainly helpful but, adding `npx create-react-app` in [Start a New React Project](https://react.dev/learn/start-a-new-react-project#production-grade-react-frameworks) would be great for beginners wanting to create a react app for static content

x0samnan:
- Closes #5781.
- Added npx create-react-app for beginners trying to create static content.
- Moved a element accordingly inside a element.

### DIFF
--- a/src/content/learn/start-a-new-react-project.md
+++ b/src/content/learn/start-a-new-react-project.md
@@ -14,6 +14,29 @@ If you want to build a new app or a new website fully with React, we recommend p
 
 </Note>
 
+## React App without Frameworks {/*react-app-without-frameworks*/}
+
+<TerminalBlock>
+npx create-react-app
+</TerminalBlock>
+<Note>
+<DeepDive>
+
+#### Can I use React without a framework? {/*can-i-use-react-without-a-framework*/}
+
+You can definitely use React without a framework--that's how you'd [use React for a part of your page.](/learn/add-react-to-an-existing-project#using-react-for-a-part-of-your-existing-page) **However, if you're building a new app or a site fully with React, we recommend using a framework.**
+
+Here's why.
+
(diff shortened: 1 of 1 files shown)


sophiebits:
Thanks for the PR but as described in the deep dive on this page (that you moved), we don't recommend this for new apps. Our hope is that frameworks are actually more beginner-friendly, not less – and they scale with you as your app grows much better than Create React App and similar SPA-only tools do.

LeopoldLerch:
Well, that depends. In our company eg we use .Net Core for the backends. And we do not need SSR, ... and additional stuff, therefore we use JUST React (with typescript) and CSR. Of course, altough these frameworks add a lot of functionality, they also add some stuff to think of and keep in mind. 
So, eg beginners will have to know more than just how to do react in order to use the boilerplates.
In the end "create-react-app" is JUST React, not more, but not less. I think there is still place and need for this

But will it, altough documentation goes away, still be available and maintained?

LeopoldLerch:
Also
https://create-react-app.dev/
still and only mentions the "old" create-react-app. I just don´t understand, why we should hide the way, we used to do it in the past, just because there are now some additional features there. Even if the new ones are the recommended, there is nothing preventing to document and mention the "plain" version

@harish-sethuraman @sophiebits

rodrigc:
@LeopoldLerch If you need something more minimal than a full framework, you might want to give Vite a try, with the `react` or `react-ts` template: https://v3.vitejs.dev/guide/#trying-vite-online

The docs on the react.dev site lean towards recommending a framework, but it does briefly mention vite
under this popout section:
https://react.dev/learn/start-a-new-react-project#can-i-use-react-without-a-framework

PR #5791, opened 2023-03-21
docs: Remove `serverUrl` in code example

yujiseok:
Hi team!
I really like the new site.

I found `Duplicate identifier` error in code example. ([docs link](https://react.dev/learn/lifecycle-of-reactive-effects#react-verifies-that-you-specified-every-reactive-value-as-a-dependency))
In my small opinion, I'd like to suggest that get rid of it.


x0samnan:
@yujiseok Isn't this an example of error on purpose?

gaearon:
thanks!

gaearon:
no, this wasn't on purpose.

PR #5800, opened 2023-03-22
fix: removed ts-ignore from useEffect by creating new interface

hemant-dev-8:
removed ts-ignore from useEffect by creating new interface


sophiebits:
It's not clear that this is better. ts-ignore is fine in moderation IMO.

PR #5807, opened 2023-03-23
Update "top level" language to be less ambiguous with "levelContext"

jbonta:
We use "top level" all over the docs, but this specific topic uses a different meaning of "level". If you're learning about passing context across multiple "levels" of components, this is actually pretty confusing and makes you think you can only useContext at, for example, the top level of your app.


gaearon:
interesting, thanks

PR #5809, opened 2023-03-24
fix: incorrect description

alveifbklsiu259:
Hello, I think useEffect's cleanup function will run right before the component unmounts, instead of after, here is my testing, Thank you.
test


apmeehan:
Hi @alveifbklsiu259, I don't think your test demonstrates that the cleanup function is running before the unmount, as what's currently visible on the page doesn't necessarily reflect the current state of the DOM. For example, if you check devtools element inspector after the alert is triggered, you should see the `<h1>Text.js</h1>` element is actually no longer there.

Also, you can try querying the DOM in the cleanup function, and should again find that the relevant element can't be found ([example](https://codesandbox.io/s/cleanup-test-11u6c8?file=/src/Test.jsx)).

alveifbklsiu259:
> Hi @alveifbklsiu259, I don't think your test demonstrates that the cleanup function is running before the unmount, as what's currently visible on the page doesn't necessarily reflect the current state of the DOM. For example, if you check devtools element inspector after the alert is triggered, you should see the `<h1>Text.js</h1>` element is actually no longer there.
> 
> Also, you can try querying the DOM in the cleanup function, and should again find that the relevant element can't be found ([example](https://codesandbox.io/s/cleanup-test-11u6c8?file=/src/Test.jsx)).

Sorry for late reply and thanks for explanation and demonstration, I'll close the pull request.

PR #5810, opened 2023-03-24
fix: remove redundant spaces 

Davont:

### DIFF
--- a/src/sidebarLearn.json
+++ b/src/sidebarLearn.json
@@ -186,7 +186,7 @@
         },
         {
           "title": "Removing Effect Dependencies",
-          "path": "/learn/removing-effect-dependencies"              
+          "path": "/learn/removing-effect-dependencies"
         },
         {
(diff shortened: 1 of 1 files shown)


src/sidebarLearn.json
```diff
@@ -186,7 +186,7 @@
         },
         {
           "title": "Removing Effect Dependencies",
-          "path": "/learn/removing-effect-dependencies"              
+          "path": "/learn/removing-effect-dependencies"
```
harish-sethuraman, line 189: Ideally it should have been fixed by prettier. Lets add JSON to prettier run if it was not there before?

Davont: yes. I actually noticed this omission while saving the file locally. It's possible that this vulnerability was reported before formatting was put in place.

Davont: Perhaps you can format it with one click before submitting.

Issue #5812, opened 2023-03-24
Tailwind: Block and Flex in conflict


x0samnan:
Line 44 of SideNavbar.tsx seems to have a conflict between block and flex for large screens. Can we remove the flex?

![image](https://user-images.githubusercontent.com/92310785/227483753-4b14790e-4e04-41d5-9080-5dc31d3b8772.png)

PR #5814, opened 2023-03-24
Resolved the Tailwind CSS conflicts

abhinav-m22:
* Resolved the CSS conflict by removing the 'lg:flex' property as 'lg:flex' applies the same CSS properties as 'lg:block' in the file [SidebarNav.tsx](https://github.com/reactjs/react.dev/blob/main/src/components/Layout/SidebarNav/SidebarNav.tsx)

* Removed some Tailwind CSS properties which were added multiple times in [HomeContent.js](https://github.com/reactjs/react.dev/blob/main/src/components/Layout/HomeContent.js)

* Fixes the issue - #5812

### DIFF
--- a/src/components/Layout/HomeContent.js
+++ b/src/components/Layout/HomeContent.js
@@ -124,7 +124,7 @@ export function HomeContent() {
         <div className="mx-5 mt-12 lg:mt-24 mb-20 lg:mb-32 flex flex-col justify-center">
           <Logo
             className={cn(
-              'mt-4 mb-3 text-link dark:text-link-dark w-24 lg:w-28 self-center text-sm mr-0 text-link dark:text-link-dark flex origin-center transition-all ease-in-out'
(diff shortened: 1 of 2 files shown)


abhinav-m22:
> @abhinav-m22 Thanks for the PR but I am not sure yet, if it is flex that has to be removed. Also, there are a few more tailwind properties that are in conflict so, also have a look at them before the PR is merged.

The warning is displayed only on these two properties as they have same properties. It won't affect the other properties as they are completely different. So keeping any one of the two properties is fine.

Attaching the warning screenshot for reference:

image

x0samnan:
Great! Also, @abhinav-m22, search for Sidebar in Vscode and there are a few more tailwindcss conflicts, can you look for and resolve as it would be better than opening a new PR. This PR may get a little long though.

x0samnan:
LGTM!

PR #5836, opened 2023-03-28
Explain how `null` ends up in `ref.current` for React-managed refs

eps1lon:
Accompanying https://github.com/DefinitelyTyped/DefinitelyTyped/pull/64918 to reduce confusion in second gotcha of https://twitter.com/mattpocockuk/status/1636098722982404096

### DIFF
--- a/src/content/learn/referencing-values-with-refs.md
+++ b/src/content/learn/referencing-values-with-refs.md
@@ -284,7 +284,7 @@ You also don't need to worry about [avoiding mutation](/learn/updating-objects-i
 
 ## Refs and the DOM {/*refs-and-the-dom*/}
 
-You can point a ref to any value. However, the most common use case for a ref is to access a DOM element. For example, this is handy if you want to focus an input programmatically. When you pass a ref to a `ref` attribute in JSX, like `<div ref={myRef}>`, React will put the corresponding DOM element into `myRef.current`. Once the element is removed from the DOM, React will write `null` into `myRef.current`. You can read more about this in [Manipulating the DOM with Refs.](/learn/manipulating-the-dom-with-refs)
+You can point a ref to any value. However, the most common use case for a ref is to access a DOM element. For example, this is handy if you want to focus an input programmatically. When you pass a ref to a `ref` attribute in JSX, like `<div ref={myRef}>`, React will put the corresponding DOM element into `myRef.current`. Once the element is removed from the DOM, React will update `myRef.current` to be `null`. You can read more about this in [Manipulating the DOM with Refs.](/learn/manipulating-the-dom-with-refs)
 
 <Recap>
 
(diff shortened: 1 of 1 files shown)


ericanderson:
I think the problem in the tweet is that its weird that you have to specify `useRef<Foo | null>()` when the behavior of react using references is that the null is implied. Especially since `useRef<Foo>(null)` magically implies that the returned value is effectively typed `{ current Foo | null }` but the usage of `useRef<Foo>()` is not, its ` { current: Foo | undefined }`.

Given that there is zero documentation about `useRef()` with zero arguments, I think the biggest part of the confusion comes from being able to call it without any arguments.

eps1lon:
> Given that there is zero documentation about useRef() with zero arguments, I think the biggest part of the confusion comes from being able to call it without any arguments.

Yeah I'm not so sure about that one but I don't feel strongly about it. So we'll make the initial value mandatory.

This doesn't block this PR though. I think it's valueable to call out where `null` is coming from in `ref` props regardless of types.

src/content/learn/referencing-values-with-refs.md
```diff
@@ -284,7 +284,7 @@ You also don't need to worry about [avoiding mutation](/learn/updating-objects-i
 
 ## Refs and the DOM {/*refs-and-the-dom*/}
 
-You can point a ref to any value. However, the most common use case for a ref is to access a DOM element. For example, this is handy if you want to focus an input programmatically. When you pass a ref to a `ref` attribute in JSX, like `<div ref={myRef}>`, React will put the corresponding DOM element into `myRef.current`. You can read more about this in [Manipulating the DOM with Refs.](/learn/manipulating-the-dom-with-refs)
+You can point a ref to any value. However, the most common use case for a ref is to access a DOM element. For example, this is handy if you want to focus an input programmatically. When you pass a ref to a `ref` attribute in JSX, like `<div ref={myRef}>`, React will put the corresponding DOM element into `myRef.current`. Once the element is removed from the DOM, React will write `null` into `myRef.current`. You can read more about this in [Manipulating the DOM with Refs.](/learn/manipulating-the-dom-with-refs)
```
lunaleaps, line 287: nit: I find "React will write..." to be odd phrasing. How about:
`Once the element is removed from the DOM, React will update `myRef.current` to be `null`

lunaleaps (approved):

src/content/learn/referencing-values-with-refs.md
eps1lon:
```suggestion
You can point a ref to any value. However, the most common use case for a ref is to access a DOM element. For example, this is handy if you want to focus an input programmatically. When you pass a ref to a `ref` attribute in JSX, like `<div ref={myRef}>`, React will put the corresponding DOM element into `myRef.current`. Once the element is removed from the DOM, React will update `myRef.current` to be `null`. You can read more about this in [Manipulating the DOM with Refs.](/learn/manipulating-the-dom-with-refs)
```

eps1lon: That sounds better, thank you.

PR #5846, opened 2023-03-30
fix: illustration type check working correctly

fport:


lunaleaps:
What is this fixing? Can you elaborate?

PR #5853, opened 2023-04-01
fix(gh-workflow): skip comment when push directly

### LINKED ISSUE #5801: [GH workflow] analyze_bundle.txt missed on main branch or the workflow set incorrectly
When I merge PR into main branch, I always received the notification that the workflow failed like [this](https://github.com/reactjs/zh-hans.reactjs.org/actions/runs/4495876442).

I found the PR workflow is different from main branch. The latter doesn't generate comment indeed but it upload it and should be used the other workflow I metion at start.

### [The PR workflow](https://github.com/reactjs/zh-hans.reactjs.org/actions/runs/4495163315/jobs/7908500574):
pr
### [The main workflow](https://github.com/reactjs/zh-hans.reactjs.org/actions/runs/4495862925/jobs/7909947887):
main

awxiaoxian2020:
close #5801.

It works fine in [zh-hans repo workflow](https://github.com/reactjs/zh-hans.react.dev/actions/runs/4580485453/workflow).

I will explain my changes in code review.

### DIFF
--- a/.github/workflows/analyze.yml
+++ b/.github/workflows/analyze.yml
@@ -48,10 +48,10 @@ jobs:
 
       - name: Download base branch bundle stats
         uses: dawidd6/action-download-artifact@v2
-        if: success() && github.event.number
+        if: success()
         with:
           workflow: analyze.yml
-          branch: ${{ github.event.pull_request.base.ref }}
+          branch: ${{ github.event.pull_request.base.ref || 'main' }} 
           name: bundle_analysis.json
           path: .next/analyze/base/bundle
 
@@ -69,7 +69,7 @@ jobs:
       # Either of these arguments can be changed or removed by editing the `nextBundleAnalysis`
       # entry in your package.json file.
       - name: Compare with base branch bundle
-        if: success() && github.event.number
+        if: success()
         run: ls -laR .next/analyze/base && npx -p nextjs-bundle-analysis compare
 
       - name: Upload analysis comment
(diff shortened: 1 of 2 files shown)


.github/workflows/analyze.yml
```diff
@@ -48,10 +48,10 @@ jobs:
 
       - name: Download base branch bundle stats
         uses: dawidd6/action-download-artifact@v2
-        if: success() && github.event.number
+        if: success()
         with:
           workflow: analyze.yml
-          branch: ${{ github.event.pull_request.base.ref }}
+          branch: ${{ github.event.pull_request.base.ref || 'main' }}
```
awxiaoxian2020, line 54: When we push directly, we use main branch.

.github/workflows/analyze.yml
```diff
@@ -69,14 +69,21 @@ jobs:
...
+        if: success()
         run: ls -laR .next/analyze/base && npx -p nextjs-bundle-analysis compare
 
       - name: Upload analysis comment
         uses: actions/upload-artifact@v2
         with:
           name: analysis_comment.txt
           path: .next/analyze/__bundle_analysis_comment.txt
+          
+  number:
```
awxiaoxian2020, line 81: We shouldn't upload the pr_number when we push directly. So I split a new job and add a condition in order to run it on pull_request only.

.github/workflows/analyze_comment.yml
```diff
@@ -10,8 +10,7 @@ jobs:
   comment:
     runs-on: ubuntu-latest
     if: >
-      ${{ github.event.workflow_run.event == 'pull_request' &&
-      github.event.workflow_run.conclusion == 'success' }}
+        github.event.workflow_run.workflows.event_name == 'pull_request' && github.event.workflow_run.workflows.conclusion == 'success'
```
awxiaoxian2020, line 13: This is the key about this problem.

We should use `github.event.workflow_run.workflows.event_name` instead of `github.event.workflow_run.event`.

And we shouldn't use the `${{}}`, I think.

PR #5872, opened 2023-04-03
Use opening fragment instead of div

vedala:
In Declaring a Winner section,  code example shows `<div>` as outermost tag as first line within the return statement. The `<div>` should be replaced by `<>` instead, so that it is consistent with full code listing show at other places in the document.


harish-sethuraman (approved):
Thanks :)

Issue #5868, opened 2023-04-03
page not working


Surya9263:
- [ ] There is no content when you click on "rare use cases" in the "static getDerivedStateFromProps(props, state)" page.
[https://react.dev/reference/react/Component#static-getderivedstatefromprops](url)

PR #5874, opened 2023-04-03
fix: url

harish-sethuraman:
Fixes https://github.com/reactjs/react.dev/issues/5868

### DIFF
--- a/vercel.json
+++ b/vercel.json
@@ -140,11 +140,6 @@
       "permanent": false
     },
     {
-      "source": "/link/derived-state",
-      "destination": "https://legacy.reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html",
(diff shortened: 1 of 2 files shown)


harish-sethuraman:
We have setup redirect in `vercel.json`?

gaearon:
Yes but I'd like the links to be direct because this is clearer.

harish-sethuraman:
Removed in redirect as well. Is that fine?

vercel.json
```diff
@@ -139,11 +139,6 @@
       "destination": "/reference/react-dom/components/common#dangerously-setting-the-inner-html",
       "permanent": false
     },
-    {
```
gaearon, line 142: We should keep the redirect. We'll keep using it from React error messages.

PR #5883, opened 2023-04-05
Fix improper idiom usage

jhylacey:
The phrase "most of the times" is incorrect and would only make sense
in the context of a larger sentence. The phrase is actually "most of the
time" and is used in place of "usually".

### DIFF
--- a/src/content/learn/importing-and-exporting-components.md
+++ b/src/content/learn/importing-and-exporting-components.md
@@ -138,10 +138,10 @@ How you export your component dictates how you must import it. You will get an e
 
 | Syntax           | Export statement                           | Import statement                          |
 | -----------      | -----------                                | -----------                               |
-| Default  | `export default function Button() {}` | `import Button from './button.js';`     |
-| Named    | `export function Button() {}`         | `import { Button } from './button.js';` |
+| Default  | `export default function Button() {}` | `import Button from './Button.js';`     |
+| Named    | `export function Button() {}`         | `import { Button } from './Button.js';` |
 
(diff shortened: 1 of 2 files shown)


src/content/learn/importing-and-exporting-components.md
```diff
@@ -138,10 +138,10 @@ How you export your component dictates how you must import it. You will get an e
...
 | -----------      | -----------                                | -----------                               |
-| Default  | `export default function Button() {}` | `import Button from './button.js';`     |
-| Named    | `export function Button() {}`         | `import { Button } from './button.js';` |
+| Default  | `export default function Button() {}` | `import Button from './Button.js';`     |
+| Named    | `export function Button() {}`         | `import { Button } from './Button.js';` |
 
-When you write a _default_ import, you can put any name you want after `import`. For example, you could write `import Banana from './button.js'` instead and it would still provide you with the same default export. In contrast, with named imports, the name has to match on both sides. That's why they are called _named_ imports!
+When you write a _default_ import, you can put any name you want after `import`. For example, you could write `import Banana from './Button.js'` instead and it would still provide you with the same default export. In contrast, with named imports, the name has to match on both sides. That's why they are called _named_ imports!
```
harish-sethuraman, line 144: These changes were merged in some other PR.

jhylacey: @harish-sethuraman I'm very new to using Github, could you explain to me why it was that I pushed that old commit? Or direct to me a resource that explains? I did the same process for the second of the two commits, ie made sure everything was up to date from main branch, but for some reason the same one went through twice.

gaearon: you probably created the second branch _from_ your first branch rather than from `main`. so it included your last commit.

PR #5885, opened 2023-04-06
fix: minor wording adjustment

escwxyz:
`for` your components which are wrapped inside the context


gaearon:
but also _from_ the components providing context (which is what the example shows).

PR #5893, opened 2023-04-07
Upgrade react-collapsed

roginfarrer:
Replaces #5511 

`react-collapsed` v4 includes React 18 support (ie, uses `useId`), improves types, and fixes some edge case issues with animations. Doesn't involve the vanilla JS core refactor as discussed in the original PR (I haven't given up on that, but I didn't want to leave the original package unmaintained while I'm working on it).

Tested by running locally, I don't see any layout jumps as observed in the original PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "next-remote-watch": "^1.0.0",
     "parse-numeric-range": "^1.2.0",
     "react": "^0.0.0-experimental-16d053d59-20230506",
-    "react-collapsed": "npm:@gaearon/react-collapsed@3.1.0-forked.1",
+    "react-collapsed": "4.0.2",
     "react-dom": "^0.0.0-experimental-16d053d59-20230506",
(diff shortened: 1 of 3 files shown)


lunaleaps:
Thanks for the update! Can you share the instructions to test locally?

roginfarrer:
> Thanks for the update! Can you share the instructions to test locally?

@lunaleaps Sure! The collapse is used for the sidebar. You can test it be running the app and clicking through the different sections. 

CleanShot 2023-10-03 at 19 50 28@2x

lunaleaps:
This looks good to me! However in the process of testing, I noticed that there's a scrollbar UI issue but I was able to also reproduce on `main`. I'll file an issue.

Screenshot 2023-10-10 at 9 23 37 PM

lunaleaps:
Filed the issue here: https://github.com/reactjs/react.dev/issues/6348

PR #5895, opened 2023-04-07
Improved explanations of a solution to a challenge

Aleksandar15:
Improved solution to a challenge # 3. at [Challenges](https://react.dev/learn/javascript-in-jsx-with-curly-braces#challenges).

Current explanation:
current explanations

Improved explanation:
improved explanations

### DIFF
--- a/src/content/learn/javascript-in-jsx-with-curly-braces.md
+++ b/src/content/learn/javascript-in-jsx-with-curly-braces.md
@@ -480,9 +480,9 @@ To check that your fix worked, try changing the value of `imageSize` to `'b'`. T
 
 You can write it as `src={baseUrl + person.imageId + person.imageSize + '.jpg'}`.
 
-1. `{` opens the JavaScript expression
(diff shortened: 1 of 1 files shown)


gaearon:
I find this a bit too wordy — it makes it sounds more complex than it is. Can you think of a more concise explanation?

Aleksandar15:
> I find this a bit too wordy — it makes it sounds more complex than it is. Can you think of a more concise explanation?

That was my initial intention, then I went on with the step-by-step guide because the whole section is beginner-friendly and keeps reminding users of the JavaScript's fundamentals, but this explanation was kinda "*neglected*".

I was also thinking of when:
a) More experienced users read this, the first sentence will be enough to give them *aha-moment* and scroll down to the code.
b) Beginners read this, still have no idea what's happening, and continue to read the steps to give them *aha-moment*.

A workaround I can think of is to keep the 1-sentence above the Sandbox, but move the steps below the Sandbox, and have it say "*Here's what happened:*"

What's your take on it? Should I rather shorten the steps, or move them below ?

Aleksandar15:
Shortened explanation:

image

PR #5930, opened 2023-04-18
Add <meta> tags for translations

gaearon:
- Add language code to the site config for translations to override
- Set `<html lang>` based on that code
- Add `<link ref="canonical">` for the current page
- Add `<link rel="alternate" hrefLang="x-default">` pointing to the English version
- Add `<link rel="alternate" hrefLang="...">` for each language (currently, only English)
- Move `react.dev/docs/some-old.url.html -> legacy.reactjs.org/docs/some-old.url.html` redirect from Vercel config to Next.js config. This lets us send it to the current translation's legacy site instead of the English legacy site.

### DIFF
--- a/src/components/Seo.tsx
+++ b/src/components/Seo.tsx
@@ -22,7 +22,7 @@ const deployedTranslations = [
   // We'll add more languages when they have enough content.
 ];
 
-function getDomain(languageCode) {
+function getDomain(languageCode: string): string {
   const subdomain = languageCode === 'en' ? '' : languageCode + '.';
   return subdomain + 'react.dev';
 }
--- a/src/components/Seo.tsx
+++ b/src/components/Seo.tsx
@@ -55,13 +55,13 @@ export const Seo = withRouter(
         <link
           rel="alternate"
           href={canonicalUrl.replace(siteDomain, getDomain('en'))}
-          hreflang="x-default"
+          hrefLang="x-default"
         />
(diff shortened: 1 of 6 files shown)


src/components/Seo.tsx
```diff
@@ -16,6 +17,16 @@ export interface SeoProps {
...
 }
 
+const deployedTranslations = [
+  'en',
+  // We'll add more languages when they have enough content.
+];
+
+function getDomain(languageCode: string): string {
+  const subdomain = languageCode === 'en' ? '' : languageCode + '.';
+  return subdomain + 'react.dev';
```
gaearon, line 27: We're going to assume this is where they canonically point — even before some translation's docs are stable enough to be published at the domain.

src/siteConfig.js
```diff
@@ -2,8 +2,12 @@
  * Copyright (c) Facebook, Inc. and its affiliates.
  */
 
-export const siteConfig = {
-  editUrl: 'https://github.com/reactjs/react.dev/tree/main/src/pages',
+exports.siteConfig = {
+  // --------------------------------------
+  // Translations should replace these lines:
+  languageCode: 'en',
+  hasLegacySite: true,
```
gaearon, line 9: Translators should change these two lines above ^^

gaearon: (To clarify, we'll probably do this ourselves though as we're going through fixing the syncs.)

PR #5949, opened 2023-04-21
onFocus fires on focus not on blur

Binney:
Fairly sure this is just a typo as the copy was exactly the same as the description for `onBlur`.

(disclaimer: I have not verified this behaviour myself)


lunaleaps:
Looks like this has been fixed, thanks for flagging @Binney!

https://github.com/reactjs/react.dev/blob/2dbd19531f5031e81a67eba532203c9abedc054a/src/content/reference/react-dom/components/common.md?plain=1#L99

Issue #5950, opened 2023-04-21
Important note about `set` functions missing from new `useState` docs


iFreilicht:
I was attending a class about `useCallback` in conjunction with `useState` in a custom hook, and the solution to my exercise used an empty dependency array in `useCallback`, even though it used a `setValue` from a state:

```js
  const [value, setValue] = React.useState(initialValue);

  const toggleValue = React.useCallback(
    () => setValue((currentValue) => !currentValue),
    [] # I put [setValue] here
  );

  return [value, toggleValue];
```

I wasn't sure if this was ok, but didn't find any info about this in the [new docs of useState](https://react.dev/reference/react/useState#setstate). I asked around and found that [the old docs **do** contain the specific information I was looking for](https://legacy.reactjs.org/docs/hooks-reference.html#usestate):

![image](https://user-images.githubusercontent.com/9742635/233661115-511e35cc-af69-4dc6-8ef4-3ce793578118.png)

I'd like to suggest to add this note or something similar in the new docs as well. It might have been omitted for a good reason, but it does feel like important information to me.

PR #5956, opened 2023-04-24
Important note about `set` functions missing from new `useState` docs #5950

umairraza96:
Added the note that `setState` wont change on re-renders.

closes issue number #5950


umairraza96:
Closes #5950 

image

PR #5907, opened 2023-04-10
fix: update analyze_comment `body-includes`

awxiaoxian2020:
We can see https://github.com/reactjs/react.dev/actions/runs/4654784574/jobs/8236816631#step:6:13
It has updated to `<!-- __NEXTJS_BUNDLE_react-dev -->`

PR #5971, opened 2023-04-28
Make bundle analysis workflow less fragile

gaearon:
- Pin the version (https://github.com/hashicorp/nextjs-bundle-analysis/issues/52#issuecomment-1527700321).
- Use the "sticky comment" workflow (https://github.com/hashicorp/nextjs-bundle-analysis/pull/54) to fix the regression introduced by https://github.com/hashicorp/nextjs-bundle-analysis/pull/44. Alternative to https://github.com/reactjs/react.dev/pull/5907.


gaearon:
This needed an extra fix in https://github.com/reactjs/react.dev/commit/496c5784a971aee2980abdd0b3b9ff1d9be30505, but works now.

PR #5976, opened 2023-04-28
Add basic reference pages for use client and server

sophiebits:
I guess this turned into a bit more of an FAQ but I think it's useful to have this written down in a canonical place.


rickhanlonii (approved):
This is great, @sophiebits do you think we should add some kind of indicator that this is a "canary" feature? Approving because we can follow up to add that.

sophiebits:
> do you think we should add some kind of indicator that this is a "canary" feature?

This is what the Note at the top is for; don’t think it’s enough?

src/content/reference/react/use-client.md
```diff
@@ -0,0 +1,57 @@
...
+
+#### Caveats {/*caveats*/}
+
+* It's not necessary to add `'use client'` to every file that uses client-only React features, only the files that are imported from server component files. `'use client'` denotes the _boundary_ between server-only and client code; any components further down the tree will automatically be executed on the client. In order to be rendered from server components, components exported from `'use client'` files must have serializable props.
+* When a `'use client'` file is imported from a server file, the imported values can be rendered as a React component or passed via props to a client component. Any other use will throw an exception.
+* When a `'use client'` file is imported from another client file, the directive has no effect. This allows you to write client-only components that are simultaneously usable from server and client components.
+push down
```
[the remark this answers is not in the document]
sophiebits: thanks, fixed

sophiebits:
¯\\\_(ツ)\_/¯ I think this is better than nothing

src/content/reference/react/use-server.md
```diff
@@ -0,0 +1,48 @@
...
+
+<InlineToc />
+
+---
+
+## Reference {/*reference*/}
+
+### `'use server'` {/*use-server*/}
+
+Add `'use server';` at the very top of an async function to mark that the function can be executed by the client.
```
harish-sethuraman, line 28:
```suggestion
Add `'use server';` at the very top of an async function to mark that the function can be executed by the server.
```
is it me or should this be executed by server?

sophiebits: hm. I meant that it can be imported + triggered by the client

harish-sethuraman: Yea nvm got confused 😅

rickhanlonii:
Yeah, I think I was looking for it to mention the "canary" specifically with a link to the versioning policy.

sophiebits:
For most APIs we’ll add I agree (probably with a specific version number noted) but for these specifically I’m not sure it’s helpful because checking the version won’t tell you if you’re already using an RSC framework and upgrading to canary won’t help you if you’re not.

sophiebits:
(Not opposed to adding it here for consistency though once we have a standard pattern!)

PR #5977, opened 2023-04-28
update chronological order of the React Miami conf

M0nica:
Moved React Miami 2023 to previous conferences since it just passed!


sophiebits:
thanks!

PR #5981, opened 2023-04-30
fix Illustration and IllustrationBlock alt color on dark mode

oxcened:
### Motivation

When dark mode is turned on, the alt text rendered by the Illustration and IllustrationBlock components is not visible.
On dark mode, the text color inherited is `text-primary-dark`, which is not visible on the white background.

### Change

Applying a fixed `text-primary` ensures that the text is always visible regardless of the current theme.

### Demonstration

Source of the images have been removed for the sake of testing.

**Current behaviour:**

Screenshot 2023-04-30 at 17 29 58
Screenshot 2023-04-30 at 17 32 03

**Fixed behaviour:**

Screenshot 2023-04-30 at 17 30 02
Screenshot 2023-04-30 at 17 32 06

### DIFF
--- a/src/components/MDX/MDXComponents.tsx
+++ b/src/components/MDX/MDXComponents.tsx
@@ -239,7 +239,7 @@ function Illustration({
           src={src}
           alt={alt}
           style={{maxHeight: 300}}
-          className="bg-white rounded-lg"
+          className="rounded-lg"
         />
         {caption ? (
           <figcaption className="text-center leading-tight mt-4">
@@ -271,7 +271,12 @@ function IllustrationBlock({
   const images = imageInfos.map((info, index) => (
     <figure key={index}>
       <div className="bg-white rounded-lg p-4 flex-1 flex xl:p-6 justify-center items-center my-4">
-        <img src={info.src} alt={info.alt} height={info.height} />
+        <img
+          className="text-primary"
+          src={info.src}
+          alt={info.alt}
+          height={info.height}
+        />
       </div>
       {info.caption ? (
         <figcaption className="text-secondary dark:text-secondary-dark text-center leading-tight mt-4">


lunaleaps:
Is it possible to also make the alt text dark mode when the theme is dark mode?

oxcened:
> Is it possible to also make the alt text dark mode when the theme is dark mode?

Not sure I understood your question. The alt text does currently inherit the dark mode, but since the background is fixed to white, it is not readable when dark mode is on.

lunaleaps:
@oxcened Oh I guess I was asking if the alt text background could also turn dark during dark mode, that way we don't need to address the text color.

oxcened:
> @oxcened Oh I guess I was asking if the alt text background could also turn dark during dark mode, that way we don't need to address the text color.

Ah I got you. We could do that, but as far as I can tell, all illustration images have a white background, and I'm not sure it would look very clean.

Screenshot 2023-09-28 at 09 31 56

Screenshot 2023-09-28 at 09 32 02

---

Otherwise you could keep the white background on the image container, and set the dark background on the image only.

Screenshot 2023-09-28 at 09 40 44

Not sure there any other solutions?

harish-sethuraman:
There are two illustrations
- single illustration that has a white bg in image so text isnt visible in dark mode
- illustration block that has a parent div with rounded white bg

in the single illustration bg white is not needed at all? it doesn't show up until the image is not loaded. so ideally the text is visible when bg is removed.  (can remove `bg-white`?)

in the block we can consider having `text-primary` so that the text is visible in both modes and we do not interfere with the background white of the image that has rounded borders?  (can add `text-primary` to image?)

oxcened:
> in the single illustration bg white is not needed at all? it doesn't show up until the image is not loaded. so ideally the text is visible when bg is removed. (can remove `bg-white`?)

Yeah it is different because the background is applied directly to the img. So as long as the images have their own white background, the style has no visible effect. I think this is currently true for all images, so it depends whether it might change in the future I guess?

> in the block we can consider having `text-primary` so that the text is visible in both modes and we do not interfere with the background white of the image that has rounded borders? (can add `text-primary` to image?)

I think this is the best option. Currently the PR applies it to the outer div instead of the img, though.

oxcened:
hey @lunaleaps @harish-sethuraman I did some changes according to your comments, what do you think?

lunaleaps (approved):

lunaleaps:
Thanks for fixing, this looks good to me!

PR #5983, opened 2023-04-30
fix(adding-interactivity): fixed Render and Commit link label

oxcened:
Fixes the label of the link that navigates to `Render and Commit`.

The label says that React updates the UI in **two** phases, however the page it links to affirms multiple times that the phases are **three** (Trigger, Render, Commit).


gaearon:
I'm not super happy with the wording on that phase, but "Trigger" isn't really meant as a phase, it just refers to kicking off the process.

oxcened:
> I'm not super happy with the wording on that phase, but "Trigger" isn't really meant as a phase, it just refers to kicking off the process.

Ah, got it. Maybe we could opt for something like "Request" or "Queue".

PR #5996, opened 2023-05-04
Fix grammatical error: change 'are' to 'is'

Tunzeki:
The sentence reads:

With this approach, none of the existing state items are being mutated, and the bug is fixed:

It should read:

With this approach, none of the existing state items is being mutated, and the bug is fixed:

### DIFF
--- a/src/content/learn/updating-arrays-in-state.md
+++ b/src/content/learn/updating-arrays-in-state.md
@@ -567,7 +567,7 @@ setMyList(myList.map(artwork => {
 
 Here, `...` is the object spread syntax used to [create a copy of an object.](/learn/updating-objects-in-state#copying-objects-with-the-spread-syntax)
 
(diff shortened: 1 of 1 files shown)


src/content/learn/updating-arrays-in-state.md
```diff
@@ -567,7 +567,7 @@ setMyList(myList.map(artwork => {
 
 Here, `...` is the object spread syntax used to [create a copy of an object.](/learn/updating-objects-in-state#copying-objects-with-the-spread-syntax)
 
-With this approach, none of the existing state items are being mutated, and the bug is fixed:
```
BartoszKlonowski, line 570: I think this is correct as written.

BartoszKlonowski (commented):
The usage of `are` is correct in this case, so I will close this PR. However, please feel free to comment if there are some specific grammar rules that would say otherwise.

PR #6005, opened 2023-05-06
Remove pitfall saying NextJS app router is in beta

nahal04:
The Next.js app router is now stable as of Next.js 13.4. Therefore I have removed the Pitfall on 'Start a New React Project' page, which says "Next.js’s App Router is currently in beta and is not yet recommended for production"

### DIFF
--- a/src/content/learn/start-a-new-react-project.md
+++ b/src/content/learn/start-a-new-react-project.md
@@ -92,11 +92,6 @@ These features are getting closer to being production-ready every day, and we've
 **[Next.js's App Router](https://beta.nextjs.org/docs/getting-started) is a redesign of the Next.js APIs aiming to fulfill the React team’s full-stack architecture vision.** It lets you fetch data in asynchronous components that run on the server or even during the build.
 
 Next.js is maintained by [Vercel](https://vercel.com/). You can [deploy a Next.js app](https://nextjs.org/docs/deployment) to any Node.js or serverless hosting, or to your own server. Next.js also supports [static export](https://beta.nextjs.org/docs/configuring/static-export) which doesn't require a server.
-<Pitfall>
-
-Next.js's App Router is **currently in beta and is not yet recommended for production** (as of Mar 2023). To experiment with it in an existing Next.js project, [follow this incremental migration guide](https://beta.nextjs.org/docs/upgrade-guide#migrating-from-pages-to-app).
-
-</Pitfall>
 
 <DeepDive>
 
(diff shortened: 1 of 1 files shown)


gaearon:
There's a few things I'd like to do:

- We should make sure the links from both Pages Router an App Router go to the corresponding docs sections and that installation instructions are correct for both. Would you mind fixing that?
- I think we should keep a note here, but change the wording to suggest that although App Router is stable, it is still bleeding edge tech.

nahal04:
Sure. Will work on that :+1:

src/content/learn/start-a-new-react-project.md
```diff
@@ -26,7 +26,12 @@ npx create-next-app
...
 
-Next.js is maintained by [Vercel](https://vercel.com/). You can [deploy a Next.js app](https://nextjs.org/docs/deployment) to any Node.js or serverless hosting, or to your own server. [Fully static Next.js apps](https://nextjs.org/docs/advanced-features/static-html-export) can be deployed to any static hosting.
+Next.js is maintained by [Vercel](https://vercel.com/). You can [deploy a Next.js app](https://nextjs.org/docs/pages/building-your-application/deploying) to any Node.js or serverless hosting, or to your own server. [Fully static Next.js apps](https://nextjs.org/docs/pages/building-your-application/deploying/static-exports) can be deployed to any static hosting.
+
+<Note>
+There are two ways of building a Next.js app - using [*App Router*](#nextjs-app-router) or *Pages Router*. During installation, you will be prompted to use the *App Router*. Choose *'No'* to stick with the older *Pages Router*. However, using an *App Router* is the current recommended way of building a Next.js app. The *App Router* is built on bleeding-edge React features like **React Server Components**
+See [Bleeding Edge React Frameworks](#bleeding-edge-react-frameworks).
+</Note>
```
nickmccurdy, line 34: I feel like this wording could be confusing, as it goes back and forth between suggesting app router and pages router, and it starts by explaining "no" even though "yes" is recommended.

nickmccurdy:
@nahal04 Thanks for updating this. How about adding `npx create-next-app@latest` to the first section too? It's important that pages router users still get the latest template, especially if they plan on migrating to app router later.

Also, do we still need the bleeding edge section? I personally find it confusing that Next is in both sections even though both its routers are considered stable, and AFAIK all meta frameworks that are production ready and support RSCs still support rendering without RSCs.

lunaleaps:
Hi @nahal04, thanks for making this PR. We want to re-visit this page but for now, https://github.com/reactjs/react.dev/pull/6318 will minimally remove the outdated information. We need some discussion around how we want to frame each framework and will make that change. For now, I'll close this PR. Feel free if you have other concerns.

Issue #6006, opened 2023-05-06
Update the "start a new react project page"


thesuryavivek:
## App router in Nextjs 13.4 is stable and the warning section that it's still in beta can be removed.

<img width="960" alt="image" src="https://user-images.githubusercontent.com/72210941/236633061-2439388e-74d2-474c-8cbe-6442d57083f2.png">

PR #6007, opened 2023-05-06
deleted pitfall component for NextJS app router beta warning

thesuryavivek:
Issue #6006 

App router is now stable in NextJS 13.4 

So, The pitfall component for the NextJS app router still in beta from the file "start-a-new-react-project.md" is removed.

I think the pitfall component is only used there. But I didn't deleted the pitfall component and it's icon component files from the codebase. Let me know if these components can be deleted.


thesuryavivek:
This is already in work by another contributor.

PR #6016, opened 2023-05-10
fix: ambiguous meaning

yionr:
First, let me describe an aspect of the pure function:
```
let a = 1

add(a,1)

console.log(a) // if output is 2 , then the function is not a pure functionn ,right?
```

I think the original sentence _'It does not change any objects or variables that existed before it was called.'_ can simply devide into two parts: 
First part is _'It does not change any objects or variables that existed'_ , 
And the second is _'befor it was called'_ .

It is confuse: What is the relationship between changes in external variables and the function itself before it is called?

Perhaps the original intention of the author was to express: 
_'It does not change any objects or variables that existed before after it was called.'_ 
Which just insert a word 'after' , then the sentence can split into _'It does not change any objects or variables that existed before'_ and _'after it was called.'_ 

ps. I'm not a native English speaker,If any wrong with it,i'm feel so sorry , but in chinese translation , it did really translated into a wrong meaning :(

### DIFF
--- a/src/content/learn/describing-the-ui.md
+++ b/src/content/learn/describing-the-ui.md
@@ -466,7 +466,7 @@ Read **[Rendering Lists](/learn/rendering-lists)** to learn how to render a list
 
 Some JavaScript functions are *pure.* A pure function:
 
-* **Minds its own business.** It does not change any objects or variables that existed before it was called.
+* **Minds its own business.** It does not change any objects or variables that existed before after it was called.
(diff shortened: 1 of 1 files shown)


yionr:
Execuse me! I think i have understand it ,the sentence means that before the function is called,the objects and variables was just existed, and should not be changed after function invoke.I'll PR on translation repo.
But I still feel that it may be slightly ambiguous. I would greatly appreciate if a native English speaker could offer their input on this and suggest if there's a better way to phrase it.

PR #6024, opened 2023-05-12
fix: invalid description for onFocus dom prop

### LINKED ISSUE #5933: Invalid description for `onFocus` DOM prop in the reference page
The [common-props](https://react.dev/reference/react-dom/components/common#common-props) reference page is providing invalid description for `onFocus` DOM prop

**Actual Behaviour:**

image

**Expected Behaviour:**

`onFocus` method is fired only when the element **receives or enters** focus not **lose** focus.

So the description should look like the following 
- onFocus: A [FocusEvent handler](https://react.dev/reference/react-dom/components/common#focusevent-handler) function. Fires when an element receives [focus](https://developer.mozilla.org/en-US/docs/Web/API/Element/focus_event). Unlike the built-in browser focus event, in React the onFocus event bubbles.

Also I can make the PR for it, if the issue is verified.

jitiendran:
Updated the description for `onFocus` DOM prop in the reference page

fixes #5933


rickhanlonii:
Fixed in https://github.com/reactjs/react.dev/pull/6168

PR #6026, opened 2023-05-12
Add accessibility note to interactivity section in the new React docs

kmiddleton14:
We want to add accessibility into the new React Docs. This PR adds a note in the interactivity section letting users know to use the html tag `<button>` instead of `<div>` when using onClick handlers.

Screenshot 2023-05-12 at 1 17 42 PM

### DIFF
--- a/src/content/learn/adding-interactivity.md
+++ b/src/content/learn/adding-interactivity.md
@@ -26,12 +26,6 @@ React lets you add *event handlers* to your JSX. Event handlers are your own fun
 
 Built-in components like `<button>` only support built-in browser events like `onClick`. However, you can also create your own components, and give their event handler props any application-specific names that you like.
 
-<Note>
-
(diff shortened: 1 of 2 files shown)


jessebeach (commented):
Makes sense and conforms to best practice recommendations.

sophiebits:
This overview is really supposed to be a brief 1-minute summary of each page within the section. How about we instead add a note to the "Responding to Events" page at an appropriate place? 

I also noticed that the linked page is a little terse and jargony and I'm not sure that it is accessible (ha) to many of our site visitors. I found https://developer.mozilla.org/en-US/docs/Learn/Accessibility/HTML which might be a better introductory resource to link to.

kmiddleton14:
Added to the "Responding to Events" page and updated the link to https://developer.mozilla.org/en-US/docs/Learn/Accessibility/HTML

jessebeach:
Looks good. It conveys the right message.

PR #6027, opened 2023-05-12
Remove extra "on"

Tunzeki:
A comment reads: 

So you tell React that this Effect "depends on" on props and state

It should read: 

So you tell React that this Effect "depends on" props and state


musllim (commented):
LGTM

PR #6029, opened 2023-05-13
Synonyms are update in the Home Page

rakshithcgowda:
Screenshot (41)
Just changed the phrase to the more meaning full one.


sophiebits:
Appreciate the thought but I prefer it as is.

PR #6043, opened 2023-05-17
Fix wrong explanation in preserving-and-resetting-state

smikitky:
In "Preserving and Resetting State", I found that a part of the description did not correctly reflect the corresponding example code and diagrams. The correct behavior (as seen in the code and the diagrams) is as follows:

| phase | `isFancy` state | container |
| - | :-: | :-: |
| initial state | `false` | `section` |
| after checking the checkbox | `true` | `div` |

### DIFF
--- a/src/content/learn/preserving-and-resetting-state.md
+++ b/src/content/learn/preserving-and-resetting-state.md
@@ -688,7 +688,7 @@ label {
 
 </Sandpack>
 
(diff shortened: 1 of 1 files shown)


smikitky:
@harish-sethuraman Could you take a look at this? I believe this is a fairly simple error.

rammba:
Hey @smikitky, thanks for this PR, I also noticed that and wanted to change it.
Hello @rickhanlonii, @poteto, @eps1lon can you please check this one?

PR #6044, opened 2023-05-17
Fix incorrect inner component reference

tdd:
The inner component is `EditForm`, not `EditContact`.


smikitky:
I noticed the same problem while translating this article. I believe this needs to be merged.

harish-sethuraman (approved):

PR #6047, opened 2023-05-17
Fixing inconsistent prop names in useMemo/useCallback examples

bobziroll:
In the useCallback docs page, the examples pass a prop called `referrerId` to the `ProductPage` component. But the `ProductPage` component is receiving a prop called `referrer` instead of `referredId`. This PR changes the example prop being passed to say `referrer` instead of `referrerId`.

### DIFF
--- a/src/content/reference/react/useCallback.md
+++ b/src/content/reference/react/useCallback.md
@@ -269,7 +269,7 @@ export default function App() {
       </label>
       <hr />
       <ProductPage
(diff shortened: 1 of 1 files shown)


jaba9531:
Hello! I was just reading through the docs and found this mistake, and was going to submit a PR but found that this one already exists. Can we move forward with merging it? I’m happy to make any changes if needed.

PR #6050, opened 2023-05-18
fix next.js broken link on suspense reference

abdulhdr1:
There was a broken link on the Suspense page when it mentions the possibility to use suspense on data fetching with Next.js, leading to a 404 page, this links it correctly to the suspense section on next's docs. Also, I'm not sure if it's a good practice on the docs to use a link with an anchor, as they can be fragile but link directly to the content.


gaearon (changes requested):
This link also exists on other pages. Let's fix it consistently?

abdulhdr1:
fixed https://github.com/reactjs/react.dev/pull/6053

PR #6059, opened 2023-05-22
docs: list Vite as viable option for new projects

### LINKED ISSUE #5797: List Vite at the same level as SSR frameworks as a way to start a new React project
In the _"Start a New React Project"_ section of the docs, Vite (and Parcel) are currently mentioned at the bottom of a collapsible toggle _"Can I use React without a framework"_, hidden by default. They are listed as an option only if your app has "unusual" constraints (no further explanation or example of what that could be).

As Dan recently explained, the main reason behind this is that SSR frameworks offer the optionality of, well, SSR. The argument is that although you may not need it now, you might in the future, so you should still start with an SSR framework. But it isn't a free lunch - this way, you get immediately locked into a specific framework that makes a lot of other decisions for you that you haven't even had a chance to evaluate yet.

The recent article ["New React docs pretend SPAs don't exist anymore"](https://wasp-lang.dev/blog/2023/03/17/new-react-docs-pretend-spas-dont-exist) caused a lot of reactions, and many developers expressed their opinion on [Reddit (300+ comments)](https://www.reddit.com/r/reactjs/comments/11u1r3w/new_react_docs_pretend_spas_dont_exist_anymore/) and [HackerNews (200+ comments)](https://news.ycombinator.com/item?id=35201848).

This was the top comment in the Reddit thread:
Screenshot 2023-03-22 at 12 52 27

My point is not at all that an SSR framework is a bad/wrong choice for any kind of app, and I believe it is a totally valid recommendation, even for SPAs. But it is evident that developers still widely recognize other solutions as equally good for particular use cases. Also, with the current setup, beginners in React are now forced to immediately choose between three frameworks which all add a lot on the top of React. It is almost an impossible task for a beginner (in the end, 80% will choose Next simply because it is listed on top).

**Vite should be listed next to Next, Remix, and Gatsby. I would clearly depict the pros and cons and possible pitfalls.**

I understand that the aim here was to protect developers from using react in a "wrong" way, but, at the end of the day, React is a library, not a framework, and there are a lot of benefits to it. The docs should respect and objectively reflect what the developer community identified as sensible ways to use it, not the other way around.

scarf005:
## Summary

image

- resolves #5797

### Renamed `Can I use React without a framework?` to `Should I use React with a framework?`

- reflected as Vite is now listed as a viable and default option for starting react project.
- changed link `add-react-to-an-existing-project.md` to account for the change in link.

### DIFF
--- a/src/content/learn/add-react-to-an-existing-project.md
+++ b/src/content/learn/add-react-to-an-existing-project.md
@@ -24,7 +24,7 @@ Here's how we recommend to set it up:
 2. **Specify `/some-app` as the *base path*** in your framework's configuration (here's how: [Next.js](https://nextjs.org/docs/api-reference/next.config.js/basepath), [Gatsby](https://www.gatsbyjs.com/docs/how-to/previews-deploys-hosting/path-prefix/)).
 3. **Configure your server or a proxy** so that all requests under `/some-app/` are handled by your React app.
 
-This ensures the React part of your app can [benefit from the best practices](/learn/start-a-new-react-project#can-i-use-react-without-a-framework) baked into those frameworks.
+This ensures the React part of your app can [benefit from the best practices](/learn/start-a-new-react-project#should-i-use-react-with-a-framework) baked into those frameworks.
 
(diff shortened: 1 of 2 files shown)


julian-alarcon:
Any status on this??

rickhanlonii:
Replaced by https://github.com/reactjs/react.dev/pull/7495 and https://github.com/reactjs/react.dev/pull/7624.

PR #6074, opened 2023-05-30
Fix hover background color in StrictMode examples

chousheng:
The StoryTray examples in StrictMode.md intend to change the background color on hover.

But it is not working. The reason is that `<li>` elements are styled with `float: left` CSS property, which makes the height of the parent `<ul>` element zero and thus makes the background color invisible.

This PR fixes the problem by adding `overflow: auto` CSS property to the `<ul>` element to clear floats, which is well supported by modern browsers.

| Before fix     | After fix      |
| -------------- | -------------- |
| <video src="https://github.com/reactjs/react.dev/assets/38355699/3a5e8785-ff2e-4de9-9728-bc3739619ce0"> | <video src="https://github.com/reactjs/react.dev/assets/38355699/153dacaf-86e2-4e63-9b5e-5f7a71329b14"> |

### DIFF
--- a/src/content/reference/react/StrictMode.md
+++ b/src/content/reference/react/StrictMode.md
@@ -196,6 +196,7 @@ ul {
   margin: 0;
   list-style-type: none;
   height: 100%;
+  overflow: auto;
 }
 
 li {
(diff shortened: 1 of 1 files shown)


chousheng:
I just signed CLA.

PR #6087, opened 2023-06-08
Update responding-to-events.md

jiangmaniu:
More clear expression


lunaleaps:
This change is making the statement less accurate are there are multiple button components. Will close for now

PR #6098, opened 2023-06-13
fix(docs): Component reference render parameters

### LINKED ISSUE #6097: Wrong doc for Component, render parameters
Check it: https://react.dev/reference/react/Component#render-parameters

The content for the render parameters of Class Components is a copy pasta of the [parameters of getSnapshotBeforeUpdate](https://react.dev/reference/react/Component#getsnapshotbeforeupdate)

It should probably say instead: `render` does not take any parameters.

image

brunoeduardodev:
Closes #6097


rickhanlonii:
Nice catch!

PR #6109, opened 2023-06-21
fix(typo): Remove meaningless words, 'nt' in common.md

xilucks:
Hi, I recently found the word "nt." while looking at the document. It was hard to understand what the word meant, and I thought it was an abbreviation, so I tried to understand it through an various document, but it was hard to understand exactly.
Also i ask for deletion because it seems to be a wrong word because it exists on a different line.
thank you.


rickhanlonii:
Thanks, this was fixed along with others in this file in https://github.com/reactjs/react.dev/pull/6167

PR #6112, opened 2023-06-21
fix(search): removing unused imports

mateoguzmana:
Removing unused imports from the Search component


rickhanlonii:
Thanks, this was fixed on main!

PR #6115, opened 2023-06-25
Fix broken page anchor

yuta-ike:
The anchor ID string in "useRef" reference page was incorrect. The correct anchor ID string should be "where-you-_can_-cause-side-effects," but it was mistakenly written as "here-you-can-cause-side-effect."

https://github.com/reactjs/react.dev/blob/feb5ba400f25b85c768ab72951e5b702bde509a8/src/content/learn/keeping-components-pure.md?plain=1#L192


harish-sethuraman (approved):
Thanks!

Issue #5960, opened 2023-04-25
Add TypeScript-related pages to new docs


markerikson:
Per https://twitter.com/mattpocockuk/status/1650611319312113665 , the new React docs currently do not provide any information on using React with TypeScript:

> One super-frustrating thing about libraries that don't ship TypeScript types themselves (i.e. React) is that they don't bother documenting their types.
> 
> In React, super important types like ComponentProps<T> aren't in the official docs, and a lot of devs don't know about them 🤯

When we did a survey of /r/reactjs in 2019, about 48% of respondents said they were using TS with React, and all the anecdata I've read since then says that percentage is now _way_ higher.

It would be extremely helpful if the docs had _some_ kind of official page on React+TS usage.

The current de-facto React+TS reference is https://react-typescript-cheatsheet.netlify.app/ , which has numerous pages on setup, basic usage, and advanced usage, as well as an abbreviated page listing some of the utility types such as `ComponentProps`.

There's enough material there that I don't expect _all_ of it to get added to the React docs.  But, at a minimum, there ought to be at least _one_ "TS Usage" page that covers things like:

- Setup and needing to install the `@types/react*` packages
- Typing function components
- Signatures for the major hooks
- Typing event handlers

and then points over to the React+TS Cheatsheet for further reference.

I could see potential for having more than one page, such as splitting it across "Setup", "Basic Usage", "Advanced Usage", and "TS Types API Reference", but that would be a stretch goal.

Dan mentioned that [at some point the React team might want to do cleanup on the TS types to better reflect the React team's view of those concepts](https://twitter.com/dan_abramov/status/1650625217222385664).  That can be done later, but **I strongly think that there needs to be _some_ TS usage material added to the React docs _today_ to provide guidance to the community on how to use React+TS properly**.

Would the React team be open to accepting a PR adding a "TS Usage" page along these lines?

CalleEklund:
I'll be happy to work on this and give a first draft. I'm new to contributing to opensource project but do have experience in react and typescript.

rickhanlonii:
Merged, thanks for the suggestion!

PR #6120, opened 2023-06-28
Adds a TypeScript overview page

orta:
Re: #5960 

Adds a 1st draft of a React + TypeScript page. Covers:

- Describing React Components
- Hooks, with a special focus on first showing how the inference works and then how to override it 
- Using common React types for Event handling, describing children and typing `style`
- Jump off points for deeper understanding

I based this from @markerikson's list in #5960 but also went through a few of my react codebases and looked at what `React.` types had been used which I think gives a pretty pragmatic set to work from. Open to opinions though!

As there is no type-checking with Sandpack, I added a feature to the Sandpack component to open any `.tsx` files in the TypeScript playground. It only shows on `*.tsx` files, of which there are no others.

It is worth noting that the Playground is going to be getting a reduced feature set over time ( https://github.com/microsoft/TypeScript-Website/issues/2804 ) but that it's _very_ unlikely to be removing the set of features used in this document (JSX support, type-acquisition, URL hash links.) Regardless, I have been working with the owner of a popular link shortener to host a fully-featured copy of the current playground outside of the TypeScript website https://github.com/gillchristian/tsplay.dev/pull/115 as a backup which URLs could be switched to in the future.

### DIFF
--- a/src/components/Layout/Footer.tsx
+++ b/src/components/Layout/Footer.tsx
@@ -290,6 +290,7 @@ export function Footer() {
           </FooterLink>
           <FooterLink href="/learn/">Quick Start</FooterLink>
           <FooterLink href="/learn/installation">Installation</FooterLink>
+          <FooterLink href="/learn/typescript">Using TypeScript</FooterLink>
           <FooterLink href="/learn/describing-the-ui">
             Describing the UI
           </FooterLink>
--- a/src/content/learn/typescript.md
+++ b/src/content/learn/typescript.md
@@ -13,8 +13,7 @@ TypeScript is a popular way to add type definitions to JavaScript codebases. Out
 
 * [TypeScript with React Components](/learn/typescript#typescript-with-react-components)
 * [Typing common hooks](/learn/typescript#typing-hooks)
-* [How to set up your editor](/learn/editor-setup)
-* [How to install React Developer Tools](/learn/react-developer-tools)
+* [Typing Events](/learn/typescript/#typing-dom-events)
 
 </YouWillLearn>
 
@@ -141,7 +140,7 @@ The [`useReducer` hook](/reference/react/useReducer) is a more complex hook that
 <Sandpack>
 
 ```tsx App.tsx active
-import {useReducer, useCallback} from 'react';
+import {useReducer} from 'react';
 
 interface State {
    count: number 
@@ -167,8 +166,8 @@ function stateReducer(state: State, action: CounterAction): State {
 export default function App() {
   const [state, dispatch] = useReducer(stateReducer, initialState);
 
-  const addFive = useCallback(() => dispatch({ type: "setCount", value: state.count + 5 }), [state.count]);
-  const reset = useCallback(() => dispatch({ type: "reset" }), []);
+  const addFive = () => dispatch({ type: "setCount", value: state.count + 5 });
+  const reset = () => dispatch({ type: "reset" });
 
   return (
     <div>
@@ -333,4 +332,46 @@ export default App = AppTSX;
 </Sandpack>
 
 
-### `useMemo` / `useCallback` {/*typing-other-hooks*/}
+### `useMemo` / `useCallback` {/*typing-memo-callback*/}
+
+The [`useMemo`](/reference/react/useMemo) and [`useCallback`](/reference/react/useCallback) hooks follow the same pattern with determining their type from the parameter passed to them. If needed you can pass a type argument to them to explicitly set the type.
+
+`useCallback` requires adding your 
+
+### `useEffect` {/*typing-useeffect*/}
+
+The [`useEffect` hook](/reference/react/useEffect) is used to perform side effects in a component. It is called after every render by default, but can be configured to only run when certain values change. The types for this hook allow for either returning a cleanup function or not returning anything.
+
+## Typing DOM Events {/*typing-dom-events*/}
+
+When working with DOM events in React, the type of the event is inferred from the event handler. However, when you want to extract a function to be passed to an event handler, you will need to explicitly set the type of the event.
+
+<Sandpack>
+
+```tsx App.tsx active
+import { useState } from 'react';
+
+export default function Form() {
+  const [value, setValue] = useState("Change me");
+
+  function handleChange(event: React.ChangeEvent<HTMLInputElement>) {
+    setValue(event.target.value);
+  }
+
+  return (
+    <>
+      <input value={value} onChange={handleChange} />
+      <p>Value: {value}</p>
+    </>
+  );
+}
+```
+
+```js App.js hidden
+import AppTSX from "./App.tsx";
+export default App = AppTSX;
+```
+
+</Sandpack>
+
+There are many types of events provided in React - the full list can be found [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/b580df54c0819ec9df62b0835a315dd48b8594a9/types/react/index.d.ts#L1247C1-L1373) which is based on the [most popular events from the DOM](https://developer.mozilla.org/en-US/docs/Web/Events). If you need to use an event that is not included in this list, you can use the `React.SyntheticEvent` type, which is the base type for all events.
--- a/src/content/learn/typescript.md
+++ b/src/content/learn/typescript.md
@@ -342,7 +342,11 @@ The [`useMemo`](/reference/react/useMemo) and [`useCallback`](/reference/react/u
 
 The [`useEffect` hook](/reference/react/useEffect) is used to perform side effects in a component. It is called after every render by default, but can be configured to only run when certain values change. The types for this hook allow for either returning a cleanup function or not returning anything.
 
-## Typing DOM Events {/*typing-dom-events*/}
+## Useful Types {/*useful-types*/}
+
+There is quite an expansive set of types which come from the `@types/react` package, it is worth a read when you feel comfortable with how React and TypeScript interact. You can find them [in React's folder in DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts). We will cover a few of the more common types here.
+
+### DOM Events {/*typing-dom-events*/}
 
 When working with DOM events in React, the type of the event is inferred from the event handler. However, when you want to extract a function to be passed to an event handler, you will need to explicitly set the type of the event.
 
@@ -374,4 +378,56 @@ export default App = AppTSX;
 
 </Sandpack>
 
-There are many types of events provided in React - the full list can be found [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/b580df54c0819ec9df62b0835a315dd48b8594a9/types/react/index.d.ts#L1247C1-L1373) which is based on the [most popular events from the DOM](https://developer.mozilla.org/en-US/docs/Web/Events). If you need to use an event that is not included in this list, you can use the `React.SyntheticEvent` type, which is the base type for all events.
+There are many types of events provided in the React types - the full list can be found [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/b580df54c0819ec9df62b0835a315dd48b8594a9/types/react/index.d.ts#L1247C1-L1373) which is based on the [most popular events from the DOM](https://developer.mozilla.org/en-US/docs/Web/Events). If you need to use an event that is not included in this list, you can use the `React.SyntheticEvent` type, which is the base type for all events.
+
+### Children {/*typing-children*/}
+
+There are three common paths to describing the children of a component. The first is to use the `React.ReactNode` type, which is a union of all the possible types that can be passed as children in JSX:
+
+```ts
+interface ModalRendererProps {
+  title: string;
+  children: React.ReactNode;
+}
+```
+
+The second is to use the `React.ReactElement` type, which is only JSX elements and not JavaScript primitives like strings or numbers:
+
+```ts
+interface ModalRendererProps {
+  title: string;
+  children: React.ReactElement;
+}
+```
+
+The third is to use `React.PropsWithChildren` which is a utility type that takes an object type and adds a optional `children` field of `React.ReactNode` to it:
+
+```ts
+type ModalRendererProps = React.PropsWithChildren<{
+  title: string;
+}>
+```
+
+Note, that you cannot use TypeScript to describe that the children are a certain type of elements. You can see all three of these in action with the type-checker in [this TypeScript playground](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAJQKYEMDG8BmUIjgIilQ3wChSB6CxYmAOmXRgDkIATJOdNJMGAZzgwAFpxAR+8YADswAVwGkZMJFEzpOjDKw4AFHGEEBvUnDhphwADZsi0gFw0mDWjqQBuUgF9yaCNMlENzgAXjgACjADfkctFnYkfQhDAEpQgD44AB42YAA3dKMo5P46C2tbJGkvLIpcgt9-QLi3AEEwMFCItJDMrPTTbIQ3dKywdIB5aU4kKyQQKpha8drhhIGzLLWODbNs3b3s8YAxKBQAcwXpAThMaGWDvbH0gFloGbmrgQfBzYpd1YjQZbEYARkB6zMwO2SHSAAlZlYIBCdtCRkZpHIrFYahQYQD8UYYFA5EhcfjyGYqHAXnJAsIUHlOOUbHYhMIIHJzsI0Qk4P9SLUBuRqXEXEwAKKfRZcNA8PiCfxWACecAAUgBlAAacFm80W-CU11U6h4TgwUv11yShjgJjMLMqDnN9Dilq+nh8pD8AXgCHdMrCkWisVoAet0R6fXqhWKhjKllZVVxMcavpd4Zg7U6Qaj+2hmdG4zeRF10uu-Aeq0LBfLMEe-V+T2L7zLVu+FBWLdLeq+lc7DYFf39deFVOotMCACNOCh1dq219a+30uC8YWoZsRyuEdjkevR8uvoVMdjyTWt4WiSSydXD4NqZP4AymeZE072ZzuUeZQKheRKGoMUbX4AB1YARAAYXfNlgEEJkoFVbheBgGRzjgGQMNkBQABpzAgBC0K4bE4AgTAXWCFBpDYOBpAgN8Kjsb0mj9cCoPfKoumDEpQ2cEC2OEaDGKqLIjC8dI8xyfJY2iBNhOqWpU2Y9M4gEoSk2kbMuMkgk1I46Qi1eVtewNKs8T0ioql0iDBP0htHk2bsPnbfsuyMns61cwcAXMmz1I4AzKSGCybCstcEBCgLMjgaFIqs3ckVWOLAq3ZKTyxHEkr8uzYuyyyDOvUlyTSoLqQASRwaRgDQFBsVVO4oHZThpBQBY8Jq6RiP4ei6OfRlmRgqpcvY-L+QGf8gA).
+
+### Style Props {/*typing-style-props*/}
+
+When using inline styles in React, you can use `React.CSSProperties` to describe the object passed to the `style` prop. This type is a union of all the possible CSS properties, and is a good way to ensure you are passing valid CSS properties to the `style` prop, and to get auto-complete in your editor.
+
+```ts
+interface MyComponentProps {
+  style: React.CSSProperties;
+}
+```
+
+## Further learning {/*further-learning*/}
+
+This guide has covered the basics of using TypeScript with React, but there is a lot more to learn. We recommend the following resources:
+
+ - [The TypeScript handbook](https://www.typescriptlang.org/docs/handbook/) is the official documentation for TypeScript, and covers most key language features.
+
+ - [The TypeScript release notes](https://devblogs.microsoft.com/typescript/) covers a each new features in-depth.
+
+ - [React TypeScript Cheatsheet](https://react-typescript-cheatsheet.netlify.app/) is a community-maintained cheatsheet for using TypeScript with React, covering a lot of useful edge cases and providing more breadth than this document.
+
+ - [TypeScript Community Discord](discord.com/invite/typescript) is a great place to ask questions and get help with TypeScript and React issues.
--- a/src/content/learn/typescript.md
+++ b/src/content/learn/typescript.md
@@ -10,6 +10,7 @@ TypeScript is a popular way to add type definitions to JavaScript codebases. Out
 </Intro>
 
 <YouWillLearn>
+
 * [TypeScript with React Components](/learn/typescript#typescript-with-react-components)
 * [Typing common hooks](/learn/typescript#typing-hooks)
 * [How to set up your editor](/learn/editor-setup)
@@ -186,6 +187,7 @@ export default function App() {
 import AppTSX from "./App.tsx";
 export default App = AppTSX;
 ```
+
 </Sandpack>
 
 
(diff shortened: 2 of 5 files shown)


src/content/learn/typescript.md
```diff
@@ -0,0 +1,433 @@
...
+    <div>
+      <p>Current object: {object.kind}</p>
+    </div>
+  )
+}
+```
+
+### `useRef` {/*typing-useref*/}
+
+The [`useRef` hook](/reference/react/useRef) returns a mutable ref object whose `.current` property is initialized to the passed argument. The returned object will persist for the full lifetime of the component.
```
phryneas, line 297: It's probably important to explain that `useRef<Value>` will give you an immutable ref, and `useRef<Value | null>` (or an initial value) will give you a mutable one.

Reference: https://github.com/typescript-cheatsheets/react#option-2-mutable-value-ref

src/content/learn/typescript.md
```diff
@@ -0,0 +1,433 @@
...
+<Sandpack>
+
+```tsx App.tsx active
+import { useState } from 'react';
+
+export default function Form() {
+  const [value, setValue] = useState("Change me");
+
+  function handleChange(event: React.ChangeEvent<HTMLInputElement>) {
+    setValue(event.target.value);
```
phryneas, line 362: It might generally be a best practice to show `currentTarget` in event handlers, even though it's not really the point here.

```suggestion
    setValue(event.currentTarget.value);
```

phryneas:
I love this! I left some nitpicks, but generally: 💯!

src/content/learn/typescript.md
```diff
@@ -0,0 +1,433 @@
...
+
+
+## Hooks {/*typing-hooks*/}
+
+The type definitions from `@types/react-dom` include types for the built-in hooks, so you can use them in your components without any additional setup. They are built to take into account the code you write in your component, so you will get inferred types a lot of the time and ideally do not need to handle the minutiae of providing the types. However, we can look at a few examples of how to provide types for hooks.
+
+### `useState` {/*typing-usestate*/}
+
+The [`useState` hook](/reference/react/useReducer) will re-use the value passed in as the initial state to determine what the type of the value should be. For example:
```
tom-sherman, line 104:
```suggestion
The [`useState` hook](/reference/react/useState) will re-use the value passed in as the initial state to determine what the type of the value should be. For example:
```

src/content/learn/typescript.md
```diff
@@ -0,0 +1,433 @@
...
+
+interface State {
+   count: number 
+};
+
+type CounterAction =
+  | { type: "reset" }
+  | { type: "setCount"; value: State["count"] }
+
+const initialState: State = { count: 0 };
```
tom-sherman, line 153: Is this a good place to showcase `satisfies`? Or would that be distracting from the example?

src/content/learn/typescript.md
```diff
@@ -0,0 +1,433 @@
...
+```
+
+```js App.js hidden
+import AppTSX from "./App.tsx";
+export default App = AppTSX;
+```
+
+</Sandpack>
+
+This technique works when you have an obvious default value - but often with context you do not, and [in those cases `null` is an appropriate default value](/reference/react/useContext#specifying-a-fallback-default-value). However, you likely would not want `null` to be in the type when consuming the type, our recommendation is to have the hook do a runtime check for it's existence and throw an error when not present:
```
tom-sherman, line 254: This example is IMO an anti-pattern if it's included it should at least be marked as such. That's why the doc that's linked is specifically talking about specifying a default value _other than_ null.

I think this can be done pretty simply by re-iterating what the linked doc says eg.

> Often, instead of null, there is some more meaningful value you can use as a default. For the occasions where this not...

src/content/learn/typescript.md
tom-sherman: This is 100% the weirdest part about the React types and definitely deserves proper attention.

src/content/learn/typescript.md
orta: Satisfies is an interesting case, but not what we want here, for example:

```ts
type State = {
  something?: string
  count: number
}

const initial: State = { count: 0 }
const initial2 = { count: 0} satisfies State
```

Would mean that the reducer does not know about `something?: string` on `initial2` but does in `initial`. Because it doesnt modify the type exposed in initial - just the raw literal as the type.

src/content/learn/typescript.md
```diff
@@ -0,0 +1,433 @@
...
+```
+
+</Sandpack>
+
+The type describing your component's props can be as simple or as complex as you need, though they should probably be an object type. You can learn about how TypeScript describes objects in [Object Types](https://www.typescriptlang.org/docs/handbook/2/objects.html) but you may also be interested in using [Union Types](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#union-types) to describe a prop that can be one of a few different types and the [Creating Types from Types](https://www.typescriptlang.org/docs/handbook/2/types-from-types.html) guide for more advanced use cases.
+
+
+## Hooks {/*typing-hooks*/}
+
+The type definitions from `@types/react-dom` include types for the built-in hooks, so you can use them in your components without any additional setup. They are built to take into account the code you write in your component, so you will get inferred types a lot of the time and ideally do not need to handle the minutiae of providing the types. However, we can look at a few examples of how to provide types for hooks.
```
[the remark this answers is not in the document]
orta: Yes! However, as I mentioned `@types/react-dom` at the top, I felt like introducing this further below (and maybe needing to explain the distinction) didn't make sense

src/content/learn/typescript.md
orta: Thanks, I'll internalize and add

src/content/learn/typescript.md
```diff
@@ -0,0 +1,433 @@
...
+
+The [`useEffect` hook](/reference/react/useEffect) is used to perform side effects in a component. It is called after every render by default, but can be configured to only run when certain values change. The types for this hook allow for either returning a cleanup function or not returning anything.
+
+## Useful Types {/*useful-types*/}
+
+There is quite an expansive set of types which come from the `@types/react` package, it is worth a read when you feel comfortable with how React and TypeScript interact. You can find them [in React's folder in DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts). We will cover a few of the more common types here.
+
+### DOM Events {/*typing-dom-events*/}
+
+When working with DOM events in React, the type of the event is inferred from the event handler. However, when you want to extract a function to be passed to an event handler, you will need to explicitly set the type of the event.
```
tom-sherman, line 351: It would be nice to explain how you choose the `event` type ie. Why `ChangeEvent`? And why `HTMLInputElement`?

src/content/learn/typescript.md
orta: I think that seems reasonable, having read MDN on the topic - the distinction is new to me 👍🏻

src/content/learn/typescript.md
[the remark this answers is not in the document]
tom-sherman: I don't understand 🤔 folks need to install both packages to get the full types, right? If you just installed @types/react-dom you wouldn't have types for useState and friends.

src/content/learn/typescript.md
orta: Yeah, I think this makes sense 👍🏻

src/content/learn/typescript.md
orta: There's a very real chance this is just me - but in the current codebase I work in there are zero cases where context has a default value (there's also a lot of `{} as any` or `null as any`)   - in some of the Artsy codebases I looked at it was about half of the `createContext` had a value like those too.

I'm sure I can go back and maybe add a fake object which conforms to the same shape with NOOP funcs etc, but doing all that dancing for no useful value seems like busywork vs this technique which raises in the right location when there's an issue and acts like it is supposed to act inside components

But I agree, there's a contradiction in what this side is saying vs what the docs maybe there's a way to re-frame this paragraph 👍🏻

markerikson:
At first glance this looks fantastic!

Will come back and give a proper review later

src/content/learn/typescript.md
tom-sherman: In my experience it's a _very prevalent_ anti pattern 😅 if it counts for anything pretty sure I've seen React core team members say similar things about context before, but I don't want to derail this discussion too much about context idioms 😆 

It's a pattern common enough to be documented here, just needs to match the tone of "this is probably not a good thing to be doing" IMO.

src/content/learn/typescript.md
tom-sherman: Yeah makes sense. I think there are cases where `satisfies` is useful but it probably would warrant a whole separate example of it's own.

src/content/learn/typescript.md
```diff
@@ -0,0 +1,433 @@
+---
+title: Using TypeScript
+re: https://github.com/reactjs/react.dev/issues/5960
+---
+
+<Intro>
+
+TypeScript is a popular way to add type definitions to JavaScript codebases. Out of the box, TypeScript [supports JSX](/learn/writing-markup-with-jsx) and you can get full React support by adding [`@types/react-dom`](https://www.npmjs.com/package/@types/react-dom) to your project.
```
eps1lon, line 8: `@types/react` is the main one since that's what you mostly interact with (hooks, utility types etc). `@types/react-dom` is onyl for `react-dom`

src/content/learn/typescript.md
```diff
@@ -0,0 +1,433 @@
...
+```
+
+```js App.js hidden
+import AppTSX from "./App.tsx";
+export default App = AppTSX;
+```
+
+</Sandpack>
+
+The type describing your component's props can be as simple or as complex as you need, though they should probably be an object type. You can learn about how TypeScript describes objects in [Object Types](https://www.typescriptlang.org/docs/handbook/2/objects.html) but you may also be interested in using [Union Types](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#union-types) to describe a prop that can be one of a few different types and the [Creating Types from Types](https://www.typescriptlang.org/docs/handbook/2/types-from-types.html) guide for more advanced use cases.
```
eps1lon, line 95: I think we can make the wording stronger here about the object type. There's not a use case for using anything else here I can think of. If there is, we should mention it.

src/content/learn/typescript.md
```diff
@@ -0,0 +1,433 @@
...
+}
+```
+
+### `useRef` {/*typing-useref*/}
+
+The [`useRef` hook](/reference/react/useRef) returns a mutable ref object whose `.current` property is initialized to the passed argument. The returned object will persist for the full lifetime of the component.
+
+`useRef` is often used to access the underlying DOM element of a component, but it can also be used to store any mutable value.
+
+When interacting with the DOM, the type of the ref should be set to the type of the underlying DOM element. The naming rule is `HTML` + the name of the element + `Element`, for example `HTMLDivElement` or `HTMLButtonElement`. You can see the full list from TypeScript 5.1 [here](https://github.com/microsoft/TypeScript/blob/a3773ec590c4f0308d546f0e65818cd0d12402f3/src/lib/dom.generated.d.ts#L26899-L27012). These are provided globally by the "DOM" lib, which is included by default in TypeScript projects.
```
eps1lon, line 301: We should rather link to MDN or some other DOM documentation. TS provides the declarations. But what interfaces there are is not a TS thing but a HTML thing.

src/content/learn/typescript.md
```diff
@@ -0,0 +1,433 @@
...
+The second is to use the `React.ReactElement` type, which is only JSX elements and not JavaScript primitives like strings or numbers:
+
+```ts
+interface ModalRendererProps {
+  title: string;
+  children: React.ReactElement;
+}
+```
+
+The third is to use `React.PropsWithChildren` which is a utility type that takes an object type and adds a optional `children` field of `React.ReactNode` to it:
```
eps1lon, line 403: I would not recommend this since it was only added to ease migration. Especially because the name is inaccurate since it's optional children not required children. `children?: React.ReactNode` also makes you more familiar with the concept of a React node and that you maybe want to use a slot other than `children`.

eps1lon (commented):
Really nice. Has the right balance between too much and too little types.

src/content/learn/typescript.md
eps1lon: Extra attention to `target` vs `currentTarget` would be nice since this is a recurring issue where people think `target` should be `T` instead of just `EventTarget` for every event

src/content/learn/typescript.md
tom-sherman: It also gently introduces the idea of having ReactNode in another prop if you need it. Maybe there should be a specific example for this actually.

src/content/learn/typescript.md
```diff
@@ -0,0 +1,433 @@
...
...n are a certain type of elements. You can see all three of these in action with the type-checker in [this TypeScript playground](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAJQKYEMDG8BmUIjgIilQ3wChSB6CxYmAOmXRgDkIATJOdNJMGAZzgwAFpxAR+8YADswAVwGkZMJFEzpOjDKw4AFHGEEBvUnDhphwADZsi0gFw0mDWjqQBuUgF9yaCNMlENzgAXjgACjADfkctFnYkfQhDAEpQgD44AB42YAA3dKMo5P46C2tbJGkvLIpcgt9-QLi3AEEwMFCItJDMrPTTbIQ3dKywdIB5aU4kKyQQKpha8drhhIGzLLWODbNs3b3s8YAxKBQAcwXpAThMaGWDvbH0gFloGbmrgQfBzYpd1YjQZbEYARkB6zMwO2SHSAAlZlYIBCdtCRkZpHIrFYahQYQD8UYYFA5EhcfjyGYqHAXnJAsIUHlOOUbHYhMIIHJzsI0Qk4P9SLUBuRqXEXEwAKKfRZcNA8PiCfxWACecAAUgBlAAacFm80W-CU11U6h4TgwUv11yShjgJjMLMqDnN9Dilq+nh8pD8AXgCHdMrCkWisVoAet0R6fXqhWKhjKllZVVxMcavpd4Zg7U6Qaj+2hmdG4zeRF10uu-Aeq0LBfLMEe-V+T2L7zLVu+FBWLdLeq+lc7DYFf39deFVOotMCACNOCh1dq219a+30uC8YWoZsRyuEdjkevR8uvoVMdjyTWt4WiSSydXD4NqZP4AymeZE072ZzuUeZQKheRKGoMUbX4AB1YARAAYXfNlgEEJkoFVbheBgGRzjgGQMNkBQABpzAgBC0K4bE4AgTAXWCFBpDYOBpAgN8Kjsb0mj9cCoPfKoumDEpQ2cEC2OEaDGKqLIjC8dI8xyfJY2iBNhOqWpU2Y9M4gEoSk2kbMuMkgk1I46Qi1eVtewNKs8T0ioql0iDBP0htHk2bsPnbfsuyMns61cwcAXMmz1I4AzKSGCybCstcEBCgLMjgaFIqs3ckVWOLAq3ZKTyxHEkr8uzYuyyyDOvUlyTSoLqQASRwaRgDQFBsVVO4oHZThpBQBY8Jq6RiP4ei6OfRlmRgqpcvY-L+QGf8gA).
```
tom-sherman, line 411: > Note, that you cannot use TypeScript to describe that the children are a certain type of elements.

I think this should be clarified and expanded upon. The question of "How do I only accept specific components as children?" is very frequently asked and so deserves some attention.

Secondly, `{children: number}` is valid usage but IMO "certain type of elements" alludes to it being not allowed which is wrong.

src/content/learn/typescript.md
orta: I agree with this, I did look and https://developer.mozilla.org/en-US/docs/Web/HTML/Element was the best list of the element types but didn't contain the actual names. This glossary is _ok_ https://developer.mozilla.org/en-US/docs/Web/API#h_2 but not a great thing to link to as it's got a bunch of extras in it

The TypeScript link provided is _really_ good because it provides a list of `"a": HTMLAnchorElement`  which takes _the thing you know_ and shows you the _thing you need to figure out_. 

I think these some word-smithing which could highlight how these are named based on the Web APIs though

src/content/learn/typescript.md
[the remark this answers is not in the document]
orta: `@types/react` is a dependency of `@types/react-dom` so you get both with that one import. But maybe you're right - I should probably just mention both at the top and the right one down here

src/content/learn/typescript.md
phryneas: It could actually be safest to show a horrible
```js
type State = {
  something?: string
  count: number
}

const initial = { count: 0 } satisfies State as State
```
as from my experience, if someone were to move the `initial` declaration into the component before the `useReducer` call, TypeScript code flow analysis might otherwise kick in and the reducer state ends up typed as `{ count: 0 }`, since that it "the only value it could have had on that code path".

src/content/learn/typescript.md
```diff
@@ -0,0 +1,433 @@
...
+```
+
+</Sandpack>
+
+
+### `useMemo` / `useCallback` {/*typing-memo-callback*/}
+
+The [`useMemo`](/reference/react/useMemo) and [`useCallback`](/reference/react/useCallback) hooks follow the same pattern with determining their type from the parameter passed to them. If needed you can pass a type argument to them to explicitly set the type.
+
+`useCallback` requires adding your
```
kj800x, line 339: Just in case you missed it, I noticed this sentence is only half written

src/content/learn/typescript.md
```diff
@@ -0,0 +1,433 @@
...
+</Sandpack>
+
+The type describing your component's props can be as simple or as complex as you need, though they should probably be an object type. You can learn about how TypeScript describes objects in [Object Types](https://www.typescriptlang.org/docs/handbook/2/objects.html) but you may also be interested in using [Union Types](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#union-types) to describe a prop that can be one of a few different types and the [Creating Types from Types](https://www.typescriptlang.org/docs/handbook/2/types-from-types.html) guide for more advanced use cases.
+
+
+## Hooks {/*typing-hooks*/}
+
+The type definitions from `@types/react-dom` include types for the built-in hooks, so you can use them in your components without any additional setup. They are built to take into account the code you write in your component, so you will get inferred types a lot of the time and ideally do not need to handle the minutiae of providing the types. However, we can look at a few examples of how to provide types for hooks.
+
+### `useState` {/*typing-usestate*/}
```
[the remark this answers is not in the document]
tom-sherman: I don't think this is a good pattern to be recommending in the docs. Usually you want to wrap your setters into an "event-based" instead of a "mutation-based" API to preserve state encapsulation. eg. `onChangeTeamSubmission` or `onSymbolKeysChange`.

Passing setters down can also indicate your state should be moved to be more local.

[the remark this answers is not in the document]
tom-sherman: In general I think this doc should only reference patterns defined elsewhere in the docs, it shouldn't introduce new ones.

src/content/learn/typescript.md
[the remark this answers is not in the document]
orta: I've updated this to include both types/react and types/react-dom at the intro, and specify the right one here 👍🏻

src/content/learn/typescript.md
orta: Switched this down to just node and element in the examples, updated the playground to account for this also

src/content/learn/typescript.md
orta: I've re-framed this section a bit (more "this _is_ the DOM naming system", not "it kinda works like this", and also linked to the MDN source of them

src/content/learn/typescript.md
orta: @eps1lon can you please re-explain this point? I'm not quite sure I get the distinction on `target` vs `currentTarget`  enough to get the value

src/content/learn/typescript.md
orta: Thanks!

src/content/learn/typescript.md
eps1lon: We should do this in a follow-up since it depends on the event. It doesn't block this from being merged.

eps1lon (approved):
I think this is a really great start. Thank you!

src/content/learn/typescript.md
phryneas: Looks good to me :)

src/content/learn/typescript.md
totomakers: Can we juste use the type `const handleChange: ComponentProps<'input'>['onChange'] = (e) => {}` instead manualy typing all args ?

src/content/learn/typescript.md
tom-sherman: Sounds good!

src/content/learn/typescript.md
```diff
@@ -0,0 +1,474 @@
...
+
+```js App.js hidden
+import AppTSX from "./App.tsx";
+export default App = AppTSX;
+```
+</Sandpack>
+
+ <Note>
+
+These sandboxes can handle TypeScript code, but they do not run the type-checker. This means you can amend the TypeScript sandboxes to learn, but you won't get any type errors or warnings. To get type-checking, you can use the [TypeScript Playground](https://www.typescriptlang.org/play) or use a more fully-featured online sandbox.
```
rickhanlonii, line 57: What do we need in order to support this? It would be nice to have an example here where the type is wrong to see the benefit.

src/content/learn/typescript.md
```diff
@@ -0,0 +1,474 @@
+---
+title: Using TypeScript
+re: https://github.com/reactjs/react.dev/issues/5960
+---
+
+<Intro>
+
+TypeScript is a popular way to add type definitions to JavaScript codebases. Out of the box, TypeScript [supports JSX](/learn/writing-markup-with-jsx) and you can get full React Web support by adding [`@types/react`](https://www.npmjs.com/package/@types/react) and [`@types/react-dom`](https://www.npmjs.com/package/@types/react-dom) to your project.
```
rickhanlonii, line 8: Is it worth mentioning that the types are official / maintained by the React team?

src/content/learn/typescript.md
```diff
@@ -0,0 +1,474 @@
...
+```
+
+</Sandpack>
+
+The type describing your component's props can be as simple or as complex as you need, though they should be an object type described with either a `type` or `interface`. You can learn about how TypeScript describes objects in [Object Types](https://www.typescriptlang.org/docs/handbook/2/objects.html) but you may also be interested in using [Union Types](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#union-types) to describe a prop that can be one of a few different types and the [Creating Types from Types](https://www.typescriptlang.org/docs/handbook/2/types-from-types.html) guide for more advanced use cases.
+
+
+## Hooks {/*typing-hooks*/}
+
+The type definitions from `@types/react` include types for the built-in hooks, so you can use them in your components without any additional setup. They are built to take into account the code you write in your component, so you will get inferred types a lot of the time and ideally do not need to handle the minutiae of providing the types. However, we can look at a few examples of how to provide types for hooks.
```
rickhanlonii, line 101: If you're new to typescript, it's probably not clear what "inferred" means. What about a Deep Dive section here describing this?

src/content/learn/typescript.md
```diff
@@ -0,0 +1,474 @@
...
+
+### `useState` {/*typing-usestate*/}
+
+The [`useState` hook](/reference/react/useState) will re-use the value passed in as the initial state to determine what the type of the value should be. For example:
+
+```ts
+const [enabled, setEnabled] = useState(false);
```
rickhanlonii, line 108: It's not clear that this is using the inferred type, what about:

```suggestion
// Infer the type as "boolean"
const [enabled, setEnabled] = useState(false);
```

src/content/learn/typescript.md
```diff
@@ -0,0 +1,474 @@
...
+```
+
+Will assign the type of `boolean` to `enabled`, and `setEnabled` will be a function accepting either a `boolean` argument, or a function that returns a `boolean`. If you want to explicitly provide a type for the state, you can do so by providing a type argument to the `useState` call:
+
+```ts 
+const [enabled, setEnabled] = useState<boolean>(false);
```
rickhanlonii, line 114:
```suggestion
// Explicitly set the type to "boolean"
const [enabled, setEnabled] = useState<boolean>(false);
```

src/content/learn/typescript.md
```diff
@@ -0,0 +1,474 @@
...
+ - `const initialState: State` provides a type for the initial state, and also the type which is used by `useReducer` by default.
+ - `stateReducer(state: State, action: CounterAction): State` sets the types for the reducer function's arguments and return value.
+
+A more explicit alternative to setting the type on `initialState` is to provide a type argument to `useReducer`:
+
+```ts
+const initialState = { count: 0 };
+
+export default function App() {
+  const [state, dispatch] = useReducer<State>(stateReducer, initialState);
```
rickhanlonii, line 207: Where is State coming from?

src/content/learn/typescript.md
```diff
@@ -0,0 +1,474 @@
...
+```js App.js hidden
+import AppTSX from "./App.tsx";
+export default App = AppTSX;
+```
+
+</Sandpack>
+
+This technique works when you have an default value which makes sense - but there are occasionally cases when you do not, and in those cases `null` can feel reasonable as a default value. However, to allow the type-system to understand your code, you need to explicitly set `ContextShape | null` on the `createContext`. 
+
+This causes the knock-on issue that you need to eliminate the `| null` in the type for context consumers. Our recommendation is to have the hook do a runtime check for it's existence and throw an error when not present:
```
rickhanlonii, line 257: I'm not familiar with the phrase "knock-on"?

src/content/learn/typescript.md
```diff
@@ -0,0 +1,474 @@
...
+
+  return (
+    <div>
+      <p>Current object: {object.kind}</p>
+    </div>
+  )
+}
+```
+
+### `useRef` {/*typing-useref*/}
```
rickhanlonii, line 298: Refs and effects are "escape hatches" in the learn docs so let's skip them?

rickhanlonii (commented):
Great start! I have a few comments!

src/content/learn/typescript.md
orta: I did look into it, and it's possible, https://github.com/codesandbox/sandpack/discussions/237#discussioncomment-3870506  preview: here https://codesandbox.io/s/github/danilowoz/sandpack-tsserver

It is a bit of a trade-off though, TypeScript is a ~6mb download for the client (.d.ts files and the compiler) and it's something that will occasionally want version bumping to keep up-to-date with the typescript runtime (though it could try re-use the version used by this repo) - all that for a few code samples is generally not worth it _IMO_.

For the TS website, I hardcoded all this information at build time by making https://shikijs.github.io/twoslash/ - but those code samples are static which I think is a bit restrictive WRT how the rest of the site works?

src/content/learn/typescript.md
```diff
@@ -0,0 +1,474 @@
...
+
+* [TypeScript with React Components](/learn/typescript#typescript-with-react-components)
+* [Typing common hooks](/learn/typescript#typing-hooks)
+* [Common types from `@types/react`](/learn/typescript/#useful-types)
+* [Further learning locations](/learn/typescript/#further-learning)
+
+</YouWillLearn>
+
+
+All of the production-grade frameworks mentioned in [Start a New React Project](/learn/start-a-new-react-project) offer support for using TypeScript with React, we recommend consulting their documentation for more information on setup. This guide assumes you have your project configured to support writing `*.tsx` TypeScript React files, and have finished the Quick Start guide.
```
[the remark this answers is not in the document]
orta: If it's a single sentence, it might not need that, but I'm open to that - you don't want to be too hand-hold-y here because how setting up TypeScript on each platform is probably a bit different and can change over time

src/content/learn/typescript.md
orta: Is it "classed" as being officially maintained by the React team? /cc @eps1lon ?

I thought it wasn't, but I've been out of the game for a few years. If it is, this is definitely something we should add 👍🏻

src/content/learn/typescript.md
orta: I like this idea!

src/content/learn/typescript.md
orta: It's meant to be chopped, I should do an `import` to indicate its something "off-camera" - thanks

src/content/learn/typescript.md
orta: 👍🏻

src/content/learn/typescript.md
orta: I agree, and I think this can be a React team decision which I'll be happy with. I have no pushback on removing `useEffect` as it's really only there because it would be the only one _not_ there.

`useRef` though has this tricky _"either it's mutable, or it's not"_ thing that you kinda have to learn or it doesn't really fit in your head when you first try it - because for using it inline in JSX you need to really have that `|null` . I think beginners get stuck on this bit, and "react typescript useRef" finding this link would describe everything they need to know.

src/content/learn/typescript.md
eps1lon: > Is it "classed" as being officially maintained by the React team? 

[Yes](https://react.dev/community/team#sebastian-silbermann)

src/content/learn/typescript.md
eps1lon: It's probably best add to the `useRef` docs in a "TS usage" section or something. At least for the APIs we consider escape hatches.

phryneas: Tbh., it would feel very weird if this was the one hook where the TS instructions would be buried somewhere completely else. In that case, TS usage instrucations would need to be added to pretty much every single API page.

Maybe just keep it here at the bottom of the page under "other hooks"?

eps1lon: I don't think this page will be digestable if we add TS usage instructions for every API. Mid-term, React with TypeScript will fit naturally into docs and won't be an after-thought that a single page would imply. This page is really just to get you started.

I'm fine if this is just temporary until we figure out TS in demos but if the intention is to keep TS in a single page, I'd like to discuss this a bit more.

phryneas: I'd *love* to see TS everywhere in the docs, and then no complaints from me - but so far it seemed like a very hard thing to have any form of TS in the docs, so I didn't expect there to be a high chance for that.

src/content/learn/typescript.md
eps1lon: I also added a link to https://www.typescriptlang.org/docs/handbook/type-inference.html here. We call out how inferred types behave on the specific hook sections. Should we explain inference more here?

src/content/learn/typescript.md
eps1lon: Removed "knock-on": https://github.com/reactjs/react.dev/pull/6120/commits/42fd5a4c694981f2764b87932fc190977ec32124. Looked redundant to me.

src/content/learn/typescript.md
eps1lon: Moved `useRef` to its reference page while mentioning at the bottom that individual pages may contain TS usage if necessary. I removed the `useEffect` section since it didn't seem like it contained relevant information. 
[`c0e6963` (#6120)](https://github.com/reactjs/react.dev/pull/6120/commits/c0e6963a549fda1b845817d5e61ca8fe11d170b1)

src/content/learn/typescript.md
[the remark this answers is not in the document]
eps1lon: Added "add to existing" project and moved everything under "installation" section

orta:
Thanks @eps1lon! ( Sorry been really heads down the last few weekends in a row )

src/components/Layout/Footer.tsx
```diff
@@ -290,6 +290,7 @@ export function Footer() {
           </FooterLink>
           <FooterLink href="/learn/">Quick Start</FooterLink>
           <FooterLink href="/learn/installation">Installation</FooterLink>
+          <FooterLink href="/learn/typescript">Using TypeScript</FooterLink>
```
rickhanlonii, line 293: Let's remove this, the links here are for the top-level grouping, and "Using Typescript" is under "Installation" listed above.

src/content/learn/typescript.md
rickhanlonii: We can follow up with the DeepDive section.

rickhanlonii (approved):
This is a great start, let's ship this initial version and iterate as needed.

PR #6125, opened 2023-06-30
Update correct link common.md

NGHdeveloper:
Correct link to https://developer.mozilla.org/en-US/docs/Web/API/AnimationEvent/pseudoElement


lunaleaps:
Thanks for catching!

PR #6127, opened 2023-07-02
Remove the 'App router is not ready for production' part.

thatgurkangurk:
The app router got stable in Next 13.4, so the docs shouldn't state that it isn't ready for production.


lunaleaps:
This has been addressed in https://github.com/reactjs/react.dev/pull/6318

PR #6131, opened 2023-07-05
Update your-first-component.md

NGHdeveloper:
Correction of the plural "elements" by its singular "element", for better understanding of the text.


rickhanlonii:
Thanks but in the example code this is referring to, there are multiple Profile components so the plural here is accurate.

PR #6133, opened 2023-07-05
add rendercon kenya 2023 to conferences

orama254:
Added Rendercon Kenya 2023 to the list of conferences


orama254:
Hello @harish-sethuraman 

Your assistance with this will be appreciated.

PR #6134, opened 2023-07-05
Fix rounded corners for code sandbox.

cpojer:
The rounded corners in code examples don't match the ones in code sandboxes. This PR fixes it by aligning the rounded borders of the code sandbox (`<CustomPreset>`) with the ones from code examples (`CodeBlock`).

<table>
<tr>
 <td>
	**Before**
 <td>
    **After**
<tr>
 <td>
CleanShot 2023-07-06 at 07 01 44@2x
 <td>
CleanShot 2023-07-06 at 07 01 23@2x
</table>

Screenshot of the whole box:
CleanShot 2023-07-06 at 07 02 20@2x

Notice how in the before screenshot the bottom box has a different rounded border, and in the after screenshot the borders are the same.

This has been driving me a bit crazy since the new docs launched 😅

### DIFF
--- a/src/components/MDX/Sandpack/CustomPreset.tsx
+++ b/src/components/MDX/Sandpack/CustomPreset.tsx
@@ -74,7 +74,7 @@ const SandboxShell = memo(function SandboxShell({
   return (
     <>
       <div
-        className="shadow-lg dark:shadow-lg-dark rounded-lg"
+        className="shadow-lg dark:shadow-lg-dark rounded-2xl"
         ref={containerRef}
         style={{
           contain: 'content',


src/components/MDX/Sandpack/CustomPreset.tsx
```diff
@@ -74,7 +74,7 @@ const SandboxShell = memo(function SandboxShell({
   return (
     <>
       <div
-        className="shadow-lg dark:shadow-lg-dark rounded-lg"
+        className="shadow-lg dark:shadow-lg-dark rounded-2xl"
```
harish-sethuraman, line 77: makes sense tbh 😅 I've also felt the same just want someone from the core team to confirm cc: @rickhanlonii

rickhanlonii:
I think this is intentional so there is some separation in the design of the two. @gaearon @lebo do you remember?

PR #6137, opened 2023-07-07
Fix root-unmount returns

solar3070:
The `root.unmount() Returns` in "hydrateRoot" reference was incorrect. The correct Returns should be "`root.unmount` returns `undefined`".


rickhanlonii:
Nice catch!

PR #6138, opened 2023-07-07
RTL Languages Support

A7med3bdulBaset:
This PR addresses left-to-right (RTL) language styles, directionality, and other relevant adjustments for the React documentation.

## Types of changes

- Updated `tailwindcss` to the latest version to utilize logical properties.
- Added the `.isRTL` property to the `siteConfig` object <sup>[note 1](#note-1)</sup>.
- Set the `dir` attribute on the `<html>` element to either `ltr` or `rtl` based on the `isRTL` property.
- Maintained the `dir` attributes of `<InlineCode />`, `<CodeBlock />`, and `<Sandpack />` as `ltr` regardless of the `isRTL` property.
- Converted physical and directional properties to logical properties <sup>[+96% of browsers support logical properties](https://caniuse.com/css-logical-props)</sup>:
  - Replaced `pl-*`, `pr-*`, `ml-*`, and `mr-*` with `ps-*`, `pe-*`, `ms-*`, and `me-*`.
  - Replaced `text-left` and `text-right` with `text-start` and `text-end`.
  - Replaced `left-*` and `right-*` with `start-*` and `end-*`.
  - Replaced `rounded-l-*` and `rounded-r-*` with `rounded-s-*` and `rounded-e-*`.
  - Replaced `border-l-*` and `border-r-*` with `border-s-*` and `border-e-*`.
  - And so on.
- For the components `<IconNavArrow />`, `<IconArrow />`, `<IconArrowSmall />`, and `<IconChevron />`, added two additional values to the `displayDirection` prop: `start` and `end`. These values behave differently based on the page direction and should be used instead of `left` and `right` values <sup>[note 2](#note-2)</sup>.
- This PR does not affect anything in LTR languages.

## Live demo:

See this [live demo](https://rtl-react-dev.vercel.app/learn) of the documentation with some Arabic data (from [ar.react.dev](https://github.com/reactjs/ar.react.dev))

## Notes:

<details>
<summary id="note-1">Note 1: The `.isRTL` property</summary>

We had several options to choose from:
- Added a new property to the `siteConfig` object and set it manually for each language.
  ```js
  exports.siteConfig = {
    // ...
    languageCode: 'ar',
    isRTL: true,
    // ...
  };
  ```

- Used language detection like [rtl-detect](https://www.npmjs.com/package/rtl-detect) to automatically detect the language direction and set the property.
  ```js
  const siteConfig = {
    // ...
    languageCode: 'ar',
    hasLegacySite: true,
    isRTL: false, // Do NOT ever set this to true. It is deprecated.
    // ...
  };

  const rtlDetect = require('rtl-detect');
  siteConfig.isRTL = rtlDetect.isRtlLang(siteConfig.languageCode);

  exports.siteConfig = siteConfig;
  ```

</details>

<details>
<summary id="note-2">Note 2: New values for the <code>displayDirection</code> prop</summary>

```jsx
export const IconNavArrow = memo<
  JSX.IntrinsicElements['svg'] & {
    /**
     * The direction the arrow should point.
     * `start` and `end` are relative to the current locale.
     * For example, in LTR, `start` is left and `end` is right.
     */
    displayDirection: 'start' | 'end' | 'right' | 'left' | 'down';
  }
>(function IconNavArrow({displayDirection = 'end', className}) {
  const classes = cn(
    'duration-100 ease-in transition',
    {
      'rotate-0': displayDirection === 'down',
      'rotate-90': displayDirection === 'left',
      '-rotate-90': displayDirection === 'right',
      'rotate-90 rtl:-rotate-90': displayDirection === 'start',
      '-rotate-90 rtl:rotate-90': displayDirection === 'end',
    },
    className
  );

  return (
    <svg
      // ...
      className={classes}
      //...
  );
});

```

</details>

@gaearon, @rickhanlonii, @acdlite, and React Community, could you please review this PR?


A7med3bdulBaset:
@lunaleaps and @mattcarrollcode, could you please review this? As Arabic users, we have difficulties reading the Arabic docs with LTR directions.

src/components/Icon/IconChevron.tsx
```diff
@@ -7,15 +7,20 @@ import cn from 'classnames';
...
+     * for example, in LTR, `start` is left and `end` is right.
+     */
+    displayDirection: 'start' | 'end' | 'right' | 'left' | 'up' | 'down';
   }
 >(function IconChevron({className, displayDirection}) {
   const classes = cn(
     {
       'rotate-0': displayDirection === 'down',
-      'rotate-90': displayDirection === 'left',
+      'rotate-90': displayDirection === 'left' || displayDirection === 'start',
```
lunaleaps, line 21: Why didn't we create a separate `start` key here? re:
`'rotate-90 rtl:-rotate-90': displayDirection === 'start'`

src/components/Layout/HomeContent.js
```diff
@@ -1537,7 +1529,7 @@ function Cover({background, children}) {
         src={background}
         width={500}
         height={263}
-        alt=""
+        alt="cover"
```
lunaleaps, line 1532: Is this useful to add? Guess I'm not sure the context of when this is displayed

src/components/Layout/HomeContent.js
```diff
@@ -1612,13 +1606,13 @@ function Thumbnail({video}) {
                 key={i}
                 className="h-8 w-8 border-2 shadow-md border-gray-70 object-cover rounded-full"
                 src={src}
-                alt=""
+                alt="Speaker"
```
lunaleaps, line 1609: I'd say this be better left empty or perhaps actually add an alt description to our `videos` data list, or perhaps more generically "video thumbnail"?

lunaleaps (approved):
Changes look good to me! Just a couple small q's about alt properties and a CSS clarification.

Thanks for doing this work!

Question - is there a lint we can set up to ensure we don't start re-introducing left/right css properties?

src/components/Layout/HomeContent.js
A7med3bdulBaset: I pushed a commit to remove it. thanks

src/components/Layout/HomeContent.js
A7med3bdulBaset: This too has been deleted

src/components/Icon/IconChevron.tsx
A7med3bdulBaset: Nice catch
 resolved

lunaleaps:
Thanks for addressing, looks like just conflicting files and then I can merge!

A7med3bdulBaset:
> Thanks for addressing, looks like just conflicting files and then I can merge!

Resolved

rickhanlonii:
Wow this is awesome, thanks a ton for all that went into this!

A7med3bdulBaset:
> @lunaleaps: **Question:** Is there a lint we can set up to ensure we don't start re-introducing left/right CSS properties?

Regarding icons, we can eliminate the `"left"` and `"right"` in `displayDirection` to enforce future updates to employ `"start"` and `"end"` directions.

For Tailwindcss classes, I have released [`eslint-plugin-rtl-friendly`](https://github.com/A7med3bdulBaset/eslint-plugin-rtl-friendly) to serve this purpose. It reports in cases where you utilize physical properties like `pl-*`, prompting you to use `ps-*` instead, with auto-fix by `eslint --fix`

Note: Up til now, It doesn't detect like `className={cn(...)}`. It works only for string up to this moment `v.0.1.0` and doesn't lint eslint file ([StyleLint](https://stylelint.io/) is needed for that).

We have the option to install and extend it. By configuring `no-physical-properties` as `error` (while the default "recommended config" is "warn") to enhance strictness.

What do you think? @lunaleaps @rickhanlonii

lunaleaps:
Thanks @A7med3bdulBaset for looking into that! I'm curious if we do still need some left/right properties and how we exclude them from the lint? Otherwise, this looks like a great idea!

A7med3bdulBaset:
> Thanks @A7med3bdulBaset for looking into that! I'm curious if we do still need some left/right properties and how we exclude them from the lint? Otherwise, this looks like a great idea!
 
Could you provide an example of left/right properties we need where start/end wouldn't resolve the issue?

A: There are NONE.

The only places we seem to need them are in the `<CodeBlock />`, `<Sandbox/>`, and similar components. However, even in those cases, we use `dir="ltr"`, so `ps-*` becomes `pl-*`, regardless of the page direction.

PR #6144, opened 2023-07-09
add create-react-app in main docs

DIWAKARKASHYAP:


DIWAKARKASHYAP:
check

rickhanlonii:
Thanks but we no longer recommend using Create React App for new projects. If one of the recommended production-ready frameworks don't work for your use case we recommend using Vite (see the "Can I use React without a framework?" section on the page lower down).

PR #6148, opened 2023-07-11
fix cloneElement code example

mykolavlasov:
Fixing the "log" of the `cloneElement` function. It's mentioned that provided props(2nd argument) are merged with `element.props`(1st argument), but the code snippet was wrong.


mykolavlasov:
@harish-sethuraman Could you please check it? Seems that this repo doesn't allow to specify the list of reviewers

PR #6149, opened 2023-07-11
docs: fix a broken link in react-labs-june-2022

MatanBobi:
The link was broken (due to the domain changes).
I believe that a redirect should probably happen but I'm not sure where is that defined.
If I can help with the redirect rules please let me know :)


rickhanlonii:
@mattcarrollcode @lunaleaps @gaearon looks like blog link redirects are broken, any way to fix that?

MatanBobi:
@rickhanlonii Looks like the redirects are defined here:
https://github.com/reactjs/legacy.reactjs.org/blob/main/vercel.json

and it is missing a redirect for that specific page. I can create a PR there if it helps :)

rickhanlonii:
The old site needs the .html at the end I believe, so maybe the fix here is just merging this broken link.

MatanBobi:
Thanks @rickhanlonii :)

PR #6153, opened 2023-07-13
doc(cloneElement): fix example log

MarvelSQ:
## bug

https://react.dev/reference/react/cloneElement#cloneelement

Screenshot 2023-07-13 at 15 08 15

missing override props from `{ isHighlighted: true }`


lunaleaps:
This looks like it's been fixed, thanks for raising!

PR #6158, opened 2023-07-15
fix: wrong parameter name defined in explanation

MatanBobi:
The function signature had a parameter named `commitTime` but the parameters explanation referenced it as `endTime`. This PR aligns them to be `commitTime`:
```
function onRender(id, phase, actualDuration, baseDuration, startTime, commitTime) {
  // Aggregate or log render timings...
}
```


rickhanlonii:
Nice catch!

PR #6164, opened 2023-07-19
Fix typo: "server-only" to "client-only"

smikitky:
Fixed a heading that did not correctly reflect the following text.

This section describes how to display server-side **fallback** for client-only **content**. Saying "server-only **content**" is inappropriate.


rickhanlonii:
Yeah nice catch. I think "client only" is what's meant here, given the code throws on the server.

PR #6167, opened 2023-07-19
Remove junk lines in common.md

smikitky:
During the translation of `common.md`, I found three lines that were clearly junk, so I removed them.


rickhanlonii:
Wow nice find, thanks!

PR #6173, opened 2023-07-21
Add UI for canary releases

rickhanlonii:
## Overview

Trying an idea for adding badges for APIs that are in Canary.

## Light

Screenshot 2023-08-10 at 5 07 04 PM

## Dark

Screenshot 2023-08-10 at 5 07 01 PM


src/components/Layout/Sidebar/SidebarLink.tsx
```diff
@@ -71,12 +71,17 @@ export function SidebarLink({
...
-        })}>
-        {title}
-      </span>
+      <div>
+        {title}{' '}
+        {canary && (
+          <span
+            title="This feature is available in the latest React Canary"
+            className="ml-2 border border-purple-40 dark:border-purple-50 rounded-full pt-0 pb-0.5 px-2 text-sm text-purple-40 dark:text-purple-50 leading-3">
+            canary
```
poteto, line 80: I prefer the look of it capitalized, but I get that that's super subjective haha. Looks great regardless! Very easy to scan and tell. 

I like the purple but was wondering if the colors should be more "traditional"? I think yellow is typically associated with canary (yellow Chrome icon, yellow Meta app icons, etc).

src/components/Layout/Sidebar/SidebarLink.tsx
```diff
@@ -71,12 +71,17 @@ export function SidebarLink({
...
-          'text-gray-400 dark:text-gray-500': wip,
-        })}>
-        {title}
-      </span>
+      <div>
+        {title}{' '}
+        {canary && (
+          <span
+            title="This feature is available in the latest React Canary"
+            className="ml-2 border border-purple-40 dark:border-purple-50 rounded-full pt-0 pb-0.5 px-2 text-sm text-purple-40 dark:text-purple-50 leading-3">
```
poteto, line 79: I like the purple but was wondering if the colors should be more "traditional" (while fitting our color palette)? I think yellow is typically associated with canary (yellow Chrome icon, yellow Meta app icons, etc).

src/components/Layout/Sidebar/SidebarLink.tsx
rickhanlonii: Capitalizing it makes the pill feel bigger I think.

Good call on the yellow. Here's yellow in our current color scheme, which looks more brown than yellow:

Screenshot 2023-07-21 at 12 39 40 PM

Screenshot 2023-07-21 at 12 39 30 PM

poteto: I think you might have pasted the wrong screenshot, I still see purple

rickhanlonii: @poteto oops, fixed!

src/components/Layout/Sidebar/SidebarLink.tsx
```diff
@@ -15,7 +15,7 @@ interface SidebarLinkProps {
   selected?: boolean;
   title: string;
   level: number;
-  wip: boolean | undefined;
+  canary: boolean | undefined;
```
harish-sethuraman, line 18:
```suggestion
  canary?: boolean;
```
may be no difference at all 🙈

rickhanlonii: fixed thanks

src/components/Layout/Sidebar/SidebarLink.tsx
rickhanlonii: Ok I pushed the yellow

sophiebits:
Do you think the sidebar badges are too prominent? We want them to make people turn away from those items, not towards them. I wonder if a small monochrome icon would be better (I think that’s what MDN does?)

src/content/reference/react/directives.md
```diff
@@ -1,10 +1,17 @@
 ---
 title: "Directives"
+canary: true
 ---
 
+<Canary>
+
+Directives are available in the [React Canary](https://react.dev/community/versioning-policy#canary-channel) release channel. They are needed only if you're [using React Server Components](/learn/start-a-new-react-project#bleeding-edge-react-frameworks) or building a library compatible with them.
```
sophiebits, line 8: nit: the channel is called Canary, not React Canary. “release channel” would need to be linked before “React”

eps1lon:
>  I wonder if a small monochrome icon would be better (I think that’s what MDN does?)

They add a small, blue-colored icon:
Screenshot 2023-07-25 at 14 31 29

With Twitter's rebranding we might get away with a blue bird (i.e. Canary) 😉

rickhanlonii:
@sophiebits @eps1lon @poteto pushed a change to use an icon and the callout is gray to match. What do you think?

src/components/Layout/Sidebar/SidebarLink.tsx
```diff
@@ -72,12 +73,16 @@ export function SidebarLink({
...
-          'text-gray-400 dark:text-gray-500': wip,
-        })}>
-        {title}
-      </span>
+      <div>
+        {title}{' '}
+        {canary && (
+          <IconCanary
+            title="This feature is available in the latest Canary"
+            className="ml-2 text-gray-30 dark:text-gray-60 inline-block w-4 h-4"
```
sophiebits, line 81: can we add `align-[-2px]`

src/components/Layout/Sidebar/SidebarLink.tsx
```diff
@@ -72,12 +73,16 @@ export function SidebarLink({
...
+          />
+        )}
```
harish-sethuraman, line 83: I feel like we can choose to hide the icon near title as we are showing a callout + icon on the left pane?

src/components/PageHeading.tsx
```diff
@@ -27,7 +31,15 @@ function PageHeading({
       <div className="max-w-4xl ml-0 2xl:mx-auto">
         {breadcrumbs ? <Breadcrumbs breadcrumbs={breadcrumbs} /> : null}
         <H1 className="mt-0 text-primary dark:text-primary-dark -mx-.5 break-words">
-          {title}
+          <div className="flex items-center">
```
sophiebits, line 34: probably don't want this to be flex (if it's a long title, the icon should probably be on the last line)

src/content/reference/react/use-client.md
```diff
@@ -1,12 +1,12 @@
...
+canary: true
 ---
 
-<Note>
+<Canary>
 
-These directives are needed only if you're [using React Server Components](/learn/start-a-new-react-project#bleeding-edge-react-frameworks) or building a library compatible with them.
-
-</Note>
+`'use client'` available in the [Canary](https://react.dev/community/versioning-policy#canary-channel) release channel. It is needed only if you're [using React Server Components](/learn/start-a-new-react-project#bleeding-edge-react-frameworks) or building a library compatible with them.
```
sophiebits, line 8: still think its weird to talk about these directives being "available" in a certain channel since react doesn't interpret these. I would personally stick to only mentioning RSC in the text (but the callout icon/heading seems ok)

sophiebits (commented):
can you fix the vertical alignment? eg it's 1px or 2px too low in the sidebar and it's too high in the title (should be baseline-aligned). looks ok in the callout

otherwise seems good to me

src/content/reference/react/use-client.md
rickhanlonii: I see, yeah makes sense for these ones.

rickhanlonii:
@sophiebits fixed

PR #6174, opened 2023-07-21
Fix lint warnings

rickhanlonii:
Also updates the lint rules so the dependency warning is an error, and ignores `_` as unused.

### DIFF
--- a/src/components/Layout/HomeContent.js
+++ b/src/components/Layout/HomeContent.js
@@ -8,12 +8,10 @@ import {
   useState,
   useContext,
   useId,
-  Fragment,
   Suspense,
   useEffect,
   useRef,
   useTransition,
-  useReducer,
 } from 'react';
 import cn from 'classnames';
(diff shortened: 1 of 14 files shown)


src/components/Layout/HomeContent.js
```diff
@@ -90,19 +79,23 @@ function FullBleed({children}) {
...
     minute: 'numeric',
   });
-  let [, forceUpdate] = useReducer((n) => n + 1, 0);
   useEffect(() => {
-    const timeout = setTimeout(forceUpdate, nextMinute - Date.now());
+    const msPerMinute = 60 * 1000;
+    let nextMinute = Math.floor(+date / msPerMinute + 1) * msPerMinute;
+
+    const timeout = setTimeout(() => {
+      if (Date.now() > nextMinute) {
```
sophiebits, line 92: seems like if your system clock moved backwards (even slightly) then the clock display would stop advancing.

PR #6177, opened 2023-07-24
Add use hook API reference page

mattcarrollcode:
Add initial documentation for the use hook as a API reference page

https://react-dev-git-fork-mattcarrollcode-use-ref-fbopensource.vercel.app/reference/react/use


src/content/reference/react/use.md
```diff
@@ -0,0 +1,769 @@
...
+title: use
+---
+
+<Wip>
+This section is incomplete. See also the [draft use RFC](https://github.com/acdlite/rfcs/blob/first-class-promises/text/0000-first-class-support-for-promises.md#usepromise).
+</Wip>
+
+[//]: # (TODO: replace the `<Note></Note>` below with `<Canary></Canary>` once https://github.com/reactjs/react.dev/pull/6173 is merged)
+<Note>
+The `use` hook is only supported when using [React Server Components](/learn/start-a-new-react-project#bleeding-edge-react-frameworks). React Server Components are only available in React’s canary and experimental channels. Learn more about [React's release channels here](/community/versioning-policy#all-release-channels).
```
sophiebits, line 11: Isn't `use` supported in canary and experimental without RSC?

src/content/reference/react/use.md
```diff
@@ -0,0 +1,769 @@
...
+
+Call `use` in your component to read the value of a resource like a [promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) or [context](/learn/passing-data-deeply-with-context).
+
+```jsx
+import { use } from "react";
+
+function Message({ messagePromise }) {
+  const message = use(messagePromise);
+  const theme = use(ThemeContext);
+  //...
```
sophiebits, line 40: nit: we normally write `// ...` with a space (looks like the only occurrences of `//...` are in the tic-tac-toe tutorial 😄)

src/content/reference/react/use.md
```diff
@@ -0,0 +1,769 @@
...
+</Wip>
+
+[//]: # (TODO: replace the `<Note></Note>` below with `<Canary></Canary>` once https://github.com/reactjs/react.dev/pull/6173 is merged)
+<Note>
+The `use` hook is only supported when using [React Server Components](/learn/start-a-new-react-project#bleeding-edge-react-frameworks). React Server Components are only available in React’s canary and experimental channels. Learn more about [React's release channels here](/community/versioning-policy#all-release-channels). 
+</Note>
+
+<Intro>
+
+`use` is a React Hook that lets you read the value of a resource like a [promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) or [context](/learn/passing-data-deeply-with-context).
```
sophiebits, line 16: (is the term "resource" an intentional creation here?)

src/content/reference/react/use.md
```diff
@@ -0,0 +1,769 @@
...
+```jsx
+import { use } from "react";
+
+function Message({ messagePromise }) {
+  const message = use(messagePromise);
+  const theme = use(ThemeContext);
+  //...
+```
+
+Unlike all other React Hooks, the `use` Hook can be called within conditional statements like loops and `if` statements. Like other React Hooks, the parent function of a Hook must be a React component or Hook.
```
sophiebits, line 43: nit: "parent function" here and below is a bit jargony; do you think "must be called from" or "called within" suffices? or is that confusing because it sounds like it could be a called from an intermediate helper

src/content/reference/react/use.md
```diff
@@ -0,0 +1,769 @@
...
+
+---
+
+## Usage {/*usage*/}
+
+### Reading promises with `use` {/*reading-promises-with-use*/}
+
+`use` functions similarly to the `await` with JavaScript promises but is compatible with React's renderer and [`Suspense`](/reference/react/Suspense). Browser APIs or data fetching libraries that return a promise can be passed to `use` to read the value of the promise.
+
+In the example below we'll use <CodeStep step={1}>[`navigator.storage.estimate()`](https://developer.mozilla.org/en-US/docs/Web/API/StorageManager/estimate)</CodeStep>, a browser API that provides information on how much storage is being used by the current page and how much space is available. <CodeStep step={1}>`navigator.storage.estimate()`</CodeStep> is an async function that returns a Promise which resolves once the information is available.
```
lunaleaps, line 71: Inconsistent capitalization of promise. I'm not sure which it should be

src/content/reference/react/use.md
```diff
@@ -0,0 +1,769 @@
...
+
+function Message({ messagePromise }) {
+  const message = use(messagePromise);
+  const theme = use(ThemeContext);
+  //...
+```
+
+Unlike all other React Hooks, the `use` Hook can be called within conditional statements like loops and `if` statements. Like other React Hooks, the parent function of a Hook must be a React component or Hook.
+
+A component that passes a promise to `use` integrates with [`Suspense`](/reference/react/Suspense) and [`ErrorBoundary`](/reference/react/Component#catching-rendering-errors-with-an-error-boundary). When a pending promise is passed to `use`, the fallback of the closest Suspense Boundary will be displayed. When the promise is resolved, the component calling `use` will be rendered using the data from the resolved promise. The rendered component will then replace the Suspense Boundary fallback. If the promise passed to `use` is rejected, the fallback of the nearest Error Boundary will be displayed.
```
sophiebits, line 45: > A component that passes a promise to use integrates with Suspense and ErrorBoundary.

this is very passive voice, and it's also confusing to me because it makes a statement about the user's components that at a glance could be read as a requirement that the user must fulfill. maybe just "When called with a promise, the `use` Hook integrates with"?

looks like house style is `Suspense boundary` and `error boundary`

I also think this description isn't quite right because it doesn't account for startTransition.

src/content/reference/react/use.md
```diff
@@ -0,0 +1,769 @@
...
+
+---
+
+## Troubleshooting {/*troubleshooting*/}
+
+### I'm getting this error: "An unsupported type was passed to use" {/*im-getting-this-error-an-unsupported-type-was-passed-to-use*/}
+
+The `use` hook only supports [JavaScript promises](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) and React's [Context](/learn/passing-data-deeply-with-context) as parameters.
+
+### I'm getting this error: "Suspense Exception: This is not a real error!" {/*suspense-exception-error*/}
```
lunaleaps, line 725: This is a really strange error message...not necessarily about this but should we follow up with a better error?

src/content/reference/react/use.md
```diff
@@ -0,0 +1,769 @@
...
+
+In the example below we'll use <CodeStep step={1}>[`navigator.storage.estimate()`](https://developer.mozilla.org/en-US/docs/Web/API/StorageManager/estimate)</CodeStep>, a browser API that provides information on how much storage is being used by the current page and how much space is available. <CodeStep step={1}>`navigator.storage.estimate()`</CodeStep> is an async function that returns a Promise which resolves once the information is available.
+
+```js [[1,6, "navigator.storage.estimate()"], [2,6, "storagePromise"], [3,6, "useState"], [4,7, "use"], [5,8, "storage.quota"]]
+"use client";
+
+import { use, useState } from "react"
+
+function StorageInfo() {
+  const [storagePromise] = useState(navigator.storage.estimate());
```
sophiebits, line 79: Is this usage correct @acdlite? I thought at present the promise needs to be created before render.

src/content/reference/react/use.md
```diff
@@ -0,0 +1,769 @@
...
+## Usage {/*usage*/}
+
+### Reading promises with `use` {/*reading-promises-with-use*/}
+
+`use` functions similarly to the `await` with JavaScript promises but is compatible with React's renderer and [`Suspense`](/reference/react/Suspense). Browser APIs or data fetching libraries that return a promise can be passed to `use` to read the value of the promise.
+
+In the example below we'll use <CodeStep step={1}>[`navigator.storage.estimate()`](https://developer.mozilla.org/en-US/docs/Web/API/StorageManager/estimate)</CodeStep>, a browser API that provides information on how much storage is being used by the current page and how much space is available. <CodeStep step={1}>`navigator.storage.estimate()`</CodeStep> is an async function that returns a Promise which resolves once the information is available.
+
+```js [[1,6, "navigator.storage.estimate()"], [2,6, "storagePromise"], [3,6, "useState"], [4,7, "use"], [5,8, "storage.quota"]]
+"use client";
```
sophiebits, line 74: I think we use single quotes. Also there are two semicolons missing in this block.

src/content/reference/react/use.md
```diff
@@ -0,0 +1,769 @@
...
+  // ✅ This works! The parent function is a component. 
+  const message = use(messagePromise);
+  //...
+```
+
+
+
+### How do I prevent the component calling `use` from abruptly appearing in the UI? {/*how-do-i-prevent-the-component-calling-use-from-abruptly-appearing-in-the-ui*/}
+
+Components that pass a promise to `use` must be wrapped in [`Suspense`](/reference/react/Suspense) to prevent the Component from "popping in" when the promise is resolved. In the examples below, the `Message` Component calls the `use` hook. Because the `Message` Component is not wrapped in [`Suspense`](/reference/react/Suspense) the `Message` component will abruptly pop into the page when `messagePromise` is resolved:
```
lunaleaps, line 750: When we saying "popping in", are we talking about a layout jump?

src/content/reference/react/use.md
```diff
@@ -0,0 +1,769 @@
...
+<Pitfall>
+
+When passing promises to `use`, the promise must be cached with the `cache` function, a state variable, or context to prevent the promise from being recreated on every render.
+
+```js
+function Component() {
+  use(new Promise(...)) // ❌ promise will be recreated on every render
+  // ...
+```
+instead, pass the promise to a state variable or context or wrap the function that returns the promise in `cache`:
```
sophiebits, line 98: Instead

src/content/reference/react/use.md
```diff
@@ -0,0 +1,769 @@
...
+2. Next, <CodeStep step={2}>`storagePromise`</CodeStep> is passed to <CodeStep step={4}>`use`</CodeStep> to read the resolved value of the promise.
+3. Finally, the resolved value of the promise can now be used safely. This includes surfacing data read by <CodeStep step={4}>`use`</CodeStep> like <CodeStep step={5}>`storage.quota`</CodeStep> in the JSX returned by the component.
+
+<Pitfall>
+
+When passing promises to `use`, the promise must be cached with the `cache` function, a state variable, or context to prevent the promise from being recreated on every render.
+
+```js
+function Component() {
+  use(new Promise(...)) // ❌ promise will be recreated on every render
```
sophiebits, line 95: ;

src/content/reference/react/use.md
```diff
@@ -0,0 +1,769 @@
...
+root.render(
+  <StrictMode>
+    <App />
+  </StrictMode>
+);
+```
+
+```js data-fetching-lib.js hidden
+export function fetchMessage() {
+  return new Promise((resolve) => setTimeout(resolve, 3000, "⚛️"));
```
sophiebits, line 254: 3000 is too long, maybe 1000 or 1500

src/content/reference/react/use.md
```diff
@@ -0,0 +1,769 @@
...
+```
+</Sandpack>
+
+<Solution />
+
+</Recipes>
+
+
+
+### Reading context with `use` {/*reading-context-with-use*/}
```
sophiebits, line 276: this should probably answer the question somewhere: is there a reason to use useContext over use?

also, examples don't show conditional use

src/content/reference/react/use.md
```diff
@@ -0,0 +1,769 @@
...
+
+</Sandpack>
+
+### Streaming data from the server to the client {/*streaming-data-from-server-to-client*/}
+
+Data can be streamed from the server to the client by passing a promise as a prop from a server component to a client component:
+
+```js App.js
+import { fetchMessage } from "./lib.js";
+import { Message } from "./message.js";
```
sophiebits, line 415: nit: would be nicer if Message had a name that more obviously was a component

src/content/reference/react/use.md
```diff
@@ -0,0 +1,769 @@
...
+
+```js
+// message.js
+"use client";
+
+import { use } from "react";
+
+export function Message({ messagePromise }) {
+  const messageContent = use(messagePromise);
+  return <p>Here is the message: {messageContent}</p>;
```
sophiebits, line 437: ideally this could show something that would need a client component (even if behind a `...`)

src/content/reference/react/use.md
```diff
@@ -0,0 +1,769 @@
...
+Data can be streamed from the server to the client by passing a promise as a prop from a server component to a client component:
+
+```js App.js
+import { fetchMessage } from "./lib.js";
+import { Message } from "./message.js";
+
+export default function App() {
+  const messagePromise = fetchMessage();
+  return (
+    <Suspense fallback={<p>waiting for message...</p>}>
```
sophiebits, line 420: nit
```suggestion
    <Suspense fallback={<p>Waiting for message...</p>}>
```

src/content/reference/react/use.md
```diff
@@ -0,0 +1,769 @@
...
+"use client";
+
+import { use } from "react";
+
+export function Message({ messagePromise }) {
+  const messageContent = use(messagePromise);
+  return <p>Here is the message: {messageContent}</p>;
+}
+```
+Because `Message` is wrapped in [`Suspense`](/reference/react/Suspense) the fallback will be displayed until the promise is resolved. When the promise is resolve the value will be unwrapped by the `use` Hook and the `Message` component will replace the [`Suspense`](/reference/react/Suspense) fallback.
```
sophiebits, line 440: typos

src/content/reference/react/use.md
```diff
@@ -0,0 +1,769 @@
...
+  return <p>Here is the message: {messageContent}</p>;
+}
+```
+
+```js App.js
+import { Suspense, useState } from "react";
+import { Message } from "./message.js";
+
+function fetchMessage() {
+  return new Promise((resolve) => setTimeout(resolve, 3000, "⚛️"));
```
sophiebits, line 460: faster

src/content/reference/react/use.md
```diff
@@ -0,0 +1,769 @@
...
+      <Message messageContent={messageContent} />
+    </Suspense>
+  );
+}
+```
+
+But using `await` in a [Server Component](/reference/react/components#server-components) will block the rendering of the [Server Component](/reference/react/components#server-components) until the `await` statement is finished.
+
+Passing a promise from a server component to a client component prevents the promise from blocking the rendering of the server component. This enables a suspense fallback to be displayed to the user. When the promise resolves, the value of the promise is read by `use`. This value is used to render the client component which replaces the suspense fallback.
+
```
sophiebits, line 538: What about if it was split so that Suspense was in a parent component?

src/content/reference/react/use.md
```diff
@@ -0,0 +1,769 @@
...
+In some cases a promise passed to `use` could be rejected. You can handle rejected promises by either:
+
+1. [Displaying an error to users with ErrorBoundary.](#displaying-an-error-to-users-with-errorboundary)
+2. [Providing an alternative value with `Promise.catch`](#providing-an-alternative-value-with-promise-catch)
+
+<Pitfall>
+`use` cannot be called in a try-catch block. Instead of a try-catch block [wrap your component in an Error Boundary](#displaying-an-error-to-users-with-errorboundary), or [provide an alternative value to use with the promise's `.catch` method](#providing-an-alternative-value-with-promise-catch).
+</Pitfall>
+
+#### Displaying an error to users with `ErrorBoundary` {/*displaying-an-error-to-users-with-errorboundary*/}
```
sophiebits, line 552:
```suggestion
#### Displaying an error to users with an error boundary {/*displaying-an-error-to-users-with-errorboundary*/}
```

and within this section; "`ErrorBoundary`" is not a concept

src/content/reference/react/use.md
```diff
@@ -0,0 +1,769 @@
...
+import { Message } from "./message.js";
+
+function fetchMessage() {
+  return new Promise((resolve, reject) => {
+    setTimeout(reject, 3000, new Error("no new message found."));
+  }).catch((error) => {
+    console.log(error.toString());
+    return "no new message found.";
+  });
+}
```
sophiebits, line 651: This is really clunky. Does this need to show a timeout? Plain .catch might be simpler.

src/content/reference/react/use.md
```diff
@@ -0,0 +1,769 @@
...
+```
+</Sandpack>
+
+---
+
+## Troubleshooting {/*troubleshooting*/}
+
+### I'm getting this error: "An unsupported type was passed to use" {/*im-getting-this-error-an-unsupported-type-was-passed-to-use*/}
+
+The `use` hook only supports [JavaScript promises](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) and React's [Context](/learn/passing-data-deeply-with-context) as parameters.
```
sophiebits, line 723:
```suggestion
The `use` hook only supports [JavaScript promises](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) and React [context](/learn/passing-data-deeply-with-context) as parameters.
```

src/content/reference/react/use.md
```diff
@@ -0,0 +1,769 @@
...
+
+### I'm getting this error: "Suspense Exception: This is not a real error!" {/*suspense-exception-error*/}
+
+You are either calling `use` outside of a React component or Hook function, or calling `use` in a try-catch block. If you are calling `use` inside a `try-catch` block wrap your component in an error boundary, or call the promise's `catch` to catch the error and resolve the promise with another value. [See these examples](#dealing-with-rejected-promises).
```
sophiebits, line 727: nit: try–catch should use an en dash (and not be in backticks unless you split `try` and `catch`)

src/content/reference/react/use.md
```diff
@@ -0,0 +1,769 @@
...
+function Message({messagePromise}) {
+  function download() {
+    // ❌ `use`'s parent function is not a Component or Hook!
+    const message = use(messagePromise);
+    //...
+```
+instead, call `use` outside any component closures, where the parent function is a component or Hook:
+```jsx
+function Message({messagePromise}) {
+  // ✅ This works! The parent function is a component.
```
sophiebits, line 741:
```suggestion
  // ✅ The parent function is a component. 
```

src/content/reference/react/use.md
```diff
@@ -0,0 +1,769 @@
...
+### I'm getting this error: "Suspense Exception: This is not a real error!" {/*suspense-exception-error*/}
+
+You are either calling `use` outside of a React component or Hook function, or calling `use` in a try-catch block. If you are calling `use` inside a `try-catch` block wrap your component in an error boundary, or call the promise's `catch` to catch the error and resolve the promise with another value. [See these examples](#dealing-with-rejected-promises).
+
+If you are calling `use` outside a React component or Hook function, move the `use` call to a React component or Hook function.
+
+```jsx
+function Message({messagePromise}) {
+  function download() {
+    // ❌ `use`'s parent function is not a Component or Hook!
```
sophiebits, line 734:
```suggestion
    // ❌ `use`'s parent function is not a component or Hook
```

src/content/reference/react/components.md
```diff
@@ -22,3 +22,23 @@ React exposes a few built-in components that you can use in your JSX.
 ## Your own components {/*your-own-components*/}
 
 You can also [define your own components](/learn/your-first-component) as JavaScript functions.
+
+--
+
+## Client and Server Components {/*client-and-server-components*/}
```
sophiebits, line 28: Can you make clear to people not using RSC what (if anything) to take away from this section?

src/content/reference/react/components.md
```diff
@@ -22,3 +22,23 @@ React exposes a few built-in components that you can use in your JSX.
...
+
+There are two main types of components in React: Server Components and Client Components. Both types follow the same basic principles of React - they are reusable, composable, and stateful - but they are used in different scenarios and have different capabilities.
+
+### Server Components {/*server-components*/}
+
+Server Components are a new addition to React that allow developers to leverage the power of server-side rendering while maintaining the interactivity and richness of a client-side React application. They are rendered only on the server and have zero impact on the client-side bundle size, making them an excellent choice for parts of your application that are not interactive and don't need to be updated in response to user interaction.
+
+### Client Components {/*client-components*/}
+
+Client Components are the traditional React components that run on the client. They can access the full range of React features, including state, effects, and access to the DOM. They are essential for creating rich, interactive features where users expect immediate feedback.
```
sophiebits, line 42: Not sure if this is temporary but we should be working towards server-first docs, not describing server components as a "new addition" and client components as "traditional"

src/content/reference/react/components.md
```diff
@@ -22,3 +22,23 @@ React exposes a few built-in components that you can use in your JSX.
...
+
+--
+
+## Client and Server Components {/*client-and-server-components*/}
+
+<Wip>
+[React Server Components](/learn/start-a-new-react-project#bleeding-edge-react-frameworks) are only available in React’s canary and experimental channels. Learn more about [React's release channels here](/community/versioning-policy#all-release-channels).
+</Wip>
+
+There are two main types of components in React: Server Components and Client Components. Both types follow the same basic principles of React - they are reusable, composable, and stateful - but they are used in different scenarios and have different capabilities.
```
sophiebits, line 34: did we discuss whether to capitalize these? in 'use client' page, I didn't and I would lean towards not

src/content/reference/react/use.md
sophiebits: also confused about what this section means

src/content/reference/react/use.md
sophiebits: the full text is

```
Suspense Exception: This is not a real error! It's an implementation detail of
`use` to interrupt the current render. You must either rethrow it immediately,
or move the `use` call outside of the `try/catch` block. Capturing without
rethrowing will lead to unexpected behavior.

To handle async errors, wrap your component in an error boundary, or call the
promise's `.catch` method and pass the result to `use`
```

sophiebits (commented):
added a few thoughts

src/content/reference/react/components.md
```diff
@@ -22,3 +22,23 @@ React exposes a few built-in components that you can use in your JSX.
 ## Your own components {/*your-own-components*/}
 
 You can also [define your own components](/learn/your-first-component) as JavaScript functions.
+
+--
```
lunaleaps, line 26:
```suggestion
---
```

src/content/reference/react/use.md
```diff
@@ -0,0 +1,769 @@
...
+    "react-error-boundary": "4.0.3"
+  },
+  "main": "/index.js"
+}
+```
+</Sandpack>
+
+#### Providing an alternative value with `Promise.catch` {/*providing-an-alternative-value-with-promise-catch*/}
+
+If you'd like to display provide an alternative value when a promise passed to `use` is rejected you can use an the promise's [`catch`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/catch) method. To use the promise's `catch` method, call `catch` on the promise object. `catch` takes a single argument: a function that takes an error message as an argument. Whatever is returned by the function passed to `catch` will be used as the resolved value of the promise.
```
TheSavior, line 636: > If you'd like to display provide an alternative value

typo?

src/content/reference/react/use.md
TheSavior, line 723: Does this error also come up when using a promise, but not a /supported/ shape of value of that promise? Like when we tried to pass the fetch itself, vs the json value of it

TheSavior (commented):
I'm so excited for folks to get access to this doc

src/content/reference/react/use.md
mattcarrollcode: somewhat? I originally called this `promiseOrContext` but changed it to leave us some flexibility for `use` to consume other resources in the future like a data store. I haven't put any more thought into it beyond that.

src/content/reference/react/use.md
mattcarrollcode: I'm not sure I understand what you are asking. `use` will throw an error if you pass it anything but a promise or a context. If you pass JSON to `use` it will throw an error. I think we were confusing passing a promise to `use` vs passing an object.
```js
use(fetch('/some/api'); // works
```

```js
const response = fetch('/some/api')
use(response); // error
```

```js
const response = fetch('/some/api')
use(response.json()); // works
```

src/content/reference/react/use.md
mattcarrollcode: The example does work in `18.3.0-canary-9377e1010-20230712` ([see this](https://codesandbox.io/s/dwvpmr)) which is the current canary version Next.js uses. But if its not supposed to work/isn't something we want to support let me know and I'll fix this.

sophiebits: What I'm unclear on is the guarantees for when state is preserved when a component suspends before mounting. If this works then I suppose useMemo also would, although since navigator.storage.estimate() isn't pure I suppose it's technically wrong to call it from useMemo.

src/content/reference/react/components.md
leerob, line 42: Should we cross-link to the use client directive?

src/content/reference/react/use.md
```diff
@@ -0,0 +1,769 @@
...
+}
+```
+
+1. First, the promise returned by <CodeStep step={1}>`navigator.storage.estimate()`</CodeStep> is stored in the <CodeStep step={2}>`storagePromise`</CodeStep> state variable created by <CodeStep step={3}>`useState`</CodeStep>. This prevents the call to <CodeStep step={1}>`navigator.storage.estimate()`</CodeStep> and the promise it returns from being recreated across renders.
+2. Next, <CodeStep step={2}>`storagePromise`</CodeStep> is passed to <CodeStep step={4}>`use`</CodeStep> to read the resolved value of the promise.
+3. Finally, the resolved value of the promise can now be used safely. This includes surfacing data read by <CodeStep step={4}>`use`</CodeStep> like <CodeStep step={5}>`storage.quota`</CodeStep> in the JSX returned by the component.
+
+<Pitfall>
+
+When passing promises to `use`, the promise must be cached with the `cache` function, a state variable, or context to prevent the promise from being recreated on every render.
```
leerob, line 91: We mention `cache` here for the first time, but it's not yet available in the docs. That might be okay, but I don't know if that creates a dependency on shipping that first.

src/content/reference/react/use.md
```diff
@@ -0,0 +1,769 @@
...
+```
+
+</Pitfall>
+
+
+<Recipes>
+
+#### Read promises from async browser APIs {/*read-promises-from-async-browser-apis*/}
+
+You can read promises returned by async browser APIs like [`navigator.storage.estimate()`](https://developer.mozilla.org/en-US/docs/Web/API/StorageManager/estimate) with `use`. In this example, the promise from `navigator.storage.estimate()` is stored in a state variable and passed to the `StorageInfo` component as a prop. The promise is then read by `use` and the result is displayed to the user. While waiting for the promise to resolve, the fallback of the parent Suspense component is displayed.
```
leerob, line 116: This is duplicated from above. Intentional?

leerob: It's mentioned `use` can be called conditionally. That could be a great example to include 😄

leerob (commented):
Great work! Exciting progress 🎉

src/content/reference/react/use.md
```diff
@@ -0,0 +1,769 @@
...
+
+* `resource`: this is the source of the data you want to read a value from. A resources can be a [promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) or a [context](/learn/passing-data-deeply-with-context).
+
+#### Returns {/*returns*/}
+
+The `use` Hook returns the value that was read from the resource like the resolved value of a [promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) or [context](/learn/passing-data-deeply-with-context).
+
+#### Caveats {/*caveats*/}
+
+* The parent function of the `use` Hook must be a React component or your own Hooks.
```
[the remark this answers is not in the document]
mattcarrollcode: Fixed!

src/content/reference/react/use.md
mattcarrollcode: Looks like another page [capitalizes Promise](https://react.dev/reference/react-dom/server/renderToReadableStream#returns) I've updated all the references to Promise on this page to be capitalized.

Seems like async and asynchronous are both used on various react.dev pages. asynchronous does seem more appropriate. I've updated all references to asynchronous instead of async unless otherwise required (e.g. the `async` keyword before a function).

src/content/reference/react/use.md
mattcarrollcode: I've removed this section.

src/content/reference/react/use.md
mattcarrollcode: yes, you're right, this is a mistake! fixed

src/content/reference/react/use.md
mattcarrollcode: 😄 nice catch! Fixed.

src/content/reference/react/use.md
mattcarrollcode: Good point, I've updated this to "Like other React Hooks, the function that calls `use` must be React component or Hook." and updated the rest of this page's references to "parent function" to be less jargony. What do you think?

src/content/reference/react/use.md
mattcarrollcode: Took your suggested fix for passive voice and fixed the Suspense boundary and error boundary.

I reworded to the explanation to remove the problematic reference to rendering. I think that is what you were referring to what you said the description isn't consistent with `startTransition`?  What do you think?

src/content/reference/react/use.md
mattcarrollcode: Fixed.

src/content/reference/react/use.md
mattcarrollcode: Fixed!

src/content/reference/react/use.md
mattcarrollcode: fixed!

src/content/reference/react/use.md
mattcarrollcode: Fixed

src/content/reference/react/use.md
mattcarrollcode: "is there a reason to use useContext over use?" probably not except that it's not available in stable. Added a sentence to indicate `use` is preferred over `useContext`

Altered example to demonstrate conditional use

src/content/reference/react/use.md
mattcarrollcode: Fixed!

src/content/reference/react/use.md
mattcarrollcode: Done!

src/content/reference/react/use.md
mattcarrollcode: Fixed

src/content/reference/react/use.md
mattcarrollcode: fixed

src/content/reference/react/use.md
mattcarrollcode: done!

src/content/reference/react/use.md
mattcarrollcode: Added a note that you need to wrap the client component that is passed the promise in Suspense for this to work. Is that enough or would you think a different approach would be better?

src/content/reference/react/use.md
mattcarrollcode: fixed

src/content/reference/react/use.md
mattcarrollcode: Fixed

src/content/reference/react/use.md
mattcarrollcode: Fixed!

src/content/reference/react/use.md
mattcarrollcode: fixed

src/content/reference/react/use.md
mattcarrollcode: fixed

src/content/reference/react/components.md
mattcarrollcode: added a note. What do you think?

src/content/reference/react/components.md
mattcarrollcode: fixed.

src/content/reference/react/components.md
mattcarrollcode: Dan suggested that Server Components and Client Components be capitalized but I don't have a strong opinion either way. I've changed this to be lower case for now.

src/content/reference/react/components.md
mattcarrollcode: fixed

src/content/reference/react/use.md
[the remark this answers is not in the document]
mattcarrollcode: Yes, this is a mistake, use is support in canary in and out of server components. Fixed.

src/content/reference/react/components.md
```diff
@@ -22,3 +22,23 @@ React exposes a few built-in components that you can use in your JSX.
...
+
+## Client and Server Components {/*client-and-server-components*/}
+
+<Wip>
+[React Server Components](/learn/start-a-new-react-project#bleeding-edge-react-frameworks) are only available in React’s canary and experimental channels. Learn more about [React's release channels here](/community/versioning-policy#all-release-channels).
```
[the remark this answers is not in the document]
mattcarrollcode: fixed

src/content/reference/react/use.md
```diff
@@ -0,0 +1,769 @@
...
+In the example below we'll use <CodeStep step={1}>[`navigator.storage.estimate()`](https://developer.mozilla.org/en-US/docs/Web/API/StorageManager/estimate)</CodeStep>, a browser API that provides information on how much storage is being used by the current page and how much space is available. <CodeStep step={1}>`navigator.storage.estimate()`</CodeStep> is an async function that returns a Promise which resolves once the information is available.
+
+```js [[1,6, "navigator.storage.estimate()"], [2,6, "storagePromise"], [3,6, "useState"], [4,7, "use"], [5,8, "storage.quota"]]
+"use client";
+
+import { use, useState } from "react"
+
+function StorageInfo() {
+  const [storagePromise] = useState(navigator.storage.estimate());
+  const storage = use(storagePromise);
```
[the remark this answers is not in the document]
mattcarrollcode: "people might expect the component to be live to storage changes" very good point, I've add a note at the end of the pitfall section to capture that the results are "cached". Do you think its confusing enough I should lead with the Data fetching example instead of a browser API?

src/content/reference/react/use.md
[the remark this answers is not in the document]
mattcarrollcode: Good note. I tried to reword this. What do you think?

src/content/reference/react/use.md
```diff
@@ -0,0 +1,769 @@
...
+function Form() {
+  // ... renders buttons inside ...
+}
+```
+
+It doesn't matter how many layers of components there are between the provider and the `Button`. When a `Button` *anywhere* inside of `Form` calls `use(ThemeContext)`, it will receive `"dark"` as the value.
+
+<Pitfall>
+
+`use(context)` always looks for the closest context provider *above* the component that calls it. It searches upwards and **does not** consider context providers in the component from which you're calling `use()`.
```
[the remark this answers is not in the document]
mattcarrollcode: I just straight copied this from the useContext docs! I've added a note that useContext has the same restrictions

src/content/reference/react/use.md
```diff
@@ -0,0 +1,769 @@
...
+  return (
+    <section className={className}>
+      <h1>{title}</h1>
+      {children}
+    </section>
+  )
+}
+
+function Button({ children }) {
+  const theme = use(ThemeContext);
```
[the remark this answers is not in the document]
mattcarrollcode: Altered this to use a conditional

src/content/reference/react/use.md
mattcarrollcode: fixed!

src/content/reference/react/components.md
mattcarrollcode: Good call! Added.

src/content/reference/react/use.md
mattcarrollcode: @lunaleaps is working on the `cache` doc now, I think its ok to ship this as is and add a link once the `cache` doc is publish. Does that sound ok to you?

src/content/reference/react/use.md
mattcarrollcode: I was thinking that folks might jump to the example, but you're right this is a little wordy. I've thinned it out a bit.

I've added a conditional use example too!

src/content/reference/react/use.md
mattcarrollcode: fixed

src/content/reference/react/components.md
```diff
@@ -22,3 +22,29 @@ React exposes a few built-in components that you can use in your JSX.
...
+
+## Client and server components {/*client-and-server-components*/}
+
+<Wip>
+[React server components](/learn/start-a-new-react-project#bleeding-edge-react-frameworks) are only available in React’s canary and experimental channels. Learn more about [React's release channels here](/community/versioning-policy#all-release-channels).
+</Wip>
+
+<Note>
+
+React server components require the use of [a supported framework](/learn/start-a-new-react-project#bleeding-edge-react-frameworks). If you're aren't working with a framework that support React server components, all the components you interact with are treated as client components.
```
harish-sethuraman, line 36:
```suggestion
React server components require the use of [a supported framework](/learn/start-a-new-react-project#bleeding-edge-react-frameworks). If you aren't working with a framework that supports React server components, all the components you interact with are treated as client components.
```

src/content/reference/react/components.md
```diff
@@ -22,3 +22,29 @@ React exposes a few built-in components that you can use in your JSX.
...
+
+There are two main types of components in React: server components and client components. Both types follow the same basic principles of React - they are reusable, composable, and stateful - but they are used in different scenarios and have different capabilities.
+
+### Server components {/*server-components*/}
+
+Server components allow developers to leverage the power of server-side rendering while maintaining the interactivity and richness of a client-side React application. They are rendered only on the server and have zero impact on the client-side bundle size, making them an excellent choice for parts of your application that are not interactive and don't need to be updated in response to user interaction.
+
+### Client components {/*client-components*/}
+
+Client components can access the full range of React features, including state, effects, and access to the DOM. They are essential for creating rich, interactive features where users expect immediate feedback. Client component must have the [`'use client'` directive](/reference/react/use-client#use-client) place at the top of the file they are defined in.
```
harish-sethuraman, line 48:
```suggestion
Client components can access the full range of React features, including state, effects, and access to the DOM. They are essential for creating rich, interactive features where users expect immediate feedback. Client component must have the [`'use client'` directive](/reference/react/use-client#use-client) placed at the top of the file they are defined in.
```

src/content/reference/react/use.md
```diff
@@ -0,0 +1,886 @@
...
+
+When called with a promise, the `use` Hook integrates with [`Suspense`](/reference/react/Suspense) and [error boundaries](/reference/react/Component#catching-rendering-errors-with-an-error-boundary). The component calling `use` *suspends* while the Promise passed to `use` is pending. If the component that calls `use` is wrapped in a Suspense boundary the fallback of that Suspense component will be displayed.  Once the Promise is resolved, React hides the Suspense component fallback and renders the component(s) inside the Suspense boundary with the data returned by the `use` Hook. If the Promise passed to `use` is rejected, the fallback of the nearest Error Boundary will be displayed.
+
+[See more examples below.](#usage)
+
+#### Parameters {/*parameters*/}
+
+* `resource`: this is the source of the data you want to read a value from. A resources can be a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) or a [context](/learn/passing-data-deeply-with-context).
```
harish-sethuraman, line 51:
```suggestion
* `resource`: this is the source of the data you want to read a value from. A resource can be a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) or a [context](/learn/passing-data-deeply-with-context).
```

src/content/reference/react/use.md
```diff
@@ -0,0 +1,886 @@
...
+
+#### Returns {/*returns*/}
+
+The `use` Hook returns the value that was read from the resource like the resolved value of a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) or [context](/learn/passing-data-deeply-with-context).
+
+#### Caveats {/*caveats*/}
+
+* The `use` Hook must be called inside a React component or a Hook.
+* `use` is not recommended for data fetching from a [server component](/reference/react/components#server-components). `async` and `await` are preferred when fetching data from a server component.
+* When passing Promises to `use`, you must store the Promise in a state variable or context to avoid recreating the Promise on every rerender. When invoking a function that returns a Promise, you can to store the returned Promise with a cache mechanism like React's `cache` function.
```
harish-sethuraman, line 61:
```suggestion
* When passing Promises to `use`, you must store the Promise in a state variable or context to avoid recreating the Promise on every rerender. When invoking a function that returns a Promise, you can store the returned Promise with a cache mechanism like React's `cache` function.
```

src/content/reference/react/use.md
```diff
@@ -0,0 +1,886 @@
...
+  "main": "/index.js"
+}
+```
+</Sandpack>
+
+<Solution />
+
+#### Read Promises from data fetching libraries {/*read-promises-from-data-fetching-libraries*/}
+
+You can read the values of Promises returned from data fetching libraries with the `use` Hook. In this example the function `fetchMessage` is imported from a mock data fetching library. The Promise returned by `fetchMessage` is stored in the `messagePromise` state variable and then passed to `use` to read. The result is then displayed to the user. While waiting for the Promise to resolve, the fallback of the Suspense component that wraps `Message` is displayed.
```
harish-sethuraman, line 214:
```suggestion
You can read the values of Promises returned from data fetching libraries with the `use` Hook. In this example, the function `fetchMessage` is imported from a mock data fetching library. The Promise returned by `fetchMessage` is stored in the `messagePromise` state variable and then passed to `use` to read. The result is then displayed to the user. While waiting for the Promise to resolve, the fallback of the Suspense component that wraps `Message` is displayed.
```

src/content/reference/react/use.md
```diff
@@ -0,0 +1,886 @@
...
+
+import { use, Suspense, useState } from 'react';
+import { fetchByline } from './api.js';
+
+function Article({ title, body, shouldIncludeByline, authorId }) {
+  const [byline, setByline] = useState(null);
+
+  if (shouldIncludeByline && !byline) {
+    // Because `use` is inside a conditional block, we avoid blocking
+    // unncessarily when `shouldIncludeByline` is false.
```
harish-sethuraman, line 310:
```suggestion
    // unnecssarily when `shouldIncludeByline` is false.
```

harish-sethuraman (commented):
have added some typos I found while reading :)

src/content/reference/react/use.md
TheSavior: I guess what I'm asking about is more is more around the serialization of promises over server components, vs what `use` supports. Promises that resolve to objects that contained functions couldn't be serialized obv so they failed when passed to `use`, right? But they would have failed anyways.

src/content/reference/react/components.md
TheSavior: I'm finding it weird to be lower case. In my head, React Server Components is essentially a proper concept. It has a capitalized acronym; RSC. Similarly, SSR is Server Side Rendering, not server side rendering. 

When we've talked about RSC in past communication, like blog posts, it's been capitalized. https://react.dev/blog/2022/03/29/react-v18#server-components-is-still-in-development

> [Server Components](https://react.dev/blog/2020/12/21/data-fetching-with-react-server-components) is an upcoming feature that allows developers to build apps that span the server and client, combining the rich interactivity of client-side apps

https://react.dev/blog/2021/12/17/react-conf-2021-recap#streaming-server-rendering-with-suspense

> In the keynote, we also shared our vision for Suspense, Server Components, new React working groups, and our long-term many-platform vision for React Native.

src/content/reference/react/use.md
```diff
@@ -29,99 +29,120 @@ const value = use(resource);
...
 function Component() {
-  const [promise] = useState(() => new Promise(...)) // ✅ Promise persists across renders
-  const data = use(cache(fnThatReturnsAPromise())) // ✅ Promise persists across renders
-  //...
+  const [promise] = useState(() => new Promise(...)); // ✅ Promise persists across renders
+  const data = use(cache(fnThatReturnsAPromise())); // ✅ Promise persists across renders
+  // ...
+```
+
+Storing the results of a API call in a state variable, context, or cache means that value returned by `use` will not be updated if the underlying API returns a new value. If the underlying data changes you must call the API again and update the state variable or context to the new Promise returned by the API call:
```
TheSavior, line 110: of /an/ API call

src/content/reference/react/use.md
```diff
@@ -29,99 +29,120 @@ const value = use(resource);
...
+
+```js
+function StorageInfo() {
+  const [storagePromise, setStoragePromise] = useState(
```
TheSavior, line 114: You changed the example above to use a callback here, does this need to be a callback too?

src/content/reference/react/use.md
```diff
@@ -269,13 +290,130 @@ export function fetchMessage() {
...
+    setShow(true);
+  }
+
+  if (!show) {
+    return <button onClick={showArticle}>Show article</button>;
+  }
+  return (
+    <Suspense fallback={<p>⌛Loading article...</p>}>
+      <Article
+        title={'Article Title'}
```
TheSavior, line 335: Is the docs code style to write this as `title="Article Title"` or `title={'Article Title'}`. I'm used to seeing the former

src/content/reference/react/use.md
```diff
@@ -269,13 +290,130 @@ export function fetchMessage() {
...
+</Sandpack>
+
+<Solution />
+
+</Recipes>
 
 ### Reading context with `use` {/*reading-context-with-use*/}
 
-When a [context](/learn/passing-data-deeply-with-context) is passed to `use`, it works similarly to [`useContext`](/reference/react/useContext). While `useContext` must be called at the top level of your component, `use` can be called inside conditionals like `if` and loops like `for`.
+When a [context](/learn/passing-data-deeply-with-context) is passed to `use`, it works similarly to [`useContext`](/reference/react/useContext). While `useContext` must be called at the top level of your component, `use` can be called inside conditionals like `if` and loops like `for`. `use` is preferred over `useContext` because it is more flexible.
```
TheSavior, line 416: Does this mean that people generally shouldn't be using `useContext` anymore? That's how I'd interpret this guidance

src/content/reference/react/components.md
sophiebits: Haven’t caught up on all the messages here but I had been thinking “React Server Components” when talking about the feature but not when talking about the type of the component, if that makes sense.

Like if it was called Full-Stack React and had server components and client components. (Maybe that’s too confusing.)

src/content/reference/react/use.md
mattcarrollcode: Yes, I  added this afer @sophiebits suggested we answer the question "is there a reason to use useContext over use?" I don't think there is, so I added this sentence to indicate that you should use `use` over `useContext`: https://github.com/reactjs/react.dev/pull/6177#discussion_r1281183548

src/content/reference/react/components.md
TheSavior: Yep, I agree! When talking about a specific server component it is lower case, when talking about React Server Components, it is upper case. I was thinking about it the same way.

When talking about a React feature being supported in a Server Component it is uppercase (but maybe that should always just be Server Component*s*)

sophiebits: My inclination would be:

* Frameworks that support React Server Components let you write both server components and client components.
* Only server components can use async/await at present, but client component support is planned. [this is different [this is different from what you said @TheSavior]
* I wrote a client component today and it made me sad.

i.e. capitalize it only if and only if you could replace the usage with "The React Architecture" or similar (if that were an official term)

mattcarrollcode: I've changed it back to capitalize "Server Components" and "Client Components" to be consistent with the React blog:

* https://react.dev/blog/2022/06/15/react-labs-what-we-have-been-working-on-june-2022#server-components
* https://react.dev/blog/2023/03/22/react-labs-what-we-have-been-working-on-march-2023#react-server-components
* https://react.dev/blog/2022/03/29/react-v18#server-components-is-still-in-development
* https://react.dev/blog/2023/05/03/react-canaries#example-react-server-components

which has more history than the react.dev doc pages that reference lowercase client component or server component:

* https://react.dev/reference/react/use-server#caveats
* https://react.dev/reference/react/useId#why-is-useid-better-than-an-incrementing-counter
* https://react.dev/reference/react/use-client#caveats

src/content/reference/react/use.md
mattcarrollcode: Fixed!

src/content/reference/react/use.md
mattcarrollcode: yes! good catch, fixed.

src/content/reference/react/use.md
mattcarrollcode: Looks like the docs style is `title="Article Title"`! Fixed.

src/content/reference/react/components.md
sophiebits: ok, sgtm – not convinced we should capitalize these long term (eg: we don't capitalize component, element, context) but seems fine for now, thanks for investigating

src/content/reference/react/use.md
leerob: Works for me 👍

src/content/reference/react/use.md
mattcarrollcode: per @acdlite creating promises during rendering is not a supported pattern. In the future it might be a good idea to write docs for how to work with promises when writing a data fetching library that integrates with Suspense. I'll be rewriting the examples to avoid creating promises as part of rendering.

src/content/reference/react/index.md
```diff
@@ -110,6 +110,7 @@ To prioritize rendering, use one of these Hooks:
 
 These Hooks are mostly useful to library authors and aren't commonly used in the application code.
 
+- [`use`](/reference/react/use) lets you read the value of a resource like a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) or [context](/learn/passing-data-deeply-with-context)
```
sophiebits, line 113: imo `use` is or will be a lot more important than useDebugValue, useId, useSyncExternalStore. if you want to leave it here while it's only in canary that's probably OK but before stable it should probably move up, maybe to its own section?

src/content/reference/react/use.md
mattcarrollcode: Gotcha. Added a note in the Streaming section  and a troubleshooting section indicating that the resolved value of the promise must be serializable.

src/content/reference/react/use.md
mattcarrollcode: I've rewritten this page to reflect that promises can't be created as a part of rendering.

src/content/reference/react/components.md
```diff
@@ -22,3 +22,29 @@ React exposes a few built-in components that you can use in your JSX.
...
+[React server components](/learn/start-a-new-react-project#bleeding-edge-react-frameworks) are only available in React’s canary and experimental channels. Learn more about [React's release channels here](/community/versioning-policy#all-release-channels).
+</Wip>
+
+<Note>
+
+React server components require the use of [a supported framework](/learn/start-a-new-react-project#bleeding-edge-react-frameworks). If you aren't working with a framework that supports React server components, all the components you interact with are treated as client components.
+
+</Note>
+
+There are two main types of components in React: server components and client components. Both types follow the same basic principles of React - they are reusable, composable, and stateful - but they are used in different scenarios and have different capabilities.
```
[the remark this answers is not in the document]
mattcarrollcode: I've removed this section since we are planning to add this information into the "use server" and "use client" directive reference pages

src/content/reference/react/components.md
```diff
@@ -22,3 +22,29 @@ React exposes a few built-in components that you can use in your JSX.
...
+
+### Server components {/*server-components*/}
+
+Server components allow developers to leverage the power of server-side rendering while maintaining the interactivity and richness of a client-side React application. They are rendered only on the server and have zero impact on the client-side bundle size, making them an excellent choice for parts of your application that are not interactive and don't need to be updated in response to user interaction.
+
+### Client components {/*client-components*/}
+
+Client components can access the full range of React features, including state, effects, and access to the DOM. They are essential for creating rich, interactive features where users expect immediate feedback. Client component must have the [`'use client'` directive](/reference/react/use-client#use-client) placed at the top of the file they are defined in.
+
+In the following sections, we will delve deeper into the characteristics, capabilities, and use cases of both server components and client Components.
```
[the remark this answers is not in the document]
mattcarrollcode: I've removed this section since we are planning to add this information into the "use server" and "use client" directive reference pages

src/content/reference/react/components.md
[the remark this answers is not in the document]
mattcarrollcode: I've removed this section since we are planning to add this information into the "use server" and "use client" directive reference pages

src/content/reference/react/use.md
```diff
@@ -0,0 +1,886 @@
...
+
+function MessageComponent({ messagePromise }) {
+  const message = use(messagePromise);
+  const theme = use(ThemeContext);
+  // ...
+```
+
+Unlike all other React Hooks, the `use` Hook can be called within conditional statements like loops and `if` statements. Like other React Hooks, the function that calls `use` must be a Component or Hook.
+
+When called with a Promise, the `use` Hook integrates with [`Suspense`](/reference/react/Suspense) and [error boundaries](/reference/react/Component#catching-rendering-errors-with-an-error-boundary). The component calling `use` *suspends* while the Promise passed to `use` is pending. If the component that calls `use` is wrapped in a Suspense boundary the fallback of that Suspense component will be displayed.  Once the Promise is resolved, React hides the Suspense component fallback and renders the component(s) inside the Suspense boundary with the data returned by the `use` Hook. If the Promise passed to `use` is rejected, the fallback of the nearest Error Boundary will be displayed.
```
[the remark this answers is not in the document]
mattcarrollcode: I've implemented your suggestions, thanks!

src/content/reference/react/use.md
```diff
@@ -0,0 +1,886 @@
...
+
+#### Returns {/*returns*/}
+
+The `use` Hook returns the value that was read from the resource like the resolved value of a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) or [context](/learn/passing-data-deeply-with-context).
+
+#### Caveats {/*caveats*/}
+
+* The `use` Hook must be called inside a Component or a Hook.
+* `use` is not recommended for data fetching from a [Server Component](/reference/react/components#server-components). `async` and `await` are preferred when fetching data from a Server Component.
+* When passing Promises to `use`, you must store the Promise in a state variable or context to avoid recreating the Promise on every rerender. When invoking a function that returns a Promise, you can store the returned Promise with a cache mechanism like React's `cache` function.
```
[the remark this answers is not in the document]
mattcarrollcode: I've removed the reference to `cache` for now.

src/content/reference/react/use.md
```diff
@@ -0,0 +1,886 @@
...
+
+<InlineToc />
+
+---
+
+## Reference {/*reference*/}
+
+### `use(resource)` {/*use*/}
+
+Call `use` in your component to read the value of a resource like a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) or [context](/learn/passing-data-deeply-with-context).
```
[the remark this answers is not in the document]
mattcarrollcode: I have this here to follow react.dev convention (e.g. [useState](https://react.dev/reference/react/useState)), to introduce the code sample, and because the intro sentence is about ~page height away once you account for the ToC, titles and code samples. I think I'll leave this for now, please let me know if I'm missing something.

src/content/reference/react/use.md
```diff
@@ -0,0 +1,886 @@
...
+* `resource`: this is the source of the data you want to read a value from. A resource can be a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) or a [context](/learn/passing-data-deeply-with-context).
+
+#### Returns {/*returns*/}
+
+The `use` Hook returns the value that was read from the resource like the resolved value of a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) or [context](/learn/passing-data-deeply-with-context).
+
+#### Caveats {/*caveats*/}
+
+* The `use` Hook must be called inside a Component or a Hook.
+* `use` is not recommended for data fetching from a [Server Component](/reference/react/components#server-components). `async` and `await` are preferred when fetching data from a Server Component.
```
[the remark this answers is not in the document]
mattcarrollcode: If I understand this correctly `fetch` is preferred when fetching data in a RSC in Next.js because of the all the work Next.js has done to dedupe requests, cache, etc. but my understanding is that this is a framework-level choice and another framework could choose to have some other data loading API that they prefer. To keep the React docs neutral I think I'll keep this is as a "generic" preference for `async`/`await` and omit `fetch` for now. Please let me know if I'm missing something.

src/content/reference/react/use.md
```diff
@@ -0,0 +1,886 @@
...
+
+#### Parameters {/*parameters*/}
+
+* `resource`: this is the source of the data you want to read a value from. A resource can be a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) or a [context](/learn/passing-data-deeply-with-context).
+
+#### Returns {/*returns*/}
+
+The `use` Hook returns the value that was read from the resource like the resolved value of a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) or [context](/learn/passing-data-deeply-with-context).
+
+#### Caveats {/*caveats*/}
```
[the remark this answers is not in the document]
mattcarrollcode: Excellent suggestion! I've incorporated it into the Caveats section

src/content/reference/react/index.md
mattcarrollcode: Good callout, I added a "Resource Hooks" section.

src/content/reference/react/use.md
```diff
@@ -0,0 +1,504 @@
...
+
+  return (
+    <Suspense fallback={<p>waiting for message...</p>}>
+      <Message messagePromise={messagePromise} />
+    </Suspense>
+  );
+}
+```
+
+To use the Promise's <CodeStep step={1}>`catch`</CodeStep> method, call <CodeStep step={1}>`catch`</CodeStep> on the Promise object. <CodeStep step={1}>`catch`</CodeStep> takes a single argument: a function that takes an error message as an argument. Whatever is <CodeStep step={2}>returned</CodeStep> by the function passed to <CodeStep step={1}>`catch`</CodeStep> will be used as the resolved value of the Promise.
```
lunaleaps, line 469: Is this limited to defining the catch on the server? Could we add a `catch` to a promise from the server in a client component?

src/content/reference/react/use.md
```diff
@@ -0,0 +1,504 @@
...
+
+---
+
+## Troubleshooting {/*troubleshooting*/}
+
+### I'm getting this error: "An unsupported type was passed to use" {/*im-getting-this-error-an-unsupported-type-was-passed-to-use*/}
+
+The `use` Hook only supports [JavaScript Promises](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) and React [Context](/learn/passing-data-deeply-with-context) as parameters.
```
lunaleaps, line 477: Would this error trigger if you `use`-d a client-created promise?

src/content/reference/react/use.md
```diff
@@ -0,0 +1,504 @@
...
+
+```jsx
+function MessageComponent({messagePromise}) {
+  function download() {
+    // ❌ the function calling `use` is not a Component or Hook
+    const message = use(messagePromise);
+    // ...
+```
+
+instead, call `use` outside any component closures, where the function that calls `use` is a component or Hook:
```
lunaleaps, line 493:
```suggestion
Instead, call `use` outside any component closures, where the function that calls `use` is a component or Hook:
```

src/content/reference/react/use.md
```diff
@@ -0,0 +1,504 @@
...
+### I'm getting this error: "Suspense Exception: This is not a real error!" {/*suspense-exception-error*/}
+
+You are either calling `use` outside of a React component or Hook function, or calling `use` in a try–catch block. If you are calling `use` inside a try–catch block, wrap your component in an error boundary, or call the Promise's `catch` to catch the error and resolve the Promise with another value. [See these examples](#dealing-with-rejected-promises).
+
+If you are calling `use` outside a React component or Hook function, move the `use` call to a React component or Hook function.
+
+```jsx
+function MessageComponent({messagePromise}) {
+  function download() {
+    // ❌ the function calling `use` is not a Component or Hook
```
lunaleaps, line 488: In this example, wouldn't `download` be called by the `MessageComponent` since it's only defined in the component scope? Or else, how else would it be called? In that case, would this still be invalid? If that is what you're trying to demonstrate, maybe we should explicitly add the `download()` call in the example?

lunaleaps:
Thank for getting this in shape to review again! No major blockers for me, I think the main thing I'm concerned about is the clarification of unsupported client-created promises with `use` and how we should highlight that

lunaleaps (approved):

lunaleaps (approved):

src/content/reference/react/use.md
```diff
@@ -0,0 +1,504 @@
...
+
+#### Returns {/*returns*/}
+
+The `use` Hook returns the value that was read from the resource like the resolved value of a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) or [context](/learn/passing-data-deeply-with-context).
+
+#### Caveats {/*caveats*/}
+
+* The `use` Hook must be called inside a Component or a Hook.
+* `use` is not recommended for data fetching from a [Server Component](/reference/react/use-server). `async` and `await` are preferred when fetching data from a Server Component.
+* Creating a Promise inside a [Client Component](/reference/react/use-client) is not supported.
```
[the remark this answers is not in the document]
mattcarrollcode: You can create a promise in a client component and you can pass it to `use` and it will all work properly. The problem is that the work to optimize how promises work across rerenders hasn't happened. If you create a promise or call a API that returns a Promise in a client component that Promise will be recreated on every rerender. `cache` actually fixes this (see Jack Herrington's video: https://www.youtube.com/watch?v=T3m-MZkuadU) but as you well know, `cache` is not ready on the client yet either.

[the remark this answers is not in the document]
mattcarrollcode: I'll take another stab at explaining this better

src/content/reference/react/use.md
```diff
@@ -0,0 +1,504 @@
...
+* `resource`: this is the source of the data you want to read a value from. A resource can be a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) or a [context](/learn/passing-data-deeply-with-context).
+
+#### Returns {/*returns*/}
+
+The `use` Hook returns the value that was read from the resource like the resolved value of a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) or [context](/learn/passing-data-deeply-with-context).
+
+#### Caveats {/*caveats*/}
+
+* The `use` Hook must be called inside a Component or a Hook.
+* `use` is not recommended for data fetching from a [Server Component](/reference/react/use-server). `async` and `await` are preferred when fetching data from a Server Component.
```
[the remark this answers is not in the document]
mattcarrollcode: I think we want to leave room for lib authors to be able to use it in some cases, but I don't think there is `use` use case for developers using React to build apps

src/content/reference/react/use.md
```diff
@@ -0,0 +1,504 @@
...
+
+```js [[2, 4, "theme"], [1, 4, "ThemeContext"]]
+import { use } from 'react';
+
+function Button() {
+  const theme = use(ThemeContext);
+  // ... 
+```
+
+`use` returns the <CodeStep step={2}>context value</CodeStep> for the <CodeStep step={1}>context</CodeStep> you passed. To determine the context value, React searches the component tree and finds **the closest context provider above** for that particular context.
```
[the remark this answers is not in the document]
mattcarrollcode: I copied this from the (`useContext` docs)[https://react.dev/reference/react/useContext#passing-data-deeply-into-the-tree] so if we change it here we should probably change it there too.

src/content/reference/react/use.md
```diff
@@ -0,0 +1,504 @@
...
+    </ThemeContext.Provider>
+  );
+}
+
+function Form() {
+  // ... renders buttons inside ...
+}
+```
+
+It doesn't matter how many layers of components there are between the provider and the `Button`. When a `Button` *anywhere* inside of `Form` calls `use(ThemeContext)`, it will receive `"dark"` as the value.
```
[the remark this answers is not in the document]
mattcarrollcode: I copied this from the [`useContext` docs](https://react.dev/reference/react/useContext#passing-data-deeply-into-the-tree) so if we change it here we should probably change it there too.

src/content/reference/react/use.md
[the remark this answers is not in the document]
sophiebits: The case when you would want to is if you are using another hook in the same component. Which would probably mean just useId right now.

Whenever we say "not recommended" it would be good to say why. In this case I think it's because use() requires React to run the component a second time once the data resolves, while async/await can pick up right where it left off.

src/content/reference/react/use.md
```diff
@@ -0,0 +1,504 @@
...
+  const messagePromise = fetchMessage();
+  return (
+    <Suspense fallback={<p>waiting for message...</p>}>
+      <Message messagePromise={messagePromise} />
+    </Suspense>
+  );
+}
+```
+
+The Client Component (<CodeStep step={2}>`Message`</CodeStep>) then takes the Promise it received as a prop (<CodeStep step={4}>`messagePromise`</CodeStep>) and passes it to the <CodeStep step={5}>`use`</CodeStep> Hook. This allows the Client Component (<CodeStep step={2}>`Message`</CodeStep>) to read the value from the Promise that was initially created by the Server Component (<CodeStep step={4}>`messagePromise`</CodeStep>):
```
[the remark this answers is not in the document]
mattcarrollcode: That's a great idea! Fixed.

src/content/reference/react/use.md
```diff
@@ -0,0 +1,504 @@
...
+<Note>
+
+When passing a Promise from a Server Component to a Client Component, its resolved value must be serializable to pass between server and client. Data types like functions aren't serializable and cannot serve as the resolved value of such a Promise.
+
+</Note>
+
+
+<DeepDive>
+
+#### Server Components: `await` vs passing a Promise to a Client Component {/*await-vs-passing-a-promise*/}
```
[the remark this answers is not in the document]
mattcarrollcode: Great idea! Changed the title to "Should I resolve a Promise in a Server or Client Component?"

src/content/reference/react/use.md
```diff
@@ -0,0 +1,504 @@
...
+Instead of passing a Promise from a Server Component to a Client Component you could `await` the Promise in the Server Component and pass the data to the Client Component as a prop:
+
+```js
+export default function App() {
+  const messageContent = await fetchMessage();
+  return <Message messageContent={messageContent} />
+}
+```
+
+But using `await` in a [Server Component](/reference/react/components#server-components) will block the rendering of the [Server Component](/reference/react/components#server-components) until the `await` statement is finished. Passing a Promise from a Server Component to a Client Component prevents the Promise from blocking the rendering of the Server Component.
```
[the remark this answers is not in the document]
mattcarrollcode: I've hesitated to provide general advice because I think is pretty nuanced. If you're all the point in developing your app where you are looking at the `use` Hook documentation and have read this far I think you'll have a pretty good idea if you need to stream a promise or not. Generally I think folks should be doing data fetching on the server (especially if the result of the Promise impacts the layout of the page) but there are plenty of cases where streaming the result to the client would make more sense (e.g. a operation that takes a long time, the user won't notice if this data takes a little while to populate, their users expect low latency for some areas of the page but not others, etc.). Overall I think trying to provide general advice here will do more harm than good.

src/content/reference/react/use.md
mattcarrollcode: The `catch` method returns a new Promise (which is why you can chain method calls on a Promise). If you called `catch` on a Promise in a client component you'd be creating a new Promise in render which would be recreated on every rerender. So for now `catch` should only be called outside of render.

src/content/reference/react/use.md
mattcarrollcode: Good catch! Fixed.

src/content/reference/react/use.md
mattcarrollcode: `download` will be called by `MessageComponent` but `use`'s parent function is `download` and `download` isn't a Component or Hook. The parent function of `use` must be a Component or Hook or an error will be thrown.

The code displayed in this snippet is wrong and there isn't any code that could be written elsewhere in the component or tree to make it right. You could simply write code to not call `download` and then you wouldn't trigger an error but the code would still be "incorrect". If you call `use` outside a component or hook there will be an error: output

src/content/reference/react/use.md
mattcarrollcode: Unfortunately, no. You can pass a client-created Promise without any visible indication that anything is incorrect or unsupported.

src/content/reference/react/use.md
```diff
@@ -0,0 +1,504 @@
...
+    "react-error-boundary": "4.0.3"
+  },
+  "main": "/index.js"
+}
+```
+</Sandpack>
+
+#### Providing an alternative value with `Promise.catch` {/*providing-an-alternative-value-with-promise-catch*/}
+
+If you'd like to provide an alternative value when the Promise passed to `use` is rejected you can use an the Promise's <CodeStep step={1}>[`catch`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/catch)</CodeStep> method:
```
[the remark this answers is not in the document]
mattcarrollcode: Thanks, fixed!

src/content/reference/react/use.md
```diff
@@ -0,0 +1,504 @@
...
+  }
+  return false;
+}
+```
+
+<CodeStep step={2}>`use`</CodeStep> is called from inside a <CodeStep step={1}>`if`</CodeStep> statement, allowing you to conditionally read values from a Context.
+
+<Pitfall>
+
+Like `useContext`, `use(context)` always looks for the closest context provider *above* the component that calls it. It searches upwards and **does not** consider context providers in the component from which you're calling `use(context)`.
```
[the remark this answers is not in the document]
mattcarrollcode: The default value is `null` and is covered in the [createContext docs](https://react.dev/reference/react/createContext). The current `useContext` and  and `createContext` docs don't have an example to showcase how React searches the tree for Contexts so I think I'll omit that on this page too, at least until the `use` hook is stable.

src/content/reference/react/use.md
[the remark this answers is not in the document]
mattcarrollcode: Updated the caveat wording to better explain the limitations of creating Promises in Client components.

src/content/reference/react/use.md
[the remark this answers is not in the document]
mattcarrollcode: Thanks for the explanation Sophie! I've updated this caveat to include this context.

src/content/reference/react/use.md
```diff
@@ -0,0 +1,496 @@
...
+}
+
+function Form() {
+  // ... renders buttons inside ...
+}
+```
+
+It doesn't matter how many layers of components there are between the provider and the `Button`. When a `Button` *anywhere* inside of `Form` calls `use(ThemeContext)`, it will receive `"dark"` as the value.
+
+Unlike [`useContext`](/reference/react/useContext), <CodeStep step={2}>`use`</CodeStep> can be called in conditionals and loops like <CodeStep step={1}>`if`</CodeStep>.
```
sophiebits, line 97: feedback: I think the use of CodeStep on this line and later in the section is a bit overbearing. I'm not sure it adds anything that isn't already plainly evident from looking at the text/code, and also since you used the same colors to mean something else in this same section, it seems like it's related (but then it's not)

src/content/reference/react/use.md
```diff
@@ -0,0 +1,496 @@
...
+  },
+  "main": "/index.js"
+}
+```
+
+</Sandpack>
+
+### Streaming data from the server to the client {/*streaming-data-from-server-to-client*/}
+
+Data can be streamed from the server to the client by passing a Promise as a prop from a <CodeStep step={1}>Server Component</CodeStep> to a <CodeStep step={2}>Client Component</CodeStep>.
```
sophiebits, line 217: similarly, I'm not sure that CodeStep in this section adds much and I think it can be a bit overwhelming. instead, I would use prose to introduce the setup of a server component called App passing a promise to a client component called Message, which seems to be what you're trying to convey with the colors.

src/content/reference/react/use.md
```diff
@@ -0,0 +1,496 @@
...
+    "react-error-boundary": "4.0.3"
+  },
+  "main": "/index.js"
+}
+```
+</Sandpack>
+
+#### Providing an alternative value with `Promise.catch` {/*providing-an-alternative-value-with-promise-catch*/}
+
+If you'd like to provide an alternative value when the Promise passed to `use` is rejected you can use the Promise's <CodeStep step={1}>[`catch`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/catch)</CodeStep> method.
```
sophiebits, line 449: (and ditto here – don't think the colors add anything, so it's a bit distracting)

PR #6193, opened 2023-08-02
chores: remove unused deps and args

hawtim:
When I am learning this part from [react.dev](https://react.dev/learn/managing-state#scaling-up-with-reducer-and-context), I found some unused imports and args, so I remove them to make it clear.


hawtim (commented):
double checked

PR #6194, opened 2023-08-03
Fix grammar in a sentence.

Aleksandar15:
Fixed grammar in an explanatory part.


mattcarrollcode:
Thanks for the fix! Merging....

PR #6210, opened 2023-08-10
Remove useless prop in `ContactList` code examples

danivazx:

### DIFF
--- a/src/content/learn/managing-state.md
+++ b/src/content/learn/managing-state.md
@@ -319,7 +319,6 @@ export default function Messenger() {
     <div>
       <ContactList
         contacts={contacts}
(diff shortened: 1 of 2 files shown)


src/content/learn/managing-state.md
```diff
@@ -336,7 +335,6 @@ const contacts = [
 
 ```js ContactList.js
 export default function ContactList({
-  selectedContact,
```
lunaleaps, line 339: Instead of removing this prop, can we add bold-en the selected contact name in this example? I think that will make this example clearer when you've selected a name.

lunaleaps (changes requested):

Issue #6014, opened 2023-05-09
No non-legacy page for Error Decoder


martincostello:
If one follows a link for a React error in a website, you are directed to the [legacy react documentation](https://legacy.reactjs.org/docs/error-decoder.html/).

![image](https://github.com/reactjs/react.dev/assets/1439341/346abbf3-ec6b-4582-93e3-320aa4226be5)

There doesn't _appear_ to be a non-legacy equivalent of this page. It would make sense for developers following these error messages from the browser console to a non-legacy page.

When/if there is one, then the URL [here](https://github.com/facebook/react/blob/7ac5e9a602347f3b7d26c60a549c483d3bc88bbf/packages/shared/formatProdErrorMessage.js#L14) should be updated to point to it.

rickhanlonii:
Yeah, nice catch. Dupe of https://github.com/reactjs/react.dev/issues/5572 let's track it there.

PR #6214, opened 2023-08-11
Feat: error-decoder

### LINKED ISSUE #5572: [Beta] Migrating Error Decoder
`beta.reactjs.org` currently missing `/error-decoder` (E.g. https://reactjs.org/docs/error-decoder.html/?invariant=421). https://beta.reactjs.org/docs/error-decoder.html/?invariant=421 will be redirected to ~~https://legacy.reactjs.org/error-decoder.html?invariant=421 which becomes 404.~~ https://legacy.reactjs.org/docs/error-decoder.html?invariant=421

SukkaW:
The PR fixes #5572 and #6014.

@rickhanlonii 

Preview:

- No error code in URL: https://react-dev-git-fork-sukkaw-error-decoder-fbopensource.vercel.app/errors
- Error code only: https://react-dev-git-fork-sukkaw-error-decoder-fbopensource.vercel.app/errors/421
- Error code and single argument: https://react-dev-git-fork-sukkaw-error-decoder-fbopensource.vercel.app/errors/137?args[]=%3Cbr%20/%3E
- Error code and multiple arguments: https://react-dev-git-fork-sukkaw-error-decoder-fbopensource.vercel.app/errors/377?args[0]=114514n&args[1]=count
- Error code and missing the only argument: https://react-dev-git-fork-sukkaw-error-decoder-fbopensource.vercel.app/errors/137
- Error code and missing some arguments: https://react-dev-git-fork-sukkaw-error-decoder-fbopensource.vercel.app/errors/377?args[0]=&args[1]=count
- Error message with URL: https://react-dev-git-fork-sukkaw-error-decoder-fbopensource.vercel.app/errors/119


src/pages/[[...markdownPath]].js
```diff
@@ -216,10 +220,31 @@ export async function getStaticProps(context) {
...
+        toc,
+        meta,
+        errorCodes: await (
+          await fetch(
+            'https://raw.githubusercontent.com/facebook/react/main/scripts/error-codes/codes.json'
+          )
+        ).json(),
+      },
+    };
+  }
```
Comment (SukkaW) [inline src/pages/[[...markdownPath]].js:242]: The legacy site implements a Gatsby plugin to download the `codes.json` from GitHub RAW during the build: https://github.com/reactjs/reactjs.org/blob/d696cb1d2a1ea1d7eb92d5b098b51dde8fcc536c/plugins/gatsby-source-react-error-codes/gatsby-node.js#L13-L30

I am fetching `codes.json` in `getStaticProps`, but IMHO this doesn't look right.

src/components/MDX/ErrorDecoder.tsx
```diff
@@ -0,0 +1,104 @@
...
+export default function ErrorDecoder() {
+  const {isReady} = useRouter();
+  const errorCodes = useContext(ErrorCodesContext);
+
+  const [code, msg] = useMemo(() => {
+    let code = null;
+    let msg = '';
+
+    if (typeof window !== 'undefined' && isReady) {
+      const parseResult = parseQueryString(window.location.search);
```
SukkaW, line 89: I copied the `parseQueryString` from the legacy site for the initial prototype. I am migrating this to the Next.js router's `query` object.

SukkaW: Done in [`1927c29` (#6214)](https://github.com/reactjs/react.dev/pull/6214/commits/1927c299a63d8fbbe898f363ca235f8fc25f8eb0).

src/pages/[[...markdownPath]].js
Comment (rickhanlonii) [reply src/pages/[[...markdownPath]].js:243]: What if we also use `getStaticPaths` to generate all of the error code pages at build time? We should also namespace the route with something like `errors/[[error_code]].js`, so the new site URLs would be something like:

`https://react.dev/errors/377?args%5B0%5D=114514n&args%5B1%5D=count`

That would require a few other things:
- For when new errors are pushed but the site hasn't deployed, we would need a fallback for Not Found that looks up the error codes at runtime to see if they exist.
- For redirects to work (where the error code is in the URL for the static paths) we would need a query-param based redirect. Maybe `https://react.dev/errors?invariant=337` redirects to `https://react.dev/errors/337`. We can update the error messages in React to use the static path and keep the redirect around as a 301 for existing links.

Comment (rickhanlonii) [reply src/pages/[[...markdownPath]].js:243]: We could avoid the later if vercel allows for moving query params to the route in a redirect, but skimming the docs I don't see a way to do it.

Comment (SukkaW) [reply src/pages/[[...markdownPath]].js:243]: > What if we also use `getStaticPaths` to generate all of the error code pages at build time? We should also namespace the route with something like `errors/[[error_code]].js`, so the new site URLs would be something like:

I like the idea of `https://react.dev/errors/377?args%5B0%5D=114514n&args%5B1%5D=count`, but SSG all 483 (and the number will continue to grow) pages at the build time really sounds like a bad idea to me.

> For when new errors are pushed but the site hasn't deployed, we would need a fallback for Not Found that looks up the error codes at runtime to see if they exist.

It is possible to implement the on-demand static generation with the Next.js' `getStaticPaths` + `fallback: 'blocking'` (https://nextjs.org/docs/pages/api-reference/functions/get-static-paths#fallback-blocking). However, `getStaticPaths` + `fallback: 'blocking'` doesn't work with Next.js SSG, which means the site now has to run on a server/serverless platform.

Should we fetch the `code.json` on the client side (which can guarantee to be the latest version) instead?

Comment (rickhanlonii) [reply src/pages/[[...markdownPath]].js:243]: What's the concern with SSG? If it's build-time perf, what about using ISR with [`fallback: true`](https://nextjs.org/docs/pages/api-reference/functions/get-static-paths#fallback-true)? This will avoid generating the files during build time and instead generate them on-demand, forcing blocking mode where it makes sense (e.g. Google). Vercel will automatically configure the server for deployment. 

Fetching on the client would create an unnecessary delay/waterfall.

Comment (SukkaW) [reply src/pages/[[...markdownPath]].js]: Done in [`4c89c2b` (#6214)](https://github.com/reactjs/react.dev/pull/6214/commits/4c89c2beb2373d4db21fdaa301d60c303108c59c). I have created a `src/pages/errors/[error_code].tsx` and a `src/pages/errors/index.tsx` file.

src/components/ErrorDecoder.tsx
```diff
@@ -0,0 +1,94 @@
...
+    if (typeof window !== 'undefined' && isReady) {
+      return replaceArgs(
+        errorMessages,
+        parseQueryString(query),
+        '[missing argument]'
+      );
+    }
+
+    return replaceArgs(errorMessages, [], '[parsing argument]');
+  }, [errorMessages, isReady, query]);
```
SukkaW, line 73: @rickhanlonii 

Turns out that Next.js SSG is fast enough. The total build time is at 95 seconds now.

Note that although the page itself has been SSG-ed at the build time, the `router.query` still requires client-side rendering and will only be populated after the initial hydration, thus `isReady` can not be avoided.

rickhanlonii:
I pushed https://github.com/facebook/react/pull/27240 to switch the error messages over to this format.

src/components/ErrorDecoder.tsx
```diff
@@ -0,0 +1,94 @@
...
+  let argIdx = 0;
+  return msg.replace(/%s/g, function () {
+    const arg = argList[argIdx++];
+    // arg can be an empty string: ?invariant=377&args[0]=&args[1]=count
+    return arg === undefined || arg === '' ? replacer : arg;
+  });
+}
+
+// When the message contains a URL (like https://fb.me/react-refs-must-have-owner),
+// make it a clickable link.
```
rickhanlonii, line 18: Can you add an example of this to the PR description?

src/components/ErrorDecoder.tsx
```diff
@@ -0,0 +1,94 @@
...
+    );
+  }
+
+  return (
+    <div>
+      <p>
+        <b>The full text of the error you just encountered is:</b>
+      </p>
+      <code className="block bg-red-100 text-red-600 py-4 px-6 mt-5 rounded-lg">
+        <b>{urlify(msg)}</b>
```
rickhanlonii, line 90: Move the urlify inside useMemo?

src/components/ErrorDecoder.tsx
rickhanlonii: When the query changes, won't that trigger an update? Why is the `isReady` needed?

src/pages/[[...markdownPath]].js
```diff
@@ -216,10 +217,14 @@ export async function getStaticProps(context) {
   const fm = require('gray-matter');
   const meta = fm(mdx).data;
 
+  // Serialize MDX into JSON.
```
Comment (rickhanlonii) [inline src/pages/[[...markdownPath]].js:220]: Can probably revert this now.

src/pages/errors/[error_code].tsx
```diff
@@ -0,0 +1,96 @@
...
+        <ErrorDecoder errorMessages={errorMessages} />
+      </MaxWidth>
+    </Page>
+  );
+}
+
+/**
+ * Next.js Page Router doesn't have a way to cache specific data fetching request.
+ * But since Next.js uses limited number of workers, keep "cachedErrorCodes" as a
+ * module level memory cache can reduce the number of requests down to once per worker.
```
Comment (rickhanlonii) [inline src/pages/errors/[error_code].tsx:48]: Nice optimization!

rickhanlonii (commented):
Looking really good, what do you think about my markdown comment?

src/pages/errors/[error_code].tsx
```diff
@@ -0,0 +1,96 @@
...
+      meta={{title: 'Error Decoder'}}
+      routeTree={sidebarLearn as RouteItem}
+      section="unknown">
+      <MaxWidth>
+        <Intro>
+          <P>
+            In the minified production build of React, we avoid sending down
+            full error messages in order to reduce the number of bytes sent over
+            the wire.
+          </P>
```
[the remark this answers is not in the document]
Comment (rickhanlonii) [reply src/pages/errors/[error_code].tsx:30]: The reason this is important is because it limits the number of merge conflicts in the translation repo when we change the code of the page. When the only differences between the main repo and the translation repos are in the markdown files, it makes it easier for translations to stay up to date with the main site since the only conflicts are in the actual content.

src/components/ErrorDecoder.tsx
SukkaW: > When the query changes, won't that trigger an update? Why is the `isReady` needed?

Per Next.js documentation: https://nextjs.org/docs/pages/building-your-application/rendering/automatic-static-optimization#how-it-works

> To be able to distinguish if the query is fully updated and ready for use, you can leverage the `isReady` field on [`next/router`](https://nextjs.org/docs/pages/api-reference/functions/use-router#router-object).

This also prevents SSR hydration mismatch errors. On the server, Next.js SSR the page without the `query`, and Next.js will also use the empty query object during the initial hydration.

The `isReady: true` will only be set after the initial hydration has finished and the `query` object has been populated (the `isReady` here acts similarly with [React's two-pass rendering approach](https://react.dev/reference/react-dom/client/hydrateRoot#handling-different-client-and-server-content)).

The `useSearchParams` of the Next.js App Router uses the Suspense approach, which is useful when we migrate to the App Router.

src/components/ErrorDecoder.tsx
SukkaW: I can't find this kind of error message in the recent version of React. This kind of error message might only exist in the earlier version of React.

SukkaW: @rickhanlonii Found an example: https://react-dev-git-fork-sukkaw-error-decoder-fbopensource.vercel.app/errors/119

The PR description has also been updated.

src/pages/errors/[error_code].tsx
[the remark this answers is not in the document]
Comment (SukkaW) [reply src/pages/errors/[error_code].tsx:30]: I like the idea. But IMHO this might be too much for this PR now. We could implement this in the future PR.

src/components/ErrorDecoder.tsx
```diff
@@ -0,0 +1,94 @@
...
+          {message}
+        </a>
+      );
+    }
+    return message;
+  });
+}
+
+// `?invariant=123&args[]=foo&args[]=bar`
+// or `// ?invariant=123&args[0]=foo&args[1]=bar`
```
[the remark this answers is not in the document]
SukkaW: Updated in [`90586e4` (#6214)](https://github.com/reactjs/react.dev/pull/6214/commits/90586e42ad98c72de3c285087ca231c51bdfa62d).

src/components/ErrorDecoder.tsx
SukkaW: Done in [`90586e4` (#6214)](https://github.com/reactjs/react.dev/pull/6214/commits/90586e42ad98c72de3c285087ca231c51bdfa62d).

src/components/ErrorDecoder.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+function replaceArgs(
+  msg: string,
+  argList: Array<string | undefined>,
+  replacer = '[missing argument]'
+): string {
+  let argIdx = 0;
+  return msg.replace(/%s/g, function () {
+    const arg = argList[argIdx++];
+    // arg can be an empty string: ?invariant=377&args[0]=&args[1]=count
```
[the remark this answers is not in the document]
SukkaW: Updated in [`90586e4` (#6214)](https://github.com/reactjs/react.dev/pull/6214/commits/90586e42ad98c72de3c285087ca231c51bdfa62d).

src/content/index.md
```diff
@@ -4,4 +4,4 @@ title: React
 permalink: index.html
 ---
 
-{/* See HomeContent.js */}
\ No newline at end of file
+{/* See HomeContent.js */}
```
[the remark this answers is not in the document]
SukkaW: The change has been discarded in [`56bac50` (#6214)](https://github.com/reactjs/react.dev/pull/6214/commits/56bac504dc96b00a63c061a2d8d17a8203fa9879).

src/pages/[[...markdownPath]].js
Comment (SukkaW) [reply src/pages/[[...markdownPath]].js]: Reverted in [`90586e4` (#6214)](https://github.com/reactjs/react.dev/pull/6214/commits/90586e42ad98c72de3c285087ca231c51bdfa62d).

src/components/ErrorDecoder.tsx
rickhanlonii: Ok that's a bummer. I also see that there are multiple updates that happen before `query` and `isReady` are set, delaying the time it takes to render the final message.

Instead, why don't we read the query params ourselves instead of from the router. That will allow us to set the message in an effect, which fires immediately after hydration, and that will set it at the first moment possible.

Finally, we can add a fade in animation to avoid the flash of unparsed args. Something like this:

```js
function parseQueryString(search: string): Array<string> {
  const rawQueryString = search.substring(1);
  if (!rawQueryString) {
    return [];
  }

  let args = [];

  const queries = rawQueryString.split('&');
  for (let i = 0; i < queries.length; i++) {
    const query = decodeURIComponent(queries[i]);
    if (query.indexOf('args[') === 0) {
      args.push(query.slice(query.indexOf(']=') + 2));
    }
  }

  return args;
}

export default function ErrorDecoder({errorMessages}: ErrorDecoderProps) {
  const [message, setMessage] = useState<string | null>(errorMessages);

  useEffect(() => {
    if (errorMessages == null) {
      return;
    }
    setMessage(
      replaceArgs(
        errorMessages,
        parseQueryString(window.location.search || ''),
        '[missing argument]'
      )
    );
  }, [errorMessages]);

  const displayMessage = useMemo(() => {
    if (!message) {
      return null;
    }
    return urlify(
      replaceArgs(message, parseQueryString(''), '[missing argument]')
    );
  }, [message]);

  if (!displayMessage) {
    return (
      <p>
        When you encounter an error, you{"'"}ll receive a link to this page for
        that specific error and we{"'"}ll show you the full error text.
      </p>
    );
  }

  return (
    <div>
      <p>
        <b>The full text of the error you just encountered is:</b>
      </p>
      <code className="block bg-red-100 text-red-600 py-4 px-6 mt-5 rounded-lg animate-fade-up">
        <b>{displayMessage}</b>
      </code>
    </div>
  );
}
```

```js
// add to tailwind.config.js
{
  theme: {
    extend: {
      animation: {
        'fade-up': 'fade-up 1s both',
        // ...
      } ,
      keyframes: {
        'fade-up': {
          '0%': {
            opacity: '0',
            transform: 'translateY(2rem)',
          },
          '100%': {
            opacity: '1',
            transform: 'translateY(0)',
          },
        },
        // ...
      }
      // ...
    }
    // ...
  }

```

Here's what the animation looks like:


https://github.com/reactjs/react.dev/assets/2440089/e3a9b26d-ac41-4575-8cb8-3d7913b30c41

src/pages/errors/[error_code].tsx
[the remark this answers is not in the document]
Comment (rickhanlonii) [reply src/pages/errors/[error_code].tsx:30]: Unfortunately that would cause even more conflicts for the translations, because some folks might start translating this now and then we would change the code. Is it OK if I push a change to add it?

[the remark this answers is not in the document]
Comment (SukkaW) [reply src/pages/errors/[error_code].tsx:30]: > Unfortunately that would cause even more conflicts for the translations, because some folks might start translating this now and then we would change the code. Is it OK if I push a change to add it?

Sure! Feel free to do so. `Allow edits and access to secrets by maintainers` has always been enabled.

src/components/ErrorDecoder.tsx
SukkaW: @rickhanlonii Done in [`4383ac9` (#6214)](https://github.com/reactjs/react.dev/pull/6214/commits/4383ac921d3eb568b5b7bdc8fc787b82c276a310)

Note that, inside `useEffect` I add an early return when the `errorMessages` doesn't contain `%s`, to avoid extra re-renders for most of the errors.

SukkaW: @rickhanlonii Would you like to review the change in [`4383ac9` (#6214)](https://github.com/reactjs/react.dev/pull/6214/commits/4383ac921d3eb568b5b7bdc8fc787b82c276a310)?

In the meantime, I am going to start addressing `<ErrorDecoder>` for MDX you have mentioned.

rickhanlonii:
Ok, this is great, I pushed some updates:

- Refactored to a single `compileMDX` function used by `[errorCode]` and `[[…markdownPath]]`.
- Removed the animation because it’s distracting. Instead, if there are no params we just display the error. If there are params, we delay showing the error until they’re inserted, like the old site did.
- Moved some of the text that needs translated into the markdown files, so the `ErrorDecoder` only shows the non-translatable error content.
- Added a `generic.md` separate from `index.md`. We may want to do more with `index.md` like list common errors with links to the page, so this keeps them separate, and I updated the content in `index.md`.
- Renamed `errorMessages` to `errorMessage` since it’s singular
- Added `errorCode` to the `ErrorDecoderContext` because I plan to use it soon. I have a WIP that uses it but I don’t want to block on it. 
- Moved `ErrorDecoderContext` out of `/MDX`, since it’s not needed in MDX files

src/pages/errors/[errorCode].tsx
```diff
@@ -0,0 +1,157 @@
...
+  let path = params?.errorCode || 'index';
+  let mdx;
+  try {
+    mdx = fs.readFileSync(rootDir + '/' + path + '.md', 'utf8');
+  } catch {
+    // if [errorCode].md is not found, fallback to generic.md
+    mdx = fs.readFileSync(rootDir + '/generic.md', 'utf8');
+  }
+
+  const {content, toc, meta} = await compileMDX(mdx, path, {code, errorCodes});
```
Comment (rickhanlonii) [inline src/pages/errors/[errorCode].tsx:124]: I also added the `errorCodes` to the cache, since when new errors are pushed upstream we need to break the cache and rebuild the pages. It's not often, but we do change the error message text.

PR #6227, opened 2023-08-15
Update correct link in select.md

NGHdeveloper:
Correct link


mattcarrollcode:

PR #6229, opened 2023-08-15
Clarify these features are not related to Next

lunaleaps:
There was some discussion online that the warning mentioning Next may give the impression this is a NextJS feature. 

What if we leave it just as a blank warning for now? I can see the appeal of linking to something vs. nothing but I do hope to get these fleshed out soon. What do you think?

### DIFF
--- a/src/content/reference/react/use-client.md
+++ b/src/content/reference/react/use-client.md
@@ -50,8 +50,4 @@ When a file marked `'use client'` is imported from a server component, [compatib
 
 ## Usage {/*usage*/}
 
-<Wip>
-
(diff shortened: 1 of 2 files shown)


sophiebits (approved):
I guess this is fine, could also go with something like

> This API can be used in any framework that supports React Server Components, but the Next.js documentation may answer additional questions.

rickhanlonii:
@sophiebits @lunaleaps yeah, and when other frameworks add support and docs, we can also add links to those docs.

> This API can be used in any framework that supports React Server Components: 
> - <link>Next.js Server Components</link>
> - More coming soon

PR #6232, opened 2023-08-16
Fix link in new typescript docs

maxjacobson:
While reading the new typescript docs introduced in #6120, I noticed this link points to a fork. This commit changes it to a relative link so it should work everywhere.


mattcarrollcode:

PR #6240, opened 2023-08-18
Add cache API docs

lunaleaps:
Initial docs for cache API 

## [Preview](https://react-dev-git-fork-lunaleaps-cache-docs-fbopensource.vercel.app/reference/react/cache)

### DIFF
--- a/src/content/reference/react/cache.md
+++ b/src/content/reference/react/cache.md
@@ -6,7 +6,7 @@ canary: true
 <Canary>
 * `cache` is only for use with [React Server Components](https://react.dev/blog/2023/03/22/react-labs-what-we-have-been-working-on-march-2023#react-server-components). See [frameworks](https://react.dev/learn/start-a-new-react-project#bleeding-edge-react-frameworks) that support React Server Components.
 
-* `cache` is currently only available in React’s [Canary](https://react.dev/community/versioning-policy#canary-channel) and [experimental](https://react.dev/community/versioning-policy#experimental-channel) channels. Please ensure you understand the limitations before using `cache` in production. Learn more about [React's release channels here](/community/versioning-policy#all-release-channels).
+* `cache` is only available in React’s [Canary](https://react.dev/community/versioning-policy#canary-channel) and [experimental](https://react.dev/community/versioning-policy#experimental-channel) channels. Please ensure you understand the limitations before using `cache` in production. Learn more about [React's release channels here](/community/versioning-policy#all-release-channels).
 </Canary>
 
 <Intro>
@@ -41,7 +41,7 @@ function Chart({data}) {
 }
 ```
 
-`getMetrics` will first check if there is a cached result for `data`, otherwise call `calculateMetrics(data)` and cache the returned result.
+When `getMetrics` is first called with `data`, `getMetrics` will call `calculateMetrics(data)` and store the result in cache. If `getMetrics` is called again with the same `data`, it will return the cached result instead of calling `calculateMetrics(data)` again.
 
 [See more examples below.](#usage)
 
@@ -68,7 +68,7 @@ The optimization of caching return values based on inputs is known as [_memoizat
 - React will invalidate the cache for all memoized functions for each server request. 
 - Each call to `cache` creates a new function. This means that calling `cache` with the same function multiple times will return different memoized functions that do not share the same cache.
 - `cachedFn` will also cache errors. If `fn` throws an error for certain arguments, it will be cached, and the same error is re-thrown when `cachedFn` is called with those same arguments.
-- `cache` is for use in [Server Components](https://react.dev/blog/2023/03/22/react-labs-what-we-have-been-working-on-march-2023#react-server-components) only. There are plans to introduce `cache` for Client Components, but it is not a supported feature today. There are plans to add a warning when `cache` is used in Client Components.
+- `cache` is for use in [Server Components](https://react.dev/blog/2023/03/22/react-labs-what-we-have-been-working-on-march-2023#react-server-components) only.
 
 ---
 
@@ -105,9 +105,12 @@ Assume `Profile` is rendered first. It will call <CodeStep step={1}>`getUserMetr
 When `TeamReport` renders its list of `users` and reaches the same `user` object, it will call <CodeStep step={2}>`getUserMetrics`</CodeStep> and read the result from cache.
 
 <Pitfall>
-Calling different memoized functions will read from different caches. To access the same cache, components must call the same memoized function.
 
-```js [[1, 7, "getWeekReport = cache(calculateWeekReport)"], [1, 8, "getWeekReport(cityData)"], [2, 17, "getWeekReport = cache(calculateWeekReport)"], [2, 20, "getWeekReport(cityData)"]]
+##### Calling different memoized functions will read from different caches. {/*pitfall-different-memoized-functions*/}
+
+To access the same cache, components must call the same memoized function.
+
+```js [[1, 7, "getWeekReport"], [1, 7, "cache(calculateWeekReport)"], [1, 8, "getWeekReport"]]
 // Temperature.js
 import {cache} from 'react';
 import {calculateWeekReport} from './report';
@@ -115,10 +118,12 @@ import {calculateWeekReport} from './report';
 export function Temperature({cityData}) {
   // 🚩 Wrong: Calling `cache` in component creates new `getWeekReport` for each render
   const getWeekReport = cache(calculateWeekReport);
-	const report = getWeekReport(cityData);
+  const report = getWeekReport(cityData);
   // ...
 }
+```
 
+```js [[2, 6, "getWeekReport"], [2, 6, "cache(calculateWeekReport)"], [2, 9, "getWeekReport"]]
 // Precipitation.js
 import {cache} from 'react';
 import {calculateWeekReport} from './report';
@@ -138,13 +143,15 @@ In addition, `Temperature` creates a <CodeStep step={1}>new memoized function</C
 
 To maximize cache hits and reduce work, the two components should call the same memoized function to access the same cache. Instead, define the memoized function in a dedicated module that can be [`import`-ed](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import) across components.
 
-```js [[3, 5, "export default cache(calculateWeekReport)"], [3, 8, "getWeekReport", 0], [3, 11, "getWeekReport(cityData)"], [3, 16, "getWeekReport", 0], [3, 19, "getWeekReport(cityData)"]]
+```js [[3, 5, "export default cache(calculateWeekReport)"]]
 // getWeekReport.js
 import {cache} from 'react';
 import {calculateWeekReport} from './report';
 
 export default cache(calculateWeekReport);
+```
 
+```js [[3, 2, "getWeekReport", 0], [3, 5, "getWeekReport"]]
 // Temperature.js
 import getWeekReport from './getWeekReport';
 
@@ -152,7 +159,9 @@ export default function Temperature({cityData}) {
 	const report = getWeekReport(cityData);
   // ...
 }
+```
 
+```js [[3, 2, "getWeekReport", 0], [3, 5, "getWeekReport"]]
 // Precipitation.js
 import getWeekReport from './getWeekReport';
 
@@ -279,7 +288,7 @@ If the promise is already settled, either to an error or the _fulfilled_ result,
 
 <Pitfall>
 
-Calling a memoized function outside of a component will not use the cache.
+##### Calling a memoized function outside of a component will not use the cache. {/*pitfall-memoized-call-outside-component*/}
 
 ```jsx [[1, 3, "getUser"]]
 import {cache} from 'react';
@@ -288,7 +297,7 @@ const getUser = cache(async (userId) => {
   return await db.user.query(userId);
 });
 
-// 🚩 Wrong: Calling memoized function outside of render will not memoize.
+// 🚩 Wrong: Calling memoized function outside of component will not memoize.
 getUser('demo-id');
 
 async function DemoProfile() {
@@ -332,7 +341,7 @@ function App() {
   );
 }
 ```
-In this example, `App` renders two `WeatherReport`s with the same record. Even though both components do the same work, they cannot share work. `useMemo` is only a local cache to the component instance. 
+In this example, `App` renders two `WeatherReport`s with the same record. Even though both components do the same work, they cannot share work. `useMemo`'s cache is only local to the component.
 
 However, `useMemo` does ensure that if `App` re-renders and the `record` object doesn't change, each component instance would skip work and use the memoized value of `avgTemp`. `useMemo` will only cache the last computation of `avgTemp` with the given dependencies. 
 
@@ -399,11 +408,15 @@ Compared to `useMemo`, `memo` memoizes the component render based on props vs. s
 
 ### My memoized function still runs even though I've called it with the same arguments {/*memoized-function-still-runs*/}
 
-Beyond the mentioned Pitfalls, it may be a problem with how React checks if something exists in cache.
+See prior mentioned pitfalls
+* [Calling different memoized functions will read from different caches.](#pitfall-different-memoized-functions)
+* [Calling a memoized function outside of a component will not use the cache.](#pitfall-memoized-call-outside-component)
+
+If none of the above apply, it may be a problem with how React checks if something exists in cache.
+
+If your arguments are not [primatives](https://developer.mozilla.org/en-US/docs/Glossary/Primitive) (ex. objects, functions, arrays), ensure you're passing the same object reference.
 
 When calling a memoized function, React will look up the input arguments to see if a result is already cached. React will use shallow equality of the arguments to determine if there is a cache hit.
-
-This means that if your arguments are not [primatives](https://developer.mozilla.org/en-US/docs/Glossary/Primitive) (ex. objects, functions, arrays), ensure you're passing the same object reference.
 
 ```js
 import {cache} from 'react';
--- a/src/content/reference/react/cache.md
+++ b/src/content/reference/react/cache.md
@@ -79,7 +79,7 @@ The optimization of caching return values based on inputs is known as [_memoizat
 
 ### Share a snapshot of data {/*take-and-share-snapshot-of-data*/}
 
-Wrap `cache` around a data-fetching function and use in multiple components. If multiple components make the same data fetch, only one request is made and the data returned is cached and shared across components. In a sense, all components refer to the same snapshot of data across the render.
+To share a snapshot of data between multiple components, wrap `cache` around a data-fetching function like `fetch`. When multiple components make the same data fetch, only one request is made and the data returned is cached and shared across components. In a sense, all components refer to the same snapshot of data across each render. 
 
 ```jsx
 import {cache} from 'react';
--- a/src/content/reference/react/cache.md
+++ b/src/content/reference/react/cache.md
@@ -14,7 +14,7 @@ Learn more about [React's release channels here](/community/versioning-policy#al
 `cache` lets you cache the result of a data fetch or computation.
 
 ```js
-const cachingFn = cache(fn);
+const cachedFn = cache(fn);
 ```
 
 </Intro>
@@ -27,7 +27,7 @@ const cachingFn = cache(fn);
 
 ### `cache(fn)` {/*cache*/}
 
-Wrap a function with `cache` to receive a caching version of that function.
+Wrap a function with `cache` to receive a version of that function with caching.
 ```jsx
 import { cache } from 'react';
 import fetchUser from 'lib/user';
@@ -39,7 +39,7 @@ function UserProfile(id) {
   //...
 }
 ```
-`getUser` will first see if there is a cached result for `id`, otherwise call `fetchUser(id)` and cache the returned result.
+`getUser` will first check if there is a cached result for `id`, otherwise call `fetchUser(id)` and cache the returned result.
 
 [See more examples below.](#usage)
 
@@ -49,25 +49,25 @@ function UserProfile(id) {
 
 #### Returns {/*returns*/}
 
-`cache` returns a caching version of `fn` with the same type signature.
+`cache` returns a cached version of `fn` with the same type signature.
 
-When calling `cachingFn` with given arguments, it first checks if a cached result exists in the cache. If a cached result exists, it returns the result. If not, it calls `fn` with the arguments, stores the result in the cache, and returns the result.
+When calling `cachedFn` with given arguments, it first checks if a cached result exists in the cache. If a cached result exists, it returns the result. If not, it calls `fn` with the arguments, stores the result in the cache, and returns the result.
 
 <Note>
 
-Caching return values based on inputs is also known as [*memoization*](https://en.wikipedia.org/wiki/Memoization). We refer to `cachingFn` as a memoizing function.
+The optimization of caching return values based on inputs is known as [*memoization*](https://en.wikipedia.org/wiki/Memoization). We refer to `cachedFn` as a memoized function.
 
 </Note>
 
 #### Caveats {/*caveats*/}
 
 [//]: # (TODO: add links to Server/Client Component reference once https://github.com/reactjs/react.dev/pull/6177 is merged)
-* `cache` is recommended for React Server Component environments only. There are plans to introduce `cache` for Client Components but it is not recommended today.
-* React will invalidate the cache for a memoizing function across server requests.
-* The benefit of `cache` is to skip duplicate work across components. To leverage this benefit, `cachingFn`, should be defined in a scope that is accessible to components. In most cases, this means calling `cache` and defining the memoizing function in the global scope.
-* Each call to `cache` creates a new memoizing version of `fn`. This means that wrapping `cache` around the same function multiple times will return different memoizing functions and they will not share the same cache.
-* Cache access only occurs during a component render. This means a call to `cachingFn` outside of a Server Component will not update or use the cache.
-* `cachingFn` will also cache errors. If `fn` throws an error for certain arguments, it will be cached, and rethrown when `cachingFn` is called with those arguments. 
+* `cache` is recommended for Server Components only. There are plans to introduce `cache` for Client Components but it is not recommended today.
+* React will invalidate the cache for a memoized function across server requests.
+* Each call to `cache` creates a new function. This means that wrapping `cache` around the same function multiple times will return different memoized functions and they will not share the same cache.
+* The benefit of `cache` is to skip duplicate work by sharing a cache. To promote cache sharing, `cachedFn`, should be defined in a scope that is accessible to multiple components. In most cases, this means calling `cache` and defining the memoized function in the global scope.
+* Cache access only occurs during a component render. This means a call to `cachedFn` outside of a Server Component will not update or use the cache.
+* `cachedFn` will also cache errors. If `fn` throws an error for certain arguments, it will be cached, and re-thrown when `cachedFn` is called with those arguments. 
 
 ---
 
@@ -99,7 +99,7 @@ If `AnimatedWeatherCard` and `MinimalWeatherCard` both render for the same city,
 
 <Pitfall>
 
-To leverage the cache, components should use the same memoizing function instance.
+To order to share the cache, components must use the same memoized function.
 
 ```js
 // temperature.js
@@ -107,7 +107,7 @@ export async function fetchTemperature(city) {
   return await fetch(`https://...`);
 }
 
-// minimalWeatherCard.js
+// MinimalWeatherCard.js
 import {cache} from 'react';
 import {fetchTemperature} from './temperature';
 
@@ -118,19 +118,20 @@ export default async function MinimalWeatherCard({city}) {
   // ...
 }
 
-// animatedWeatherCard.js
+// AnimatedWeatherCard.js
 import {cache} from 'react';
 import {fetchTemperature} from './temperature';
 
 export default async function AnimatedWeatherCard({city}) {
+  // 🚩 Wrong: cache should not be called in render.
 	const getTemperature = cache(fetchTemperature);
 	const temperature = await getTemperature(city);
   // ...
 }
 
 // app.js
-import AnimatedWeatherCard from './animatedWeatherCard';
-import AnimatedWeatherCard from './animatedWeatherCard';
+import AnimatedWeatherCard from './AnimatedWeatherCard';
+import MinimalWeatherCard from './MinimalWeatherCard';
 
 function App() {
 	return (<>
@@ -142,9 +143,9 @@ function App() {
 
 In this example, `MinimalWeatherCard` and `AnimatedWeatherCard` are not leveraging the same cache and so `fetchTemperature` is double invoked with the same argument `"Los Angeles"`.
 
-This is because each call to `cache` creates a new memoizing function that uses a different cache for look-up. For `AnimatedWeatherCard` and `MinimalWeatherCard` to use the same cache, `getTemperature` should be defined in a mutually accessible scope, like moving the definition to `temperature.js`.
+This is because each call to `cache` creates a new memoized function that uses a different cache for look-up. For `AnimatedWeatherCard` and `MinimalWeatherCard` to use the same cache, `getTemperature` should be defined in a mutually accessible scope, like moving the definition to `temperature.js`.
 
-In addition, `AnimatedWeatherCard` is calling `cache` in its component render. This means that each time `AnimatedWeatherCard` is rendered, a new memoizing function is created so there will never be an opportunity for cache-sharing.
+In addition, `AnimatedWeatherCard` is calling `cache` in its component render. This means that each time `AnimatedWeatherCard` is rendered, a new memoized function is created so there will never be an opportunity for cache-sharing.
 
 </Pitfall>
 
@@ -159,10 +160,14 @@ const getUser = cache(async (id) => {
 
 async function Profile({id}) {
   const user = await getUser(id);
-  // ...
+  return (<section>
+    <img src={user.profilePic} />
+    <h2>{user.name}</h2>
+    </section>);
 }
 
 function Page({id}) {
+  // ✅ Good: start fetching the user data 
   getUser(id);
   // ... some computational work
   return (
@@ -189,17 +194,17 @@ const getUser = cache(async (userId) => {
   return await db.user.query(userId);    
 });
 
-// This call to `getUser` will not update the cache.
+// 🚩 Wrong: This will not update the cache as its not in a component render.
 getUser('demo-id');
 
 async function DemoProfile() {
-  // This `getUser` will have a cache miss
+  // ✅ Good: This `getUser` will access the cache.
   const user = await getUser('demo-id');
-  return (<Profile user={user} />);
+  return <Profile user={user} />;
 }
 ```
 
-You may attempt to preload some work outside of a component. React only provides cache access to the memoizing function in a component render. 
+You may attempt to preload some work outside of a component. React only provides cache access to the memoized function in a component render. 
 </Pitfall>
 
 <DeepDive>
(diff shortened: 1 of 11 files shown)


src/content/reference/react/cache.md
```diff
@@ -0,0 +1,305 @@
...
+When rendering `Page`, the component calls `getUser` but note that it doesn't use the returned data. This early `getUser` call kicks off the asynchronous database query that occurs while `Page` is doing other computational work and rendering children. 
+
+When rendering `Profile`, we call `getUser` again. In the ideal case, the initial `getUser` call has already returned and cached the user data. So when `Profile` asks and waits for this data, it can simply read from the cache without requiring another remote procedure call. Even if the data request hasn't been completed, preloading data in this pattern reduces delay in data-fetching..
+
+<Pitfall>
+
+Memoization only occurs within component renders.
+
+```jsx
+import {cache} from "react";
```
lunaleaps, line 185:
```suggestion
import {cache} from 'react';
```

src/content/reference/react/cache.md
```diff
@@ -0,0 +1,305 @@
...
+getUser('demo-id');
+
+async function DemoProfile() {
+  // This `getUser` will have a cache miss
+  const user = await getUser('demo-id');
+  return (<Profile user={user} />);
+}
+```
+
+You may attempt to preload some work outside of a component. React only provides cache access to the memoizing function in a component render.
```
lunaleaps, line 201: I think I may have missed why this is a limitation, does anyone know?

rickhanlonii:
Added a branch preview to the PR description for easy access.

rickhanlonii:
Pushed additional canary icons

Screenshot 2023-08-18 at 3 17 44 PM

src/content/reference/react/cache.md
```diff
@@ -0,0 +1,306 @@
...
+	    ...
+      <Profile id={id} />
+    </>
+  );
+}
+```
+
+When rendering `Page`, the component calls `getUser` but note that it doesn't use the returned data. This early `getUser` call kicks off the asynchronous database query that occurs while `Page` is doing other computational work and rendering children. 
+
+When rendering `Profile`, we call `getUser` again. In the ideal case, the initial `getUser` call has already returned and cached the user data. So when `Profile` asks and waits for this data, it can simply read from the cache without requiring another remote procedure call. Even if the data request hasn't been completed, preloading data in this pattern reduces delay in data-fetching..
```
rickhanlonii, line 179: It would be neat to have a `<DeepDive /> here that explain how promises work in JS and why this actually works.

src/content/reference/react/cache.md
```diff
@@ -0,0 +1,306 @@
...
+
+<Pitfall>
+
+Memoization only occurs within component renders.
```
rickhanlonii, line 183:
```suggestion
Cache is only saved during component renders.
```

src/content/reference/react/cache.md
```diff
@@ -0,0 +1,306 @@
...
+  return await db.user.query(userId);    
+});
+
+// This call to `getUser` will not update the cache.
+getUser('demo-id');
+
+async function DemoProfile() {
+  // This `getUser` will have a cache miss
+  const user = await getUser('demo-id');
+  return (<Profile user={user} />);
```
rickhanlonii, line 198: No parens needed.

src/content/reference/react/cache.md
rickhanlonii: The [cache uses context](https://github.com/facebook/react/blob/f101c2d0d3a6cb5a788a3d91faef48462e45f515/packages/react-reconciler/src/ReactFiberCache.js#L29)

src/content/reference/react/cache.md
```diff
@@ -0,0 +1,306 @@
...
+[`useMemo`](/reference/react/useMemo) is a hook that is used in a client component or another hook. `useMemo` takes a function and a list of dependencies. During the component render, `useMemo` will only call the function when the dependencies have changed and will cache the result. 
+
+You can think of `useMemo`'s cache as a size of one, where it can only cache the last work it did based on the last value of its dependencies. Only the component instance can benefit from this cache to skip work across re-renders. If you rendered two instances of the same component, the `useMemo` of one wouldn't help the other to skip work. The cache is invalidated each time the dependencies change or the component is unmounted.
+
+```jsx
+'use client';
+
+function Weather(city) {
+    // assume `getTemp` is an expensive computation
+	const temp = useMemo(() => getTemp(city)), city);
```
rickhanlonii, line 219: Nit: `getTemp` sounds like a data fetch, which isn't allowed in `useMemo`. Maybe `calculateTemp` instead, or something more realistic like `filterTodos`.

rickhanlonii: Maybe the issue here is that we're comparing `cache` with `useMemo` when it's really more like `useCallback`.

src/content/reference/react/cache.md
```diff
@@ -0,0 +1,306 @@
...
+function App() {
+  return (<>
+    <MapMarker x={10} y={10} z={10} />
+    <MapMarker x={10} y={10} z={10} />
+  </>)
+}
+```
+In this case the two `MapMarker`s look like they're doing the same work and calling `calculateNorm` with the same value of `{x: 10, y: 10, z:10}`. Even though the objects contain the same values, they are not the same object reference as each component creates its own `props` object. 
+
+React will call [`Object.is`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) on the input to verify if there is a cache hit.
```
rickhanlonii, line 306: Can we add an example of the right way to do it?

src/content/reference/react/cache.md
```diff
@@ -0,0 +1,306 @@
...
+
+```jsx
+import {cache} from 'react';
+
+const calculateNorm = cache((vector) => {
+  // ...
+});
+
+function MapMarker(props) {
+  const length = calculateNorm(props);
```
rickhanlonii, line 293:
```suggestion
  // 🚩 Wrong: props is an object that changes every render.
  const length = calculateNorm(props);
```

src/content/reference/react/cache.md
```diff
@@ -0,0 +1,306 @@
...
+
+Learn more about [React's release channels here](/community/versioning-policy#all-release-channels). 
+</Canary>
+
+<Intro>
+
+`cache` lets you cache the result of a data fetch or computation.
+
+```js
+const cachingFn = cache(fn);
```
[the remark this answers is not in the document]
lunaleaps: Yea I've struggled with naming on this. What would be the difference for _does_ vs. _with_? The function returned will cache the result given the input. Is that not a caching function? I'm not strongly opposed to say "function _with_ caching" but want to build up a strong mental model for it

src/content/reference/react/cache.md
```diff
@@ -0,0 +1,306 @@
...
+
+#### Returns {/*returns*/}
+
+`cache` returns a caching version of `fn` with the same type signature.
+
+When calling `cachingFn` with given arguments, it first checks if a cached result exists in the cache. If a cached result exists, it returns the result. If not, it calls `fn` with the arguments, stores the result in the cache, and returns the result.
+
+<Note>
+
+Caching return values based on inputs is also known as [*memoization*](https://en.wikipedia.org/wiki/Memoization). We refer to `cachingFn` as a memoizing function.
```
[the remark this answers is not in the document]
lunaleaps: Hmm, yea this is a great discussion point and something I went back and forth on.

I'm fine with saying `memoized` but I guess what does the verb `memoize` mean? When I think of something as "memoized" in the past tense, I think of it as in the cache. _"The result for a certain input is memoized"_.

For `memo`, you are memoizing the one component for its last prop values. But for `cache`, you're returned a function that has memoizing ability. The function returned itself is not "memoized". It adds the special ability to memoize future inputs to that returned function. Does this perspective make sense?

src/content/reference/react/cache.md
```diff
@@ -0,0 +1,306 @@
...
+* `cache` is recommended for React Server Component environments only. There are plans to introduce `cache` for Client Components but it is not recommended today.
+* React will invalidate the cache for a memoizing function across server requests.
+* The benefit of `cache` is to skip duplicate work across components. To leverage this benefit, `cachingFn`, should be defined in a scope that is accessible to components. In most cases, this means calling `cache` and defining the memoizing function in the global scope.
+* Each call to `cache` creates a new memoizing version of `fn`. This means that wrapping `cache` around the same function multiple times will return different memoizing functions and they will not share the same cache.
+* Cache access only occurs during a component render. This means a call to `cachingFn` outside of a Server Component will not update or use the cache.
+* `cachingFn` will also cache errors. If `fn` throws an error for certain arguments, it will be cached, and rethrown when `cachingFn` is called with those arguments. 
+
+---
+
+## Usage {/*usage*/}
```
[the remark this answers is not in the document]
lunaleaps: Yea definitely. I saw Matt's `use` docs leveraging this. Will do a second pass later and add more concrete highlights in my explanations

src/content/reference/react/cache.md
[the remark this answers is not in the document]
lunaleaps: Somewhat related to your point, I'm intentionally not using the term "memoize" until the note mentions it

src/content/reference/react/cache.md
```diff
@@ -0,0 +1,306 @@
+---
+title: cache
+canary: true
+---
+
+<Canary>
+`cache` is currently only available in React’s [canary](https://react.dev/community/versioning-policy#canary-channel) and [experimental](https://react.dev/community/versioning-policy#experimental-channel) channels. Please ensure you understand the limitations before using `cache` in production.
```
smikitky, line 7:
```suggestion
`cache` is currently only available in React’s [Canary](https://react.dev/community/versioning-policy#canary-channel) and [Experimental](https://react.dev/community/versioning-policy#experimental-channel) channels. Please ensure you understand the limitations before using `cache` in production. 
```

"Canary" is usually capitalized throughout the docs.

src/content/reference/react/cache.md
lunaleaps: I think the memoized function returned from `cache` is more like `useMemo`. It is caching the returned value of the function. I can update the example

src/content/reference/react/cache.md
```diff
@@ -0,0 +1,443 @@
...
+```js {3,9}
+import {cache} from 'react';
+
+const calculateNorm = cache((x, y, z) => {
+  // ...
+});
+
+function MapMarker(props) {
+  // ✅ Good: Pass primitives to memoized function
+  const length = calculateNorm(props.x, props.y, props.z);
```
eps1lon, line 402: Especially for the preload example, it seems more ergonomic to still use the `preload(props)` pattern. So as an alternative, your `calculateNorm` would be a wrapper that unpacks the props instead. That keeps the order of arguments next to the implementation and people don't have to recall the right order of arguments:


```js
const cachedCalculateNorm = cache((x, y, z) => {
  // ...
});
const calculateNorm = props => cachedCalculateNorm(props.x, props.y, props.z)
```

The additional implementation boilerplate may seem a bit annoying but this is probably nice to keep handy for people complaining about argument order. Especially for cached functions with a lot of spread out callsites. If you don't share the cached function across modules, it's probably not worth adding the "unpacking objects"-wrapper.

I don't think we need to include this pattern but probably good to keep in mind if people focus too much on being annoyed with recalling argument order.

src/content/reference/react/cache.md
```diff
@@ -0,0 +1,443 @@
...
+
+<Note>
+
+[//]: # 'TODO: add links to Server Components when merged.'
+
+<CodeStep step={3}>Asynchronous rendering</CodeStep> is only supported for Server Components.
+
+```js [[3, 1, "async"], [3, 2, "await"]]
+async function AnimatedWeatherCard({city}) {
+	const temperature = await getTemperature(city);
```
eps1lon, line 113: I love how the highlighting connects "Asynchronous rendering" with `async/await`.

src/content/reference/react/cache.md
```diff
@@ -0,0 +1,443 @@
...
+#### Caveats {/*caveats*/}
+
+[//]: # 'TODO: add links to Server/Client Component reference once https://github.com/reactjs/react.dev/pull/6177 is merged'
+
+- `cache` is recommended for Server Components only. There are plans to introduce `cache` for Client Components, but it is not recommended today.
+- React will invalidate the cache for a memoized function across server requests.
+- Each call to `cache` creates a new function. This means that wrapping `cache` around the same function multiple times will return different memoized functions and they will not share the same cache.
+- The benefit of `cache` is to skip duplicate work by sharing a cache. To promote cache sharing, `cachedFn`, should be defined in a scope that is accessible to multiple components. In most cases, this means calling `cache` and defining the memoized function in the global scope.
+- Cache access only occurs during a component render. This means a call to `cachedFn` outside of a Server Component will not update or read the cache.
+- `cachedFn` will also cache errors. If `fn` throws an error for certain arguments, it will be cached, and re-thrown when `cachedFn` is called with those arguments.
```
eps1lon, line 74: Will it re-throw with the original callstack or the callstack that's specific to the current call?

eps1lon (approved):

src/content/reference/react/cache.md
```diff
@@ -0,0 +1,443 @@
...
+import {fetchTemperature} from './api/temperature';
+
+export default async function WeatherCard({city}) {
+  // 🚩 Wrong: cache should not be called in render.
+  const getTemperature = cache(fetchTemperature);
+  const temperature = await getTemperature(city);
+  // ...
+}
+```
+Each time `WeatherCard` is rendered, a new memoized function is created that will have it's own cache look-up. This means, across renders, `getTemperature` is referring to a new memoized function and there will be no cache-sharing, losing the benefit of memoiziation.
```
lunaleaps, line 137:
```suggestion
For every render, `WeatherCard` creates a new memoized function that will have its own cache look-up. Across renders, `getTemperature` refers to the new memoized function, losing the benefit of cache-sharing and memoization.
```

src/content/reference/react/cache.md
```diff
@@ -0,0 +1,443 @@
...
+
+```js {3,9,14}
+import {cache} from 'react';
+
+const calculateNorm = cache((vector) => {
+  // ...
+});
+
+function MapMarker(props) {
+  // ✅ Good: Pass primitives to memoized function
```
lunaleaps, line 428:
```suggestion
  // ✅ Good: Pass the same `vector` object
```

src/content/reference/react/cache.md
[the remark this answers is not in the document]
rickhanlonii: > What would be the difference for does vs. with?

The `cache` function _does_ the caching, and returns an instance of the provided function with the caching that `cache` provides. When I read `cachingFn`, I read it as "a function that can cache something" which makes me think that I will use that function to then cache something, something like:

```js
const cachingFn = cache(fn);

// Wrong
const cachedFn = cachingFn(otherFn);
```

I think it's because caching is a verb, so I read it as a function that will do caching, instead of a function that has had caching done to it. Similar to memoize, the returned value is not a memoizing or caching function, it's a memoized or cached function.

TheSavior (approved):

src/content/reference/react/cache.md
```diff
@@ -0,0 +1,443 @@
...
+
+<InlineToc />
+
+---
+
+## Reference {/*reference*/}
+
+### `cache(fn)` {/*cache*/}
+
+Wrap a function with `cache` to receive a version of that function with caching.
```
yungsters, line 31: Borrowing phrasing from the other documentation pages (e.g. `createContext`), how about:

> Call `cache` outside of any components to create a version of the function with caching.

In particular, I find the "Call [...] outside of any components" to be pretty valuable in setting the expectation about where in a developer's code this function is intended to be used.

---

I also noticed that we refer to the behavior as "caching" instead of "memoization".

I understand if the intent is to make this more accessible to developers, but I couldn't help but notice that [the documentation for `memo`](https://react.dev/reference/react/memo) already uses "memoization".

src/content/reference/react/cache.md
```diff
@@ -0,0 +1,443 @@
...
+
+```js {4,7}
+import {cache} from 'react';
+import fetchUser from 'lib/user';
+
+const getUser = cache(fetchUser);
+
+function UserProfile(id) {
+  const user = getUser(id);
+  //...
```
yungsters, line 41:
```suggestion
  // ...
```

src/content/reference/react/cache.md
```diff
@@ -0,0 +1,443 @@
...
+}
+```
+
+`getUser` will first check if there is a cached result for `id`, otherwise call `fetchUser(id)` and cache the returned result.
```
yungsters, line 45: Is it intentionally implied that `fetchUser` returns a `user` value`?

My expectation would be for a function named `fetchUser` to be async (and return a `Promise`). Therefore, I would expect `getUser` to also be async (and return a `Promise`).

And if this example *is* implying that they are both async, I might be confused why the example does not show:

```
const user = await getUser(id);
```

src/content/reference/react/cache.md
```diff
@@ -0,0 +1,443 @@
...
+
+[See more examples below.](#usage)
+
+#### Parameters {/*parameters*/}
+
+- `fn`: The function you want to cache results for. `fn` can take any arguments and return any value.
```
yungsters, line 51: Are there any expectations about whether it can return a `Promise` or not a `Promise`? (It might not need to be called out explicitly, but my question above *does* leave me to wonder…)

src/content/reference/react/cache.md
```diff
@@ -0,0 +1,443 @@
...
+
+#### Returns {/*returns*/}
+
+`cache` returns a cached version of `fn` with the same type signature.
+
+When calling `cachedFn` with given arguments, it first checks if a cached result exists in the cache. If a cached result exists, it returns the result. If not, it calls `fn` with the arguments, stores the result in the cache, and returns the result.
```
yungsters, line 57: This references `cachedFn`. Should it say `fn`?  (Same with some content after this paragraph.)

```suggestion
When calling `fn`, it first checks if a cached result exists in the cache for the given arguments. If a cached result exists, it returns the result. If not, it calls `fn` with the arguments, stores the result in the cache, and returns the result.
```

src/content/reference/react/cache.md
lunaleaps: Good question -- it looks like the original callstack: https://github.com/facebook/react/blob/456d153bb582798effa76c09bec2405ab2e392cf/packages/react/src/ReactCache.js#L122

src/content/reference/react/cache.md
```diff
@@ -0,0 +1,443 @@
...
+
+The optimization of caching return values based on inputs is known as [_memoization_](https://en.wikipedia.org/wiki/Memoization). We refer to `cachedFn` as a memoized function.
+
+</Note>
+
+#### Caveats {/*caveats*/}
+
+[//]: # 'TODO: add links to Server/Client Component reference once https://github.com/reactjs/react.dev/pull/6177 is merged'
+
+- `cache` is recommended for Server Components only. There are plans to introduce `cache` for Client Components, but it is not recommended today.
```
yungsters, line 69: > [...], but it is not recommended today.

This implies it is usable but may not work correctly. Is that the case? Or are we actively discouraging it? (And if we are, we could probably use stronger language here.)

src/content/reference/react/cache.md
```diff
@@ -0,0 +1,443 @@
...
+- React will invalidate the cache for a memoized function across server requests.
+- Each call to `cache` creates a new function. This means that wrapping `cache` around the same function multiple times will return different memoized functions and they will not share the same cache.
```
yungsters, line 71: Nit: The usage of "wrapping" feels appropriate in a guide or tutorial, but less appropriate for a technical reference document.

```suggestion
- Each call to `cache` creates a new function. This means that calling `cache` with the same function multiple times will return different memoized functions that do not share the same cache.
```

src/content/reference/react/cache.md
```diff
@@ -0,0 +1,443 @@
...
+- The benefit of `cache` is to skip duplicate work by sharing a cache. To promote cache sharing, `cachedFn`, should be defined in a scope that is accessible to multiple components. In most cases, this means calling `cache` and defining the memoized function in the global scope.
```
yungsters, line 72: Question: Are we intentionally promoting the use of global scope? Or should we be suggesting something else like module scope or separate modules (similar to our recommendation for context providers).

Consider borrowing phrasing from our existing documentation about [importing and exporting context from a file](https://react.dev/reference/react/createContext#importing-and-exporting-context-from-a-file).

src/content/reference/react/cache.md
```diff
@@ -0,0 +1,443 @@
...
+- Cache access only occurs during a component render. This means a call to `cachedFn` outside of a Server Component will not update or read the cache.
```
yungsters, line 73: What happens if called outside of a component render? (I believe you mentioned offline that it will passthrough to `fn`. This may be worth calling out, here or otherwise.)

src/content/reference/react/cache.md
```diff
@@ -0,0 +1,443 @@
...
+	// ...
+}
+
+async function MinimalWeatherCard({city) {
+	const temperature = await getTemperature(city);
+	// ...
+}
+```
+
+If `AnimatedWeatherCard` and `MinimalWeatherCard` both render for the same <CodeStep step={1}>city</CodeStep>, they will receive the same snapshot of data from the <CodeStep step={2}>memoized function</CodeStep>. The <CodeStep step={1}>city</CodeStep> acts as a cache key.
```
yungsters, line 103: It might be helpful to call out that `fetchTemperature` is not called until the first time that `getTemperature` is called.

---

> The `city` acts as a cache key.

Maybe before this sentence, explicitly mention that…

> If `AnimatedWeatherCard` and `MinimalWeatherCard` supply different `city` arguments to `getTemperature`, then `fetchTemperature` will be called twice and each call site will receive different data.

src/content/reference/react/cache.md
lunaleaps: Missed this, yes agreed. I do kind of gloss over it. I'll add a deep dive section about this

src/content/reference/react/cache.md
```diff
@@ -0,0 +1,443 @@
...
+The optimization of caching return values based on inputs is known as [_memoization_](https://en.wikipedia.org/wiki/Memoization). We refer to `cachedFn` as a memoized function.
+
+</Note>
+
+#### Caveats {/*caveats*/}
+
+[//]: # 'TODO: add links to Server/Client Component reference once https://github.com/reactjs/react.dev/pull/6177 is merged'
+
+- `cache` is recommended for Server Components only. There are plans to introduce `cache` for Client Components, but it is not recommended today.
+- React will invalidate the cache for a memoized function across server requests.
```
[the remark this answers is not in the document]
lunaleaps: all, updated

src/content/reference/react/cache.md
```diff
@@ -0,0 +1,443 @@
...
+
+export default async function AnimatedWeatherCard({city}) {
+  const temperature = await getTemperature(city);
+  // ...
+}
+```
+
+If you want to optimize cache hits across components, you'll need to define the memoized function in a scope that is accessible across multiple components.
+
+In the above example, each component <CodeStep step={1}>creates a new memoized version of `fetchTemperature`</CodeStep> that is only defined locally. This narrows the access to the memoized function to only the component in the module.
```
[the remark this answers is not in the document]
lunaleaps: I'll update this to say "only the module".

src/content/reference/react/cache.md
lunaleaps: Yea `memo` does, but they do also define memoization https://react.dev/reference/react/memo#skipping-re-rendering-when-props-are-unchanged. This is how `useMemo` also does it: https://react.dev/reference/react/useMemo#caveats

I think the lack of "memoize" in the early section, before we define it, feels sufficient, is there something that doesn't to you?

lunaleaps: Updated to use similar wording to `createContext`

src/content/reference/react/cache.md
lunaleaps: Updated!

src/content/reference/react/cache.md
lunaleaps: Hmm.. I don't think there are any limitations here. In the examples below, a lot of what we `cache` are functions that return promises

src/content/reference/react/cache.md
lunaleaps: No, this is correct. The memoized function is `cachedFn`.

src/content/reference/react/cache.md
lunaleaps: We are actively discouraging it. We are aiming to add a warning when using `cache` on the client. How about
```suggestion
- `cache` is recommended for Server Components only. There are plans to introduce `cache` for Client Components, but it is not a supported feature. Today, `cache` will no-op and warn if used in a Client Component.
```

That last bit I'd remove since we don't actually warn (yet) and add it in when we do. cc @acdlite

src/content/reference/react/cache.md
lunaleaps: Updated, and also mention it in the "Pitfall" section that we also talk about this

src/content/reference/react/cache.md
lunaleaps: I guess I assumed that separate modules, and importing the module, is eventually global scope. I can try and clarify this

src/content/reference/react/cache.md
lunaleaps: Re: second point, added!

src/content/reference/react/cache.md
lunaleaps: Clarified that this is the same error re-thrown

src/content/reference/react/cache.md
```diff
@@ -0,0 +1,485 @@
...
+  const getWeekReport = cache(calculateWeekReport);
+	const report = getWeekReport(cityData);
+  // ...
+}
+
+// Precipitation.js
+import {cache} from 'react';
+import {calculateWeekReport} from './report';
+
+// 🚩 Wrong: `getWeekReport` is only accessible for `Precipitation` component.
```
mattcarrollcode, line 126: couldn't the code for `Precipitation.js` be correct depending on the relationship between the Temperature component and `Precipitation.js`? e.g. if the Precipitation component was the parent component of the Temperature component the code for `Precipitation.js` might be correct as is, as long as it made the `getWeekReport` function created in `Precipitation.js` is made accessible to the Temperature component.

src/content/reference/react/cache.md
```diff
@@ -0,0 +1,485 @@
...
+}
+```
+
+If the same `user` object is rendered in both `Profile` and `TeamReport`, the two components can share work and only call `calculateUserMetrics` once for that `user`. 
+
+Assume `Profile` is rendered first. It will call <CodeStep step={1}>`getUserMetrics`</CodeStep>, and check if there is a cached result. Since it is the first time `getUserMetrics` is called with that `user`, there will be a cache miss. `getUserMetrics` will then call `calculateUserMetrics` with that `user` and write the result to cache. 
+
+When `TeamReport` renders its list of `users` and reaches the same `user` object, it will call <CodeStep step={2}>`getUserMetrics`</CodeStep> and read the result from cache.
+
+<Pitfall>
```
mattcarrollcode, line 107: It feels like this is calling out two pitfalls at the same time:

1. calling cache inside a component will recreate the cached function (and the cache) on every render
2. you need to share the cached function across components or the cached won't be shared

I found it a bit difficult to follow. Have you considered factoring 1 out of this pitfall into a troubleshooting section? e.g. a title for a troubleshooting section based on 1 might be somethign like "My memoized function isn't caching my data fetch/computation"

src/content/reference/react/cache.md
```diff
@@ -0,0 +1,485 @@
...
+
+export default function Precipitation({cityData}) {
+  const report = getWeekReport(cityData);
+  // ...
+}
+```
+Here, both components call the <CodeStep step={3}>same memoized function</CodeStep> exported from `./getWeekReport.js` to read and write to the same cache. 
+</Pitfall>
+
+### Share a snapshot of data {/*take-and-share-snapshot-of-data*/}
```
mattcarrollcode, line 167: This section is very clear and concise!

src/content/reference/react/cache.md
```diff
@@ -0,0 +1,485 @@
...
+
+Notice that the first <CodeStep step={2}>`getData`</CodeStep> call does not `await` whereas the <CodeStep step={3}>second</CodeStep> does. [`await`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await) is a JavaScript operator that will wait and return the settled result of the promise. The first <CodeStep step={2}>`getData`</CodeStep> call simply initiates the `fetch` to cache the promise for the second <CodeStep step={3}>`getData`</CodeStep> to look-up.
+
+If by the <CodeStep step={3}>second call</CodeStep> the promise is still _pending_, then `await` will pause for the result. The optimization is that while we wait on the `fetch`, React can continue with computational work, thus reducing the wait time for the <CodeStep step={3}>second call</CodeStep>. 
+
+If the promise is already settled, either to an error or the _fulfilled_ result, `await` will return that value immediately. In both outcomes, there is a performance benefit.
+</DeepDive>
+
+<Pitfall>
```
mattcarrollcode, line 280: Have you considered moving this pitfall into a troubleshooting section? e.g. this could be merged into the troubleshooting section I proposed in a earlier comment with the title "My memoized function isn't caching my data fetch/computation

src/content/reference/react/cache.md
```diff
@@ -0,0 +1,485 @@
...
+  const record = getRecord();
+  return (
+    <>
+      <WeatherReport record={record} />
+      <WeatherReport record={record} />
+    </>
+  );
+}
+```
+In this example, `App` renders two `WeatherReport`s with the same record. Even though both components do the same work, they cannot share work. `useMemo` is only a local cache to the component instance.
```
mattcarrollcode, line 335: > `useMemo` is only a local cache to the component instance. 

nit: this wording is a little awkward, maybe try something like "`useMemo`'s cache is local to the component"?

src/content/reference/react/cache.md
```diff
@@ -0,0 +1,485 @@
...
+}
+```
+
+React only provides cache access to the memoized function in a component. When calling <CodeStep step={1}>`getUser`</CodeStep> outside of a component, it will still evaluate the function but not read or update the cache.
+
+This is because cache access is provided through a [context](/learn/passing-data-deeply-with-context) which is only accessibile from a component. 
+
+</Pitfall>
+
+<DeepDive>
```
mattcarrollcode, line 307: I love how you evolve a single example to show how to use all 3 APIs!

src/content/reference/react/cache.md
```diff
@@ -0,0 +1,485 @@
...
+
+</DeepDive>
+
+---
+
+## Troubleshooting {/*troubleshooting*/}
+
+### My memoized function still runs even though I've called it with the same arguments {/*memoized-function-still-runs*/}
+
+Beyond the mentioned Pitfalls, it may be a problem with how React checks if something exists in cache.
```
mattcarrollcode, line 402: Could you link to the relevant Pitfalls with descriptions of when to follow their advice or remove this sentence? It feels a bit strange to mention there might be relevant answers to someone who finds this section but not to at least help them find the relevant info.

src/content/reference/react/cache.md
```diff
@@ -0,0 +1,485 @@
...
+
+When calling a memoized function, React will look up the input arguments to see if a result is already cached. React will use shallow equality of the arguments to determine if there is a cache hit.
+
+This means that if your arguments are not [primatives](https://developer.mozilla.org/en-US/docs/Glossary/Primitive) (ex. objects, functions, arrays), ensure you're passing the same object reference.
```
mattcarrollcode, line 406: >if your arguments are not [primatives](https://developer.mozilla.org/en-US/docs/Glossary/Primitive) (ex. objects, functions, arrays), ensure you're passing the same object reference.

Could you bubble this to the top of the section? I think many folks will know exactly what to do after reading this sentence so it'd be nice to highlight it to folks that are skimming for an answer.

mattcarrollcode (commented):
Looking really good! I have a few comments/suggestions/questions but they are all relatively minor.

src/content/reference/react/cache.md
```diff
@@ -0,0 +1,485 @@
...
+
+const getMetrics = cache(calculateMetrics);
+
+function Chart({data}) {
+  const report = getMetrics(data);
+  // ...
+}
+```
+
+`getMetrics` will first check if there is a cached result for `data`, otherwise call `calculateMetrics(data)` and cache the returned result.
```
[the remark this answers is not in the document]
lunaleaps: Updated to 

```
When `getMetrics` is first called with `data`, `getMetrics` will call `calculateMetrics(data)` and store the result in cache. If `getMetrics` is called again with the same `data`, it will return the cached result instead of calling `calculateMetrics(data)` again.
```

src/content/reference/react/cache.md
```diff
@@ -0,0 +1,485 @@
...
+</Note>
+
+#### Caveats {/*caveats*/}
+
+[//]: # 'TODO: add links to Server/Client Component reference once https://github.com/reactjs/react.dev/pull/6177 is merged'
+
+- React will invalidate the cache for all memoized functions for each server request. 
+- Each call to `cache` creates a new function. This means that calling `cache` with the same function multiple times will return different memoized functions that do not share the same cache.
+- `cachedFn` will also cache errors. If `fn` throws an error for certain arguments, it will be cached, and the same error is re-thrown when `cachedFn` is called with those same arguments.
+- `cache` is for use in [Server Components](https://react.dev/blog/2023/03/22/react-labs-what-we-have-been-working-on-march-2023#react-server-components) only. There are plans to introduce `cache` for Client Components, but it is not a supported feature today. There are plans to add a warning when `cache` is used in Client Components.
```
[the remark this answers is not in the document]
lunaleaps: Removed

src/content/reference/react/cache.md
```diff
@@ -0,0 +1,485 @@
...
+
+The optimization of caching return values based on inputs is known as [_memoization_](https://en.wikipedia.org/wiki/Memoization). We refer to the function returned from `cache` as a memoized function.
+
+</Note>
+
+#### Caveats {/*caveats*/}
+
+[//]: # 'TODO: add links to Server/Client Component reference once https://github.com/reactjs/react.dev/pull/6177 is merged'
+
+- React will invalidate the cache for all memoized functions for each server request.
```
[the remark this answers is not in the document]
mattcarrollcode: Luna and I spoke and validated that a cached function's cache is invalidated after every request even on a long lived node/dev server. We decided that situations like using cache in a server component when compiling to a static app that is deployed to a CDN is more of a edge case and doesn't need to be addressed in this section. Resolving.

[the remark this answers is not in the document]
lunaleaps: Chatted offline, I think for now we will keep it as "server request". It's not ideal but we can iterate on it as we get more documentation out.

src/content/reference/react/cache.md
```diff
@@ -0,0 +1,485 @@
...
+
+<Pitfall>
+Calling different memoized functions will read from different caches. To access the same cache, components must call the same memoized function.
+
+```js [[1, 7, "getWeekReport = cache(calculateWeekReport)"], [1, 8, "getWeekReport(cityData)"], [2, 17, "getWeekReport = cache(calculateWeekReport)"], [2, 20, "getWeekReport(cityData)"]]
+// Temperature.js
+import {cache} from 'react';
+import {calculateWeekReport} from './report';
+
+export function Temperature({cityData}) {
```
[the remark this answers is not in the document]
lunaleaps: Ah yes, I did this because I want `Temperature` and `Precipitation` to refer to their own memoized function calls -- not necessarily to the component.

src/content/reference/react/cache.md
lunaleaps: Hmm.. in a way, but the way this example is written doesn't demonstrate that use-case. 

Like yes you can re-write this example to be an example that does demonstrate cache-sharing but in its current implementation it doesn't. It is implicit that these components are rendered as siblings, I could make that clearer with an `App` component.

mattcarrollcode (approved):

src/content/reference/react/cache.md
lunaleaps: I will keep it for now. I had tried separating it out in a previous iteration, I think it won't be a major footgun as long as the larger pitfall is addressed. We can always update it if we see evidence of this being a common pattern

src/content/reference/react/cache.md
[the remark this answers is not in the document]
lunaleaps: I'll keep this for now, I've separated the example into separate code blocks to make the separation for components easier. The ideal as we discussed is to use Sandbox to capture this. This requires a bit more work and we can re-visit when we get those changes in

src/content/reference/react/cache.md
lunaleaps: I've added linking to the pitfall sections and now mention them in the troubleshooting section

PR #6248, opened 2023-08-23
add: adding documentation for custom hook useForm

nvbale:
The idea for this PR is to add in the react documentation a custom hook that is too much used by the community.


sophiebits:
Thanks for the suggestion but the documentation is about the hooks that React itself provides.

PR #6253, opened 2023-08-23
Fix text in docs

smdmori:
## The question and it's solutions had a minor mismatch in text
image
image

<hr />

## useState is deleted, so we can delete it's import too
image
react dev

### DIFF
--- a/src/content/learn/removing-effect-dependencies.md
+++ b/src/content/learn/removing-effect-dependencies.md
@@ -926,7 +926,7 @@ function ChatRoom() {
 
   useEffect(() => {
     const options = createOptions();
-    const connection = createConnection();
+    const connection = createConnection(options);
     connection.connect();
     return () => connection.disconnect();
   }, []); // ✅ All dependencies declared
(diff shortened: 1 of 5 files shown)


src/content/learn/removing-effect-dependencies.md
smdmori, line 1: It should pass options object like the codes below and after it

src/content/learn/synchronizing-with-effects.md
smdmori, line 1: I think it's clearer in this way, maybe I'm wrong?

src/content/learn/state-as-a-snapshot.md
```diff
@@ -79,7 +79,7 @@ When React re-renders a component:
 
 1. React calls your function again.
 2. Your function returns a new JSX snapshot.
-3. React then updates the screen to match the snapshot you've returned.
+3. React then updates the screen to match the snapshot your function have returned.
```
sophiebits, line 82:
```suggestion
3. React then updates the screen to match the snapshot your function returned.
```

src/content/learn/synchronizing-with-effects.md
```diff
@@ -767,7 +767,7 @@ Buying is not caused by rendering; it's caused by a specific interaction. It sho
   }
 ```
 
-**This illustrates that if remounting breaks the logic of your application, this usually uncovers existing bugs.** From the user's perspective, visiting a page shouldn't be different from visiting it, clicking a link, and pressing Back. React verifies that your components abide by this principle by remounting them once in development.
+**This illustrates that if remounting breaks the logic of your application, this usually uncovers existing bugs.** From the user's perspective, visiting a page shouldn't be different from re-visiting it (clicking a link on the page, and pressing Back). React verifies that your components abide by this principle by remounting them once in development.
```
sophiebits, line 770:
```suggestion
**This illustrates that if remounting breaks the logic of your application, this usually uncovers existing bugs.** From a user's perspective, visiting a page shouldn't be different from visiting it, clicking a link, then pressing Back to view the page again. React verifies that your components abide by this principle by remounting them once in development.
```

sophiebits (commented):
thanks for the fixes!

PR #6257, opened 2023-08-25
Update synchronizing-with-effects.md

niranjanghule:
Add missing word 'error' which completes the sentence


sophiebits:
It's correct as written.

PR #6270, opened 2023-08-31
Adds examples of common typescript component patterns

BenJanecke:
Hey folks, I really appreciated the recent addition of Typescript to the documentation!

I felt that the following additions are highly likely t be beneficial as in my experience these specific signatures are something that I've seen either completely omitted because the team couldn't figure them out or implemented in with some wild workarounds in every single TS React codebase I've either directly worked on or acted as consultant for over the last handful of years.

I'm no technical writer so I'm super open to recommendations on adjusting the copy for clarity and style, but I'm rather adamant on trying to get some form of these patterns into the docs :)

### DIFF
--- a/src/content/learn/typescript.md
+++ b/src/content/learn/typescript.md
@@ -122,9 +122,91 @@ export default App = AppTSX;
 The type describing your component's props can be as simple or as complex as you need, though they should be an object type described with either a `type` or `interface`. You can learn about how TypeScript describes objects in [Object Types](https://www.typescriptlang.org/docs/handbook/2/objects.html) but you may also be interested in using [Union Types](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#union-types) to describe a prop that can be one of a few different types and the [Creating Types from Types](https://www.typescriptlang.org/docs/handbook/2/types-from-types.html) guide for more advanced use cases.
 
 
+### Common patterns for components in Typescript {/*common-patterns-for-components-in-typescript*/}
+
+#### Exposing the api of the component being decorated {/*exposing-the-api-of-the-component-being-decorated*/}
+
+Often when we create components that decorate other components we want to expose the api of the component being decorated.
+For example if we're making a button component that adds a customProp we would still like the consumer to expose the props of the underlying button component.
+
+
+```tsx App.tsx active
+import { ComponentPropsWithoutRef, ElementType } from "react";
+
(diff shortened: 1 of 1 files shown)


BenJanecke:
Hey folks @sophiebits @kassens  @lunaleaps@TheSavior @mattcarrollcode
Apologies for the broadcast to all of you, I'm not sure if there's an appropriate tag to get the teams attention?

This PR has been open for a while with no feedback, kindly let me know if there's anything I can change to get it closer to being merged, if there's any feedback you'd like to provide, or if I should close it?

lunaleaps:
Thanks for the contribution! I think my overall sense is that these usage examples are helpful but it might help to have them listed after the "Useful Types" section. It feels like the sections "Example Hooks" and "Useful Types" are more oriented towards newcomers to TS and the examples you mention are a little more involved. 

So I see the layout being:

* Overview
* Installation
* TypeScript with React Components
* Example Hooks
* Useful Types
* Common Patterns <-- what you're adding
* Further Learning

Other feedback I'd give is to provide more concrete usage examples. 

For example in ["Decorating components"](https://react-9bxij8ggk-fbopensource.vercel.app/learn/typescript#exposing-the-api-of-the-component-being-decorated), you do mention a small example but anchoring a bit more in the use case of why you're doing this, in a succinct way would be helpful for newcomers to the concept. Also I wonder if the term "decoration" is best here. I can also see this being called higher order components. I can see good links out to Wikipedia/MDN about decoration as a CS concept but also mentioning what this may be known as within the community may be helpful. 

[Typesafe ForwardRef](https://react-9bxij8ggk-fbopensource.vercel.app/learn/typescript#type-safe-forwardref)
Similar to above, fleshing out a use-case for a forwardedRef (a bit more will be helpful. Also linking to forwardRef docs. 

For [Memoize and forwardref the same component](https://react-9bxij8ggk-fbopensource.vercel.app/learn/typescript#memoize-and-forwardref-the-same-component-while-maintaining-type-saftey), it's not clear to me the problem / difficulty that you're trying to address is. Can you expand that use-case? I want to understand how it's a common use-case people will run into. 

Other general thoughts
* Simplify the language around concepts. The docs aim to be understandable to beginners to CS. Terms like "inference" or "generic arguments" should be linked out to deeper reading and only used if they're necessary for the usage. 
* Let's turn these examples into CodeSandboxes. Also I think we should separate out the type imports in the import statements.

lunaleaps (changes requested):
See overall comment

PR #6279, opened 2023-09-04
Update HomeContent.js

hvlxh:
Added missing bracket in the Example2 component


smikitky:
The original code is correct. The parentheses correctly wrap the `video => ...` function, not the `<Video />` tag. The `<Video />` tag doesn't need to be wrapped by anything here.
```
{videos.map(video =>
  <Video key={video.id} video={video} />
)}
```

PR #6285, opened 2023-09-06
Update lazy.md to mention default exports

sophiebits:
React.lazy dots into the `default` property. Document this.


justjake (commented):
Thanks!

lunaleaps (approved):

PR #6286, opened 2023-09-07
fix links

awxiaoxian2020:


sophiebits:
Correct me if I'm wrong but I don't believe there was an issue with the links as they are?

awxiaoxian2020:
@sophiebits  We use the "internal link" without domains, which helps consistence in translated sites, and reduce the cost of http request.(maybe)

sophiebits:
The point about translated sites is a good one. Thanks.

PR #6296, opened 2023-09-12
Update tutorial-tic-tac-toe.md added import statement for react

mostakiiiim:
I encountered an error in my React application that said: "Line 2: 'React' must be in scope when using JSX."

To resolve this issue, I made sure to import React at the beginning of my JavaScript file, along with StrictMode.


sophiebits:
When using the latest JSX runtime, you do not need to import React.

https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html

If you are only seeing the lint error but the code works otherwise, you can disable it: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/react-in-jsx-scope.md#when-not-to-use-it

PR #6301, opened 2023-09-15
Update 'optimizing re-renders' section

SonDo580:
Just adding some minor changes. I hope it can help.
I see that the doc already mention skipping re-rendering in 'memo', 'useCallback', 'useMemo'. But in this section of 'useContext', the explanation stops at memoizing the context value and say that the children will not re-render, without warning the developer that if a parent component re-renders, its children will be re-render even if props and context value are the same.
So I think we should mention the use of 'memo' here. I know, experienced developers can clearly see the problem. But for a novice developer starting to learn React, it can be a bit misleading.
If there are anything I should improve, please let me know. Thanks.

### DIFF
--- a/src/content/reference/react/useContext.md
+++ b/src/content/reference/react/useContext.md
@@ -1339,6 +1339,61 @@ function MyApp() {
 
 As a result of this change, even if `MyApp` needs to re-render, the components calling `useContext(AuthContext)` won't need to re-render unless `currentUser` has changed.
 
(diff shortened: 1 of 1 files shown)


sophiebits:
Thanks for the PR but this section of the documentation is not about how to prevent Page from rerendering but rather the components that refer to AuthContext. I feel it is confusing to incorporate this guidance about memo right here especially since it is also not specific to useContext or context in any way.

PR #6308, opened 2023-09-21
Add first version appearance for API parts

mkarajohn:
This PR introduces a new "Added in <version_number>" section for each part of the API that is currently included in the stable versions.

Only part that gave me some trouble and could not exactly pinpoint when it was added was `flushSync` which must have been released somewhere during v16 according to [this](https://github.com/facebook/react/issues/11527#issuecomment-360199710) but it only got officially documented in the v18 docs.

If someone knows the exact version this was added I will gladly add it.

image


mkarajohn:
...any thoughts on this?

sophiebits:
Thoughts @mattcarrollcode? I don't think this is quite the right design but the rough idea seems good. There might also be certain parts of the API that were added later. I wonder if something more like the MDN browser compatibility table later in the page would make sense (perhaps comparing RSC/SSR/Client…? not sure).

src/content/reference/react-dom/flushSync.md
```diff
@@ -2,6 +2,10 @@
 title: flushSync
 ---
 
+<AddedInVersion version="16" />
+
+[//]: # (According to the linked answer from Dan Abramov, this was supposed to be released during v16, but is only officially documented in v18.0.0 --> https://github.com/facebook/react/issues/11527#issuecomment-360199710)
```
A7med3bdulBaset, line 7: Remove this and add as a review comment or comment it in mdx

```suggestion
{/*(According to the linked answer from Dan Abramov, this was supposed to be released during v16, but is only officially documented in v18.0.0 --> https://github.com/facebook/react/issues/11527#issuecomment-360199710)*/}
```

src/content/reference/react/Children.md
```diff
@@ -2,6 +2,10 @@
 title: Children
 ---
 
+<AddedInVersion version="0.10.0" />
+
+[//]: # (Does not show up in changelogs but only appears in v0.10.0 docs based on https://web.archive.org/ research)
```
A7med3bdulBaset, line 7:
```suggestion
{/*(Does not show up in changelogs but only appears in v0.10.0 docs based on https://web.archive.org/ research)*/}
```

mkarajohn: Thanks. I just `ctrl + /` in my editor and it added the syntax for what it thought was an mdx comment

src/content/reference/react-dom/flushSync.md
mkarajohn: Resolved in 2fbf483152fd0ca43fc4d399d319f295841d91b5

mkarajohn:
BTW, this is one example of why I believe this PR has value. People not being sure since when a particular feature exists, falsely leading them to believe they can't use it 
image

Also, not trying to be annoying here, but is this PR something you consider giving feedback or merging at some point? 

Thanks

PR #6312, opened 2023-09-22
Add <form> Component reference documentation

mattcarrollcode:
Add initial reference documentation for the `<form>` Component

[Preview](https://react-dev-git-form-ref-docs-fbopensource.vercel.app/reference/react-dom/components/form)

### DIFF
--- a/src/content/reference/react-dom/components/form.md
+++ b/src/content/reference/react-dom/components/form.md
@@ -31,7 +31,7 @@ The [built-in browser `<form>` component](https://developer.mozilla.org/en-US/do
 
 ### `<form>` {/*form*/}
 
-To create interactive controls for submitting information, render the [built-in browser `<form>` component](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form). 🚧
+To create interactive controls for submitting information, render the [built-in browser `<form>` component](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form).
 
 ```js
 <form action={search}>
@@ -42,81 +42,21 @@ To create interactive controls for submitting information, render the [built-in 
 
 [See more examples below.](#usage)
 
-#### 🚧 Props {/*props*/}
+#### Props {/*props*/}
 
 `<form>` supports all [common element props.](/reference/react-dom/components/common#props)
 
-Additionally, `<form>` supports passing a function directly to the `action` and `formAction` props to handle form submission.
+* [`action`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#action): a URL or function. When a URL is passed to `action` the form will behave like the HTML form component. When a function is passed to `action` the function will handle the form submission. The function passed to `action` may be async and will be called with a single argument containing the [form data](https://developer.mozilla.org/en-US/docs/Web/API/FormData) of the submitted form. The `action` prop can be overridden by a `formAction` attribute on a `<button>`, `<input type="submit">`, or `<input type="image">` component.
 
-* [`action`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/option#disabled): A boolean. Specifies what to do when a form is submitted. The form data can be handled with client or server side code.
-* [`<button>`]
-    * [`formAction`] prop that overrides `<form>`'s `action` prop
-    * [`formAction`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#formaction): A string. Overrides the parent `<form action>` for `type="submit"` and `type="image"`.
+#### Caveats {/*caveats*/}
 
-
-
-#### 🚧 Caveats {/*caveats*/}
-
-* 🚧
+* When a function is passed to `action` or `formAction` the HTTP method will be POST regaudless of value of the `method` prop.
 
 ---
 
 ## Usage {/*usage*/}
 
-### 🚧 Submitting forms to the server {/*progressively-enhanching-forms*/}
-
-Render a `<form>` with a input and submit button. Pass a function marked with `'use server';` to the `action` prop of form to run the function when the form is submitted.
-
-<Sandpack>
-
-```js App.js
-'use client'
-
-import { search } from './actions.js'
-
-export default function Search() {
-  return (
-    <form action={search}>
-        <input name="query" />
-        <button type="submit">Search</button>
-    </form>
-
-  );
-}
-```
-
-```js actions.js
-'use server'
-
-import { queryDb } from './lib.js';
-
-export async function search(formData) {
-    const query = formData.get('query')
-    const results = queryDb(query)
-}
-```
-```js lib.js hidden
-export function queryDb(query) {
-    return "hi"
-}
-```
-
-```json package.json hidden
-{
-  "dependencies": {
-    "react": "experimental",
-    "react-dom": "experimental",
-    "react-scripts": "^5.0.0"
-  },
-  "main": "/index.js",
-  "devDependencies": {}
-}
-```
-
-</Sandpack>
-
-
-### Handling form submissions on the client {/*handle-form-submission-on-the-client*/}
+### Handling form submission without a server {/*handle-form-submission-on-the-client*/}
 
 Render a `<form>` with a input and submit button. Pass a function to the `action` prop of form to run the function when the form is submitted.
 
@@ -140,8 +80,8 @@ export default function Search() {
 ```json package.json hidden
 {
   "dependencies": {
-    "react": "experimental",
-    "react-dom": "experimental",
+    "react": "18.3.0-canary-6db7f4209-20231021",
+    "react-dom": "18.3.0-canary-6db7f4209-20231021",
     "react-scripts": "^5.0.0"
   },
   "main": "/index.js",
@@ -151,16 +91,60 @@ export default function Search() {
 
 </Sandpack>
 
+### Submitting forms without JavaScript {/*submitting-forms-without-javascript*/}
+
+Render a `<form>` with a input and submit button. Pass a function marked with `'use server'` to the `action` prop of form to run the function when the form is submitted. When used with [Server Components](/reference/react/use-client), the form is [progressively enhanced](https://developer.mozilla.org/en-US/docs/Glossary/Progressive_Enhancement).
+
+```jsx
+import { updateCart } from './lib.js';
+
+function AddToCart({productId}) {
+  async function addToCart(formData) {
+    'use server'
+    const productId = formData.get('productId')
+    await updateCart(productId)
+  }
+  return (
+    <form action={addToCart}>
+        <input type="hidden" name="productId" value={productId} />
+        <button type="submit">Add to Cart</button>
+    </form>
+
+  );
+}
+```
+
+In leiu, of using hidden form fields to provide data to the function passed to the `action` prop of `<form>` you can call the `bind` method on the function passed to the `action` prop of `<form>`.
+
+```jsx
+import { updateCart } from './lib.js';
+
+function AddToCart({productId}) {
+  async function addToCart(productId, formData) {
+    "use server";
+    await updateCart(productId)
+  }
+  const addProductToCart = addToCart.bind(null, productId);
+  return (
+    <form action={addProductToCart}>
+      <button type="submit">Add to Cart</button>
+    </form>
+  );
+}
+```
+
 ### Display a pending state during form submission {/*display-a-pending-state-during-form-submission*/}
-You can track the status of a form submission with the `useFormStatus` Hook. The `useFormStatus` Hook provides information on the form's state, like when the form is pending. The `pending` property indicates whether the form is currently in the process of being submitted. You can use this property to update the UI while the form is being submitted like disabling buttons or alterning wording.
+To display a pending state while a form is submitting, you can call the `useFormStatus` Hook in a component rendered in a `<form>` and read the `pending` property returned.
+
+Here, we use the `pending` property to indicate the form is submitting.
 
 <Sandpack>
 
 ```js App.js
-import { experimental_useFormStatus as useFormStatus } from "react-dom";
+import { useFormStatus } from "react-dom";
 import { submitForm } from "./actions.js";
 
-function Button() {
+function Submit() {
   const { pending } = useFormStatus();
   return (
     <button type="submit" disabled={pending}>
@@ -172,7 +156,7 @@ function Button() {
 function Form({ action }) {
   return (
     <form action={action}>
-      <Button />
+      <Submit />
     </form>
   );
 }
@@ -191,41 +175,53 @@ export async function submitForm(query) {
 ```json package.json hidden
 {
   "dependencies": {
-    "react": "experimental",
-    "react-dom": "experimental",
+    "react": "canary",
+    "react-dom": "canary",
     "react-scripts": "^5.0.0"
   },
   "main": "/index.js",
   "devDependencies": {}
 }
 ```
-</Sandpack>  
+</Sandpack>
 
 <Pitfall>
 
-The `useFormStatus` Hook must be called from a Component that is located inside `<form>`. You cannot call `useFormStatus` from the same Component that returns `<form>`.
+##### `useFormStatus` will not return status information for a `<form>` rendered in the same component. {/*useformstatus-will-not-return-status-information-for-a-form-rendered-in-the-same-component*/}
+
+The `useFormStatus` Hook only returns status information for a parent `<form>` and not for any `<form>` rendered in the same component calling the Hook, or child components.
 
 ```js
 function Form() {
-  const { pending } = useFormStatus() // 🚩 `pending` will never be `true`
-  return <form action={submit}> //...
+  // 🚩 `pending` will never be true
+  // useFormStatus does not track the form rendered in this component
+  const { pending } = useFormStatus();
+  return <form action={submit}></form>;
 }
 ```
 
-Instead call `useFormStatus` from inside a Component that is located inside `<form>`.
+Instead call `useFormStatus` from inside a component that is located inside `<form>`.
 
 ```js
-function Button(){
-    const { pending } = useFormStatus() // ✅ `pending` will be derived from the parent form state
-    return <button disabled={pending}> //...
+function Submit() {
+  // ✅ `pending` will be derived from the form that wraps the Submit component
+  const { pending } = useFormStatus();
+  return <button disabled={pending}>...</button>;
 }
 
 function Form() {
-  return <form> <Button /> //...
+  // This is the <form> `useFormStatus` tracks
+  return (
+    <form action={submit}>
+      <Submit />
+    </form>
+  );
 }
 ```
 
 </Pitfall>
+
+To learn more about the `useFormStatus` Hook see the [reference documentation](/reference/react-dom/hooks/useFormStatus).
 
 ### Optimistically updating form data {/*optimistically-updating-form-data*/}
 The `useOptimistic` Hook provides a way to optimistically update the user interface before a background operation, like a network request, completes. In the context of forms, this technique helps to make apps feel more responsive. When a user submits a form, instead of waiting for the server's response to reflect the changes, the interface is immediately updated with the expected outcome. 
@@ -236,54 +232,48 @@ For example, when a user types a message into the form and hits the "Send" butto
 
 
 ```js App.js
-import { experimental_useOptimistic as useOptimistic, useState } from "react";
+import { useOptimistic, useState } from "react";
 import { deliverMessage } from "./actions.js";
 
 function Thread({ messages, sendMessage }) {
+  async function formAction(formData) {
+    addOptimisticMessage(formData.get("message"));
+    await sendMessage(formData);
+  }
   const [optimisticMessages, addOptimisticMessage] = useOptimistic(
     messages,
     (state, newMessage) => [
       ...state,
       {
-        message: newMessage,
+        text: newMessage,
         sending: true
       }
     ]
   );
 
   return (
-    <div>
-      {optimisticMessages.map((m, i) => (
-        <div key={i}>
-          {m.message}
-          <small>{m.sending ? " (Sending...)" : ""}</small>
+    <>
+      {optimisticMessages.map((message, index) => (
+        <div key={index}>
+          {message.text}
+          {!!message.sending && <small> (Sending...)</small>}
         </div>
       ))}
-      <form
-        action={sendMessage}
-        onSubmit={async (e) => {
-          e.preventDefault();
-          const form = e.currentTarget;
-          const formData = new FormData(form);
-          addOptimisticMessage(formData.get("message"));
-          form.reset();
-          await sendMessage(formData);
-        }}
-      >
+      <form action={formAction}>
         <input type="text" name="message" placeholder="Hello!" />
         <button type="submit">Send</button>
       </form>
-    </div>
+    </>
   );
 }
 
 export default function App() {
   const [messages, setMessages] = useState([
-    { message: "Hello there!", sending: false, key: 1 }
+    { text: "Hello there!", sending: false, key: 1 }
   ]);
   async function sendMessage(formData) {
     const sentMessage = await deliverMessage(formData.get("message"));
-    setMessages([...messages, { message: sentMessage }]);
+    setMessages([...messages, { text: sentMessage }]);
   }
   return <Thread messages={messages} sendMessage={sendMessage} />;
 }
@@ -300,8 +290,8 @@ export async function deliverMessage(message) {
 ```json package.json hidden
 {
   "dependencies": {
-    "react": "experimental",
-    "react-dom": "experimental",
+    "react": "18.3.0-canary-6db7f4209-20231021",
+    "react-dom": "18.3.0-canary-6db7f4209-20231021",
     "react-scripts": "^5.0.0"
   },
   "main": "/index.js",
@@ -313,7 +303,7 @@ export async function deliverMessage(message) {
 
 ### Handling form submission errors {/*handling-form-submission-errors*/}
 
-In some cases the function called by a `<form>`'s `action` prop throw an error. You can handle these errors by wrapping `<form>` in an Error Boundary. If function called by a `<form>`'s `action` prop throws an error the fallback for the error boundary will be displayed.
+In some cases the function called by a `<form>`'s `action` prop throw an error. You can handle these errors by wrapping `<form>` in an Error Boundary. If the function called by a `<form>`'s `action` prop throws an error, the fallback for the error boundary will be displayed.
 
 <Sandpack>
 
@@ -343,8 +333,8 @@ export async function search() {
 ```json package.json hidden
 {
   "dependencies": {
-    "react": "experimental",
-    "react-dom": "experimental",
+    "react": "18.3.0-canary-6db7f4209-20231021",
+    "react-dom": "18.3.0-canary-6db7f4209-20231021",
     "react-scripts": "^5.0.0",
     "react-error-boundary": "4.0.3"
   },
@@ -353,54 +343,65 @@ export async function search() {
 }
 ```
 
-To surface specific error messages to your users from a form action you can use the `useFormState` Hook. `useFormState` takes two parameters: an action and a inital state, and returns two values, a state varible and a new action. When you use the action returned by `useFormState` it will call the action passed to `useFormState` with the current state as an argument. The value that the actions passed to `useFormState` returns will be used as the new value for the state var.
+</Sandpack>
+
+To surface error messages to your users from a form action you can use the `useFormState` Hook. `useFormState` takes two parameters: an action and a inital state, and returns two values, a state varible and a form action. The form action is passed to the `action` prop of the form. The value returned by the action passed to `useFormState` is used to update the state varible. This can be used to return error message that can be displayed to a user. When use in conjunction with [Server Actions](/reference/react/use-server) and [Server Components](/reference/react/use-client), the error message is progressively enhanced.
 
 <Sandpack>
 
 ```js App.js
-import { useState } from "react";
 import { useFormState } from "react-dom";
+import { signUpNewUser } from "./api";
 
-let maxUsers = 1;
-
-export default function NewsletterSignUp() {
-  const [signups, setSignups] = useState(0);
-  async function addEmailToNewsletter(prevState) {
-    if (signups >= maxUsers) {
-      return "maximum users reached";
+export default function Page() {
+  async function signup(prevState, formData) {
+    "use server";
+    const email = formData.get("email");
+    try {
+      await signUpNewUser(email);
+      alert(`Added "${email}"`);
+    } catch (err) {
+      return err.toString();
     }
-    setSignups(signups + 1);
   }
-  const [errorMessage, formAction] = useFormState(addEmailToNewsletter, null);
+  const [message, formAction] = useFormState(signup, null);
   return (
-    <form action={formAction}>
-      <label>
-        Email:
-        <input type="text" />
-      </label>
-      <button disabled={!!errorMessage} type="submit">
-        Sign up
-      </button>
-      <p style={{ color: "red" }}>{errorMessage}</p>
-    </form>
+    <>
+      <h1>Signup for my newsletter</h1>
+      <p>Signup with the same email twice to see an error</p>
+      <form action={formAction} id="signup-form">
+        <label for="email">Email: </label>
+        <input name="email" id="email" placeholder="react@example.com" />
+        <button>Sign up</button>
+        {!!message && <p>{message}</p>}
+      </form>
+    </>
   );
+}
+```
+
+```js api.js hidden
+let emails = [];
+
+export async function signUpNewUser(newEmail) {
+  if (emails.includes(newEmail)) {
+    throw new Error("This email address has already been added");
+  }
+  emails.push(newEmail);
 }
 ```
 
 ```json package.json hidden
 {
   "dependencies": {
-    "react": "canary",
-    "react-dom": "canary",
+    "react": "18.3.0-canary-6db7f4209-20231021",
+    "react-dom": "18.3.0-canary-6db7f4209-20231021",
     "react-scripts": "^5.0.0"
   },
   "main": "/index.js",
   "devDependencies": {}
 }
 ```
-
-</Sandpack>
-
 
 </Sandpack>
 
@@ -412,20 +413,21 @@ Forms can be designed to handle multiple submission actions based on the button 
 
 ```js App.js
 export default function Search() {
-  function publish(formAction) {
-    const content = formAction.get("content");
-    const button = formAction.get("button");
+  function publish(formData) {
+    const content = formData.get("content");
+    const button = formData.get("button");
     alert(`'${content}' was published with the '${button}' button`);
   }
 
-  function save(formAction) {
-    const content = formAction.get("content");
+  function save(formData) {
+    const content = formData.get("content");
     alert(`Your draft of '${content}' has been saved!`);
   }
 
   return (
     <form action={publish}>
       <textarea name="content" rows={4} cols={40} />
+      <br />
       <button type="submit" name="button" value="submit">Publish</button>
       <button formAction={save}>Save draft</button>
     </form>
@@ -436,8 +438,8 @@ export default function Search() {
 ```json package.json hidden
 {
   "dependencies": {
-    "react": "experimental",
-    "react-dom": "experimental",
+    "react": "18.3.0-canary-6db7f4209-20231021",
+    "react-dom": "18.3.0-canary-6db7f4209-20231021",
     "react-scripts": "^5.0.0"
   },
   "main": "/index.js",
--- a/src/content/reference/react-dom/components/form.md
+++ b/src/content/reference/react-dom/components/form.md
@@ -224,7 +224,7 @@ function Form() {
 To learn more about the `useFormStatus` Hook see the [reference documentation](/reference/react-dom/hooks/useFormStatus).
 
 ### Optimistically updating form data {/*optimistically-updating-form-data*/}
-The `useOptimistic` Hook provides a way to optimistically update the user interface before a background operation, like a network request, completes. In the context of forms, this technique helps to make apps feel more responsive. When a user submits a form, instead of waiting for the server's response to reflect the changes, the interface is immediately updated with the expected outcome. 
+The `useOptimistic` Hook provides a way to optimistically update the user interface before a background operation, like a network request, completes. In the context of forms, this technique helps to make apps feel more responsive. When a user submits a form, instead of waiting for the server's response to reflect the changes, the interface is immediately updated with the expected outcome.
 
 For example, when a user types a message into the form and hits the "Send" button, the `useOptimistic` Hook allows the message to immediately appear in the list with a "Sending..." label, even before the message is actually sent to a server. This "optimistic" approach gives the impression of speed and responsiveness. The form then attempts to truly send the message in the background. Once the server confirms the message has been received, the "Sending..." label is removed.
 
@@ -238,6 +238,7 @@ import { deliverMessage } from "./actions.js";
 function Thread({ messages, sendMessage }) {
   async function formAction(formData) {
     addOptimisticMessage(formData.get("message"));
+    document.getElementById("message-form").reset();
     await sendMessage(formData);
   }
   const [optimisticMessages, addOptimisticMessage] = useOptimistic(
@@ -259,7 +260,7 @@ function Thread({ messages, sendMessage }) {
           {!!message.sending && <small> (Sending...)</small>}
         </div>
       ))}
-      <form action={formAction}>
+      <form action={formAction} id="message-form">
         <input type="text" name="message" placeholder="Hello!" />
         <button type="submit">Send</button>
       </form>
--- a/src/content/reference/react-dom/components/form.md
+++ b/src/content/reference/react-dom/components/form.md
@@ -58,7 +58,7 @@ To create interactive controls for submitting information, render the [built-in 
 
 ### Handle form submission on the client {/*handle-form-submission-on-the-client*/}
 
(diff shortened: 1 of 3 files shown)


src/content/reference/react-dom/components/form.md
```diff
@@ -0,0 +1,445 @@
+---
+title: "<form>"
+canary: true
+---
+
+<Canary>
+
+React's `form` Component is currently only available in React's canary and experimental channels but you can use the [standard HTML `form` tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form) in your React's stable channel. Learn more about [React's release channels here](/community/versioning-policy#all-release-channels).
```
TheSavior, line 8: This makes it sound like React provides a different form component, implying that you might need to import `form` from React. Maybe instead the wording should be something like "React's additions to the browser's `form` component are only in canary and experimental"

src/content/reference/react-dom/components/form.md
```diff
@@ -0,0 +1,445 @@
...
+`<form>` supports all [common element props.](/reference/react-dom/components/common#props)
+
+Additionally, `<form>` supports these props:
+
+* [`action`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/option#disabled): A boolean. Specifies what to do when a form is submitted. The form data can be handled with client or server side code.
+* 🚧
+
+
+🚧 Forms can contain one or more of the following tags:
+* [`<input>`]
```
TheSavior, line 56: Oof, it's hard to know whether to list everything that the browser supports here, or only the things that have special React behavior. How are you thinking about that?

src/content/reference/react-dom/components/form.md
```diff
@@ -0,0 +1,445 @@
...
+
+* 🚧
+
+---
+
+## Usage {/*usage*/}
+
+### 🚧 Submitting forms to the server {/*progressively-enhanching-forms*/}
+
+Render a `<form>` with a input and submit button. Pass a function marked with `'use server';` to the `action` prop of form to run the function when the form is submitted.
```
TheSavior, line 78: Hmm, this is hard to explain and probably is dependent on the use server / use client docs. For example, if you have a file with "use client" and within that file you have a function with "use server", that won't work, right?

src/content/reference/react-dom/components/form.md
```diff
@@ -0,0 +1,445 @@
...
+    <div>
+      {optimisticMessages.map((m, i) => (
+        <div key={i}>
+          {m.message}
+          <small>{m.sending ? " (Sending...)" : ""}</small>
+        </div>
+      ))}
+      <form
+        action={sendMessage}
+        onSubmit={async (e) => {
```
TheSavior, line 274: This function feels really clunky. Is this really the recommended way to do optimistic updates for forms?

src/content/reference/react-dom/components/form.md
```diff
@@ -0,0 +1,445 @@
...
+  "main": "/index.js",
+  "devDependencies": {}
+}
+```
+
+</Sandpack>
+
+### Handling form submission errors {/*handling-form-submission-errors*/}
+
+In some cases the function called by a `<form>`'s `action` prop throw an error. You can handle these errors by wrapping `<form>` in an Error Boundary. If function called by a `<form>`'s `action` prop throws an error the fallback for the error boundary will be displayed.
```
TheSavior, line 326: s/If function/If the function

src/content/reference/react-dom/components/form.md
```diff
@@ -0,0 +1,445 @@
...
+
+```js App.js
+export default function Search() {
+  function publish(formAction) {
+    const content = formAction.get("content");
+    const button = formAction.get("button");
+    alert(`'${content}' was published with the '${button}' button`);
+  }
+
+  function save(formAction) {
```
TheSavior, line 382: the arguments here are `formData` not `formAction`, right?

src/content/reference/react-dom/components/form.md
```diff
@@ -0,0 +1,445 @@
...
+    "react-scripts": "^5.0.0"
+  },
+  "main": "/index.js",
+  "devDependencies": {}
+}
+```
+
+</Sandpack>
+
+### Managing state inside a form {/*managing-state-inside-a-form*/}
```
TheSavior, line 411: Woah. This could use some more explanation. It isn't clear to me when you'd want to use this

TheSavior (commented):
This is super cool! I appreciate the examples. As this gets closer to shipping we'll want to figure out more user friendly and less code-centric examples of course. I'm sure you have a list of things you want to do and I jumped the gun on reviewing this anyways 😀

src/content/reference/react-dom/components/form.md
mattcarrollcode: Yeah, good note. I was thinking something similar. This copy definitely needs more nuance

src/content/reference/react-dom/components/form.md
mattcarrollcode: I was starting by blindly copying what other DOM component pages are doing like [input](https://react-kbffi0laq-fbopensource.vercel.app/reference/react-dom/components/input). I know this is different because its supported tags inside `form` instead of attributes of `form` but that's why this list is here, so I don't forget to address this. Its going to be a little tricky, we'll at least need to document submit buttons pretty well because of `formAction`. I haven't given it much thought yet overall though, this is mostly a placeholder for me to remember to address this.

src/content/reference/react-dom/components/form.md
mattcarrollcode: Correct.

I started writing the "use server" docs and wanted to have a better idea of how forms were going to look like before I started on "use server" but they are going to be very interconnected in places like this. This explination was copied from the "Handling form submissions on the client" (which I wrote before this, and is itself a bad placeholder) and it needs to be completely rewritten.

I put this usage example first because I think that's what people are going to be most interested in: getting their form data to a server to do something with, but this example is just a placeholder because we don't have nodebox working. I think I figured out how to get all the other examples for `form` working on the client without nodebox but this example definitely needs nodebox to be a "proper" sample.

src/content/reference/react-dom/components/form.md
mattcarrollcode: yeah, I'm not sure what a good use-case for this Hook is yet, either. I just wanted to document what I got working so far.

src/content/reference/react-dom/components/form.md
```diff
@@ -0,0 +1,445 @@
...
+
+[See more examples below.](#usage)
+
+#### 🚧 Props {/*props*/}
+
+`<form>` supports all [common element props.](/reference/react-dom/components/common#props)
+
+Additionally, `<form>` supports these props:
+
+* [`action`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/option#disabled): A boolean. Specifies what to do when a form is submitted. The form data can be handled with client or server side code.
```
kmiddleton14, line 51: Should this link to the [form action attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#attributes_for_form_submission) instead of the html option?

eps1lon:
FYI: 
- stable preview link for latest changes for branches on upstream (such as this branch): `react-dev-git-<branch-name>-fbopensource.vercel.app`
- stable preview link for latest changes for branches forks: `react-dev-git-fork-<username>-<branch-name>-fbopensource.vercel.app`

More details in https://vercel.com/docs/deployments/generated-urls and https://github.com/orgs/vercel/discussions/970#discussioncomment-3918729 for forks.

src/content/reference/react-dom/components/form.md
sophiebits: https://github.com/facebook/react/pull/27453 and https://github.com/facebook/react/pull/27454 fix it to be a noop and to warn when called out of a transition.

src/content/reference/react-dom/components/form.md
davidmccabe: My view is that we should only document the differences and just link to MDN for the standard stuff. Otherwise, the standard stuff drowns out the signal of what's different, while also not being a very good reference for the standard stuff because we're never going to go into as much detail as MDN does.

src/content/reference/react-dom/components/form.md
davidmccabe: The type is wrong, it should say 'a function' instead of 'a boolean'.

src/content/reference/react-dom/components/form.md
```diff
@@ -0,0 +1,458 @@
...
+
+* 🚧
+
+---
+
+## Usage {/*usage*/}
+
+### 🚧 Submitting forms to the server {/*progressively-enhanching-forms*/}
+
+Render a `<form>` with a input and submit button. Pass a function marked with `'use server';` to the `action` prop of form to run the function when the form is submitted.
```
davidmccabe, line 78: I don't feel we should have the semicolon  in 'use server' here when it's used inline in a sentence as it makes the sentence too hard to parse. Your code formatter will add the semicolon if you perchance don't know or remember to, so I don't think this will lead to frustration.

mattcarrollcode: Fixed!

src/content/reference/react-dom/components/form.md
mattcarrollcode: THanks, fixed.

src/content/reference/react-dom/components/form.md
mattcarrollcode: good catch, fixed.

src/content/reference/react-dom/components/form.md
mattcarrollcode: Thanks all! Fixed.

src/content/reference/react-dom/components/form.md
mattcarrollcode: Fixed

src/content/reference/react-dom/components/form.md
mattcarrollcode: I removed this list

src/content/reference/react-dom/components/form.md
mattcarrollcode: Thanks, fixed both issues

src/content/reference/react-dom/components/form.md
mattcarrollcode: Added a note on the relationship between server components and progressive enhancement

src/content/reference/react-dom/components/form.md
mattcarrollcode: Fixed, thanks @sophiebits your suggestion was very helpful!

src/content/reference/react-dom/components/form.md
```diff
@@ -0,0 +1,451 @@
...
+
+#### Props {/*props*/}
+
+`<form>` supports all [common element props.](/reference/react-dom/components/common#props)
+
+* [`action`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#action): a URL or function. When a URL is passed to `action` the form will behave like the HTML form component. When a function is passed to `action` the function will handle the form submission. The function passed to `action` may be async and will be called with a single argument containing the [form data](https://developer.mozilla.org/en-US/docs/Web/API/FormData) of the submitted form. The `action` prop can be overridden by a `formAction` attribute on a `<button>`, `<input type="submit">`, or `<input type="image">` component.
+
+#### Caveats {/*caveats*/}
+
+* When a function is passed to `action` or `formAction` the HTTP method will be POST regaudless of value of the `method` prop.
```
lunaleaps, line 53:
```suggestion
* When a function is passed to `action` or `formAction` the HTTP method will be POST regardless of value of the `method` prop.
```

src/content/reference/react-dom/components/input.md
```diff
@@ -32,6 +32,8 @@ To display an input, render the [built-in browser `<input>`](https://developer.m
 
 `<input>` supports all [common element props.](/reference/react-dom/components/common#props)
 
+* [`formAction`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#formaction): A string or function. Overrides the parent `<form action>` for `type="submit"` and `type="image"`. When a URL is passed to `action` the form will behave like a standard HTML form. When a function is passed to `formAction` the function will handle the form submission. See [`<form action>`](/reference/react-dom/components/form#props).
```
lunaleaps, line 35: Should we put this as a separate point (like not a sub-point under "supports all common element props" and call out that this is a React extension and only available for canary?

src/content/reference/react-dom/components/form.md
```diff
@@ -0,0 +1,451 @@
...
+</form>
+```
+
+[See more examples below.](#usage)
+
+#### Props {/*props*/}
+
+`<form>` supports all [common element props.](/reference/react-dom/components/common#props)
+
+* [`action`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#action): a URL or function. When a URL is passed to `action` the form will behave like the HTML form component. When a function is passed to `action` the function will handle the form submission. The function passed to `action` may be async and will be called with a single argument containing the [form data](https://developer.mozilla.org/en-US/docs/Web/API/FormData) of the submitted form. The `action` prop can be overridden by a `formAction` attribute on a `<button>`, `<input type="submit">`, or `<input type="image">` component.
```
lunaleaps, line 49: Same as mentioned for `input.md`, should we not nest this point under the title "supports all common element props" as if anything this is calling out the React extensions we're adding to `form`... like I guess my root issue is that I wish we were clearer that "here are the things that we've done that change HTML form behavior"

src/content/reference/react-dom/components/form.md
```diff
@@ -0,0 +1,451 @@
...
+
+#### Caveats {/*caveats*/}
+
+* When a function is passed to `action` or `formAction` the HTTP method will be POST regaudless of value of the `method` prop.
+
+---
+
+## Usage {/*usage*/}
+
+### Handling form submission without a server {/*handle-form-submission-on-the-client*/}
```
lunaleaps, line 59: Nit: How about "Client-only form submissions" or "Client forms"

src/content/reference/react-dom/components/form.md
```diff
@@ -0,0 +1,451 @@
...
+
+Render a `<form>` with a input and submit button. Pass a function to the `action` prop of form to run the function when the form is submitted.
```
lunaleaps, line 61:
```suggestion
Pass a function to the `action` prop of form to run the function when the form is submitted.
```
Should we also add that this won't call a server endpoint, which is traditional with HTML `action`?

src/content/reference/react-dom/components/form.md
```diff
@@ -0,0 +1,451 @@
...
+  "main": "/index.js",
+  "devDependencies": {}
+}
+```
+
+</Sandpack>
+
+### Submitting forms without JavaScript {/*submitting-forms-without-javascript*/}
+
+Render a `<form>` with a input and submit button. Pass a function marked with `'use server'` to the `action` prop of form to run the function when the form is submitted. When used with [Server Components](/reference/react/use-client), the form is [progressively enhanced](https://developer.mozilla.org/en-US/docs/Glossary/Progressive_Enhancement).
```
lunaleaps, line 96: Isn't this the default for forms already without React? Re: server endpoints, so it's more of saying that `action` doesn't regress this ability? Also, maybe briefly mention on why/when you'd want to support form submission without JS

src/content/reference/react-dom/components/form.md
```diff
@@ -0,0 +1,451 @@
...
+    <form action={addToCart}>
+        <input type="hidden" name="productId" value={productId} />
+        <button type="submit">Add to Cart</button>
+    </form>
+
+  );
+}
+```
+
+In leiu, of using hidden form fields to provide data to the function passed to the `action` prop of `<form>` you can call the `bind` method on the function passed to the `action` prop of `<form>`.
```
lunaleaps, line 117: Oh this took a while for me to understand. Maybe we could set the scene a bit more for the first example

src/content/reference/react-dom/components/form.md
```diff
@@ -0,0 +1,451 @@
...
+
+```jsx
```
lunaleaps, line 119:
```suggestion
```jsx {8}
```

src/content/reference/react-dom/components/form.md
```diff
@@ -0,0 +1,451 @@
...
+    "react-scripts": "^5.0.0"
+  },
+  "main": "/index.js",
+  "devDependencies": {}
+}
+```
+
+</Sandpack>
+
+### Submitting forms without JavaScript {/*submitting-forms-without-javascript*/}
```
lunaleaps, line 94: Playing off of client-only forms, what about adding an example of "Submitting forms with server actions", it could just be this example and we continue to expand about how `action` also supports progressive enhancement

src/content/reference/react-dom/components/form.md
```diff
@@ -0,0 +1,451 @@
...
+    "react-error-boundary": "4.0.3"
+  },
+  "main": "/index.js",
+  "devDependencies": {}
+}
+```
+
+</Sandpack>
+
+To surface error messages to your users from a form action you can use the `useFormState` Hook. `useFormState` takes two parameters: an action and a inital state, and returns two values, a state varible and a form action. The form action is passed to the `action` prop of the form. The value returned by the action passed to `useFormState` is used to update the state varible. This can be used to return error message that can be displayed to a user. When use in conjunction with [Server Actions](/reference/react/use-server) and [Server Components](/reference/react/use-client), the error message is progressively enhanced.
```
lunaleaps, line 349:
```suggestion
To surface error messages to your users from a form action you can use the `useFormState` Hook. `useFormState` takes two parameters: an action and a inital state, and returns two values, a state variable and a form action. The form action is passed to the `action` prop of the form. The value returned by the action passed to `useFormState` is used to update the state variable. This can be used to return error message that can be displayed to a user. When use in conjunction with [Server Actions](/reference/react/use-server) and [Server Components](/reference/react/use-client), the error message is progressively enhanced.
```

lunaleaps: I'm having a hard time following this..I also haven't read the `useFormState` docs yet so bear that in mind. 

I think first off the bit about "To surface error messages to your users from a form action", that doesn't feel clear to me why I need `useFormState`, couldn't I show an error message without the hook? 

Is there a mental model around `useFormState` we could share? I think the one I'm shaping in my mind reading this is that it allows you to receive a return value from your server action? But isn't that possible just from returning a value from the server function? But I guess you'd need a wrapper to read it.. so I guess that's what `useFormState` is for?

lunaleaps: Also, I think for this "`useFormState` takes two parameters: an action and a inital state", I'd suggestion saying function instead of action. Or maybe expanding on what action means here?

src/content/reference/react-dom/components/form.md
```diff
@@ -0,0 +1,451 @@
...
+    "react-scripts": "^5.0.0"
+  },
+  "main": "/index.js",
+  "devDependencies": {}
+}
+```
+
+</Sandpack>
+
+### Processing forms differently with different submit buttons {/*processing-forms-differently-with-different-submit-buttons*/}
```
lunaleaps, line 409: Nit: What about "Handling multiple submission types"?

src/content/reference/react-dom/components/form.md
```diff
@@ -0,0 +1,451 @@
...
+
+Forms can be designed to handle multiple submission actions based on the button pressed by the user. Each button can be associated with a distinct action or behavior. When a user taps a specific button, the form is submitted, and a corresponding action, defined by that button's attributes and action, is executed. For instance, a form might allow users to either save their input as a draft or submit it for review with separate buttons.
```
lunaleaps, line 411: Could we clarify or namedrop that `formAction` allows you to customize the submission action for multiple buttons?

src/content/reference/react-dom/components/form.md
```diff
@@ -0,0 +1,451 @@
...
+    "react-dom": "canary",
+    "react-scripts": "^5.0.0"
+  },
+  "main": "/index.js",
+  "devDependencies": {}
+}
+```
+</Sandpack>
+
+<Pitfall>
```
lunaleaps, line 188: Do we need this Pitfall here? It feels mostly relevant to `useFormStatus`. I imagine if someone was having issues with `useFormStatus`, they'd check the reference on the API

src/content/reference/react-dom/components/form.md
```diff
@@ -0,0 +1,451 @@
...
+  );
+}
+```
+
+</Pitfall>
+
+To learn more about the `useFormStatus` Hook see the [reference documentation](/reference/react-dom/hooks/useFormStatus).
+
+### Optimistically updating form data {/*optimistically-updating-form-data*/}
+The `useOptimistic` Hook provides a way to optimistically update the user interface before a background operation, like a network request, completes. In the context of forms, this technique helps to make apps feel more responsive. When a user submits a form, instead of waiting for the server's response to reflect the changes, the interface is immediately updated with the expected outcome.
```
lunaleaps, line 227: I like this section and how it lays out the use-case. Do we need to go into what happens if the server fails to apply the update?

src/content/reference/react-dom/components/form.md
```diff
@@ -0,0 +1,451 @@
...
+
+For example, when a user types a message into the form and hits the "Send" button, the `useOptimistic` Hook allows the message to immediately appear in the list with a "Sending..." label, even before the message is actually sent to a server. This "optimistic" approach gives the impression of speed and responsiveness. The form then attempts to truly send the message in the background. Once the server confirms the message has been received, the "Sending..." label is removed.
```
lunaleaps, line 229: I like this usage example and how it lays out the use-case to orient me into the mindset. Do we need to go into what happens if the server fails to apply the update?

src/content/reference/react-dom/components/form.md
lunaleaps, line 117:
```suggestion
In lieu, of using hidden form fields to provide data to the function passed to the `action` prop of `<form>` you can call the `bind` method on the function passed to the `action` prop of `<form>`.
```

src/content/reference/react-dom/components/input.md
mattcarrollcode: Do you think we need a callout here in addition to the page-level canary callout at the top of the page?

mattcarrollcode: Good catch, fixed.

mattcarrollcode:
In person code review notes with Luna:
* refer the funciton passed to the action prop as a "action"
* rephrase "form action" to a "the action passed to form"

src/content/reference/react-dom/components/form.md
```diff
@@ -0,0 +1,451 @@
...
+
+
+```js App.js
+import { useOptimistic, useState } from "react";
+import { deliverMessage } from "./actions.js";
+
+function Thread({ messages, sendMessage }) {
+  async function formAction(formData) {
+    addOptimisticMessage(formData.get("message"));
+    document.getElementById("message-form").reset();
```
mattcarrollcode, line 241: From @gaearon - we shouldn’t be ever recommending getElementById inside components. use refs for this. the problem is that it makes it impossible to render more than one instance of the same component

src/content/reference/react-dom/components/form.md
mattcarrollcode: Would making `action` not a bullet here fix the issue? or were you thinking of doing something else to make this more prominent?

src/content/reference/react-dom/components/form.md
mattcarrollcode: Fixed.

src/content/reference/react-dom/components/form.md
mattcarrollcode: fixed.

src/content/reference/react-dom/components/form.md
sophiebits: useId + getElementById would be technically OK (it's essentially how aria-describedby etc work) but refs are better

I do think this is annoyingly verbose though. per chat convo last night, maybe changing keys is more elegant?

src/content/reference/react-dom/components/form.md
mattcarrollcode: fixed

src/content/reference/react-dom/components/form.md
mattcarrollcode: fixed.

src/content/reference/react-dom/components/form.md
mattcarrollcode: Fixed

src/content/reference/react-dom/components/form.md
```diff
@@ -0,0 +1,451 @@
...
+    <input name="query" />
+    <button type="submit">Search</button>
+</form>
+```
+
+[See more examples below.](#usage)
+
+#### Props {/*props*/}
+
+`<form>` supports all [common element props.](/reference/react-dom/components/common#props)
```
davidmccabe, line 47: I really like how you're only documenting the props that we change from baseline, I think that's how we should handle this.

src/content/reference/react-dom/components/form.md
```diff
@@ -0,0 +1,451 @@
...
+    <form action={addToCart}>
+        <input type="hidden" name="productId" value={productId} />
+        <button type="submit">Add to Cart</button>
+    </form>
+
+  );
+}
+```
+
+In leiu, of using hidden form fields to provide data to the function passed to the `action` prop of `<form>` you can call the `bind` method on the function passed to the `action` prop of `<form>`.
```
davidmccabe, line 117:
```suggestion
In lieu of using hidden form fields to provide data to form action function, you can call the `bind` method to supply it with extra arguments.
```

src/content/reference/react-dom/components/form.md
```diff
@@ -0,0 +1,451 @@
...
+  return (
+    <form action={addProductToCart}>
+      <button type="submit">Add to Cart</button>
+    </form>
+  );
+}
+```
+
+### Display a pending state during form submission {/*display-a-pending-state-during-form-submission*/}
+To display a pending state while a form is submitting, you can call the `useFormStatus` Hook in a component rendered in a `<form>` and read the `pending` property returned.
```
davidmccabe, line 137: Not sure if it's consistent but it seems wrong to me to capitalize "Hook" here.

davidmccabe: I think you want the passive "is being submitted" throughout. Forms don't submit, they are submitted.

src/content/reference/react-dom/components/form.md
mattcarrollcode: Added a note on how hidden form fields can be used to encode data for a form action to consume.

src/content/reference/react-dom/components/form.md
davidmccabe: Why not make the input a controlled input?

src/content/reference/react-dom/components/form.md
davidmccabe: The simple use case of useFormState, and probably what the great majority of uses will be, is to read the return value from a server action. The return value might be an error message but not of the sort that you'd want to throw an exception for — for example, if you can't add an item to your cart because it's sold out. That's an application logic thing, not a system-level failure.

The fully general use case, which explains the name of the hook, is to implement a reducer state machine that can be progressively enhanced. This is because useFormState remembers the previous return value from the server and passes it back to the server again each time the action is invoked.

The API is quite weird and difficult to explain so I'm not surprised that trying to compress it into a paragraph is hard to follow. 

Here's another attempt:

To surface error messages to your users from a form action you can use the `useFormState` Hook. This Hook creates a state variable that is set to the most recent value returned by the form action. It takes two parameters: an action and an initial state; and returns two values: the latest state and a wrapped version of the form action. You pass this wrapped action into your form. The action that you pass receives the previous state as its first argument. When use in conjunction with [Server Actions](/reference/react/use-server) and [Server Components](/reference/react/use-client), the error message is progressively enhanced.

davidmccabe (commented):
This is great.

src/content/reference/react-dom/components/form.md
mattcarrollcode: Great idea! implemented.

src/content/reference/react-dom/components/form.md
sophiebits: I'm not sure we need to fully explain the signature of useFormState here. Just a sentence about what it does / why to use it, and then the code sample and link to its own page can speak for themselves.

mattcarrollcode: gotcha, thanks for the feedback, this is pretty confusing. I broke the useFormState example out into its own section and rewrote the wording but kept the example the same.

src/content/reference/react-dom/components/form.md
mattcarrollcode: SGTM, fixed.

src/content/reference/react-dom/components/form.md
mattcarrollcode: Good catch, fixed.

src/content/reference/react-dom/components/form.md
mattcarrollcode: I think that is outside the scope of the form docs, but probably belongs in the useOptimistic Hook doc.  @kmiddleton14  FYI


I'll add a note to add a link to the useOptimistic docs once they are merged.

src/content/reference/react-dom/components/form.md
mattcarrollcode: fixed

src/content/reference/react-dom/components/form.md
mattcarrollcode: Fixed, thanks!

src/content/reference/react-dom/components/form.md
mattcarrollcode: [Capital "H" Hook seems to be the convention on react.dev](https://github.com/search?q=repo%3Areactjs%2Freact.dev+%22%60+hook%22&type=code). I agree it looks a little strange

Fixed the tense issue.

src/content/reference/react-dom/components/form.md
mattcarrollcode: > Why not make the input a controlled input?

I think a controled input would work just fine, but I didn't do it in the initial version to try to keep this already very long example as short and as simple as possible.

I think useRef is the best option here and I've updated the code to use useRef.

src/content/reference/react-dom/components/form.md
```diff
@@ -0,0 +1,433 @@
...
+    <form action={addToCart}>
+        <input type="hidden" name="productId" value={productId} />
+        <button type="submit">Add to Cart</button>
+    </form>
+
+  );
+}
+```
+
+In lieu, of using hidden form fields to provide data to the `<form>`'s action, you can call the <CodeStep step={1}>`bind`</CodeStep> method to supply it with extra arguments. This will bind a new argument (<CodeStep step={2}>`productId`</CodeStep>) to the function in addition to the <CodeStep step={3}>`formData`</CodeStep> that is passed as a argument to the function.
```
kmiddleton14, line 121: nit extra ,
```suggestion
In lieu of using hidden form fields to provide data to the `<form>`'s action, you can call the <CodeStep step={1}>`bind`</CodeStep> method to supply it with extra arguments. This will bind a new argument (<CodeStep step={2}>`productId`</CodeStep>) to the function in addition to the <CodeStep step={3}>`formData`</CodeStep> that is passed as a argument to the function.
```

src/content/reference/react-dom/components/form.md
sophiebits: We do capitalize Hook and have done so since 2018 but 5 years later it may be time to lowercase, especially since the word "hook" is now uncommon in any other context (whereas people would talk about lifecycle hooks, etc. before).

src/content/reference/react-dom/components/form.md
```diff
@@ -0,0 +1,433 @@
...
+
+<Sandpack>
+
+```js App.js
+import { ErrorBoundary } from "react-error-boundary";
+import { action } from "./actions.js"
+
+export default function Search() {
+  return (
+    <ErrorBoundary fallback={<p>There was an error while submitting the form</p>}>
```
kmiddleton14, line 291: Not sure why but the fallback error message isn't displaying for me. Is this expected?

https://github.com/reactjs/react.dev/assets/20743223/a2726e61-c7b6-49ad-aa03-3b317c5943ca

src/content/reference/react-dom/components/form.md
```diff
@@ -0,0 +1,433 @@
...
+
+* When a function is passed to `action` or `formAction` the HTTP method will be POST regardless of value of the `method` prop.
+
+---
+
+## Usage {/*usage*/}
+
+### Handle form submission on the client {/*handle-form-submission-on-the-client*/}
+
+Pass a function to the `action` prop of form to run the function when the form is submitted. Unlike a conventional HTML action, the form won't call a server endpoint when a client function is passed to `action` prop of form. [`formData`](https://developer.mozilla.org/en-US/docs/Web/API/FormData) will be passed to the function as an arrgument so you can access the data submitted by the form.
```
lunaleaps, line 61:
```suggestion
Pass a function to the `action` prop of form to run the function when the form is submitted. This differs from the conventional [HTML action](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#action), which only accepts URLs. [`formData`](https://developer.mozilla.org/en-US/docs/Web/API/FormData) will be passed to the function as an argument so you can access the data submitted by the form.
```

src/content/reference/react-dom/components/form.md
```diff
@@ -0,0 +1,433 @@
...
+  "main": "/index.js",
+  "devDependencies": {}
+}
+```
+
+</Sandpack>
+
+### Handle form submission with a Server Action {/*handle-form-submission-with-a-server-action*/}
+
+Render a `<form>` with a input and submit button. Pass a server action (a function marked with [`'use server'`](/reference/react/use-server) to the `action` prop of form to run the function when the form is submitted.
```
lunaleaps, line 96:
```suggestion
Render a `<form>` with an input and submit button. Pass a server action (a function marked with [`'use server'`](/reference/react/use-server) to the `action` prop of form to run the function when the form is submitted.
```

src/content/reference/react-dom/components/form.md
```diff
@@ -0,0 +1,433 @@
...
+
+This is similar to the way forms work when a URL is passed to the `action` prop. The ability to submit forms without JavaScript enables users to submit forms before JavaScript has had a chance to load. This can be benificial to users who have a slow connection, device, or have JavaScript disabled.
```
lunaleaps, line 98: This paragraph kinda throws me off. Maybe it should be re-ordered. Or suggestion

```suggestion
Passing a server action for `<form action>` allow users to submit forms without JavaScript enabled or before the code has loaded. This is beneficial to users who have a slow connection, device, or have JavaScript disabled and is similar to the way forms work when a URL is passed to the `action` prop.
```

src/content/reference/react-dom/components/form.md
```diff
@@ -0,0 +1,433 @@
...
+
+You can use hidden form fields to provide data to the `<form>`'s action. The hidden form field data can be access via the [`formData`](https://developer.mozilla.org/en-US/docs/Web/API/FormData) argument pass to the function that is passed to the `action` prop of `<form>`.
```
lunaleaps, line 100:
```suggestion
You can use hidden form fields to provide data to the `<form>`'s action. The server action will be called with hidden form field data as an instance of [`FormData`](https://developer.mozilla.org/en-US/docs/Web/API/FormData).
```

src/content/reference/react-dom/components/form.md
```diff
@@ -0,0 +1,433 @@
...
+
+### Display a form submission error without JavaScript {/*display-a-form-submission-error-without-javascript*/}
+
+Displaying a form submissions error message without JavaScript requires that:
+
+1. `<form>` be rendered by a [Server Component](/reference/react/use-client)
+1. the function passed to the `<form>`'s `action` prop be a [Server Action](/reference/react/use-server)
+1. the `useFormState` Hook be used to display the error message
+
+`useFormState` takes two parameters: an [Server Action](/reference/react/use-server) and a inital state. `useFormState` returns two values, a state variable and a action. The action returned by `useFormState` should be passed to the `action` prop of the form. The state varible returned by `useFormState` can be used to displayed an error message. The value returned by the [Server Action](/reference/react/use-server) passed to `useFormState` will be used to update the state variable.
```
lunaleaps, line 331:
```suggestion
`useFormState` takes two parameters: a [Server Action](/reference/react/use-server) and an inital state. `useFormState` returns two values, a state variable and an action. The action returned by `useFormState` should be passed to the `action` prop of the form. The state varible returned by `useFormState` can be used to displayed an error message. The value returned by the [Server Action](/reference/react/use-server) passed to `useFormState` will be used to update the state variable.
```

src/content/reference/react-dom/components/form.md
lunaleaps: same, I'm wondering if it's because its not using RSCs? Or do we maybe need to wrap this in a startTransition?

src/content/reference/react-dom/components/form.md
```diff
@@ -0,0 +1,433 @@
...
+  "main": "/index.js",
+  "devDependencies": {}
+}
+```
+
+</Sandpack>
+
+### Display a form submission error without JavaScript {/*display-a-form-submission-error-without-javascript*/}
+
+Displaying a form submissions error message without JavaScript requires that:
```
lunaleaps, line 325: Nit: just to tie the two concepts together more
```suggestion
Displaying a form submissions error message without JavaScript, for progressive enhancement, requires that:
```

src/content/reference/react-dom/components/form.md
```diff
@@ -0,0 +1,433 @@
...
+    "react-dom": "18.3.0-canary-6db7f4209-20231021",
+    "react-scripts": "^5.0.0"
+  },
+  "main": "/index.js",
+  "devDependencies": {}
+}
+```
+
+</Sandpack>
+
```
lunaleaps, line 390: Should we link here to `useFormState` docs?

```suggestion

Learn more about updating state from a form action with the [`useFormState`](/reference/react-dom/hooks/useFormState) docs
```

src/content/reference/react-dom/components/form.md
```diff
@@ -0,0 +1,433 @@
...
+### Handling multiple submission types {/*processing-forms-differently-with-different-submit-buttons*/}
```
lunaleaps, line 391:
```suggestion
### Handling multiple submission types {/*handling-multiple-submission-types*/}
```

src/content/reference/react-dom/components/form.md
```diff
@@ -0,0 +1,433 @@
...
+
+Forms can be designed to handle multiple submission actions based on the button pressed by the user. Each button inside a form can be associated with a distinct action or behavior by setting the `formAction` prop. When a user taps a specific button, the form is submitted, and a corresponding action, defined by that button's attributes and action, is executed. For instance, a form might  submit a article for review by default but have a seperate button with `formAction` set to save the article as a draft.
```
lunaleaps, line 393:
```suggestion
Forms can be designed to handle multiple submission actions based on the button pressed by the user. Each button inside a form can be associated with a distinct action or behavior by setting the `formAction` prop. 

When a user taps a specific button, the form is submitted, and a corresponding action, defined by that button's attributes and action, is executed. For instance, a form might submit a article for review by default but have a seperate button with `formAction` set to save the article as a draft.
```

src/content/reference/react-dom/components/form.md
TheSavior: Does it work if you have the contents in a different component?

lunaleaps (approved):
Love this! Thanks for all the updates. I have some suggestions -- mostly spelling and occasional sentence re-ordering but those aren't blocking!

I think the only blocking thing is the error boundary example not working

src/content/reference/react-dom/components/form.md
sophiebits: @TheSavior should be equivalent since there aren't hooks in Search. ErrorBoundary doesn't care if it goes `ErrorBoundary > form` or `ErrorBoundary > Foo > form`

src/content/reference/react-dom/components/form.md
sophiebits: no commas in this edit please

also fyi I try to always say "before the JavaScript bundle loads" or similar. our out-of-order streaming in general relies on a tiny runtime to swap between suspense fallback and content, and to put streamed pieces in the right places – shouldn't be a concern for low-bandwidth or flaky connections but we shouldn't set the expectation that everything will work with JS entirely disabled

src/content/reference/react-dom/components/form.md
mattcarrollcode: Moving the client action into the component seems to have fixed the issue. I'll look into why this version of the code didn't work later.

src/content/reference/react-dom/components/form.md
mattcarrollcode: Fixed.

src/content/reference/react-dom/components/form.md
mattcarrollcode: fixed.

src/content/reference/react-dom/components/form.md
mattcarrollcode: sorry for the grammar mistakes, thank you, fixed!

src/content/reference/react-dom/components/form.md
mattcarrollcode: fixed.

src/content/reference/react-dom/components/form.md
mattcarrollcode: fixed.

src/content/reference/react-dom/components/form.md
mattcarrollcode: Thanks Sophie and Luna, I reworded this to cover both issues.

PR #6314, opened 2023-09-23
Add `useFormStatus` Hook API reference documentation

mattcarrollcode:
Add initial reference documentation for the `useFormStatus` Hook

Preview: https://react-dev-git-useformstatus-fbopensource.vercel.app/reference/react-dom/hooks/useFormStatus

### DIFF
--- a/src/content/reference/react-dom/hooks/useFormStatus.md
+++ b/src/content/reference/react-dom/hooks/useFormStatus.md
@@ -29,7 +29,7 @@ const { pending, data, method, action } = useFormStatus();
 
 The `useFormStatus` Hook provides status information of the last form submission.
 
-```js [[1, 5, "status.pending"]]
+```js {4},[[1, 5, "status.pending"]]
 import action from './actions';
 
 function Submit() {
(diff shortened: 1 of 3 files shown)


src/content/reference/react-dom/hooks/useFormStatus.md
```diff
@@ -0,0 +1,166 @@
...
+function Form() {
+  const { pending } = useFormStatus() // 🚩 `pending` will never be `true`
+  return <form action={submit}></form>
+}
+```
+
+Instead call `useFormStatus` from inside a Component that is located inside `<form>`.
+
+```js
+function Button(){
```
kmiddleton14, line 124: nit: Should we name this component `FormButton` to emphasize the point to make sure it's not used outside the Form component?

lunaleaps: I've renamed the component to `Submit`, I think that better captures the purpose of that component

lunaleaps:
@mattcarrollcode @kmiddleton14 @davidmccabe @TheSavior PR is ready for another look!

src/content/reference/react-dom/hooks/useFormStatus.md
```diff
@@ -0,0 +1,259 @@
...
+---
+
+## Reference {/*reference*/}
+
+### `useFormStatus()` {/*use-form-status*/}
+
+The `useFormStatus` Hook provides status information of the last form submission.
+
+```js [[1, 5, "status.pending"]]
+import action from './actions';
```
kmiddleton14, line 33: Is there a reason we're importing action from ./actions vs. getting it from `useFormStatus()`?  What're the differences between the action returned from `useFormStatus()` vs. using the imported one and when to use one over the other

src/content/reference/react-dom/hooks/useFormStatus.md
```diff
@@ -0,0 +1,259 @@
...
+
+function Submit() {
+  const status = useFormStatus();
```
kmiddleton14, line 36: nit: should we use the same convention as earlier?
`const { pending } = useFormStatus();`

src/content/reference/react-dom/hooks/useFormStatus.md
```diff
@@ -0,0 +1,259 @@
...
+
+`useFormStatus` does not take any parameters.
+
+#### Returns {/*returns*/}
+
+A `status` object with the following properties:
+
+* `pending`: A boolean. If `true`, this means the parent `<form>` is pending submission. Otherwise, `false`.
+* `data`: An object implementing the [`FormData interface`](https://developer.mozilla.org/en-US/docs/Web/API/FormData) that contains the data the parent `<form>` is submitting. `null` if there is no active submission or no parent `<form>`.
+* `method`: A string value of either `'get'` or `'post'`. This represents whether the parent `<form>` is submitting with either `GET` or `POST` [HTTP method](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods). By default, a `<form>` will use `GET` method and can be specified by the [`method`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#attributes_for_form_submission) property.
```
kmiddleton14, line 65: nit: Instead of https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#attributes_for_form_submission

Should we link to method specifically?
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#method?

src/content/reference/react-dom/hooks/useFormStatus.md
```diff
@@ -0,0 +1,259 @@
...
+    "react": "canary",
+    "react-dom": "canary",
+    "react-scripts": "^5.0.0"
+  },
+  "main": "/index.js",
+  "devDependencies": {}
+}
+```
+</Sandpack>  
+
```
kmiddleton14, line 248: Would we want to include a usage example for `method` and/or `action` too from `useFormStatus`?

kmiddleton14 (commented):
General comments and clarification questions but overall lgtm!

src/content/reference/react-dom/hooks/useFormStatus.md
lunaleaps: The `action` prop that is given in `useFormStatus` is the action we're passing to the `<form>` tag.

src/content/reference/react-dom/hooks/useFormStatus.md
lunaleaps: I intentionally replaced this with the object as I wanted to mirror the definition that `useFormStatus` returns status information, more than just `pending`. 

True, it's cleaner to use `{pending}` but my impression is that this version puts less emphasis on the other properties that are returned in `status` -- but just slightly. So happy to change if you strongly disagree.

src/content/reference/react-dom/hooks/useFormStatus.md
lunaleaps: Oh good catch! Updating

src/content/reference/react-dom/hooks/useFormStatus.md
lunaleaps: Yea I was wondering the same but I can't think of common patterns in which you'll need to reference the `method` and even more so the `action` prop. If you have any ideas, happy to add!

mattcarrollcode:
Thank you @lunaleaps getting this done! This looks great. I love the troubleshooting section and the accessing form data example. I didn't even know about the `data` property!

PR #6323, opened 2023-09-27
docs: comma is not part of a link

andrewvasilchuk:


harish-sethuraman:
We have the same instance at more than one place in the react docs. I remember Dan telling this is not an issue as well. So closing the issue thanks.

andrewvasilchuk:
@harish-sethuraman thanks for the review. By the way, here is the [Dan's comment](https://github.com/reactjs/react.dev/pull/5007#issuecomment-1239379738).

PR #6337, opened 2023-10-05
Add Experimental Taint API Docs

sebmarkbage:
These are still experimental. I haven't linked to them from anywhere yet.

Previews:
* [taintUniqueValue](https://react-dev-git-fork-sebmarkbage-taintapis-fbopensource.vercel.app/reference/react/experimental_taintUniqueValue)
* [taintObjectReference](https://react-dev-git-fork-sebmarkbage-taintapis-fbopensource.vercel.app/reference/react/experimental_taintObjectReference)

### DIFF
--- a/src/content/reference/react/experimental_taintObjectReference.md
+++ b/src/content/reference/react/experimental_taintObjectReference.md
@@ -1,0 +1,125 @@
+---
+title: experimental_taintObjectReference
+---
+
+<Wip>
+
+**This API is experimental and is not available in a stable version of React yet.**
+
+You can try it by upgrading React packages to the most recent experimental version:
+
+- `react@experimental`
+- `react-dom@experimental`
+- `eslint-plugin-react-hooks@experimental`
+
+Experimental versions of React may contain bugs. Don't use them in production.
+
+This API is only available inside React Server Components.
+
+</Wip>
+
+
+<Intro>
+
+`taintObjectReference` lets you prevent a passing a specific object instance to a Client Component by mistake.
+
+```js
+experimental_taintObjectReference(message, object);
+```
+
+To prevent passing a crypto key, hash or token, see [`taintUniqueValue`](/reference/react/experimental_taintUniqueValue).
+
+</Intro>
+
+<InlineToc />
+
+---
+
+## Reference {/*reference*/}
+
+### `taintObjectReference(message, object)` {/*taintobjectreference*/}
+
+Call `taintObjectReference` with an object to register it with React as something that should not be allowed to be passed to the Client as is:
+
+```js
+import {experimental_taintObjectReference} from 'react';
+
+experimental_taintObjectReference(
+  'Do not pass ALL environment variables to the client.',
+  process.env
+);
+```
+
+[See more examples below.](#usage)
+
+#### Parameters {/*parameters*/}
+
+* `message`: The message you want to display if the object gets passed to a Client Component. It's logged as any other Error that is thrown.
+
+* `object`: Any kind of object that should be tainted. You can pass functions and class instances here too. They're already blocked from being passed to Client Components but this lets you customize the error message. You can also taint a specific instance of a Typed Array without tainting the value in other Typed Array copies.
+
+#### Returns {/*returns*/}
+
+`experimental_taintObjectReference` returns `undefined`.
+
+---
+
+## Usage {/*usage*/}
+
+If you're running a Server Components that has access to sensitive data, you have to be careful not to pass objects straight through:
+
+```js
+export async function getUser(id) {
+  const user = await db`SELECT * FROM users WHERE id = ${id}`;
+  return user;
+}
+```
+
+```js
+import { getUser } from '...';
+import { InfoCard } from '...';
+
+export async function Profile(props) {
+  const user = await getUser(props.userId);
+  // DO NOT DO THIS
+  return <InfoCard user={user} />;
+}
+```
+
+```js
+"use client";
+
+export async function InfoCard({ user }) {
+  return <div>{user.name}</div>;
+}
+```
+
+A `"use client"` component should never accept objects that might be carrying sensitive data. Ideally, the getUser helper should not expose data that the current user can't see. Sometimes mistakes happen during refactoring.
+
+To protect against this mistakes happening down the line we can "taint" the user object in our data API.
+
+```js
+import {experimental_taintObjectReference} from 'react';
+
+export async function getUser(id) {
+  const user = await db`SELECT * FROM users WHERE id = ${id}`;
+  experimental_taintObjectReference(
+    'Do not pass the entire user object to the client. ' +
+      'Instead, pick off the specific properties you need for this use case.',
+    user,
+  );
+  return user;
+}
+```
+
+Now whenever anyone tries to pass this object to a Client Component, or bound to a Server Action, it'll error with the passed in error message instead.
+
+<Pitfall>
+
+**Do not rely on just tainting for security.** You should design your data access APIs and access to private keys in an isolated way such that you don't easily leak it into the Component rendering in the first place. Tainting is opt-in and easy to forget.
+
+Tainting provides an extra layer of protection if someone on your team still makes a mistake but it's not intended as the primary solution.
+
+Tainting an object doesn't prevent leaking of every possible derived value. For example, `const userInfo = {name: user.name, ssn: user.ssn}` create a new object which is not tainted. Creating a new object by just cloning it means it's no longer tainted `{...object}`. It only protects against simple mistakes when the object is passed straight through unchanged.
+
+</Pitfall>
--- a/src/content/reference/react/experimental_taintObjectReference.md
+++ b/src/content/reference/react/experimental_taintObjectReference.md
@@ -66,7 +66,7 @@ experimental_taintObjectReference(
 
 ## Usage {/*usage*/}
 
-If you're running a Server Components that has access to sensitive data, you have to be careful not to pass objects straight through:
+If you're running a Server Components environment that has access to sensitive data, you have to be careful not to pass objects straight through:
 
 ```js
 export async function getUser(id) {
--- a/src/content/reference/react/experimental_taintObjectReference.md
+++ b/src/content/reference/react/experimental_taintObjectReference.md
@@ -21,13 +21,13 @@ This API is only available inside React Server Components.
 
 <Intro>
 
-`taintObjectReference` lets you prevent a passing a specific object instance to a Client Component by mistake.
+`taintObjectReference` lets you prevent a specific object instance from being passed to a Client Component like a `user` object.
 
 ```js
 experimental_taintObjectReference(message, object);
 ```
 
-To prevent passing a crypto key, hash or token, see [`taintUniqueValue`](/reference/react/experimental_taintUniqueValue).
+To prevent passing a key, hash or token, see [`taintUniqueValue`](/reference/react/experimental_taintUniqueValue).
 
 </Intro>
 
@@ -37,7 +37,7 @@ To prevent passing a crypto key, hash or token, see [`taintUniqueValue`](/refere
 
 ## Reference {/*reference*/}
 
-### `taintObjectReference(message, object)` {/*taintobjectreference*/}
+### `taintObjectReference(errMessage, object)` {/*taintobjectreference*/}
 
 Call `taintObjectReference` with an object to register it with React as something that should not be allowed to be passed to the Client as is:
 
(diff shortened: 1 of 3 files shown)


src/content/reference/react/experimental_taintUniqueValue.md
```diff
@@ -0,0 +1,162 @@
...
+Experimental versions of React may contain bugs. Don't use them in production.
+
+This API is only available inside React Server Components.
+
+</Wip>
+
+
+<Intro>
+
+`taintUniqueValue` lets you prevent a passing a password, crypto graphic key or token, as a string, bigint or typed array, to a Client Component. It blocks this string completely so it cannot be a value that's also used in other contexts. It needs to be unique.
```
mattcarrollcode, line 24: nit: "It blocks this string completely" is a little confusing since you just mentioned in the previous sentance that the value can be a string, bigint or typed array.

maybe replace "passing a password, crypto graphic key or token, as a string, bigint or typed array," with "value like a password or token" since you already mention the data types in the reference for the "value" parameter?

src/content/reference/react/experimental_taintUniqueValue.md
```diff
@@ -0,0 +1,162 @@
...
+
+---
+
+## Reference {/*reference*/}
+
+### `taintUniqueValue(message, lifetime, value)` {/*taintuniquevalue*/}
+
+Call `taintUniqueValue` with a password, token, key or hash to register it with React as something that should not be allowed to be passed to the Client as is:
+
+```js
```
mattcarrollcode, line 44: Love this example! its very clear how to use the API just from this snippet

src/content/reference/react/experimental_taintUniqueValue.md
```diff
@@ -0,0 +1,162 @@
...
+
+`experimental_taintUniqueValue` returns `undefined`.
+
+---
+
+## Usage {/*usage*/}
+
+If you're running a Server Components environment that has access to private keys or passwords such as database passwords, you have to be careful not to pass that to a Client Component.
+
+```js
```
mattcarrollcode, line 74: This example is great but its a little confusing to understand the relationship with the two files. Maybe add a comment at the top indicating the filename of the file with `Overview" in it  and a import statement in the file with `Dashboard` in it to "import" `Overview` into the file with `Dashboard`?

mattcarrollcode: Maybe once this lands in experimental we can make this into a code sandbox?

src/content/reference/react/experimental_taintUniqueValue.md
```diff
@@ -0,0 +1,162 @@
...
+    const headers = { Authorization: password };
+    fetch(url, { headers }).then(...);
+  }, [password]);
+  ...
+}
+```
+
+This example would leak the secret API token to the client. If this API token can be used to access data this particular user shouldn't have access to, it's a security problem.
+
+Ideally, secrets like this are abstracted into a single helper file that can only be imported by trusted data utilities on the server. The helper can even be tagged with `server-only` to ensure that this file isn't imported on the client.
```
mattcarrollcode, line 97: What do we need to do to transfer ownership of `server-only` to React and document `server-only` on another page? Is there anything stopping us from doing both those things?

src/content/reference/react/experimental_taintUniqueValue.md
```diff
@@ -0,0 +1,162 @@
...
+
+experimental_taintUniqueValue(
+  'Do not pass the API token password to the client. ' +
+    'Instead do all fetches on the server.'
+  process,
+  process.env.API_PASSWORD
+);
+```
+
+Now whenever anyone tries to pass this password to a Client Component, or bound to a Server Action, it'll error with the passed in error message instead.
```
mattcarrollcode, line 123: What kind of error? how is this surfaced? e.g. if you try to pass a tainted value to a client component will it error on the server side? will in trigger a error boundary? If you try to return a tainted value from a server action will the server action return an error?

src/content/reference/react/experimental_taintUniqueValue.md
```diff
@@ -0,0 +1,162 @@
...
+
+```js
+import {experimental_taintUniqueValue} from 'react';
+
+export async function getUser(id) {
+  const user = await db`SELECT * FROM users WHERE id = ${id}`;
+  experimental_taintUniqueValue(
+    'Do not pass a user session token to the client.',
+    user,
+    user.session.token,
```
mattcarrollcode, line 141: I thought you mentioned that the taint APIs would only do a shallow inspection of objects. If the value is in `user.session.token` will that still get picked up/labeled by `taintUniqueValue`? In what situations will `taintUniqueValue` *not* taint a string from a object?

Might also be good to have a sentance or two and a link to `taintObjectReference` since you're using a object in this example.

src/content/reference/react/experimental_taintUniqueValue.md
```diff
@@ -0,0 +1,162 @@
...
+  );
+  return user;
+}
+```
+
+This ensures that if this `user` object gets stored in some global store, and another request can read from it, the token is still tainted. If it gets cleaned up then we don't end up leaking memory.
```
mattcarrollcode, line 147: What does `tainted` mean in this context? Will the object still be passed from server to client but the `user.session.token` is missing? Will there be an error at any point? What happens if the client tries to access `user.session.token` will it just be missing or will React throw an error?

src/content/reference/react/experimental_taintUniqueValue.md
```diff
@@ -0,0 +1,162 @@
...
+
+If the lifetime object gets GC:ed during an on-going request, React will still keep the value tainted for the duration of this request.
+
+
+<Pitfall>
```
mattcarrollcode, line 152: Feels to me like this is important enough that it should be more prominent (e.g. further up the page)

src/content/reference/react/experimental_taintObjectReference.md
```diff
@@ -0,0 +1,125 @@
...
+
+`experimental_taintObjectReference` returns `undefined`.
+
+---
+
+## Usage {/*usage*/}
+
+If you're running a Server Components environment that has access to sensitive data, you have to be careful not to pass objects straight through:
+
+```js
```
mattcarrollcode, line 71: This is a great example but hard to see the relationship between the files. Maybe add filenames in comments at the top of code snippets and use them in place of `...` in the imports (e.g. `from '...'`)?

Alternatively, could we move this to a code sandbox after this lands in the experimental channel.

src/content/reference/react/experimental_taintObjectReference.md
```diff
@@ -0,0 +1,125 @@
...
+  experimental_taintObjectReference(
+    'Do not pass the entire user object to the client. ' +
+      'Instead, pick off the specific properties you need for this use case.',
+    user,
+  );
+  return user;
+}
+```
+
+Now whenever anyone tries to pass this object to a Client Component, or bound to a Server Action, it'll error with the passed in error message instead.
```
mattcarrollcode, line 115: Maybe we can add a similar Server Action usage example?

src/content/reference/react/experimental_taintObjectReference.md
```diff
@@ -0,0 +1,125 @@
...
+
+<Pitfall>
```
mattcarrollcode, line 117: Feels to me like this is important enough that it should be more prominent (e.g. further up the page)

src/content/reference/react/experimental_taintObjectReference.md
```diff
@@ -0,0 +1,125 @@
...
+
+* `object`: Any kind of object that should be tainted. You can pass functions and class instances here too. They're already blocked from being passed to Client Components but this lets you customize the error message. You can also taint a specific instance of a Typed Array without tainting the value in other Typed Array copies.
+
+#### Returns {/*returns*/}
+
+`experimental_taintObjectReference` returns `undefined`.
+
+---
+
+## Usage {/*usage*/}
```
mattcarrollcode, line 67: Maybe create a subsection under usage with a name for this usage example?

src/content/reference/react/experimental_taintUniqueValue.md
```diff
@@ -0,0 +1,162 @@
...
+
+* `value`: A string, bigint or TypedArray. This value must be a unique sequence of characters/bytes with high entropy such as a cryptographic token, private key, hash or a long password. This value gets banned for passing to any Client Component in the whole app.
+
+#### Returns {/*returns*/}
+
+`experimental_taintUniqueValue` returns `undefined`.
+
+---
+
+## Usage {/*usage*/}
```
mattcarrollcode, line 70: Maybe create a subsection under usage with a name for this usage example?

mattcarrollcode (commented):
This is great! Thank you for doing this. This makes a huge difference in being able to document these APIs early. You writing the initial docs takes so much back and forth out of the equation, and makes write the docs so much more pleasant!

I did a quick pass and added some thoughts. I'm also happy to take this PR over and make the changes I suggested if you like.

Another general thought: We should probably add this to the ToC with some kind of label like canary to indicate this is an experimental API (but I can do this after the fact if you like)

src/content/reference/react/experimental_taintUniqueValue.md
```diff
@@ -45,30 +45,78 @@ Call `taintUniqueValue` with a password, token, key or hash to register it with
...
+  process.env.SECRET_KEY
 );
 ```
 
 [See more examples below.](#usage)
 
 #### Parameters {/*parameters*/}
 
-* `message`: The message you want to display if the object gets passed to a Client Component. It's logged as any other Error that is thrown.
+* `errMessage`: The message you want to display if the object gets passed to a Client Component. If `value` is passed to a Client Component an Error will be thrown. `errMessage` let's you define the message displayed by this Error.
```
sebmarkbage, line 58: This is not what it's called in the code and in TypeScript definitions.

One thing here is that this is currently just passed straight to new Error but I don't think that will be the end state. We'll likely provide more context inside the error.

src/content/reference/react/experimental_taintObjectReference.md
```diff
@@ -54,30 +54,67 @@ experimental_taintObjectReference(
...
+* `object`: The object to be tainted.
 
 #### Returns {/*returns*/}
 
 `experimental_taintObjectReference` returns `undefined`.
 
+#### Caveats {/*caveats*/}
+
+- Recreating or cloning a tainting object creates a new untained object which main contain sensetive data. For example, if you have a tainted `user` object, `const userInfo = {name: user.name, ssn: user.ssn}` or `{...user}` will create new objects which are not tainted.  `taintObjectReference` only protects against simple mistakes when the object is passed through to a Client Component unchanged.
+- Functions and class instances can be passed to `taintObjectReference` as `object`. Functions and classes are already blocked from being passed to Client Components but the React's default error message will be replaced by what you defined in `errMessage`.
```
sebmarkbage, line 68: This isn't really a caveat. It's an explicit feature same with typed arrays.

src/content/reference/react/experimental_taintUniqueValue.md
```diff
@@ -45,30 +45,78 @@ Call `taintUniqueValue` with a password, token, key or hash to register it with
...
-* `value`: A string, bigint or TypedArray. This value must be a unique sequence of characters/bytes with high entropy such as a cryptographic token, private key, hash or a long password. This value gets banned for passing to any Client Component in the whole app.
+* `value`: A string, bigint or TypedArray. `value` must be a unique sequence of characters or bytes with high entropy such as a cryptographic token, private key, hash, or a long password. `value` will be blocked from being sent to any Client Component.
 
 #### Returns {/*returns*/}
 
 `experimental_taintUniqueValue` returns `undefined`.
 
+#### Caveats {/*caveats*/}
+
+- Modifying tainted values can invalidate tainting protection. Converting tainted values to upper case, concatenating tainted string values into a larger string, converting tainted values to base64, returning a substring, and similar transformations will lose tainting protection.
```
sebmarkbage, line 70: Not sure "modifying" is the right word here because you can't really mutate them which might be implied. It's more about deriving into new values than modifying the existing ones. E.g. concatenation isn't really modifying.

sebmarkbage (commented):
lgtm. Left some nits.

gnoff (commented):
Might need to go in a separate PR but we should change the API ordering to sort based on the first letter of the string after `experimental_` This will help keep the order stable when the experimental prefix is removed and make it clearer where to find the API when you only know it by it's semantic name, "taint" for instance

alternatively consider putting all experimental references at the end of the list

mattcarrollcode:
> Might need to go in a separate PR but we should change the API ordering to sort based on the first letter of the string after `experimental_` This will help keep the order stable when the experimental prefix is removed and make it clearer where to find the API when you only know it by it's semantic name, "taint" for instance
> 
> alternatively consider putting all experimental references at the end of the list

Good note. I'll all the experimental APIs at the end of the list for now.

src/content/reference/react/experimental_taintUniqueValue.md
```diff
@@ -0,0 +1,177 @@
...
+);
+```
+
+[See more examples below.](#usage)
+
+#### Parameters {/*parameters*/}
+
+* `message`: The message you want to display if `value` is passed to a Client Component. This message will be displayed as a part of the Error that will be thrown if `value` is passed to a Client Component.
+
+* `lifetime`: Any object that indicates how long `value` should be tainted. `value` will be blocked from being sent to any Client Component while this object still exists. For example, passing `globalThis` blocks the value for the lifetime of an app. `lifetime` is typically an object whose properties contains `value`.
```
lunaleaps, line 60:
```suggestion
* `lifetime`: Any object that indicates how long `value` should be tainted. `value` will be blocked from being sent to any Client Component while this object still exists. For example, passing [`globalThis`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis) blocks the value for the lifetime of an app. `lifetime` is typically an object whose properties contains `value`.
```

src/content/reference/react/experimental_taintUniqueValue.md
```diff
@@ -0,0 +1,177 @@
...
+
+* `value`: A string, bigint or TypedArray. `value` must be a unique sequence of characters or bytes with high entropy such as a cryptographic token, private key, hash, or a long password. `value` will be blocked from being sent to any Client Component.
+
+#### Returns {/*returns*/}
+
+`experimental_taintUniqueValue` returns `undefined`.
+
+#### Caveats {/*caveats*/}
+
+- Deriving new values from tainted values can compromise tainting protection. New values created by uppercasing tainted values, concatenating tainted string values into a larger string, converting tainted values to base64, substringing tainted values, and other similar transformations are not tainted unless you explicity call `taintUniqueValue` on these newly created values.
```
lunaleaps, line 70: This feels a bit repetitive with the Pitfall right after it which is essentially saying the same thing? 
Perhaps we could move the Pitfall after a usage example and showcase how deriving a new value from the example wouldn't be tainted

sophiebits:
Did you want to mention that the value taint lives through the end of a request?

sebmarkbage:
The way it's described is in terms of the reverse, that it only gets promoted if the object outlives the request.

> If this object gets stored in a global cache or is accessible by another request, the session token remains tainted.

Not sure the intricacies are important for knowing how it's used. It might be important if you're vetting the implementation details but you should probably assume it works as described otherwise.

PR #6339, opened 2023-10-06
Explain limitations of useSyncExternalStore with concurrency

sophiebits:
Doesn't seem like we have this noted anywhere.

[Preview](https://react-dev-git-sophiebits-patch-5-fbopensource.vercel.app/reference/react/useSyncExternalStore#caveats)

### DIFF
--- a/src/content/reference/react/useSyncExternalStore.md
+++ b/src/content/reference/react/useSyncExternalStore.md
@@ -59,7 +59,23 @@ The current snapshot of the store which you can use in your rendering logic.
 
 * If the store is mutated during a [non-blocking transition update](/reference/react/useTransition), React will fall back to performing that update as blocking. Specifically, React will call `getSnapshot` a second time just before applying changes to the DOM. If it returns a different value than when it was called originally, React will restart the transition update from scratch, this time applying it as a blocking update, to ensure that every component on screen is reflecting the same version of the store.
 
-* It's not recommended to _suspend_ a render based on a store value returned by `useSyncExternalStore`. For example, if you set `const expandedItemId = useSyncExternalStore(...)`, avoid rendering a [`lazy`](/reference/react/lazy) component conditionally like `{expandedItemId != null && <LazyItemDetails />}` or calling [`use`](/reference/react/use) with a Promise dependent on the value like `use(fetchItem(expandedItemId))`. The reason is that mutations to the external store cannot be [marked as non-blocking transition updates](/reference/react/useTransition), so they will trigger the nearest [`Suspense` fallback](/reference/react/Suspense), replacing already-rendered content on screen with a loading spinner, which typically makes a poor UX.
+* It's not recommended to _suspend_ a render based on a store value returned by `useSyncExternalStore`. The reason is that mutations to the external store cannot be [marked as non-blocking transition updates](/reference/react/useTransition), so they will trigger the nearest [`Suspense` fallback](/reference/react/Suspense), replacing already-rendered content on screen with a loading spinner, which typically makes a poor UX.
+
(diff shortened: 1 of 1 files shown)


sophiebits:
(@acdlite if you wouldn't mind checking the accuracy/completeness of my explanation)

src/content/reference/react/useSyncExternalStore.md
```diff
@@ -57,6 +57,10 @@ The current snapshot of the store which you can use in your rendering logic.
 
 * If a different `subscribe` function is passed during a re-render, React will re-subscribe to the store using the newly passed `subscribe` function. You can prevent this by declaring `subscribe` outside the component.
 
+* If the store is mutated during a [non-blocking transition update](/reference/react/useTransition), React will fall back to performing that update as blocking. Specifically, React will call `getSnapshot` a second time just before applying changes to the DOM. If it returns a different value than when it was called originally, React will restart the transition update from scratch, this time applying it as a blocking update, to ensure that every component on screen is reflecting the same version of the store.
```
rickhanlonii, line 60: Is this only for transitions, or also useDeferredValue, and eventually useFormState and unOptimistic? Would it be better to talk about "non urgent updates" instead of transitions specifically?

sophiebits: I’m not picky about the nomenclature here. I usually call them transition updates because I think we call it transition priority internally IIRC but fine with whatever. I wish we had a page that talked about update priorities generally.

src/content/reference/react/useSyncExternalStore.md
```diff
@@ -57,6 +57,10 @@ The current snapshot of the store which you can use in your rendering logic.
...
+
+* It's not recommended to _suspend_ a render based on a store value returned by `useSyncExternalStore`. For example, if you set `const expandedItemId = useSyncExternalStore(...)`, avoid rendering a [`lazy`](/reference/react/lazy) component conditionally like `{expandedItemId != null && <LazyItemDetails />}` or calling [`use`](/reference/react/use) with a Promise dependent on the value like `use(fetchItem(expandedItemId))`. The reason is that mutations to the external store cannot be [marked as non-blocking transition updates](/reference/react/useTransition), so they will trigger the nearest [`Suspense` fallback](/reference/react/Suspense), replacing already-rendered content on screen with a loading spinner, which typically makes a poor UX.
```
mattcarrollcode, line 62: > For example, if you set const expandedItemId = useSyncExternalStore(...), avoid rendering a lazy component conditionally like {expandedItemId != null && <LazyItemDetails />} or calling use with a Promise dependent on the value like use(fetchItem(expandedItemId)).

this sentence is a little hard to follow. Could you put the examples in to a code block? If I understand this correctly I think it should look something like this?:

```js
import { use, lazy } from 'react';

const MarkdownPreview = lazy(() => import('./MarkdownPreview.js'));

function TodosApp() {
  const todos = useSyncExternalStore(...);
  ❌ passing `todos` to the `use` Hook can suspend a render based on a store value
  use(fetchItem(todos[0]))
  ❌ conditionally rendering a lazy Component based on `todos` can suspend a render based on a store value
  return <>{todo != null && <MarkdownPreview />}</>
  // ...
}
```

mattcarrollcode (approved):

src/content/reference/react/useSyncExternalStore.md
sophiebits: I wasn't sure if I'm allowed to put a code block inside a bullet point. I can play around with other treatments if you think this isn't parseable but this is an advanced concept (and advanced hook) so I figured it is ok if it isn't super simple.

PR #6342, opened 2023-10-06
Add default case for useReducer form example

redrambles:
The current example's reducer function doesn't have a default case, so I thought I'd add one and put the error throwing inside it. :)


sophiebits:
the code is fine as-is; the way it is has the advantage of also throwing if you forget to `return` in one of the cases

PR #6344, opened 2023-10-09
docs: Remove unnecessary `when`

locol23:
Thanks for the nice documents.
I found a typo. Maybe one `when` would be fine.


sophiebits:
thanks!

Issue #6336, opened 2023-10-04
Missing early return in set function example?


alesmenzel:
Regarding [this example](https://react.dev/reference/react/useState#im-trying-to-set-state-to-a-function-but-it-gets-called-instead)

```jsx
import { useState } from 'react';

export default function CountLabel({ count }) {
  const [prevCount, setPrevCount] = useState(count);
  const [trend, setTrend] = useState(null);
  if (prevCount !== count) {
    setPrevCount(count);
    setTrend(count > prevCount ? 'increasing' : 'decreasing');
//    return; here?
  }
  return (
    <>
      <h1>{count}</h1>
      {trend && <p>The count is {trend}</p>}
    </>
  );
}
```

**It seems to me that the example should include an early `return` statement after `setTrend`** so React doesn't have to compute a tree that might be discarded immediately because the trend is changed from `null` to `increasing` / `decreasing`

first render
```jsx
return (
    <>
      <h1>{count}</h1>
    </>
  );
```

immediate second pass
```jsx
return (
    <>
      <h1>{count}</h1>
      <p>The count is increasing</p>
    </>
  );
```

wAsI7:
Working on it!!

PR #6345, opened 2023-10-09
Fix:#6336- Adding Early return in set function example

wAsI7:

### DIFF
--- a/src/content/reference/react/useState.md
+++ b/src/content/reference/react/useState.md
@@ -1114,22 +1114,36 @@ export default function App() {
 ```
 
 ```js CountLabel.js active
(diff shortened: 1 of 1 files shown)


smikitky:
Why add a new `useEffect` to adjust state? This is an anti-pattern, and the current docs [clearly says you shouldn't rely on `useEffect` to solve this](https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes).

@alesmenzel pointed out in #6336 that the code is *slightly* inefficient without `return`, but this fix worsens the situation by forcing the component to render twice.

PR #6350, opened 2023-10-12
Update Google Anaytics 4 tag on react.dev

kmiddleton14:
This PR adds the GA4 tag to our react.dev site.

Note that the GA4 dashboard is already setup, but the script wasn't added to populate the data. This PR adds it in following Google's installation steps. 

Screenshot 2023-10-12 at 5 21 50 PM

## Test plan
- `yarn check-all` passes
- Verified site visually looks ok
- Viewed page source and verified script is there
Screenshot 2023-10-12 at 5 27 20 PM

- Verified the window.dataLayer has the correct config

Screenshot 2023-10-12 at 5 08 54 PM

### DIFF
--- a/src/siteConfig.js
+++ b/src/siteConfig.js
@@ -16,4 +16,5 @@ exports.siteConfig = {
     apiKey: 'e8451218980a351815563de007648b00',
     indexName: 'beta-react',
   },
+  googleAnalyticsTagId: 'G-B1E83PJ3RT',
 };
--- a/src/pages/_document.tsx
+++ b/src/pages/_document.tsx
@@ -9,6 +9,15 @@ const MyDocument = () => {
   return (
     <Html lang={siteConfig.languageCode}>
       <Head />
+      <script
+        async
+        src={`https://www.googletagmanager.com/gtag/js?id=${siteConfig.googleAnalyticsTagId}`}
+      />
+      <script
+        dangerouslySetInnerHTML={{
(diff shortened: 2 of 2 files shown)


src/siteConfig.js
```diff
@@ -16,4 +16,5 @@ exports.siteConfig = {
     apiKey: 'e8451218980a351815563de007648b00',
     indexName: 'beta-react',
   },
+  googleAnalyticsTagId: 'G-B1E83PJ3RT',
```
harish-sethuraman, line 19: should we remove the already existing google analytics? we were using [`ga-lite`](https://github.com/reactjs/react.dev/pull/4355)

kmiddleton14: Yes but maybe in another PR to keep the changes separate

kmiddleton14:
Also verified data is being populated locally through GA tool
Screenshot 2023-10-13 at 11 30 29 AM

mattcarrollcode:
Does this have any impact on initial page/site load performance? Is this compatible with the "thumbs up" and "thumbs down" buttons on most pages?

harish-sethuraman:
The GA4 script adds upto `~80kB` to the bundle compressed (~227kB un compressed). How ever the file is served from disk cache after the first load. But some of the script is un used as per the lighthouse run. Yes we can track the `thumbsup` and  `thumbsdown` with this. (note: `ga-lite` [added around `2kB compressed` only to the bundle size](https://bundlephobia.com/package/ga-lite@2.1.6))

Screenshot 2023-10-16 at 1 01 58 PM

lunaleaps (approved):

src/siteConfig.js
kmiddleton14: Cleanup here https://github.com/reactjs/react.dev/pull/6366

PR #6353, opened 2023-10-15
fix showing of feedback container when submitted

affu72:
What's the problem here?

- After a user submits feedback, the feedback container remains persistent and does not disappear.
- If a user provides feedback for a specific page and then revisits that page, the feedback container reappears, prompting for feedback again.

This PR will solve the above two issues.

### DIFF
--- a/src/components/Layout/Feedback.tsx
+++ b/src/components/Layout/Feedback.tsx
@@ -2,15 +2,41 @@
  * Copyright (c) Facebook, Inc. and its affiliates.
  */
 
-import {useState} from 'react';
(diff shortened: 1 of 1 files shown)


harish-sethuraman:
> If a user provides feedback for a specific page and then revisits that page, the feedback container reappears, prompting for feedback again.


Im not sure if this was done getting incremental feedback as the page might have change even after the feedback was given by the user. But it makes sense to not show feedback now as the content is already there and not changing much. Would like someone from the team to verify on this cc: @mattcarrollcode

PR #6354, opened 2023-10-17
Add Error Boundary to useTransition API docs

kmiddleton14:
Error Boundary now works in React canary with useTransition API and displays the fallback.  This PR adds documentation with a simple usage example. 

Note that we want to prevent the error dialog from displaying in the Sandpack when we throw an error so that we can show the error boundary fallback.

[Preview link](https://react-nzxyyeyvn-fbopensource.vercel.app/reference/react/useTransition#displaying-an-error-to-users-with-error-boundary)

Screenshot 2023-10-17 at 5 01 53 PM

https://github.com/reactjs/react.dev/assets/20743223/f3ad1ff8-41ae-4959-b99f-d187270a7791


src/components/MDX/Sandpack/Preview.tsx
```diff
@@ -52,6 +52,12 @@ export function Preview({
     rawError = null;
   }
 
+  // When throwing a new Error in Sandpack - we want to disble the dev error dialog
```
kassens, line 55:
```suggestion
  // When throwing a new Error in Sandpack - we want to disable the dev error dialog
```

src/content/reference/react/useTransition.md
```diff
@@ -1501,6 +1502,92 @@ main {
...
+</Canary>
+
+If you'd like to display an error to your users when using `useTransition`, you can use an [error boundary](/reference/react/Component#catching-rendering-errors-with-an-error-boundary). To use an error boundary, wrap the component where you are calling the `useTransition` in an error boundary. If the function in `startTransition` is rejected the fallback for the error boundary will be displayed.
+
+<Sandpack>
+
+```js SubmitContainer.js active
+
+import { useTransition } from "react";
+import { ErrorBoundary } from "react-error-boundary";
```
kassens, line 1520: The React docs should probably use React's APIs directly?

src/content/reference/react/useTransition.md
```diff
@@ -1501,6 +1502,92 @@ main {
...
+import { SubmitContainer } from "./SubmitContainer.js";
+
+export default function App() {
+  return <SubmitContainer />;
+}
+```
+
+```js index.js hidden
+// TODO: update to import from stable
+// react instead of canary once the `use`
```
kassens, line 1557: copypasta? Did you mean `useTransition`?

src/components/MDX/Sandpack/Preview.tsx
```diff
@@ -52,6 +52,12 @@ export function Preview({
...
+  // to show the Error Boundary fallback
+  if (rawError && rawError.message.includes('throw new Error()')) {
```
kassens, line 57: are the demos interactive and changable? This hardcoded detection might break easily if someone throws a different error if that's the case.

src/content/reference/react/useTransition.md
```diff
@@ -1501,6 +1502,92 @@ main {
...
 
+### Displaying an error to users with a error boundary {/*displaying-an-error-to-users-with-error-boundary*/}
+
+<Canary>
+
+Error Boundary for useTransition is currently only available in React's canary and experimental channels. Learn more about [React's release channels here](/community/versioning-policy#all-release-channels).
+
+</Canary>
+
+If you'd like to display an error to your users when using `useTransition`, you can use an [error boundary](/reference/react/Component#catching-rendering-errors-with-an-error-boundary). To use an error boundary, wrap the component where you are calling the `useTransition` in an error boundary. If the function in `startTransition` is rejected the fallback for the error boundary will be displayed.
```
kassens, line 1513: I think "rejected" is a term typically just used with Promises. "throws" would be more general.

src/components/MDX/Sandpack/Preview.tsx
lunaleaps: As a slightly less fragile approach, is it possible to create a custom Error class for this case?

src/content/reference/react/useTransition.md
```diff
@@ -1501,6 +1502,92 @@ main {
...
+
+<Sandpack>
+
+```js SubmitContainer.js active
+
```
lunaleaps, line 1518:
```suggestion
```

src/content/reference/react/useTransition.md
```diff
@@ -1501,6 +1502,92 @@ main {
...
+  return (
+    <button 
+      disabled={pending} 
+      onClick={() => {
+        startTransition(() => {
+          throw new Error();
+        });
+      }}>
+        Submit
```
lunaleaps, line 1541: Maybe we could make this button label be like: "Trigger Error Boundary" to be clearer

lunaleaps: Or, lean into a real-world example where we'd use a `useTransition` like the example of loading indicators: https://react-nzxyyeyvn-fbopensource.vercel.app/reference/react/useTransition#preventing-unwanted-loading-indicators

src/content/reference/react/useTransition.md
sophiebits: Technically since it's an async function it is a promise that rejects after the first `await`, right? Not sure the clearest way to say this.

kassens: "throws" feels right to me for async functions as you can `catch` the error around the `await`. Just a feeling though!

sophiebits:
I'm looking at this again and I'm worried this isn't realistic enough to be helpful. Do we really expect people to throw on purpose directly in a transition? Or should we be trying to explain what happens when there is an unhandled error.

Assuming the latter (which had been my assumption), can we make the code call some a function with a real product action name that ends up throwing? Like `function addComment() {}`. It can fail unconditionally but what we should be teaching IMO is how errors in normal code bubble up when that code fails.

src/content/reference/react/useTransition.md
kmiddleton14: I'll update this example to be more product specific based on yours and @sophiebits's suggestions

src/content/reference/react/useTransition.md
kmiddleton14: ErrorBoundary isn't part of a React API. See note [here](https://react.dev/reference/react/Component#catching-rendering-errors-with-an-error-boundary) and package [here](https://www.npmjs.com/package/react-error-boundary)

kmiddleton14:
Not ready for review - had some merging issues and will resolve first

src/components/MDX/Sandpack/Preview.tsx
kmiddleton14: IMO it's ok if someone adds a different error and we show the dialog. Having this check is mostly to demonstrate the initial fallback. If they fork the code in codesandbox they'll be able to close the modal and see the fallback too. It's mostly for initial demonstration purposes. I can update this in a separate PR if we feel differently in the future.

src/content/reference/react/useTransition.md
```diff
@@ -1501,6 +1502,100 @@ main {
...
 
+### Displaying an error to users with a error boundary {/*displaying-an-error-to-users-with-error-boundary*/}
+
+<Canary>
+
+Error Boundary for useTransition is currently only available in React's canary and experimental channels. Learn more about [React's release channels here](/community/versioning-policy#all-release-channels).
+
+</Canary>
+
+If you'd like to display an error to your users when using `useTransition`, you can use an [error boundary](/reference/react/Component#catching-rendering-errors-with-an-error-boundary). To use an error boundary, wrap the component where you are calling the `useTransition` in an error boundary. If the function passed to `startTransition` is rejected, the fallback for the error boundary will be displayed.
```
lunaleaps, line 1513: Nit/Suggestion: Rephrasing to lead with why you'd want to consider an error boundary

```suggestion
If a function passed to `startTransition` throws an error, you can display an error to your user with an [error boundary](/reference/react/Component#catching-rendering-errors-with-an-error-boundary). To use an error boundary, wrap the component where you are calling the `useTransition` in an error boundary. Once the function passed to `startTransition` errors, the fallback for the error boundary will be displayed.
```

src/content/reference/react/useTransition.md
```diff
@@ -1501,6 +1502,100 @@ main {
...
+
+export function AddCommentContainer() {
+  return (
+    <ErrorBoundary fallback={<p>⚠️Something went wrong</p>}>
+        <AddCommentButton />
+    </ErrorBoundary>
+  );
+}
+
+function addComment(comment) {
```
lunaleaps, line 1529: We can probably remove the comment param and null check and just always throw?

lunaleaps (approved):

src/content/reference/react/useTransition.md
kmiddleton14: True the check isn't necessary, but I kind of wanted to make it feel more realistic when we'd throw an error

PR #6362, opened 2023-10-20
'use client' - flesh out usages and reference

lunaleaps:
Flesh out reference docs for 'use client' by adding more thorough explanation and usage examples 

### [Preview](https://react-dev-git-use-client-fbopensource.vercel.app/reference/react/use-client)


src/content/reference/react/use-client.md
```diff
@@ -11,7 +11,7 @@ canary: true
 
 <Intro>
 
-`'use client'` marks source files whose components execute on the client.
+`'use client'` marks which components render on the client.
```
sophiebits, line 14: (I realize this is a draft so not reviewing in detail but wanted to note one high-level thing…)

thought: is there a way we can be clear up top that this marks the _boundary_ between server and client components? ideally people should realize that you need serializable props and that you don't need to mark every file deep down. it's great you have FancyText but I think it could be a bit clearer throughout

(obviously my intro text you replaced did not even try to clarify this! oops. what you wrote here is a clear improvement)

public/images/docs/diagrams/use_client_render_tree.png
TheSavior, line 1: Hmm, this is hard because InspirationGenerator doesn't render Copyright, Copyright is passed in. I'm not sure how to show that here though. I think as drawn, people will ask "wait, how does a client component render a server component?!"

sophiebits: It certainly decides if and where to show it. I think the fact that this might be unintuitive is the whole point!

src/content/reference/react/use-client.md
lunaleaps: Updated to try and highlight the concept of boundary more

src/content/reference/react/use-client.md
```diff
@@ -34,26 +34,343 @@ export default function RichTextEditor(props) {
...
+
+</Sandpack>
+
+<Diagram name="use_client_module_dependency" height={250} width={545} alt="A tree graph with the top node representing the module 'App.js'. 'App.js' has three children: 'Copyright.js', 'FancyText.js', and 'InspirationGenerator.js'. 'InspirationGenerator.js' has two children: 'FancyText.js' and 'inspirations.js'. The nodes under and including 'InspirationGenerator.js' have a yellow background color to signify that this sub-graph is client-rendered due to the 'use client' directive in 'InspirationGenerator.js'.">
+`'use client'` segments the module dependency tree of the React Server Components app to mark `InspirationGenerator.js` and any of its dependencies as client-rendered.
+</Diagram>
+
+In the module dependency tree of the example app, the `'use client'` directive in `InspirationGenerator.js` marks that module and all of its transitive dependencies as client modules. It creates a subtree of client modules with `InspirationGenerator.js` as the root.
+
+Irregardless if a module exports a component, like `inspirations.js`, if a module is part of that client subtree, then it will be evaluated on the client.
```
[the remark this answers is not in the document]
lunaleaps: I think I'll just remove it as I've clarified earlier that code and component are both marked as client evaluated.

public/images/docs/diagrams/use_client_render_tree.png
lunaleaps: Yea I chose this example to highlight client components rendering server components and to emphasize that `use client` isn't a boundary on the render tree but the module dependency tree

src/content/reference/react/use-client.md
```diff
@@ -23,37 +23,364 @@ canary: true
...
+	return <p>My Component</p>
+}
+```
+
+2. A "component" can also refer to a **component callsite** of its definition.
+```js
+import MyComponent from './MyComponent';
+
+function App() {
+	// This is a usage of a component
```
lunaleaps, line 193:
```suggestion
	// This is a callsite of a component
```

src/content/reference/react/use-client.md
```diff
@@ -23,37 +23,364 @@ canary: true
...
+* If the component is imported and called in a Server Component, then the component callsite is a Server Component.
+
+
+<Diagram name="use_client_render_tree" height={150} width={450} alt="A tree graph where each node represents a component and its children as child components. The top-level node is labelled 'App' and it has two child components 'InspirationGenerator' and 'FancyText'. 'InspirationGenerator' has two child components, 'FancyText' and 'Copyright'. Both 'InspirationGenerator' and its child component 'FancyText' are marked to be client-rendered.">A render tree illustrates component callsites.</Diagram>
+
+Back to the question of `FancyText`, we see that the component definition does _not_ have a `'use client'` directive and it has two callsites.
+
+The callsite of `FancyText` as a child of `App`, marks that usage as a Server Component as root components are always Server Components. When `FancyText` is imported and called under `InspirationGenerator`, that callsite of `FancyText` is a Client Component as `InspirationGenerator` is a Client Component since it contains a `'use client'` directive.
```
lunaleaps, line 210:
```suggestion
The callsite of `FancyText` as a child of `App`, marks that usage as a Server Component as root components are always Server Components. When `FancyText` is imported and called under `InspirationGenerator`, that callsite of `FancyText` is a Client Component as `InspirationGenerator` contains a `'use client'` directive.
```

src/content/reference/react/use-client.md
```diff
@@ -23,37 +23,364 @@ canary: true
...
+`'use client'` defines the boundary between server and client code on the module dependency tree.
+</Diagram>
+
+In the module dependency tree, we see that `App.js` imports and calls `Copyright` from the `Copyright.js` module. As `Copyright.js` does not contain a `'use client'` directive, the component callsite is rendered on the server. `App` is rendered on the server as it is the root component.
+
+Client Components can render Server Components because you can pass JSX as props. In this case, `InspirationGenerator` takes [JSX as children props](/learn/passing-props-to-a-component#passing-jsx-as-children) and receives `Copyright` as a `children` prop. However, the `InspirationGenerator` module never directly imports the `Copyright` module nor calls the component, all of that is done by `App`.
```
lunaleaps, line 230:
```suggestion
Client Components can render Server Components because you can pass JSX as props. In this case, `InspirationGenerator` receives `Copyright` as [children](/learn/passing-props-to-a-component#passing-jsx-as-children). However, the `InspirationGenerator` module never directly imports the `Copyright` module nor calls the component, all of that is done by `App`.
```

src/content/reference/react/use-client.md
```diff
@@ -23,37 +23,364 @@ canary: true
...
+	* [TypedArray](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray)
+* Plain [objects](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object), those created with [object initializers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer), with serializable properties
+* Client or Server Components
+* [Promises](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise)
+
+Notably, these are not supported:
+* [Functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function) that are not exported from client-marked modules or marked with `'use server'`
+* [Classes](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Objects/Classes_in_JavaScript)
+* Objects that are not an instance of [Object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object), or [null-prototype objects](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object#null-prototype_objects)
```
lunaleaps, line 282:
```suggestion
* Objects that are not an instance of [Object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object), or are [null-prototype objects](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object#null-prototype_objects)
```

src/content/reference/react/use-client.md
```diff
@@ -23,37 +23,364 @@ canary: true
...
 ```
 
-When a file marked `'use client'` is imported from a server component, [compatible bundlers](/learn/start-a-new-react-project#bleeding-edge-react-frameworks) will treat the import as the "cut-off point" between server-only code and client code. Components at or below this point in the module graph can use client-only React features like [`useState`](/reference/react/useState).
+When a file marked with `'use client'` is imported from a Server Component, [compatible bundlers](/learn/start-a-new-react-project#bleeding-edge-react-frameworks) will treat the module import as a boundary between server-run and client-run code.
+
+As a dependency of `RichTextEditor`,  `formatDate` and `Link` will also be evaluated on the client regardless if their modules contain a `'use client'` directive.
+
+<DeepDive>
+
+#### Client and Server Rendering with React Server Components {/*client-and-server-rendering-with-react-server-components*/}
```
smikitky, line 50:
```suggestion
#### Client and server rendering with React Server Components {/*client-and-server-rendering-with-react-server-components*/}
```

Titles of Deep Dive sections are not capitalized.

src/content/reference/react/use-client.md
```diff
@@ -23,37 +23,364 @@ canary: true
 
 ### `'use client'` {/*use-client*/}
 
-Add `'use client';` at the very top of a file to mark that the file (including any child components it uses) executes on the client, regardless of where it's imported.
+Add `'use client'` at the top of a file to mark the module and its transitive dependencies as client code.
```
davidmccabe, line 26: This use of "transitive dependencies" is correct but my brain erroneously interpreted it as "transitive dependents" instead. I wonder if we could say something like "and any other modules it imports".

src/content/reference/react/use-client.md
```diff
@@ -23,37 +23,364 @@ canary: true
...
+  return
+    <Link />
+    // ...
+}
 ```
 
-When a file marked `'use client'` is imported from a server component, [compatible bundlers](/learn/start-a-new-react-project#bleeding-edge-react-frameworks) will treat the import as the "cut-off point" between server-only code and client code. Components at or below this point in the module graph can use client-only React features like [`useState`](/reference/react/useState).
+When a file marked with `'use client'` is imported from a Server Component, [compatible bundlers](/learn/start-a-new-react-project#bleeding-edge-react-frameworks) will treat the module import as a boundary between server-run and client-run code.
+
+As a dependency of `RichTextEditor`,  `formatDate` and `Link` will also be evaluated on the client regardless if their modules contain a `'use client'` directive.
```
davidmccabe, line 46: regardless if -> regardless of whether

Possible follow-on sentence: Note that the same module may be evaluated on the server when imported from server code and on the client when imported from client code.

src/content/reference/react/use-client.md
```diff
@@ -23,37 +23,364 @@ canary: true
...
+
+<DeepDive>
+
+#### Client and server rendering with React Server Components {/*client-and-server-rendering-with-react-server-components*/}
+
+React's full-stack architecture vision is realized through the [React Server Components (RSC) specification](https://github.com/reactjs/rfcs/blob/main/text/0188-server-components.md).
+
+React Server Components lets you coordinate which components of your [render tree](/learn/understanding-your-ui-as-a-tree#the-render-tree) are rendered on the server or on the client. Server rendering is traditionally more performant and allows access to the back-end, while client rendering is necessary for adding interactivity into your components.
```
davidmccabe, line 54: "traditionally" may not be the right word choice here (as opposed to actually, in the present day?) Perhaps you meant "often" or "generally".

src/content/reference/react/use-client.md
```diff
@@ -23,37 +23,364 @@ canary: true
...
+
+```js InspirationGenerator.js
+'use client';
+
+import * as React from 'react';
+import inspirations from './inspirations';
+import FancyText from './FancyText';
+
+export default function InspirationGenerator({children}) {
+  const [index, setIndex] = React.useState(0);
```
davidmccabe, line 104: We don't usually use 'React.---' for hooks but import them into the module namespace.

src/content/reference/react/use-client.md
```diff
@@ -23,37 +23,364 @@ canary: true
...
+1. A "component" can refer to a **component definition**. In most cases this will be a function.
+
+```js
+// This is a definition of a component
+function MyComponent() {
+	return <p>My Component</p>
+}
+```
+
+2. A "component" can also refer to a **component callsite** of its definition.
```
davidmccabe, line 188: A "component" can also refer to a **component instance** created by using the component as an element within some other component.

davidmccabe: I don't think we use the word "callsite" elsewhere. Technically speaking they aren't callsites as the component function is not called. They are technically elements but might more clearly be referred to as instances.

src/content/reference/react/use-client.md
```diff
@@ -23,37 +23,364 @@ canary: true
...
+	return <MyComponent />;
+}
+```
+
+Often, the imprecision is not important when explaining concepts, but in this case it is.
+
+When we talk about Server or Client Components, we are referring to component callsites.
+
+* If the component is defined in a module with a `'use client'` directive, or the component is imported and called in a Client Component, then the component callsite is a Client Component.
+* If the component is imported and called in a Server Component, then the component callsite is a Server Component.
```
davidmccabe, line 203: * If the component that is _not_ defined with `use client` is imported and called in a Server Component, then that component instance is a Server Component.

src/content/reference/react/use-client.md
```diff
@@ -23,37 +23,364 @@ canary: true
...
-* Directives like `'use client'` must be at the very beginning of a file, above any imports or other code (comments above directives are OK). They must be written with single or double quotes, not backticks. (The `'use xyz'` directive format somewhat resembles the `useXyz()` Hook naming convention, but the similarity is coincidental.)
+* `'use client'` must be at the very beginning of a file, above any imports or other code (comments are OK). They must be written with single or double quotes, but not backticks.
+* When a `'use client'` module is imported from another client-rendered module, the directive has no effect.
+* When a component module contains a `'use client'` directive, any usage of that component is guaranteed to be a Client Component. However, a component can still be evaluated on the client even if it does not have a `'use client'` directive.
+	* A component usage is considered a Client Component if it is defined in module with `'use client'` directive or it is a transitive dependency of a module that contains a `'use client'` directive. Otherwise, it is a Server Component.
```
davidmccabe, line 290: or it is -> or when it is

and clarify that just importing it from one client component doesn't taint it into running on the client when some other server component imports it.

davidmccabe (commented):
This is a great doc. Some suggestions. I'll get to the Usage section later this afternoon.

davidmccabe:
I think reading through the Usage section here has swayed me away from capitalizing Client Component.

src/content/reference/react/use-client.md
```diff
@@ -23,37 +23,364 @@ canary: true
...
+import {getCounterValueFromFile} from 'fs-utils';
+import Counter from './Counter';
+
+export default async function CounterContainer() {
+  const initialValue = await getCounterValueFromFile();
+  return <Counter initialValue={initialValue} />
+}
+```
+
+For example, the parent component of the Client Component `CounterContainer` does not require `'use client'` as it is not interactive and does not use state. To be more specific, `CounterContainer` must be a Server Component as it reads from the local file system on the server. Notably, Server Components support asynchronous rendering, which is not supported for Client Components.
```
davidmccabe, line 334: To be more specific, -->  In addition, / Moreover,  —— this is just a further fact rather than an increase of specificity.

Notably, Server Components support asynchronous rendering, which is not supported for Client Components. ---> (This is possible because Server Components, unlike Client Components, can be async functions.)

I would say "can be async functions" because this is more specific and actionable than "support asynchronous rendering" which is an abstract concept that could refer to concurrent rendering or who knows whatw.

src/content/reference/react/use-client.md
```diff
@@ -23,37 +23,364 @@ canary: true
...
+
+```js
+export default function FancyText({title, text}) {
+  return title
+    ? <h1 className='fancy title'>{text}</h1>
+    : <h3 className='fancy cursive'>{text}</h3>
+}
+```
+
+In this case, it is discouraged to use the `'use client'` directive as it pre-maturely forces all component usages of `FancyText` to be rendered on the client, which comes at a performance cost. As demonstrated in the earlier Inspirations app example, `FancyText` is used as both a Server or Client Component, depending on where it is imported and used.
```
davidmccabe, line 346:
```suggestion
In this case, it is discouraged to use the `'use client'` directive as it prematurely forces all component usages of `FancyText` to be rendered on the client, which comes at a performance cost. As demonstrated in the earlier Inspirations app example, `FancyText` is used as both a Server or Client Component, depending on where it is imported and used.
```

src/content/reference/react/use-client.md
lunaleaps: I think that weakens the definition though because 'any other module it imports' may be only interpreted as one level

src/content/reference/react/use-client.md
lunaleaps: I've heard conflicting thoughts on this from @eps1lon as I originally had them as instances. There was mention it may be confused with the runtime instance.
Before callsite, I used "use", like "component usages"

src/content/reference/react/use-client.md
lunaleaps: Maybe it should just be: "Otherwise, it is a Server Component"

src/content/reference/react/use-client.md
lunaleaps: I'll go back to usage for now

src/content/reference/react/use-client.md
```diff
@@ -23,37 +23,364 @@ canary: true
...
+
+### Using React libraries {/*using-react-libraries*/}
+
+Often in a React app, you'll leverage third-party libraries to handle common UI patterns or logic.
+
+These libraries may rely on component Hooks or client APIs. In these cases, you'll need to ensure you're using these libraries in Client Components. Depending on the nature of these libraries, this may mean adding a `'use client'` near the top of your module dependency tree – marking the majority of your React app as client-rendered.
+
+Libraries that use any of the following React APIs must be marked as client-run:
+* [createContext](/reference/react/createContext)
+* [`react`](/reference/react) and [`react-dom`](/reference/react-dom/hooks) Hooks
```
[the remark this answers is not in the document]
lunaleaps: I forget, I thought `use` wasn't recommended on the server? cc @mattcarrollcode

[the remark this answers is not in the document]
sophiebits: use and useId can be used sparingly in server components. But async/await is better if all you want to do is unwrap a promise. It's really only if you need useId then you can't do async/await, and then you need use() if you want to "await" in the same component.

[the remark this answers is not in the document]
lunaleaps: Updated to include those exceptions

src/content/reference/react/use-client.md
lunaleaps: I've updated to that, as I use similar definition in the caveat

PR #6364, opened 2023-10-20
Fix: Common components typo in example event handler, `onScroll` → `onWheel` 

insanj:
Very small typo fix visible here:
Screenshot 2023-10-20 at 12 35 00 PM


sophiebits:
thanks!

PR #6366, opened 2023-10-20
Remove unsupported ga-lite and update existing calls to GA4

kmiddleton14:
We recently updated our GA tag to include GA4 in [this PR](https://github.com/reactjs/react.dev/pull/6350)

These changes removes the no longer used `ga-lite`, and updates it to use `gtag` instead since `ga-lite` is no longer supported. 
I've also updated `NEXT_PUBLIC_GA_TRACKING_ID` to the new ID, and removed it from `src/siteConfig.js` 

## Test plan
- yarn check-all passes
- Verified site visually looks ok
- Verified locally in the network payload data is showing for all the events for site navigation, feedback button, and termination event

Screenshot 2023-10-20 at 4 49 19 PM
Screenshot 2023-10-20 at 4 49 06 PM
Screenshot 2023-10-20 at 4 20 05 PM


src/pages/_app.tsx
```diff
@@ -5,21 +5,20 @@
...
 
 if (typeof window !== 'undefined') {
-  if (process.env.NODE_ENV === 'production') {
-    ga('create', process.env.NEXT_PUBLIC_GA_TRACKING_ID, 'auto');
-    ga('send', 'pageview');
-  }
   const terminationEvent = 'onpagehide' in window ? 'pagehide' : 'unload';
   window.addEventListener(terminationEvent, function () {
-    ga('send', 'timing', 'JS Dependencies', 'unload');
+    // @ts-ignore
```
mattcarrollcode, line 17: can you install the gtag typescript typtes instead of a `ts-ignore` here? I think `yarn add @types/gtag.js --dev` should do the trick.

src/pages/_app.tsx
```diff
@@ -5,21 +5,20 @@
...
 import {useRouter} from 'next/router';
-import {ga} from '../utils/analytics';
 
 import '@docsearch/css';
 import '../styles/algolia.css';
 import '../styles/index.css';
 import '../styles/sandpack.css';
 
 if (typeof window !== 'undefined') {
-  if (process.env.NODE_ENV === 'production') {
```
mattcarrollcode, line 16: Why was this check removed? Does this mean that PR preview deploys and locally hosted environments will now send GA4 events?

mattcarrollcode (approved):

src/pages/_app.tsx
kmiddleton14: This check was for the two deprecated ga calls that were also removed so it's no longer needed

src/pages/_app.tsx
kmiddleton14: Oh nice! I can follow up and add this in a separate PR

PR #6370, opened 2023-10-23
refac: Search component type improvements

ssi02014:
Improved the type of `Hit` type in the Search component.

Instead of being of type `any`, it is now of type `props` of the actual `hitComponent` of `DocSearchProps`. 
https://github.com/algolia/docsearch/blob/main/packages/docsearch-react/src/DocSearch.tsx#L26
```ts
export interface DocSearchProps {
    // ...
    hitComponent?(props: {
        hit: InternalDocSearchHit | StoredDocSearchHit;
        children: React.ReactNode;
    }): JSX.Element;
}
```

<br />

The type of `url` is `string`.
https://github.com/algolia/docsearch/blob/main/packages/docsearch-react/src/types/DocSearchHit.ts#L45
```ts
export declare type DocSearchHit = {
    // ...
    url: string;
};
```

`hit.url.replace()`, `hit.url` will return the same result. However, `replace()` throws a type error because it actually needs to take `2 arguments`, so only hit.url is sufficient.

```ts
interface String {
  // ...
  replace(searchValue: { [Symbol.replace](string: string, replaceValue: string): string; }, replaceValue: string): string;
}

```

스크린샷 2023-10-23 오후 10 56 27

### DIFF
--- a/src/components/Search.tsx
+++ b/src/components/Search.tsx
@@ -2,6 +2,7 @@
  * Copyright (c) Facebook, Inc. and its affiliates.
  */
 
+import {DocSearchProps} from '@docsearch/react';
 import Head from 'next/head';
 import Link from 'next/link';
 import Router from 'next/router';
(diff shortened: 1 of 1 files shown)


src/components/Search.tsx
```diff
@@ -20,8 +21,10 @@ export interface SearchProps {
   onClose: () => void;
 }
 
-function Hit({hit, children}: any) {
-  return <Link href={hit.url.replace()}>{children}</Link>;
+type HitProps = Parameters<NonNullable<DocSearchProps['hitComponent']>>[0];
```
ssi02014, line 24:
```ts
type HitProps = {
    hit: InternalDocSearchHit | StoredDocSearchHit;
    children: React.ReactNode;
}
```

PR #6377, opened 2023-10-24
Add `useOptimistic Hook` API reference documentation

kmiddleton14:
Add initial reference documentation for the `useOptimistic` Hook

[Preview link](https://react-538fantv1-fbopensource.vercel.app/reference/react/useOptimistic)

Persistent Preview: 
https://react-dev-git-fork-kmiddleton14-use-optimistic-fbopensource.vercel.app/reference/react/useOptimistic

### DIFF
--- a/src/content/reference/react/useOptimistic.md
+++ b/src/content/reference/react/useOptimistic.md
@@ -29,19 +29,32 @@ The `useOptimistic` Hook is currently only available in React's canary and exper
 
 `useOptimistic` is a React Hook that lets you optimistically update the UI.
 
-🚧
+```js
+import { useOptimistic } from 'react';
+
+function AppContainer() {
+  const [optimistic, addOptimistic] = useOptimistic(
+    initialState,
+    (currentState, newState) => {
+      // merge and return new state
+    }
+  );
+}
+```
 
 [See more examples below.](#usage)
 
 #### Parameters {/*parameters*/}
 
-* `initialState`: 🚧
-* `updateFn`: 🚧
+* `initialState`: `initialState` is the source of truth so if the `initialState` is ever changed and we get a new value in from the server, the return value `optimistic` will be set the that too since it will always treat `initialState` as the final source of truth.
+* `updateFn`: `updateFn` is the mutation that will occur to `optimistic` when `addOptimistic` is called.
+
 
 #### Returns {/*returns*/}
 
-* `optimistic`: 🚧
-* `addOptimistic`: 🚧
+* `optimistic`: `optimistic` is the optimistic state, it will default to what `initialState` is.
+* `addOptimistic`: `addOptimistic` is the dispatching function to call that will run what we define in `updateFn`.
+
 
 #### Caveats {/*caveats*/}
 
@@ -114,7 +127,7 @@ export default function App() {
 
 ### Optimistically updating forms {/*optimistically-updating-with-forms*/}
 
-The `useOptimistic` Hook provides a way to optimistically update the user interface before a background operation, like a network request, completes. In the context of forms, this technique helps to make apps feel more responsive. When a user submits a form, instead of waiting for the server's response to reflect the changes, the interface is immediately updated with the expected outcome. 
+The `useOptimistic` Hook provides a way to optimistically update the user interface before a background operation, like a network request, completes. In the context of forms, this technique helps to make apps feel more responsive. When a user submits a form, instead of waiting for the server's response to reflect the changes, the interface is immediately updated with the expected outcome.
 
 For example, when a user types a message into the form and hits the "Send" button, the `useOptimistic` Hook allows the message to immediately appear in the list with a "Sending..." label, even before the message is actually sent to a server. This "optimistic" approach gives the impression of speed and responsiveness. The form then attempts to truly send the message in the background. Once the server confirms the message has been received, the "Sending..." label is removed.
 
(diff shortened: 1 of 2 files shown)


mattcarrollcode:
Notes from in person code review:

Things to fix now
* document the signature of `updateFn`
* Update code sample to import `useOptimisitic` instead of `experimental_useOptimisitic`
* Mismatch between the name of the second argument to `useOptimisitic` (`addOptimistic` and `updateFn`)

Things that can wait until later
* Cavets: maybe updateFn has some limitations we should document e.g. no side effects
* Maybe add a description of what "optimisitic" means in the reference section?
* Add example of what happens when the network request fails
* Investigate what happens if you stack useOPtimisitic calls on top of each other?

src/content/reference/react/useOptimistic.md
```diff
@@ -0,0 +1,212 @@
...
+
+<InlineToc />
+
+---
+
+## Reference {/*reference*/}
+
+### `useOptimistic(initialState, updateFn)` {/*use*/}
+
+`useOptimistic` is a React Hook that lets you optimistically update the UI.
```
davidmccabe, line 30: A few more sentences to outline how it accomplishes this, something like:

It accepts some state as an argument and returns a copy of that state that can be different during the duration of an async action such as a network request. You provide a function that takes the current state and the input to the action, and returns the _optimistic_ state to be used while the action is pending.

This state is called the "optimistic" state because it is usually used to immediately present the user with the result of performing an action, even though the action actually takes time to complete.

src/content/reference/react/useOptimistic.md
```diff
@@ -0,0 +1,212 @@
...
+    }
+  );
+}
+```
+
+[See more examples below.](#usage)
+
+#### Parameters {/*parameters*/}
+
+* `initialState`: `initialState` is the source of truth so if the `initialState` is ever changed and we get a new value in from the server, the return value `optimistic` will be set the that too since it will always treat `initialState` as the final source of truth.
```
davidmccabe, line 49: `baseState`: the value to be returned initially and whenever no action is pending.

src/content/reference/react/useOptimistic.md
```diff
@@ -0,0 +1,212 @@
...
+* `updateFn`: `updateFn` is the mutation that will occur to `optimistic` when `addOptimistic` is called.
```
davidmccabe, line 50: `reducer(currentState, input)`: a function that takes the current state and the input passed to `dispatch` and returns the resulting optimistic state.

src/content/reference/react/useOptimistic.md
```diff
@@ -0,0 +1,212 @@
...
+
+
+#### Returns {/*returns*/}
+
+* `optimistic`: `optimistic` is the optimistic state, it will default to what `initialState` is.
```
davidmccabe, line 55: `optimisticState`: The resulting optimistic state. It is equal to `baseState` unless an action is pending, in which case it is equal to the value returned by `reducer`.

src/content/reference/react/useOptimistic.md
```diff
@@ -0,0 +1,212 @@
...
+* `addOptimistic`: `addOptimistic` is the dispatching function to call that will run what we define in `updateFn`.
```
davidmccabe, line 56: `dispatch(input)`: signals the beginning of an action. The input you provide will be passed to `reducer` to help it compute the optimistic state to be used while the action is pending.

davidmccabe (commented):
Some suggestions for one way the reference section this might be reworded.

src/content/reference/react/useOptimistic.md
```diff
@@ -0,0 +1,202 @@
...
+  );
+}
+```
+
+[See more examples below.](#usage)
+
+#### Parameters {/*parameters*/}
+
+* `state`: `state` is the source of truth so if the `state` is ever changed and we get a new value in from the server, the return value `optimisticState` will be set the that too since it will always treat `state` as the final source of truth.
+* `updateFn`: `updateFn` is the mutation that will occur to `optimisticState` when `addOptimistic` is called. `updateFn` takes in two parameters. The `currentState` and the `newState`. The return value will be the merged value of the `currentState` and `newState`.
```
lunaleaps, line 51: Hmm, I'm still kind of confused by `newState` -- what about `optimisticValue` instead? State implies that it is the state object, but when I see the messages example (the last one) your updateFn is
```
(state, newMessage) => [
      ...state,
      {
        text: newMessage,
        sending: true
      }
    ]
```

src/content/reference/react/useOptimistic.md
lunaleaps: It is kinda redundant to say the same thing as the top, I think this place could be a better place to expand on what the hook does re Dave's point.

src/content/reference/react/useOptimistic.md
```diff
@@ -0,0 +1,202 @@
...
+
+
+#### Returns {/*returns*/}
+
+* `optimisticState`: `optimisticState` is the optimistic state, it will default to what `state` is.
+* `addOptimistic`: `addOptimistic` is the dispatching function to call that will run what we define in `updateFn`.
```
lunaleaps, line 57:
```suggestion
* `addOptimistic`: `addOptimistic` is the dispatching function to call when you have an optimistic update. It takes one argument, `optimisticValue`, of any type and will call the `updateFn` with `state` and `optimisticValue`.
```

Alternatively we could flush this section out like we do in `setState` like @mattcarrollcode mentioned: https://react-dev-git-fork-kmiddleton14-use-optimistic-fbopensource.vercel.app/reference/react/useState#setstate -- like how it creates another type signature explanation for the setter

src/content/reference/react/useOptimistic.md
```diff
@@ -0,0 +1,202 @@
...
+    (likes) => likes + 1
+  );
+
+  async function addLike() {
+    addOptimisticLike();
+    await updateLikes(likes);
+  }
+
+  return (
+    <button disabled={disabled} onClick={addLike}>
```
lunaleaps, line 82: It looks like we need to wrap this in a transition, that might actually be a good caveat
CleanShot 2023-10-25 at 18 14 58@2x

src/content/reference/react/useOptimistic.md
```diff
@@ -0,0 +1,202 @@
...
+  "devDependencies": {}
+}
+```
+
+</Sandpack>
+
+
+### Optimistically updating forms {/*optimistically-updating-with-forms*/}
+
+The `useOptimistic` Hook provides a way to optimistically update the user interface before a background operation, like a network request, completes. In the context of forms, this technique helps to make apps feel more responsive. When a user submits a form, instead of waiting for the server's response to reflect the changes, the interface is immediately updated with the expected outcome.
```
lunaleaps, line 126: I think this has been copied over from the form docs, which is fine, but some of the points mentioned here are generally about `useOptimistic` and might be better shared on the first example as it's kinda strange to be so verbose on the second but have no description for the first

lunaleaps (approved):
Will approve to unblock, not that you need that anyway to merge 😉, but I think the 2 priority bits for me are:

1. the precision of the term `newState` in the `updateFn` and also adding more detail on the signature of `updateFn`
2. errors on the like example
3.

src/content/reference/react/useOptimistic.md
kmiddleton14: That's a good idea and one I can add in the second iteration

src/content/reference/react/useOptimistic.md
kmiddleton14: I've updated it to `state` , but will revisit naming it to baseState

src/content/reference/react/useOptimistic.md
kmiddleton14: I'll remove this example for now, and update it with the caveat and transition

src/content/reference/react/useOptimistic.md
kmiddleton14: Will update with the description

PR #6378, opened 2023-10-24
Adds Reference Overview Page

brianjo:
This draft PR creates a home page for the Reference section of the React documentation. The purpose of this page is to provide a scannable overview of the documentation and make it easier for readers to get to the information they are looking for. The key feature is that it allows for the hooks section to be closed by default so the rest of the documentation is accessible from the Reference landing page. 

The PR does the following. 
* Creates a new overview page, which was renamed to index.md
* Renames the current index.md file to hooks.md
* Updates the left navigation to include the new page at the top. 

Screenshot 2023-10-24 at 4 34 52 PM

### DIFF
--- a/src/sidebarReference.json
+++ b/src/sidebarReference.json
@@ -7,8 +7,13 @@
       "sectionHeader": "react@18.2.0"
     },
     {
-      "title": "Hooks",
+      "title": "Overview",
       "path": "/reference/react",
+      "canary": true
+    },
+    {
+      "title": "Hooks",
+      "path": "/reference/react/hooks",
       "routes": [
         {
           "title": "use",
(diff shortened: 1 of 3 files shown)


lunaleaps (approved):
lgtm! I tried to see if there was a way for git to mark the move, but it seems to be more Github auto-detection?

src/sidebarReference.json
```diff
@@ -7,8 +7,13 @@
       "sectionHeader": "react@18.2.0"
     },
     {
-      "title": "Hooks",
+      "title": "Overview",
       "path": "/reference/react",
+      "canary": true
```
mattcarrollcode, line 12: change this to remove it

PR #6384, opened 2023-10-26
'use server' fleshing out

lunaleaps:
Preview: https://react-dev-git-use-server-fbopensource.vercel.app/reference/react/use-server

### DIFF
--- a/src/content/reference/react/use-server.md
+++ b/src/content/reference/react/use-server.md
@@ -24,7 +24,7 @@ canary: true
 
 ### `'use server'` {/*use-server*/}
 
-Add `'use server'` at the top of an async function body to mark that that function can be called by the client. We refer to server functions that have been marked with `'use server'` as server actions.
+Add `'use server'` at the top of an async function body to mark the function as callable by the client. We refer to server functions that have been marked with `'use server'` as _server actions_.
 
 ```js
 async function addToCart(data) {
@@ -33,51 +33,224 @@ async function addToCart(data) {
 }
 ```
 
-As a server action, you can pass `addToCart` as a prop to a Client Component to be called on the client.
+As a server action, you can pass `addToCart` as a prop to a Client Component to be invoked on the client.
 
 When calling a server action on the client, it will make a network request to the server that includes a serialized copy of any arguments passed. If the server action returns a value, that value will be serialized and returned to the client.
 
 Instead of individually marking functions with `'use server'`, you can add the directive to the top of a file to mark all exports within that file as server actions.
 
-### Serializable parameters and return values {/*serializable-parameters-and-return-values*/}
+### Serializable arguments and return values {/*serializable-parameters-and-return-values*/}
 
-* 
+As client code calls the server action over the network, any arguments passed will need to be serializable.
 
-### Closures {/*closures*/}
+Here are supported types for server action arguments:
 
-* This section should be about defining Server Actions and what is persisted, briefly that values in function sc
+* Primitives
+	* [string](https://developer.mozilla.org/en-US/docs/Glossary/String)
+	* [number](https://developer.mozilla.org/en-US/docs/Glossary/Number)
+	* [bigint](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt)
+	* [boolean](https://developer.mozilla.org/en-US/docs/Glossary/Boolean)
+	* [undefined](https://developer.mozilla.org/en-US/docs/Glossary/Undefined)
+	* [null](https://developer.mozilla.org/en-US/docs/Glossary/Null)
+	* [symbol](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol), only symbols registered in the global Symbol registry via [`Symbol.for`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/for)
+* Iterables
+	* [String](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String)
+	* [Array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array)
+	* [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map)
+	* [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set)
+	* [TypedArray](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray)
+* [FormData](https://developer.mozilla.org/en-US/docs/Web/API/FormData) instances
+* Plain [objects](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object), those created with [object initializers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer), with serializable properties
+* Functions that are server actions
+* [Promises](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise)
 
-### Security {/*security*/}
+Notably, these are not supported:
+* React Elements, or [JSX](https://react.dev/learn/writing-markup-with-jsx)
+* Functions, this includes component functions or any other function that is not a server action
+* [Classes](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Objects/Classes_in_JavaScript)
+* Objects that are not an instance of [Object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object), or are [null-prototype objects](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object#null-prototype_objects)
+* Symbols not registered globally, ex. `Symbol('my new symbol')`
 
-* Brief point here about tainting values
+
+Supported serializable return values are the same as [serializable props](/reference/react/use-client#passing-props-from-server-to-client-components) for a boundary Client Component.
+
+
+### Security considerations {/*security*/}
+
+Recall that parameters to server actions are fully client-controlled. For security, always treat them as untrusted input, and make sure to validate and escape arguments as appropriate.
+
+In addition, we recommend validating user authorization for any server mutations a server action may perform.
+
+<Wip>
+
+Beyond arguments, server actions also encode closures which are sent to the client and back to when the server action is called. To prevent sending sensitive data, there are experimental taint APIs to prevent unique values and objects from being passed to client code.
+
+See [experimental_taintUniqueValue](/reference/react/experimental_taintUniqueValue) and [experimental_taintObjectReference](/reference/react/experimental_taintObjectReference).
+
+</Wip>
 
 
 #### Caveats {/*caveats*/}
-
-* Remember that parameters to functions marked with `'use server'` are fully client-controlled. For security, always treat them as untrusted input, making sure to validate and escape the arguments as appropriate.
-* To avoid the confusion that might result from mixing client- and server-side code in the same file, `'use server'` can only be used in server-side files; the resulting functions can be passed to client components through props.
+* `'use server'` can only be used in server-side files; the resulting functions can be passed to Client Components through props.
 * Because the underlying network calls are always asynchronous, `'use server'` can be used only on async functions.
-* Directives like `'use server'` must be at the very beginning of their function or file, above any other code including imports (comments above directives are OK). They must be written with single or double quotes, not backticks. (The `'use xyz'` directive format somewhat resembles the `useXyz()` Hook naming convention, but the similarity is coincidental.)
+* server actions must be called in [transitions](). When used in a `<form action>`, form will ensure the server action is called in a transition. Elsewhere, you will need to manually call a server action in a transition.
+* Directives like `'use server'` must be at the very beginning of their function or file, above any other code including imports (comments above directives are OK). They must be written with single or double quotes, not backticks.
 
 ## Usage {/*usage*/}
 
+### Server actions in forms {/*server-actions-in-forms*/}
 
-### Server Actions in forms {/*server-actions-in-forms*/}
+The most common use case of server actions will be calling server functions that mutate data. On the browser, the [HTML form element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form) is the traditional approach for a user to submit a mutation. With React Server Components, React introduces first-class support for server actions in [forms](/reference/react-dom/components/form).
 
-* Server actions in Server Components, wins of progressive enhancement
-* Server action call from Client Component
-* Receiving a return value with `useFormState`
+Here is a form that allows a user to request a username.
 
-### Server Actions outside of forms {/*server-actions-outside-of-forms*/}
+```js
+// UsernameForm.js
+'use client';
 
-* Transitions, must wrap in transition
-* Cannot be progressively enhanced because cant define
+export default UsernameForm({submit}) {
+  return (
+    <form action={submit}>
+      <input type="text" name="username" />
+      <button type="submit">Request</button>
+    </form>
+  );
+}
+```
 
+```js
+// App.js
+import UsernameForm from './UsernameForm';
 
+function requestUsername(formData) {
+  'use server';
+  const username = formData.get('username');
+  // ...
+}
 
-This section is a work in progress. 
+export default App() {
+  <UsernameForm submit={requestUsername}>
+}
+```
 
-This API can be used in any framework that supports React Server Components. You may find additional documentation from them.
-* [Next.js documentation](https://nextjs.org/docs/getting-started/react-essentials)
-* More coming soon
-</Wip>
+In this example `UsernameForm` is a Client Component and receives a server action as the prop `submit`. When a user submits this form, there is a network request to the server function `requestUsername`.
+
+When calling a server action in a `<form action>`, React will supply the form's [FormData](https://developer.mozilla.org/en-US/docs/Web/API/FormData) as the first argument to the server action.
+
+<Note>
+
+We force `UsernameForm` as a Client Component with `'use client'` purely for demonstration. In general, it is not recommended to add the `'use client'` directive if the component does not use [client-only features](reference/react/use-client#usage).
+
+</Note>
+
+#### Progressive Enhancement in `<form>` {/*progressive-enhancement-in-form*/}
+
+In the above example, the `<form>` is rendered in a Client Component. By instead rendering the `<form>` in a Server Component, React can support [progressively enhanced](https://developer.mozilla.org/en-US/docs/Glossary/Progressive_Enhancement) forms.
+
+In this context, we mean that server-rendered forms can provide the baseline experience of submitting a form without JavaScript. This is useful for apps that are server-side rendered (SSR) where HTML scaffolding is sent to the client first, before any code is downloaded to add interactivity.
+
+<DeepDive>
+
+#### What is the difference between SSR and React Server Components (RSC)? {/*what-is-the-difference-between-ssr-and-react-server-components-rsc*/}
+
+Server-side rendering and React Server Components are two features for optimizing performance of React apps and can complement one another.
+
+Server-side rendering (SSR) is the concept of taking a single render pass of a React app to resolve it to its base UI primitives, HTML elements for the web. This HTML is sent down to the client to serve as "scaffolding" while JavaScript is still being downloaded and evaluated.
+
+SRR is an optimization for ["time to paint"](https://developer.mozilla.org/en-US/docs/Glossary/First_paint) but it still requires all your app code to download and be evaluated on the client. The client then needs to [hydrate](/reference/react-dom/client/hydrateRoot#hydrating-server-rendered-html) this scaffolding to add interactivity.
+
+React Server Components (RSC) is the concept of breaking up your React [module dependency tree](/reference/react/use-client#marking-client-components-with-use-client) to determine what components can be rendered on the server and what can be sent to the client. In contrast to SSR, components that are rendered on the server have access to server APIs and can make data-fetches.
+
+These rendered Server Components are streamed down to the client in a special encoding to be translated to HTML or whatever client UI primitive. There is no hydration due to [limitations of Server Components](https://react-dev-git-use-client-fbopensource.vercel.app/reference/react/use-client#limitations).
+
+RSC is an optimization for ["time to interactive"](https://developer.mozilla.org/en-US/docs/Glossary/Time_to_interactive) as it reduces the code that needs to be downloaded and run by the client.
+
+</DeepDive>
+
+Here is the same username request form but rendered as a Server Component.
+
+```js
+// App.js
+
+export default App() {
+  function requestUsername(formData) {
+    'use server';
+    const username = formData.get('username');
+    // ...
+  }
+
+  return (
+    <form action={requestUsername}>
+      <input type="text" name="username" />
+      <button type="submit">Request</button>
+    </form>
+  );
+}
+```
+
+The `<form>` is rendered in a Server Component but the form submission will take place on the client. Before [hydration](/reference/react-dom/client/hydrateRoot#hydrating-server-rendered-html) or if JavaScript is disabled, this form will still allow users to call the server action.
+
+#### Handling return values {/*handling-return-values*/}
+
+You'll notice that in a username request form, there might be the chance that a username is not available. `requestUsername` should tell us if it fails or not. How does the form receive the return value of the server action?
+
+Here is where you can use [`useFormState`](/reference/react-dom/hooks/useFormState), a hook designed for receiving a server action return value that supports progressive enhancement.
+
+```js
+// App.js
+import {useFormState} from 'react';
+
+export default App() {
+  function requestUsername(formData) {
+    'use server';
+    const username = formData.get('username');
+    if (canRequest(username)) {
+      // ...
+      return 'successful';
+    }
+    return 'failed';
+  }
+
+  const [returnValue, action] = useFormState(requestUsername, 'n/a');
+
+  return (
+    <>
+      <form action={requestUsername}>
+        <input type="text" name="username" />
+        <button type="submit">Request</button>
+      </form>
+      <p>Last submission request returned: {returnValue}</p>
+    </>
+  );
+}
+```
+
+### Calling a server action outside of `<form>` {/*calling-a-server-action-outside-of-form*/}
+
+Server actions are exposed server endpoints to client code and can in theory be called anywhere in client code.
+
+```js
+import incrementLike from './actions';
+import {useState, useTransition} from 'react';
+
+function LikeButton() {
+  const [pending, startTransition] = useTransition();
+  const [likeCount, setLikeCount] = useState(0);
+
+  const onClick = () => startTransition(async () => {
+    const currentCount = await incrementLike();
+    setLikeCount(currentCount);
+  });
+
+  return (
+    <>
+      <p>Total Likes: {likeCount}</p>
+      <button onClick={onClick} disabled={pending}>Like</button>;
+    </>
+  );
+}
+```
+
+If you call a server action outside of a [form](/reference/react-dom/components/form), which has first-class support for server actions, you can only do so in a Client Component, which lacks progressive enhancement. In addition, you'll need to manually call the server action in a transition `<form action>` takes care of this for you.
+
+To read a server action return value, you'll need to `await` the promise returned.
--- a/src/content/reference/react/use-server.md
+++ b/src/content/reference/react/use-server.md
@@ -26,18 +26,39 @@ canary: true
 
 Add `'use server'` at the top of an async function body to mark the function as callable by the client. We call these functions _server actions_.
 
-```js
+```js {2}
 async function addToCart(data) {
   'use server';
   // ...
 }
 ```
 
-Pass `addToCart` as a prop to a Client Component to be invoked on the client.
-
 When calling a server action on the client, it will make a network request to the server that includes a serialized copy of any arguments passed. If the server action returns a value, that value will be serialized and returned to the client.
 
-Instead of individually marking functions with `'use server'`, you can add the directive to the top of a file to mark all exports within that file as server actions.
+Instead of individually marking functions with `'use server'`, you can add the directive to the top of a file to mark all exports within that file as server actions that can be used anywhere, including imported in client code.
+
+#### Caveats {/*caveats*/}
+* `'use server'` must be at the very beginning of their function or module; above any other code including imports (comments above directives are OK). They must be written with single or double quotes, not backticks.
+* `'use server'` can only be used in server-side files. The resulting server actions can be passed to Client Components through props. See supported [types for serialization](#serializable-parameters-and-return-values).
+* To import a server action from [client code](/reference/react/use-client), the directive must be used on a module level.
+* Because the underlying network calls are always asynchronous, `'use server'` can only be used on async functions.
+* Always treat arguments to server actions as untrusted input and authorize any mutations. See [security considerations](#security).
+* Server actions should be called in a [transition](/reference/react/useTransition). Server actions passed to [`<form action>`](/reference/react-dom/components/form#props) or [`formAction`](/reference/react-dom/components/input#props) will automatically be called in a transition.
+* Server actions are designed for mutations that update server-side state; they are not recommended for data fetching. Accordingly, frameworks implementing server actions typically process one action at a time and do not have a way to cache the return value.
+
+### Security considerations {/*security*/}
+
+Arguments to server actions are fully client-controlled. For security, always treat them as untrusted input, and make sure to validate and escape arguments as appropriate.
+
+In any server action, make sure to validate that the logged-in user is allowed to perform that action.
+
+<Wip>
+
+To prevent sending sensitive data from a server action, there are experimental taint APIs to prevent unique values and objects from being passed to client code.
+
+See [experimental_taintUniqueValue](/reference/react/experimental_taintUniqueValue) and [experimental_taintObjectReference](/reference/react/experimental_taintObjectReference).
+
+</Wip>
 
 ### Serializable arguments and return values {/*serializable-parameters-and-return-values*/}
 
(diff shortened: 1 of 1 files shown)


src/content/reference/react/use-server.md
```diff
@@ -24,34 +24,233 @@ canary: true
...
+See [experimental_taintUniqueValue](/reference/react/experimental_taintUniqueValue) and [experimental_taintObjectReference](/reference/react/experimental_taintObjectReference).
+
+</Wip>
+
+
+#### Caveats {/*caveats*/}
+* `'use server'` can only be used in server-side files; the resulting functions can be passed to Client Components through props.
 * Because the underlying network calls are always asynchronous, `'use server'` can be used only on async functions.
-* Directives like `'use server'` must be at the very beginning of their function or file, above any other code including imports (comments above directives are OK). They must be written with single or double quotes, not backticks. (The `'use xyz'` directive format somewhat resembles the `useXyz()` Hook naming convention, but the similarity is coincidental.)
+* server actions must be called in [transitions](). When used in a `<form action>`, form will ensure the server action is called in a transition. Elsewhere, you will need to manually call a server action in a transition.
```
lunaleaps, line 96:
```suggestion
* server actions must be called in [transitions](/reference/react/useTransition). When used in a `<form action>`, form will ensure the server action is called in a transition. Elsewhere, you will need to manually call a server action in a transition.
```

src/content/reference/react/use-server.md
```diff
@@ -24,34 +24,233 @@ canary: true
...
+
+
+Supported serializable return values are the same as [serializable props](/reference/react/use-client#passing-props-from-server-to-client-components) for a boundary Client Component.
+
+
+### Security considerations {/*security*/}
 
-* Remember that parameters to functions marked with `'use server'` are fully client-controlled. For security, always treat them as untrusted input, making sure to validate and escape the arguments as appropriate.
-* To avoid the confusion that might result from mixing client- and server-side code in the same file, `'use server'` can only be used in server-side files; the resulting functions can be passed to client components through props.
+Recall that parameters to server actions are fully client-controlled. For security, always treat them as untrusted input, and make sure to validate and escape arguments as appropriate.
```
TheSavior, line 80: Is this "Recall" necessary? vs just "Parameters to server actions..."

I felt kinda dumb when reading Recall because I didn't see that mentioned before so I went back to look for it. Point gets across without this though

src/content/reference/react/use-server.md
```diff
@@ -24,34 +24,233 @@ canary: true
...
+}
+```
+
+In this example `UsernameForm` is a Client Component and receives a server action as the prop `submit`. When a user submits this form, there is a network request to the server function `requestUsername`.
+
+When calling a server action in a `<form action>`, React will supply the form's [FormData](https://developer.mozilla.org/en-US/docs/Web/API/FormData) as the first argument to the server action.
+
+<Note>
+
+We force `UsernameForm` as a Client Component with `'use client'` purely for demonstration. In general, it is not recommended to add the `'use client'` directive if the component does not use [client-only features](reference/react/use-client#usage).
```
TheSavior, line 142: > In general, it is not recommended to add the `'use client'` directive if the component does not use [client-only features](reference/react/use-client#usage).

How do we feel about this sentence?

Do we want to make such a strong recommendation or does that feed the narrative that client components are bad? 

Could say something like "UsernameForm requires no client-only features, so this example would also work without "use client""

Alternatively, should we change the example to not need the note? Like, should we just remove use client? A form doesn't need that. Or if you want the use client example, should it use state or something?

TheSavior: Ah, I see your next line builds on this being marked as a client component. So maybe use state somehow?

brainstorming an arbitrary idea:

```
export default UsernameForm({submit}) {
  const [username, setUsername] = useState('');
  return (
    <form action={submit}>
      <input type="text" name="username" value={username} onChange={(e) => setUsername(e.value)} />
      Username is {username.length} characters long!
      <button type="submit">Request</button>
    </form>
  );
}
```

src/content/reference/react/use-server.md
```diff
@@ -24,34 +24,233 @@ canary: true
...
+
+<DeepDive>
+
+#### What is the difference between SSR and React Server Components (RSC)? {/*what-is-the-difference-between-ssr-and-react-server-components-rsc*/}
+
+Server-side rendering and React Server Components are two features for optimizing performance of React apps and can complement one another.
+
+Server-side rendering (SSR) is the concept of taking a single render pass of a React app to resolve it to its base UI primitives, HTML elements for the web. This HTML is sent down to the client to serve as "scaffolding" while JavaScript is still being downloaded and evaluated.
+
+SRR is an optimization for ["time to paint"](https://developer.mozilla.org/en-US/docs/Glossary/First_paint) but it still requires all your app code to download and be evaluated on the client. The client then needs to [hydrate](/reference/react-dom/client/hydrateRoot#hydrating-server-rendered-html) this scaffolding to add interactivity.
```
TheSavior, line 160: SSR?

src/content/reference/react/use-server.md
```diff
@@ -24,34 +24,233 @@ canary: true
...
+
+Here is where you can use [`useFormState`](/reference/react-dom/hooks/useFormState), a hook designed for receiving a server action return value that supports progressive enhancement.
+
+```js
+// App.js
+import {useFormState} from 'react';
+
+export default App() {
+  function requestUsername(formData) {
+    'use server';
```
TheSavior, line 205: This is the first time in this doc we show a server action defined within the component. Should that be an example higher in the docs? The purpose being able to use a component prop or value from within the server action

src/content/reference/react/use-server.md
```diff
@@ -24,34 +24,233 @@ canary: true
...
+      </form>
+      <p>Last submission request returned: {returnValue}</p>
+    </>
+  );
+}
+```
+
+### Calling a server action outside of `<form>` {/*calling-a-server-action-outside-of-form*/}
+
+Server actions are exposed server endpoints to client code and can in theory be called anywhere in client code.
```
TheSavior, line 230: remove "in theory"? They can be called, for real, anywhere in client code, right? :D

src/content/reference/react/use-server.md
```diff
@@ -24,34 +24,233 @@ canary: true
...
+  return (
+    <>
+      <p>Total Likes: {likeCount}</p>
+      <button onClick={onClick} disabled={pending}>Like</button>;
+    </>
+  );
+}
+```
+
+If you call a server action outside of a [form](/reference/react-dom/components/form), which has first-class support for server actions, you can only do so in a Client Component, which lacks progressive enhancement. In addition, you'll need to manually call the server action in a transition `<form action>` takes care of this for you.
```
TheSavior, line 254: > server action in a transition `<form action>` takes care of this for you.

Is there missing punctuation here?

Is it supposed to be:

> server action in a transition. `<form action>` takes care of this for you.

src/content/reference/react/use-server.md
```diff
@@ -24,34 +24,233 @@ canary: true
...
 
-This API can be used in any framework that supports React Server Components. You may find additional documentation from them.
-* [Next.js documentation](https://nextjs.org/docs/getting-started/react-essentials)
-* More coming soon
-</Wip>
\ No newline at end of file
+To read a server action return value, you'll need to `await` the promise returned.
```
TheSavior, line 256: I am pretty sure this is right and reasonable, but I'm curious how to explain why you should be using await here and not using the `use` function. Is it because we are inside an async function already so it is fine, and also we don't want to make this component suspend while waiting for the server response, you handle that with the pending flag instead of suspense?

TheSavior (approved):
Looks great!

Some comments, but not everything needs to be resolved before landing. Could discuss and make more changes to the big things as needed in a follow up.

src/content/reference/react/use-server.md
TheSavior: After seeing the other other example below, *shrug*, maybe the note is fine. lol. Maybe I've gone through the same thought process you did that landed you here 😅

src/content/reference/react/use-server.md
```diff
@@ -24,34 +24,233 @@ canary: true
...
 async function addToCart(data) {
   'use server';
   // ...
 }
-
-// <ProductDetailPage addToCart={addToCart} />
 ```
 
-This function can be passed to the client. When called on the client, it will make a network request to the server that includes a serialized copy of any arguments passed. If the server function returns a value, that value will be serialized and returned to the client.
+As a server action, you can pass `addToCart` as a prop to a Client Component to be invoked on the client.
```
sophiebits, line 36: `As a server action, you` doesn't agree gramatically

src/content/reference/react/use-server.md
```diff
@@ -24,34 +24,233 @@ canary: true
 
 ### `'use server'` {/*use-server*/}
 
-Add `'use server';` at the very top of an async function to mark that the function can be executed by the client.
+Add `'use server'` at the top of an async function body to mark the function as callable by the client. We refer to server functions that have been marked with `'use server'` as _server actions_.
```
sophiebits, line 27: shorter: `We call these functions _server actions_.`?

src/content/reference/react/use-server.md
```diff
@@ -24,34 +24,233 @@ canary: true
...
+	* [boolean](https://developer.mozilla.org/en-US/docs/Glossary/Boolean)
+	* [undefined](https://developer.mozilla.org/en-US/docs/Glossary/Undefined)
+	* [null](https://developer.mozilla.org/en-US/docs/Glossary/Null)
+	* [symbol](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol), only symbols registered in the global Symbol registry via [`Symbol.for`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/for)
+* Iterables
+	* [String](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String)
+	* [Array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array)
+	* [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map)
+	* [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set)
+	* [TypedArray](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray)
```
sophiebits, line 61: ArrayBuffer too

src/content/reference/react/use-server.md
```diff
@@ -24,34 +24,233 @@ canary: true
...
+* [FormData](https://developer.mozilla.org/en-US/docs/Web/API/FormData) instances
+* Plain [objects](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object), those created with [object initializers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer), with serializable properties
+* Functions that are server actions
+* [Promises](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise)
```
sophiebits, line 65: Date should be somewhere in this list (and for use client)

src/content/reference/react/use-server.md
```diff
@@ -24,34 +24,233 @@ canary: true
...
+
+* Primitives
+	* [string](https://developer.mozilla.org/en-US/docs/Glossary/String)
+	* [number](https://developer.mozilla.org/en-US/docs/Glossary/Number)
+	* [bigint](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt)
+	* [boolean](https://developer.mozilla.org/en-US/docs/Glossary/Boolean)
+	* [undefined](https://developer.mozilla.org/en-US/docs/Glossary/Undefined)
+	* [null](https://developer.mozilla.org/en-US/docs/Glossary/Null)
+	* [symbol](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol), only symbols registered in the global Symbol registry via [`Symbol.for`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/for)
+* Iterables
```
sophiebits, line 56: maybe: Iterables containing serializable values? or is that obvious

src/content/reference/react/use-server.md
```diff
@@ -24,34 +24,233 @@ canary: true
...
+	* [String](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String)
+	* [Array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array)
+	* [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map)
+	* [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set)
+	* [TypedArray](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray)
+* [FormData](https://developer.mozilla.org/en-US/docs/Web/API/FormData) instances
+* Plain [objects](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object), those created with [object initializers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer), with serializable properties
```
sophiebits, line 63: would be nice to mention explicitly that circular values are supported and mention the (limited) guarantees on object identity, though I'm not 100% sure on the exact semantics there so maybe for a follow-up

src/content/reference/react/use-server.md
```diff
@@ -24,34 +24,233 @@ canary: true
...
+* Functions that are server actions
+* [Promises](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise)
+
+Notably, these are not supported:
+* React Elements, or [JSX](https://react.dev/learn/writing-markup-with-jsx)
```
sophiebits, line 68: lowercase `element`

src/content/reference/react/use-server.md
```diff
@@ -24,34 +24,233 @@ canary: true
...
+* Functions, this includes component functions or any other function that is not a server action
+* [Classes](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Objects/Classes_in_JavaScript)
+* Objects that are not an instance of [Object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object), or are [null-prototype objects](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object#null-prototype_objects)
```
sophiebits, line 71: Would this be clearer as:

* Instances of any class (other than the built-ins mentioned above)
* Objects with a `null` prototype

src/content/reference/react/use-server.md
```diff
@@ -24,34 +24,233 @@ canary: true
...
 
-* Remember that parameters to functions marked with `'use server'` are fully client-controlled. For security, always treat them as untrusted input, making sure to validate and escape the arguments as appropriate.
-* To avoid the confusion that might result from mixing client- and server-side code in the same file, `'use server'` can only be used in server-side files; the resulting functions can be passed to client components through props.
+Recall that parameters to server actions are fully client-controlled. For security, always treat them as untrusted input, and make sure to validate and escape arguments as appropriate.
+
+In addition, we recommend validating user authorization for any server mutations a server action may perform.
+
+<Wip>
+
+Beyond arguments, server actions also encode closures which are sent to the client and back to when the server action is called. To prevent sending sensitive data, there are experimental taint APIs to prevent unique values and objects from being passed to client code.
```
sophiebits, line 86: a bit hard to follow here but I guess you're planning on expanding? I think we could use a fuller security section (I can help outline if you need)

src/content/reference/react/use-server.md
```diff
@@ -24,34 +24,233 @@ canary: true
...
+Supported serializable return values are the same as [serializable props](/reference/react/use-client#passing-props-from-server-to-client-components) for a boundary Client Component.
+
+
+### Security considerations {/*security*/}
 
-* Remember that parameters to functions marked with `'use server'` are fully client-controlled. For security, always treat them as untrusted input, making sure to validate and escape the arguments as appropriate.
-* To avoid the confusion that might result from mixing client- and server-side code in the same file, `'use server'` can only be used in server-side files; the resulting functions can be passed to client components through props.
+Recall that parameters to server actions are fully client-controlled. For security, always treat them as untrusted input, and make sure to validate and escape arguments as appropriate.
+
+In addition, we recommend validating user authorization for any server mutations a server action may perform.
```
sophiebits, line 82: how about instead:

In any server action, make sure to validate that the logged-in user is allowed to perform that action.

src/content/reference/react/use-server.md
sophiebits: can you try another pass at the wording here? a bit hard to parse and to follow

also would be good to accommodate for component libraries doing the transition automatically in the wording

src/content/reference/react/use-server.md
```diff
@@ -24,34 +24,233 @@ canary: true
...
+    </form>
+  );
+}
+```
+
+```js
+// App.js
+import UsernameForm from './UsernameForm';
+
+function requestUsername(formData) {
```
sophiebits, line 125: async

src/content/reference/react/use-server.md
```diff
@@ -24,34 +24,233 @@ canary: true
...
+
+The most common use case of server actions will be calling server functions that mutate data. On the browser, the [HTML form element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form) is the traditional approach for a user to submit a mutation. With React Server Components, React introduces first-class support for server actions in [forms](/reference/react-dom/components/form).
+
+Here is a form that allows a user to request a username.
+
+```js
+// UsernameForm.js
+'use client';
+
+export default UsernameForm({submit}) {
```
sophiebits, line 111: function

src/content/reference/react/use-server.md
```diff
@@ -24,34 +24,233 @@ canary: true
...
+import UsernameForm from './UsernameForm';
+
+function requestUsername(formData) {
+  'use server';
+  const username = formData.get('username');
+  // ...
+}
+
+export default App() {
+  <UsernameForm submit={requestUsername}>
```
sophiebits, line 132: return? `/>`?

src/content/reference/react/use-server.md
```diff
@@ -24,34 +24,233 @@ canary: true
...
+
+<Note>
+
+We force `UsernameForm` as a Client Component with `'use client'` purely for demonstration. In general, it is not recommended to add the `'use client'` directive if the component does not use [client-only features](reference/react/use-client#usage).
+
+</Note>
+
+#### Progressive Enhancement in `<form>` {/*progressive-enhancement-in-form*/}
+
+In the above example, the `<form>` is rendered in a Client Component. By instead rendering the `<form>` in a Server Component, React can support [progressively enhanced](https://developer.mozilla.org/en-US/docs/Glossary/Progressive_Enhancement) forms.
```
sophiebits, line 148: do you mean "If you change UsernameForm to be a Server Component,"?

src/content/reference/react/use-server.md
```diff
@@ -24,34 +24,233 @@ canary: true
...
+
+In this context, we mean that server-rendered forms can provide the baseline experience of submitting a form without JavaScript. This is useful for apps that are server-side rendered (SSR) where HTML scaffolding is sent to the client first, before any code is downloaded to add interactivity.
+
+<DeepDive>
+
+#### What is the difference between SSR and React Server Components (RSC)? {/*what-is-the-difference-between-ssr-and-react-server-components-rsc*/}
```
sophiebits, line 154: is this the right page for this explanation? doesn't seem actions-specific

src/content/reference/react/use-server.md
```diff
@@ -24,34 +24,233 @@ canary: true
...
+
+SRR is an optimization for ["time to paint"](https://developer.mozilla.org/en-US/docs/Glossary/First_paint) but it still requires all your app code to download and be evaluated on the client. The client then needs to [hydrate](/reference/react-dom/client/hydrateRoot#hydrating-server-rendered-html) this scaffolding to add interactivity.
+
+React Server Components (RSC) is the concept of breaking up your React [module dependency tree](/reference/react/use-client#marking-client-components-with-use-client) to determine what components can be rendered on the server and what can be sent to the client. In contrast to SSR, components that are rendered on the server have access to server APIs and can make data-fetches.
+
+These rendered Server Components are streamed down to the client in a special encoding to be translated to HTML or whatever client UI primitive. There is no hydration due to [limitations of Server Components](https://react-dev-git-use-client-fbopensource.vercel.app/reference/react/use-client#limitations).
```
sophiebits, line 164:
```suggestion
These rendered Server Components are streamed down to the client in a special encoding to be translated to HTML or whatever client UI primitive. There is no hydration due to [limitations of Server Components](/reference/react/use-client#limitations).
```

sophiebits: don't know what you mean by "no hydration" here

src/content/reference/react/use-server.md
```diff
@@ -24,34 +24,233 @@ canary: true
...
+
+RSC is an optimization for ["time to interactive"](https://developer.mozilla.org/en-US/docs/Glossary/Time_to_interactive) as it reduces the code that needs to be downloaded and run by the client.
+
+</DeepDive>
+
+Here is the same username request form but rendered as a Server Component.
```
sophiebits, line 170: would it make sense to just remove the "use client" without combining the files? just mention that they can be combined

sophiebits (commented):
there is one big error in this which is that progressive enhancement 100% works with client components! as long as it's a server action on the form, then SSR of the client component will output the right thing

and useFormState is a client hook! it cannot be imported from a server component. the only hooks that can be used in server components are `use` and `useId` (and technically `useContext` which is being removed, `useCallback` `useDebugValue` `useMemo` which are noops)

src/content/reference/react/use-server.md
lunaleaps: Yea I wasn't sure about this, it needs to go somewhere as I just feel like if I read this I'd be confused. I guess we can punt it to learn docs.

src/content/reference/react/use-server.md
```diff
@@ -24,34 +24,233 @@ canary: true
...
+    'use server';
+    const username = formData.get('username');
+    if (canRequest(username)) {
+      // ...
+      return 'successful';
+    }
+    return 'failed';
+  }
+
+  const [returnValue, action] = useFormState(requestUsername, 'n/a');
```
Fredkiss3, line 214: We can't both define a server action and use `useFormState` in the same component, server actions can only be defined in server components or standalone file marked with `'use server'` directive at the top of the file and `useFormState` can only be used in a client component.

src/content/reference/react/use-server.md
```diff
@@ -24,34 +24,233 @@ canary: true
...
+
+#### Handling return values {/*handling-return-values*/}
+
+You'll notice that in a username request form, there might be the chance that a username is not available. `requestUsername` should tell us if it fails or not. How does the form receive the return value of the server action?
+
+Here is where you can use [`useFormState`](/reference/react-dom/hooks/useFormState), a hook designed for receiving a server action return value that supports progressive enhancement.
+
+```js
+// App.js
+import {useFormState} from 'react';
```
Fredkiss3, line 201: `useFormState` should be imported from `react-dom` not `react`

src/content/reference/react/use-server.md
lunaleaps: I think I'll remove the separation of these examples -- I wanted to keep them separate as I thought there was a difference between server / client rendering the form but I was wrong. I mean yes `useFormState` can only be used in client component, but I don't think it's worth focussing on render environments beyond that.

src/content/reference/react/use-server.md
lunaleaps: Yea I think it needs expanding where closures in server actions are talked about. I'll remove the first sentence for now and just mention the taint APIs

src/content/reference/react/use-server.md
lunaleaps: For sure, I'm not sure I completely follow or know how to express that, so feel free to suggest/add

src/content/reference/react/use-server.md
lunaleaps: I've removed it for now and simplified the example usage -- at some point we should talk about this perhaps when we expand on closures

src/content/reference/react/use-server.md
lunaleaps: Hmm that's a good callout. Yea I don't think `use` should be used in this callback..but I'm going off intuition. @sophiebits is that right? Would we ever want to call a server action in render? I guess they're always called in transitions..

Edit: Trying out `use`, I get the following error

```
Error: Hooks are not supported inside an async component. This error is often caused by accidentally adding `'use client'` to a module that was originally written for the server.
```

lunaleaps:
Updated and ready for another review!

src/content/reference/react/use-server.md
sophiebits: use() is only to unwrap a resource during the render pass (i.e. in a component body or in a setstate updater). Not in nested functions like the imperative one here.

You shouldn't call a server action in render; refactor your code so you don't have to (plus, SSR won't let you).

src/content/reference/react/use-server.md
```diff
@@ -24,34 +24,180 @@ canary: true
...
+* Iterables containing serializable values
+	* [String](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String)
+	* [Array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array)
+	* [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map)
+	* [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set)
+	* [TypedArray](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray)
+* [ArrayBuffer](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer)
+* [Date](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date)
+* [FormData](https://developer.mozilla.org/en-US/docs/Web/API/FormData) instances
+* Plain [objects](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object), those created with [object initializers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer), with serializable properties
```
sophiebits, line 65:
```suggestion
* Plain [objects](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object): those created with [object initializers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer), with serializable properties
```

src/content/reference/react/use-server.md
```diff
@@ -24,34 +24,180 @@ canary: true
...
+* Functions that are server actions
+* [Promises](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise)
+
+Notably, these are not supported:
+* React elements, or [JSX](https://react.dev/learn/writing-markup-with-jsx)
+* Functions, this includes component functions or any other function that is not a server action
```
sophiebits, line 71:
```suggestion
* Functions, including component functions or any other function that is not a server action
```

src/content/reference/react/use-server.md
```diff
@@ -24,34 +24,180 @@ canary: true
...
+<Wip>
+
+To prevent sending sensitive data from a server action, there are experimental taint APIs to prevent unique values and objects from being passed to client code.
+
+See [experimental_taintUniqueValue](/reference/react/experimental_taintUniqueValue) and [experimental_taintObjectReference](/reference/react/experimental_taintObjectReference).
+
+</Wip>
+
+
+#### Caveats {/*caveats*/}
```
sophiebits, line 95: wrong heading level (this shouldn't be nested inside security)

maybe put this at the top and mention in caveats that serializable types and security are factors to think about, detailed below? particularly for security we should at least mention it in caveats I think in case someone only reads the caveats

src/content/reference/react/use-server.md
```diff
@@ -24,34 +24,180 @@ canary: true
...
+* `'use server'` can only be used in server-side files; the resulting functions can be passed to Client Components through props.
 * Because the underlying network calls are always asynchronous, `'use server'` can be used only on async functions.
-* Directives like `'use server'` must be at the very beginning of their function or file, above any other code including imports (comments above directives are OK). They must be written with single or double quotes, not backticks. (The `'use xyz'` directive format somewhat resembles the `useXyz()` Hook naming convention, but the similarity is coincidental.)
+* Server actions must be called in a [transition](/reference/react/useTransition). Server actions passed to [`<form action>`](/reference/react-dom/components/form#props) or [`formAction`](/reference/react-dom/components/input#props) will automatically be called in a transition.
```
sophiebits, line 98:
```suggestion
* Server actions should be called in a [transition](/reference/react/useTransition). Server actions passed to [`<form action>`](/reference/react-dom/components/form#props) or [`formAction`](/reference/react-dom/components/input#props) will automatically be called in a transition.
```

src/content/reference/react/use-server.md
```diff
@@ -24,34 +24,180 @@ canary: true
...
+* Directives like `'use server'` must be at the very beginning of their function or file, above any other code including imports (comments above directives are OK). They must be written with single or double quotes, not backticks.
```
sophiebits, line 99: add something like:

```suggestion
* Server actions are designed for mutations that update server-side state; they are not recommended for data fetching. Accordingly, frameworks implementing server actions typically process one action at a time and do not have a way to caching the return value.
* Directives like `'use server'` must be at the very beginning of their function or file, above any other code including imports (comments above directives are OK). They must be written with single or double quotes, not backticks.
```

src/content/reference/react/use-server.md
```diff
@@ -24,34 +24,180 @@ canary: true
...
+
+In this example `requestUsername` is a server action passed to a `<form>`. When a user submits this form, there is a network request to the server function `requestUsername`. When calling a server action in a form, React will supply the form's <CodeStep step={1}>[FormData](https://developer.mozilla.org/en-US/docs/Web/API/FormData)</CodeStep> as the first argument to the server action.
+
+By passing a server action to the form `action`, React can [progressively enhance](https://developer.mozilla.org/en-US/docs/Glossary/Progressive_Enhancement) the form. This means that forms can be submitted before the JavaScript bundle is loaded.
+
+#### Handling return values in forms {/*handling-return-values*/}
+
+In the username request form, there might be the chance that a username is not available. `requestUsername` should tell us if it fails or not. How can the form receive the return value of the server action?
+
+Here is where you can use [`useFormState`](/reference/react-dom/hooks/useFormState), a Hook designed for receiving a server action return value which supports progressive enhancement.
```
sophiebits, line 134:
```suggestion
To update the UI based on the result of a server action while supporting progressive enhancement, use [`useFormState`](/reference/react-dom/hooks/useFormState).
```

src/content/reference/react/use-server.md
```diff
@@ -24,34 +24,180 @@ canary: true
...
+        <input type="text" name="username" />
+        <button type="submit">Request</button>
+      </form>
+      <p>Last submission request returned: {returnValue}</p>
+    </>
+  );
+}
+```
+
+Note that `useFormState` can only be called in <CodeStep step={1}>[client code](/reference/react/use-client)</CodeStep>.
```
sophiebits, line 172:
```suggestion
Note that like most Hooks, `useFormState` can only be called in <CodeStep step={1}>[client code](/reference/react/use-client)</CodeStep>.
```

src/content/reference/react/use-server.md
```diff
@@ -24,34 +24,180 @@ canary: true
...
+
+### Calling a server action outside of `<form>` {/*calling-a-server-action-outside-of-form*/}
+
+Server actions are exposed server endpoints and can be called anywhere in client code.
+
+If you call a server action outside of a [form](/reference/react-dom/components/form), you'll need to call the server action in a [transition](/reference/react/useTransition). Server actions in forms will automatically handle this for you.
```
sophiebits, line 178:
```suggestion
When using a server action outside of a [form](/reference/react-dom/components/form), call the server action in a [transition](/reference/react/useTransition), which allows you to show a loading indicator and handle unexpected errors. Server actions in forms will automatically handle this for you.
```

src/content/reference/react/use-server.md
```diff
@@ -24,34 +24,180 @@ canary: true
...
+
+```js {8-11}
+import incrementLike from './actions';
+import {useState, useTransition} from 'react';
+
+function LikeButton() {
+  const [pending, startTransition] = useTransition();
```
sophiebits, line 185: our standard name for this is

```suggestion
  const [isPending, startTransition] = useTransition();
```

src/content/reference/react/use-server.md
```diff
@@ -24,34 +24,180 @@ canary: true
...
+  const onClick = () => startTransition(async () => {
+    // `incrementLike` is a server action
+    const currentCount = await incrementLike();
+    setLikeCount(currentCount);
+  });
+
+  return (
+    <>
+      <p>Total Likes: {likeCount}</p>
+      <button onClick={onClick} disabled={pending}>Like</button>;
```
sophiebits, line 197:
```suggestion
      <button onClick={onClick} disabled={isPending}>Like</button>;
```

src/content/reference/react/use-server.md
```diff
@@ -24,34 +24,180 @@ canary: true
...
+
+function LikeButton() {
+  const [pending, startTransition] = useTransition();
+  const [likeCount, setLikeCount] = useState(0);
+
+  const onClick = () => startTransition(async () => {
+    // `incrementLike` is a server action
+    const currentCount = await incrementLike();
+    setLikeCount(currentCount);
+  });
```
sophiebits, line 192: for the docs I think this is easier to follow as a block

```suggestion
  const onClick = () => {
    startTransition(async () => {
      // `incrementLike` is a server action
      const currentCount = await incrementLike();
      setLikeCount(currentCount);
    })
  };
```

(and makes clear the return value of startTransition isn't needed)

src/content/reference/react/use-server.md
```diff
@@ -24,34 +24,180 @@ canary: true
...
+
+Note that `useFormState` can only be called in <CodeStep step={1}>[client code](/reference/react/use-client)</CodeStep>.
+
+### Calling a server action outside of `<form>` {/*calling-a-server-action-outside-of-form*/}
+
+Server actions are exposed server endpoints and can be called anywhere in client code.
+
+If you call a server action outside of a [form](/reference/react-dom/components/form), you'll need to call the server action in a [transition](/reference/react/useTransition). Server actions in forms will automatically handle this for you.
+
+```js {8-11}
```
sophiebits, line 180: if you take my edit below

```suggestion
```js {9-13}
```

sophiebits (approved):
looking great!! thanks for the hard work on this

src/content/reference/react/use-server.md
lunaleaps: Yea after Sophie's review, I've removed the separation since I mistakenly thought progressive enhancement was affected by client vs. server rendering

src/content/reference/react/use-server.md
lunaleaps: I might keep the other sections at the same level as caveats and yea bump up this level and move caveats to the top. I think having the security and serialization info in the reference outline will be helpful.

lunaleaps: Oh actually I'll move caveats under "use server" reference, as it is with all the other reference docs

sophiebits: Sounds good, thanks!

src/content/reference/react/use-server.md
```diff
@@ -203,4 +201,15 @@ function LikeButton() {
 }
 ```
 
+```js
+// actions.js
```
lunaleaps, line 205: I'm fleshing out the action as I want to illustrate that you need the directive on the module level

PR #6395, opened 2023-10-29
Fix missing id properties in docs example

jpayoung:
The [docs page for rendering lists](https://react.dev/learn/rendering-lists#filtering-arrays-of-items) uses a list of people as example. Where `id` properties are first introduced (i.e. the "data is structured"), two of the objects in the list are missing ids. This might cause confusion to new readers.

The absence of or these two objects appears to be a mistake. Where the list of people is next referenced, all 5 objects have ids. They are not added as an exercise or example anywhere in between.

This change adds those two missing ids. Thanks!


rickhanlonii (approved):
Nice catch

PR #6397, opened 2023-10-30
Fix incorrect component naming

wyozi:
`<form>` must be rendered in a client component, since it uses a hook. This was correct in link href (`use client`) but not in link copy.

### DIFF
--- a/src/content/reference/react-dom/components/form.md
+++ b/src/content/reference/react-dom/components/form.md
@@ -322,7 +322,7 @@ export default function Search() {
 
 Displaying a form submission error message before the JavaScript bundle loads for progressive enhancement requires that:
 
-1. `<form>` be rendered by a [Server Component](/reference/react/use-client)
+1. `<form>` be rendered by a [Client Component](/reference/react/use-client)
 1. the function passed to the `<form>`'s `action` prop be a [Server Action](/reference/react/use-server)
 1. the `useFormState` Hook be used to display the error message
 


Huxpro:
This paragraph is indeed confusing. 

By a glance, the example seems to be a Client Component that can fire a Server Action which will eventually return the error message back to the useFormState hook over the wire. Wouldn't that mean the client is only able to display the error message until the client-side JS bundle is loaded and finish hydration? Assuming by "progressive enchantment" we are talking about <form> being actionable when SSRing its belonging Client Component before it's hydrated?

But the title is "Display a form submission error **without JavaScript**" which pointed out that my above understanding is completely wrong...

So let's try a different direction: what if it is a <form> within a Server Component being rendered and sent over wire? In that case, form can still send a Server Action to the server and get the re-rendered result (potentially with an error message) back, without having any client-side JavaScript loaded. This seems work fine, but that seems to break two previous rules:

1. Server Components doesn't run Hook (is useFormState an exception here? A Server Hook?)
2. 'use server' directive being contained in Server module. It's originally declared to annotate Server Action in Client code 

cc @dan @

sophiebits:
You can render a form from a server component or from a client component. useFormState can only be called in a client component, but it does progressively enhance (client components are server-rendered via traditional SSR like react-dom/server, they just can't access server-only resources).

We should not be saying "without JavaScript" because our streaming protocol uses a tiny amount of JS to replace suspense fallbacks with their content and for out-of-order streaming. Instead say that it works "before the JavaScript bundle loads" or "if the JavaScript bundle doesn't load".

Huxpro:
@sophiebits how does the error be displayed if the JS bundle is not loaded? Isn't the useFormState Hook only run after hydration ("progressive enhanced")? The title is "Display a form submission error without JavaScript"

gaearon:
During SSR, the Client component will get its `useFormState` result so it will be in the HTML.

PR #6401, opened 2023-11-02
Edits for new "use client" content

sophiebits:


src/content/reference/react/use-client.md
```diff
@@ -31,20 +31,19 @@ Add `'use client'` at the top of a file to mark the module and its transitive de
...
-export default function RichTextEditor(props) {
-  formatDate();
+export default function RichTextEditor({ timestamp, text }) {
+  const date = formatDate(timestamp);
+  // ...
+  const editButton = <Button />;
   // ...
-  return
-    <Link />
-    // ...
```
sophiebits, line 41: found this syntax awkward. changed to a Button because I couldn't think of a variable name for the link

src/content/reference/react/use-client.md
```diff
@@ -179,7 +178,7 @@ First, let's clarify that the term "component" is not very precise. Here are jus
 ```js
 // This is a definition of a component
 function MyComponent() {
-	return <p>My Component</p>
+  return <p>My Component</p>
```
sophiebits, line 181: changed from tabs (which render as 4 spaces) to spaces (2)

src/content/reference/react/use-client.md
```diff
@@ -215,39 +214,39 @@ This means that the component definition for `FancyText` will both be evaluated
...
 
 ### When to use `'use client'` {/*when-to-use-use-client*/}
 
-With `'use client'`, you can determine what component usages will be Client Components. As Server Components are default, here is a brief overview of the advantages and limitations to Server Components to determine when you need to mark something as client rendered.
+With `'use client'`, you can determine when components are Client Components. As Server Components are default, here is a brief overview of the advantages and limitations to Server Components to determine when you need to mark something as client rendered.
 
 For simplicity, we talk about Server Components, but the same principles apply to all code in your app that is server run.
 
-#### Advantages {/*advantages*/}
+#### Advantages of Server Components {/*advantages*/}
```
sophiebits, line 239: I only made minor edits here but I don't love this section in its current form. I think of this less as advantages/disadvantages and more that in many cases they are for different things! more like

* unique capabilities of server components
* unique capabilities of client components
* when to prefer one or the other if you don't need either one

src/content/reference/react/use-client.md
```diff
@@ -146,12 +145,12 @@ export default [
...
-`'use client'` segments the module dependency tree of the React Server Components app to mark `InspirationGenerator.js` and any of its dependencies as client-rendered.
+`'use client'` segments the module dependency tree of the React Server Components app, marking `InspirationGenerator.js` and all of its dependencies as client-rendered.
 </Diagram>
 
-In the module dependency tree of the example app, the `'use client'` directive in `InspirationGenerator.js` marks that module and all of its transitive dependencies as client modules. It creates a subtree of client modules with `InspirationGenerator.js` as the root.
-
 During render, the framework will server-render the root component and continue through the [render tree](/learn/understanding-your-ui-as-a-tree#the-render-tree), opting-out of evaluating any code imported from client-marked code.
```
sophiebits, line 154: this whole part is still a bit confusing for SSR particularly

src/content/reference/react/use-client.md
```diff
@@ -311,18 +312,18 @@ As `Counter` requires both the `useState` hook and event handlers to increment o
...
 import Counter from './Counter';
 
 export default async function CounterContainer() {
-  const initialValue = await getCounterValueFromFile();
+  const initialValue = await readFile('/path/to/counter_value');
   return <Counter initialValue={initialValue} />
 }
 ```
 
-For example, the parent component of the Client Component `CounterContainer` does not require `'use client'` as it is not interactive and does not use state. In addition, `CounterContainer` must be a Server Component as it reads from the local file system on the server. This is possible because Server Components, unlike Client Components, can be async functions.
```
sophiebits, line 323: we will support async functions on the client so this isn't why

src/content/reference/react/use-client.md
```diff
@@ -332,13 +333,15 @@ export default function FancyText({title, text}) {
 }
 ```
 
-In this case, it is discouraged to use the `'use client'` directive as it prematurely forces all component usages of `FancyText` to be rendered on the client, which comes at a performance cost. As demonstrated in the earlier Inspirations app example, `FancyText` is used as both a Server or Client Component, depending on where it is imported and used.
```
sophiebits, line 335: (first sentence was overly prescriptive)

src/content/reference/react/use-client.md
```diff
@@ -332,13 +333,15 @@ export default function FancyText({title, text}) {
...
+In this case, we don't add the `'use client'` directive, resulting in `FancyText`'s _output_ (rather than its source code) to be sent to the browser when referenced from a Server Component. As demonstrated in the earlier Inspirations app example, `FancyText` is used as both a Server or Client Component, depending on where it is imported and used.
+
+But if `FancyText`'s HTML output was large relative to its source code (including dependencies), it might be more efficient to force it to always be a Client Component. Components that return a long SVG path string are one case where it may be more efficient to force a component to be a Client Component.
 
 ### Using client APIs {/*using-client-apis*/}
 
-Your React app may use client-specific APIs which are dependent on your targeted client. For the browser, some example client APIs include web storage, audio and video manipulation, and device hardware, among [others](https://developer.mozilla.org/en-US/docs/Web/API).
```
sophiebits, line 339: I think you were trying to be agnostic to web vs native here? but I think it was a bit confusing in practice and better to be clearer

(most of our docs are written with web in mind; I think we probably need a bigger rethink terminology-wise before we could truly update them to be agnostic)

src/content/reference/react/use-client.md
```diff
@@ -350,26 +353,27 @@ export default function Circle() {
   useLayoutEffect(() => {
     const canvas = ref.current;
     const context = canvas.getContext('2d');
+    context.reset();
```
sophiebits, line 356: effects can run multiple times :) like with offscreen in the future (and also because `[]` deps aren't specified here)

could've put this in the unmount but I put it here

src/content/reference/react/use-client.md
lunaleaps: We can re-title this and re-word to frame as capabilities!

src/content/reference/react/use-client.md
lunaleaps: We can have async components on client?

src/content/reference/react/use-client.md
```diff
@@ -215,39 +214,39 @@ This means that the component definition for `FancyText` will both be evaluated
...
 * Server Components benefit from running on the server. They can access the local filesystem and may experience low latency for data fetches and network requests.
 
-#### Limitations {/*limitations*/}
+#### Limitations of Server Components {/*limitations*/}
 * Server Components cannot support interaction as event handlers must be registered and triggered by a client.
-	* The browser, as an example client, is responsible for delegating UI clicks to the appropriate `onClick` handler.
+	* For example, event handlers like `onClick` can only be defined in Client Components.
 * Server Components cannot use most Hooks.
-	* When Server Components are rendered, they are conceptually resolved into instructions for the client to interpret and turn into UI primitives. This means that Server Components do not persist in memory after render and do not support re-rendering or state.
+	* When Server Components are rendered, their output is essentially a list of components for the client to render are conceptually resolved into instructions for the client to interpret and turn into UI primitives. Server Components do not persist in memory after render and cannot have their own state.
```
lunaleaps, line 247: I think this needs re-wording after the edit, not sure if this is what you intend

```suggestion
	* When Server Components are rendered, their output is essentially a list of components for the client to render. Server Components do not persist in memory after render and cannot have their own state.
```

src/content/reference/react/use-client.md
```diff
@@ -311,18 +312,18 @@ As `Counter` requires both the `useState` hook and event handlers to increment o
...
 
 export default async function CounterContainer() {
-  const initialValue = await getCounterValueFromFile();
+  const initialValue = await readFile('/path/to/counter_value');
   return <Counter initialValue={initialValue} />
 }
 ```
 
-For example, the parent component of the Client Component `CounterContainer` does not require `'use client'` as it is not interactive and does not use state. In addition, `CounterContainer` must be a Server Component as it reads from the local file system on the server. This is possible because Server Components, unlike Client Components, can be async functions.
+For example, `Counter`'s parent component, `CounterContainer`, does not require `'use client'` as it is not interactive and does not use state. In addition, `CounterContainer` must be a Server Component as it reads from the local file system on the server, which is possibly only in a Server Component.
```
lunaleaps, line 324:
```suggestion
For example, `Counter`'s parent component, `CounterContainer`, does not require `'use client'` as it is not interactive and does not use state. In addition, `CounterContainer` must be a Server Component as it reads from the local file system on the server, which is possible only in a Server Component.
```

src/content/reference/react/use-client.md
sophiebits: lol whoops thank you, this is indeed what I meant

sophiebits (commented):
ty

lunaleaps (approved):

PR #6413, opened 2023-11-06
Update code formatting on useTransition.md

mattcarrollcode:


src/content/reference/react/useTransition.md
```diff
@@ -1563,16 +1564,16 @@ export default function App() {
 // TODO: update to import from stable
 // react instead of canary once the `use`
 // Hook is in a stable release of React
-import React, { StrictMode } from 'react';
-import { createRoot } from 'react-dom/client';
-import './styles.css';
+import React, { StrictMode } from "react";
+import { createRoot } from "react-dom/client";
+import "./styles.css";
```
lunaleaps, line 1569: Are double-quoting these? I'm seeing single-quotes elsewhere

mattcarrollcode: This is what CodeSandbox does by default on save

https://github.com/reactjs/react.dev/assets/7158882/c83d2143-43a0-4c51-9383-99381e532694

lunaleaps: Ohoo I see, do you mind if we keep it single? This is how we format code at Meta and I also see most of our docs already use single quotes. I think we need some light-weight markdown formatting to start enforcing this. I'm wondering if we can turn it on with most formatting disabled. I looked into it before but I think this might be possible

lunaleaps (approved):

PR #6415, opened 2023-11-06
Fix feedback gtag value

kmiddleton14:
Feedback event wasn't displaying the parameter value in google analytics. By checking the boolean before passing to gtag fixes it. I've verified we're now seeing param values of the feedback event 'button' in google analytics

Screenshot 2023-11-06 at 6 23 19 PM

### DIFF
--- a/src/components/Layout/Feedback.tsx
+++ b/src/components/Layout/Feedback.tsx
@@ -45,13 +45,14 @@ const thumbsDownIcon = (
 );
 
 function sendGAEvent(isPositive: boolean) {
+  const value = isPositive ? 1 : 0;
   // Fragile. Don't change unless you've tested the network payload
   // and verified that the right events actually show up in GA.
   // @ts-ignore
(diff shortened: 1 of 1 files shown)


brianjo (approved):

src/components/Layout/Feedback.tsx
```diff
@@ -45,13 +45,14 @@ const thumbsDownIcon = (
...
 
 function sendGAEvent(isPositive: boolean) {
+  const value = isPositive ? 1 : 0;
   // Fragile. Don't change unless you've tested the network payload
   // and verified that the right events actually show up in GA.
   // @ts-ignore
   gtag('event', 'feedback', {
     event_category: 'button',
     event_label: window.location.pathname,
-    value: isPositive ? 1 : 0,
```
lunaleaps, line 54: @kmiddleton14 I might be missing something obvious but what exactly fixed this? Isn't this change the same?

PR #6417, opened 2023-11-07
Capitalize Server Action

Huxpro:
"Server Action", like "Effect", is a React-specific notion
that would be benefited from captailization to be distinguished
from its genertic meaning.

It seems like [Next doc](https://nextjs.org/docs/app/api-reference/functions/server-actions)
has also adopted such connventions and we should probably do the same.

### DIFF
--- a/src/content/reference/react-dom/components/form.md
+++ b/src/content/reference/react-dom/components/form.md
@@ -93,11 +93,11 @@ export default function Search() {
 
 ### Handle form submission with a Server Action {/*handle-form-submission-with-a-server-action*/}
 
-Render a `<form>` with an input and submit button. Pass a server action (a function marked with [`'use server'`](/reference/react/use-server)) to the `action` prop of form to run the function when the form is submitted.
+Render a `<form>` with an input and submit button. Pass a Server Action (a function marked with [`'use server'`](/reference/react/use-server)) to the `action` prop of form to run the function when the form is submitted.
 
-Passing a server action to `<form action>` allow users to submit forms without JavaScript enabled or before the code has loaded. This is beneficial to users who have a slow connection, device, or have JavaScript disabled and is similar to the way forms work when a URL is passed to the `action` prop.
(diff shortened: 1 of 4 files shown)


gaearon (approved):
yes. should also do this for Client and Server when they’re used in the React sense (eg “Client component”.

Huxpro:
@gaearon do you have a list of things that should be treated like this? It seems like "prop", "state", "ref", "context", etc. are treated in lowercase.

I'm asking because, in many languages (at least in Chinese and Japanese writings),  English words should be written in their canonical case without considering English rules (e.g. first words, title case).

gaearon:
i think it’s just Effect, Hook, and recent new terms where our meaning is overloaded (Client / Server things)

sophiebits:
Hmm I can see the argument for Server Action but I actually am not sure I agree for Client and Server. There the meaning is not any different than the English word alone. The distinction can be a little confusing especially with SSR but I am not sure the capitalization does anything to help.

gaearon:
I feel like it is different in the sense that "Server" can run at the build time while "Client" runs on the server

smikitky:
Other words that are capitalized in the docs include:

- Strict Mode
- Suspense
- Fragment
- Canary (though this is often ignored recently)

sophiebits:
We should probably have a linter for whatever we decide.

If people never capitalize it in their tweets does that mean we shouldn't capitalize it in the docs?

Huxpro:
> We should probably have a linter for whatever we decide.

@sophiebits that would be nice but it's pretty challenging to be complete. We can probably capture syntactic ones like `useBlah` Hook and "Client Component but others are tricky

Huxpro:
> Other words that are capitalized in the docs include:
> 
> * Strict Mode
> * Suspense
> * Fragment
> * Canary (though this is often ignored recently)

I've just manually updated those but yea I agreed with @sophiebits that we need...something at least. This is not going to scale at all

PR #6424, opened 2023-11-09
Capitalize word "Hook"

Huxpro:
Summary of changes:
See <https://github.com/reactjs/react.dev/pull/6417>.

There is one exception though. The error message
“Rendered fewer hooks than expected” from React is indeed "hooks"
so I kept as-is. Shall we change the error message from React?


A7med3bdulBaset:
You're doing a good job **BUT** keep in mind that this docs is internationalized. Every file you edit will conflict in the weekly rebasing for forks. 

that really hurts maintainer. So if that is not necessary please don't do.

thanks

Huxpro:
> You're doing a good job **BUT** keep in mind that this docs is internationalized. Every file you edit will conflict in the weekly rebasing for forks. 
> 
> 
> 
> that really hurts maintainer. So if that is not necessary please don't do.
> 
> 
> 
> thanks 

Not gonna lie but I AM maintainer of one of the translation. 

In fact, these editions were originated from translators being confused by words "action" being mis-capitalized so it's hard to distinguish if it was the literal word "action" vs. React-specific notion of "React Action" (which should be capitalized). 

I can see why this can bring up extra works for translators but hopefully you understand why such enforcement may contribute to better documentation and translation overall. 

We will see if a linter can be developed to help stop future commits to the docs mis-capitalize words so bulk edits like this won't happen time to time.

gaearon:
For consistency I think it's worth doing, but yes, apologies for the churn.

PR #6431, opened 2023-11-12
Update startTransition.md

prajwalkulkarni:
In the context of a function invocation, I believe it'd be more relevant to say that a function or a callback was called with no or some `arguments` instead of `parameters`


prajwalkulkarni:
@gaearon @Huxpro @sophiebits Could you kindly review my changes?

Huxpro:
LGTM ;)

Issue #6454, opened 2023-11-29
Bug: UI Navigation button overflows  issue


sanjeevkumar321:
### Website or app

https://react.dev/reference/react/startTransition

### issue image

![Screenshot 2023-11-29 103122](https://github.com/facebook/react/assets/46792207/d4f0904f-3f94-4c28-9dfb-ccc37cae3297)

### Expected image

![Screenshot 2023-11-29 103037](https://github.com/facebook/react/assets/46792207/3dfebe37-ff7e-4d64-9dd2-7f9ca09f687b)

code -
```
@media (min-width: 768px)
.md\:w-80 {
    width: 20rem;
}
```

fix code-
```
@media (min-width: 768px)
.md\:w-80 {
    width: 23rem;
}
```

### How often does this bug happen?

Every time

iamdebobrota:
# Attempted

iamdebobrota:
Have fixed the UI bugs, Raised the PR:  https://github.com/reactjs/react.dev/pull/6493

iamdebobrota:
<img width="691" alt="image" src="https://github.com/reactjs/react.dev/assets/99647439/0ddd7a67-03ec-40d3-818c-c90cb4239fe6">

rickhanlonii:
Fixed in https://github.com/reactjs/react.dev/pull/6519

PR #6456, opened 2023-12-01
fix footer Next/Previous  navigation text overflow and remove the rin…

sanjeevkumar321:
fix footer Next/Previous  navigation text overflow and remove the ring border on focus

#6454


rickhanlonii:
Fixed in https://github.com/reactjs/react.dev/pull/6519

PR #6459, opened 2023-12-02
Initial docs for DOM static resource loading functions

davidmccabe:
Add docs for the Float imperative functions: preconnect, prefetchDNS, preinit, preinitModule, preload, preloadModule.

@gnoff  Please read with a close eye for accuracy.

https://react-dev-git-fork-davidmccabe-float2-fbopensource.vercel.app/reference/react-dom#resource-preloading-apis

I think these are still canary only, is that correct?


gnoff (commented):
Going to keep reviewing but so far I've taken a look at the index and preload pages.

One question I have is whether we should convey the idea that for users of libraries and meta frameworks it is likely you don't need to call these preloading functions yourself becuase the tool you are using is probably doing it for you?

If you roll your own resource loading in React then these methods are useful
If you write libraries that manage resource loading then these methods are useful
But for likely the large majority of React developers these methods will probably not get used.

One case that may be more common is libraries probalby cannot reason about the domains you are likely to connect to so probably won't use preconnect and prefetchDNS much however App authors have this information. This may be the most commonly used methods by non-library/non-metaframework authors even if it seems like the least powerful

src/content/reference/react-dom/preload.md
```diff
@@ -0,0 +1,160 @@
...
+
+[See more examples below.](#usage)
+
+The `preload` function provides the browser with a hint that it should start downloading the given resource, which can save time.
+
+#### Parameters {/*parameters*/}
+
+* `href`: a string. The URL of the resource you want to download.
+* `options`: an object. It contains the following properties:
+  *  `as`: a required string. The type of resource. Its possible values are `audio`, `document`, `embed`, `fetch`, `font`, `image`, `object`, `script`, `style`, `track`, `video`, `worker`.
```
gnoff, line 43: The intended support semantics here are anything that a browser can preload. also the `as` property is named such because this maps to the `as` attribute of a `<link rel="preload"...` tag. Maybe we should clarify where the list of supported types comes from so if/when it evolves it's clear React's implementation will as well?

Maybe link to the relevant MDN documentation?

githrdw:
I was looking for this documentation!
Let me explain my use case to hopefully give more priority to this documentation :)
We are using Next.JS (14.0.5-canary.5) with React (18.2.0) and have tried using the Image component.
It contains a preload function and all kinds of niceties about srcset.

However, it did not meet our requirements, because it does not work with our WordPress API / CDN setup.
We are tied to use `<picture>` with `<source>`'s because of our complicated art direction, instead of Next's `` with `sizes`.
This means that we also have to take care of our own image preload.

Next briefly noticed this preload function in it's [meta docs](https://nextjs.org/docs/app/api-reference/functions/generate-metadata#resource-hints), but not all of the options.
It's thanks to TS I discovered the preload options are available, as Next and React did not document those yet.

githrdw:
In addition, I feel free to draw attention to this PR:
https://github.com/facebook/react/pull/27129

Adding `media` option to `preload` function.

The component I am working on is a hero image (LCP) and must therefore be preloaded with the highest priority.
The preload also needs to know which `<source>` is loaded into `<picture>`, depending on the media query.
This preload media query cannot yet be configured in any way.

src/content/reference/react-dom/preinit.md
```diff
@@ -0,0 +1,133 @@
...
+
+[See more examples below.](#usage)
+
+The `preinit` function provides the browser with a hint that it should start downloading and executing the given resource, which can save time. Scripts that you `preinit` are executed when they finish downloading. Stylesheets that you preinit are inserted into the document, which causes them to go into effect right away.
+
+#### Parameters {/*parameters*/}
+
+* `href`: a string. The URL of the resource you want to download and execute.
+* `options`: an object. It contains the following properties:
+  *  `as`: a required string. The type of resource. Its possible values are `script` and `style`.
```
A7med3bdulBaset, line 56:
```suggestion
  * `as`: a required string. The type of resource. Its possible values are `script` and `style`.
```

src/content/reference/react-dom/preinit.md
```diff
@@ -0,0 +1,133 @@
...
+  * `precedence`: a string. Required with stylesheets. Says where to insert the stylesheet relative to others. Stylesheets with higher precedence can override those with lower precedence. The possible values are `reset`, `low`, `medium`, `high`. 
+  *  `crossOrigin`: a string. The [CORS policy](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/crossorigin) to use. Its possible values are `anonymous` and `use-credentials`. It is required when `as` is set to `"fetch"`.
+  *  `integrity`: a string. A cryptographic hash of the resource, to [verify its authenticity](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity).
+  *  `nonce`: a string. A cryptographic [nonce to allow the resource](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/nonce) when using a strict Content Security Policy.
```
A7med3bdulBaset, line 60:
```suggestion
  * `nonce`: a string. A cryptographic [nonce to allow the resource](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/nonce) when using a strict Content Security Policy. 
```

src/content/reference/react-dom/preinit.md
```diff
@@ -0,0 +1,133 @@
...
+* `options`: an object. It contains the following properties:
+  *  `as`: a required string. The type of resource. Its possible values are `script` and `style`.
+  * `precedence`: a string. Required with stylesheets. Says where to insert the stylesheet relative to others. Stylesheets with higher precedence can override those with lower precedence. The possible values are `reset`, `low`, `medium`, `high`. 
+  *  `crossOrigin`: a string. The [CORS policy](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/crossorigin) to use. Its possible values are `anonymous` and `use-credentials`. It is required when `as` is set to `"fetch"`.
+  *  `integrity`: a string. A cryptographic hash of the resource, to [verify its authenticity](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity).
+  *  `nonce`: a string. A cryptographic [nonce to allow the resource](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/nonce) when using a strict Content Security Policy. 
+  *  `fetchPriority`: a string. Suggests a relative priority for fetching the resource. The possible values are `auto` (the default), `high`, and `low`.
```
A7med3bdulBaset, line 61:
```suggestion
  * `fetchPriority`: a string. Suggests a relative priority for fetching the resource. The possible values are `auto` (the default), `high`, and `low`.
```

src/content/reference/react-dom/preconnect.md
```diff
@@ -0,0 +1,96 @@
...
+
+#### Returns {/*returns*/}
+
+`preconnect` returns nothing.
+
+#### Caveats {/*caveats*/}
+
+* Multiple calls to `preconnect` with the same server have the same effect as a single call.
+* In the browser, you can call `preconnect` in any situation: while rendering a component, in an effect, in an event handler, and so on.
+* In server-side rendering or when rendering Server Components, `preconnect` only has an effect if you call it while rendering a component or in an async context originating from rendering a component. Any other calls will be ignored.
```
tom-sherman, line 59: > an async context originating from rendering a component

I don't think it's obvious what this refers to. I think an example of one of these contexts would help here.

rbalicki2:
This isn't related to the documentation *per se*, but since these resources aren't disposable, perhaps the docs could include information about what one wants to do if one wants to avoid memory leaks? Is cleanup React's responsibility? Are these scoped to the tree from which `preconnect` etc. are called? Presumably not, since they seem to be callable imperatively?

In addition, what happens if the cryptographic hash doesn't match? Is this exposed to the user anywhere?

PR #6463, opened 2023-12-05
Add network usage example and caveat to useOptimistic page

kmiddleton14:
- Add Caveats section to useOptimistic page
- Add demo to optimistically update the UI while waiting for a network request

https://github.com/reactjs/react.dev/assets/20743223/59a1c097-3630-433b-b9f2-4f1f8c1b0955

### DIFF
--- a/src/content/reference/react/useOptimistic.md
+++ b/src/content/reference/react/useOptimistic.md
@@ -151,7 +151,7 @@ export async function deliverMessage(message) {
 <Sandpack>
 
 ```js like-button.js active
-import { useOptimistic } from "react";
+import { startTransition, useOptimistic } from "react";
 
 export function LikeButton({ likes, updateLikes, disabled }) {
   const [optimisticLikes, addOptimisticLike] = useOptimistic(
@@ -160,8 +160,10 @@ export function LikeButton({ likes, updateLikes, disabled }) {
   );
 
   async function addLike() {
-    addOptimisticLike();
-    await updateLikes(likes);
+    startTransition(() => {
+      addOptimisticLike();
+      updateLikes(likes);
+    })
   }
 
   return (
--- a/src/content/reference/react/useOptimistic.md
+++ b/src/content/reference/react/useOptimistic.md
@@ -59,6 +59,11 @@ function AppContainer() {
 * `optimisticState`: The resulting optimistic state. It is equal to `state` unless an action is pending, in which case it is equal to the value returned by `updateFn`.
 * `addOptimistic`: `addOptimistic` is the dispatching function to call when you have an optimistic update. It takes one argument, `optimisticValue`, of any type and will call the `updateFn` with `state` and `optimisticValue`.
 
+
+#### Caveats {/*caveats*/}
+
+* An optimistic state update needs to be called in an action or wrapped with `startTransition`. If it's used outside an action or a transition React will throw an error.
+
 ---
 
 ## Usage {/*usage*/}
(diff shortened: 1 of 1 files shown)


src/content/reference/react/useOptimistic.md
```diff
@@ -59,6 +59,11 @@ function AppContainer() {
 * `optimisticState`: The resulting optimistic state. It is equal to `state` unless an action is pending, in which case it is equal to the value returned by `updateFn`.
 * `addOptimistic`: `addOptimistic` is the dispatching function to call when you have an optimistic update. It takes one argument, `optimisticValue`, of any type and will call the `updateFn` with `state` and `optimisticValue`.
 
+
+#### Caveats {/*caveats*/}
+
+* An optimistic state update needs to be called in an action or wrapped with `startTransition`. If it's used outside an action or a transition React will throw an error.
```
lunaleaps, line 65:
```suggestion
* An optimistic state update needs to be called in an action or wrapped with `startTransition`. If it's used outside an action or a transition, React will throw an error.
```

src/content/reference/react/useOptimistic.md
```diff
@@ -143,3 +148,67 @@ export async function deliverMessage(message) {
 ```
 
 </Sandpack>
+
+---
+
+### Optimistically update the UI while waiting for a network request {/*optimistically-update-while-waiting-for-network*/}
```
lunaleaps, line 154: Hmm I may be forgetting something but what was the reasoning for us to add this example? It seems quite similar to the form action example above, the only difference being that it isn't in a form? I think we'd want to add some description here of why this example is worth highlighting. 

In addition, I think it'd be valuable to expand the example and show a failed optimistic update.

kmiddleton14: Yes it is similar but wanted to showcase another example that isn't using a form. I agree on adding in a failure example and have it as a followup

lunaleaps: Can we add a line about that?
Something like 
```suggestion
### Optimistically update the UI while waiting for a network request {/*optimistically-update-while-waiting-for-
network*/}

`useOptimistic` can also be used outside of forms, for any network request but within a transition.
```

src/content/reference/react/useOptimistic.md
```diff
@@ -143,3 +148,67 @@ export async function deliverMessage(message) {
...
+
+<Sandpack>
+
+```js like-button.js active
+import { startTransition, useOptimistic } from "react";
+
+export function LikeButton({ likes, updateLikes, disabled }) {
+  const [optimisticLikes, addOptimisticLike] = useOptimistic(
```
lunaleaps, line 162: Also the example for me is not updating the like count from 0->1

CleanShot 2023-12-15 at 09 45 56

re: https://react-ac6nnqh06-fbopensource.vercel.app/reference/react/useOptimistic#noun-labs-1201738-(2)

damianstasik: @lunaleaps One way to fix this example would be to start a transition with `useTransition` hook instead of using the standalone `startTransition` function. This way you'll be able to:
- replace the `disabled` prop with the transition's pending state,
- make the arrow function (scope) passed to `startTransition` async and await `updateLikes` inside (crucial).

It fixes the issue while slightly simplifying the code, here's the preview:
https://codesandbox.io/p/sandbox/gallant-nova-8w3s2h

PR #6483, opened 2023-12-12
Text Display error in Go full-stack with a framework

nepalivai:
Fix error [27816](https://github.com/facebook/react/issues/27816)

What dose this PR do?
It fixes the issue with the landing page of the react website.

PR #6490, opened 2023-12-15
Fix option color in react-conf dropdown (Home section)

JoseSanchez54:
Currently the text is white on a white background when the field is not selected, the text color has been fixed to be black on white


rickhanlonii:
Fixed in https://github.com/reactjs/react.dev/pull/6483

Issue #6460, opened 2023-12-02
UI BUG : Navigation button text overflows from parant span box


godspeed-03:
Website or app
https://react.dev/reference/react/startTransition

Repro steps
When hovering on the next navigation button, the text overflows the Box.
issue image
![Screenshot from 2023-12-02 23-55-42](https://github.com/reactjs/react.dev/assets/119731754/3b915011-0006-43e1-a7ab-927a9168c8f0)

code -
Just needed to warp the link text with Break-word property.

How often does this bug happen?
Every time

nitesh9109:
@godspeed-03, I am also using the Brave browser. However, this problem doesn't occur in Brave, instead, it happens in Firefox.

iamdebobrota:
# Attempted

iamdebobrota:
Raised the PR, have fixed the UI : https://github.com/reactjs/react.dev/pull/6493

iamdebobrota:
Here is an example  : 
<img width="947" alt="image" src="https://github.com/reactjs/react.dev/assets/99647439/4983463c-c490-40e4-ba16-fb26c3cbaebd">

prajwalkulkarni:
I think using `break-all` or `break-word` still wouldn't fix the issue completely. What if the string is even longer? The container height would be increased to avoid overflowing and accommodate the complete string in multiline.
The same goes with increasing the parent width, or even alternatively setting the width to be `auto` with a min-width. It would mess up the design system and lead to visual inconsistencies. 
Instead, I think without altering the width of the container, we can hide the overflowing content with ellipsis and on hover show the complete string on the tooltip.

rickhanlonii:
Fixed https://github.com/reactjs/react.dev/pull/6519

PR #6493, opened 2023-12-15
fixed: UI BUG : Navigation button text overflows from parant span box…

iamdebobrota:
…, issues - #6460

### DIFF
--- a/src/components/DocsFooter.tsx
+++ b/src/components/DocsFooter.tsx
@@ -82,7 +82,9 @@ function FooterLink({
         <span className="block no-underline text-sm tracking-wide text-secondary dark:text-secondary-dark uppercase font-bold group-focus:text-link dark:group-focus:text-link-dark group-focus:text-opacity-100">
           {type}
         </span>
-        <span className="block text-lg group-hover:underline">{title}</span>
+        <span className="block text-lg group-hover:underline break-all">
+          {title}
+        </span>
       </span>
     </NextLink>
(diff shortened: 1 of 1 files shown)


iamdebobrota:
CLA signed done

src/components/DocsFooter.tsx
```diff
@@ -82,7 +82,9 @@ function FooterLink({
         <span className="block no-underline text-sm tracking-wide text-secondary dark:text-secondary-dark uppercase font-bold group-focus:text-link dark:group-focus:text-link-dark group-focus:text-opacity-100">
           {type}
         </span>
-        <span className="block text-lg group-hover:underline">{title}</span>
+        <span className="block text-lg group-hover:underline break-all">
```
harish-sethuraman, line 85: I don't feel this to be elegant 
Screenshot 2023-12-16 at 2 15 19 PM

I'd rather consider removing the  `md:w-80` or removing explicit width and adding `md:w-auto` along with a min width so that the width takes upon content's width when it is pretty long else it will take the default min width

Screenshot 2023-12-16 at 2 15 57 PM

iamdebobrota: Sure, here is the updated UI
image

rickhanlonii:
Fixed here in a more robust way so it won't word break almost ever (and not at all today): https://github.com/reactjs/react.dev/pull/6519

Issue #5724, opened 2023-03-18
Feedback button not disappearing after submission


ankitjangidx:
### I recently visited the official ReactJS documentation website and noticed a feedback button. After submitting my feedback through the button, I noticed that it did not disappear, causing me to have to scroll down to view the next section in the sidebar. I refreshed the page to see if it was a temporary issue, but the button is still visible.

https://user-images.githubusercontent.com/100501722/226102319-22465293-5507-4591-8e7b-f8044324194d.mp4

<br/>

### I believe this could be a bug in the website's functionality and would like to report it to the development team. It's important to have a seamless user experience, and having a feedback button that doesn't disappear after submission can be confusing for users.

<br/>

sambhavgupta0705:
Hey Can I work on this issue?

iamdebobrota:
Have made the changes:
- After submitting the feedback, this will disappear after 10 seconds.
-  If you move to another page, the feedback button should appear.

Have create the PR : https://github.com/reactjs/react.dev/pull/6495


Thank you.

rickhanlonii:
Fixed in https://github.com/reactjs/react.dev/pull/6533

PR #6495, opened 2023-12-15
Ui : Feedback button not disappearing after submission issue :- #5724

iamdebobrota:
Resolved this UI bugs : https://github.com/reactjs/react.dev/issues/5724


harish-sethuraman:
@mattcarrollcode I wanted to confirm if we want to keep showing the feedback buttons even after collecting it or should we allow it to disappear?

iamdebobrota:
On every page, we should take the feedback for the same. So, It will be good if we can take the feedback on each page for different content.

rickhanlonii:
Fixed in https://github.com/reactjs/react.dev/pull/6533

PR #6499, opened 2023-12-17
Improve documentation for __html

Daniel15:
The intent of the `__html` object for `dangerouslySetInnerHTML` is to act as a marker for which strings can be rendered as HTML. It should be wrapped in the `__html` wrapper at the point that it is produced.

For use cases of `dangerouslySetInnerHTML` on Facebook, `__html` rarely appears in the  JavaScript code, as often we render the HTML server-side and return the `__html` object JSON-encoded as part of the network response. For client-side generated HTML, there's a sanitization function that strips out bad things (scripts, etc) then returns the `__html` object. We've also got a lint rule ensuring that `__html` is not explicitly written in JavaScript code. This means that if we see an object with a `__html` property, it *should* be OK to render it via `dangerouslySetInnerHTML`.

Because of this, it doesn't make sense to create it inline like `<div dangerouslySetInnerHTML={{__html: markup}} />`. This bypasses the main purpose of using `__html` as a marker. There's no way to tell if `markup` is expected to be rendered as HTML (e.g. it was produced by a Markdown library, ran through a HTML sanitizer, etc) or if could be totally unsafe to use it.

This was documented in the React 0.x docs, but the latest documentation no longer mentions it. I've added a paragraph about it to the current docs.

### DIFF
--- a/src/content/reference/react-dom/components/common.md
+++ b/src/content/reference/react-dom/components/common.md
@@ -982,6 +982,8 @@ textarea { display: block; margin-top: 5px; margin-bottom: 10px; }
 
 </Sandpack>
 
+The `{__html}` object should be created as close to where the HTML is generated as possible, like the above example does in the `renderMarkdownToHTML` function. This ensures that all raw HTML being used in your code is explicitly marked as such, and that only variables that you expect to contain HTML are passed to `dangerouslySetInnerHTML`. It is not recommended to create the object inline like `<div dangerouslySetInnerHTML={{__html: markup}} />`.
+
 To see why rendering arbitrary HTML is dangerous, replace the code above with this:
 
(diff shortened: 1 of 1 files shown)


davidmccabe:
Thanks for contributing this. It points out a good nuance that might not be obvious from the example. I'd want to put it at the end of the section, though, after the paragraph beginning "The code embedded in the HTML will run." Where it's currently placed, it interrupts the example that shows why setting HTML is dangerous in the first place. Once you understand why it's dangerous, we can spell out this nuance about how better to avoid the danger.

Daniel15:
Yeah I wasn't sure where to put it... Initially I had it at the top above the example, since I thought people might read the page and not read anything below the example. I can move it below the example if you think that'd be better!

PR #6501, opened 2023-12-18
Add guiding note

smith558:
Without the note, an entire tutorial step is just assumed. It might seem obvious to many, but it took a fair time to realise when I went through the tutorial myself the first time. It broke all further examples that follow in the tutorial. This goes against being a beginner-friendly and **guided** tutorial. Additionally, look at my comment about misleading instructions that follow if the guiding note is not there https://github.com/reactjs/react.dev/pull/6501#discussion_r1457171715.

### DIFF
--- a/src/content/learn/thinking-in-react.md
+++ b/src/content/learn/thinking-in-react.md
@@ -484,15 +484,27 @@ function FilterableProductTable({ products }) {
 
 Inside the `SearchBar`, you will add the `onChange` event handlers and set the parent state from them:
 
-```js {5}
-<input 
-  type="text" 
-  value={filterText} 
-  placeholder="Search..." 
-  onChange={(e) => onFilterTextChange(e.target.value)} />
+```js {4,5,13,19}
+function SearchBar({
+  filterText,
+  inStockOnly,
+  onFilterTextChange,
+  onInStockOnlyChange
+}) {
+  return (
+    <form>
+      <input
+        type="text"
+        value={filterText}
+        placeholder="Search..."
+        onChange={(e) => onFilterTextChange(e.target.value)}
+      />
+      <label>
+        <input
+          type="checkbox"
+          checked={inStockOnly}
+          onChange={(e) => onInStockOnlyChange(e.target.checked)}
 ```
-
-(Set up changing state for the checkbox using `onInStockOnlyChange` yourself.)
 
 Now the application fully works!
 
--- a/src/content/learn/thinking-in-react.md
+++ b/src/content/learn/thinking-in-react.md
@@ -492,6 +492,8 @@ Inside the `SearchBar`, you will add the `onChange` event handlers and set the p
   onChange={(e) => onFilterTextChange(e.target.value)} />
 ```
 
+(Set up changing state for the checkbox using `onInStockOnlyChange` yourself.)
+
 Now the application fully works!
 
 <Sandpack>


rickhanlonii:
Thanks, but I'm not sure this adds much.

smith558:
> Thanks, but I'm not sure this adds much.

Without that, it is just assumed. It might seem obvious to you, but when I went through the tutorial myself the first time, it took a fair time to realise. It broke all my further examples in the tutorial. This goes against the idea of it being a beginner-friendly and *guided* tutorial.

src/content/learn/thinking-in-react.md
```diff
@@ -492,6 +492,8 @@ Inside the `SearchBar`, you will add the `onChange` event handlers and set the p
   onChange={(e) => onFilterTextChange(e.target.value)} />
 ```
 
+(Set up changing state for the checkbox using `onInStockOnlyChange` yourself.)
+
 Now the application fully works!
```
smith558, line 497: @smith558 The sentence here saying "Now the application fully works!" is misleading, since the application doesn't yet work, it needs further code added (unless the note I would like to add is there).

rickhanlonii: It does work, as shown in the next sandbox, not the previous. The guide is written as "here's the code you'll right" and "here's a working sandbox with those changes".

smith558: The sandbox works of course. I meant if you follow the steps in your own code editor. Look, it is inconsistent with the rest of the tutorial articles (as well as this one) where the pattern is that even the most obvious steps are explicitly mentioned. This is the only place in all of the tutorial articles that this pattern suddenly breaks for a moment. I just went through all of them in order.

smith558: Decide as you wish but you've probably worked with these tutorials many many times and are well proficient with React as well. I tried this tutorial for the first time and it happened to be an issue. You can hardly have a better validation that a problem may exist for your target audience.

smith558: @rickhanlonii would you please let me know if it seems to make more sense now?

rickhanlonii: @smith558 i see, yeah I agree, thanks for explaining it for me more. 

Instead of this note, can we just expand the code above this to give a full example, highlighting the code to change like this:

Screenshot 2024-01-22 at 5 00 04 PM

Here's the code:

```js
```js {4,5,13,19}
function SearchBar({
  filterText,
  inStockOnly,
  onFilterTextChange,
  onInStockOnlyChange
}) {
  return (
    <form>
      <input
        type="text"
        value={filterText}
        placeholder="Search..."
        onChange={(e) => onFilterTextChange(e.target.value)}
      />
      <label>
        <input
          type="checkbox"
          checked={inStockOnly}
          onChange={(e) => onInStockOnlyChange(e.target.checked)}
        />{" "}
        Only show products in stock
      </label>
    </form>
  );
}
```
```

rickhanlonii:
Re-opening, if you can make the change suggested above then we can merge!

smith558:
> Re-opening, if you can make the change suggested above then we can merge!

Sure! Happy to have it ready tomorrow.

src/content/learn/thinking-in-react.md
smith558: Done

rickhanlonii (approved):
Looks great, thanks!

PR #6502, opened 2023-12-18
Adding Learn docs for RSCs

lunaleaps:
This is a draft PR for the outline shared here: https://docs.google.com/document/d/1rQVqqBI7KXVRQjValsCXqsEFqmGQ6h4fWdFAQ9X0d7o/edit?usp=sharing (TODO: put this up accessible somewhere.

### DIFF
--- a/src/content/learn/fetching-data.md
+++ b/src/content/learn/fetching-data.md
@@ -1,0 +1,198 @@
+---
+title: Fetching Data
+---
+
+<Intro>
+
+The data rendered in your React app may come from a variety of sources and in many cases, your components may need to "wait" for the data to be fetched before they can render.
+
+In this chapter, we will go over common data sources and how to fetch and use that data in your React components.
+
+</Intro>
+
+<YouWillLearn>
+
+* Static and dynamic data and how it affects data access
+* How to fetch data from common data sources
+* How components "wait" for data to be fetched
+
+</YouWillLearn>
+
+## Static and Dynamic Data {/*static-and-dynamic-data*/}
+
+In our previous React examples, we've used local variables to hold data that our components use.
+
+```js
+const ESTABLISHED_YEAR = 1986;
+
+function CompanyFooter() {
+  return <p>Quality since {ESTABLISHED_YEAR}</p>;
+}
+```
+In this component, the data is the year we store in the local variable `ESTABLISHED_YEAR`.
+
+Our app may also use ESTABLISHED_YEAR in other places so we can decouple the data from our UI and provide the data in a separate module.
+
+```js
+import ESTABLISHED_YEAR from "./data";
+
+function CompanyFooter() {
+  return <p>Quality since {ESTABLISHED_YEAR}</p>;
+}
+```
+
+In both of these examples, the data is set and read once for `CompanyFooter`. The examples read the data outside of the component function so every-time the component is rendered, it refers to the same data.
+
+This works because the nature of the data in this example is unchanging. An establishment year for a company is a historical record that will not be updated in the future. We refer to this type of data as **static**.
+
+On the other hand, there may be data that is ever-changing or unique to a user or location. Consider a Weather app that renders the temperature of a local city, the temperature data powering that app will frequently update. We refer to this type of data as **dynamic** data.
+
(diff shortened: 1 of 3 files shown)


src/content/learn/fetching-data.md
```diff
@@ -0,0 +1,206 @@
...
+
+<Illustration>
+Illustration of the parallel work while getting dry cleaning
+</Illustration>
+
+Now, imagine you are trying to render the component `LATemperature`. The first thing you do is kick off the asynchronous work of reading the temperature file and you've received `tempPromise` as a Promise. Now, you're free to do the next set of work. But wait, the next instruction requires you to know the temperature, yet all we have is a Promise. The Promise may or may not be ready with the actual temperature value.
+
+Here is where `await` comes in. `await` is a keyword that is used before a Promise and tells JavaScript to pause running the current function and wait for the Promise to resolve.Then, when it is resolved, await will unwrap the Promise container to return the value. We use await on tempPromise to pause our component rendering until the file is read and receive the value returned.
```
davidmccabe, line 155:
```suggestion
Here is where `await` comes in. `await` is a keyword that is used before a Promise and tells JavaScript to pause running the current function and wait for the Promise to resolve. Then, when it is resolved, `await` will unwrap the Promise container to return the value. We use `await` on `tempPromise` to pause our component rendering until the file has been read. Once the process is done, we receive the value that was read from the file and assign it to the variable `temp`.
```

src/content/learn/fetching-data.md
```diff
@@ -0,0 +1,206 @@
...
+
+In our analogy, this would be similar to waiting at the dry cleaners for them to finish because the next chore on your todo list requires those clothes. Thus, preventing you from doing anything else in the meantime.
```
davidmccabe, line 157: The analogy seems imperfect: This makes it sound like using `await` is equivalent to using a synchronous API that pauses the main thread. 

Maybe you need to introduce another task that depends on the dry cleaning, like packing your suitcase for a trip. You can do other stuff in the meantime, but to finish packing your suitcase you need to get the dry cleaning back.

src/content/learn/fetching-data.md
```diff
@@ -0,0 +1,206 @@
...
+  // expensive calculation
+  const offset = calculateNewOffset(lastTouchPoint);
+
+  return ...
+}
+```
+
+</DeepDive>
+
+You'll notice that our component function declarations now have an `async` prefix. This is also a JavaScript keyword and any function that uses `await` will need to be marked with `async`. If a function `await`s a Promise, it itself becomes asynchronous.
```
davidmccabe, line 186: This paragraph maybe belongs in the main discussion. 

Whenever you use `await` within a function, you have to add the `async` keyword to the beginning of the function.

src/content/learn/fetching-data.md
```diff
@@ -0,0 +1,206 @@
...
+
+```js
+async function HoveringButton({ userContext, label, lastTouchPoint }) {
+  // async operation
+  const localePromise = getLocale(userContext);
+
+  // expensive calculation
+  const offset = calculateNewOffset(lastTouchPoint);
+
+  return ...
```
davidmccabe, line 180: Example incomplete? It doesn't `await` anything.

src/content/learn/fetching-data.md
```diff
@@ -0,0 +1,206 @@
...
+}
+```
+
+</DeepDive>
+
+You'll notice that our component function declarations now have an `async` prefix. This is also a JavaScript keyword and any function that uses `await` will need to be marked with `async`. If a function `await`s a Promise, it itself becomes asynchronous.
+
+Imagine being the parent function calling the `LATemperature` component function. When you call `LATemperature` and it gets to the instruction of `await`-ing `readFilePromise`, JavaScript execution returns to you and you receive a Promise that promises to give you the completed render of `LATemperature`.
```
davidmccabe, line 188: LATemperature is a component, not a normal function, so you wouldn't normally call it. Maybe the chaining aspect of promises should be explained with a non-component function: for example, you could create a function `readTemperature` that reads the file and then parses it and extracts the temperature as a number.

nickmccurdy:
I'm glad this is being worked on, thanks! lmk if you'd like community feedback.

PR #6527, opened 2023-12-30
Fix missing function keyword

A7med3bdulBaset:
This PR fixes the missing function keyword in the code snippet. and change exported server actions to named exported. Why? because, the file beginning with `'use server'` is supposed to export more than one server action.

### DIFF
--- a/src/content/reference/react/use-server.md
+++ b/src/content/reference/react/use-server.md
@@ -139,7 +139,7 @@ To update the UI based on the result of a Server Action while supporting progres
 // requestUsername.js
 'use server';
 
-export default async function requestUsername(formData) {
+export async function requestUsername(formData) {
   const username = formData.get('username');
   if (canRequest(username)) {
     // ...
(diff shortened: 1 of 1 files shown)


lunaleaps:
> Why? because, the file beginning with 'use server' is supposed to export more than one server action.

I don't believe that is true, where did you get that impression?

A7med3bdulBaset:
> I don't believe that is true, where did you get that impression?

If this is not right I revert them back to named exports

lunaleaps:
`'use server'` at the top of a module just marks that all exports from that module must be run on the server. It shouldn't be placing limitations on how many exports there are.

Slightly unrelated -- but this reminded me: 
You can use `'use server'` in a local scope. One gotcha is that you can't import a server action into client code that only has `'use server'` in its local function scope. It needs to be on the module scope. 

The only time to use `'use server'` in a local function scope is if its declared in a client code.

lunaleaps (approved):

PR #6539, opened 2024-01-09
fix: Grammar in versioning-policy.md

zoephyr:
Adds a missing word in the documentation describing canary versioning


rickhanlonii:
Thanks!

PR #6544, opened 2024-01-11
chore(lint): prefer `const`

account-noreply:
If a variable is never reassigned, using the const declaration is better.


rickhanlonii:
Thanks, but we're not strict about this.

PR #6550, opened 2024-01-13
Upgrade tailwind version

prajwalkulkarni:
# 💫 Changelog
:star: Upgrade `tailwindcss` to `3.4.1`.
:star: Adds support for min-width `min-w-*` out of the box and more (https://tailwindcss.com/blog/tailwindcss-v3-4)
:star: Removed `80: '20rem'` from `minWidth` in `tailwind.config.js`


prajwalkulkarni:
Hey Ricky, from our previous conversation (#6519 ), it was decided to upgrade the tailwindcss version to `3.4` as the latest version now supports `min-w-*`. 
I've also removed overwriting the property of `minWidth` object. Did a quick sanity after the upgrade and everything looks good.

In `tailwind.config.js` we're overwriting the value of `xs` in `maxWidth` to `21rem` in contrast to the actual value of `20rem`. Although I'm not sure why that was changed, should we consider removing it, so that the default value of `20rem` is used?
@rickhanlonii

tailwind.config.js
```diff
@@ -60,9 +60,8 @@ module.exports = {
         ...defaultTheme.maxWidth,
         xs: '21rem',
       },
-      minWidth:{
+      minWidth: {
```
rickhanlonii, line 63: You can just remove this whole key since there are no overrides anymore.

rickhanlonii:
@prajwalkulkarni not sure why it's 21 and not 20. You could either just leave it, or test every place it's used to see if 20 works just as well, or rename it to max-w-custom-xs, so max-w-xs is the default value. I don't have a strong opinion either way.

prajwalkulkarni:
The `max-w-xs` is mostly used in viewports with breakpoint `lg` and above. Using `20rem` instead of `21rem` is causing spacing issues, which requires me to update the width of the parent container as well. So, for the time being, I feel it is ok to keep it as is, although I've updated the name to `custom-xs` so that the original value corresponding to `max-w-xs` is preserved.  
I will work on updating the widths some other time, maybe as a separate PR so that we can completely get rid of overriding `maxWidth`, but for this PR, I think we can limit the scope to just upgrading the tailwind version. What do you think?

rickhanlonii:
Awesome, thanks! In future PRs, please turn off the class sorting because it makes it difficult to review the changes.

PR #6554, opened 2024-01-16
Fix async-await syntax

jamiebuilds:


rickhanlonii:
nice catch thanks @jamiebuilds

PR #6562, opened 2024-01-19
Refactor: Remove unused 'index' variable

clevermiraz:
In this commit, the 'index' variable was removed from the map function within the JSX. The 'index' variable was previously used for mapping over the 'items' array, but it became unnecessary as it wasn't utilized within the loop. This change improves code readability and eliminates potential confusion. No functional impact is expected.

### DIFF
--- a/src/content/learn/choosing-the-state-structure.md
+++ b/src/content/learn/choosing-the-state-structure.md
@@ -527,7 +527,7 @@ export default function Menu() {
     <>
       <h2>What's your travel snack?</h2>
       <ul>
-        {items.map((item, index) => (
+        {items.map((item) => (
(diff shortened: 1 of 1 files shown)


BartoszKlonowski (commented):
This doesn't break the example, but we still do have unused `index` in previous example (the one showing duplicated states problem), which is in line 463 of the page's code.
This also has to be removed.

**Note** however, that completely removing `index`, though unused, will leave the line 557
> (Alternatively, you may hold the selected index in state.)

completely without example how to do so. On the other hand, I don't think the unused `index` is emphasized enough for reader to notice the idea behind it.

clevermiraz:
Noted.

PR #6565, opened 2024-01-21
Fix typo in Tic-Tac-Toe tutorial

t-hamano:
My understanding is that the subject is plural, so I don't think the "s" is necessary.


rickhanlonii (approved):

rickhanlonii:
Nice catch

PR #6572, opened 2024-01-22
Remove redundant component from sandbox example

smith558:


rickhanlonii:
The text right above this sandbox says:

> Built-in components like <button> only support built-in browser events like onClick. However, you can also create your own components, and give their event handler props any application-specific names that you like.

So showing a separate component is the entire point of the sandbox.

smith558:
> The text right above this sandbox says:
> 
> > Built-in components like only support built-in browser events like onClick. However, you can also create your own components, and give their event handler props any application-specific names that you like.
> 
> So showing a separate component is the entire point of the sandbox.

But there is already such a component in the sandbox, precisely, `Toolbar`. The `Button` example adds no value and is not a good component use since it's identical to `<button>`, that is against good practice and a bad example then. It _doesn't_ actually use any "application-specific names" as mentioned in the text. Either `Button` should be removed or the whole sandbox example reworked to use some application-specific event handler props.

smith558:
@rickhanlonii I think this one might be worth quickly looking at once more.

rickhanlonii:
In isolation might agree, but it's certainly not wrong to write code like this, and the next doc after this builds off this and uses the simple Button element to teach responding to events, diving deeper into the ways to use Button. 

I appreciate the feedback, but we're going to keep it as is.

PR #6574, opened 2024-01-23
Remove unused variables

smith558:
The variable `nextId` is not used in either of the sandbox examples at all.


rickhanlonii (approved):
Nice catch

PR #6583, opened 2024-01-25
update createPortal docs to use a specific portal node

souporserious:
The docs for `createPortal` currently mention to use `document.body` as the main example of a portal node to use. However, this pattern is a footgun and can lead to a full style re-calculation of the `body` when the node is updated.

This PR updates the `createPortal` docs to refer to creating a specific portal node to render into that avoids this. Please see [this article](https://atfzl.com/articles/don-t-attach-tooltips-to-document-body/
) for a detailed explanation of the issue.

### DIFF
--- a/src/content/reference/react-dom/createPortal.md
+++ b/src/content/reference/react-dom/createPortal.md
@@ -29,13 +29,17 @@ To create a portal, call `createPortal`, passing some JSX, and the DOM node wher
 ```js
 import { createPortal } from 'react-dom';
 
+const portalNode = document.createElement('div')
+portalNode.id = 'portal'
+document.appendChild(portalNode)
+
 // ...
 
 <div>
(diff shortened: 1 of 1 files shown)


src/content/reference/react-dom/createPortal.md
```diff
@@ -29,13 +29,17 @@ To create a portal, call `createPortal`, passing some JSX, and the DOM node wher
 ```js
 import { createPortal } from 'react-dom';
 
+const portalNode = document.createElement('div')
+portalNode.id = 'portal'
+document.appendChild(portalNode)
```
souporserious, line 34: It's a little noisy to repeat this over and over throughout the doc. Maybe it would be better to declare this once up front and refer to something like `document.getElementById('portal')`?

PR #6593, opened 2024-01-31
Update: Usage of setState function

rituparna-ui:
The correct usage for any state updating function is to use the function form of it.
`setCount(count + 1)` vs `setCount((prevCount) => prevCount + 1)`

If the state update depends on any previous value of the state, it must be fetched from the argument of the function that is passed to the state updating function

### DIFF
--- a/src/content/learn/index.md
+++ b/src/content/learn/index.md
@@ -325,7 +325,7 @@ function MyButton() {
   const [count, setCount] = useState(0);
 
   function handleClick() {
-    setCount(count + 1);
+    setCount((prevCount) => prevCount + 1);
   }
 
   return (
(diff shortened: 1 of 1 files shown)


pwbriggs:
I'm not sure it makes sense to change this here. As per [`CONTRIBUTING.md`],

> Learn React is designed to introduce fundamental concepts in a step-by-step way. Each individual article in Learn React builds on the knowledge from the previous ones [...] It is important that the reader can start with the first article and work their way to the last Learn React article without ever having to "look ahead" for a definition.

So while in practice it may be a good practice to pass a function to the state setter, it overcomplicates things for the reader to talk about it here. Passing a function to the state setter is covered later in ["Queueing a Series of State Updates"].

[`CONTRIBUTING.md`]: https://github.com/reactjs/react.dev/blob/5d2f7105bd6374e465b8bdce8efceaeb8f01c937/CONTRIBUTING.md#guidelines-for-text

["Queueing a Series of State Updates"]: https://react.dev/learn/queueing-a-series-of-state-updates

PR #6602, opened 2024-02-02
suggestion for todo

npiyush97:
fix for todo


rickhanlonii:
Using fetch in an effect doesn't suspend, which would make for really bad docs for the page about using Suspense.

npiyush97:
> Using fetch in an effect doesn't suspend, which would make for really bad docs for the page about using Suspense.

Heat of the moment

PR #6615, opened 2024-02-07
fix: remove useless content

loveloki:
From the above, `as` does not have a `fetch` value


pwbriggs:
Hey there @loveloki 👋

I'd say it's more likely this page *missing* documentation about this `fetch` value than somehow having an extra sentence. Maybe [open an issue] reporting this contradiction instead?

Thanks! ❤️

[open an issue]: https://github.com/reactjs/react.dev/issues/new/choose

sophiebits:
Sorry for the delay but thank you!

PR #6620, opened 2024-02-08
Add useFormStatus pitfall example

mattcarrollcode:
Addresses https://github.com/facebook/react/issues/27980

### DIFF
--- a/src/content/reference/react-dom/hooks/useFormStatus.md
+++ b/src/content/reference/react-dom/hooks/useFormStatus.md
@@ -47,7 +47,7 @@ export default function App() {
 }
 ```
 
-To get status information, the `Submit` component must be rendered within a `<form>`. The Hook returns information like the <CodeStep step={1}>`pending`</CodeStep> property which tells you if the form is actively submitting. 
+To get status information, the `Submit` component must be rendered within a `<form>`. The Hook returns information like the <CodeStep step={1}>`pending`</CodeStep> property which tells you if the form is actively submitting.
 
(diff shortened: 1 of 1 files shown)


src/content/reference/react-dom/hooks/useFormStatus.md
```diff
@@ -165,6 +165,62 @@ function Form() {
...
+  );
+  function handleSubmit(event) {
+    event.preventDefault();
+    setPending(true);
+    startTransition(async () => {
+      optimisticIncreaseCount();
+      const updatedCount = await addCount(count);
+      setCount(updatedCount);
+      setPending(false);
+    });
```
A7med3bdulBaset, line 192:
```suggestion
import { useTransition, useOptimistic } from "react";
import { addCount } from "./api";

export default function Form() {
  const [pending, startTransition] = useTransition();
  const [count, setCount] = useState(0);
  const [optimisticCount, optimisticIncreaseCount] = useOptimistic(
    count,
    (state) => state + 1
  );

  function handleSubmit(event) {
    event.preventDefault();

    startTransition(async () => {
      optimisticIncreaseCount();
      const updatedCount = await addCount(count);
      setCount(updatedCount);
    });
```

eps1lon:
Do we also have a similar pitfall for useContext? I remember one or two cases where people also  didn't know they had to call useContext from within the tree that a Context wraps. 

Maybe it makes sense teach useFormStatus as a kind of useContext where the nearest form is the provider.

PR #6622, opened 2024-02-08
Update index.md

drex44:
Fix the grammar in the documentation.


drex44:
CLA signed but it says it would take about 1 hour.

rickhanlonii:
Nice catch!

PR #6627, opened 2024-02-12
Fix server/client typo in <form> docs

### LINKED ISSUE #6601: [Typo]: Inconsistency in the <form> element docs
### Summary

The "Display a form submission error without JavaScript" section on the form element docs page states the following, among other things:

```
Displaying a [...] error message [...] requires that:
- <form> be rendered by a Server Component
- [...]
- the useFormState Hook be used to display the error message
```

Then there's an example on how to do this. However, there are two concerns with the example:
- The provided example uses a client component, not a server component, so the `<form>` element is being rendered in a client component, breaking the first requirement
- If we pretend that the client component is used instead of a server component for the demonstration purposes only, the component that renders the `<form>` element uses the `useFormState` hook, which is not allowed for server components.

### Page

https://react.dev/reference/react-dom/components/form#display-a-form-submission-error-without-javascript

### Details

_No response_

mdj-uk:
Fixes #6601

With this change the example makes sense, since you can't call `useFormState` in a server component.

### DIFF
--- a/src/content/reference/react-dom/components/form.md
+++ b/src/content/reference/react-dom/components/form.md
@@ -322,7 +322,7 @@ export default function Search() {
 
 Displaying a form submission error message before the JavaScript bundle loads for progressive enhancement requires that:
 
-1. `<form>` be rendered by a [Server Component](/reference/react/use-client)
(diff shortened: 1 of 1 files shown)


rickhanlonii:
@mattcarrollcode can you review?

PR #6635, opened 2024-02-16
docs: add useEffectEvent to the sidebar menu

MatanBobi:
This PR adds the `experimental_useEffectEvent` page to the API reference menu.
As this hook is being referenced in the learn section but the page wasn't visible in the menu (while other new experimental hooks are in the menu), it seemed like this hook is not planned anymore. Adding it to the menu states that this API is something the React core team aims to ship.

image

Thanks for all your hard work! :)

### DIFF
--- a/src/sidebarReference.json
+++ b/src/sidebarReference.json
@@ -40,6 +40,11 @@
           "path": "/reference/react/useEffect"
         },
         {
+          "title": "experimental_useEffectEvent",
+          "path": "/reference/react/experimental_useEffectEvent",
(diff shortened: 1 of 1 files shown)


rickhanlonii:
Hey, sorry for the confusion, I didn't realize this was the page you were suggesting. We may be making some changes to this API before it lands so let's wait to add this page until it's closer to final.

MatanBobi:
Oh, no worries, thanks @rickhanlonii!

PR #6638, opened 2024-02-16
(fix) added closing div to line number 85 for more clarity

KishorKumarParoi:
(fix) added closing div to line number 85 , because before adding another closing div rather than keeping inline closing div sometimes error is showing on the console

### DIFF
--- a/src/content/learn/choosing-the-state-structure.md
+++ b/src/content/learn/choosing-the-state-structure.md
@@ -81,7 +81,8 @@ export default function MovingDot() {
         top: -10,
         width: 20,
         height: 20,
-      }} />
+      }}> 
+      </div>
     </div>
   )
 }


pwbriggs:
Interesting. I don't see why it would be displaying an error. Do you know of any way to reproduce the message? Can you provide a screenshot?

Thanks! ❤️

KishorKumarParoi:
I tried to get that error again. I went back by ctrl+z but I couldn't get the same error again. I have also become surprised. I just copy pasted whole code to run in vs code for my learning of react documentation.

rickhanlonii:
The previous version is correct, if you can share the error I can help you debug but is shouldn't be related to the syntax (and in fact, prettier would remove the tag when there's no child content, preferring the inline close).

PR #6643, opened 2024-02-17
Remove redundant suggestion

tomaszwysocki:
Removed the redundant alternative state management suggestion. The documentation already emphasizes the elimination of duplication by advising to hold the selected ID in state, making the alternative suggestion unnecessary.


rickhanlonii (approved):
Yeah and storing the index could be bad anyway, because if you filter or sort the list the index would be wrong, causing it to get out of sync in the same way storing the item in state would be.

Issue #6489, opened 2023-12-14
Wrong client detection in example


steida:
This:

```js
if (typeof window === 'undefined') {
    throw Error('Chat should only render on the client.');
  }
```

Should be this:

```js
if (typeof document === 'undefined') {
    throw Error('Chat should only render on the client.');
  }
```

Because Deno has window object.

rickhanlonii:
Deno 2.0 is removing `window` because this pattern is way too prevalent https://github.com/denoland/deno/pull/22057

PR #6646, opened 2024-02-18
Wrong client detection in example

praveenzsp:
Issue link : [#6489](https://github.com/reactjs/react.dev/issues/6489)

Updated `window` object to `document` in `src/content/reference/react/Suspense.md` at line no. 2531.


rickhanlonii:
Deno 2.0 is removing window because this pattern is way too prevalent https://github.com/denoland/deno/pull/22057

PR #6651, opened 2024-02-20
Update and fix gh actions

### LINKED ISSUE #5801: [GH workflow] analyze_bundle.txt missed on main branch or the workflow set incorrectly

### LINKED ISSUE #5961: Update GH actions to use latest Node
We get a bunch of warnings, eg on https://github.com/reactjs/react.dev/actions/runs/4791495069.

### LINKED ISSUE #7148: [Bug]: `analyze` workflow fails because of deprecated version of `actions/upload-artifact`
### Summary

[Analyze workflow fails](https://github.com/reactjs/react.dev/actions/runs/10769404480):
>  Error: This request has been automatically failed because it uses a deprecated version of `actions/upload-artifact: v2`. Learn more: https://github.blog/changelog/2024-02-13-deprecation-notice-v1-and-v2-of-the-artifact-actions/

Partially related to #5961

Should be fixed by #6651

### Page

-

### Details

_No response_

alinkedd:
- Updated external actions to the latest version (~~except for the `upload-artifact`~~). Close #6651
  - ~~`upload-artifact` was bumped to v3, as I'm not sure about the required unique artifact names in v4, though v3 still uses node16. Also close #7148~~
- Fixed the warning of uploading missing files if previous comparing analysis steps were canceled. (As an example, you may check this warning in the annotations section here: https://github.com/reactjs/react.dev/actions/runs/7980888987 or directly in the logs here: https://github.com/reactjs/react.dev/actions/runs/7980888987/job/21791474787 in the `Upload analysis comment` step)
- Fixed `Analyze Bundle (Comment)` action failure on the default branch by rewriting the condition, as [GitHub's runner evaluation is tricky and sometimes provides false positives](https://github.com/actions/runner/issues/1173). Close #5801

NOTE: analyze-comment depends on `workflow_run` and thus will be used only after merging updates into the default branch. This is why the deprecation warning is still present (https://github.com/reactjs/react.dev/actions/runs/7980542658). To check the proposed fix in a condition, you may look over the [same fix in my test repo](https://github.com/alinkedd/test-repo/commit/4f3fe0c75aef179ff2927de17c4a554916c9e2fb) and then [dependant workflow respecting that condition](https://github.com/alinkedd/test-repo/actions/runs/7990165729).

Thorough review is welcomed, as I'm no GitHub Actions professional :)

UPD
I added folded block to `if` condition without expression to shorten the line, and it seems to work file. Links to my latest test results are updated.

UPD 2
#7277 has already updated most of the actions, so it closes #5961 by updating dawidd6/action-download-artifact at last

### DIFF
--- a/.github/workflows/analyze.yml
+++ b/.github/workflows/analyze.yml
@@ -55,7 +55,7 @@ jobs:
           name: bundle_analysis.json
 
       - name: Download base branch bundle stats
-        uses: dawidd6/action-download-artifact@v2
+        uses: dawidd6/action-download-artifact@v6
         if: success() && github.event.number
         with:
           workflow: analyze.yml
--- a/.github/workflows/analyze.yml
+++ b/.github/workflows/analyze.yml
@@ -82,6 +82,7 @@ jobs:
 
       - name: Upload analysis comment
         uses: actions/upload-artifact@v4
+        if: success() && github.event.number
         with:
           name: analysis_comment.txt
           path: .next/analyze/__bundle_analysis_comment.txt
--- a/.github/workflows/analyze_comment.yml
+++ b/.github/workflows/analyze_comment.yml
@@ -14,7 +14,7 @@ jobs:
       github.event.workflow_run.conclusion == 'success' }}
     steps:
       - name: Download base branch bundle stats
-        uses: dawidd6/action-download-artifact@v2
+        uses: dawidd6/action-download-artifact@v6
         with:
           workflow: analyze.yml
           run_id: ${{ github.event.workflow_run.id }}
@@ -22,7 +22,7 @@ jobs:
           path: analysis_comment.txt
 
       - name: Download PR number
-        uses: dawidd6/action-download-artifact@v2
+        uses: dawidd6/action-download-artifact@v6
         with:
           workflow: analyze.yml
           run_id: ${{ github.event.workflow_run.id }}
--- a/.github/workflows/analyze_comment.yml
+++ b/.github/workflows/analyze_comment.yml
@@ -10,8 +10,8 @@ jobs:
   comment:
     runs-on: ubuntu-latest
     if: >
-      ${{ github.event.workflow_run.event == 'pull_request' &&
-      github.event.workflow_run.conclusion == 'success' }}
+      github.event.workflow_run.event == 'pull_request' &&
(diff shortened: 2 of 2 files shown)


alinkedd:
@gaearon @harish-sethuraman could you please help to find someone to review this PR regarding GH actions? It would be great to have them fixed.

duplicates https://github.com/reactjs/react.dev/pull/6179#issuecomment-1955047477
fyi @rickhanlonii

alinkedd:
@poteto hi
I noticed you've worked with GitHub Actions. Could you please take a look at this PR as well?

.github/workflows/analyze.yml
```diff
@@ -55,7 +55,7 @@ jobs:
           name: bundle_analysis.json
 
       - name: Download base branch bundle stats
-        uses: dawidd6/action-download-artifact@v2
+        uses: dawidd6/action-download-artifact@v6
```
poteto, line 58: do we really need an external action for this? could we rewrite this to use https://github.com/actions/toolkit/tree/main/packages/artifact instead?

poteto (changes requested):
thanks for the contribution! rather than update the existing external action, i would rather we make use of the github ones. could you help rewrite them to use it instead?

alinkedd:
Close in favor of #7667

.github/workflows/analyze.yml
alinkedd: @poteto I've completely forgot about this one :) I've moved the fixes to #7667 since they're small and ready to go.

However, I'm not available to work on the improvement of switching from `dawidd6/action-download-artifact` to the latest version of `actions/download-artifact`, perhaps it could be converted into a separate issue

PR #6656, opened 2024-02-21
Add Troubleshooting Section to react-developer-tools.md

fkellner:
Taken from https://github.com/facebook/react/issues/27338, these are the helpful points which helped me resolve my issue (old version of Chrome) and which I would have expected to be part of the documentation.


fkellner:
Why would I need to have a CLA for this? Does this not entirely defeat the point of proposing changes with a low barrier of entry? I literally copy-pasted an Answer from a Github issue, so I cannot even say those changes are mine...

PR #6657, opened 2024-02-22
Please carefully look up to this here  as setIsActive is not defined yet but called inside code at line 163

KishorKumarParoi:
Please carefully look up to this here setIsActive is not defined yet at line 163, isActive is only passing by props from Accordion, I know that what I did it's not best solution,I added this lines for proper understanding and updated setIsActive to setCheckActive to remove errors, Please take necessary steps so that learners like myself don't get confused by ReferenceError of variables which isn't declared yet and got curious about why it was like that

### DIFF
--- a/src/content/learn/sharing-state-between-components.md
+++ b/src/content/learn/sharing-state-between-components.md
@@ -147,13 +147,20 @@ export default function Accordion() {
 }
 
 function Panel({ title, children, isActive }) {
+  // Please carefully look up to this here setIsActive is not defined yet
+  // isActive is only passing by props
+  // I added this lines and updated setIsActive to setCheckActive to remove errors
+  // Please take necessary step so that learners like myself don't get confused by error
+  
+  const [checkActive, setCheckActive] = useState(isActive);
+
   return (
     <section className="panel">
       <h3>{title}</h3>
       {isActive ? (
         <p>{children}</p>
       ) : (
-        <button onClick={() => setIsActive(true)}>
+        <button onClick={() => setCheckActive(true)}>
           Show
         </button>
       )}


pwbriggs:
Hey there @KishorKumarParoi! 👋

The error in this code is intentional. It's showing the second step of a 3-step refactoring to [lift state up].

That said, I agree that this may be confusing. Perhaps it would be a good idea to add a comment next to the confusing code? Something like this:

```jsx
function Panel({ title, children, isActive }) {
  return (
    <section className="panel">
      <h3>{title}</h3>
      {isActive ? (
        <p>{children}</p>
      ) : (
        <button onClick={() => setIsActive(true)}>
        // 👆 This onClick handler doesn't work yet, because we removed setIsActive in step 1
          Show
        </button>
      )}
    </section>
  );
}
```

What do you think?

[lift state up]: https://react.dev/learn/sharing-state-between-components

KishorKumarParoi:
> Hey there @KishorKumarParoi! 👋
> 
> The error in this code is intentional. It's showing the second step of a 3-step refactoring to [lift state up](https://react.dev/learn/sharing-state-between-components).
> 
> That said, I agree that this may be confusing. Perhaps it would be a good idea to add a comment next to the confusing code? Something like this:
> 
> ```js
> function Panel({ title, children, isActive }) {
>   return (
>     <section className="panel">
>       <h3>{title}</h3>
>       {isActive ? (
>         <p>{children}</p>
>       ) : (
>         <button onClick={() => setIsActive(true)}>
>         // 👆 This onClick handler doesn't work yet, because we removed setIsActive in step 1
>           Show
>         </button>
>       )}
>     </section>
>   );
> }
> ```
> 
> What do you think?

Yes I am agreed with you

pwbriggs:
@KishorKumarParoi how about you update the PR with these changes then?

KishorKumarParoi:
> @KishorKumarParoi how about you update the PR with these changes then?

Updated Thank you

Mavludin:
Hope this one is going to be merged soon

PR #6658, opened 2024-02-23
Remove useMemo from useFormStatus example

rickhanlonii:
## Overview

The useMemo here breaks the rules of React, is unnecessary and makes a worse experience:

- **Breaks the rules of React**: setting a timer in memo, setting state, and reading/writing to refs are all side effects in render, which is unsafe. useMemo must be a pure function with no side effects.
- **Unnecessary**: the formData is automatically updated as state, so there's no need to store it in a ref ref, you can just read it. The `showSubmitted` state is also unnecessary since the `pending` state already tells you if it's being submitted.
- **Worse experience**: by using a timer, there is an extra state update to show the submitted state (causing a multi-pass render update), and there is an artificial delay between when the form submission completes and when the "Submitted request" text dismisses. By using the `pending` and `formData` values directly, the "Submitted request` text is tied directly to the form submission transition "pending" state, so the disabled state and the text update in the same commit.

## Additional improvements
I also made a couple more improvements:
- CSS to space things out
- Reset the form after submission
- Increase form submission time to 2 seconds

## Before

https://github.com/reactjs/react.dev/assets/2440089/7352d743-a9c7-4025-a16b-26185530af37

## After

https://github.com/reactjs/react.dev/assets/2440089/e70f0603-7be0-44ed-be4a-4122374b65ed

### DIFF
--- a/src/content/reference/react-dom/hooks/useFormStatus.md
+++ b/src/content/reference/react-dom/hooks/useFormStatus.md
@@ -182,36 +182,17 @@ import {useFormStatus} from 'react-dom';
 export default function UsernameForm() {
   const {pending, data} = useFormStatus();
 
-  const [showSubmitted, setShowSubmitted] = useState(false);
-  const submittedUsername = useRef(null);
-  const timeoutId = useRef(null);
-
-  useMemo(() => {
-    if (pending) {
-      submittedUsername.current = data?.get('username');
-      if (timeoutId.current != null) {
-        clearTimeout(timeoutId.current);
-      }
-
-      timeoutId.current = setTimeout(() => {
-        timeoutId.current = null;
-        setShowSubmitted(false);
-      }, 2000);
-      setShowSubmitted(true);
-    }
-  }, [pending, data]);
-
   return (
-    <>
-      <label>Request a Username: </label><br />
-      <input type="text" name="username" />
+    <div>
+      <label>Request a Username: </label>
+      <br />
+      <input type="text" name="username" disabled={pending}/>
       <button type="submit" disabled={pending}>
-        {pending ? 'Submitting...' : 'Submit'}
+        Submit
       </button>
-      {showSubmitted ? (
-        <p>Submitted request for username: {submittedUsername.current}</p>
-      ) : null}
-    </>
+      <br />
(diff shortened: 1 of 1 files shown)


lunaleaps:
I think I added this example and I don't recall why I didn't disable the input while submitting -- is there a reason why we changed that?

lunaleaps (approved):

src/content/reference/react-dom/hooks/useFormStatus.md
```diff
@@ -182,47 +182,33 @@ import {useFormStatus} from 'react-dom';
...
-      setShowSubmitted(true);
-    }
-  }, [pending, data]);
-
   return (
-    <>
-      <label>Request a Username: </label><br />
-      <input type="text" name="username" />
+    <div>
+      <label>Request a Username: </label>
```
sophiebits, line 187: not how the label element works — maybe this can be a heading? (then don't need the `br` below)

sophiebits (approved):
ty

src/content/reference/react-dom/hooks/useFormStatus.md
```diff
@@ -182,47 +182,33 @@ import {useFormStatus} from 'react-dom';
...
       <button type="submit" disabled={pending}>
-        {pending ? 'Submitting...' : 'Submit'}
+        Submit
       </button>
-      {showSubmitted ? (
-        <p>Submitted request for username: {submittedUsername.current}</p>
-      ) : null}
-    </>
+      <br />
+      <p>{pending ? `Requesting ${data?.get("username")}...`: ''}</p>
```
acdlite, line 194: Nit
```suggestion
      <p>{data ? `Requesting ${data.get("username")}...`: ''}</p>
```

src/content/reference/react-dom/hooks/useFormStatus.md
rickhanlonii: nice catch

rickhanlonii:
@lunaleaps if you don't disable the input, then the user can continue typing but when the action finishes the reset will blow away their changes, which is sucky


https://github.com/reactjs/react.dev/assets/2440089/8eff4a67-e087-47d9-9819-ad9aa70d4f54

Issue #6644, opened 2024-02-17
[Bug]: Challenges section not connecting to sandbox


sumit181sk:
### Summary

After solving the challenge correctly, the code gets disconnected from sandbox.
![Screenshot (116)](https://github.com/reactjs/react.dev/assets/70004987/89b64f93-4fc1-4eec-9478-96155f2c9258)

### Page

https://react.dev/learn/writing-markup-with-jsx

esau-morais:
i've got to reproduce once, but it might be related to codesandbox, not to react.dev

pwbriggs:
Reproduced in Chrome, Edge, and Firefox. Sometimes the error appears before the challenge is completely solved, or even with no interaction. I checked the network log in DevTools, and I don't see any failed requests, so I don't think it's (directly) related to the network.

rickhanlonii:
Also confirmed, but no idea what's failing. Have you seen this before @danilowoz?

danilowoz:
It must be a bug in the timeout handle on the Sandpack instance, I will take a look as soon as possible

PR #6659, opened 2024-02-23
fix(sandpack): clear up bundler timeout when there's a syntax error

danilowoz:
Update @codesandbox/sandpack-react version to 2.13.1, which fixes an issue when the bundler responds with a syntax error and the Sandpack instance didn't use to clear the bundler timeout (https://github.com/codesandbox/sandpack/pull/1087)

Close https://github.com/reactjs/react.dev/issues/6644

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "check-all": "npm-run-all prettier lint:fix tsc"
   },
   "dependencies": {
-    "@codesandbox/sandpack-react": "2.6.0",
+    "@codesandbox/sandpack-react": "2.13.1",
     "@docsearch/css": "3.0.0-alpha.41",
     "@docsearch/react": "3.0.0-alpha.41",
     "@headlessui/react": "^1.7.0",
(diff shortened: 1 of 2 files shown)


danilowoz:
Not fixed yet, it must be something else on the Sandpack integration

rickhanlonii:
Damn, nice attempt tho

danilowoz:
Okay, I figured it out! The React.dev uses a pinned Sandpack bundler URL, which makes it more stable and predictable, but on the other hand, it doesn't include any latest fix. 2 months ago, we shipped an important fix on the Sandpack bundler (https://github.com/codesandbox/sandpack-bundler/pull/62), and we didn't update it here.

So, it should be fixed now, and I'll make sure to always update the URL here in case we publish any important fix.

danilowoz:
That's weird, could you please describe how to reproduce it and which page that happens?

rickhanlonii (approved):
Awesome find, thanks for jumping on this so quick

bondz:
@danilowoz  see #6662

Issue #6662, opened 2024-02-26
[Bug]: Issue with Sandpacks


bondz:
### Summary

Sandpack preview no longer reloads after scroll.

Pick any page with a sandpack.... On scroll to the sandpack, the preview loads correctly, scroll away from the sandpack and then back, the sandpack's preview doesn't show up anymore.

### Page

https://react.dev/learn/describing-the-ui

All pages with sandpacks are also affected
### Details

https://github.com/reactjs/react.dev/assets/4683867/58c303b5-a121-409a-a11c-92d17ba201e8

Shown above. The preview loads correctly the first time but if the user navigates away from the sandpack and then later back to it, it fails to load.

Tried to fix it locally and the last working version of sandpack is `2.6.1`. Tested in Safari and Edge

I think this is related to #6659 cc @danilowoz @rickhanlonii

bondz:
All pages with Sandpacks are affected.

danilowoz:
That was unexpected! For now, I'd try to fix the previous bug applying the following change
https://github.com/reactjs/react.dev/pull/6663/files#diff-8dd8662806ac667355c6bff8e8274df548eea48029d2c960160b7adb424501a4

I was able to reproduce it on Sandpack, but no idea what's causing that

bondz:
Thank you @danilowoz 

Just to note that the sandpack for https://react.dev/learn/describing-the-ui#writing-markup-with-jsx

Shows the error the first time but not after scroll. Every other sandpack works correctly.

bondz:
I think it was introduced with this change. With `2.6.0` locally, it continues to show the Error message after the user scrolls away and back.

harish-sethuraman:
In main I can see that the sandpack that @bondz has shared isnt working as expected. Shows blank screen occasionally.
have to confirm if this is affecting other sandpacks as well. From what I have seen it doesnt affect any other. 

checked with older commits and seems like the preview takes time to load even after entering viewport and flickers. It didnt show loader rather when a preview (error) was already there it force replaced the preview again only after entering the viewport. Have attached the video from older commit without your changes (dan's commit) (url: https://react-1ctcjdyrh-fbopensource.vercel.app/)

https://github.com/reactjs/react.dev/assets/32865581/688623c3-3996-4129-b41f-884c2e126400

danilowoz:
Unfortunately, I old PR broke a couple of things in Sandpack. These changes introduced some edge cases and I hope this was the last one 🤞

PR #6664, opened 2024-02-27
fix: Update @codesandbox/sandpack-react version to 2.13.4

danilowoz:
Fixes https://github.com/reactjs/react.dev/issues/6662

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "check-all": "npm-run-all prettier lint:fix tsc"
   },
   "dependencies": {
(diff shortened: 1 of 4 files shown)


danilowoz:
@harish-sethuraman, it should be fixed now. Could you please review it?

harish-sethuraman (approved):

harish-sethuraman:
I think its fine. We can look into the issue if it pops up again. I was not able to reproduce the same error.

PR #6668, opened 2024-02-28
fix: Update sandpack-react version to 2.13.5

danilowoz:
This latest Sandpack version makes sure it unregisters all clients when the sandpack instance is out of viewport.

cc @harish-sethuraman

Closes https://github.com/reactjs/react.dev/issues/6662.


danilowoz:
Pinging @harish-sethuraman and @rickhanlonii

PR #6670, opened 2024-03-01
Add /link redirects

rickhanlonii:
Adds /link redirects so the reactjs.org URLs in warning messages can be changed to react.dev.

### DIFF
--- a/vercel.json
+++ b/vercel.json
@@ -13,6 +13,137 @@
       "source": "/learn/meet-the-team",
       "destination": "/community/team",
       "permanent": true
+    },
+    {
+      "source": "/link/warning-keys",
+      "destination": "/learn/rendering-lists#keeping-list-items-in-order-with-key",
+      "permanent": false
+    },
+
+    {
+      "source": "/link/invalid-hook-call",
+      "destination": "/warnings/invalid-hook-call-warning",
+      "permanent": false
+    },
+    {
+      "source": "/link/hooks-data-fetching",
+      "destination": "/reference/react/useEffect#fetching-data-with-effects",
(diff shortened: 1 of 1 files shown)


vercel.json
```diff
@@ -13,6 +13,137 @@
...
+      "permanent": false
+    },
+    {
+      "source": "/link/dangerously-set-inner-html",
+      "destination": "/reference/react-dom/components/common#dangerously-setting-the-inner-html",
+      "permanent": false
+    },
+    {
+      "source": "/link/event-pooling",
+      "destination": "/reference/react/useEffect#fetching-data-with-effects",
```
eps1lon, line 45: This seems off since we link it from 
```
"This synthetic event is reused for performance reasons. If you're seeing this, " +
          "you're %s `%s` on a released/nullified synthetic event. %s. " +
          'If you must keep the original synthetic event around, use event.persist(). ' +
          'See https://react.dev/link/event-pooling for more information.',
```

eps1lon (approved):
Clicked through all the links printed by `cat vercel.json | jq -r '.redirects[] | "source: " + .source + ", destination: " + (if .destination | startswith("https") then .destination else "https://react.dev" + .destination end)'`

`/link/event-pooling` -> `/reference/react/useEffect#fetching-data-with-effects` seems off. The rest looks good.

vercel.json
rickhanlonii: Nice catch, this must be an existing mistake from here: https://github.com/reactjs/legacy.reactjs.org/blob/main/vercel.json#L266

I'll update to the old event pooling docs

PR #6673, opened 2024-03-02
Update small typo in useDeferredValue.md

ninjaPixel:
There are 4 instances where the phrase "in background" is used on this page. I have updated these to "in the background".

For example, in the following paragraph:
> During the initial render, the returned deferred value will be the same as the value you provided. During updates, React will first attempt a re-render with the old value (so it will return the old value), and then try another re-render in background with the new value (so it will return the updated value).

I have changed the final bit of text from "re-render in background" to "re-render in **the** background".


rickhanlonii (approved):

rickhanlonii:
Thanks!

PR #6680, opened 2024-03-05
Add initial Rules of React docs

poteto:
Rendered: https://react-dev-git-lt-rules-of-react-fbopensource.vercel.app/reference/rules


src/content/reference/rules/index.md
```diff
@@ -0,0 +1,38 @@
...
+
+The upcoming compiler automatically makes writing simple and intuitive React code run efficiently by default. It understands JavaScript semantics and relies on the Rules of React in order to correctly and precisely optimise your codebase.
+
+However, while the compiler can detect most cases of Rules of React breakages, because of the dynamic and flexible nature of JavaScript, it is not possible to exhaustively detect all edge cases. Future iterations of the compiler will continue to improve its static detection, but it's important to understand and follow the Rules of React so that your code continues to run correctly and efficiently even if it's not possible to statically detect.
+
+If the Rules of React are broken, at best the upcoming compiler will skip optimising the components that broke the rules; and at worst, if the breakage is not statically detectable the compiled code may break in unexpected ways.
```
josephsavona, line 24: I know we've iterated on this deep dive a bunch, but it doesn't really answer the question posed in the heading: why the rules are necessary. There's a strong emphasis on React Compiler here that could mislead people into thinking that the rules exist to make the compiler possible. Instead, it seems better to remove references to the compiler (or just lightly mention "the upcoming React Compiler will also depend on these rules"). And then focus more on the core issues of predictability, ability to start/stop, local reasoning, etc.

poteto: Yeah I think you're right, it's too compiler centric right now. Let me take another pass at cleaning this up, or I might consider dropping this deep dive section altogether since it doesn't really add all that much to the page

src/content/reference/rules/components-must-be-idempotent.md
```diff
@@ -0,0 +1,36 @@
...
+
+Put simply, idempotence means that you [always get the same result everytime](learn/keeping-components-pure) you run that piece of code.
+
+```js {3}
+function NewsFeed({ items }) {
+  // ✅ Array.filter doesn't mutate `items`
+  const filteredItems = items.filter(item => item.isDisplayed === true);
+  return (
+    <ul>
+      {filteredItems.map(item => <li>{item.text}</li>}
```
kassens, line 19: missing key

src/content/reference/rules/never-call-component-functions-directly.md
```diff
@@ -0,0 +1,36 @@
...
+  return <Layout><Article /></Layout>; // ✅ Only use components in JSX
+}
+```
+
+```js {3}
+function BlogPost() {
+  ReactDOM.render(
+    Layout({ children: Article() }), // ❌ Never call them directly
+    /* ... */
+  );
```
kassens, line 26: This is a deprecated API, but I think you meant something like this?

```suggestion
  return Layout({ children: Article() }); // ❌ Never call them directly
```

src/content/reference/rules/never-call-component-functions-directly.md
```diff
@@ -0,0 +1,36 @@
...
+}
+```
+
+Letting React orchestrate rendering allows a number of benefits:
```
kassens, line 30: The biggest reason is that if a component contains hooks, it's easy to violate rules of hooks when components are called directly in a loop or conditionally. I think these benefits are very secondary to that?

src/content/reference/rules/never-pass-around-hooks-as-regular-values.md
```diff
@@ -0,0 +1,58 @@
...
+function useDataWithLogging() {
+  // ... Logic should go in here
+}
+```
+
+Hooks should also not be dynamically used: for example, instead of doing dependency injection in a component:
+
+```js {2}
+function ChatInput() {
+  return <Button useData={useDataWithLogging} /> // ❌
```
eps1lon, line 38: I'm not aware that we talked about this rule before publicly. Especially because it doesn't violate the original rules of hooks as far as I can tell.

I checked our codebase for potential violations by just grepping for `use*` as props with `use[A-Z]\w+=`.

Out of the 8000+ components we have (including the ones without any props), only 2 accepted hooks as props. So the rule wouldn't prohibit a popular pattern in our case. 

The two cases where very trivial where you could've just passed the return value of the hook instead of the hook itself:

```diff
-function Child({ useSettings }) {
-   const settings = useSettings
+function Child({ settings }) {
 }


 function Parent() {
+ const settings = useSettings()
-  return <Child useSettings={useSettings} />
+  return <Child settings={settings} />
 }
```

Though this is a slightly different change now because there may be cases where you wanted  to spread hook calls across different components for perf and moving all calls into a single component may be bad since it makes a single render call too expensive.

There's another instance in our codebase where we pass hooks as value but I'm all for breaking that up since nobody understands that piece of code.

eps1lon (commented):
Love it. Finally something to link to 🥳

src/content/reference/rules/never-call-component-functions-directly.md
jackpope: Not sure if the particular pattern matters here but what I've seen most often in product code is the mix of JSX/functions, like:

```
function BlogPost() {
  return <Layout>{Article()}</Layout>
}
```

src/content/reference/rules/never-pass-around-hooks-as-regular-values.md
```diff
@@ -0,0 +1,58 @@
...
+title: Never pass around hooks as regular values
+---
+
+<Intro>
+Hooks should only be called inside of components. Never pass it around as a regular value.
+</Intro>
+
+---
+
+Hooks allow you to augment a component with React features. They should always be called as a function, and never passed around as a regular value. This enables _local reasoning_, or the ability for developers to understand everything a component can do by looking at that component in isolation.
```
jackpope, line 11: > enables local reasoning

This makes it sound like this is less of a rule and more of a recommended pattern. It's unclear what the consequence of the bad example below is.

src/content/reference/rules/never-pass-around-hooks-as-regular-values.md
```diff
@@ -0,0 +1,58 @@
...
+}
+```
+
+```js {2}
+function ChatInput() {
+  const data = useDataWithLogging(); // ✅
+}
+```
+
+Hooks should be immutable and not be mutated. Instead of mutating a hook dynamically, create a static version of the hook with the desired functionality.
```
jackpope, line 26: Should there be a bad example for "mutating a hook dynamically"?

src/content/reference/rules/only-call-hooks-at-the-top-level.md
```diff
@@ -0,0 +1,92 @@
...
+```js
+useEffect(function persistForm() {
+  // 👍 We're not breaking the first rule anymore
+  if (name !== '') {
+    localStorage.setItem('formData', name);
+  }
+});
+```
+
+Note that you don’t need to worry about this problem if you use the provided lint rule. But now you also know why Hooks work this way, and which issues the rule is preventing.
```
jackpope, line 92: Should we promote lint/compiler closer to the top of each page whenever it will catch rule breakages for you?

src/content/reference/rules/side-effects-must-run-outside-of-render.md
```diff
@@ -0,0 +1,73 @@
...
+title: Side effects must run outside of render
+---
+
+<Intro>
+Side effects should not run in render, as React can render components multiple times to create the best possible user experience.
+</Intro>
+
+---
+
+While render must be kept pure, side effects are necessary at some point in order for your app to do anything interesting, like showing something on the screen! The key point of this rule is that side effects should not run in render, as React can render components multiple times. In most cases, you'll use [event handlers](learn/responding-to-events) to handle side effects.
```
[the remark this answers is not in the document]
poteto: I'm using more of the general functional programming version of "side effects" to describe anything interesting happening, not specifically `useEffect`. Basically a program fully containing only pure functions doesn't do anything interesting apart from calculating some result. At some point you need to have a side effect to draw something to a screen.

src/content/reference/rules/index.md
```diff
@@ -0,0 +1,28 @@
...
+These constraints are known as the **Rules of React**. They are rules – and not just guidelines – in the sense that if they are broken, your app likely has bugs. Your code also becomes unidiomatic and harder to understand and reason about.
+
+We strongly recommend using Strict Mode alongside React's ESLint plugin to help your codebase follow the Rules of React. By following the Rules of React, you'll be able to find and address these bugs, as well as prepare your codebase to work out of the box with the upcoming compiler.
+
+<DeepDive>
+#### Why are rules necessary in React? {/*why-are-rules-necessary-in-react*/}
+
+You can think of React's constraints like the grammatical rules and patterns of languages: they constrain what we can do with words, so that we can correctly and efficiently communicate our thoughts.
+
+These rules have been used in the design of all of React's features over the years. React's Strict Mode enforces several of these rules at runtime in DEV mode, and with the release of React's upcoming compiler, more rules will now be statically checked to help you find more bugs as well as allow for correct optimization of your code.
```
pwbriggs, line 20: I don't think there's anywhere else in the docs where you refer to it as "DEV mode," maybe you should clarify?

poteto: Yup good call out, will edit. Thanks!

src/content/reference/rules/only-call-hooks-at-the-top-level.md
poteto: Hmm good idea, I'll do that in a follow up!

src/content/reference/rules/components-and-hooks-must-be-pure.md
```diff
@@ -0,0 +1,275 @@
...
+[Purity](/learn/keeping-components-pure) makes your code easier to understand, debug, and also allows the compiler to optimize your components and hooks correctly.
+</Intro>
+
+<InlineToc />
+
+---
+
+## Components must be idempotent {/*components-must-be-idempotent*/}
+
+React components are assumed to always return the same output with respect to their props. This is known as [idempotency](https://stackoverflow.com/questions/1077412/what-is-an-idempotent-operation).
```
kassens, line 15: Shouldn't this include state and context?

src/sidebarReference.json
```diff
@@ -10,6 +10,24 @@
...
+          "title": "Components and Hooks must be pure",
+          "path": "/reference/rules/components-and-hooks-must-be-pure"
+        },
+        {
+          "title": "React orchestrates Components and Hooks",
+          "path": "/reference/rules/react-orchestrates-components-and-hooks"
+        },
+        {
+          "title": "Rules of Hooks",
+          "path": "/reference/rules/rules-of-hooks"
```
kassens, line 27: Where the rules of hooks unlisted before?

poteto: Seems like we never had them on the new docs site, only in the legacy one

rickhanlonii: There wasn't a nav item for them, but there is an error page for them that is linked to.

The page is here: https://react.dev/warnings/invalid-hook-call-warning

src/content/reference/rules/index.md
```diff
@@ -0,0 +1,27 @@
+---
```
rickhanlonii, line 1: Reminder to update the /reference landing page, which is an index of all the reference pages and where we navigate to when you click "reference" in the navbar: https://react.dev/reference

poteto: Ooh thanks! Will do

src/content/reference/rules/index.md
```diff
@@ -0,0 +1,27 @@
...
+title: Rules of React
+---
+
+<Intro>
+JavaScript rendering libraries and frameworks like React have constraints or "rules" to make the programming model cohesive and easy to reason about, while also helping you prevent bugs in your code. The rules also have the added benefit of creating a safe space for React to optimize and run your code more efficiently. This page lists all the Rules of React.
```
rickhanlonii, line 6: > This page lists all the Rules of React.

We could do what the other section index pages do, which is preview each section:

https://react.dev/reference/react/hooks

src/content/reference/react/index.md
```diff
@@ -10,6 +10,14 @@ This section provides detailed reference documentation for working with React. F
 
 The React reference documentation is broken down into functional subsections:
 
+## Rules of React {/*rules-of-react*/}
+
+JavaScript rendering libraries and frameworks like React have constraints or "rules" to make the programming model cohesive and easy to reason about, while also helping you prevent bugs in your code. The rules also have the added benefit of creating a safe space for React to optimize and run your code more efficiently.
```
josephsavona, line 15: Everything in this paragraph is true, but it also feels like it downplays or is missing the most important piece: _the rules make React what it is_. It isn't that there is using React and separately that there are some rules around it, the rules _define_ React (as a programming model).

As the text here points out, this is also true of other rendering libraries (though I think React's model is particular coherent if you approach it from a more formal perspective). 

One way to reconcile this in the text would be something like:

React is a programming paradigm for declarative user interfaces, in which developers describe what should be rendered given the current state and the framework is responsible for updating the UI. Just as different programming languages have their own ways of expressing concepts, React has its own idioms — or rules — for how to express patterns in a way that is easy to understand and yields high-quality applications.

To learn more about expressing UIs with React, we recommend reading [Thinking in React]. This section describes the React programming paradigm from the perspective of its key rules, which may differ from approaches you've used elsewhere. Many of these rules are common to other UI libraries and non-UI related "reactive" systems, in addition to being a strong approach to building robust software generally.

src/content/reference/react/index.md
```diff
@@ -10,6 +10,14 @@ This section provides detailed reference documentation for working with React. F
...
+
+* [Components and Hooks must be pure](/reference/rules/components-and-hooks-must-be-pure) – Purity makes your code easier to understand, debug, and also allows the compiler to optimize your components and hooks correctly.
```
josephsavona, line 17: I would recommend reducing the emphasis on the compiler, and focusing more heavily on correctness. Purity is foremost about correctness — it avoids bugs by design, which then in turns makes applications easier to understand and debug. I would maybe mention that purity allows React to automatically optimize your applications, not necessarily call out the compiler.

If we do mention "compiler", it should be "the upcoming React Compiler" (which is tedious, hence maybe just talk about potential for optimization instead)

src/content/reference/react/index.md
```diff
@@ -10,6 +10,14 @@ This section provides detailed reference documentation for working with React. F
...
+* [React orchestrates Components and Hooks](/reference/rules/react-orchestrates-components-and-hooks) – React takes care of when your code runs for you so that your application has a great user experience.
```
josephsavona, line 18: Could make this more succinct: "React is responsible for rendering components and hooks when necessary to optimize the user experience."

src/content/reference/react/index.md
```diff
@@ -10,6 +10,14 @@ This section provides detailed reference documentation for working with React. F
...
+* [Rules of Hooks](/reference/rules/rules-of-hooks) – Hooks are JavaScript functions, but you need to follow two rules when using them.
```
josephsavona, line 19: I still think of this as the second category. Hooks are expressed as functions but are conceptually distinct (similar to components). 

But if we're going to keep it here, i would maybe emphasize this fact. "Hooks are defined using JavaScript functions, but they represent a special type of reusable UI logic with restrictions on where they can be called."

src/content/reference/rules/components-and-hooks-must-be-pure.md
```diff
@@ -0,0 +1,275 @@
+---
+title: Components and Hooks must be pure
+---
+
+<Intro>
+[Purity](/learn/keeping-components-pure) makes your code easier to understand, debug, and also allows the compiler to optimize your components and hooks correctly.
```
josephsavona, line 6: same here as the comment above

src/content/reference/rules/components-and-hooks-must-be-pure.md
```diff
@@ -0,0 +1,275 @@
...
+
+<Note>
+Side effects are a broader term than Effects. Effects specifically refer to code that's wrapped in `useEffect`, while a side effect is a general term for code that has any observable effect other than its primary result of returning a value to the caller.
+</Note>
+
+While render must be kept pure, side effects are necessary at some point in order for your app to do anything interesting, like showing something on the screen! The key point of this rule is that side effects should not run in render, as React can render components multiple times. In most cases, you'll use [event handlers](learn/responding-to-events) to handle side effects.
+
+For example, you might have an event handler that displays a confirmation dialog after the user clicks a button. Using an event handler explicitly tells React that this code doesn't need to run during render, keeping render pure. If you've exhausted all options – and only as a last resort – you can also handle side effects using `useEffect`.
+
+<DeepDive>
```
josephsavona, line 58: Maybe lift this deep dive to the top of the page? It felt abrupt to have it in the middle

src/content/reference/rules/components-and-hooks-must-be-pure.md
```diff
@@ -0,0 +1,275 @@
...
+Even though it looks like we are mutating `items`, the key point to note is that this code only does so locally – the mutation isn't "remembered" when the component is rendered again. In other words, `items` only stays around as long as the component does. Because `items` is always recreated every time `<FriendList />` is rendered, the component will always return the same result.
+
+On the other hand, if `items` was created outside of the component, it holds on to its previous values and remembers changes:
+
+```js {1}
+let items = []; // ❌ created outside of the component
+function FriendList({ friends }) {
+  // Push `friends` into `items`...
+  return <section>{items}</section>;
+}
```
josephsavona, line 94: nit: let's show the mutation here, otherwise the text below "continue appending ..." doesn't make sense

src/content/reference/rules/components-and-hooks-must-be-pure.md
```diff
@@ -0,0 +1,275 @@
...
+  SuperCalculator.initializeIfNotReady(); // ✅ Fine if it doesn't affect other components
+  // Continue rendering...
+}
+```
+
+Side effects that are directly visible to the user are not allowed in the render logic of React components. In other words, merely calling a component function shouldn’t by itself produce a change on the screen.
+
+```js {2}
+function ProductDetailPage({ product }) {
+  document.window.title = product.title; // ❌
```
josephsavona, line 112: should we also show the correct version (with a `useEffect()`)? or link to a section about using effects?

src/content/reference/rules/components-and-hooks-must-be-pure.md
```diff
@@ -0,0 +1,275 @@
...
+
+## Props and state are immutable {/*props-and-state-are-immutable*/}
+
+A component's props and state are immutable [snapshots](learn/state-as-a-snapshot) with respect to a single render. Never mutate them directly.
+
+You can think of the props and state values as snapshots that are updated after rendering. For this reason, you don't modify the props or state variables directly: instead you pass new props, or use the setter function provided to you to tell React that state needs to update the next time the component is rendered.
+
+### Don't mutate Props {/*props*/}
+When followed, this rule allows React to understand that values that flow from props aren't mutated when they're passed as arguments to functions, allowing certain optimizations to be made. Mutating props may also indicate a bug in your app – changing values on the props object doesn't cause the component to update, leaving your users with an outdated UI.
```
josephsavona, line 125: This is a bit circular: squinting, it's "the rule allows React to understand that the rule is followed".

Let's maybe update to highlight the correctness issue: props are immutable because if you mutate them the application will produce inconsistent output, which can be hard to debug since it may or may not work depending on the circumstance.

src/content/reference/rules/index.md
josephsavona, line 1: is this file still used?

src/content/reference/rules/index.md
```diff
@@ -0,0 +1,56 @@
...
```
josephsavona, line 6: see comment above — the same text from the reference/react/index.md page could be adapted here

src/content/reference/rules/index.md
```diff
@@ -0,0 +1,56 @@
...
+These rules have been used in the design of all of React's features over the years. Enabling Strict Mode catches bugs caused by breaking these rules, and with the release of React's upcoming compiler, more rules will now be statically checked to help you find more bugs as well as allow for correct optimization of your code.
+
+The Rules of React are proven rules used at companies like Meta that help you maintain an application and codebase that scales with you. When followed, your codebase becomes easier to understand and maintain, is less buggy, and helps React ensure your code runs efficiently by default.
+</DeepDive>
+
+---
+
+## Components and Hooks must be pure {/*components-and-hooks-must-be-pure*/}
+
+[Purity in Components and Hooks](/reference/rules/components-and-hooks-must-be-pure) is a key rule of React that makes your app predictable, easy to debug, and creates a safe space for the compiler to safely optimize your code.
```
josephsavona, line 31: same text here as above, emphasize correctness more and maybe talk about optimization opporunities but not the compiler

src/content/reference/react/index.md
poteto: Thanks, incorporated!

src/content/reference/rules/components-and-hooks-must-be-pure.md
poteto: I was trying to avoid writing any examples with useEffect since we generally want to discourage people to use it unless truly necessary. I think I'll just link to the existing Effect pages since they do a good job explaining!

src/content/reference/rules/components-and-hooks-must-be-pure.md
```diff
@@ -0,0 +1,282 @@
...
+When render is kept pure, React can understand how to prioritize which updates are most important for the user to see first. This is made possible because of render purity: since components don't have side effects in render, React can pause rendering components that aren't as important to update, and only come back to them later when it's needed.
+
+Concretely, this means that rendering logic can be run multiple times in a way that allows React to give your user a pleasant user experience. However, if your component has an untracked side effect – like modifying the value of a global variable during render – when React runs your rendering code again, your side effects will be triggered in a way that won't match what you want. This often leads to unexpected bugs that can degrade how your users experience your app.
+</DeepDive>
+
+---
+
+## Components must be idempotent {/*components-must-be-idempotent*/}
+
+React components are assumed to always return the same output with respect to their inputs – props, state, and context. This is known as [idempotency](https://stackoverflow.com/questions/1077412/what-is-an-idempotent-operation).
```
josephsavona, line 24: nit: passive voice, maybe "React assumes that components always return..."

src/content/reference/rules/components-and-hooks-must-be-pure.md
```diff
@@ -0,0 +1,282 @@
...
+On the other hand, if `items` was created outside of the component, it holds on to its previous values and remembers changes:
+
+```js {1}
+let items = []; // ❌ created outside of the component
+function FriendList({ friends }) {
+  for (let i = 0; i < friends.length; i++) {
+    let friend = friends[i];
+    items.push(
+      <Friend key={friend.id} friend={friend} />
+    );
```
josephsavona, line 96: sorry to be a pain, let's add a " ❌ mutates a value created outside of render" comment here just to be super clear about the problem

src/content/reference/rules/components-and-hooks-must-be-pure.md
```diff
@@ -0,0 +1,282 @@
+---
+title: Components and Hooks must be pure
```
mattcarrollcode, line 2: It might make this page easier to write and understand if you had a small section at the beginning that desribes what purity is, why its important and kicks people off to the /learn/keeping-components-pure page for a refresher and have the rest of the page focused on more philosophy of why components should be pure. Here is an sample high level outline:

What is pure component?
Idempotency
Side effects
Immutability

Right now this page feels like its entangled with /learn/keeping-components-pure and I'm not sure when I would send someone to this page vs /learn/keeping-components-pure. I think creating a very accessible summary, then listing /learn/keeping-components-pure as a "prerequesite" to the rest of the page will allow this page to have its own purpose and identity distinct from /learn/keeping-components-pure.

src/content/reference/rules/components-and-hooks-must-be-pure.md
```diff
@@ -0,0 +1,282 @@
...
+## Components must be idempotent {/*components-must-be-idempotent*/}
+
+React assumes that components always return the same output with respect to their inputs – props, state, and context. This is known as [idempotency](https://stackoverflow.com/questions/1077412/what-is-an-idempotent-operation).
+
+Put simply, idempotence means that you [always get the same result everytime](learn/keeping-components-pure) you run that piece of code.
+
+```js {3}
+function NewsFeed({ items }) {
+  // ✅ Array.filter doesn't mutate `items`
+  const filteredItems = items.filter(item => item.isDisplayed === true);
```
mattcarrollcode, line 31: Can you use a simplier example and make it interactive? Maybe something with numbers so its clear that if you input is num1 you'll always get f(num1) displayed if the component?

src/content/reference/rules/components-and-hooks-must-be-pure.md
```diff
@@ -0,0 +1,282 @@
...
+
+`new Date()` is not idempotent as it always returns the current date and changes its result every time it's called. When you render the above component, the time displayed on the screen will stay stuck on the time that the component was rendered. Similarly, functions like `Math.random()` also aren't idempotent, because they return different results every time they're called, even when the inputs are the same.
+
+This doesn't mean you shouldn't use non-idempotent functions like `new Date()` _at all_ – you should just avoid using them during render. One quick heuristic to tell if code runs during render is to examine where it is: if it's written at the top level like in the example above, there's a good chance it runs during render.
+
+Try building a component that displays the time in real-time in our [challenge](learn/keeping-components-pure#challenges) to see if you follow this rule!
+
+## Side effects must run outside of render {/*side-effects-must-run-outside-of-render*/}
+
+[Side effects](/learn/keeping-components-pure#side-effects-unintended-consequences) should not run in render, as React can render components multiple times to create the best possible user experience.
```
mattcarrollcode, line 57: I think it's worth a quick definition of "side effects" here vs. linking to a subheading in a learn page. Since this whole section is about side effects, making sure everyone is on the same page up front seems like a good place to start to me.

src/content/reference/rules/components-and-hooks-must-be-pure.md
```diff
@@ -0,0 +1,282 @@
...
+
+<Note>
+Side effects are a broader term than Effects. Effects specifically refer to code that's wrapped in `useEffect`, while a side effect is a general term for code that has any observable effect other than its primary result of returning a value to the caller.
+</Note>
+
+While render must be kept pure, side effects are necessary at some point in order for your app to do anything interesting, like showing something on the screen! The key point of this rule is that side effects should not run in render, as React can render components multiple times. In most cases, you'll use [event handlers](learn/responding-to-events) to handle side effects.
```
mattcarrollcode, line 63: > run in render

I don't think we've built up enough context with the reader at this point for them to understand what it means to be "in render"

src/content/reference/rules/components-and-hooks-must-be-pure.md
```diff
@@ -0,0 +1,282 @@
...
+  const filteredItems = items.filter(item => item.isDisplayed === true);
+  return (
+    <ul>
+      {filteredItems.map(item => <li key={item.id}>{item.text}</li>}
+    </ul>
+  );
+}
+```
+
+This means that _all_ code that runs during render must also be idempotent in order for this rule to hold. For example, this line of code is not idempotent (and therefore, neither is the component) and breaks this rule:
```
mattcarrollcode, line 40: > breaks this rule

which rule?

src/content/reference/rules/components-and-hooks-must-be-pure.md
```diff
@@ -0,0 +1,282 @@
...
+      <Friend key={friend.id} friend={friend} />
+    ); // ❌ mutates a value created outside of render
+  }
+  return <section>{items}</section>;
+}
+```
+
+When `<FriendList />` runs again, we will continue appending `friends` to `items` every time that component is run, leading to multiple duplicated results. This version of `<FriendList />` has observable side effects during render and **breaks the rule**.
+
+Similarly, lazy initialization is fine despite not being fully "pure":
```
mattcarrollcode, line 104: > Similarly

What is this similar to? the previous paragraph seems to talk about how a rule is broken but this paragraph says breaking this rule is fine?

src/content/reference/rules/components-and-hooks-must-be-pure.md
```diff
@@ -0,0 +1,282 @@
...
+
+```js {2}
+function ExpenseForm() {
+  SuperCalculator.initializeIfNotReady(); // ✅ Fine if it doesn't affect other components
+  // Continue rendering...
+}
+```
+
+Side effects that are directly visible to the user are not allowed in the render logic of React components. In other words, merely calling a component function shouldn’t by itself produce a change on the screen.
```
mattcarrollcode, line 113: > Side effects that are directly visible to the user are not allowed in the render logic of React components.

How is the reader supposed to know which changes will make a visable change vs. a non-visable change?

src/content/reference/rules/components-and-hooks-must-be-pure.md
```diff
@@ -0,0 +1,282 @@
...
+
+```js {2}
+function ProductDetailPage({ product }) {
+  document.window.title = product.title; // ❌
+}
+```
+
+One way to achieve the desired result of updating `window.title` outside of render is to [synchronize the component with `window`](http://localhost:3000/learn/synchronizing-with-effects).
+
+As long as calling a component multiple times is safe and doesn’t affect the rendering of other components, React doesn’t care if it’s 100% pure in the strict functional programming sense of the word. It is more important that [components must be idempotent](/reference/rules/components-must-be-idempotent).
```
mattcarrollcode, line 123: Link is broken?

mattcarrollcode, line 123: > As long as calling a component multiple times is safe and doesn’t affect the rendering of other components, React doesn’t care if it’s 100% pure in the strict functional programming sense of the word. It is more important that components must be idempotent.

It's not clear to me from this section what the difference is between a component that is "safe" to call multiple times and idempotency. There are some exceptions listed but I'd have a hard time looking at a component and knowing what to do to check if a compnent is "safe" to call multiple times based on this seciton.

src/content/reference/rules/components-and-hooks-must-be-pure.md
```diff
@@ -0,0 +1,282 @@
...
+
+## Props and state are immutable {/*props-and-state-are-immutable*/}
+
+A component's props and state are immutable [snapshots](learn/state-as-a-snapshot) with respect to a single render. Never mutate them directly.
+
+You can think of the props and state values as snapshots that are updated after rendering. For this reason, you don't modify the props or state variables directly: instead you pass new props, or use the setter function provided to you to tell React that state needs to update the next time the component is rendered.
```
mattcarrollcode, line 129: This is excellent and exactly what I was expecting from a page in the rules of react section. We're telling people how React works in a way that is relevent to how they write and think about React code.

src/content/reference/rules/components-and-hooks-must-be-pure.md
```diff
@@ -0,0 +1,282 @@
...
+}
+```
+
+One way to achieve the desired result of updating `window.title` outside of render is to [synchronize the component with `window`](http://localhost:3000/learn/synchronizing-with-effects).
+
+As long as calling a component multiple times is safe and doesn’t affect the rendering of other components, React doesn’t care if it’s 100% pure in the strict functional programming sense of the word. It is more important that [components must be idempotent](/reference/rules/components-must-be-idempotent).
+
+## Props and state are immutable {/*props-and-state-are-immutable*/}
+
+A component's props and state are immutable [snapshots](learn/state-as-a-snapshot) with respect to a single render. Never mutate them directly.
```
mattcarrollcode, line 127: > A component's props and state are immutable [snapshots](learn/state-as-a-snapshot) with respect to a single render. Never mutate them directly.

What is this sentence? Is it the rule? is it a reality of how React works? is it something people should do? what happens if they mutate props or state?

> respect to a single render

I don't think  we've defined what a render is at this point in the doc?

> Never mutate them directly.

What is "them"?

src/sidebarReference.json
```diff
@@ -10,6 +10,24 @@
       "title": "Overview",
       "path": "/reference/react"
     },
+    {
```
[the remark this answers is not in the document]
poteto: Good idea, will incorporate!

src/content/reference/react/index.md
```diff
@@ -10,6 +10,14 @@ This section provides detailed reference documentation for working with React. F
 
 The React reference documentation is broken down into functional subsections:
 
+## Rules of React {/*rules-of-react*/}
+
+React is a programming paradigm for declarative user interfaces, in which developers describe what should be rendered given the current state and the framework is responsible for updating the UI. Just as different programming languages have their own ways of expressing concepts, React has its own idioms — or rules — for how to express patterns in a way that is easy to understand and yields high-quality applications.
```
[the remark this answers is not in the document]
poteto: > Just as different programming languages have their own ways of expressing concepts,

This isn't superfluous, it's an explicit comparison of React to programming languages, which we haven't typically made the case for before. I think it's important to draw this new comparison because it makes the React programming model make sense. I will shorten this section here though because it's just the top level index page, and keep it in the actual Rules of React index.

src/content/reference/react/index.md
```diff
@@ -10,6 +10,14 @@ This section provides detailed reference documentation for working with React. F
...
+
+* [Components and Hooks must be pure](/reference/rules/components-and-hooks-must-be-pure) – Purity makes your code easier to understand, debug, and allows React to automatically optimize your components and hooks correctly.
```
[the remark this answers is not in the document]
poteto: > I don't think we should focus on the complier for the rules of React pages. I think the complier optimization is a byproduct of the values React some of which can be expressed via these rules.

Yeah I've removed all mention of "compiler" (except one) already. I'm not seeing how this specific sentence puts the compiler as the focus though, it's just a short factual sentence of the "consequence" of breaking the rules – one of which is the fact that React can't optimize code that breaks the rules. I don't think we should hide that from people.

src/content/reference/rules/components-and-hooks-must-be-pure.md
```diff
@@ -0,0 +1,282 @@
...
+When render is kept pure, React can understand how to prioritize which updates are most important for the user to see first. This is made possible because of render purity: since components don't have side effects in render, React can pause rendering components that aren't as important to update, and only come back to them later when it's needed.
+
+Concretely, this means that rendering logic can be run multiple times in a way that allows React to give your user a pleasant user experience. However, if your component has an untracked side effect – like modifying the value of a global variable during render – when React runs your rendering code again, your side effects will be triggered in a way that won't match what you want. This often leads to unexpected bugs that can degrade how your users experience your app.
+</DeepDive>
+
+---
+
+## Components must be idempotent {/*components-must-be-idempotent*/}
+
+React assumes that components always return the same output with respect to their inputs – props, state, and context. This is known as [idempotency](https://stackoverflow.com/questions/1077412/what-is-an-idempotent-operation).
```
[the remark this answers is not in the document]
poteto: I can add our own definition

src/content/reference/rules/components-and-hooks-must-be-pure.md
poteto: Could you share more about what's not straightforward about this example? I'm not sure an interactive example would do anything here, it's meant to just illustrate that Array.filter is idempotent.

src/content/reference/rules/components-and-hooks-must-be-pure.md
poteto: It's defined in the note directly below the sentence, do you think that's not clear enough?

src/content/reference/rules/components-and-hooks-must-be-pure.md
```diff
@@ -0,0 +1,273 @@
+---
+title: Components and Hooks must be pure
+---
+
+<Intro>
+[Purity](/learn/keeping-components-pure) makes your code easier to understand, debug, and also allows the compiler to optimize your components and hooks correctly.
```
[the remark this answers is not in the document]
poteto: Yeah I'll try my best to come up with something!

src/content/reference/rules/components-and-hooks-must-be-pure.md
poteto: > What is this sentence? Is it the rule? is it a reality of how React works? is it something people should do? what happens if they mutate props or state?

This sentence is the introduction of this section, the actual rules are covered below

> What is "them"?

I thought it was obvious from the heading and sentence before "props... state" that this referred to the same. Do you want to mention it again, like:

"A component's props and state are immutable [snapshots](https://github.com/reactjs/react.dev/pull/learn/state-as-a-snapshot). Never mutate props and state directly. Instead, pass new props down and use the setter function from `useState`."

src/content/reference/rules/components-and-hooks-must-be-pure.md
```diff
@@ -0,0 +1,283 @@
...
+```js {2}
+function Clock() {
+  const date = new Date(); // ❌ always returns a different result!
+  return <div>{date}</div>
+}
+```
+
+`new Date()` is not idempotent as it always returns the current date and changes its result every time it's called. When you render the above component, the time displayed on the screen will stay stuck on the time that the component was rendered. Similarly, functions like `Math.random()` also aren't idempotent, because they return different results every time they're called, even when the inputs are the same.
+
+This doesn't mean you shouldn't use non-idempotent functions like `new Date()` _at all_ – you should just avoid using them during render. One quick heuristic to tell if code runs during render is to examine where it is: if it's written at the top level like in the example above, there's a good chance it runs during render.
```
harish-sethuraman, line 52: Can we provide an example of how we can use `new Date()` outside of render? I get a little bit struck at this part. May be adding example should provide better understanding?

src/content/reference/rules/components-and-hooks-must-be-pure.md
```diff
@@ -0,0 +1,283 @@
...
+
+<Note>
+Side effects are a broader term than Effects. Effects specifically refer to code that's wrapped in `useEffect`, while a side effect is a general term for code that has any observable effect other than its primary result of returning a value to the caller.
+</Note>
+
+While render must be kept pure, side effects are necessary at some point in order for your app to do anything interesting, like showing something on the screen! The key point of this rule is that side effects should not run in render, as React can render components multiple times. In most cases, you'll use [event handlers](learn/responding-to-events) to handle side effects.
+
+For example, you might have an event handler that displays a confirmation dialog after the user clicks a button. Using an event handler explicitly tells React that this code doesn't need to run during render, keeping render pure. If you've exhausted all options – and only as a last resort – you can also handle side effects using `useEffect`.
```
harish-sethuraman, line 66: possible to add code snippet here as well? 

I feel it would be better to come up with something like [ESLint's docs](https://eslint.org/docs/latest/rules/constructor-super#rule-details) with a correct and incorrect example to show which is better than the other one.

src/content/reference/rules/components-and-hooks-must-be-pure.md
```diff
@@ -0,0 +1,282 @@
...
+
+<InlineToc />
+
+<DeepDive>
+#### Why does render need to be pure? {/*why-does-render-need-to-be-pure*/}
+UI libraries like React take care of when your code runs for you so that your application has a great user experience. React is declarative: you tell React what to render in your component's logic, and React will figure out how best to display it to your user!
+
+When render is kept pure, React can understand how to prioritize which updates are most important for the user to see first. This is made possible because of render purity: since components don't have side effects in render, React can pause rendering components that aren't as important to update, and only come back to them later when it's needed.
+
+Concretely, this means that rendering logic can be run multiple times in a way that allows React to give your user a pleasant user experience. However, if your component has an untracked side effect – like modifying the value of a global variable during render – when React runs your rendering code again, your side effects will be triggered in a way that won't match what you want. This often leads to unexpected bugs that can degrade how your users experience your app.
```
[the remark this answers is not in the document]
poteto: This is the deep dive section though, so presumably someone's reading this because they want to know more.

> Concretely, this means that rendering logic can be run multiple times in a way that allows React to give your user a pleasant user experience.

this sets the background for why pure components are needed: they enable concurrency (although I've purposefully avoided using the term)

> This is a pretty confusing for someone who hasn't encountered this before. Maybe add an example showing a side effect leading to a undesirable outcome?

there already is an example in the [Keeping Components Pure learn page](http://localhost:3000/learn/keeping-components-pure#side-effects-unintended-consequences). I think there's quite a lot of overlap between these 2 pages, but what I'm imagining the purpose of this page is more of a scannable reference for people who are debugging or in the middle of building something. I added a note to the top of this reference page to first read Keeping Components Pure before this page, so hopefully that gives people the required background

src/content/reference/rules/components-and-hooks-must-be-pure.md
```diff
@@ -0,0 +1,282 @@
...
+[Purity](/learn/keeping-components-pure) makes your code easier to understand, debug, and allows React to automatically optimize your components and hooks correctly.
+</Intro>
+
+<InlineToc />
+
+<DeepDive>
+#### Why does render need to be pure? {/*why-does-render-need-to-be-pure*/}
+UI libraries like React take care of when your code runs for you so that your application has a great user experience. React is declarative: you tell React what to render in your component's logic, and React will figure out how best to display it to your user!
+
+When render is kept pure, React can understand how to prioritize which updates are most important for the user to see first. This is made possible because of render purity: since components don't have side effects in render, React can pause rendering components that aren't as important to update, and only come back to them later when it's needed.
```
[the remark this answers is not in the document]
poteto: Clarified that folks should read KCP first

src/content/reference/rules/components-and-hooks-must-be-pure.md
```diff
@@ -0,0 +1,282 @@
+---
+title: Components and Hooks must be pure
+---
+
+<Intro>
+[Purity](/learn/keeping-components-pure) makes your code easier to understand, debug, and allows React to automatically optimize your components and hooks correctly.
```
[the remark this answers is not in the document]
poteto: Done!

src/content/reference/rules/components-and-hooks-must-be-pure.md
poteto: Sure, added one!

src/content/reference/rules/components-and-hooks-must-be-pure.md
poteto: There's already a lot of documentation around `useEffect` and mostly to not use them, so I don't want to add a new one here which might give someone the wrong impression that it's not a bad thing to use them. They're meant to be a last resort.

src/sidebarReference.json
mattcarrollcode, line 1: Can you move this to the bottom (before or after the legacy APIs section) and make each page its own heading?
```suggestion
{
  "title": "API Reference",
  "path": "/reference/react",
  "routes": [
    {
      "hasSectionHeader": true,
      "sectionHeader": "react@18.2.0"
    },
    {
      "title": "Overview",
      "path": "/reference/react"
    },
    {
      "title": "Hooks",
      "path": "/reference/react/hooks",
      "routes": [
        {
          "title": "use",
          "path": "/reference/react/use",
          "canary": true
        },
        {
          "title": "useCallback",
          "path": "/reference/react/useCallback"
        },
        {
          "title": "useContext",
          "path": "/reference/react/useContext"
        },
        {
          "title": "useDebugValue",
          "path": "/reference/react/useDebugValue"
        },
        {
          "title": "useDeferredValue",
          "path": "/reference/react/useDeferredValue"
        },
        {
          "title": "useEffect",
          "path": "/reference/react/useEffect"
        },
        {
          "title": "useId",
          "path": "/reference/react/useId"
        },
        {
          "title": "useImperativeHandle",
          "path": "/reference/react/useImperativeHandle"
        },
        {
          "title": "useInsertionEffect",
          "path": "/reference/react/useInsertionEffect"
        },
        {
          "title": "useLayoutEffect",
          "path": "/reference/react/useLayoutEffect"
        },
        {
          "title": "useMemo",
          "path": "/reference/react/useMemo"
        },
        {
          "title": "useOptimistic",
          "path": "/reference/react/useOptimistic",
          "canary": true
        },
        {
          "title": "useReducer",
          "path": "/reference/react/useReducer"
        },
        {
          "title": "useRef",
          "path": "/reference/react/useRef"
        },
        {
          "title": "useState",
          "path": "/reference/react/useState"
        },
        {
          "title": "useSyncExternalStore",
          "path": "/reference/react/useSyncExternalStore"
        },
        {
          "title": "useTransition",
          "path": "/reference/react/useTransition"
        }
      ]
    },
    {
      "title": "Components",
      "path": "/reference/react/components",
      "routes": [
        {
          "title": "<Fragment> (<>)",
          "path": "/reference/react/Fragment"
        },
        {
          "title": "<Profiler>",
          "path": "/reference/react/Profiler"
        },
        {
          "title": "<StrictMode>",
          "path": "/reference/react/StrictMode"
        },
        {
          "title": "<Suspense>",
          "path": "/reference/react/Suspense"
        }
      ]
    },
    {
      "title": "APIs",
      "path": "/reference/react/apis",
      "routes": [
        {
          "title": "cache",
          "path": "/reference/react/cache",
          "canary": true
        },
        {
          "title": "createContext",
          "path": "/reference/react/createContext"
        },
        {
          "title": "forwardRef",
          "path": "/reference/react/forwardRef"
        },
        {
          "title": "lazy",
          "path": "/reference/react/lazy"
        },
        {
          "title": "memo",
          "path": "/reference/react/memo"
        },
        {
          "title": "startTransition",
          "path": "/reference/react/startTransition"
        },
        {
          "title": "experimental_taintObjectReference",
          "path": "/reference/react/experimental_taintObjectReference",
          "canary": true
        },
        {
          "title": "experimental_taintUniqueValue",
          "path": "/reference/react/experimental_taintUniqueValue",
          "canary": true
        }
      ]
    },
    {
      "title": "Directives",
      "path": "/reference/react/directives",
      "canary": true,
      "routes": [
        {
          "title": "'use client'",
          "path": "/reference/react/use-client",
          "canary": true
        },
        {
          "title": "'use server'",
          "path": "/reference/react/use-server",
          "canary": true
        }
      ]
    },
    {
      "hasSectionHeader": true,
      "sectionHeader": "react-dom@18.2.0"
    },
    {
      "title": "Hooks",
      "path": "/reference/react-dom/hooks",
      "routes": [
        {
          "title": "useFormState",
          "path": "/reference/react-dom/hooks/useFormState",
          "canary": true
        },
        {
          "title": "useFormStatus",
          "path": "/reference/react-dom/hooks/useFormStatus",
          "canary": true
        }
      ]
    },
    {
      "title": "Components",
      "path": "/reference/react-dom/components",
      "routes": [
        {
          "title": "Common (e.g. <div>)",
          "path": "/reference/react-dom/components/common"
        },
        {
          "title": "<form>",
          "path": "/reference/react-dom/components/form",
          "canary": true
        },
        {
          "title": "<input>",
          "path": "/reference/react-dom/components/input"
        },
        {
          "title": "<option>",
          "path": "/reference/react-dom/components/option"
        },
        {
          "title": "<progress>",
          "path": "/reference/react-dom/components/progress"
        },
        {
          "title": "<select>",
          "path": "/reference/react-dom/components/select"
        },
        {
          "title": "<textarea>",
          "path": "/reference/react-dom/components/textarea"
        },
        {
          "title": "<link>",
          "path": "/reference/react-dom/components/link",
          "canary": true
        },
        {
          "title": "<meta>",
          "path": "/reference/react-dom/components/meta",
          "canary": true
        },
        {
          "title": "<script>",
          "path": "/reference/react-dom/components/script",
          "canary": true
        },
        {
          "title": "<style>",
          "path": "/reference/react-dom/components/style",
          "canary": true
        },
        {
          "title": "<title>",
          "path": "/reference/react-dom/components/title",
          "canary": true
        }
      ]
    },
    {
      "title": "APIs",
      "path": "/reference/react-dom",
      "routes": [
        {
          "title": "createPortal",
          "path": "/reference/react-dom/createPortal"
        },
        {
          "title": "flushSync",
          "path": "/reference/react-dom/flushSync"
        },
        {
          "title": "findDOMNode",
          "path": "/reference/react-dom/findDOMNode"
        },
        {
          "title": "hydrate",
          "path": "/reference/react-dom/hydrate"
        },
        {
          "title": "preconnect",
          "path": "/reference/react-dom/preconnect",
          "canary": true
        },
        {
          "title": "prefetchDNS",
          "path": "/reference/react-dom/prefetchDNS",
          "canary": true
        },
        {
          "title": "preinit",
          "path": "/reference/react-dom/preinit",
          "canary": true
        },
        {
          "title": "preinitModule",
          "path": "/reference/react-dom/preinitModule",
          "canary": true
        },
        {
          "title": "preload",
          "path": "/reference/react-dom/preload",
          "canary": true
        },
        {
          "title": "preloadModule",
          "path": "/reference/react-dom/preloadModule",
          "canary": true
        },
        {
          "title": "render",
          "path": "/reference/react-dom/render"
        },
        {
          "title": "unmountComponentAtNode",
          "path": "/reference/react-dom/unmountComponentAtNode"
        }
      ]
    },
    {
      "title": "Client APIs",
      "path": "/reference/react-dom/client",
      "routes": [
        {
          "title": "createRoot",
          "path": "/reference/react-dom/client/createRoot"
        },
        {
          "title": "hydrateRoot",
          "path": "/reference/react-dom/client/hydrateRoot"
        }
      ]
    },
    {
      "title": "Server APIs",
      "path": "/reference/react-dom/server",
      "routes": [
        {
          "title": "renderToNodeStream",
          "path": "/reference/react-dom/server/renderToNodeStream"
        },
        {
          "title": "renderToPipeableStream",
          "path": "/reference/react-dom/server/renderToPipeableStream"
        },
        {
          "title": "renderToReadableStream",
          "path": "/reference/react-dom/server/renderToReadableStream"
        },
        {
          "title": "renderToStaticMarkup",
          "path": "/reference/react-dom/server/renderToStaticMarkup"
        },
        {
          "title": "renderToStaticNodeStream",
          "path": "/reference/react-dom/server/renderToStaticNodeStream"
        },
        {
          "title": "renderToString",
          "path": "/reference/react-dom/server/renderToString"
        }
      ]
    },
    {
      "hasSectionHeader": true,
      "sectionHeader": "Rules of React"
    },
    {
      "title": "Overview",
      "path": "/reference/rules"
    },
    {
      "title": "Components and Hooks must be pure",
      "path": "/reference/rules/components-and-hooks-must-be-pure"
    },
    {
      "title": "React calls Components and Hooks",
      "path": "/reference/rules/react-calls-components-and-hooks"
    },
    {
      "title": "Rules of Hooks",
      "path": "/reference/rules/rules-of-hooks"
    },
    {
      "hasSectionHeader": true,
      "sectionHeader": "Legacy APIs"
    },
    {
      "title": "Legacy React APIs",
      "path": "/reference/react/legacy",
      "routes": [
        {
          "title": "Children",
          "path": "/reference/react/Children"
        },
        {
          "title": "cloneElement",
          "path": "/reference/react/cloneElement"
        },
        {
          "title": "Component",
          "path": "/reference/react/Component"
        },
        {
          "title": "createElement",
          "path": "/reference/react/createElement"
        },
        {
          "title": "createFactory",
          "path": "/reference/react/createFactory"
        },
        {
          "title": "createRef",
          "path": "/reference/react/createRef"
        },
        {
          "title": "isValidElement",
          "path": "/reference/react/isValidElement"
        },
        {
          "title": "PureComponent",
          "path": "/reference/react/PureComponent"
        }
      ]
    }
  ]
}

```

src/content/reference/rules/index.md
```diff
@@ -0,0 +1,60 @@
+---
+title: Rules of React
+---
+
+<Intro>
+Just as different programming languages have their own ways of expressing concepts, React has its own idioms — or rules — for how to express patterns in a way that is easy to understand and yields high-quality applications. This page lists all the Rules of React.
```
mattcarrollcode, line 6: > This page lists all the Rules of React.

I think this should be removed. It would be easy for someone to misinterpret this as saying that the rules of React won't ever change/be remove/added/evolve or if you follow these rules your code will always be good.

src/content/reference/rules/index.md
```diff
@@ -0,0 +1,60 @@
...
+</Intro>
+
+<InlineToc />
+
+---
+
+<Note>
+To learn more about expressing UIs with React, we recommend reading [Thinking in React](/learn/thinking-in-react). This section describes the React programming paradigm from the perspective of its key rules, which may differ from approaches you've used elsewhere. Many of these rules are common to other UI libraries and non-UI related "reactive" systems, in addition to being a strong approach to building robust software generally.
```
mattcarrollcode, line 14: > This section describes the React programming paradigm from the perspective of its key rules, which may differ from approaches you've used elsewhere. Many of these rules are common to other UI libraries and non-UI related "reactive" systems, in addition to being a strong approach to building robust software generally.

I think taking another approach to these sentences would help readers that aren't React experts understand what this section is about. I'd rewrite this to say something more along these lines:


This section describes the rules you need to follow to write idiomatic React code. Writing idiomatic React code can help you write well organized, safe, and composable applications. These properties make your app more resilient to changes and makes it easier to work with other developers, libraries, and tools.

src/sidebarReference.json
poteto: Done! Moved it above Legacy APIs

src/content/reference/rules/index.md
poteto: Love it, incorporated!

PR #6686, opened 2024-03-08
Fixed useEffectEvent import path under escape hatch section

TheMaskedGuy:
Ignore this PR If the intention is to have the docs have final import for useEffectEvent api (which is not valid as api is in experimental stage and thus has different import path than the currently specified in docs)

If the above case is true, then I think it would be better to specify the current experimental import statement (for copying or learning) in the "under construction" banner or a comment or as a note as the real import is specified very down below in challenges section.

tl;dr - Updated the useEffectEvent import path to reflect the current experimental stage.


rickhanlonii:
Yeah the goal is to show the example with the final API

PR #6691, opened 2024-03-11
Changed the documentation of the subscribe argument to a more accurat…

Tinttori:
The documentation of the subscribe argument in useSyncExternalStore was inaccurate. Now it says the following:

subscribe: A function that takes a single callback argument and subscribes it to the store. When the store changes, it should invoke the provided callback. This alerts React to potential changes in the store, but it will only cause the component to re-render if the snapshot, evaluated subsequently, has also changed. The subscribe function should return a function that removes the subscription when it's no longer needed. This helps prevent unnecessary re-renders and optimizes performance.

### DIFF
--- a/src/content/reference/react/useSyncExternalStore.md
+++ b/src/content/reference/react/useSyncExternalStore.md
@@ -41,7 +41,7 @@ It returns the snapshot of the data in the store. You need to pass two functions
 
 #### Parameters {/*parameters*/}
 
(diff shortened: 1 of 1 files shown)


pwbriggs:
@Tinttori, I may be wrong so don't quote me on this, but I always thought "rendering" is the process of *running the component functions,* and React updates the DOM when the *result* of rendering is different.

sophiebits (approved):
thanks for the suggestion! I made a couple tweaks

PR #6696, opened 2024-03-14
Remove links to localhost

poteto:
It seems like these localhost references were being stripped out anyway, so this isn't a bugfix but just a small cleanup.


kassens:
shipit

Issue #6713, opened 2024-03-27
Capitalize React concepts in docs


rickhanlonii:
### Summary

We should be capitalizing proper React concepts in the docs. This was flagged by @samselikoff [here](https://x.com/samselikoff/status/1772448138449596745?s=20) and I took a first pass at correct Effects, Hooks, and Transitions.

### Page

All of them

## Remaining items

The remaing items are:

- [ ] Components
- [ ] State
- [ ] Actions
- [ ] Context
- [ ] Suspsense
- [ ] Error Boundaries
- [ ] ... more?

## Submitting a PR

**Please do not ask to take an item, just submit a PR.** If multiple people submit the same PR, I'll add co-author credit to the commit.

When you submit a PR, please try and cover all usages in all the docs, on all pages. Avoid fixing one or two instances and favor full coverage to minimize the number of PRs and reviews.

Note that headings like `{/* usestate */}` must stay lower case, so a find and replace won't work.

anthonylaflamme:
I would like to take a go at this any first read I should do for my first contribution ? I read this already : https://github.com/reactjs/react.dev/blob/4df3124fa2116cdb228caa0f1f05a3449b53bd09/CONTRIBUTING.md

dlitsman:
Curious if Props should also be capitalized similar to State/Context?

Also, updated my pull requests to remove blog/* changes

smikitky:
I would like to kindly remind you that most languages that do not belong to the [Indo-European language family](https://en.wikipedia.org/wiki/Indo-European_languages) don't have the concept of uppercase/lowercase to begin with. Many languages in the world are [unicase](https://en.wikipedia.org/wiki/Unicase).

I agree with introducing distinctions between terms like "state" and "State" or "(side) effect" and "Effect", as these have actually confused developers for many years. However, I hope you will proceed with caution when continuously introducing new uppercase React concepts, such as "Action" or "Offscreen".

In many languages, including my own, it's really challenging to coming up with separate translations for "action" and "Action" without causing confusion among readers. Sometimes we just give up on the translation altogether and denote the React-specific concepts like `Action`, in English, but this is not ideal.

kassens:
@smikitky this must be seen as a style guide for English where these rules are often used. It also doesn't even apply in German which I know because all nouns like Action would always be capitalized.
Other translations should develop their own style guide. It's good to think about the style and do It consistently for one language so it's matching the norm of that language and "Action" for example (if translated) would have 1 translation for the language.

[...]

PR #6717, opened 2024-03-27
fix: Fix word component(s) capitalization

anthonylaflamme:
This pr fixes the word component(s) capitalization for issue #6713


anthonylaflamme:
just signed the claim

anthonylaflamme:
there is two instances of the word "subcomponent" should these be sub Components or we leave them as is ?

anthonylaflamme:
I also wonder about places where we talk about HTML Element but refer to them as component ex: 

" The [built-in browser `<link>` **Component**](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link) lets you use external resources such as stylesheets or annotate the document with link metadata."

This might be out of scope for this MR but should these be Element instead of Component ?

rickhanlonii:
@anthonylaflamme thanks for taking a stab!

Can you revert the homepage, /blog and /community changes? Let's just keep those as-is.

Can you also revert the capitalization for things that are not referring to the React Component concept directly? I update https://github.com/reactjs/react.dev/issues/6713 with a style guide to try and make it clearer when to capitalize.

Some example from this PR that should not be capitalized:
- "the `<link>` component"
- "such as preserving `this` context inside a component"
- "can update 50,000+ components alone"
- "upgrade your apps one component at a time"

Admittedly this is tricky, my mental model for this is "would it make sense to link to the docs for components here" or "could this be replaced with 'React Component'"?

anthonylaflamme:
Thank you for making this more precise. I will continue to update them as I read through the context of some of them I should be able to pull this in a timely manner.

src/content/learn/preserving-and-resetting-state.md
```diff
@@ -561,11 +561,11 @@ label {
 
 </Sandpack>
 
-Here, you switch between _different_ component types at the same position. Initially, the first child of the `<div>` contained a `Counter`. But when you swapped in a `p`, React removed the `Counter` from the UI tree and destroyed its state.
+Here, you switch between _different_ Component types at the same position. Initially, the first child of the `<div>` contained a `Counter`. But when you swapped in a `p`, React removed the `Counter` from the UI tree and destroyed its state.
```
anthonylaflamme, line 564: This one feels tricky @rickhanlonii do you think this should be Component or component ? Since we are talking about the react context here but also referencing one React Component (Counter) but also built ins (<p>) ?

Gitstar-OC:
Do I have to do this with all the file's out there, Tell me and assign this issue I will do it!

PR #6727, opened 2024-03-30
chore: update link for code of conduct

ahmoin:
`code.facebook.com/codeofconduct` -> `opensource.fb.com/code-of-conduct`


rickhanlonii:
Let's keep it the same, I think the Meta OSS team keeps this in sync with the redirects and will update as necessary.

rickhanlonii:
Let's keep it the same, I think the Meta OSS team keeps this in sync with the redirects and will update as necessary.

PR #6737, opened 2024-04-06
fix(rules-of-react): "everytime" -> "every time"

SiarheiBobryk:
As far as I remember the grammar, _every time_ is not an acceptable compound word, although it's fine to say _everywhere_, _everyday_ or _everyone_.

Please correct me if I missed anything.


eps1lon (approved):
Seems right: https://www.grammarly.com/blog/everytime-every-time/

Thanks

PR #6743, opened 2024-04-09
Fixes #6693: fixes search functionality on iOS devices

### LINKED ISSUE #6693: [Bug]: Search functionality not working correctly on react.dev in Safari on iOS real device (iPhone 12 Pro)
## The current behavior
Search functionality does not work correctly on iOS real device (iPhone 12 Pro). 

https://github.com/reactjs/react.dev/assets/57572988/f50ff514-d84b-4e8c-b067-628eedd2d762

Since search functionality is currently working as expected on macOS, both in Chrome and Safari, the issue might be in implementation of [Algolia InstantSearch iOS](https://www.algolia.com/doc/guides/building-search-ui/what-is-instantsearch/ios/):

https://github.com/reactjs/react.dev/assets/57572988/44ddea01-cd10-4323-bcdb-aec1788da64d

---

## The expected behavior

Search functionality is working as expected on macOS, both in Chrome and Safari. Should work similarly on iOS real device:

**macOs Chrome**

https://github.com/reactjs/react.dev/assets/57572988/41c9cd76-3ac8-4820-9917-52141d68c8ed

**macOS Safari**

https://github.com/reactjs/react.dev/assets/57572988/4daf9a68-2671-446c-a9c3-8a39bfefc8e5

### Page

https://react.dev

## Steps To Reproduce

1. Open react.dev in Safari on iPhone (real device)
2. Tap search icon: note that search function is working as intended
3. Scroll down to "Write components with code and markup" section
4. Tap search icon: note that action doesn't open search
5. Scroll down again and tap the search icon several times:  note that action doesn't open search
6. Repeat step 5: note that eventually screen freezes
___

The iOS real device that this behavior was recorded on is using the current iOS version:

react_github-bug-report_ios-version

MeetYourCreator:
Fixes search functionality on iOS devices

Fixes #6693
___

Current: Behavior of search functionality on iPhone 12 Pro

https://github.com/reactjs/react.dev/assets/57572988/89f03c9f-7d66-4bb8-914c-a11e3e16dcfe

___

Fixes #6693: Behavior of search functionality on iPhone 12 Pro 

https://github.com/reactjs/react.dev/assets/57572988/0f8b9b75-082c-4d95-840b-55d8bb72659f

### DIFF
--- a/src/components/Search.tsx
+++ b/src/components/Search.tsx
@@ -110,7 +110,6 @@ export function Search({
         createPortal(
           <DocSearchModal
             {...options}
-            initialScrollY={window.scrollY}
             searchParameters={searchParameters}
             onClose={onClose}
             navigator={{
--- a/src/components/Layout/TopNav/TopNav.tsx
+++ b/src/components/Layout/TopNav/TopNav.tsx
@@ -142,7 +142,7 @@ export default function TopNav({
   breadcrumbs: RouteItem[];
   section: 'learn' | 'reference' | 'community' | 'blog' | 'home' | 'unknown';
 }) {
-  const [isOpen, setIsOpen] = useState(false);
+  const [isMenuOpen, setIsMenuOpen] = useState(false);
   const scrollParentRef = useRef<HTMLDivElement>(null);
   const {asPath} = useRouter();
   const [isScrolled, setIsScrolled] = useState(false);
@@ -154,18 +154,18 @@ export default function TopNav({
 
   // While the overlay is open, disable body scroll.
   useEffect(() => {
-    if (isOpen) {
+    if (isMenuOpen) {
       const preferredScrollParent = scrollParentRef.current!;
       disableBodyScroll(preferredScrollParent);
       return () => enableBodyScroll(preferredScrollParent);
     } else {
       return undefined;
     }
-  }, [isOpen]);
+  }, [isMenuOpen]);
 
   // Close the overlay on any navigation.
   useEffect(() => {
-    setIsOpen(false);
+    setIsMenuOpen(false);
   }, [asPath]);
 
   // Also close the overlay if the window gets resized past mobile layout.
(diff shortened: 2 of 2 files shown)


src/components/Search.tsx
```diff
@@ -110,7 +110,6 @@ export function Search({
         createPortal(
           <DocSearchModal
             {...options}
-            initialScrollY={window.scrollY}
```
rickhanlonii, line 113: Awesome stuff.

is this the actual fix? Can you explain what the problem was and how this fixes it?

MeetYourCreator: Hi @rickhanlonii 👋

Thanks so much! Yep, line 113 is the actual fix. I didn't take a look into the internals of the library, but it looks like the bug might actually be in the [Autocomplete library](https://autocomplete-experimental.netlify.app/docs/docsearchmodal/).  It looks like their DocSearchModal component has a bug in the modal design, which is causing the screen to freeze once the device width is less than about 755px.

https://github.com/reactjs/react.dev/assets/57572988/305696e2-52b5-4b7b-a362-cb66140d0ebe

MeetYourCreator: Hi @rickhanlonii 👋

Thanks so much! Yep, line 113 is the actual fix. I didn't take a look into the internals of the library, but it looks like the bug might actually be in the [Autocomplete library](https://autocomplete-experimental.netlify.app/docs/docsearchmodal/).  It looks like their DocSearchModal component has a bug in the modal design, which is causing the screen to freeze once the device width is less than about 755px.

https://github.com/reactjs/react.dev/assets/57572988/305696e2-52b5-4b7b-a362-cb66140d0ebe

rickhanlonii (approved):
LGTM - I tested on desktop and mobile to confirm nothing broke, and the fix works.

If you can sign the CLA I'll merge!

MeetYourCreator:
Thanks so much, @rickhanlonii 😀

PR #6746, opened 2024-04-10
Revert inappropriate capitalization of "Transition"

smikitky:
PR #6712 capitalized instances of "transition" to "Transition". I found one instance that shouldn't have been capitalized.

https://github.com/reactjs/react.dev/blob/a7103d8cc2ee8b10e314aea096538b5057e0ba10/src/content/reference/react/Suspense.md?plain=1#L1744

Here, "transition" in "state transition" does not refer to the React mechanism itself, so it should remain lowercase.

@rickhanlonii

### DIFF
--- a/src/content/reference/react/Suspense.md
+++ b/src/content/reference/react/Suspense.md
@@ -1741,7 +1741,7 @@ function Router() {
   // ...
 ```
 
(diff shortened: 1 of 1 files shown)


rickhanlonii (approved):

rickhanlonii:
Nice catch

PR #6756, opened 2024-04-15
Remove canary label from cache

kassens:
An idea how to signal the version support of `cache`.


kassens:
Alternative proposal in #6757 which I think I like better. It puts the change at the bottom to keep the eyes on the current API.

PR #6757, opened 2024-04-15
Add Changelog for cache

kassens:
An alternative proposal to add a changelog to the bottom of the page to show newly added APIs.

This is similar to

- php.net: https://www.php.net/manual/en/function.get-magic-quotes-runtime.php#refsect1-function.get-magic-quotes-runtime-changelog
- MDN: https://developer.mozilla.org/en-US/docs/Web/API/Node/insertBefore#browser_compatibility

This approach in having 1 evergreen docs page reduces a lot of headache of having older versions that might still want occasional updates, especially with a very stable API like React.

### DIFF
--- a/src/content/reference/react/cache.md
+++ b/src/content/reference/react/cache.md
@@ -1,13 +1,6 @@
 ---
 title: cache
-canary: true
 ---
-
-<Canary>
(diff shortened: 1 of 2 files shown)


rickhanlonii:
I like it, and we can make the later versions link to a Version History page, which we keep a pinned docs version for, for removed APIs. The Version History page would list probably just the major versions with a link to the docs for those major versions. 

It can also link to the changelog entries for each version, so maybe we do include all versions there at least until the page is enormous?

PR #6763, opened 2024-04-19
Update SocialBanner.tsx

kassens:
Tickets seem to be sold out, update the text.


kassens:
@mattcarrollcode what do you think?

mattcarrollcode (approved):

rickhanlonii:
Fwiw "Join us" was inclusive of btoh people who watch the stream or attend in person, so I feel like it was better before.

PR #6769, opened 2024-04-23
fix: appropriately set prevRoute and nextRoute in getRouteMeta.tsx

idorenyinudoh:
at the moment, on the blog, the link to the previous route is actually the next route and that of the next route, is the article before the current one (the open page).

that default comes from this file and to fix it, i set `prevRoute` to `ctx.route` initially and when the current route is found, `prevRoute` is assigned the last breadcrumb if the current index is greater than 1.

for example, in this screenshot below, the next article was published in March 2023, while the previous was in February 2024. Meanwhile, it should be the other way round

Screenshot 2024-04-23 at 10 42 44

### DIFF
--- a/src/components/Layout/getRouteMeta.tsx
+++ b/src/components/Layout/getRouteMeta.tsx
@@ -85,8 +85,8 @@ function buildRouteMeta(
 
   const {routes} = currentRoute;
 
-  if (ctx.route && !ctx.nextRoute) {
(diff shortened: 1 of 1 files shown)


rickhanlonii:
In the preview it looks like `prev === current === next`?

Screenshot 2024-05-18 at 8 21 39 PM
Screenshot 2024-05-18 at 8 21 22 PM

PR #6770, opened 2024-04-23
Add documentation for ref cleanup functions

jackpope:


noahlemen (commented):
the content of this looks good to me overall, but considering versioning, maybe it would be best to update _just_ the `reference/` page, with _all_ the changes within a `CanaryBadge` (that is, the information about cleanup functions / ref callbacks called with null would also be broken out to a canary badge)

then, after release, we'd update the`learn/` page as well? it seems like so far `learn/` does not make reference to the canary, and that's probably not relevant information for that audience until a release.

jackpope:
Updated reference changes to be inside a Canary UI block. Also reverted the null argument change in the list example, but kept the other changes since the old image service is broken and the updates make it easier for us to drop in the cleanup function usage later.

src/content/learn/manipulating-the-dom-with-refs.md
```diff
@@ -325,13 +342,13 @@ In this example, `itemsRef` doesn't hold a single DOM node. Instead, it holds a
...
   ref={node => {
     const map = getMap();
-    if (node) {
-      // Add to the Map
-      map.set(cat.id, node);
-    } else {
-      // Remove from the Map
-      map.delete(cat.id);
-    }
+    // Add to the map
```
rickhanlonii, line 345: This doesn't indicate canary? I also don't understand why we'd need to add/remove cats?

rickhanlonii (approved):
Looking good, left a couple nits

src/content/learn/manipulating-the-dom-with-refs.md
jackpope: Ah thanks I missed this spot. Updating to a Canary block. 

For the add/remove, this was to make it more obvious how/when the cleanups are fired if you want to extend the example with logs or other behavior. Was helpful when putting it together but I'll update to simplify here.

src/content/learn/manipulating-the-dom-with-refs.md
```diff
@@ -244,34 +245,25 @@ export default function CatFriends() {
...
-                  map.delete(cat.id);
+                  map.delete(cat);
                 }
               }}
             >
-              <img
-                src={cat.imageUrl}
-                alt={'Cat #' + cat.id}
-              />
+              <img src={cat} />
```
noahlemen, line 266: should we leave all this as is? and keep the `{ id, imageUrl }` structure in `setupCatList`? while this change is probably what i would prefer in actual code, i think the object makes things a bit more immediately grokkable in the context of examples

noahlemen: whoops, didn't see you already merged. i don't feel strongly about this.

PR #6771, opened 2024-04-23
RFC for a docs version selector

kassens:
An idea how we could do versioning of docs that I think might be sustainable.

- In the reference we have a version selector (design TBD)
- `sidebarReference.json` can contain a version now to conditionally include entries
- a `<VersionCondidion />` component allows to conditionally include content in a doc when APIs change over time

Code quality is very rough. Would probably put the version into the path somewhere, but that needs some more changes.


rickhanlonii:
Replaced by https://github.com/reactjs/react.dev/pull/6814

PR #6774, opened 2024-04-24
Move `use` to APIs

kassens:
`use` is not a Hook as it can be called conditionally. This moves it to the React APIs section in the reference and refers to it as a React API instead of React Hook.

### DIFF
--- a/src/content/reference/react/hooks.md
+++ b/src/content/reference/react/hooks.md
@@ -106,24 +106,6 @@ To prioritize rendering, use one of these Hooks:
 
 ---
 
-## Resource Hooks {/*resource-hooks*/}
-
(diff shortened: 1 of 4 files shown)


src/content/reference/react/hooks.md
```diff
@@ -106,24 +106,6 @@ To prioritize rendering, use one of these Hooks:
 
 ---
 
-## Resource Hooks {/*resource-hooks*/}
```
rickhanlonii, line 109: Can we move this to apis.md? It can also explain that it can only be called in render, like a hook, but can be called conditionally, unlike a hook.

rickhanlonii (approved):
LGTM, just port the resource info

PR #6775, opened 2024-04-24
Update test warning content

jackpope:
- Make it more clear that shallow rendering is not supported
- Provide recommendations for incrementally migrating tests before major version upgrade

### DIFF
--- a/src/content/warnings/react-test-renderer.md
+++ b/src/content/warnings/react-test-renderer.md
@@ -4,11 +4,20 @@ title: react-test-renderer Deprecation Warnings
 
 ## ReactTestRenderer.create() warning {/*reacttestrenderercreate-warning*/}
 
-react-test-renderer is deprecated. A warning will fire whenever calling ReactTestRenderer.create() or ReactShallowRender.render(). The react-test-renderer package will remain available on NPM but will not be maintained and may break with new React features or changes to React's internals.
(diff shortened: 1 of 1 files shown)


src/content/warnings/react-test-renderer.md
```diff
@@ -4,11 +4,20 @@ title: react-test-renderer Deprecation Warnings
...
 
-react-test-renderer is deprecated. A warning will fire whenever calling ReactTestRenderer.create() or ReactShallowRender.render(). The react-test-renderer package will remain available on NPM but will not be maintained and may break with new React features or changes to React's internals.
+`react-test-renderer` is deprecated. An error will log whenever calling ReactTestRenderer.create(). The package will remain available on NPM but will not be maintained and will break with new React features or changes to React's internals.
 
+<Note>
 The React Team recommends migrating your tests to [@testing-library/react](https://testing-library.com/docs/react-testing-library/intro/) or [@testing-library/react-native](https://callstack.github.io/react-native-testing-library/docs/getting-started) for a modern and well supported testing experience.
 
+If your application has extensive `react-test-renderer` usage, consider pinning the package to an alias such as `react-test-renderer-DO_NOT_USE` and incrementally migrating to `testing-library`. When your tests have been migrated, you can upgrade to React 19 with more confidence.
```
rickhanlonii, line 12: The pinning strategy is to not block upgrading to 19, so you don't need to upgrade them all at first. The caveat is that you won't be able to use 19 features like actions in your tests.

jackpope: Updated

PR #6776, opened 2024-04-24
Rename useFormState to useActionState

kassens:
We renamed this and moved it to the `react` package because it's not tied to DOM forms.

### DIFF
--- a/src/content/reference/react/hooks.md
+++ b/src/content/reference/react/hooks.md
@@ -113,6 +113,7 @@ These Hooks are mostly useful to library authors and aren't commonly used in the
 - [`useDebugValue`](/reference/react/useDebugValue) lets you customize the label React DevTools displays for your custom Hook.
 - [`useId`](/reference/react/useId) lets a component associate a unique ID with itself. Typically used with accessibility APIs.
 - [`useSyncExternalStore`](/reference/react/useSyncExternalStore) lets a component subscribe to an external store.
+* [`useActionState`](/reference/react/useActionState) allows you to manage state of actions.
 
 ---
 
--- a/src/content/reference/react/useActionState.md
+++ b/src/content/reference/react/useActionState.md
@@ -9,6 +9,12 @@ The `useActionState` Hook is currently only available in React's Canary and expe
 
 </Canary>
 
+<Note>
+
+In earlier React Canary versions, this API was part of `react-dom` and called `useFormState`.
+
+</Note>
+
 <Intro>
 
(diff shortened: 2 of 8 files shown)


src/content/reference/react/hooks.md
```diff
@@ -113,6 +113,7 @@ These Hooks are mostly useful to library authors and aren't commonly used in the
 - [`useDebugValue`](/reference/react/useDebugValue) lets you customize the label React DevTools displays for your custom Hook.
 - [`useId`](/reference/react/useId) lets a component associate a unique ID with itself. Typically used with accessibility APIs.
 - [`useSyncExternalStore`](/reference/react/useSyncExternalStore) lets a component subscribe to an external store.
+* [`useActionState`](/reference/react/useActionState) allows you to manage state of actions.
```
kassens, line 116: Should there be a section `Action Hooks` maybe? Not sure where this should go.

rickhanlonii (commented):
Nice!! One thing is that we need to keep `useFormState` for a bit because it's only deprecated in 19 and there are a ton of users using it (for example, everyone on Next.js App Router) and there are links around the net for it. We could do that with a shell that has a warning and a link to the new page, or add a redirect in `vercel.json`

src/content/reference/react/useActionState.md
```diff
@@ -283,7 +281,7 @@ form button {
 
 ### My action can no longer read the submitted form data {/*my-action-can-no-longer-read-the-submitted-form-data*/}
 
-When you wrap an action with `useFormState`, it gets an extra argument *as its first argument*. The submitted form data is therefore its *second* argument instead of its first as it would usually be. The new first argument that gets added is the current state of the form.
+When you wrap an action with `useActionState`, it gets an extra argument *as its first argument*. The submitted form data is therefore its *second* argument instead of its first as it would usually be. The new first argument that gets added is the current state of the form.
```
kassens, line 284: A lot is referring to forms. Probably needs another pass to abstract this more from DOM forms specifically and explain how to use actions w/o forms.

src/content/reference/react/hooks.md
rickhanlonii: Yeah I'd create a section with useTransition, useActionState, and useDeferredValue and call it like "UI hooks" or something

jackpope (approved):

koba04:
@kassens Do you have a plan to add the `isPending` value returned from `useActionState` into the docs?

rickhanlonii:
@koba04 yeah I'm updating the docs, got a little delayed with React Conf

Issue #6123, opened 2023-06-28
React 18 strict mode docs are inconsistent about how unmount and remount actually works (and whether it's simulated)


AleksandrHovhannisyan:
**Problem**

The [React v18.0 release article](https://react.dev/blog/2022/03/29/react-v18) states the following about React 18's new strict mode behavior. Emphasis is my own:

> To help surface these issues, React 18 introduces a new development-only check to Strict Mode. This new check will **automatically unmount and remount every component**, whenever a component mounts for the first time, restoring the previous state on the **second mount**.

However, in the following code demo, I'm able to use a ref to keep track of the number of times that `useEffect` runs, which I would not be able to do if the component were truly unmounting since refs are destroyed: https://codesandbox.io/s/run-mount-logic-only-once-zpywhc?file=/src/App.js.

```jsx
import { useEffect, useRef } from "react";

export default function App() {
  const hasMounted = useRef(false);

  useEffect(() => {
    if (hasMounted.current === false) {
      hasMounted.current = true;
      console.count("mounted");
    }
  }, []);

  return <></>;
}
```

If the component truly unmounted and remounted and then restored state on the second mount, then the ref would not persist because 1) refs are not state (and hence not restored as described in the above quote), and 2) refs are normally destroyed on unmount. In that case, we would see the following logs:

```
mounted: 1
mounted: 2
```

However, what we actually see is:

```
mounted: 1
```

Which seems to suggest that React is not _actually_ unmounting and remounting every component.

But later on, the same article clarifies that React _simulates_ unmounting and remounting:

> With Strict Mode in React 18, React will simulate unmounting and remounting the component in development mode:

Which makes a lot more sense: Effects are _invoked_ twice to _simulate_ an unmount and remount, but components are not _truly_ unmounted. The React docs on [Fixing bugs found by re-running Effects in development]() further clarify this:

> When Strict Mode is on, React will also run one extra setup+cleanup cycle in development for every Effect. This may feel surprising, but it helps reveal subtle bugs that are hard to catch manually.

**Suggestion**: 

There are two possibilities as far as I can tell:

1. React _does_ actually unmount and remount components, and it restores not only "state" but also anything within the scope of the component (e.g., refs).
2. (More likely) React _simulates_ remounting by simply _invoking_ every effect twice.

[...]

PR #6777, opened 2024-04-24
Add information about ref handling in strict mode

noahlemen:

### DIFF
--- a/src/content/reference/react/StrictMode.md
+++ b/src/content/reference/react/StrictMode.md
@@ -44,7 +44,7 @@ Strict Mode enables the following development-only behaviors:
 
 - Your components will [re-render an extra time](#fixing-bugs-found-by-double-rendering-in-development) to find bugs caused by impure rendering.
 - Your components will [re-run Effects an extra time](#fixing-bugs-found-by-re-running-effects-in-development) to find bugs caused by missing Effect cleanup.
-- <CanaryBadge title="This feature is only available in the Canary channel" /> Your components will [detach and re-attach refs to components](#fixing-bugs-found-by-cleaning-up-and-re-attaching-dom-refs-in-development) to find bugs caused by missing ref cleanup.
+- <NextMajorBadge title="This feature is available in React 19 beta and the React canary channel" /> Your components will [detach and re-attach refs to components](#fixing-bugs-found-by-cleaning-up-and-re-attaching-dom-refs-in-development) to find bugs caused by missing ref cleanup.
 - Your components will [be checked for usage of deprecated APIs.](#fixing-deprecation-warnings-enabled-by-strict-mode)
 
 #### Props {/*props*/}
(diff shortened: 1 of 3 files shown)


src/content/reference/react/StrictMode.md
```diff
@@ -825,7 +827,19 @@ Without Strict Mode, it was easy to miss that your Effect needed cleanup. By run
...
 
 ---
+### Fixing bugs found by detaching and re-attaching DOM refs in development {/*fixing-bugs-found-by-cleaning-up-and-re-attaching-dom-refs-in-development*/}
 
+<Canary> 
+In canary and experimental channels, React will run an extra setup+cleanup cycle in development for DOM refs, much like it does for Effects. 
+
+For [`ref` callbacks](/reference/react-dom/components/common#ref-callback), React will call the callback function with the DOM node as its argument. It will then call the callback's [cleanup function](reference/react-dom/components/common#returns) before calling the `ref` callback function again with the DOM node as its argument.
+
+Similarly, React will detach DOM refs that were created via `useRef` by setting `ref.current` to `null` before setting it back to the DOM node.
```
jackpope, line 837: useRef is probably more common than callback refs. Maybe just flip it to put this statement before the line above?

src/content/reference/react/StrictMode.md
```diff
@@ -825,7 +827,19 @@ Without Strict Mode, it was easy to miss that your Effect needed cleanup. By run
 [Read more about implementing Effect cleanup.](/learn/synchronizing-with-effects#how-to-handle-the-effect-firing-twice-in-development)
 
 ---
+### Fixing bugs found by detaching and re-attaching DOM refs in development {/*fixing-bugs-found-by-cleaning-up-and-re-attaching-dom-refs-in-development*/}
```
jackpope, line 830: I think the `<Canary>` component should start above this line since this category is only applicable in canary

jackpope (approved):

src/content/reference/react/StrictMode.md
noahlemen: i was debating doing this, but that's a good justification - thanks

src/content/reference/react/StrictMode.md
noahlemen: it seems like the header doesn't show up properly in the outline if i do that. maybe can fix that though

jackpope: Not a big deal if it messes up the formatting

eps1lon: We should just say "refs" instead of "DOM refs". Double invoking works for all of them now: https://ccp74m.csb.app/

noahlemen: we don't currently have terminology to differentiate clearly between `ref`s as a general store for data (eg: a `ref` object created via `useRef` which points to some arbitrary value) versus `ref`s as passed to an element. we currently seem to conceptualize this in the docs as being centered around interation with DOM elements ([example](https://react.dev/learn/referencing-values-with-refs#refs-and-the-dom) even though this doesn't clearly encompass the kinds of refs like your example shows. open to suggestions though – perhaps i'm overthinking this and just saying "refs" is fine.

eps1lon: what about "detaching and re-attachin refs to components" to make it clear we're not talking about refs as datastores like `useRef`?

noahlemen: i think that makes sense. i wasn't sure if "components" would be inclusive of DOM elements, but it looks like elsewhere in the reference we refer to these as ["common components"](https://react.dev/reference/react-dom/components/common), so i think that works.

AleksandrHovhannisyan:
See also this issue that I opened a while back: https://github.com/reactjs/react.dev/issues/6123. In it, I proposed that we update the React docs to discourage developers from relying on `useRef` to run useEffect logic once on mount in strict mode.

Given that React 19 is now fixing refs so they run twice in strict mode (https://github.com/facebook/react/issues/24670), I think it's worth flagging the "useRef to run useEffect once on mount" hack in the docs as an anti-pattern. Our code base naively adopted this widespread pattern back when we were upgrading to React 18, and now it seems like we've only delayed the inevitable.

For example, search up "React 18 strict mode run useEffect once" or something along those lines, and nearly every answer that comes up suggests using `useRef` as a workaround. For example:

https://stackoverflow.com/questions/72238175/why-useeffect-running-twice-and-how-to-handle-it-well-in-react

rickhanlonii:
@AleksandrHovhannisyan note that this only impacts element refs like DOM nodes and refs with `useImperitiveEffect`, refs created with useRef will not be destroyed, so this seems like a separate thing from your concern.

I added docs for the pattern you suggested here: https://github.com/reactjs/react.dev/pull/6846

rickhanlonii (approved):

PR #6796, opened 2024-04-26
Updated react-19.md

christianbmartinez:
Updated "feature" to "features" as this sentence was presumably intended to be plural by the original author.

Also added a comma in the beginning of paragraph as it felt like a run on sentence. 

Starting the next new sentence with "These" instead of "So". 

That's all i've got for now :) 

It would be my first PR, I hope to get into more technical PR's down the road. 

<3


eps1lon (approved):
Thanks

PR #6798, opened 2024-04-26
Fix type issues with `useActionState` docs

eps1lon:
Fixes various issues with how we document `useActionState`. Some of it related to `initialState` being required (still being discussed) but also other issues like putting the permalink in the second parameter or forgetting initial state when it was actually required (e.g. unconditionally accessing `state.error`)


rickhanlonii (approved):
ope

rickhanlonii:
Conflicts

PR #6811, opened 2024-04-28
Convert "Canary" callouts to "React 19 beta"

rickhanlonii:
## Overview

**Update: The base branch is `v19` since these are React 19 specific**

_Note: this is included in one PR to minimize the number of conflicts with the translation repos._

Now that the React 19 beta is out, the callouts on the page can be converted to specify "React 19 beta" for the features that are in React 19. I also switched the React Server Components callouts to it's own callout type, since the important notice there is not the specific release but the environment they run in.

Included in this PR:

- Adds the `<NextMajor>` and `<RSC> callout components, similar to the `<Canary>` callout.
- Adds the `<NextMajorBadge>` and `<RSCBadge>` components, similar to `<CanaryBadge>`
- Switches the sidebar metadata from just `canary` with an enum `canary | major`
- Adds `React 19` badges to sidebar items
- Updates text for callouts

Note on the text: the text says "available in React 19 beta and the React canary channel". This is so users who are using the canary with these features today will not be confused and think they're not available in their release. We can drop the "React canary channel" part once we ship the pinned version for react 18, but it's probably better to keep it until the stable release so we don't churn external links.

### React 19 callout and sidebar

Screenshot 2024-04-28 at 2 50 28 PM

### React Server Component callout 

Screenshot 2024-04-28 at 2 50 52 PM

### Inline React 19 badges

Screenshot 2024-04-28 at 2 55 04 PM


src/content/learn/manipulating-the-dom-with-refs.md
```diff
@@ -341,9 +341,9 @@ In this example, `itemsRef` doesn't hold a single DOM node. Instead, it holds a
 
 This lets you read individual DOM nodes from the Map later.
 
-<Canary>
+<NextMajor>
 
-This example shows another approach for managing the Map with a `ref` callback cleanup function.
+Starting in React 19 callback refs can return a cleanup function. When the cleanup function is provided, React will not call the `ref` callback with `null` and call the cleanup function instead:
```
harish-sethuraman, line 346: Should we mention beta here as well? 
```suggestion
Starting from React 19 beta, callback refs can return a cleanup function. When the cleanup function is provided, React will not call the `ref` callback with `null` and call the cleanup function instead:
```

jackpope (approved):
Nice improvement!

rickhanlonii:
Changes the base to `v19`

src/content/learn/manipulating-the-dom-with-refs.md
rickhanlonii: Added the comma, but I think "starting in" works better here, and since this is already versioned to the beta I don't think we need to mention the beta inline.

PR #6817, opened 2024-04-30
Add logo option

rickhanlonii:
## Overview

Adds a query param to switch the React logo to https://github.com/SAWARATSUKI/ServiceLogos:
- on: `?uwu=true`
- off: `?uwu=false`

Setting will persist until you turn it off.

## Screen

Screenshot 2024-04-30 at 1 00 19 AM

## Credits

Credit to @SAWARATSUKI in https://github.com/SAWARATSUKI/ServiceLogos

This PR also adds a console.log and footer credit for the amazing artwork.


src/pages/_document.tsx
```diff
@@ -19,6 +19,41 @@ const MyDocument = () => {
         }}
       />
       <body className="font-text font-medium antialiased text-lg bg-wash dark:bg-wash-dark text-secondary dark:text-secondary-dark leading-base">
+        <script
+          dangerouslySetInnerHTML={{
+            __html: `
```
Daniel15, line 24: Creating a `__html` object inline is explicitly discouraged in the React docs 😅 we should follow the best practices here. Can't the script just go in a regular .js file?

rickhanlonii: There is no other way to use `dangerouslySetInnerHTML`, which we have to use since the head rendered by React and browsers ignore script text set by innerText for security reasons. 

The cautions about it being unsafe are not applicable, because the string is static. We can't use a `.js` file because it would require a round trip to fetch the script, so it would run later and flash the wrong theme or logo (see the existing logic for setting the theme in the script tag below this).

When we convert to the App Router, this is supported out of the box with React 19 with the new Resources support https://react.dev/reference/react-dom/components#resource-and-metadata-components

PR #6830, opened 2024-05-03
Fix React 19 upgrade guide typo

Ryczko:
While reading the article about the React 19 Beta update, I came across a small typo which I fixed.


eps1lon (approved):
Thank you!

PR #6836, opened 2024-05-05
Add a redirect entry for feed.xml => rss.xml

### LINKED ISSUE #6835: [Bug]: missing redirect for feed.xml
### Summary

I've been consuming the react blog via rss feed for a long time. This is the feed URL I've subscribed to in my feed reader.

`https://facebook.github.io/react/feed.xml`

It's now redirecting to `https://react.dev/feed.xml` which returns a 404. I can see that the latest blog feeds are at `https://react.dev/rss.xml` but someone probably missed or forgot to setup redirect from `/feed.xml` to `/rss.xml`.

I can fix my feed URL but for backwards compatibility for all the subscribers who are still subscribed to `/feed.xml`, I would recommend setting up a redirect to `/rss.xml`.

### Page

https://react.dev/feed.xml

### Details

_No response_

flexdinesh:
fixes #6835

### DIFF
--- a/vercel.json
+++ b/vercel.json
@@ -173,6 +173,11 @@
       "source": "/reference/react/use-server",
       "destination": "/reference/rsc/use-server",
       "permanent": true
+    },
+    {
+      "source": "/feed.xml",
+      "destination": "/rss.xml",
+      "permanent": true
     }
   ],
   "headers": [


rickhanlonii (approved):
Thanks!

PR #6837, opened 2024-05-05
Update transition docs for React 19

rickhanlonii:
### [Preview](https://19-react-dev-git-fork-rickhanlonii-19-async-e9fd9f-fbopensource.vercel.app/reference/react/useTransition#creating-an-action-with-async-transitions)

This PR updates the `useTransition` docs to include actions. 
- Reworks existing content that says it's only sync
- Updated all the examples with the proper action naming and to use `use`
- Adds "Perform non-blocking updates with Actions" to explain actions
- Adds "Exposing action prop from components" to explain how to expose an action prop
- Adds troubleshooting for set state after await
- Adds troubleshooting for out-of-order requests
- Removes "Marking a state update as a non-blocking Transition ", since there's already a section below 

Also makes minor edits to `startTransition` to use actions.

### DIFF
--- a/src/content/reference/react/useTransition.md
+++ b/src/content/reference/react/useTransition.md
@@ -107,7 +107,7 @@ function SubmitButton({ submitAction }) {
 
 * You can wrap an update into a Transition only if you have access to the `set` function of that state. If you want to start a Transition in response to some prop or a custom Hook value, try [`useDeferredValue`](/reference/react/useDeferredValue) instead.
 
-* The function you pass to the of `startTransition` is called immediately, marking all state updates that happen while it executes as Transitions. If you try to perform state updates in a setTimeout, they won't be marked as Transitions.
+* The function you pass to the of `startTransition` is called immediately, marking all state updates that happen while it executes as Transitions. If you try to perform state updates in a `setTimeout`, they won't be marked as Transitions.
 
 * You must wrap any state updates after any async requests in another `startTransition` to mark them as Transitions. This is a known limitation that we will fix in the future (see [Troubleshooting](#react-doesnt-treat-my-state-update-after-await-as-a-transition)).
 
--- a/src/content/reference/react/useTransition.md
+++ b/src/content/reference/react/useTransition.md
@@ -163,7 +163,7 @@ function CheckoutForm() {
 
 The function passed to `startTransition` is called the "Action". In Actions, you can perform side effects and update state within a Transition, without blocking the UI. A Transition can include multiple Actions, and while the Transition is in progress, your UI stays responsive in the middle of a re-render.
 
-For example, if the user clicks a tab but then change their mind and click another tab, they can do that without waiting for the first re-render to finish. But if the Action is performing a side effect, such as submitting a form, React will wait for the side effect to finish before changing the tab.
+For example, if the user clicks a tab but then changes their mind and clicks another tab, they can do that without waiting for the first re-render to finish. But if the Action is performing a side effect, such as submitting a form, React will wait for the side effect to finish before changing the tab.
 
 To give the user feedback about in-progress Transitions, to `isPending` state switches to `true` at the first call to `startTransition`, and stays `true` until all Actions complete and the final state is shown to the user. Transitions wait for side effects in Actions to complete in order to [prevent unwanted loading indicators](#preventing-unwanted-loading-indicators), and you can provide immediate feedback while the Transition is in progress with `useOptimistic`.
 
--- a/src/content/reference/react/useTransition.md
+++ b/src/content/reference/react/useTransition.md
@@ -4,7 +4,7 @@ title: useTransition
 
 <Intro>
 
-`useTransition` is a React Hook that lets you update the state without blocking the UI.
+`useTransition` is a React Hook that lets you update without blocking the UI.
 
 ```js
 const [isPending, startTransition] = useTransition()
@@ -42,20 +42,20 @@ function TabContainer() {
 `useTransition` returns an array with exactly two items:
 
 1. The `isPending` flag that tells you whether there is a pending Transition.
-2. The [`startTransition` function](#starttransition) that lets you mark a state update as a Transition.
+2. The [`startTransition` function](#starttransition) that lets you mark updates as a Transition.
 
 ---
 
-### `startTransition(fn)` {/*starttransition*/}
+### `startTransition(action)` {/*starttransition*/}
 
-The `startTransition` function returned by `useTransition` lets you mark a state update as a Transition.
+The `startTransition` function returned by `useTransition` lets you mark a updates as a Transition.
 
 ```js {6,8}
 function TabContainer() {
   const [isPending, startTransition] = useTransition();
   const [tab, setTab] = useState('about');
 
-  function selectTab(nextTab) {
+  function selectTabAction(nextTab) {
     startTransition(() => {
       setTab(nextTab);
     });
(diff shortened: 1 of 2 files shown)


src/content/reference/react/useTransition.md
```diff
@@ -64,9 +64,30 @@ function TabContainer() {
...
+* `scope`: A function that updates some state by calling one or more [`set` functions](/reference/react/useState#setstate). React immediately calls `scope` with no parameters and marks all state updates scheduled synchronously during the `scope` function call as Transitions. Any async calls awaited in the `scope` will be included in the transition, but currently require wrapping any `set` functions after the request in an additional `startTransition` (see [Troubleshooting](#react-doesnt-treat-my-state-update-after-await-as-a-transition)). State updates marked as Transitions will be [non-blocking](#marking-a-state-update-as-a-non-blocking-transition) and [will not display unwanted loading indicators.](#preventing-unwanted-loading-indicators).
```
harish-sethuraman, line 90:
```suggestion
* `scope`: A function that updates some State by calling one or more [`set` functions](/reference/react/useState#setstate). React immediately calls `scope` with no parameters and marks all state updates scheduled synchronously during the `scope` function call as Transitions. Any async calls awaited in the `scope` will be included in the transition, but currently require wrapping any `set` functions after the request in an additional `startTransition` (see [Troubleshooting](#react-doesnt-treat-my-state-update-after-await-as-a-transition)). State updates marked as Transitions will be [non-blocking](#marking-a-state-update-as-a-non-blocking-transition) and [will not display unwanted loading indicators.](#preventing-unwanted-loading-indicators).
```

harish-sethuraman: I see a lot of state references from react to be in smallcase? Is this intentional?

src/content/reference/react/useTransition.md
```diff
@@ -409,6 +430,447 @@ b { display: inline-block; margin-right: 10px; }
...
+function CheckoutForm() {
+  const [isPending, startTransition] = useTransition();
+  const [quantity, setQuantity] = useState(1);
+
+  function updateQuantityAction(newQuantity) {
+    startTransition(async () => {
+      const savedQuantity = await updateQuantity(newQuantity);
+      startTransition(() => {
+        setQuantity(savedQuantity);
+      });
```
slorber, line 457: It's probably useful to add a small note in this sandbox that explains why we need the extra `startTransition` here. I was looking for this info but it only appears later in the troubleshooting section.

I also tried to remove the extra `startTransition()` to see if I got a bug, but in this case I didn't really see any behavior change. It's probably only due to that specific example since `setQuantity` happens last

tom-sherman:
To be honest I really don't see async transitions (with useTransition) being practically adoptable without the race condition handling built in. Without this, they are simply a target for libraries as the queue handling is very tricky.

There are so many footguns with using them on their own that I think there needs to be some massive disclaimer, and for them to be called out as a tool for library and framework authors.

rickhanlonii:
@tom-sherman it's going to be more common for app developers to use `useActionState` for form actions, instead of using raw transitions directly. Using async transitions directly is targeted to library developers who intend to write their own queuing logic or component library authors to expose an `action` prop.

rickhanlonii:
It's similar to how doing event handlers to submit data on your own is tricky (for example, this case isn't handled in a raw event handler either), and libraries make that easier using raw events and exposing props to limit the complexity to app developers.

rickhanlonii:
Converting back to WIP, I think there is a better way to explain this.

src/content/reference/react/useTransition.md
```diff
@@ -1672,13 +2138,17 @@ startTransition(async () => {
...
-await someAsyncFunction();
-startTransition(() => {
-  // ✅ Setting state *during* startTransition call
-  setPage('/about');
+startTransition(async () => {
+  await someAsyncFunction();
+  // ✅ Using startTransition *after* await
+  startTransition(() => {
+    setPage('/about');
+  });
```
slorber, line 2146: Something that isn't immediately clear to me:

Do we need this nested `startTransition` for:
- regular form actions too?
- calling `setOptimisticValue`?

I assume we do, but not 100% sure since I don't see any example in the docs.

The `<form>` actions examples are likely to use `useActionState` instead of doing `startTransition(() => setState())`

The `useOptimistic` doc only calls the setter as the first step, probably the most common case, but maybe it can make sense in some rare cases to call it after `await`? I don't know 🤷‍♂️

```tsx
  async function formAction(formData) {
    addOptimisticMessage(formData.get("message"));
    formRef.current.reset();
    await sendMessage(formData);
  }
```

rickhanlonii: Form actions will call that out if needed on the form page, the examples there will not all use useActionState, but even if they do you would need to add the startTransition after the await there too. For useOptimistic, you could have multiple awaits, and want to call it between the two.

rickhanlonii:
Updated this page based on feedback, some notable changes:
- An Action is not just an async transition, you can have sync actions too (like mutations to local data that doesn't need a network request). The new framing is that the function passed to `startTransition` (sync or async) is the "action" and by convention if a function is called inside startTransition then it's called an "Action".
- Added "Perform non-blocking updates with Actions" to explain actions
- Added "Exposing action prop from components" to explain how to expose an `action` prop
- Removed "Marking a state update as a non-blocking Transition ", since there's already a section below for "Building a Suspense-enabled router" for navigation, but I think we need a new usage example for this.
- Updated all the examples with the proper `action` naming and to use `use`

src/content/reference/react/useTransition.md
rickhanlonii: Yeah, i'll do a pass before merging. It's annoying to have every other word capitalized though

noahlemen (commented):
added some minor suggestions

src/content/reference/react/useTransition.md
```diff
@@ -78,7 +107,9 @@ function TabContainer() {
 
 * You can wrap an update into a Transition only if you have access to the `set` function of that state. If you want to start a Transition in response to some prop or a custom Hook value, try [`useDeferredValue`](/reference/react/useDeferredValue) instead.
 
-* The function you pass to `startTransition` must be synchronous. React immediately executes this function, marking all state updates that happen while it executes as Transitions. If you try to perform more state updates later (for example, in a timeout), they won't be marked as Transitions.
+* The function you pass to the of `startTransition` is called immediately, marking all state updates that happen while it executes as Transitions. If you try to perform state updates in a `setTimeout`, they won't be marked as Transitions.
+
+* You must wrap any state updates after any async requests in another `startTransition` to mark them as Transitions. This is a known limitation that we will fix in the future (see [Troubleshooting](#react-doesnt-treat-my-state-update-after-await-as-a-transition)).
```
jherr, line 112: This one is complex enough where an example would be ideal.

src/content/reference/react/useTransition.md
```diff
@@ -78,7 +107,9 @@ function TabContainer() {
 
 * You can wrap an update into a Transition only if you have access to the `set` function of that state. If you want to start a Transition in response to some prop or a custom Hook value, try [`useDeferredValue`](/reference/react/useDeferredValue) instead.
 
-* The function you pass to `startTransition` must be synchronous. React immediately executes this function, marking all state updates that happen while it executes as Transitions. If you try to perform more state updates later (for example, in a timeout), they won't be marked as Transitions.
+* The function you pass to the of `startTransition` is called immediately, marking all state updates that happen while it executes as Transitions. If you try to perform state updates in a `setTimeout`, they won't be marked as Transitions.
```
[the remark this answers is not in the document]
rickhanlonii:
```suggestion
* The function you pass to the of `startTransition` is called immediately, marking all state updates that happen while it executes as Transitions. If you try to perform state updates in a `setTimeout`, for example, they won't be marked as Transitions.
```

PR #6839, opened 2024-05-06
Fix out of bound error after list length is reached.

mahmedtan:
State breaks after list length is reached.


harish-sethuraman:
This is expected. We actually have a challenge at the end of the page for this. (first challenge at the end of page)

PR #6844, opened 2024-05-07
[19] Remove <NextMajor> callouts

rickhanlonii:
Base branch: `v19`

## Overview

In the 19 version of the docs, these APIs have landed, so we don't need the callouts.


src/content/reference/react-dom/components/common.md
```diff
@@ -246,45 +246,42 @@ These events fire for resources like [`<audio>`](https://developer.mozilla.org/e
...
 
 React will also call your `ref` callback whenever you pass a *different* `ref` callback. In the above example, `(node) => { ... }` is a different function on every render. When your component re-renders, the *previous* function will be called with `null` as the argument, and the *next* function will be called with the DOM node.
 
 #### Parameters {/*ref-callback-parameters*/}
 
-* `node`: A DOM node or `null`. React will pass you the DOM node when the ref gets attached, and `null` when the `ref` gets detached. Unless you pass the same function reference for the `ref` callback on every render, the callback will get temporarily detached and re-attached during every re-render of the component.
+* `node`: A DOM node. React will pass you the DOM node when the ref gets attached. Unless you pass the same function reference for the `ref` callback on every render, the callback will get temporarily cleanup and re-create during every re-render of the component.
```
kassens, line 266: Should this be documented as nullable here still? In case you don't need a cleanup function and don't return a cleanup you'll suddenly get a null passed, maybe not expecting that.

Is not returning a cleanup function deprecated then? Should that deprecated behavior be called out more strongly below so in the "returns" section it's not said to be optional?

rickhanlonii: There is a <Note> section on the line below this that explains the behavior for null.

PR #6846, opened 2024-05-07
Add docs for useRef no no to fix double effects

rickhanlonii:
Fixes the request in https://github.com/reactjs/react.dev/issues/6123 to document this pattern.


penx:
@rickhanlonii I wonder if

> Don’t use refs to prevent Effects from firing

Is too broad?

i.e. does this pitfall specifically refer to the dev mode double render?


It's not clear to me from the docs whether it would also apply to the following:

```ts
useEffect(() => {
    // only track an event on initial component render, even if eventData changes later
    if (hasTracked.current) return;
    hasTracked.current = true;
    trackEvent(eventData);
  }, [eventData]);
```

PR #6848, opened 2024-05-09
checks list has duplicated, delete one.

zhangenming:


rickhanlonii:
This is intentional, since it's a reference each section could be linked to independently and the context is important in both locations.

Issue #6854, opened 2024-05-10
[Bug]: New node.js version is not compatible with react.dev


Rekl0w:
### Summary

New node.js version ^22.0.0 needs to be added to `package.json`.

### Page

https://react.dev

### Details

![ss1](https://github.com/reactjs/react.dev/assets/91488737/0d52f12c-8eaa-4c45-9831-4757105ba1c1)

PR #6855, opened 2024-05-10
Fix Node.js version update at `package.json`

Rekl0w:
Fix Node.js version update at `package.json` #6854


package.json
```diff
@@ -98,7 +98,7 @@
     "webpack-bundle-analyzer": "^4.5.0"
   },
   "engines": {
-    "node": "^16.8.0 || ^18.0.0 || ^19.0.0 || ^20.0.0 || ^21.0.0"
+    "node": "^16.8.0 || ^18.0.0 || ^19.0.0 || ^20.0.0 || ^21.0.0 || ^22.0.0"
```
harish-sethuraman, line 101: at this point we should actually not tag any versions but rather have node to be `>=16.8.0` is what I feel? @rickhanlonii 

similar to https://github.com/vercel/next.js/blob/cc10bf5d1fa4fd2bead34661070c936d360516b3/packages/next/package.json#L340

Rekl0w: I can update it like this if you want.

harish-sethuraman: We might also want to verify if everything is working fine before we add these node versions. Like the start scripts. Seems like 17 alone is missing in between. If its verified we are GTG.

rickhanlonii: Yes please, I loath node version checks - let's verify the oldest supported version and let's just be `>=` to that version, likely the same version Next requires. In the future, if a newer node version is broken, we can try to change the APIs so we support the newer version.

rickhanlonii (approved):
sick thank you

PR #6856, opened 2024-05-10
Update synchronizing-with-effects.md

segmentationfaulter:
Currently documentation says:

>By default, your Effect will run after every render.

Which is in contrast with another [section of the same page](https://react.dev/learn/synchronizing-with-effects#what-are-effects-and-how-are-they-different-from-events) where it says:

>Effects run at the end of a commit after the screen updates

That's why I am proposing this change. Thank you

### DIFF
--- a/src/content/learn/synchronizing-with-effects.md
+++ b/src/content/learn/synchronizing-with-effects.md
@@ -45,7 +45,7 @@ Here and later in this text, capitalized "Effect" refers to the React-specific d
 
 To write an Effect, follow these three steps:
 
-1. **Declare an Effect.** By default, your Effect will run after every render.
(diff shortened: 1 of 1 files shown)


src/content/learn/synchronizing-with-effects.md
```diff
@@ -45,7 +45,7 @@ Here and later in this text, capitalized "Effect" refers to the React-specific d
 
 To write an Effect, follow these three steps:
 
-1. **Declare an Effect.** By default, your Effect will run after every render.
+1. **Declare an Effect.** By default, your Effect will run after [commit phase](/learn/render-and-commit) i.e. when changes are comitted to the dom.
```
rickhanlonii, line 48:
```suggestion
1. **Declare an Effect.** By default, your Effect will run after every [commit](/learn/render-and-commit).
```

PR #6859, opened 2024-05-11
use a better way to describe it.

zhangenming:
my english not so good
He made me misunderstand
Hopefully that's a simpler description


rickhanlonii:
Hm, sorry this is confusing but it's correct grammatically. Changing to "except" changes the meaning subtly and I don't think it's correct.

PR #6866, opened 2024-05-14
add frontend nation conference

danielkellyio:
Front end nation is covering a full track of React talks including speakers such as:

* Kent C Dodds
* Sara Vieira
* Tejas Kumar


rickhanlonii (approved):
sweet!

PR #6894, opened 2024-05-21
fix: remove unused id from render-list example component

mostafaroshdy1:
- Removed unused  id from challenge number 3 here https://react.dev/learn/rendering-lists#challenges


rickhanlonii:
The Recipe is spread with an id so it needs to be in the props (if it was typed, it would fail otherwise). The other solution would be to only pass the other props without the ID from the parent, but it's not a big deal IMO.

PR #6901, opened 2024-05-22
Add act documentation

jackpope:
Content from legacy docs https://legacy.reactjs.org/docs/test-utils.html#act

### DIFF
--- a/src/content/reference/react/act.md
+++ b/src/content/reference/react/act.md
@@ -4,24 +4,45 @@ title: act
 
 <Intro>
 
-To prepare a component for assertions, wrap the code rendering it and performing updates inside an act() call. This makes your test run closer to how React works in the browser.
+`act` lets you wrap rendering and updates in tests to ensure changes are applied to the DOM before making assertions.
+
+```js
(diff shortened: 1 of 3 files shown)


rickhanlonii:
Can you update this to match our other Reference pages?

The top section should be a one-line short description, followed by a psudoish code example, and a TOC:

Screenshot 2024-05-22 at 2 54 07 PM

Then a Reference with a short example, parameters, returns, and caveats

 
Screenshot 2024-05-22 at 2 55 29 PM

Then a usage section with headings that explain the use case and are visible in the right sidebar:

Screenshot 2024-05-22 at 2 56 10 PM

Then a troubleshooting section:

Screenshot 2024-05-22 at 2 56 55 PM

We can defer the troubleshooting section, but it would include things like not using `await` or not setting the `IS_ACT_ENV` global.

rickhanlonii:
And can we change the base branch to `main`? Since this was added to 18.3, it should be added to those docs. I'll merge it into the v19 branch

rickhanlonii (approved):

PR #6916, opened 2024-05-28
docs: update links @testing-library/react-native

germanolira:
# Issue
The link in the React Native library documentation is broken.
https://react.dev/warnings/react-test-renderer

# The old way
[@testing-library/react-native](https://callstack.github.io/react-native-testing-library/docs/getting-started)

# The new way
[@testing-library/react-native](https://callstack.github.io/react-native-testing-library/docs/start/intro)

# Fix
Text updated with new link


eps1lon (approved):
Thank you. Worth filing an issue with React Native Testing Library so that they create a redirect. We're probably not the only ones linking to this page that was moved.

PR #6929, opened 2024-06-02
code improvement

JohnKimaiyo:


rickhanlonii:
As you can see by the failed tests, this breaks our formatting rules. We don't generally accept PRs that only change formatting.

PR #6939, opened 2024-06-09
docs(api-act): api act parts word error

yaolifeng0629:
- Question Description:
	-  	There is a misspelling in [api-act](https://react.dev/reference/react/act#rendering-components-in-tests) parts of the document. If it is intentional, please ignore this PR.
	- 	As shown below:
	image

-	Amend as follows:
	-	Replace `wwe` with `we` , The modification code is as follows.
	image


eps1lon (approved):
thanks

PR #6947, opened 2024-06-12
Changing Scroll behavior of the page

imparth7:
I raise this PR for the smooth the React.js app in-page navigation, when we navigating through the side in-page navigation that time scroll behavior is auto so I smooth that scroll behavior. ( This is my first PR in React ).


imparth7:
Now Scroll behavior of the react.dev official site is smooth.

PR #6985, opened 2024-06-20
Update meetups page by adding `React Summit US` and replacing the link for `React Advanced London`

BartoszKlonowski:
This pull request
* adds `React Summit US` to the meetups page
* replaces the guild.host based link with the direct one for the `React Advanced London`

(If the usage of `guild.host` is intentional, let me know, so that the commit can be reverted 👍)

### DIFF
--- a/src/content/community/meetups.md
+++ b/src/content/community/meetups.md
@@ -72,7 +72,7 @@ Do you have a local React.js meetup? Add it here! (Please keep the list alphabet
 ## England (UK) {/*england-uk*/}
 * [Manchester](https://www.meetup.com/Manchester-React-User-Group/)
 * [React.JS Girls London](https://www.meetup.com/ReactJS-Girls-London/)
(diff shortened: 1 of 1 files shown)


rickhanlonii:
The guild.host links are intentional, since that's how you register for the meetup and know what events are coming up. Just like how the other meetups link to meetup.com

rickhanlonii:
Is react summit a meetup or a conference? react advanced is both, is react summit too? if so, do they have a meetup link? the site we're linking too doesn't have any meetup information i could see

BartoszKlonowski:
@rickhanlonii Indeed, this is a conference, and from what I can see it's already added to the page:
https://github.com/reactjs/react.dev/blob/169d5c1820cd1514429bfac2a923e51dd782d37e/src/content/community/conferences.md?plain=1#L13
Let me close this PR as not needed.

PR #6988, opened 2024-06-21
fix: links about import & export

leedom92:

### DIFF
--- a/src/content/learn/importing-and-exporting-components.md
+++ b/src/content/learn/importing-and-exporting-components.md
@@ -59,8 +59,8 @@ These currently live in a **root component file,** named `App.js` in this exampl
 What if you want to change the landing screen in the future and put a list of science books there? Or place all the profiles somewhere else? It makes sense to move `Gallery` and `Profile` out of the root component file. This will make them more modular and reusable in other files. You can move a component in three steps:
 
 1. **Make** a new JS file to put the components in.
-2. **Export** your function component from that file (using either [default](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Statements/export#using_the_default_export) or [named](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Statements/export#using_named_exports) exports).
-3. **Import** it in the file where you’ll use the component (using the corresponding technique for importing [default](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Statements/import#importing_defaults) or [named](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Statements/import#import_a_single_export_from_a_module) exports).
+2. **Export** your function component from that file (using either [default or named](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export#description) exports).
+3. **Import** it in the file where you’ll use the component (using the corresponding technique for importing [default or named](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#description) exports).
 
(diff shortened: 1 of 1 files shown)


src/content/learn/importing-and-exporting-components.md
```diff
@@ -59,8 +59,8 @@ These currently live in a **root component file,** named `App.js` in this exampl
...
 
 1. **Make** a new JS file to put the components in.
-2. **Export** your function component from that file (using either [default](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Statements/export#using_the_default_export) or [named](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Statements/export#using_named_exports) exports).
-3. **Import** it in the file where you’ll use the component (using the corresponding technique for importing [default](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Statements/import#importing_defaults) or [named](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Statements/import#import_a_single_export_from_a_module) exports).
+2. **Export** your function component from that file (using either [default or named](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export#description) exports).
```
BartoszKlonowski, line 62: It is not the good practice to put explicit language into the link. Without it, the link will still lead to where needed, but the language can be selected either default, or redirected to what the website uses/selects for the user.

src/content/learn/importing-and-exporting-components.md
```diff
@@ -59,8 +59,8 @@ These currently live in a **root component file,** named `App.js` in this exampl
 What if you want to change the landing screen in the future and put a list of science books there? Or place all the profiles somewhere else? It makes sense to move `Gallery` and `Profile` out of the root component file. This will make them more modular and reusable in other files. You can move a component in three steps:
 
 1. **Make** a new JS file to put the components in.
-2. **Export** your function component from that file (using either [default](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Statements/export#using_the_default_export) or [named](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Statements/export#using_named_exports) exports).
```
BartoszKlonowski, line 62: I don't think there's a reason for unifying the `default` and `named`. Having them mentioned and linked separately seems to be correct as well.

src/content/learn/importing-and-exporting-components.md
```diff
@@ -59,8 +59,8 @@ These currently live in a **root component file,** named `App.js` in this exampl
...
-3. **Import** it in the file where you’ll use the component (using the corresponding technique for importing [default](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Statements/import#importing_defaults) or [named](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Statements/import#import_a_single_export_from_a_module) exports).
+2. **Export** your function component from that file (using either [default or named](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export#description) exports).
+3. **Import** it in the file where you’ll use the component (using the corresponding technique for importing [default or named](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#description) exports).
```
BartoszKlonowski, line 63: If we want to link to the `Import` page in general, there's no reason to directing to `#description`.

BartoszKlonowski (changes requested):
I don't really see a reason for modifying that part, also there's no description of the PR with any reasoning.
If there's one please provide.

src/content/learn/importing-and-exporting-components.md
leedom92: You have a point.

Issue #6987, opened 2024-06-20
[Suggestion]: Add `isPending` example to `useActionState` docs


JakeSaterlay:
### Summary

A section to describe the `isPending` value you can extract from `useActionState`

### Page

https://react.dev/reference/react/useActionState

### Details

You can do `const [state, action, isPending] = useActionState(myAction, null);`  to extract the pending state from `useActionState` however I can't see an example of this anywhere in the documentation, adding this as an example would be quite useful

JakeSaterlay:
Opened https://github.com/reactjs/react.dev/pull/6989 to work on this

PR #6989, opened 2024-06-21
`useActionState` pending example

JakeSaterlay:
This aims to describe the third item returned in the array when using the `useActionState` hook `isPending`. 

https://github.com/reactjs/react.dev/issues/6987

### DIFF
--- a/src/content/reference/react/useActionState.md
+++ b/src/content/reference/react/useActionState.md
@@ -159,17 +159,15 @@ export default function App() {
 ```js src/actions.js
 "use server";
 
-
-
 export async function addToCart(prevState, queryData) {
   const itemID = queryData.get('itemID');
   if (itemID === "1") {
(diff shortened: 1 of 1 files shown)


juligris:
Is this going to be added?

BillyDyball:
`usaActionState` is talked about in this [blog](https://react.dev/blog/2024/04/25/react-19#new-hook-useactionstate) but the docs show an incorrect return type.

This needs to be merged to fix this problem.

sebastiancarlos:
This looks great! Just to make sure we're all on the same page, could we get some confirmation from some of the core devs that the `isPending` prop in `useActionState` and its documentation here are valid at this point in time? Thanks!

eps1lon (approved):

eps1lon (approved):

PR #7007, opened 2024-07-01
chore: Refactor Accordion component to use state hook for panel activation

### LINKED ISSUE #7001: [Typo]: In Sharing State Between Components
### Summary

When explaining state lifting up, there's a small error in the example function for the **second step** :  **Pass hardcoded data from the common parent**:
```
function Panel({ title, children, isActive }) {
  return (
    <section className="panel">
      <h3>{title}</h3>
      {isActive ? (
        <p>{children}</p>
      ) : (
        <button onClick={() => setIsActive(true)}>
          Show
        </button>
      )}
    </section>
  );
}

```

The **onClick** calls **setIsActive**, but since there's no state declared, it should be removed.

### Page

https://react.dev/learn/sharing-state-between-components

### Details

_No response_

Rekl0w:
Fix #7001: chore: Refactor Accordion component to use state hook for panel activation

### DIFF
--- a/src/content/learn/sharing-state-between-components.md
+++ b/src/content/learn/sharing-state-between-components.md
@@ -133,27 +133,29 @@ In this example, it's the `Accordion` component. Since it's above both panels an
 import { useState } from 'react';
 
 export default function Accordion() {
+  const [isActive, setIsActive] = useState(false);
+
   return (
     <>
       <h2>Almaty, Kazakhstan</h2>
-      <Panel title="About" isActive={true}>
+      <Panel title="About" isActive={isActive} onShow={() => setIsActive(true)}>
         With a population of about 2 million, Almaty is Kazakhstan's largest city. From 1929 to 1997, it was its capital city.
       </Panel>
-      <Panel title="Etymology" isActive={true}>
+      <Panel title="Etymology" isActive={isActive} onShow={() => setIsActive(true)}>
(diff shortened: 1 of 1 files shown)


Rekl0w:
@rickhanlonii

Rekl0w:
@rickhanlonii

rickhanlonii:
This portion of the guide is intentionally half finished. In Step 3 below this is walks through making these changes.

src/content/learn/sharing-state-between-components.md
```diff
@@ -133,27 +133,29 @@ In this example, it's the `Accordion` component. Since it's above both panels an
 import { useState } from 'react';
 
 export default function Accordion() {
+  const [isActive, setIsActive] = useState(false);
+
   return (
     <>
       <h2>Almaty, Kazakhstan</h2>
-      <Panel title="About" isActive={true}>
+      <Panel title="About" isActive={isActive} onShow={() => setIsActive(true)}>
```
rickhanlonii: @Rekl0w this change goes too far - this is what Step 3 shows how to do. For Step 2, it should be something like:

```suggestion
      <Panel title="About" isActive={true} onShow={() => { } }>
```

And in the Step 2 text, we should explain that we're passing the state, _and an event handler to update it_ (moved from from Step 3. Then Step 3 explains how to add the state and update it in the event handler. 

Importantly, the text at the end of Step 2:

> Try editing the hardcoded isActive values in the Accordion component and see the result on the screen.

Should still work. Maybe add a sentence after that says:

> In the next step, we'll replace the hardcoded values with state in the parent updated in the `onShow` event handler.

rickhanlonii: Like this comment says: https://github.com/reactjs/react.dev/issues/7001#issuecomment-2267973635

PR #7028, opened 2024-07-05
chore: Remove startTransition's synchronous examples and infos

### LINKED ISSUE #7027: [Mistake]: React 19 says that startTransition callback must be synchronous. It is not true.
### Summary

Since this change: https://github.com/facebook/react/blob/main/CHANGELOG-canary.md#october-5-2023-1830-canary-546178f91-20231005

The line

> The function you pass to startTransition must be synchronous. React immediately executes this function, marking all state updates that happen while it executes as Transitions. If you try to perform more state updates later (for example, in a timeout), they won’t be marked as Transitions.

is no longer true.

### Page

https://19.react.dev/reference/react/useTransition

### Details

_No response_

Rekl0w:
Fix #7027: chore: Remove startTransition's synchronous examples and infos


vezaynk:
Deleting it is probably not enough. It should include documentation for the async callbacks.

PR #7031, opened 2024-07-05
Update compiler docs for Rspack and Rsbuild apps

chenjiahan:
> Rspack: The fast Rust-based web bundler with webpack-compatible API
> Rsbuild: The Rspack-based build tool.

Rspack and Rsbuild already provides documentation for using React Compiler. It would be helpful if we could show a link to the React Compiler documentation.

- https://rspack.dev/guide/tech/react#react-compiler
- https://rsbuild.dev/guide/framework/react#react-compiler


eps1lon (approved):
Thank you

PR #7047, opened 2024-07-10
Update /link/legacy-context to removal blog post

kassens:
The link is used in the strict mode warning and I'm looking to use it more. This link doesn't go to the legacy website and has more context on the removal.

I think it stays true to the purpose for links from older versions so I can use the same link in other warnings.

### DIFF
--- a/vercel.json
+++ b/vercel.json
@@ -91,7 +91,7 @@
     },
     {
       "source": "/link/legacy-context",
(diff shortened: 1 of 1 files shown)


jackpope (approved):
This would usually be a separate warning page like https://react.dev/warnings/react-test-renderer . But this still seems like an improvement

PR #7050, opened 2024-07-11
Add both Meetups in Chennai

CarreraPHP:
Add both Meetups in Chennai


gsathya:
Let's add this if this meetup is active, but I don't see any past or upcoming events on the meetup page?

rickhanlonii:
Are these still active?

PR #7057, opened 2024-07-12
placekitten.com is not available anymore

smith558:


rickhanlonii:
Fixed in https://github.com/reactjs/react.dev/pull/7189

PR #7061, opened 2024-07-13
docs(input): convert `age` to string before setting to match initial type

EastSun5566:
Hi React team,

Thanks for the great doc!

I noticed in the following file that the initial state of the `age` is set as a string:

https://github.com/reactjs/react.dev/blob/4c91abc78b20be10e7d40cf57a80a6a6247e6e9a/src/content/reference/react-dom/components/input.md?plain=1#L351

https://github.com/reactjs/react.dev/blob/4c91abc78b20be10e7d40cf57a80a6a6247e6e9a/src/content/reference/react-dom/components/input.md?plain=1#L376

Given this, I believe converting the `age` back to a string would be more consistent before setting it with `setAge`. This would maintain type consistency throughout the component.

Thank you!

### DIFF
--- a/src/content/reference/react-dom/components/input.md
+++ b/src/content/reference/react-dom/components/input.md
@@ -359,7 +359,7 @@ function Form() {
           onChange={e => setAge(e.target.value)}
           type="number"
         />
-        <button onClick={() => setAge(ageAsNumber + 10)}>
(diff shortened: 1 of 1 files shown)


rickhanlonii:
Kinda the point of this is that they don't have to be the same type. That's why it does:

```js
const ageAsNumber = Number(age);
```

PR #7066, opened 2024-07-18
Fix typo

smith558:


rickhanlonii:
Fixed in https://github.com/reactjs/react.dev/pull/7182, sorry I saw that first.

smith558:
> Fixed in #7182, sorry I saw that first.

That's a 2nd one!

PR #7073, opened 2024-07-20
fix: capitalize react concept Action

Jay-Karia:
This PR fixes capitalization of word `Actions` for issue #6713


Jay-Karia:
Can we go for **Trademark request** ? (if possible)

rickhanlonii (changes requested):
Most of these are referring to reducer actions which are different than Actions.

Jay-Karia:
Will look at it and update it soon.

PR #7096, opened 2024-08-07
Document behaviour of setting state inside `useLayoutEffect`

OliverJAsh:
As discussed: https://github.com/facebook/react/issues/17334#issuecomment-2272162502

@eps1lon You will probably have a better idea of how to word this. I was also wondering if we should add a note under "[Measuring layout before the browser repaints the screen](https://react.dev/reference/react/useLayoutEffect#measuring-layout-before-the-browser-repaints-the-screen)". This is exactly what I was trying to achieve before realising the behaviour with regards to `useEffect`.

### DIFF
--- a/src/content/reference/react/useLayoutEffect.md
+++ b/src/content/reference/react/useLayoutEffect.md
@@ -67,7 +67,7 @@ function Tooltip() {
 
 * The code inside `useLayoutEffect` and all state updates scheduled from it **block the browser from repainting the screen.** When used excessively, this makes your app slow. When possible, prefer [`useEffect`.](/reference/react/useEffect)
 
-* If you trigger a render inside `useLayoutEffect`, React will flush all remaining effects including `useEffect`.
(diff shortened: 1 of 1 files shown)


src/content/reference/react/useLayoutEffect.md
```diff
@@ -67,6 +67,8 @@ function Tooltip() {
 
 * The code inside `useLayoutEffect` and all state updates scheduled from it **block the browser from repainting the screen.** When used excessively, this makes your app slow. When possible, prefer [`useEffect`.](/reference/react/useEffect)
 
+* If you trigger a render inside `useLayoutEffect`, React will flush all remaining effects including `useEffect`.
```
eps1lon, line 70: do we use "flush" somewhere else? It sounds a bit to jargon-y.

We also use uppercase words for React specific stuff so we use "Effect". 

```suggestion
* If you trigger a state update inside `useLayoutEffect`, React will execute all remaining Effects immediately including `useEffect`.
```

eps1lon (approved):
thank you!

PR #7103, opened 2024-08-13
[Draft] Do Not Merge. Initial styling of a react.dev/careers page

TheSavior:
We are exploring having a react.dev/careers page with open positions we are hiring for. Putting this page together to prepare for when we have positions to hire for and validate the design first. 

The content prose in this PR doesn't make any sense, it has snippets from different random job postings. None of which are reflective of roles we are considering hiring for. 

I'm looking for feedback on the design elements: site footer, careers sidebar, careers landing page, role description / apply.

I also need to have a version of the careers page that shows a message when we have no open positions (while being able to leave code in the codebase as an example for the next person to see how to add one.


src/content/careers/index.md
```diff
@@ -0,0 +1,19 @@
+---
+title: React Careers
+---
+
+<Intro>
+
+This is where you can find a career on the React team
```
TheSavior, line 7: This needs better prose

TheSavior:
Feedback received offline:

Add the Meta logo to make it even more clear that this is a Meta role. 

Consider adding a note at the top noting that this is a generic meta job posting and applying is through Meta.

PR #7105, opened 2024-08-13
fix: lint issues blocking Merge Action

CarreraPHP:
Disable Eslint for code that is not supported by Eslint & also add key prop for the component within map function issues in code,

### DIFF
--- a/src/components/Layout/HomeContent.js
+++ b/src/components/Layout/HomeContent.js
@@ -915,6 +915,7 @@ function Example1() {
           isFromPackageImport={false}
           noShadow={true}
           noMargin={true}>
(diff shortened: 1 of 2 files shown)


Neoxs:
🚀 Just completed an ESLint v9 migration, and wow… what a ride! 😵‍💫

I ran into breaking changes, weird errors, and unexpected surprises—but I documented everything so you don’t have to suffer the same way.

If you’re upgrading (or still hesitating), this guide might save you some headaches:
📖 [neoxs.me/blog/migration-to-eslint-v9](https://www.neoxs.me/blog/migration-to-eslint-v9)

PR #7107, opened 2024-08-14
reference

eps1lon:
[Preview]()


src/content/reference/react-markup/renderToHTML.md
```diff
@@ -0,0 +1,74 @@
...
+
+You can call `renderToHTML` to render a non-interactive tree of React components to HTML.
+By default, it supports shared components and built-in components.
+Server Components are only allowed if used in a React Server environment (e.g. in Server Actions).
+You can also use it during Client-Side Rendering but only without Server Components.
+
+When a `<html>` tag is rendered, `renderToHTML` will automatically add `<!DOCTYPE html>` doctype.
+
+```js
+import { experimental_renderToHTML as renderToHTML } from 'react-dom/server';
```
slorber, line 33:
```suggestion
import { experimental_renderToHTML as renderToHTML } from 'react-markup';
```

looks like an unmodified copy/paste

src/content/reference/react-markup/renderToHTML.md
```diff
@@ -0,0 +1,74 @@
...
+---
+
+## Usage {/*usage*/}
+
+### Rendering a React tree as an Email {/*rendering-a-react-tree-as-an-email*/}
+
+Await the call of `renderToHTML` :
+
+```js {7}
+import { experimental_renderToHTML as renderToHTML } from 'react-markup';
```
harish-sethuraman, line 64: BTW there seems to be a package with the exact name already? https://www.npmjs.com/package/react-markup

slorber: also saw that it it looks like @eps1lon took it over

CleanShot 2024-08-14 at 17 42 08

eps1lon: Published a new major version as a placeholder to avoid confusion: https://www.npmjs.com/package/react-markup

harish-sethuraman: Also I see that the versioning starts with 2.0.0. It was whole different package before that? Is it possible to remove/unpublish the older packages before the downloads kick in? Because in ESLint npm didn't allow us to un publish package when it start getting downloads. They informed us to mark it as deprecated instead.

eps1lon: We deprecated earlier versions. I don't think we should unpublish. Even though it has only 3 weekly downloads, people were still using it so it feels overly aggressive to remove references considering the previous code is harmless.

rickhanlonii:
@eps1lon should these be added to the 19 docs instead (base branch v19 instead of main)?

eps1lon:
We need to get https://github.com/facebook/react/pull/30736 working first.

PR #7128, opened 2024-08-26
Adds React Summit US to the list of conferences

PixelsCommander:


TheSavior:
Thanks, this has been added through another PR!

PR #7132, opened 2024-08-28
Remove wrong reference of window

youngvform:
https://developer.mozilla.org/en-US/docs/Web/API/Document/title
window contains document property so i remove reference of window.


rickhanlonii (approved):

rickhanlonii:
Nice catch

PR #7143, opened 2024-09-03
Explicitly redirect from legacy 'thinking-in-react' page

### LINKED ISSUE #7142: The Link do not work
https://react.dev/docs/thinking-in-react.html

Not Found
This page doesn’t exist.

If this is a mistake, [let us know](https://github.com/reactjs/react.dev/issues/new), and we will try to fix it!

BartoszKlonowski:
This pull request fixes #7142 

Although the `thinking-in-react` page redirect already works for https://reactjs.org/docs/thinking-in-react.html, making it explicit redirect also fixes the scenario where the link has the legacy path `/docs/` but the current website `react.dev`.

### DIFF
--- a/vercel.json
+++ b/vercel.json
@@ -15,6 +15,11 @@
       "permanent": true
     },
     {
+      "source": "/docs/thinking-in-react",
+      "destination": "/learn/thinking-in-react",
+      "permanent": false
+    },
+    {
       "source": "/learn/meet-the-team",
       "destination": "/community/team",
       "permanent": true


sophiebits:
Hmm, are we linking to this path somewhere? I would have thought that this path has never been valid on this domain so we shouldn't need a redirect.

rickhanlonii:
Similar to the comment here, we don't do redirects like this: https://github.com/reactjs/react.dev/pull/7162#issuecomment-2383673202

Issue #7151, opened 2024-09-09
[Bug]: React 16 docs redirect to React 17 with no explanation


jonrajavuori-academia:
### Summary

The 'Previous versions' page does not mention that the React 16 docs don't exist because the React 17 docs covers it, nor do the React 17 docs themselves. This is confusing and doesn't leave any way to understand more about the seemingly broken flow.

### Page

https://react.dev/versions

### Details

Disclaimer: This has previously been filed at https://github.com/reactjs/react.dev/issues/7011, and other prior discussion is at https://github.com/facebook/react/issues/24802. Please feel free to immediately close this issue if the status remains "won't fix." However, I'd like to make the case for it again.

# Proposed solutions

## Solution 1: React 17 docs header

Add a clear header to the React 17 docs, perhaps only visible when redirected from the React 16 docs URL. The header could explicitly state that React 17 is close enough to React 16's API that the docs are usable for both, with a link to a page explaining in more detail and listing the few important changes. I feel that this would be well worth the time as a big improvement to the current flow, which is confusing and probably leads many developers to hunt for closed issues around this, like I did.

The advantage of this solution is that it would handle links to the React 16 documentation coming from anywhere, from bookmarks / blogs / etc.

## Solution 2: Change copy of 'Previous versions' page

A more lightweight solution that would require fewer changes would be to update the 'Previous versions' page to mention this situation explicitly, and to only point to the React 17 docs. This would at least reduce confusion for developers arriving at that page.

# Problem

Here is how my flow went, and I'm sure this is typical (at least it is for the several people who took the time to comment on the bugs above, and probably many more since the bugs are closed and most people won't take the time):

- Currently use React 16 and want to refer to the docs for an API I already know has been changing from 15 -> 18, but am unsure how exactly. In other words, I definitely want the React 16 docs. I've been around this block before with many APIs, and I don't yet have the special knowledge that 16/17 are very close.
- Visit the React docs at https://react.dev/
- Search for 'Previous versions'
- Find this page, leading me to reasonably believe that there are in fact up-to-date React 16 docs and that I can reach them by clicking this "16.react.dev" link:

<img width="1028" alt="image" src="https://github.com/user-attachments/assets/57496982-6f0b-4b6c-9485-6a38110d0f60">

[...]

PR #7153, opened 2024-09-10
doc: Update link to React v16.0 documentation

Rekl0w:
doc: Update link to React v16.0 documentation

Releated: #7151


rickhanlonii:
The TODO here is to fix the deployment of the 16 docs. Until then, linking to 17 is fine because there's really no differences between 16 and 17.

PR #7156, opened 2024-09-11
docs(act.md): correct ReactDOM to ReactDOMClient

regchiu:
Correct `ReactDOM` to `ReactDOMClient` in `act.md`.

### DIFF
--- a/src/content/reference/react/act.md
+++ b/src/content/reference/react/act.md
@@ -90,7 +90,7 @@ To test the render output of a component, wrap the render inside `act()`:
 
 ```js  {10,12}
 import {act} from 'react';
-import ReactDOM from 'react-dom/client';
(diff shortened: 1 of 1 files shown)


src/content/reference/react/act.md
```diff
@@ -127,7 +127,7 @@ it.only('can render and update a counter', async () => {
   document.body.appendChild(container);
   
   await act( async () => {
-    ReactDOMClient.createRoot(container).render(<Counter />);
+    ReactDOM.createRoot(container).render(<Counter />);
```
rickhanlonii, line 130: Could you switch the import to ReactDOMClient instead?

regchiu: OK.

rickhanlonii (approved):

PR #7157, opened 2024-09-11
docs: fix name of `MUI` and reference link

soartec-lab:
Hi, At first, thank you for maintenance this site.
Corrected the name and reference list of `MUI` since `MUI` was rebranded from `material-ui` to `MUI`. Check it out.


rickhanlonii:
I feel like "Material UI" is right? From the link:

> Material UI is an open-source React component library that implements Google's Material Design.

soartec-lab:
Oops, this seems to be confusing because there are two types of notation. thank you. I'll close this.

Issue #7158, opened 2024-09-11
`rss.js` page error for the author, date and description


imparth7:
### Summary

When I start the react.dev in my local system that time I get this error.

This error comes because of some of the react.dev markdown file doesn't have an author, date, or description. It can be solve using proper error handling.

### Page

https://react.dev/

### Details

When I comment that unhandled area of the code then it run properly.

rammba:
Related to #7085.

imparth7:
I fix that issue in #7159, but this is not merged at that time so the issue is continued in others.

rickhanlonii:
Fixed by https://github.com/reactjs/react.dev/pull/7159

PR #7159, opened 2024-09-11
fix:#7158 `rss.js` issue

imparth7:
The problem is getting the basename from the file path in the `rss.js` file.

Code before my changes
``` js
```

Code after my changes
``` js
```

issue: #7158


rickhanlonii (approved):
Thanks, tested locally and works great

Issue #7160, opened 2024-09-11
Always Fail PR at analyze job.


imparth7:
Recently I started solving react.dev issues and when I solve any issue or raise PR for that change my PR request cannot Merge because of a failed Analyze Bundle.

Error: This request has been automatically failed because it uses a deprecated version of `actions/upload-artifact: v2`. Learn more: https://github.blog/changelog/2024-02-13-deprecation-notice-v1-and-v2-of-the-artifact-actions/

Is there any one who can guide me for that issue or the my mistakes to contributing on react.dev.

alinkedd:
Duplicate of #7148

PR #7161, opened 2024-09-11
Update workflow version from 2 to 3.

imparth7:
In the react.dev workflow I facing the analyze job rejection issue because of the old version of the action.

Issue: #7160


harish-sethuraman (approved):
Thank you!

PR #7162, opened 2024-09-11
Redirect legacy events page to commmon components overview

### LINKED ISSUE #6940: Events are missing
Events are missing

https://react.dev/docs/events.html

BartoszKlonowski:
This pull request closes #6940 
It redirects the `docs/events` page from legacy to `Common components (e.g. <div>)` overview page where all events handlers are listed.

### DIFF
--- a/vercel.json
+++ b/vercel.json
@@ -30,6 +30,11 @@
       "permanent": false
     },
     {
+      "source": "/docs/events",
+      "destination": "/reference/react-dom/components/common",
+      "permanent": false
+    },
+    {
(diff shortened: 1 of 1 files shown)


rickhanlonii:
We don't do redirects like this, any existing links to this page should be at `reactjs.org/docs/events.html` which will correctly redirect to `https://legacy.reactjs.org/docs/events.html`, which has a note that links to the new docs page. If anyone is linking to `react.dev/docs/events.html` then the link is broken upstream, and whoever is linking to it should fix it there.

PR #7166, opened 2024-09-16
fixing:reactjs#7158 issue

imparth7:
The problem is getting the basename from the file path in the `rss.js` file.

Code before my changes
``` js
```

Code after my changes
``` js
```

issue: #7158

### DIFF
--- a/vercel.json
+++ b/vercel.json
@@ -30,6 +30,11 @@
       "permanent": false
     },
     {
+      "source": "/docs/events",
+      "destination": "/reference/react-dom/components/common",
+      "permanent": false
+    },
+    {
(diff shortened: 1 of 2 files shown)


vercel.json
```diff
@@ -29,6 +29,11 @@
       "destination": "/learn/rendering-lists#keeping-list-items-in-order-with-key",
       "permanent": false
     },
+    {
+      "source": "/docs/events",
+      "destination": "/reference/react-dom/components/common",
+      "permanent": false
```
rickhanlonii, line 35: Why?

rickhanlonii:
Dupe of https://github.com/reactjs/react.dev/pull/7159

Issue #7134, opened 2024-08-29
[Suggestion]: use toSorted and toReversed for the examples in Updating Array in State Page


Snaylaker:
### Summary

Update this section  [Updating arrays without mutation ](https://react.dev/learn/updating-arrays-in-state#updating-arrays-without-mutation) by adding  [toReversed()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toReversed) and  [toSorted()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSorted)  instead of suggestion cloning the array and mutating it.

### Page

https://react.dev/learn/updating-arrays-in-state

### Details

Since  [toReversed()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toReversed) and  [toSorted()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSorted) have been added to Javascript, should the docs reflect the changes and remove the  section about mutating the array before sorting and reversing?

Or does it stay just to give an example of cloning before mutating an array ?

BrianWoolfolk:
I would like to work on this issue

BrianWoolfolk:
You can see my PR #7174 for this issue.

I considered appropriate to also add a Note block regarding the similarities between the new methods and the spread syntax.

PR #7174, opened 2024-09-19
Added note for toSorted and toReversed methods (Issue #7134)

BrianWoolfolk:
This PR addresses the issue #7134 regarding the new JavaScript methods `toSort()` and `toReversed()`.

The approach here was to mention the existence of these new methods in the table at the beginning of the page [Updating Arrays in State](https://react.dev/learn/updating-arrays-in-state#making-other-changes-to-an-array) (and made the table bigger).

**Before**: "copy the array first".
**Now**: "`toReversed`, `toSorted`, copy the array first".

Also, in the [Making other changes to an array](https://react.dev/learn/updating-arrays-in-state#making-other-changes-to-an-array) section, there's now a **Note** block regarding the similitudes with these new methods and the JS spread syntax.

### DIFF
--- a/src/content/learn/updating-arrays-in-state.md
+++ b/src/content/learn/updating-arrays-in-state.md
@@ -24,12 +24,12 @@ Instead, every time you want to update an array, you'll want to pass a *new* arr
 
 Here is a reference table of common array operations. When dealing with arrays inside React state, you will need to avoid the methods in the left column, and instead prefer the methods in the right column:
 
-|           | avoid (mutates the array)           | prefer (returns a new array)                                        |
-| --------- | ----------------------------------- | ------------------------------------------------------------------- |
-| adding    | `push`, `unshift`                   | `concat`, `[...arr]` spread syntax ([example](#adding-to-an-array)) |
-| removing  | `pop`, `shift`, `splice`            | `filter`, `slice` ([example](#removing-from-an-array))              |
-| replacing | `splice`, `arr[i] = ...` assignment | `map` ([example](#replacing-items-in-an-array))                     |
-| sorting   | `reverse`, `sort`                   | copy the array first ([example](#making-other-changes-to-an-array)) |
+|           | avoid (mutates the array)           | prefer (returns a new array)                                                                  |
+| --------- | ----------------------------------- | --------------------------------------------------------------------------------------------- |
+| adding    | `push`, `unshift`                   | `concat`, `[...arr]` spread syntax ([example](#adding-to-an-array))                           |
+| removing  | `pop`, `shift`, `splice`            | `filter`, `slice` ([example](#removing-from-an-array))                                        |
+| replacing | `splice`, `arr[i] = ...` assignment | `map` ([example](#replacing-items-in-an-array))                                               |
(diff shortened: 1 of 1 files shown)


homayoun-mohammadii:
there's now a Note block regarding the similitudes with these new methods and the JS spread syntax.

rickhanlonii (changes requested):
Instead of documenting the old way, and having a note about the new way, can you flip the docs to document the new way with a note on how to do it in older versions?

rickhanlonii:
To do that, you'd need to come up with another example of something you can’t do with the spread syntax and non-mutating methods, and I'd add a section specific to "Sorting items in an array".

BrianWoolfolk:
Gotcha, although I can't think of any other examples aside from the sorting/reversing ones. There's also a "toSpliced" method for deletion and/or insertion that I think would be worth mentioning.

The problem I had with this is that I consider best practice to use the spread syntax every time we work with an array (or even objects) to correctly tell React we're modifying the array in some way.

I decided to go with a note on "how to do things now" because technically we can use "toSpliced" to add, remove, replace and insert, but that would mean to modify the entire sections above (the ones that teaches you to use the spread syntax).

I you allow me; I can change the initial sections to demonstrate the use of toSpliced (and the toReversed/toSorted methods too) and create at the end a new section for "spread syntax and it's benefits", explicitly saying that it allows us to change the *memory direction* of the array so React can see it as a new array entirely.

BrianWoolfolk:
@rickhanlonii, I was thinking of creating a section for the [structuredClone() method](https://developer.mozilla.org/en-US/docs/Web/API/Window/structuredClone) inside both docs of "Updating Objects/Arrays in State".

This provides a deep clone of the array/object that could easily replace the need for spread syntax, and even enable us to directly mutate the cloned object (take the [original example](https://react.dev/learn/updating-objects-in-state#updating-a-nested-object) from React.dev and we could mutate the cloned object just like `person.artwork.city = 'New Delhi'`).

I've worked with structuredClone before, and I think it can be more useful than the spread syntax

npar0005:
> I was thinking of creating a section for the [structuredClone() method](https://developer.mozilla.org/en-US/docs/Web/API/Window/structuredClone) inside both docs of "Updating Objects/Arrays in State".
> 
> This provides a deep clone of the array/object that could easily replace the need for spread syntax, and even enable us to directly mutate the cloned object (take the [original example](https://react.dev/learn/updating-objects-in-state#updating-a-nested-object) from React.dev and we could mutate the cloned object just like `person.artwork.city = 'New Delhi'`).
> 
> I've worked with structuredClone before, and I think it can be more useful than the spread syntax

The issue with using `structuredClone()` is that it clones all nested objects. You wouldn't/shouldn't be creating new object references for nested data if they aren't changing, as that could lead to unnecessary rerenders. 
For the example in the docs, using `structuredClone()` is fine as there's just one nested object that needs to be changed, but if there were more nested objects that weren't changing, then using `structuredClone()` would be overkill there. So we'd have to carefully spell out to readers when `structuredClone()` would be appropriate to use vs when to use the spread syntax (for finer control over what object references change for the state update).

PR #7175, opened 2024-09-19
Nit: server actions can't be passed events

rickhanlonii:
The current example would error with:

> Error: Only plain objects, and a few built-ins, can be passed to Server Actions. Classes or null prototypes are not supported

The issue is that the `onClick` passes an event handler to the server function. 

Thanks [`@bayes`](https://x.com/bayes/status/1835790101378421102) for reporting.


rickhanlonii:
Also filed https://github.com/facebook/react/pull/31005 to update the text to say "server function" instead of "server action"

PR #7179, opened 2024-09-22
Docs: Fix typos in thinking-in-react.md

rammba:


src/content/learn/thinking-in-react.md
```diff
@@ -197,7 +197,7 @@ td {
 
 (If this code looks intimidating, go through the [Quick Start](/learn/) first!)
 
-After building your components, you'll have a library of reusable components that render your data model. Because this is a static app, the components will only return JSX. The component at the top of the hierarchy (`FilterableProductTable`) will take your data model as a prop. This is called _one-way data flow_ because the data flows down from the top-level component to the ones at the bottom of the tree.
+After building your components, you'll have a library of reusable components that renders your data model. Because this is a static app, the components will only return JSX. The component at the top of the hierarchy (`FilterableProductTable`) will take your data model as a prop. This is called _one-way data flow_ because the data flows down from the top-level component to the ones at the bottom of the tree.
```
rickhanlonii, line 200:
```suggestion
After building your components, you'll have a library of reusable components that render your data model. Because this is a static app, the components will only return JSX. The component at the top of the hierarchy (`FilterableProductTable`) will take your data model as a prop. This is called _one-way data flow_ because the data flows down from the top-level component to the ones at the bottom of the tree.
```

I think "render" is correct here. The "render" refers to the plural "components" not the singular "library". It's not that the library renders, it's that the components render.

rickhanlonii (approved):

PR #7180, opened 2024-09-22
[19] s/"Server Action"/"Server Function"

rickhanlonii:
## Overview

This PRs re-names "Server Actions" to "Server Functions" in the docs starting in React 19.

If you review the changes here, you can actually see the reason this makes sense: Many of the calls to "use server" functions are not actions. 

For example, this example from the Server "Action" docs is not an Action:

```js
"use client";
import {createNote} from './actions';

function EmptyNote() {
  console.log(createNote);
  // {$$typeof: Symbol.for("react.server.reference"), $$id: 'createNoteAction'}
  <button onClick={() => createNote()} />
}
```

This is just calling a server function on click. Really, it should be converted to an Action for better UX. 

## What are Server Actions then

Server Actions are Server Functions used in an Action context. For example, there are three examples on the Server Function page in this PR that show:

- Using Server Functions with Actions 
- Form Actions with Server Functions
- Server Functions with `useActionState`

## Redirect

This change redirects links to "/server-action" to the 18 docs, which I'll add a callout for saying it's been renamed. We should also probably update the 19 blog post with a note about the name change.


src/content/blog/2023/03/22/react-labs-what-we-have-been-working-on-march-2023.md
```diff
@@ -29,7 +29,7 @@ Since our last update, we have merged the [React Server Components RFC](https://
 
 The biggest change is that we introduced [`async` / `await`](https://github.com/reactjs/rfcs/pull/229) as the primary way to do data fetching from Server Components. We also plan to support data loading from the client by introducing a new Hook called `use` that unwraps Promises. Although we can't support `async / await` in arbitrary components in client-only apps, we plan to add support for it when you structure your client-only app similar to how RSC apps are structured.
 
-Now that we have data fetching pretty well sorted, we're exploring the other direction: sending data from the client to the server, so that you can execute database mutations and implement forms. We're doing this by letting you pass Server Action functions across the server/client boundary, which the client can then call, providing seamless RPC. Server Actions also give you progressively enhanced forms before JavaScript loads.
```
sophiebits, line 32: Can we leave old blog posts unedited? I think it’s a bit confusing to change them.

vercel.json
```diff
@@ -184,6 +184,11 @@
       "destination": "/reference/rsc/use-server",
       "permanent": true
     },
+    {
+      "source": "/reference/rsc/server-actions",
+      "destination": "https://18.react.dev/reference/rsc/server-actions",
```
sophiebits, line 189: I don’t think we want people to end up on the 18. site just because they followed a link to server actions docs. Given that the content of the new server functions page is almost the same can we just link to it?

sophiebits (approved):
…are server components server functions

slorber:
Just passing by to report some details that confused me a bit 

---

On the v19 docs site there are still examples using `createNoteAction` while the function is not an action. Maybe it would be better to rename these as well?

CleanShot 2024-09-30 at 11 02 09


---

Also wondering if there's a canonical place in docs that explains what are actions and server actions (as far as I understand an action is not always necessarily a server action? 🤔 ) 

As far as I know, it's explained in the v19 RC blog post:

CleanShot 2024-09-30 at 11 04 27

But I'm not sure it's explained in the docs yet, or I couldn't find it. Maybe having a docs page about the concept of action could be useful?

src/content/reference/rsc/server-functions.md
```diff
@@ -0,0 +1,222 @@
...
+import {updateName} from './actions';
+
+function UpdateName() {
+  const [name, setName] = useState('');
+  const [error, setError] = useState(null);
+
+  const [isPending, startTransition] = useTransition();
+
+  const submitAction = async () => {
+    startTransition(async () => {
```
slorber, line 127: Also wondered here why `submitAction` is async considering `startTransition` returns nothing

Is wrapping in transition necessary in this case when using `<form action>`?

PR #7181, opened 2024-09-22
Add stable fn notes to useMemo, useTransition, useState, and useReducer

rickhanlonii:
`useCallback` has this, but not `useState`, `useReducer`, `useMemo`, or `useTransition`.

### DIFF
--- a/src/content/reference/react/useMemo.md
+++ b/src/content/reference/react/useMemo.md
@@ -1056,6 +1056,83 @@ Keep in mind that you need to run React in production mode, disable [React Devel
 
 ---
 
+### Preventing an Effect from firing too often {/*preventing-an-effect-from-firing-too-often*/}
+
+Sometimes, you might want to use a value inside an [Effect:](/learn/synchronizing-with-effects)
+
+```js {4-7,10}
+function ChatRoom({ roomId }) {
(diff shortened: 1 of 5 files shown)


src/content/reference/react/useMemo.md
```diff
@@ -1056,6 +1056,83 @@ Keep in mind that you need to run React in production mode, disable [React Devel
...
+
+  const options = useMemo(() => {
+    return {
+      serverUrl: 'https://localhost:1234',
+      roomId: roomId
+    };
+  }, [roomId]); // ✅ Only changes when roomId changes
+
+  useEffect(() => {
+    const options = createOptions();
```
Mamaduka, line 1104: I think this is leftover from the `useCallback` example.

src/content/reference/react/useMemo.md
```diff
@@ -1056,6 +1056,83 @@ Keep in mind that you need to run React in production mode, disable [React Devel
...
+    const connection = createConnection(options);
+    connection.connect();
+    return () => connection.disconnect();
+  }, [options]); // ✅ Only changes when createOptions changes
```
Mamaduka, line 1108: > // ✅ Only changes when cached `options` object changes

PR #7183, opened 2024-09-23
Fix search results ranking and grouping

rickhanlonii:
## Overview

Shout out @markerikson for flagging the search issue and @shortcuts for the help fixing it!

Makes a few improvements to searching, some of which are applied to the search index already outside of this PR:

- Fixes ranking so that React APIs and Learn are ranked at the top again. I think this regressed at some point. The fix was to move the `pageRank` attribute above `exact` in the sorting and ranking settings.
- Split up the "React APIs" into different groups:
  - React APIs
  - React DOM APIs
  - React Server Components
  - Rules of React
  - Legacy React APIs (downranked)

The changes included here are:
- Display the groups again (previously it was just a flat list with no grouping headings like React APIs)
- Display more results (previously there were a very small number of results)

To do that I updated DocSearch depts, and added/updated the `searchParameters`.

## Before

Screenshot 2024-09-23 at 3 35 08 PM

## After

Screenshot 2024-09-23 at 3 32 01 PM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,14544 @@
+{
+  "name": "react-dev",
+  "version": "1.0.0",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "react-dev",
+      "version": "1.0.0",
+      "hasInstallScript": true,
+      "license": "CC",
+      "dependencies": {
+        "@codesandbox/sandpack-react": "2.13.5",
+        "@docsearch/css": "^3.6.1",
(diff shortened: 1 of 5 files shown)


markerikson:
Looks great!

package-lock.json
poteto, line 1: I think this might have been accidentally added?

shortcuts (commented):
looks perfect! the bug is being fixed on our side too :)

package-lock.json
rickhanlonii: oh snap fixed here https://github.com/reactjs/react.dev/pull/7184

PR #7188, opened 2024-09-25
Replaced outdated placekitten calls with placeCats and loremflikcr URLS

pauljones0:
placeKittens no longer is working as intended after months of inconsistency. I have replaced the placeKittens calls with urls from loremflickr (replicating existing code for loop iterations) and that I have checked, along with placeCat calls (for individual, labelled calls.) I was unable to test as I wasn't comfortable installing an old yarn version, and it wouldn't build with the latest yarn version.

I feel that the changes are small enough that if this causes issues, a revert is simple. I doubt there will be any issues though.


pauljones0:
messed this MR up. Going to close this PR and try again

PR #7200, opened 2024-10-01
more renames from actions to functions

MelvinVermeer:
If have changed the following to prevent confusion. In these cases `updateName` and `createNote` will most likely be Server Functions, importing them from `./actions` will not help in understanding the new terminology.

```diff
--import {updateName} from './actions';
++import {updateName} from './server';
```

The same reasoning applies to the rename from `createNoteAction` to `createNote`.


rickhanlonii (changes requested):
Nice catch!

Can you change from `./functions` to `./server`?

MelvinVermeer:
@rickhanlonii sure thing ✅

MelvinVermeer:
@rickhanlonii do you require me to make any more changens?

PR #7201, opened 2024-10-01
Clarify Server Function caveats

rickhanlonii:
Nice catch by @dai-shi https://x.com/dai_shi/status/1840967455101108241

### DIFF
--- a/src/content/reference/rsc/use-server.md
+++ b/src/content/reference/rsc/use-server.md
@@ -43,8 +43,9 @@ Instead of individually marking functions with `'use server'`, you can add the d
 * To import a Server Functions from [client code](/reference/rsc/use-client), the directive must be used on a module level.
 * Because the underlying network calls are always asynchronous, `'use server'` can only be used on async functions.
 * Always treat arguments to Server Functions as untrusted input and authorize any mutations. See [security considerations](#security).
-* Server Functions should be called in a [Transition](/reference/react/useTransition). Server Functions passed to [`<form action>`](/reference/react-dom/components/form#props) or [`formAction`](/reference/react-dom/components/input#props) will automatically be called in a transition.
-* Server Functions are designed for mutations that update server-side state; they are not recommended for data fetching. Accordingly, frameworks implementing Server Functions typically process one action at a time and do not have a way to cache the return value.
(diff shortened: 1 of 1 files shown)


src/content/reference/rsc/use-server.md
```diff
@@ -43,8 +43,9 @@ Instead of individually marking functions with `'use server'`, you can add the d
...
 * Always treat arguments to Server Functions as untrusted input and authorize any mutations. See [security considerations](#security).
-* Server Functions should be called in a [Transition](/reference/react/useTransition). Server Functions passed to [`<form action>`](/reference/react-dom/components/form#props) or [`formAction`](/reference/react-dom/components/input#props) will automatically be called in a transition.
-* Server Functions are designed for mutations that update server-side state; they are not recommended for data fetching. Accordingly, frameworks implementing Server Functions typically process one action at a time and do not have a way to cache the return value.
+* Server Functions are typically used in Actions to update server-side state; if you need to fetch data from the server, consider using a [Server Component](/reference/rsc/server-components).
+* By convention, Server Functions used in Actions such as inside a [Transition](/reference/react/useTransition), passed to [`<form action>`](/reference/react-dom/components/form#props), or passed to [`formAction`](/reference/react-dom/components/input#props) are called Server Actions.
```
sophiebits, line 47: Hmm some Transitions are navigations right? And not Actions.

sophiebits (approved):

src/content/reference/rsc/use-server.md
rickhanlonii: Yeah, I want this to say "in Actions" and link directly to that section of Transitions, but that doesn't exist yet. I didn't intend for this to imply all Transitions are Actions, any idea how to clarify?

sophiebits: How about like:

By convention, Server Functions that perform a mutation of some kind (i.e., that do not merely fetch data) and are called in a context that integrates with React's handling of async operations (such as inside …) are called Server Actions.

That's wordy and maybe can be refactored but I think this is correct? For Actions in general I think it means:
* State updates are transition priority [server actions couldn't possibly trigger sync updates, of course]
* Ties into isPending, optimistic, and error handling
* Might have side effects (other than calling setState) so must be called in a context where React will call them exactly once / can't be interrupted or rebased

Note that the first two are true of navigation-style transitions too.

PR #7205, opened 2024-10-03
Fix: remove broken meetup links

juliogarciape:
I was reviewing the meetups page and noticed that there are many broken links. Several groups that were on [meetup.com](https://meetup.com) no longer appear, possibly because they are inactive or have changed their names. 

Therefore, I decided to remove those links, as they only redirected to a page like this:

Screenshot

I ended up removing some countries because they only had one link, and it was broken, leaving countries without any links


harish-sethuraman (approved):
OMG! thanks for taking time to find these broken links

juliogarciape:
> OMG! thanks for taking time to find these broken links

I have some free time and I enjoy contributing to large projects like React, even if it's just to their documentation. If you need help with similar cases, I'd be happy to assist.

Issue #7203, opened 2024-10-03
[Suggestion]: Add Collapse Feature for Left-Side Menu


aojunhao123:
### Summary

The left-side navigation menu in the React official Chinese documentation currently lacks a collapse feature. When navigating through long lists, users have to scroll to the bottom to access other sections. This can be inconvenient, especially when switching between categories such as “Hook” or “Components”.

<img width="1512" alt="image" src="https://github.com/user-attachments/assets/940e0076-74d5-47b0-bc42-204cfadc794b">

### Details

**Proposal:** 
Add a collapse/expand functionality for the left-side navigation menu. This would allow users to collapse currently opened sections and easily navigate to other sections without extensive scrolling.

**Additional Information:**
I would be happy to contribute to this feature and submit a pull request to implement the collapse functionality. Please let me know if this is something the team is open to, and I can start working on it.

Thank you for considering this feature request!

rammba:
I would like to see this feature as well.

Ayush-1110:
we can build togather

rammba:
It would be great to implement this one, but I'm not sure if @rickhanlonii or some other maintainer agrees with the proposal?

PR #7226, opened 2024-10-14
Add expand/collapse sidebar handler to the arrow icon

bauertim:
### Overview

Added manual expand/collapse toggle handler for the arrow icon in the sidebar while keeping the original design, functionality and structure of the sidebar.

### Details

`SidebarRouteTree` component recursively renders routes which make up the sidebar using route tree and breadcrumbs, so I have added state to keep track of the expanded and collapsed routes, and toggle and expand handlers. 

The main functionality of the sidebar is kept intact:

- Always load the route path and expand children routes when clicking on the route
- Only 1 route group is expanded at a time for better navigation
- After refresh selected route is expanded

Red bg to illustrate the size of the manual toggle. Clicking the route elsewhere functions exactly like before.

Screenshot from 2024-10-12 23-56-42

 ### Open Issues
There have been several issues opened in the past years of users intuitively wanting to manually collapse and expand routes for easier navigation, just recently #7203. Which shows the desire and need for better functionality and navigation with manual collapsing and expanding when searching instead of needing to always load the route in order to expand it.
But from what I have seen, all previous PR's messed up the original logic, which shouldn't be the case. My solution keeps the original functionality while adding the new functionality only to the arrow icon for those who want to use it.

### Testing
I have ran several profiler tests and compared memory usage, load times and cpu usage and couldn't find any noticeable differences with the new additional state. I have also tested memoized version of the of the sidebarroutree but there wasn't any significant difference since the tree of routes is pretty small.

### Conclusion
After playing around and using the new sidebar, it is hard to go back to the previous version. Test it yourself in the preview.

### DIFF
--- a/src/components/Layout/Sidebar/SidebarLink.tsx
+++ b/src/components/Layout/Sidebar/SidebarLink.tsx
@@ -21,6 +21,8 @@ interface SidebarLinkProps {
   isExpanded?: boolean;
   hideArrow?: boolean;
   isPending: boolean;
+  onExpand?: (event: React.MouseEvent<HTMLAnchorElement, MouseEvent>) => void;
+  onToggle?: (event: React.MouseEvent<HTMLSpanElement, MouseEvent>) => void;
 }
 
 export function SidebarLink({
@@ -32,6 +34,8 @@ export function SidebarLink({
   isExpanded,
   hideArrow,
   isPending,
+  onExpand,
+  onToggle,
 }: SidebarLinkProps) {
   const ref = useRef<HTMLAnchorElement>(null);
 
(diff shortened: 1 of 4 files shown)


rammba:
Hi @bauertim, I'm glad to see the progress regarding this feature, great job! 🎉🚀

I've tested a little bit and there are two things that I noticed:
- Sometimes `Quick Start` section is not closed when some other section from `learn react` part is opened.
Quick Start
- It would be great to enable collapse by clicking on whole label, not just on the arrow.

Since I'm not a maintainer, but a translator, I just wanted to share my UX about this. I hope this can help you!
Good luck!

PR #7227, opened 2024-10-15
Update useTocHighlight.tsx

viniciusgiani:
ToC Highlight on Click: A new useEffect is added to listen for clicks on the ToC links. When a link is clicked, the currentIndex is updated to highlight the clicked item.

Handling Few Headings: The updateActiveLink function now ensures that when there are very few headers or the user is near the bottom of the page, the last header is correctly highlighted.

### DIFF
--- a/src/components/Layout/useTocHighlight.tsx
+++ b/src/components/Layout/useTocHighlight.tsx
@@ -32,8 +32,13 @@ export function useTocHighlight() {
       const scrollPosition = window.scrollY + window.innerHeight;
       const headersAnchors = getHeaderAnchors();
 
+      if (headersAnchors.length === 0) {
+        setCurrentIndex(0);
+        return;
+      }
+
       if (scrollPosition >= 0 && pageHeight - scrollPosition <= 0) {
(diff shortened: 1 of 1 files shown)


src/components/Layout/useTocHighlight.tsx
```diff
@@ -76,7 +81,28 @@ export function useTocHighlight() {
...
+    function updateActiveLinkOnClick(event: MouseEvent) {
+      const target = event.target as HTMLAnchorElement;
+      const headersAnchors = getHeaderAnchors();
+      
+      headersAnchors.forEach((anchor, index) => {
+        if (anchor.href === target.href) {
+          setCurrentIndex(index);
+        }
+      });
+    }
```
Parvezkhan0, line 95: The `updateActiveLinkOnClick `function currently assumes that `event.target` is always an `HTMLAnchorElement`. To avoid runtime errors, you should first check if `event.target` is indeed an anchor element before proceeding.

**Suggested Fix:**

`if (target.tagName !== 'A') return;`

PR #7230, opened 2024-10-17
[compiler] Move React 17/18 section to its own subheading

poteto:
Currently it's a little hidden, let's move it to its own subheading for more prominence


poteto:
git can't diff this properly but this was just moving the section out with no edits

PR #7235, opened 2024-10-20
Fix stylesheet precedence example

EricCote:
The `precedence` prop documentation for `<style>` and `<link>` was fixed last May in this pull request:
https://github.com/reactjs/react.dev/pull/6908
(Precedence is now using any name, instead of "high", "medium" or "low")

But the code example accompanying the `<link>` documentation was not fixed.  

There was even a comment in the doc about the fact that precedence didn't seem to work.  -- It did work, it's just that it's more of a grouping behavior than an ordering behavior. So you need at least 3 `<link>` to visibly show that grouping behavior.


gaearon (approved):
omg thank you, this was so confusing

PR #7246, opened 2024-10-22
Add ESLint flat config example, fix ESLint name

karlhorky:
ESLint v8 is end-of-life and ESLint v9 uses flat config by default, so starting with this as a first example seems like the right option (https://x.com/karlhorky/status/1848758371127583015)

Also, fixed the ESLint brand name typos

cc @poteto

### DIFF
--- a/src/content/learn/react-compiler.md
+++ b/src/content/learn/react-compiler.md
@@ -149,7 +149,7 @@ export default [
 ]
 ```
 
-Or, in legacy config:
+Or, in the deprecated eslintrc config format:
 
 ```js
 module.exports = {
--- a/src/content/learn/react-compiler.md
+++ b/src/content/learn/react-compiler.md
@@ -13,7 +13,7 @@ These docs are still a work in progress. More documentation is available in the 
 <YouWillLearn>
 
 * Getting started with the compiler
-* Installing the compiler and eslint plugin
+* Installing the compiler and ESLint plugin
 * Troubleshooting
 
 </YouWillLearn>
@@ -26,7 +26,7 @@ The latest Beta release can be found with the `@beta` tag, and daily experimenta
 
 React Compiler is a new compiler that we've open sourced to get early feedback from the community. It is a build-time only tool that automatically optimizes your React app. It works with plain JavaScript, and understands the [Rules of React](/reference/rules), so you don't need to rewrite any code to use it.
 
-The compiler also includes an [eslint plugin](#installing-eslint-plugin-react-compiler) that surfaces the analysis from the compiler right in your editor. **We strongly recommend everyone use the linter today.** The linter does not require that you have the compiler installed, so you can use it even if you are not ready to try out the compiler.
+The compiler also includes an [ESLint plugin](#installing-eslint-plugin-react-compiler) that surfaces the analysis from the compiler right in your editor. **We strongly recommend everyone use the linter today.** The linter does not require that you have the compiler installed, so you can use it even if you are not ready to try out the compiler.
 
 The compiler is currently released as `beta`, and is available to try out on React 17+ apps and libraries. To install the Beta:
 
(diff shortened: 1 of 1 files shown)


src/content/learn/react-compiler.md
```diff
@@ -126,29 +126,46 @@ In addition to these docs, we recommend checking the [React Compiler Working Gro
...
+Or, in the deprecated eslintrc config format:
 
 ```js
 module.exports = {
   plugins: [
     'eslint-plugin-react-compiler',
   ],
   rules: {
-    'react-compiler/react-compiler': "error",
+    'react-compiler/react-compiler': 'error',
```
karlhorky, line 160: I made the quote style here consistent with the other quotes

src/content/learn/react-compiler.md
```diff
@@ -126,29 +126,46 @@ In addition to these docs, we recommend checking the [React Compiler Working Gro
...
+```js
+import reactCompiler from 'eslint-plugin-react-compiler'
+
+export default [
+  {
+    plugins: {
+      'eslint-plugin-react-compiler': reactCompiler,
+    },
+    rules: {
+      'react-compiler/react-compiler': 'error',
```
AnthonyLzq, line 146: I think this should be:

```js
'eslint-plugin-react-compiler/react-compiler': 'error'
```

If not, you will get the following error:

```sh
[Error] Request textDocument/diagnostic failed.
  Message: Request textDocument/diagnostic failed with message: Key "rules": Key "react-compiler/react-compiler": Could not find plugin "react-compiler".
  Code: -32603 
[Error] Document pull failed for text document file:///.../eslint.config.mjs
  Message: Request textDocument/diagnostic failed with message: Key "rules": Key "react-compiler/react-compiler": Could not find plugin "react-compiler".
  Code: -32603
  ```

Comment (karlhorky) [reply src/content/learn/react-compiler.md:146]: No, that's not how it works - check other plugin docs:

- https://www.npmjs.com/package/eslint-plugin-react-x

Comment (harish-sethuraman) [reply src/content/learn/react-compiler.md:146]: Yea @AnthonyLzq is right here

in flat config you can infer the plugin namespace in the plugins object [ref here](https://eslint.org/docs/latest/use/configure/plugins#configure-plugins)

in the example you shared the plugin namespace is `react-x` thats why rules are referred with `react-x/`

Comment (harish-sethuraman) [inline src/content/learn/react-compiler.md:146]:
```suggestion
export default [
  {
    plugins: {
      'react-compiler': reactCompiler,
    },
    rules: {
      'react-compiler/react-compiler': 'error',
```

karlhorky: sorry @AnthonyLzq @harish-sethuraman I see what you mean now

I fixed it the more conventional way, by using `'react-compiler'` as a key in the `plugins` object:

- 426a8abb04a943295e4a5e403a2bc14d42ceb04c

karlhorky: already done in 426a8abb04a943295e4a5e403a2bc14d42ceb04c, thanks!

harish-sethuraman (approved):
LGTM thanks! would like someone from the compiler team verify before we merge this cc: @poteto @gsathya

poteto:
Sweet, thanks!

src/content/learn/react-compiler.md
```diff
@@ -126,29 +126,46 @@ In addition to these docs, we recommend checking the [React Compiler Working Gro
...
+export default [
+  {
+    plugins: {
+      'react-compiler': reactCompiler,
+    },
+    rules: {
+      'react-compiler/react-compiler': 'error',
+    },
+  },
+]
```
karlhorky, line 149: in case this would be an accepted change, I actually tend to prefer:

1. a separate `configs` variable
2. usage of [the ESLint `Linter.Config[]` TS type](https://github.com/eslint/eslint/blob/d474443109762f3b92811df0411965cf64f595c2/lib/types/index.d.ts#L1402-L1406) with a JSDoc annotation

```suggestion
import reactCompiler from 'eslint-plugin-react-compiler'

/** @type {import('eslint').Linter.Config[]} */
const configs = [
  {
    plugins: {
      'react-compiler': reactCompiler,
    },
    rules: {
      'react-compiler/react-compiler': 'error',
    },
  },
]

export default configs
```

karlhorky: @poteto @harish-sethuraman oh lol too slow (thanks for the merge!)

in case this is a desirable change, I can open a separate PR

poteto: I'd prefer leaving it off since this is more about personal preference. We just want the minimal config in the docs.

Issue #6931, opened 2024-06-05
[Suggestion]: Section "Queueing a Series of State Updates": await in event handler cause rerender


0815Hunter:
### Summary

The documentation says here: "React waits until all code in the event handlers has run before processing your state updates"

This not the case when using `await` inside the handler!

Example
```
const [myNumber, setMyNumber] = useState(0);

async function handleClick(){
    setMyNumber(i => i + 1);
    await delay(100);
    setMyNumber(i => i + 1);
}

function delay(ms) {
    return new Promise(resolve => {
        setTimeout(resolve, ms);
    });
}
```
This will rerender the component twice!

### Page

https://react.dev/learn/queueing-a-series-of-state-updates

### Details

I suggest it should be mentioned as caveat/pitfall that if you call `await` inside an event handler,
all changes made to states before that `await` will then be flushed and trigger a rerender immediately, not waiting
for the event handler to finish completely.

JZZICK:
Can i try this？

JZZICK:
I try to add this 
![Uploading 屏幕截图 2024-10-30 170423.png…]()

JZZICK:
![Uploading 屏幕截图 2024-10-30 170423.png…]()

PR #7268, opened 2024-10-31
Section "Queueing a Series of State Updates": await in event handler …

JZZICK:
…cause immediate rerender


JZZICK:
fixed#6931 
added Important Caveat on Using await in Event Handlers

JZZICK:
#6931

PR #7275, opened 2024-11-03
Update Tailwind dark mode setting from 'class' to 'selector'

iwmywn:
reopen


tailwind.config.js
```diff
@@ -11,7 +11,7 @@ module.exports = {
     './src/pages/**/*.{js,ts,jsx,tsx}',
     './src/styles/**/*.{js,ts,jsx,tsx}',
   ],
-  darkMode: 'class',
+  darkMode: 'selector',
```
harish-sethuraman, line 14: I can see that class was replaced with selector from tailwind 3.4.1 but do we see any bugs / deviations? We were using `class` option in darkMode for these many days?

PR #7277, opened 2024-11-03
[ci] Speed up CI

poteto:
Amends `analyze` and `site_lint` jobs to properly cache installed modules. Also updates deprecated actions.

This speeds up `site_lint` from ~4m 30s to 30s on a cache hit.


alinkedd:
Partially fixes #5961

PR #7296, opened 2024-11-18
[conditional-rendering] Make object data easier for translation

rammba:
While translating [conditional-rendering](https://react.dev/learn/conditional-rendering) page I encountered one small issue. Namely, last (3rd) [challenge solution](https://react.dev/learn/conditional-rendering#refactor-a-series-of---to-if-and-variables) has list of objects in the code which are referenced by its names. Name of the object is not the property of the object, so I needed to make workaround. Since I'm trying to translate all the text, including object property values, I needed to introduce `name` field to the object, similar to the example at the end of [thinking-in-react](http://localhost:3000/learn/thinking-in-react#step-5-add-inverse-data-flow) page:
```
const PRODUCTS = [
  {category: "Fruits", price: "$1", stocked: true, name: "Apple"},
  {category: "Fruits", price: "$1", stocked: true, name: "Dragonfruit"},
  {category: "Fruits", price: "$2", stocked: false, name: "Passionfruit"},
  {category: "Vegetables", price: "$2", stocked: true, name: "Spinach"},
  {category: "Vegetables", price: "$4", stocked: false, name: "Pumpkin"},
  {category: "Vegetables", price: "$1", stocked: true, name: "Peas"}
];
```

I saw that some other languages also have similar workaround:
[French](https://fr.react.dev/learn/conditional-rendering#refactor-a-series-of---to-if-and-variables):
French

[Spanish](https://es.react.dev/learn/conditional-rendering#refactor-a-series-of---to-if-and-variables):
Spanish

I didn't want to to use either english names as in the french solution nor non-english names of variables as in the spanish solution. My idea was to have all functions/variables/etc. in english and all object values on my mother tongue, Serbian.

### DIFF
--- a/src/content/learn/conditional-rendering.md
+++ b/src/content/learn/conditional-rendering.md
@@ -719,24 +719,16 @@ Another solution would be to remove the condition altogether by moving the infor
 <Sandpack>
 
 ```js
-const drinks = {
-  tea: {
-    part: 'leaf',
-    caffeine: '15–70 mg/cup',
-    age: '4,000+ years'
-  },
-  coffee: {
-    part: 'bean',
-    caffeine: '80–185 mg/cup',
-    age: '1,000+ years'
-  }
-};
+const drinks = [
+  {name: 'tea', part: 'leaf', caffeine: '15–70 mg/cup', age: '4,000+ years'},
+  {name: 'coffee', part: 'bean', caffeine: '80–185 mg/cup', age: '1,000+ years'}
+];
 
 function Drink({ name }) {
-  const info = drinks[name];
+  const info = drinks.find(d => d.name === name);
   return (
     <section>
-      <h1>{name}</h1>
+      <h1>{info.name}</h1>
       <dl>
         <dt>Part of plant</dt>
         <dd>{info.part}</dd>


rammba:
Hey @gaearon @rickhanlonii @eps1lon, can you please take a look?

Parvezkhan0 (commented):
The conditional rendering page in React documentation contains object data (like lists of products or drinks), where object names (e.g., "Apple", "Tea") are being used directly within the code. These object names are part of the data, but they are not treated as translatable content.

This becomes a problem when trying to translate the content for different languages.

rammba:
On many places object data are used within the code, but on other places it's possible to translate object content, which is not the case here.

rickhanlonii:
What's wrong with the French language solution, isn't this how you would write it if you were writing it yourself? Using a normalized data structure is better here if possible since you don't need the slower `find` lookup.

rammba:
Hey @rickhanlonii, thanks for clarifying this topic. I didn't think about speed difference between `find()` and direct access by the key. If you agree I can incorporate French's solution here?

rickhanlonii:
Yeah please do, and that means we can close this?

rammba:
Hey @rickhanlonii, I've incorporated French's solution here.

rickhanlonii:
Ok sorry, I said the French version but I meant the Spanish version. For native speakers, wouldn't you want to write:

```js
<Drink name="name-in-your-language" />
```

Not

```js
<Drink name="tea" />
```

Since `tea` is in english? Is there a blocker to this do to characters or something?

rammba:
As I said in the description: `I didn't want to to use non-english names of variables as in the spanish solution`. For me it looks ugly to name variables with non-ASCII characters, while object data is totally correct to have those characters. But, if you prefer that way, I will do that.

Issue #7252, opened 2024-10-24
[Typo]: Code example mistake in useMemo page


khkh0109:
### Summary

There is a mistake in the ChatRoom example code on the useMemo page, specifically in the 'Preventing an Effect from firing too often' section.

### Page

https://react.dev/reference/react/useMemo#preventing-an-effect-from-firing-too-often

### Details

There are two mistakes in the ChatRoom example under the 'Preventing an Effect from firing too often' section:

1. The line `const options = createOptions()` suddenly appears in the code without context. It seems like this should be removed.
2. In the comment 'Only changes when createOptions changes', the reference to 'createOptions' should be replaced with 'options object'. This is because options is the actual dependency in the useEffect.

Screenshot of typo:
<img width="923" alt="스크린샷 2024-10-24 오후 7 28 57" src="https://github.com/user-attachments/assets/ee17d564-c645-4e9d-be2e-d12bc3f83738">

JZZICK:
Can i do this？

JZZICK:
Can i deal with this？

JZZICK:
#7261

JZZICK:
[https://github.com/reactjs/react.dev/pull/7261#issue-2621809632](url)

JZZICK:
#7261 has fixed

JZZICK:
#7261

PR #7319, opened 2024-12-03
Code example mistake in useMemo page

JZZICK:

### DIFF
--- a/src/content/reference/react/useMemo.md
+++ b/src/content/reference/react/useMemo.md
@@ -1105,7 +1105,7 @@ function ChatRoom({ roomId }) {
     const connection = createConnection(options);
     connection.connect();
     return () => connection.disconnect();
-  }, [options]); // ✅ Only changes when createOptions changes
+  }, [options]); // ✅ Only changes when create options changes
   // ...
 ```
 


JZZICK:
#7252

src/content/reference/react/useMemo.md
```diff
@@ -1105,7 +1105,7 @@ function ChatRoom({ roomId }) {
     const connection = createConnection(options);
     connection.connect();
     return () => connection.disconnect();
-  }, [options]); // ✅ Only changes when createOptions changes
+  }, [options]); // ✅ Only changes when create options changes
```
ZhenyaGoroh, line 1108: Maybe it should be "Only changes when options changes"? What is create options?

lukahartwig: Additionally, line 1104 `const options = createOptions();` should be removed.

JZZICK: ok，thankyou

JZZICK: Please check #7335
@ZhenyaGoroh @lukahartwig 
Thank you all

ZhenyaGoroh: > Please check #7335 @ZhenyaGoroh @lukahartwig Thank you all

I left the comment. Also I'd suggest to commit changed to this pr instead of creating a new one.

PR #7320, opened 2024-12-05
[19] Add docs for prerender APIs

rickhanlonii:


src/content/reference/react-dom/static/prerender.md
```diff
@@ -0,0 +1,298 @@
...
+  * **optional** `identifierPrefix`: A string prefix React uses for IDs generated by [`useId`.](/reference/react/useId) Useful to avoid conflicts when using multiple roots on the same page. Must be the same prefix as passed to [`hydrateRoot`.](/reference/react-dom/client/hydrateRoot#parameters)
+  * **optional** `namespaceURI`: A string with the root [namespace URI](https://developer.mozilla.org/en-US/docs/Web/API/Document/createElementNS#important_namespace_uris) for the stream. Defaults to regular HTML. Pass `'http://www.w3.org/2000/svg'` for SVG or `'http://www.w3.org/1998/Math/MathML'` for MathML.
+  * **optional** `onError`: A callback that fires whenever there is a server error, whether [recoverable](#recovering-from-errors-outside-the-shell) or [not.](#recovering-from-errors-inside-the-shell) By default, this only calls `console.error`. If you override it to [log crash reports,](#logging-crashes-on-the-server) make sure that you still call `console.error`. You can also use it to [adjust the status code](#setting-the-status-code) before the shell is emitted.
```
gnoff, line 58: maybe call this an error during prerendering since it's not a server per se

src/content/reference/react-dom/static/prerender.md
```diff
@@ -0,0 +1,298 @@
...
+  let content = '';
+  while (true) {
+    const {done, value} = await reader.read();
+    if (done) {
+      return content;
+    }
+    content += Buffer.from(value).toString('utf8');
+  }
+}
+```
```
gnoff, line 243: seems like you are mixing node and non node in this example.

```suggestion
async function renderToString() {
  const {prelude} = await prerender(<App />, {
    bootstrapScripts: ['/main.js']
  });
  
  const reader = stream.getReader();
  const decoder = new TextDecoder()
  let content = '';
  while (true) {
    const {done, value} = await reader.read();
    if (done) {
      content += decoder.decode()
      return content;
    }
    content += decoder.decode(value, { stream: true });
  }
}
```

gnoff (approved):

PR #7329, opened 2024-12-06
feat: react 19 ftw

harish-sethuraman:
- upgraded react to 19 for site
- upgraded next 15
- upgraded mdx to 3
- updated remark and rehype plugins

update: works in local properly

build fails due to 

```
./node_modules/next/dist/server/lib/router-utils/filesystem.d.ts:65:72
Type error: '>' expected.

  63 |     dynamicRoutes: FilesystemDynamicRoute[];
  64 |     nextDataRoutes: Set<string>;
> 65 |     exportPathMapRoutes: undefined | ReturnType<typeof buildCustomRoute<Rewrite>>[];
     |                                                                        ^
  66 |     devVirtualFsItems: Set<string>;
  67 |     prerenderManifest: PrerenderManifest;
  68 |     middlewareMatcher: MiddlewareRouteMatch | undefined;
```


src/utils/compileMDX.ts
```diff
@@ -60,13 +60,15 @@ export default async function compileMDX(
       .join('\n');
 
   // Turn the MDX we just read into some JS we can execute.
-  const {remarkPlugins} = require('../../plugins/markdownToHtml');
+  // const {remarkPlugins} = require('../../plugins/markdownToHtml');
```
harish-sethuraman, line 63: check why we aren't able to use remark / rehype plugins after upgrading it.

getting 

```
Error: Expected usable value but received an empty preset, which is probably a mistake: presets typically come with `plugins` and sometimes with `settings`, but this has neither
```

src/utils/compileMDX.ts
```diff
@@ -36,13 +36,13 @@ export default async function compileMDX(
...
-    );
-    return cached;
-  }
+  // const cached = await store.get(hash);
+  // if (cached) {
+  //   console.log(
+  //     'Reading compiled MDX for /' + path + ' from ./node_modules/.cache/'
+  //   );
+  //   return cached;
+  // }
```
harish-sethuraman, line 45: for now removed caching re enable before we push it to main

src/pages/[[...markdownPath]].js
```diff
@@ -89,12 +89,15 @@ function reviveNodeOnClient(key, val) {
...
-      $$typeof: Symbol.for('react.element'),
+      $$typeof: Symbol.for('react.transitional.element'),
       type: type,
       key: key,
       ref: null,
       props: props,
       _owner: null,
+      _store: {
+        validate: false,
+      },
```
Comment (harish-sethuraman) [inline src/pages/[[...markdownPath]].js]: https://github.com/facebook/react/blob/c56c6234328a29930487295afe61597db48f058c/packages/react/src/ReactChildren.js#L242-L250

Im unaware of what this piece of code does so Im currently adding _store to be `null`. Should I update react to soft check if `_store` is present or not?

src/pages/[[...markdownPath]].js
```diff
@@ -89,12 +89,13 @@ function reviveNodeOnClient(key, val) {
...
     }
     return {
-      $$typeof: Symbol.for('react.element'),
+      $$typeof: Symbol.for('react.transitional.element'),
       type: type,
       key: key,
       ref: null,
       props: props,
       _owner: null,
+      _store: null,
```
Comment (harish-sethuraman) [inline src/pages/[[...markdownPath]].js:98]: https://github.com/facebook/react/blob/c56c6234328a29930487295afe61597db48f058c/packages/react/src/ReactChildren.js#L245

Im unsure what this is used for and also should I update react to soft check for _store?

PR #7332, opened 2024-12-06
update useRef documentation to use ref prop instead of forwardRef

joshuaslate:
I noticed a couple of spots in the documentation that were suggesting using `forwardRef`. I updated them to use the `ref` prop instead.

### DIFF
--- a/src/content/learn/manipulating-the-dom-with-refs.md
+++ b/src/content/learn/manipulating-the-dom-with-refs.md
@@ -360,7 +360,7 @@ export default function MyForm() {
   const inputRef = useRef(null);
 
   function handleClick() {
-    inputRef.current?.focus();
+    inputRef.current.focus();
   }
 
   return (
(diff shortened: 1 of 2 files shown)


gaearon (changes requested):
there are a lot of unrelated whitespace changes

src/content/learn/manipulating-the-dom-with-refs.md
```diff
@@ -376,40 +376,31 @@ export default function MyForm() {
 
 </Sandpack>
 
-To help you notice the issue, React also prints an error to the console:
-
-<ConsoleBlock level="error">
```
gaearon, line 381: and here instead you'll describe the crash

src/content/learn/manipulating-the-dom-with-refs.md
```diff
@@ -376,40 +376,31 @@ export default function MyForm() {
...
 Now clicking the button to focus the input works:
 
 <Sandpack>
 
 ```js
-import { forwardRef, useRef } from 'react';
+import { useRef } from 'react';
 
-const MyInput = forwardRef((props, ref) => {
+function MyInput({ ref, ...props }) {
```
gaearon, line 401: these props are not being used so you can drop them

src/content/learn/manipulating-the-dom-with-refs.md
```diff
@@ -442,13 +433,9 @@ In the above example, `MyInput` exposes the original DOM input element. This let
...
 ```js
-import {
-  forwardRef, 
-  useRef, 
-  useImperativeHandle
-} from 'react';
+import { useRef, useImperativeHandle } from 'react';
 
-const MyInput = forwardRef((props, ref) => {
+function MyInput({ ref, ...props }) {
```
gaearon, line 438: same, these are not used

gaearon (changes requested):
let's remove the optional chaining, use the crash _to_ explain the problem, and remove the unnecessary `...props` thing

src/content/learn/manipulating-the-dom-with-refs.md
joshuaslate: I just added a brief explanation of what causes the crash, because trying the example will show the console output. If you'd like, I can add back the `<ConsoleBlock />` that shows the console output anyway:

```
Cannot read properties of null (reading 'focus')
```

gaearon: yup let's add that

gaearon:
sorry for the churn, we ended up merging https://github.com/reactjs/react.dev/pull/7364, seems like maybe this overlaps fully and can be closed?

PR #7335, opened 2024-12-07
Code example mistake in useMemo page

JZZICK:


ZhenyaGoroh:
Yeah, I think this one is a way better

ZhenyaGoroh:
Also I found, that useEffect _fires_, not changes, what do u this on it?

ZhenyaGoroh:
_Changes_ is used for memo as changing roomId recalculates the options value, but changing options doesn't trigger the useEffect _change_, its change fires the effect.

Issue #7008, opened 2024-07-02
[Suggestion]: Adding a Scroll-To-top Button .


AnkitLuhar:
### Summary

Please add a "Scroll to Top" button to improve user experience. The button should appear after scrolling down and allow users to quickly return to the top of the page.

### Details

I would like to request the addition of a "Scroll to Top" button to the website. This feature would enhance the user experience by allowing users to quickly navigate back to the top of the page without needing to manually scroll up, especially on longer pages.

-->Expected Behavior:

1. A button should appear at the bottom right (or another appropriate location) of the screen after the user has scrolled down a certain distance.

2.Clicking the button should smoothly scroll the user back to the top of the page.

3.The button should have a user-friendly design that fits well with the website's overall theme.

-->Benefits:

1.Improves user navigation and accessibility.
2.Enhances overall user experience, especially for pages with extensive content.

-->Additional Context:

This feature is commonly seen on many modern websites and is appreciated by users for its convenience.

JZZICK:
i want to try this

PR #7339, opened 2024-12-07
"Adding a Scroll-To-top Button

JZZICK:


JZZICK:
Adding a Scroll-To-top Button #7008

JZZICK:
Adding a Scroll-To-top Button #7008

PR #7346, opened 2024-12-10
fix: align counter example code with Effect dependencies documentation

Windyskr:
The original code didn't work properly - the counter only incremented once and then stopped. It failed to demonstrate the tutorial's point about dependency warnings.
After changes, the counter now increments every second, but remains unresponsive to button clicks, which suppressing the dependency linter so dangerous.

### DIFF
--- a/src/content/learn/removing-effect-dependencies.md
+++ b/src/content/learn/removing-effect-dependencies.md
@@ -311,7 +311,7 @@ export default function Timer() {
   const [increment, setIncrement] = useState(1);
 
   function onTick() {
-	setCount(count + increment);
+	setCount(count => count + increment);
   }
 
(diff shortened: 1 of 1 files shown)


Windyskr:
Thank you!
Thank you for taking the time to register as an open source contributor. You may print the submission below for your records. You can resubmit this form to update your name or email. If you would like to make any other changes, please email [cla@meta.com](mailto:cla@meta.com) with your details.

We will now be able to accept any pull requests that you make to projects on our [GitHub account](https://github.com/facebook). If you already submitted a pull request before signing the CLA, then we will add a confirmation comment on that pull request (this might take several minutes). If you haven't submitted a pull request yet, then we look forward to your future contributions!

PR #7364, opened 2024-12-13
Remove forwardRef references from useRef and Manipulating the DOM with Refs pages

mattcarrollcode:
Update the "useRef" and "Manipulating the DOM with Refs" pages to no longer use forwardRef since its no longer needed in React 19

### DIFF
--- a/src/content/learn/manipulating-the-dom-with-refs.md
+++ b/src/content/learn/manipulating-the-dom-with-refs.md
@@ -382,7 +382,7 @@ export default function MyForm() {
 
 #### Exposing a subset of the API with an imperative handle {/*exposing-a-subset-of-the-api-with-an-imperative-handle*/}
 
-In the above example, `MyInput` ref props is passed on to the original DOM input element. This lets the parent component call `focus()` on it. However, this also lets the parent component do something else--for example, change its CSS styles. In uncommon cases, you may want to restrict the exposed functionality. You can do that with [`useImperativeHandle`](/reference/react/useImperativeHandle):
+In the above example, the ref passed to `MyInput` is passed on to the original DOM input element. This lets the parent component call `focus()` on it. However, this also lets the parent component do something else--for example, change its CSS styles. In uncommon cases, you may want to restrict the exposed functionality. You can do that with [`useImperativeHandle`](/reference/react/useImperativeHandle):
 
 <Sandpack>
 
--- a/src/content/learn/manipulating-the-dom-with-refs.md
+++ b/src/content/learn/manipulating-the-dom-with-refs.md
@@ -347,15 +347,32 @@ Read more about [how this helps find bugs](/reference/react/StrictMode#fixing-bu
 Refs are an escape hatch that should be used sparingly. Manually manipulating _another_ component's DOM nodes makes your code even more fragile.
 </Pitfall>
 
-When you put a ref on a built-in component that outputs a browser element like `<input />`, React will set that ref's `current` property to the corresponding DOM node (such as the actual `<input />` in the browser).
+You can pass refs from parent component to child components [just like any other prop](/learn/passing-props-to-a-component).
+
+```js {3-4,9}
+import { useRef } from 'react';
+
+function MyInput({ref}) {
+  return <input ref={ref} />;
+}
+
+function MyForm() {
+  const inputRef = useRef(null);
+  return <><MyInput ref={inputRef} /></>
+}
+```
+
+In the above example, a ref is created in the parent component, `MyForm`, and is passed to the child component, `MyInput`. `MyInput` then passes the ref to `<input>`. Because `<input>` is a [built-in component](/reference/react-dom/components/common) React sets the `.current` property of the ref to the `<input>` DOM element.
+
+The `inputRef` created in `MyForm` now points to the `<input>` DOM element returned by `MyInput`. A click handler created in `MyForm` can access `inputRef` and call `focus()` to set the focus on `<input>`.
 
 <Sandpack>
 
 ```js
 import { useRef } from 'react';
 
-function MyInput(props) {
-  return <input {...props} />;
+function MyInput({ref}) {
+  return <input ref={ref} />;
 }
 
 export default function MyForm() {
@@ -530,7 +547,7 @@ export default function TodoList() {
     const newTodo = { id: nextId++, text: text };
     flushSync(() => {
       setText('');
-      setTodos([ ...todos, newTodo]);      
+      setTodos([ ...todos, newTodo]);
     });
     listRef.current.lastChild.scrollIntoView({
       behavior: 'smooth',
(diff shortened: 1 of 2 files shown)


src/content/learn/manipulating-the-dom-with-refs.md
```diff
@@ -343,9 +343,11 @@ Read more about [how this helps find bugs](/reference/react/StrictMode#fixing-bu
 
 ## Accessing another component's DOM nodes {/*accessing-another-components-dom-nodes*/}
 
-When you put a ref on a built-in component that outputs a browser element like `<input />`, React will set that ref's `current` property to the corresponding DOM node (such as the actual `<input />` in the browser).
+<Pitfall>
+Refs are an escape hatch that should be used sparingly. Manually manipulating _another_ component's DOM nodes makes your code even more fragile.
+</Pitfall>
 
-However, if you try to put a ref on **your own** component, like `<MyInput />`, by default you will get `null`. Here is an example demonstrating it. Notice how clicking the button **does not** focus the input:
+When you put a ref on a built-in component that outputs a browser element like `<input />`, React will set that ref's `current` property to the corresponding DOM node (such as the actual `<input />` in the browser).
```
gaearon, line 350: This is confusing. You need to walk through these stages:

- Passing a ref to `<input>`
- Creating your own component — it doesn't work anymore
- Fixing it by spreading props

Spreading props isn't a default thing people would reach for.

gaearon: Actually it might be better to remove spreading from examples below and just do `ref={ref}` so that it's obvious it works like any other prop

src/content/reference/react/useRef.md
```diff
@@ -448,16 +448,16 @@ button { display: block; margin-bottom: 20px; }
...
+Sometimes, you may want to let the parent component manipulate the DOM inside of your component. For example, maybe you're writing a `MyInput` component, but you want the parent to be able to focus the input (which the parent has no access to). You can create a `ref` in the parent and pass the `ref` as prop to the child component. Read a [detailed walkthrough](/learn/manipulating-the-dom-with-refs#accessing-another-components-dom-nodes) here.
 
 <Sandpack>
 
 ```js
-import { forwardRef, useRef } from 'react';
+import { useRef } from 'react';
 
-const MyInput = forwardRef((props, ref) => {
+const MyInput = (props, ref) => {
```
gaearon, line 458: Ref is a prop now, so it won't be a second argument. I'd suggest to directly pass it as a prop in this example and remove spreading to make the pattern very clear

gaearon (changes requested):
see above:

- let's fix the examples to show passing `ref` directly without spreading
- fix code example that has second argument, it's wrong
- keep the troubleshooting section but reword to match the new problem

src/content/reference/react/useRef.md
```diff
@@ -448,16 +448,16 @@ button { display: block; margin-bottom: 20px; }
...
 <Sandpack>
 
 ```js
-import { forwardRef, useRef } from 'react';
+import { useRef } from 'react';
 
-const MyInput = forwardRef((props, ref) => {
-  return <input {...props} ref={ref} />;
-});
+const MyInput = ({ref}) => {
```
gaearon, line 458: nit: we generally put spaces around destructuring like `({ ref })`. the rest of the docs also uses a function declaration style, like `function Input({ ref })`, no need for an arrow here

src/content/learn/manipulating-the-dom-with-refs.md
```diff
@@ -343,75 +343,39 @@ Read more about [how this helps find bugs](/reference/react/StrictMode#fixing-bu
...
 
-<Sandpack>
-
-```js
+```js {3-4,9}
 import { useRef } from 'react';
 
-function MyInput(props) {
-  return <input {...props} />;
+function MyInput({ref}) {
```
gaearon, line 355: same nit

src/content/learn/manipulating-the-dom-with-refs.md
```diff
@@ -343,75 +343,39 @@ Read more about [how this helps find bugs](/reference/react/StrictMode#fixing-bu
 
 ## Accessing another component's DOM nodes {/*accessing-another-components-dom-nodes*/}
 
-When you put a ref on a built-in component that outputs a browser element like `<input />`, React will set that ref's `current` property to the corresponding DOM node (such as the actual `<input />` in the browser).
+<Pitfall>
+Refs are an escape hatch that should be used sparingly. Manually manipulating _another_ component's DOM nodes makes your code even more fragile.
```
gaearon, line 347: so i'm wondering if this is a bit too strong here... we're describing legit cases on this page (like focusing) but this makes it sound like even the stuff below is fragile

src/content/learn/manipulating-the-dom-with-refs.md
```diff
@@ -343,75 +343,39 @@ Read more about [how this helps find bugs](/reference/react/StrictMode#fixing-bu
...
-
-  return (
-    <>
-      <MyInput ref={inputRef} />
-      <button onClick={handleClick}>
-        Focus the input
-      </button>
-    </>
-  );
+  return <><MyInput ref={inputRef} /></>
```
gaearon, line 361: this `<>` fragment isn't doing anything here

src/content/learn/manipulating-the-dom-with-refs.md
```diff
@@ -343,75 +343,39 @@ Read more about [how this helps find bugs](/reference/react/StrictMode#fixing-bu
...
 <Sandpack>
 
 ```js
-import { forwardRef, useRef } from 'react';
+import { useRef } from 'react';
 
-const MyInput = forwardRef((props, ref) => {
-  return <input {...props} ref={ref} />;
-});
+function MyInput({ref}) {
```
gaearon, line 374: nit: spacing

src/content/reference/react/useRef.md
```diff
@@ -573,20 +573,18 @@ export default function MyInput({ value, onChange }) {
...
 
-And then wrap it in [`forwardRef`](/reference/react/forwardRef) like this:
-
-```js {3,8}
-import { forwardRef } from 'react';
+And then add `ref` to the list of props your component accepts and pass `ref` as a prop to the relevent child [built-in component](/reference/react-dom/components/common) like this:
 
-const MyInput = forwardRef(({ value, onChange }, ref) => {
+```js {1,6}
+const MyInput = ({ value, onChange, ref}) => {
```
gaearon, line 579: this should remain a function declaration, no need to change it to an arrow

gaearon (approved):
looks good if nits are addressed. please have a look at https://github.com/reactjs/react.dev/pull/7332 which was aiming to do the same — maybe something can be synthesized if you like any of those edits, these PRs seem similar in goals

src/content/learn/manipulating-the-dom-with-refs.md
mattcarrollcode: Softened the wording here.

src/content/learn/manipulating-the-dom-with-refs.md
mattcarrollcode: Fixed

src/content/reference/react/useRef.md
mattcarrollcode: Fixed

PR #7365, opened 2024-12-14
Fix startTransition troubleshooting links

Mamaduka:
The `startTransition` API reference does not include a troubleshooting section. Update links and point them to `useTransition`.


Mamaduka:
Closing in favor of #7367.

PR #7367, opened 2024-12-15
Fix troubleshooting links in startTransition reference

jimmycallin:
Noticed some broken links in startTransition that I updated to link to the equivalent section in useTransition.


Mamaduka:
It looks like we both noticed the same issue, though I've missed one link :)

I'm going to close my PR in favor of yours.

PR #7369, opened 2024-12-16
docs[compiler]: clarify React DevTools support for Compiler Badges in React Native

piotrski:
Updated docs to note that React Native supports DevTools Compiler Badges from version 0.75.0, due to its custom backend.

### DIFF
--- a/src/content/learn/react-compiler.md
+++ b/src/content/learn/react-compiler.md
@@ -347,7 +347,7 @@ React Compiler can verify many of the Rules of React statically, and will safely
 
 ### How do I know my components have been optimized? {/*how-do-i-know-my-components-have-been-optimized*/}
 
-[React DevTools](/learn/react-developer-tools) (v5.0+) and [React Native DevTools](https://reactnative.dev/docs/react-native-devtools) has built-in support for React Compiler and will display a "Memo ✨" badge next to components that have been optimized by the compiler.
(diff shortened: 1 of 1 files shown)


src/content/learn/react-compiler.md
```diff
@@ -347,7 +347,7 @@ React Compiler can verify many of the Rules of React statically, and will safely
 
 ### How do I know my components have been optimized? {/*how-do-i-know-my-components-have-been-optimized*/}
 
-[React Devtools](/learn/react-developer-tools) (v5.0+) has built-in support for React Compiler and will display a "Memo ✨" badge next to components that have been optimized by the compiler.
+[React Devtools](/learn/react-developer-tools) (v5.0+) has built-in support for React Compiler and will display a "Memo ✨" badge next to components that have been optimized by the compiler. In React Native, this feature is available from version 0.75.0 onwards, since React Native bundles its own DevTools backend.
```
gaearon, line 350: i think we should remove the "5.0" thing and instead refer to the new React Native DevTools (default since 0.76)

piotrski:
@gaearon I think it would be better to keep the info about React Native version, just like we have info about the DevTools version, since there are still many projects using <0.75. It might mislead some as to why React DevTools v5.0+ badges are not showing up despite being available.  At least, that was my idea, since I had to check the changelog to find this :)

gaearon:
"React Native DevTools" is a new brand name and it links to a new tool. Maybe it's not ideal but let's see if people get confused?

PR #7372, opened 2024-12-17
Add hanging ref troubleshooting section

mattcarrollcode:

### DIFF
--- a/src/content/reference/react-dom/components/common.md
+++ b/src/content/reference/react-dom/components/common.md
@@ -282,6 +282,90 @@ To support backwards compatibility, if a cleanup function is not returned from t
 * When Strict Mode is on, React will **run one extra development-only setup+cleanup cycle** before the first real setup. This is a stress-test that ensures that your cleanup logic "mirrors" your setup logic and that it stops or undoes whatever the setup is doing. If this causes a problem, implement the cleanup function.
 * When you pass a *different* `ref` callback, React will call the *previous* callback's cleanup function if provided. If no cleanup function is defined, the `ref` callback will be called with `null` as the argument. The *next* function will be called with the DOM node.
 
+#### Troubleshooting: My ref points to a unmounted DOM node {/*my-ref-points-to-a-unmounted-dom-node*/}
+
+A `ref` callback function with a cleanup function that does not set `ref.current` to `null` can result in a `ref` to a unmounted node. Uncheck "Show Input" below and click "Submit" to see how the `ref` to the unmounted `<input>` is still accessible by the click handler for the form.
+
+<Sandpack>
+
+```js
+import { useRef, useState } from "react";
+
+export default function MyForm() {
+  const [showInput, setShowInput] = useState(true);
+  let inputRef = useRef();
+  const handleCheckboxChange = (event) => {
+    setShowInput(event.target.checked);
+  };
+  const handleSubmit = (event) => {
+    event.preventDefault();
+    if (inputRef.current) {
+      alert(`Input value is: "${inputRef.current.value}"`);
+    } else {
+      alert("no input");
+    }
+  };
+  const inputRefCallback = (node) => {
+    inputRef.current = node;
+    return () => {
+      // ⚠️ You must set `ref.current` to `null`
+      // in this cleanup function e.g.
+      // `inputRef.current = null;`
+      // to prevent hanging refs to unmounted DOM nodes
+    };
+  };
+
+  return (
+    <form onSubmit={handleSubmit}>
+      <div>
+        <label>
+          <input
+            type="checkbox"
+            checked={showInput}
+            onChange={handleCheckboxChange}
+          />
+          Show Input
+        </label>
+      </div>
+      {showInput && (
+        <div>
+          <label>
+            Input:
+            <input
+              type="text"
+              defaultValue="value from input DOM node"
+              ref={inputRefCallback}
+            />
+          </label>
+        </div>
+      )}
+      <button type="submit">Submit</button>
+    </form>
+  );
+}
+```
+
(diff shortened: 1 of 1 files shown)


src/content/reference/react-dom/components/common.md
```diff
@@ -282,6 +282,90 @@ To support backwards compatibility, if a cleanup function is not returned from t
...
+
+To fix the hanging ref to the DOM node that is no longer rendered, set `ref.current` to `null` in the `ref` callback cleanup function.
+
+```js
+import { useRef } from "react";
+
+function MyInput() {
+  inputRef = useRef()
+  const inputRefCallback = (node) => {
+    ref.current = node;
```
gaearon, line 358: this should be inputRef i think

src/content/reference/react-dom/components/common.md
```diff
@@ -282,6 +282,90 @@ To support backwards compatibility, if a cleanup function is not returned from t
...
+
+</Sandpack>
+
+To fix the hanging ref to the DOM node that is no longer rendered, set `ref.current` to `null` in the `ref` callback cleanup function.
+
+```js
+import { useRef } from "react";
+
+function MyInput() {
+  inputRef = useRef()
```
gaearon, line 356: this should have `const`

src/content/reference/react-dom/components/common.md
```diff
@@ -282,6 +282,90 @@ To support backwards compatibility, if a cleanup function is not returned from t
...
+A `ref` callback function with a cleanup function that does not set `ref.current` to `null` can result in a `ref` to a unmounted node. Uncheck "Show Input" below and click "Submit" to see how the `ref` to the unmounted `<input>` is still accessible by the click handler for the form.
+
+<Sandpack>
+
+```js
+import { useRef, useState } from "react";
+
+export default function MyForm() {
+  const [showInput, setShowInput] = useState(true);
+  let inputRef = useRef();
```
gaearon, line 296: why `let` here? other declarations are `const`

src/content/reference/react-dom/components/common.md
```diff
@@ -282,6 +282,90 @@ To support backwards compatibility, if a cleanup function is not returned from t
 * When Strict Mode is on, React will **run one extra development-only setup+cleanup cycle** before the first real setup. This is a stress-test that ensures that your cleanup logic "mirrors" your setup logic and that it stops or undoes whatever the setup is doing. If this causes a problem, implement the cleanup function.
 * When you pass a *different* `ref` callback, React will call the *previous* callback's cleanup function if provided. If no cleanup function is defined, the `ref` callback will be called with `null` as the argument. The *next* function will be called with the DOM node.
 
+#### Troubleshooting: My ref points to a unmounted DOM node {/*my-ref-points-to-a-unmounted-dom-node*/}
```
gaearon, line 285: generally Troubleshooting is a top-level section that goes after Reference _and_ after Usage on every other page. i'm hesitant to add a one-off troubleshooting in the middle of a Reference. i'm also not convinced this specific problem is super common or super necessary to document in detail since it doesn't actually break things. 

i'd actually suggest killing this section and instead adding a Usage recipe to `useRef` page that describes how to _set a ref and also do something else_ — like Chance's original example of _what he was actually trying to do_. and show the correct example there.

gaearon (changes requested):
see the second part of https://github.com/reactjs/react.dev/pull/7372#discussion_r1888285532. don't add troubleshooting just yet, instead add a recipe to `useRef` usage docs

rickhanlonii:
+1 to @gaearon's feedback, just to add some more context: we have a usage example in the [Learn docs](https://react-pb6teic0q-fbopensource.vercel.app/learn/manipulating-the-dom-with-refs#how-to-manage-a-list-of-refs-using-a-ref-callback) and we also explain how [Strict Mode helps catch this bug](https://react-pb6teic0q-fbopensource.vercel.app/reference/react/StrictMode#fixing-bugs-found-by-re-running-ref-callbacks-in-development), but having it on this page too would be useful. 

In that thread I was actually saying we should document the behavior where callback refs unmount and remount when the callback function reference changes (so you should wrap it in useCallback) which isn't mentioned anywhere in the docs. Here's a sandbox showing [the difference with/without useCallback](https://codesandbox.io/p/sandbox/react-dev-forked-lgyvf6?file=%2Fsrc%2FApp.js%3A3%2C29).

src/content/reference/react-dom/components/common.md
mattcarrollcode: Sounds good to me. I didn't do this initially because the docs for a ref callback are here and not in useRef.

Do you think moving the "`ref` callback function" section from `/reference/react-dom/components/common.md` to  the useRef page make sense too? or just this "troubleshooting" content?

mattcarrollcode:
> In that thread I was actually saying we should document the behavior where callback refs unmount and remount when the callback function reference changes (so you should wrap it in useCallback) which isn't mentioned anywhere in the docs. Here's a sandbox showing [the difference with/without useCallback](https://codesandbox.io/p/sandbox/react-dev-forked-lgyvf6?file=%2Fsrc%2FApp.js%3A3%2C29).

Thanks for the CodeSandbox! I did see that and was planning on adding a mention of this as well but I wrote this content to better understand the problem space for myself.  I'll add this in this PR or a followup.

mattcarrollcode (commented):
Rewrote this to be a usage example in useRef for detecting DOM changes

src/content/reference/react-dom/components/common.md
mattcarrollcode: Rewrote this PR to be a usage example in useRef for detecting DOM changes

PR #7375, opened 2024-12-19
change the way that merge state within useOptimistic demo

felismargarita:
This code change aims to help developers better understand the usage of useOptimistic that it's important to set a correct merge state logic to avoid incorrect mid-state rendering.

Here is a screenshot that the original code does not handle the merge logic properly that makes a same message displayed twice:

image

#7379


karthik-awebon:
Hello @felismargarita the existing code effectively demonstrates the use of the UseOptimistic Hook with a simple example. Adding unnecessary complexity to handle duplicate messages detracts from the core purpose of the demo. 

Additionally, your PR does not function as intended. The use of time as a unique identifier for messages is flawed, as identical messages can still appear multiple times if their timestamps differ, as evidenced in the screenshot.
image

PR #7411, opened 2025-01-08
Update react-19-upgrade-guide.md

amirsaeed671:
The current link for `18.3.0` release points to the release page but did not take the user to the exact `18.3.0` release section. this change will redirect user to the exact section of `18.3.0`


rickhanlonii (approved):
sgtm

PR #7414, opened 2025-01-08
fix: Fix reference error in second code snippet under Server Functions with Actions

hyundonmoon:
The second code snippet under Server Functions with Actions references `state.error` when `state` doesn't exist.

Fixed to reference `error` instead.

```javascript
function UpdateName() {
  const [name, setName] = useState('');
  const [error, setError] = useState(null);

  const submitAction = async () => {
    startTransition(async () => {
      const {error} = await updateName(name);
      if (!error) {
        setError(error);
      } else {
        setName('');
      }
    })
  }
  
  return (
    <form action={submitAction}>
      <input type="text" name="name" disabled={isPending}/>
      {state.error && <span>Failed: {state.error}</span>} // state doesn't exist
    </form>
  )
}
```

### DIFF
--- a/src/content/reference/rsc/server-functions.md
+++ b/src/content/reference/rsc/server-functions.md
@@ -137,7 +137,7 @@ function UpdateName() {
   return (
     <form action={submitAction}>
       <input type="text" name="name" disabled={isPending}/>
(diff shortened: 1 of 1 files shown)


rickhanlonii (approved):
nice catch

PR #7415, opened 2025-01-08
Only show scrollbar when necessary in TerminalBlock component

hichemfantar:
Adjust the overflow property from `overflow-x-scroll` to `overflow-x-auto` in the TerminalBlock component for improved layout handling.

before
image

after
image


harish-sethuraman (approved):
Thanks!

PR #7416, opened 2025-01-08
Add quick setup section for Vite in documentation

### LINKED ISSUE #5797: List Vite at the same level as SSR frameworks as a way to start a new React project

hichemfantar:
Introduce a new section in the React project setup guide to provide a quick start for using Vite as a build tool. This addition aims to enhance the onboarding experience for developers looking to create projects with Vite.

Description is an abridged version of the overview section from the vite [site](https://vite.dev/guide/#overview).

closes #5797

image

### DIFF
--- a/src/content/learn/start-a-new-react-project.md
+++ b/src/content/learn/start-a-new-react-project.md
@@ -8,7 +8,6 @@ If you want to build a new app or a new website fully with React, we recommend p
 
 </Intro>
 
-
 You can use React without a framework, however we’ve found that most apps and sites eventually build solutions to common problems such as code-splitting, routing, data fetching, and generating HTML. These problems are common to all UI libraries, not just React.
 
 By starting with a framework, you can get started with React quickly, and avoid essentially building your own framework later.
@@ -30,6 +29,16 @@ Even if you don't need routing or data fetching at first, you'll likely want to 
 If your app has unusual constraints not served well by these frameworks, or you prefer to solve these problems yourself, you can roll your own custom setup with React. Grab `react` and `react-dom` from npm, set up your custom build process with a bundler like [Vite](https://vitejs.dev/) or [Parcel](https://parceljs.org/), and add other tools as you need them for routing, static generation or server-side rendering, and more.
 
 </DeepDive>
+
+## Quick Setup {/*quick-setup*/}
+
+### Vite {/*vite*/}
+
+**[Vite](https://vite.dev/) is a build tool that aims to provide a faster and leaner development experience for modern web projects.** It's opinionated and comes with sensible defaults out of the box. To create a new Vite project, run in your terminal:
+
+<TerminalBlock>
+npx create-vite@latest
+</TerminalBlock>
 
 ## Production-grade React frameworks {/*production-grade-react-frameworks*/}
 


src/content/learn/start-a-new-react-project.md
```diff
@@ -31,6 +30,16 @@ If your app has unusual constraints not served well by these frameworks, or you
 
 </DeepDive>
 
+## Quick Setup {/*quick-setup*/}
```
karlhorky, line 33: Instead of emphasizing that this option is "quick", maybe it would be better to emphasize what the drawbacks are and what developers miss out on with a simple Vite setup:

1. Maybe a better heading here would be "Frontend-only React" or something (since [Vite is not integrated with React Server Components yet](https://github.com/vitejs/vite/discussions/4591))
2. Since it has these drawbacks, I guess it makes more sense further down, after the "Production-grade React frameworks"

karlhorky: Hmm, thinking more on it, it seems as if the "Can I use React without a framework?" section above already covers a lot of the drawbacks to using setups like Vite, and also links to it

Maybe it doesn't need to be more visible than that?

hichemfantar: The issue is that it isn't visible at all.

Check the issue connected to this PR.

douglasjunior: Please stop forcing users to choose NextJS. This makes no sense. SPA is still the best choice for most projects that don't need SEO optimization.

karlhorky: Next.js and other production-ready frameworks still have SPA support:

- https://nextjs.org/docs/app/building-your-application/upgrading/single-page-applications

douglasjunior: You still need to incur additional infrastructure costs to maintain a Nextjs application compared to a SPA.

A SPA, when combined with a static site bundler like Vite, Rspack, or Webpack, remains the best cost-benefit option in most cases. Therefore, the React documentation should not influence users' choices.

Moreover, NextJS often serves as bait to push users toward adopting Vercel.

karlhorky: > You still need to incur additional infrastructure costs to maintain a Nextjs application compared to a SPA

Hm, where do you see this information? With a static export, I think you can deploy to the same (free) infrastructure as other static site generator output: 

> Since Next.js supports this static export, it can be deployed and hosted on any web server that can serve HTML/CSS/JS static assets.

- https://nextjs.org/docs/app/building-your-application/deploying/static-exports

hichemfantar: Static export is not an spa.
Spa is a single html page application, static export generates an html file for each route

nickserv: Hosting platforms generally let you configure a reverse proxy to maintain SPA and MPA like routing, so that shouldn't cause issues with static exports specifically.

hichemfantar: Next js Static export still isn’t an spa though.
Remix has an spa mode but even that doesn’t behave like a real spa
https://github.com/remix-run/remix/discussions/9008

askonomm: I do find it strange how React is supposedly just a library, and yet it almost dictates that you should use a full-stack framework. Almost as if there is some financial incentive involved between React and Vercel, but I'm a cynic so ... 

In either case, the current wording makes it seem as if the only correct way to use React is with a full-stack framework, and that everything else is bad - this wording drives people away from using React as just a front-end SPA, which is still very much possible to do. If this is the intent then perhaps the documentation should say that you do not recommend using React as a SPA at all, and then people can simply start using Vue or something else instead. Right now the wording is sort of open to interpretation in a way, although heavily leaning towards telling you to use Next.js.

nickserv: React is not forcing you to do anything. Next doesn't require Vercel, and React doesn't require Next, it just recommends a meta framework for a better experience.

douglasjunior: "Better" experience for who, for Vercel? 

Since from CRA deprecation, the React docs have changed from a impartial to a Next/Vercel oriented instructions.

Next is a vendor lock-in, Next is not SPA, Next only make sense for Server Side Rendering (SEO), deploying Next outside Vercel is a pain close to the death.

"Better" experience is far away from Next.

hichemfantar: Apparently the docs have been refactored and now Vite is pushed all the way down under a "Note" section 👀

rickhanlonii:
Replaced by https://github.com/reactjs/react.dev/pull/7495 and https://github.com/reactjs/react.dev/pull/7624.

PR #7427, opened 2025-01-13
Reference docs for experimental `captureOwnerStack`

eps1lon:
React error handlers now only show production error logging in one example. 

`captureOwnerStack` docs show usage in an instrumented console.error for a dev overlay. The example with an instrumented `console.error` doesn't show interaction with error handlers

Preview: 
[captureOwnerStack](https://react-dev-git-sebbie-ownerstacks-fbopensource.vercel.app/reference/react/captureOwnerStack)
[createRoot error handlers](https://react-dev-git-sebbie-ownerstacks-fbopensource.vercel.app/reference/react-dom/client/createRoot#error-logging-in-production)
[hydrateRoot error handlers](https://react-dev-git-sebbie-ownerstacks-fbopensource.vercel.app/reference/react-dom/client/hydrateRoot#error-logging-in-production)

Enabling it on Canary: https://github.com/facebook/react/pull/32053


src/content/reference/react-dom/client/createRoot.md
```diff
@@ -45,9 +45,9 @@ An app fully built with React will usually only have one `createRoot` call for i
 
 * **optional** `options`: An object with options for this React root.
 
-  * **optional** `onCaughtError`: Callback called when React catches an error in an Error Boundary. Called with the `error` caught by the Error Boundary, and an `errorInfo` object containing the `componentStack`.
-  * **optional** `onUncaughtError`: Callback called when an error is thrown and not caught by an Error Boundary. Called with the `error` that was thrown, and an `errorInfo` object containing the `componentStack`.
-  * **optional** `onRecoverableError`: Callback called when React automatically recovers from errors. Called with an `error` React throws, and an `errorInfo` object containing the `componentStack`. Some recoverable errors may include the original error cause as `error.cause`.
+  * **optional** `onCaughtError`: Callback called when React catches an error in an Error Boundary. Called with the `error` caught by the Error Boundary, and an `errorInfo` object containing the parent Component stack in `componentStack`.
```
rickhanlonii, line 48: What is a "parent Component Stack" isn't that just the Component Stack? Adding "parent" makes it sound like it doesn't include the component that errored.

This page should also add a usage item for calling `captureOwnerStack` to get the owner stack.

src/content/reference/react/captureOwnerStack.md
```diff
@@ -0,0 +1,164 @@
...
+Experimental versions of React may contain bugs. Don't use them in production.
+
+</Wip>
+
+<Intro>
+
+`captureOwnerStack` reads the current **owner** Component stack and returns it as a string if available.
+
+```js
+captureOwnerStack();
```
rickhanlonii, line 22:
```suggestion
const stack = captureOwnerStack();
```

src/content/reference/react-dom/client/createRoot.md
```diff
@@ -369,7 +369,7 @@ root.render(<App />);
 The <CodeStep step={1}>onUncaughtError</CodeStep> option is a function called with two arguments:
 
 1. The <CodeStep step={2}>error</CodeStep> that was thrown.
-2. An <CodeStep step={3}>errorInfo</CodeStep> object that contains the <CodeStep step={4}>componentStack</CodeStep> of the error.
+2. An <CodeStep step={3}>errorInfo</CodeStep> object that contains the parent Component stack in <CodeStep step={4}>componentStack</CodeStep> of the error.
```
rickhanlonii, line 372: The capitalization of only "Component" seems weird, seems like it should either be Component Stack or component stack.

src/content/reference/react/captureOwnerStack.md
```diff
@@ -0,0 +1,164 @@
...
+`captureOwnerStack` does not take any parameters.
+
+#### Returns {/*returns*/}
+
+`captureOwnerStack` returns `string | null`.
+
+If no owner stack is available, it returns an empty string.
+Outside of development builds, `null` is returned.
+
+## Owner Component stacks vs parent Component stacks {/*owner-component-stacks-vs-parent-component-stacks*/}
```
rickhanlonii, line 59: Let's wrap this in a `<DeepDive />`

src/content/reference/react/captureOwnerStack.md
```diff
@@ -0,0 +1,164 @@
...
+
+The owner Component stack is different from the **parent** Component stack available error handlers like [`errorInfo.componentStack` in `onUncaughtError`](http://localhost:3000/reference/react-dom/client/hydrateRoot#show-a-dialog-for-uncaught-errors).
```
rickhanlonii, line 61: localhost link

src/content/reference/react/captureOwnerStack.md
```diff
@@ -0,0 +1,164 @@
...
+```
+at SubComponent
+at Component
+```
+
+Neither `App` nor the DOM components (e.g. `fieldset`) are considered owners in this stack since they didn't contribute to "creating" the node containing `SubComponent`. `App` and DOM components only forwarded the node. `App` just rendered the `children` node as opposed to `Component` which created a node containing `SubComponent` via `<SubComponent />`.
+
+Neither `Navigation` nor `legend` are in the stack at all since it's only a sibling to a node containing `<SubComponent />`.
+
+### When to use which {/*when-to-use-which*/}
```
rickhanlonii, line 129: This explanation seems too technical to really understand when to use each. This is more "what they represent" and not "when to use". It's not really an either/or either. Owner stacks can be added in addition to component stacks when you're in DEV and want to add additional debugging info.

src/content/reference/react/captureOwnerStack.md
```diff
@@ -0,0 +1,164 @@
...
+import { hydrateRoot } from 'react-dom/client';
+
+const root = hydrateRoot(
+  document.getElementById('root'),
+  <App />,
+  {
+    onCaughtError: (error, errorInfo) => {
+      if (process.env.NODE_ENV !== 'production') {
+        const ownerStack = captureOwnerStack();
+        error.stack += ownerStack;
```
rickhanlonii, line 153: Do we really recommend appending it to the existing stack? Won't that add it to the end?

IMO this should be like:

```
if (__DEV__) { 
  showDialog(error, ownerStack, componentStack);
} else {
  logProdError(error, componentStack);
}
```

src/content/reference/react/captureOwnerStack.md
```diff
@@ -0,0 +1,164 @@
...
+
+- `react@experimental`
+
+Experimental versions of React may contain bugs. Don't use them in production.
+
+</Wip>
+
+<Intro>
+
+`captureOwnerStack` reads the current **owner** Component stack and returns it as a string if available.
```
rickhanlonii, line 19: We should make it more clear that this is dev-only. I would mention it here, in a Caveat, and a Troubleshooting "My owner stacks are empty in production" section.

rickhanlonii (commented):
I think the naming we should go with is:
- Component Stack: what you're calling "parent component stack" here
- Owner Stack: what you're calling "owner component stack" here

sophiebits:
Aren't we getting rid of parent stacks entirely eventually? I thought the plan was for owner stacks to subsume them. In which case I might still discuss which stack frames you'll see but not emphasize the difference so much.

sebmarkbage:
> Aren't we getting rid of parent stacks entirely eventually? I thought the plan was for owner stacks to subsume them.

Not for production logging. They're the ones passed to the deriveStateFromError/componentDidCatch and onError on the server.

rickhanlonii:
There should also be runable sandboxes for the Usage examples

src/content/reference/react/captureOwnerStack.md
eps1lon: We do that in Next.js and it's conceptually how you arrive at the error if every JSX callsite would just be a function call.

gaearon:
> There should also be runable sandboxes for the Usage examples

To expand on that — it’s not just for it to be nice but because it’s a proof the API actually works end to end and does what the docs page says it does. We’ve had cases where that wasn’t the case and the Usage examples helped uncover the misunderstandings.

src/content/reference/react/captureOwnerStack.md
eps1lon: >  It's not really an either/or either.

I agree. This is not what I wanted to convey with this section.

src/content/reference/react-dom/client/createRoot.md
eps1lon: I feel like expanding the existing usages is sufficient. Is that what you meant? Adding a whole new section feels like overkill.

src/content/reference/react/captureOwnerStack.md
eps1lon: Done

src/content/reference/react/captureOwnerStack.md
eps1lon: Removed the section. It didn't really say anything new we already explained earlier.

src/content/reference/react/captureOwnerStack.md
eps1lon: Did all that. Also explained this in the createRoot and hydrateRoot usage examples. Though possible I missed some spots.

src/content/reference/react/captureOwnerStack.md
```diff
@@ -0,0 +1,301 @@
+---
+title: captureOwnerStack
+---
+
+<Wip>
+
+**This API is experimental and is not available in a stable version of React yet.**
```
rickhanlonii, line 7: I think we should update this to the Canary badge, and just wait until we add this to Canary to merge the docs.

src/content/reference/react/captureOwnerStack.md
```diff
@@ -0,0 +1,301 @@
...
+#### Parameters {/*parameters*/}
+
+`captureOwnerStack` does not take any parameters.
+
+#### Returns {/*returns*/}
+
+`captureOwnerStack` returns `string | null`.
+
+If no Owner Stack is available, it returns an empty string.
+Outside of development builds, `null` is returned.
```
rickhanlonii, line 57:
```suggestion
If no Owner Stack is available, it returns an empty string. Outside of development builds, `null` is returned.
```

Also, instead of "if no owner stack is available" we should say when owner stacks are not available. Like maybe not too specific, but something like "if it's not called inside a component lifecycle".

src/content/reference/react/captureOwnerStack.md
```diff
@@ -0,0 +1,301 @@
...
+```js src/index.js
+import {captureOwnerStack} from 'react';
+import {createRoot} from 'react-dom/client';
+import App, {Component} from './App.js';
+import './styles.css';
+
+createRoot(document.createElement('div'), {
+  onUncaughtError: (error, errorInfo) => {
+    // The stacks are logged instead of showing them in the UI directly to highlight that browsers will apply sourcemaps to the logged stacks.
+    // Note that sourcemapping is only applied in the real browser console not in the fake one displayed on this page.
```
rickhanlonii, line 116: Would shorten the line length of this comment so you don't force a hscroll on desktop at least:

Screenshot 2025-01-19 at 10 42 53 AM

src/content/reference/react/captureOwnerStack.md
```diff
@@ -0,0 +1,301 @@
...
+    console.log(errorInfo.componentStack);
+    console.log(captureOwnerStack());
```
rickhanlonii, line 118: Would add labels:

```suggestion
    console.log('Component Stack:', errorInfo.componentStack);
    console.log('Owner Stack:', captureOwnerStack());
```
Screenshot 2025-01-19 at 10 46 52 AM

src/content/reference/react/captureOwnerStack.md
```diff
@@ -0,0 +1,301 @@
...
+  <body>
+    <p>Check the console output.</p>
+  </body>
+</html>
+```
+
+</Sandpack>
+
+`SubComponent` would throw an error.
+The Component Stack of that error would be
```
rickhanlonii, line 160: I've seen this in a few places - think there's way too many newlines in these docs, instead I would write it more paragraph style:

```suggestion
`SubComponent` would throw an error. The Component Stack of that error would be:
```

But you should also expand here, because it's not clear what SubComponent is:

```suggestion
In this example, `SubComponent` throws an error in render, which is not caught by an error boundary. The root `onUncaughtError` method is called, and logs the Component Stack and Owner Stack. The Component Stack of that error is:
```

src/content/reference/react/captureOwnerStack.md
```diff
@@ -0,0 +1,301 @@
...
+
+Neither `Navigation` nor `legend` are in the stack at all since it's only a sibling to a node containing `<SubComponent />`.
+
+`SubComponent` is omitted because it's already part of the callstack.
+
+</DeepDive>
+
+## Usage {/*usage*/}
+
+### Improve error reporting {/*improve-error-reporting*/}
```
rickhanlonii, line 187: When I read this and the usage below, I hear "reporting for production logging" which isn't the case. I really think the usages here are limited to just instrumenting DEV dialogs.

src/content/reference/react/captureOwnerStack.md
```diff
@@ -0,0 +1,301 @@
...
+
+```js src/index.js [[1, 8, "error.stack"], [2, 7, "captureOwnerStack()"]]
+import {captureOwnerStack} from 'react';
+import {createRoot} from 'react-dom/client';
+
+const root = createRoot(document.getElementById('root'), {
+  onUncaughtError: (error, errorInfo) => {
+    if (process.env.NODE_ENV !== 'production') {
+      const ownerStack = captureOwnerStack();
+      error.stack += ownerStack;
```
rickhanlonii, line 210: Why would you do this? Isn't it already in the stack if you have DevTools open? And if DevTools is open, won't this double add the owner stack?

src/content/reference/react/captureOwnerStack.md
```diff
@@ -0,0 +1,301 @@
...
+
+const root = createRoot(document.getElementById('root'), {
+  onUncaughtError: (error, errorInfo) => {
+    if (process.env.NODE_ENV !== 'production') {
+      const ownerStack = captureOwnerStack();
+      error.stack =
+        // The stack is only split because these sandboxes don't implement ignore-listing 3rd party frames via sourcemaps.
+        // A framework would ignore-list stackframes from React via sourcemaps and then you could just `error.stack += ownerStack`.
+        // To learn more about ignore-listing see https://developer.chrome.com/docs/devtools/x-google-ignore-list
+        error.stack.split('\n    at react-stack-bottom-frame')[0] + ownerStack;
```
rickhanlonii, line 248: This should be a separate usage section like "How to collapse frames in DEV dialogs without sourcemaps"

rickhanlonii: Or maybe Troubleshooting

src/content/reference/react/captureOwnerStack.md
```diff
@@ -0,0 +1,301 @@
...
+```
+
+</Sandpack>
+
+## Troubleshooting {/*troubleshooting*/}
+
+### The Owner Stack is `null` {/*the-owner-stack-is-null*/}
+
+`captureOwnerStack` was called outside of development builds.
+For performance reasons, React will not keep track of Owners in production.
```
rickhanlonii, line 296:
```suggestion
`captureOwnerStack` was called outside of development builds. For performance reasons, React will not keep track of Owner Stacks in production.
```

src/content/reference/react/captureOwnerStack.md
```diff
@@ -0,0 +1,301 @@
...
+
+### The Owner Stack is empty {/*the-owner-stack-is-empty*/}
+
+The call of `captureOwnerStack` happened outside of a React controlled function e.g. in a `setTimeout` callback.
+During render, Effects, Events, and React error handlers (e.g. `hydrateRoot#options.onCaughtError`) Owner Stacks should be available.
```
rickhanlonii, line 301:
```suggestion
The call of `captureOwnerStack` happened outside of a React controlled function e.g. in a `setTimeout` callback. During render, Effects, Events, and React error handlers (e.g. `hydrateRoot#options.onCaughtError`) Owner Stacks should be available.
```

And can we add a sandbox with the issue and fix?

src/content/reference/react/captureOwnerStack.md
rickhanlonii: This also seems weird to me because React DevTools adds the owner stack in the stack trace? Why would you want to do this?

rickhanlonii (changes requested):
The other comments are nits, but I think there's a usage section still missing here for patching console.error. If you only instrument owner stacks for DEV dialogs in the root error handlers, you will miss owner stacks in console.error calls. So you need to patch console.error, but you need to patch it _after_ dev tools, which is tricky to get right and needs docs.

src/content/reference/react/captureOwnerStack.md
rickhanlonii: We call this out in the troubleshooting section so we should just also include that explanation here.

src/content/reference/react/captureOwnerStack.md
```diff
@@ -0,0 +1,301 @@
...
+      error.stack =
+        // The stack is only split because these sandboxes don't implement ignore-listing 3rd party frames via sourcemaps.
+        // A framework would ignore-list stackframes from React via sourcemaps and then you could just `error.stack += ownerStack`.
+        // To learn more about ignore-listing see https://developer.chrome.com/docs/devtools/x-google-ignore-list
+        error.stack.split('\n    at react-stack-bottom-frame')[0] + ownerStack;
+    }
+
+    // The stacks are logged instead of showing them in the UI directly to highlight that browsers will apply sourcemaps to the logged stacks.
+    // Note that sourcemapping is only applied in the real browser console not in the fake one displayed on this page.
+    console.error('Uncaught', error);
```
rickhanlonii, line 253: If you've patched `console.error` to add owner stacks, and you call console.error here, it will result in multiple owner stacks being applied. It's not clear to me whether you should only patch console.error, or if you should instrument the root handlers and console.error without them overlapping somehow. 

@sebmarkbage how should this work ideally? Maybe we're over-emphasizing the root handler case?

The way I expected this to work is like:

```
onUncaughtError() {
  if (process.env.NODE_ENV !== 'production') {
      const ownerStack = captureOwnerStack();
      sendDevOnlyLogs(error, ownerStack);
  } else {
    sendProductionLogs(error);
  }
  
  // Patch or DevTools will add logs to the console output.
  console.error(error);
}
```

src/content/reference/react/captureOwnerStack.md
eps1lon: This i due to how we render markdown. Normally markdown doesn't render newlines if you put two sentences on two lines. I just got into the habit of it since it makes review and suggestions easier.

Will adjust.

src/content/reference/react/captureOwnerStack.md
eps1lon: Makes sense. Need to remember to update the sandboxes to use the Canary channel. I'll leave them on experimental for now so that we can verify they're working.

src/content/reference/react/captureOwnerStack.md
eps1lon: component lifecycle can read too specific. It's not obvious that `onCaughtError` is part of the component lifecycle. 

I'll be more specific here though.

src/content/reference/react/captureOwnerStack.md
eps1lon: done

src/content/reference/react/captureOwnerStack.md
eps1lon: Should've caught all accidental newlines now.

src/content/reference/react/captureOwnerStack.md
eps1lon: Removed this section. Better leave this to React Devtools instead of polyfills.

src/content/reference/react/captureOwnerStack.md
eps1lon: Added one for custom DOM event handlers which is also a likely pitfall.

src/content/reference/react/captureOwnerStack.md
eps1lon: Yeah I think the usage in error handlers is the better example. If there's a need to document polyfilling what React Devtools is doing, we can add it later.

src/content/reference/react/captureOwnerStack.md
eps1lon: Done

src/content/reference/react-dom/client/createRoot.md
rickhanlonii: Isn't there a use case for calling it in user land to add to logs? Or should this only be called in the root handlers and error patch?

src/content/reference/react/captureOwnerStack.md
rickhanlonii: @eps1lon how do you implement a dev dialog without patching console.error? You can't only do the root handlers because you'll miss react warnings like key errors.

eps1lon: This seems like a separate task and not strictly tied to `captureOwnerStack`. We haven't documented how to implement a dev dialog that displays `console.error` before so I don't see why this needs to be coupled with `captureOwnerStack`. 

In Next.js we're just reading errors from console. Owner stacks are then always displayed in the overlay via `captureOwnerStack` (no option). But displaying Component stacks directly in the console is left to React DevTools.

rickhanlonii: > We haven't documented how to implement a dev dialog that displays console.error

Yeah but we have for the other root handlers, and if you did that plus your own console.error patch before, then owner stacks break that implementation so you need docs for how to add it back. So we need to at least tell people how to get back to what worked before owner stacks.

eps1lon:
Reduced error handler usages to just production logging. Showing a minimal dev overlay in captureOwnerStack usage section that only intercepts `console.error`. In the blog post, we'll expand on Owner Stack with

rickhanlonii (approved):
Sick

PR #7437, opened 2025-01-19
feat: migrate React.dev to the App Router

feedthejim:
## What

This PR converts all of the react.dev codebase to use the App Router.

## Why

I wanted to prepare it for dogfooding future features of React and Next.js.

## In this PR

- implemented/fixed HMR when editing mdx files
- rewrote the post-processing we did to be actual remark plugins
- rewrote the RSS script to use static route handlers
- rewrote the page rendering code to use server components, streamlining some of the logic
- leveraged Next.js's metadata methods
- enabled the new Next.js Error Overlay
- added myself as a site contributor
- added a few minor polishes for perf, namely some font optimisations

## Not in this PR

- leveraging parallel routes for the sidebars
- moving more content to be server components
- a deeper refactor of the code to actually use pathname normally
- a rewrite of the TOC, whose code is kinda weird today

### Todo

- [x] convert main page
- [x] re-do MDX generation
- [x] convert error page
- [x] convert not found
- [x] convert errors/[error]
- [x] add back languages
- [ ] optional: split sidebars with parallel routes
- [x] fix metadata gen
- [x] add /errors back
- [x] see if I can hack some router events alternative
- [x] fix TOC generation
- [x] re-do RSS generation
- [x] double check analytics logic
- [x] fix font flashing
- [x] make sure the path logic transformation is actually valid
- [x] check that the router methods have been replaced correctly


src/app/error.tsx
```diff
@@ -1,18 +1,24 @@
+'use client';
```
rickhanlonii, line 1: Not a server component?

src/content/learn/index.md
```diff
@@ -4,7 +4,7 @@ title: Quick Start
 
 <Intro>
 
-Welcome to the React documentation! This page will give you an introduction to 80% of the React concepts that you will use on a daily basis.
+Welcome to the React! This page will give you an introduction to 80% of the React concepts that you will use on a daily basis.
```
gaearon, line 7: ?

src/app/error.tsx
sophiebits: error page must be a client component https://nextjs.org/docs/app/api-reference/file-conventions/error

src/content/learn/index.md
rickhanlonii:
```suggestion
Welcome to the React documentation! This page will give you an introduction to 80% of the React concepts that you will use on a daily basis.
```

rickhanlonii (approved):
sick thanks!

gaearon:
Noticed a significant UX regression.

Previously the sidebar would animate and preserve scroll position on click:


https://github.com/user-attachments/assets/c3b0f971-6b97-46e8-b963-04e730920e31


Now it doesn't:


https://github.com/user-attachments/assets/2f699a9c-efa9-4fdd-826d-e66623cdb3f7


The lack of animation is *maybe* tolerable but the jumping is not tolerable — it immediately makes the site feel unpolished. We specifically took care for it not to jump in the previous implementation.

IMO this has to either be a fast-follow fix or a revert until App Router can handle this (if App Router doesn't let us preserve scroll position easily there). I assume this has to be fixable? Is this what the parallel route refactor will do?

gaearon:
TOC slightly regressed. Before:

Screenshot 2025-02-01 at 19 08 02

After:

Screenshot 2025-02-01 at 19 08 09


(Note the lonely `set`.)

gaearon:
Links to specific example tabs are broken.

For example: https://react-gu30t1oio-fbopensource.vercel.app/reference/react/useState#passing-the-initial-state-directly

This works on the old site.

On the new site, https://react.dev/reference/react/useState#passing-the-next-state-directly still displays the first tab.

gaearon:
After navigating to a fragment link and then to another page, the Back button gets completely borked:

https://github.com/user-attachments/assets/653ccf4d-9444-439b-b1b4-2ce4795e4217

gaearon:
I suggest reverting at least until the navigation bugs are fixed since they interfere with people using the site.

https://github.com/reactjs/react.dev/pull/7466

gaearon:
Also, have we done a webpagetest comparison for the index page and for a few docs pages? Is this neutral or an improvement? Seems like it would be bad to land without verifying that it hasn't regressed on perf.

gaearon:
(Sorry to be annoying btw! Really appreciate your work there. I think this is close but just needs a bit of polish so that it doesn’t regress the functionality.)

feedthejim:
No problem at all, there was some transformed code that I was not sure where they exactly applied. 

@rickhanlonii lets revert it and I can dig deeper.

@gaearon if you have some other specific scenarios I can verify, please let me know 🙏

gaearon:
I was mostly just stress-testing navigation (all kinds — via sidebar, via footer links, normal links, fragment and example and TOC links) and jumping back and forth using Back/Forward.

feedthejim:
I didn't get time to look back into this @gaearon, I'm not sure those bugs were App Router bugs rather than just poor conversion of the current navigation code on my end.

gaearon:
Gotcha. Can’t promise anything but maybe I’ll try to take a look in the coming weeks

clicktodev:
is this a good time to revisit this?

gaearon:
Yes, happy to review if someone fixes the above issues!

clicktodev:
I'll take a look and confirm if the problems mentioned above are fixed by the latest version of the app router.
also while i have your attention, can we merge this quick fix #8121

feedthejim:
The bugs that were found were mostly due to my faulty implementation, updating Next.js won't solve that.

SukkaW:
@gaearon @feedthejim

IMHO, we might wanna perform a gradual migration here.

First, we could extract the layout as a full client component that can be used for both Pages Router's `_app.tsx` and App Router `layout.tsx`. For simplicity, this shared layout component should be a full client component (that can accept server JSX via `children` prop). We can always refactor the layout to be "more server" later.

And then, we can migrate the error decoder to the App Router. Previously, I helped migrate the React error decoder from the legacy site to the current site (https://github.com/reactjs/react.dev/pull/6214), so I am somewhat familiar with it.

The first PR of the migration process, IMHO, should only contain these two parts, which verify that the existing layout works with App Router. I can get my hand on this when I have some time.

Only after that should we figure out how to handle all the MDX stuff. Making the existing client component "more server" should be the last step.

sophiebits:
Doing a gradual migration makes sense to me.

I also have an attempt at converting the site from 2 years ago if it's helpful. https://github.com/reactjs/react.dev/compare/main...sophiebits:react.dev:next-13-rsc

PR #7462, opened 2025-01-30
Fix Vite URLs

karlhorky:
Hi, thanks again for the react.dev docs and the other projects in the React ecosystem

Quick PR to avoid the redirect from https://vitejs.dev/ to https://vite.dev/


rickhanlonii:
Can you fix the other URLs too? There are two on `add-react-to-an-existing-project.md`

karlhorky:
> Can you fix the other URLs too? There are two on `add-react-to-an-existing-project.md`

Good point, fixed the 3 URLs in 4d39d74

rickhanlonii (approved):
ty!

karlhorky:
Glad to help, thanks for the review and merge!

PR #7479, opened 2025-02-05
add doc for importMap

anisjonischkeit:
Implemented in https://github.com/facebook/react/pull/27260


rickhanlonii:
@gnoff is this in stable? Are these doc correct?

PR #7482, opened 2025-02-05
Update Server Functions with Actions logic

timmparsons:
In the docs it says:
```const submitAction = async () => {
    startTransition(async () => {
      const {error} = await updateName(name);
      if (!error) {
        setError(error);
      } else {
        setName('');
      }
    })
  }```
  but looks like the `if (!error)` is incorrect and should be ` if (error)`


rickhanlonii (approved):
thanks!

PR #7490, opened 2025-02-12
docs: useEffect for data fetching section

GustavEikaas:
Hello!

I was reading the useEffect docs today and I got a bit curious about the useEffect for data fetching scenario.
The way I understand it, this is sort of a 101 section with super simple examples just to teach the users the basics.
However I was wondering if the example provided is sufficient? No loading states, no error states? On the off-chance anyone copies the example and uses it in a codebase it is missing some basic functionality. 

I tried to make a change that doesnt overcomplicate the example but also provides the absolute minimum of what you would need should you decide to use a useEffect for data fetching.

Would love to hear your thoughts. I know there are more places further down in the docs that needs to be updated as a result of this but I would love to discuss the changes first. If they are acceptable I will update all places

### DIFF
--- a/src/content/reference/react/useEffect.md
+++ b/src/content/reference/react/useEffect.md
@@ -907,19 +907,27 @@ import { fetchBio } from './api.js';
 export default function Page() {
   const [person, setPerson] = useState('Alice');
   const [bio, setBio] = useState(null);
+  const [isLoading, setIsLoading] = useState(false)
+  const [error, setError] = useState(null)
 
   useEffect(() => {
-    let ignore = false;
-    setBio(null);
-    fetchBio(person).then(result => {
-      if (!ignore) {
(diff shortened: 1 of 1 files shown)


mattcarrollcode:
@GustavEikaas Thank you for taking the time to open a PR and think through what might be missing from the docs!

> However I was wondering if the example provided is sufficient?

By design, these usage examples are *not* sufficient. The reference docs are intended to be the minimum needed to describe an API so you can get started using it. The usage examples document common patterns and ways of using the API to help the reader better understand how to use the API. They are not meant to be a example to copy-paste into your app or represent best practices for the surrounding code that isn't directly related to using the API.

GustavEikaas:
I see.

Thanks for taking the time to review this pull request 😁

PR #7617, opened 2025-02-16
fix(docs): webpack branding states it should be lower case

zackarychapple:
This is a super minor PR but [webpack branding guidelines](https://webpack.js.org/branding/) state that it should be lower case and the only allowed upper case is in the beginning of a sentence.


harish-sethuraman:
Thanks!

PR #7622, opened 2025-02-17
fix(docs): prevent mdx from transforming -- into – for commands

### LINKED ISSUE #7616: [Bug]: Can't copy the right command from the command block | double dashes -- gets merged into a single dash -
### Summary

When I try to copy a command, the double dashes `--` get merged into a single dash `-`.

Command Block :

Image

When I copy the command :

Image

### Page

https://react.dev/learn/building-a-react-framework#step-1-install-a-build-tool

### Details

The command is correctly written in the markdown file :
Image

Note that the issue seems to be with how MDX gets processed and not the `<TerminalBlock>`  component directly.

Found this bug while fixing the issue #7614 with the PR #7615 (wrong Vite command)

elitalpa:
# Overview

- preventing mdx from transforming -- into – for commands (#7616 )
 
Apparently any text in markdown gets processed and double dashes transformed.
If a command contains `--` and we want it to be treated as a string, it should be put between {\` and \`} or between {" and "}.

Fixes #7616

### DIFF
--- a/src/content/learn/building-a-react-framework.md
+++ b/src/content/learn/building-a-react-framework.md
@@ -33,7 +33,7 @@ The first step is to install a build tool like `vite`, `parcel`, or `rsbuild`. T
 [Vite](https://vite.dev/) is a build tool that aims to provide a faster and leaner development experience for modern web projects.
 
 <TerminalBlock>
-npm create vite@latest my-app -- --template react
+{`npm create vite@latest my-app -- --template react`}
 </TerminalBlock>
 
 Vite is opinionated and comes with sensible defaults out of the box. Vite has a rich ecosystem of plugins to support fast refresh, JSX,  Babel/SWC, and other common features. See Vite's [React plugin](https://vite.dev/plugins/#vitejs-plugin-react) or [React SWC plugin](https://vite.dev/plugins/#vitejs-plugin-react-swc) and [React SSR example project](https://vite.dev/guide/ssr.html#example-projects) to get started.
@@ -45,7 +45,7 @@ Vite is already being used as a build tool in one of our [recommended frameworks
 [Parcel](https://parceljs.org/) combines a great out-of-the-box development experience with a scalable architecture that can take your project from just getting started to massive production applications.
 
 <TerminalBlock>
-npm install --save-dev parcel
+{`npm install --save-dev parcel`}
 </TerminalBlock>
 
 Parcel supports fast refresh, JSX, TypeScript, Flow, and styling out of the box. See [Parcel's React recipe](https://parceljs.org/recipes/react/#getting-started) to get started.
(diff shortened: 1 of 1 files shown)


elitalpa:
# Comparing

Notice the `--` in commands.

## Before
Screenshot 2025-02-17 085243

## After
Screenshot 2025-02-17 085319

src/content/learn/building-a-react-framework.md
```diff
@@ -33,7 +33,7 @@ The first step is to install a build tool like `vite`, `parcel`, or `rsbuild`. T
 [Vite](https://vite.dev/) is a build tool that aims to provide a faster and leaner development experience for modern web projects.
 
 <TerminalBlock>
-npm create vite@latest my-app -- --template react
+{`npm create vite@latest my-app -- --template react`}
```
harish-sethuraman, line 36: I dont think we should do this (making it string) We might want to check if this was working before and fix the root cause in TerminalBlock component. This component is in a lot of places.

elitalpa: ## not a bug but a feature ?

Actually it seems to be a feature of a markdown plugin, not a bug.
The plugin name is "[retext-smartypants](https://github.com/retextjs/retext-smartypants)" and is implementing the smartypants syntax which transforms :
> Dashes (“--” and “---”) into en- and em-dash entities
> source (smartypant syntax docs) : [daringfireball.net/projects/smartypants]( https://daringfireball.net/projects/smartypants/)

The text inside the component gets treated as normal text because it's not put between {\`  \`} or {" "} so it modifies the string before it gets to the TerminalBlock component.

Disabling this feature of the plugin might not be an option as this feature might be used in normal text across the docs.

## Fast and Simple Solution

Putting the code between {\`  \`} or {" "} makes the component receive the raw string. 
Although his change doesn't need to be done everywhere, only when needed like here.

Using {\`  \`} or \{\" \"\} or \{\' \'\} is something common in React's JSX syntax (the docs are written in MDX which is based on JSX) used for making sure certain characters are escaped when used as childrens of a component or html tag.

## Alternative solution

The other alternative is to stop using the component directly but rather use markdown code blocks :
\`\`\`sh
\`\`\`
and maybe configure mdx to replace every code block with the TerminalBlock component. But this can require a lot more effort and impact many other pages.

elitalpa: tldr : 
- root cause doesn't seem to be the TerminalBlock component but the retext-smartypants markdown plugin which transforms any text
- this feature of the plugin might be used across the docs and disabling it doesn't seem to be an option
- possibles solutions are 
  - using curly braces (common in React)
  - never using a React component for commands and using basic markdown codeblocks

rickhanlonii: Yeah the fix makes sense, we can fix it in terminal block if needed later

rickhanlonii:
Thanks, merged this into #7624.

If you have a chance, it would be nice to audit all of the other <TerminalBlock> usages to see if they have issues.

elitalpa:
> Thanks, merged this into #7624.
> 
> If you have a chance, it would be nice to audit all of the other usages to see if they have issues.

From what I noticed, other commands are not using the TerminalBlock component but a markdown code block like :
image

So this explains why there wasn't an issue like this previously.

Or -D is used instead of --save-dev :
image

✅ Checked all the other usages, there seems to be no issue.

PR #7624, opened 2025-02-17
Updates from feedback

rickhanlonii:
## Overview

We received some good feedback on the Sunsetting Create React blog post and new Build a Framework docs. This PR attempts to address some of the main concerns by clarifying the language and migration steps.

### Concern 1: no migration steps to a build tool

Heard a lot of feedback that we're not providing migration steps to a build tool for users who have use cases where a framework isn't appropriate. The intent was to direct users to the "build a framework" docs to select a build tool to migrate to, but this wasn't clear. ""

Addresses this feedback by adding links to build tool migration guides in the blog post when a framework is not a good fit:

Screenshot 2025-02-17 at 3 21 24 PM

### Concern 2: "building a framework" title unclear

We intended for the "building a framework" page to also cover the use cases where using a framework isn't a good fit or users just want to learn how react works. In those cases, many users will end up building their own framework, so we titled it that, but it's not alway the case that you will need to build a framework (e.g. [triplex](https://bsky.app/profile/douges.dev/post/3lidkreblgc24)). To address, this PR re-titles the page to "Build a React App from Scratch", and reframes step 2 as "build common application patterns":

Screenshot 2025-02-17 at 3 22 01 PM

### Concern 3: "Create a React app" does not direct users to build tools

Heard feedback that the Create a React App page doesn't direct users to docs to use a build tool. Although the Installation page is the true landing page for the installation steps, which mentions both options, users still seem confused that it's missing. To address, this pulls in the change from #7618 to add a "Start from scratch" option on the Create a React App page:

Screenshot 2025-02-17 at 3 22 32 PM

### Concern 4: We "recommend Vite with a framework" isn't well received

Heard feedback "reading this paragraph is the most indicative thing that I've seen that React has completely lost touch. It's justifiable sure... but it isn't React" ([post](https://bsky.app/profile/ryansolid.bsky.social/post/3li5wf7bnqs26)) and that the tone you can read that sound begrudging. The intent of this section was to call out that you can use Vite with a framework, or use it directly to build from scratch, but that's not clear from the text. 

Replaced this to  "Do I need a framework?" that explains that "there are valid cases to build a React app from scratch":

Screenshot 2025-02-17 at 3 27 51 PM

### Other changes

There were a lot of good changes from #7618 under Build a React App from scratch for routing and data fetching. I've pulled those over as well. I'm not including the Web Application Architecture guide from that PR because it's fairly large and will need many iterations on feedback that will take some time before landing, but this PR should incorporate all of the main changes.

### DIFF
--- a/src/content/blog/2025/02/14/sunsetting-create-react-app.md
+++ b/src/content/blog/2025/02/14/sunsetting-create-react-app.md
@@ -2,7 +2,7 @@
 title: "Sunsetting Create React App"
 author: Matt Carroll and Ricky Hanlon
 date: 2025/02/14
-description: Today, we’re deprecating Create React App for new apps, and encouraging existing apps to migrate to a framework. We’re also providing docs for when a framework isn’t a good fit for your project, or you prefer to start by building a framework.
+description: Today, we’re deprecating Create React App for new apps, and encouraging existing apps to migrate to a framework, or to migrate to a build tool like Vite, Parcel, or RSBuild. We’re also providing docs for when a framework isn’t a good fit for your project, you want to build your own framework, or you just want to learn how React works by building a React app from scratch.
 ---
 
 February 14, 2025 by [Matt Carroll](https://twitter.com/mattcarrollcode) and [Ricky Hanlon](https://bsky.app/profile/ricky.fm)
@@ -11,7 +11,9 @@ February 14, 2025 by [Matt Carroll](https://twitter.com/mattcarrollcode) and [Ri
 
 <Intro>
 
-Today, we’re deprecating [Create React App](https://create-react-app.dev/) for new apps, and encouraging existing apps to migrate to a [framework](/learn/creating-a-react-app). We’re also providing docs for when a framework isn’t a good fit for your project, or you prefer to start by [building a framework](/learn/building-a-react-framework).
+Today, we’re deprecating [Create React App](https://create-react-app.dev/) for new apps, and encouraging existing apps to migrate to a [framework](#how-to-migrate-to-a-framework), or to [migrate to a build tool](how-to-migrate-to-a-build-tool) like Vite, Parcel, or RSBuild. 
+
+We’re also providing docs for when a framework isn’t a good fit for your project, you want to build your own framework, or you just want to learn how React works by [building a React app from scratch](/learn/build-a-react-app-from-scratch).
 
 </Intro>
 
@@ -46,6 +48,9 @@ This error message will only be shown once per install.
 </ConsoleLogLine>
 </ConsoleBlockMulti>
 
+We've also added a deprecation notice to the Create React App [website](https://create-react-app.dev/) and GitHub [repo](https://github.com/facebook/create-react-app). Create React App will continue working in maintenance mode, and we've published a new version of Create React App to work with React 19.
+
+## How to Migrate to a Framework {/*how-to-migrate-to-a-framework*/}
 We recommend [creating new React apps](/learn/creating-a-react-app) with a framework. All the frameworks we recommend support client-only SPAs, and can be deployed to a CDN or static hosting service without a server.
 
 For existing apps, these guides will help you migrate to a client-only SPA:
@@ -54,27 +59,33 @@ For existing apps, these guides will help you migrate to a client-only SPA:
 * [React Router’s framework adoption guide](https://reactrouter.com/upgrading/component-routes).
 * [Expo webpack to Expo Router migration guide](https://docs.expo.dev/router/migrate/from-expo-webpack/)
 
-Create React App will continue working in maintenance mode, and we've published a new version of Create React App to work with React 19.
+## How to Migrate to a Build Tool {/*how-to-migrate-to-a-build-tool*/}
 
 If your app has unusual constraints, or you prefer to solve these problems by building your own framework, or you just want to learn how react works from scratch, you can roll your own custom setup with React using Vite, Parcel or Rsbuild.
 
-To help users get started with Vite, Parcel or Rsbuild, we've published new docs for [Building a Framework](/learn/building-a-react-framework). Continue reading to learn more about the [limitations of Create React App](#limitations-of-create-react-app) and [why we recommend frameworks](#why-we-recommend-frameworks).
+For existing apps, these guides will help you migrate to a build tool:
 
-<Note>
+* [Vite Create React App migration guide](https://www.robinwieruch.de/vite-create-react-app/)
+* [Parcel Create React App migration guide](https://stackoverflow.com/a/49605484)
+* [Rsbuild Create React App migration guide](https://rsbuild.dev/guide/migration/cra)
 
-#### Do you recommend Vite? {/*do-you-recommend-vite*/}
+To help get started with Vite, Parcel or Rsbuild, we've added new docs for [Building a React App from Scratch](/learn/build-a-react-app-from-scratch).
 
-We provide several Vite-based recommendations.
+<DeepDive>
 
-React Router v7 is a Vite based framework which allows you to use Vite's fast development server and build tooling with a framework that provides routing and data fetching. Just like the other frameworks we recommend, you can build a SPA with React Router v7.
+#### Do I need a framework? {/*do-i-need-a-framework*/}
 
-We also recommend using Vite when [adding React to an existing project](/learn/add-react-to-an-existing-project), or [building a framework](/learn/building-a-react-framework).
+Most apps would benefit from a framework, but there are valid cases to build a React app from scratch. A good rule of thumb is if your app needs routing, you would probably benefit from a framework. 
 
-Just like Svelte has Sveltekit, Vue has Nuxt, and Solid has SolidStart, React recommends using a framework that integrates with build tools like Vite for new projects.
+Just like Svelte has Sveltekit, Vue has Nuxt, and Solid has SolidStart, [React recommends using a framework](#why-we-recommend-frameworks) that fully integrates routing into features like data-fetching and code-splitting out of the box. This avoids the pain of needing to write your own complex configurations and essentially build a framework yourself.
 
-</Note>
+However, you can always [build a React app from scratch](/learn/build-a-react-app-from-scratch) using a build tool like Vite, Parcel, or Rsbuild.
 
-## Limitations of Create React App {/*limitations-of-create-react-app*/}
+</DeepDive>
+
+Continue reading to learn more about the [limitations of build tools](#limitations-of-create-react-app) and [why we recommend frameworks](#why-we-recommend-frameworks).
+
+## Limitations of Build Tools {/*limitations-of-build-tools*/}
 
 Create React App and build tools like it make it easy to get started building a React app. After running `npx create-react-app my-app`, you get a fully configured React app with a development server, linting, and a production build.
 
--- a/src/content/blog/2025/02/14/sunsetting-create-react-app.md
+++ b/src/content/blog/2025/02/14/sunsetting-create-react-app.md
@@ -51,7 +51,7 @@ This error message will only be shown once per install.
 We've also added a deprecation notice to the Create React App [website](https://create-react-app.dev/) and GitHub [repo](https://github.com/facebook/create-react-app). Create React App will continue working in maintenance mode, and we've published a new version of Create React App to work with React 19.
 
 ## How to Migrate to a Framework {/*how-to-migrate-to-a-framework*/}
-We recommend [creating new React apps](/learn/creating-a-react-app) with a framework. All the frameworks we recommend support client-only SPAs, and can be deployed to a CDN or static hosting service without a server.
+We recommend [creating new React apps](/learn/creating-a-react-app) with a framework. All the frameworks we recommend support client-side rendering ([CSR](https://developer.mozilla.org/en-US/docs/Glossary/CSR)) and single-page apps ([SPA](https://developer.mozilla.org/en-US/docs/Glossary/SPA)), and can be deployed to a CDN or static hosting service without a server.
 
 For existing apps, these guides will help you migrate to a client-only SPA:
 
--- a/src/content/learn/build-a-react-app-from-scratch.md
+++ b/src/content/learn/build-a-react-app-from-scratch.md
@@ -34,7 +34,7 @@ The first step is to install a build tool like `vite`, `parcel`, or `rsbuild`. T
 [Vite](https://vite.dev/) is a build tool that aims to provide a faster and leaner development experience for modern web projects.
 
 <TerminalBlock>
-npx create-vite@latest my-app --template react
+npm create vite@latest my-app -- --template react
 </TerminalBlock>
 
 Vite is opinionated and comes with sensible defaults out of the box. Vite has a rich ecosystem of plugins to support fast refresh, JSX,  Babel/SWC, and other common features. See Vite's [React plugin](https://vite.dev/plugins/#vitejs-plugin-react) or [React SWC plugin](https://vite.dev/plugins/#vitejs-plugin-react-swc) and [React SSR example project](https://vite.dev/guide/ssr.html#example-projects) to get started.
--- a/src/content/learn/build-a-react-app-from-scratch.md
+++ b/src/content/learn/build-a-react-app-from-scratch.md
@@ -1,28 +1,30 @@
 ---
-title: Building a React Framework
+title: Build a React app from Scratch
 ---
 
 <Intro>
 
-If your app has constraints not well-served by existing frameworks, or you prefer to solve these problems yourself, you can build your own framework.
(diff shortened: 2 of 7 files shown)


src/content/blog/2025/02/14/sunsetting-create-react-app.md
```diff
@@ -46,35 +48,44 @@ This error message will only be shown once per install.
...
-Create React App will continue working in maintenance mode, and we've published a new version of Create React App to work with React 19.
+## How to Migrate to a Build Tool {/*how-to-migrate-to-a-build-tool*/}
 
 If your app has unusual constraints, or you prefer to solve these problems by building your own framework, or you just want to learn how react works from scratch, you can roll your own custom setup with React using Vite, Parcel or Rsbuild.
 
-To help users get started with Vite, Parcel or Rsbuild, we've published new docs for [Building a Framework](/learn/building-a-react-framework). Continue reading to learn more about the [limitations of Create React App](#limitations-of-create-react-app) and [why we recommend frameworks](#why-we-recommend-frameworks).
+For existing apps, these guides will help you migrate to a build tool:
 
-<Note>
+* [Vite Create React App migration guide](https://www.robinwieruch.de/vite-create-react-app/)
```
rickhanlonii, line 68: What's the best guide for migrating from CRA from Vite? (cc @yyx990803)

src/content/blog/2025/02/14/sunsetting-create-react-app.md
```diff
@@ -46,35 +48,44 @@ This error message will only be shown once per install.
...
+* [Parcel Create React App migration guide](https://stackoverflow.com/a/49605484)
```
rickhanlonii, line 69: What's the best guide for migrating from Parcel? (cc @devongovett)

src/content/blog/2025/02/14/sunsetting-create-react-app.md
```diff
@@ -46,35 +48,44 @@ This error message will only be shown once per install.
...
+Most apps would benefit from a framework, but there are valid cases to build a React app from scratch. A good rule of thumb is if your app needs routing, you would probably benefit from a framework. 
 
-We also recommend using Vite when [adding React to an existing project](/learn/add-react-to-an-existing-project), or [building a framework](/learn/building-a-react-framework).
+Just like Svelte has Sveltekit, Vue has Nuxt, and Solid has SolidStart, [React recommends using a framework](#why-we-recommend-frameworks) that fully integrates routing into features like data-fetching and code-splitting out of the box. This avoids the pain of needing to write your own complex configurations and essentially build a framework yourself.
 
-Just like Svelte has Sveltekit, Vue has Nuxt, and Solid has SolidStart, React recommends using a framework that integrates with build tools like Vite for new projects.
+However, you can always [build a React app from scratch](/learn/build-a-react-app-from-scratch) using a build tool like Vite, Parcel, or Rsbuild.
 
-</Note>
+</DeepDive>
```
rickhanlonii, line 84: @ryansolid does this feel less "begrudging" and more in line with other library recommendations?

josephsavona (approved):
These changes look great. This was definitely the intent of the original changes, glad that we’re making it even more clear in light of feedback.

src/content/blog/2025/02/14/sunsetting-create-react-app.md
```diff
@@ -46,35 +48,44 @@ This error message will only be shown once per install.
...
+To help get started with Vite, Parcel or Rsbuild, we've added new docs for [Building a React App from Scratch](/learn/build-a-react-app-from-scratch).
 
-#### Do you recommend Vite? {/*do-you-recommend-vite*/}
+<DeepDive>
 
-We provide several Vite-based recommendations.
+#### Do I need a framework? {/*do-i-need-a-framework*/}
 
-React Router v7 is a Vite based framework which allows you to use Vite's fast development server and build tooling with a framework that provides routing and data fetching. Just like the other frameworks we recommend, you can build a SPA with React Router v7.
+Most apps would benefit from a framework, but there are valid cases to build a React app from scratch. A good rule of thumb is if your app needs routing, you would probably benefit from a framework.
```
johnnyreilly, line 78: > A good rule of thumb is if your app needs routing, you would probably benefit from a framework.

This feels like a surprising rule of thumb.  I'm not sure I've ever built a web app without routing. It's probably quite unusual to do so.  This feels like the advice is, in not so many words, "you always need a framework".  I'm sure it's not intended as such, but it feels like you're saying "actually, you always need a framework".

Would a better rule of thumb be something like "if SEO is a strong concern for your app you might benefit from a framework"?

rickhanlonii: You can build a client-only app with a framework so it wouldn't be correct to say it's only for SEO. Framework !== Server 

Frameworks allow you to choose the rendering strategy per route, and optimize the performance for that strategy, so most apps that need routing would benefit from that setup. 

It's not always the case though, which is why it's a rule of thumb and not black and white.

johnnyreilly: You can indeed build a client-only app with a framework, but my experience has been that when you're doing so, it constantly feels as though you're doing the "not the advised thing".  

If, like me, you're not running JavaScript on the server - then using the likes of RSCs that a framework offers and may benefit you, isn't an option.  

This advice does very much feel like "always use a framework".  For keen React users like me that are happy and productive building SPAs, this is a little frustrating.  I'm sure it's not intended to be, but I suspect that others feel the same way as me on this.

shrutikapoor08:
Glad to see the feedback being incorporated!

markerikson:
Ideally I'd still like to see the actual mentions of Vite/Parcel/Rsbuild directly on the "Creating a React App" page :) I know it's a sticking point, but it's one of the things contributing to the complaints I've seen.

That said, the revised "Build a React app from Scratch" page is a significant improvement in explanations and details at this point!

A few other thoughts:

- Can we pull over the "Do I need a server to deploy a React app?" section that I had replacing "All these frameworks can export a SPA"?  I felt that helped both make the "frameworks and servers" part have more context, and also answers the common question of "do I have to have Node to use React?" etc.
- Still don't think the list of "And more..." items is needed, especially given that the recommended frameworks don't do things like auth out of the box
- The "Metro bundler" note probably needs to be at least tweaked since it still says "if you'd like your framework to support RN".  Not sure the note still makes sense in the revised page context.

src/content/blog/2025/02/14/sunsetting-create-react-app.md
rickhanlonii: It's a common misunderstanding that RSCs require running JS on a server, or that the frameworks we recommend don't allow building SPAs. 

For example, if you use the NextJS SPA guide, you'll use RSCs for the HTML shell of the app, which is built on your CI server and deployed to static files to a CDN, so there's no JS running on the server and no SSR even though you're using RSCs. Another example is `react.dev` which runs RSCs during CI to statically generate all the docs (SSG), then deploy without running JS on a server. 

If that feels like you're doing it wrong, I love to hear why that's the case so I can forward the feedback and get it addressed. It might just feel that way because we share to opinion of the Chrome team who [has encouraged](https://developers.google.com/web/updates/2019/02/rendering-on-the-web) developers to consider static or server-side rendering over a full rehydration approach. 

There are tradeoffs to using a server, of course, which is why CSR is still supported in the options we recommend, and why we're adding the starting from scratch docs. But React isn't unique in encouraging using SSG or SSR. UI libraries like [Solid](https://docs.solidjs.com/solid-start/getting-started) and [Svelte](https://svelte.dev/docs/svelte/getting-started) encourage using SolidStart and SvelteKit, frameworks like [Angular](https://angular.dev/guide/performance) recommend it for performance, and islands architecture frameworks like [Astro](https://docs.astro.build/en/concepts/islands/) and [Marko](https://markojs.com/) do too.

elitalpa (commented):
You can see in the screenshots that when copying the commands they don't work because markdown text gets transformed (fix in #7622 ).

## Preview of the Page (notice there are no double dashes -- as in the code) :

image

## When copying the commands :

vite command screenshot

parcel command screenshot

rsbuild command screenshot

rickhanlonii:
@markerikson thanks for the feedback, I made these updates:
- Added the words "Vite", "Parcel" and "Rspack" to Creating a React App
- Removed the list from "And More..." and tweaked the wording
- Updated the Metro note, but kept it there as important context

> Can we pull over the "Do I need a server to deploy a React app?" 

I considered the section you wrote, but it was confusing and inaccurate IMO. 

It starts with "if you've built a client-side [...]" but this page is for creating a new app, in which case you haven't built anything, which is confusing. It also says that SPAs don't need Node to run, but they do need node in CI to bundle the app, and it doesn't distinguish between CSR, SPAs, and SSG (which also all need Node in CI). It also says SPAs "can work with backends written in any language" which implies SSR can't, which is misleading at best.

Most importantly though, it buries the "all of the frameworks support SPAs" point to the end of the paragraph, but this is one of the biggest pieces of FUD in the community right now so I think it needs to be short, clear, and direct.

src/content/learn/build-a-react-app-from-scratch.md
```diff
@@ -33,7 +34,7 @@ The first step is to install a build tool like `vite`, `parcel`, or `rsbuild`. T
 [Vite](https://vite.dev/) is a build tool that aims to provide a faster and leaner development experience for modern web projects.
 
 <TerminalBlock>
-npm create vite@latest my-app -- --template react
+npx create-vite@latest my-app --template react
```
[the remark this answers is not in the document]
rickhanlonii: Awesome thanks, I'll merge your PR in and revert back to the command you used!

markerikson:
Gotcha. Yeah, the Node mention I put in there was intended to cover running in production, vs bundling in CI, since "Do I need Node to run React?" is a fairly common question.  I was also trying to indicate that SPAs work well if your backend isn't JS.

rickhanlonii:
Going to merge, if there are better guides to link to or tweaks to make we can fix forward 🚀

src/content/learn/build-a-react-app-from-scratch.md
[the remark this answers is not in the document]
elitalpa: Glad I could help.

markerikson:
Thank you for listening to the feedback and making these updates!

From my standpoint, I think this should be sufficient to address the main concerns:

- There's a clear statement that you can use React without a framework, and direct links to instructions
- Vite et al are mentioned in the setup page and "From Scratch" page
- phrasing / content is much improved

I'd say this is a win for everyone!

markerikson:
Heh. Just thought of one more thing that would have been good to pull over: the "DON'T USE CRA!" note I had added at the bottom :)

DogPawHat:
Oh yeah, I want that in there as well

rickhanlonii:
@markerikson I kinda of added that, I merged it with the note that was on the Installation page.

Screenshot 2025-02-17 at 6 48 57 PM


This is the page that you go to when you click "Installation" (Creating a React App is collapsed so you can't even click directly to it) so I think it makes the most sense there.

src/content/blog/2025/02/14/sunsetting-create-react-app.md
```diff
@@ -46,35 +48,44 @@ This error message will only be shown once per install.
...
 For existing apps, these guides will help you migrate to a client-only SPA:
 
 * [Next.js’ Create React App migration guide](https://nextjs.org/docs/app/building-your-application/upgrading/from-create-react-app)
 * [React Router’s framework adoption guide](https://reactrouter.com/upgrading/component-routes).
 * [Expo webpack to Expo Router migration guide](https://docs.expo.dev/router/migrate/from-expo-webpack/)
 
-Create React App will continue working in maintenance mode, and we've published a new version of Create React App to work with React 19.
+## How to Migrate to a Build Tool {/*how-to-migrate-to-a-build-tool*/}
 
 If your app has unusual constraints, or you prefer to solve these problems by building your own framework, or you just want to learn how react works from scratch, you can roll your own custom setup with React using Vite, Parcel or Rsbuild.
```
yyx990803, line 64: Saw this PR a bit late but I want to point out the "unusual constraints" phrasing here is among those that makes users who find Vite + React just fine feel that their preference is being invalidated - when in fact there are massive amount of users falling into this category.

As evidence, here's a [chart comparing number of plain Vite + React users (calculated from total of the two official Vite React plugins minus Remix) vs. Next](https://npmtrends.com/@remix-run/react-vs-@vitejs/plugin-react-vs-@vitejs/plugin-react-swc-vs-next) - the numbers are quite close and we expect this to overtake Next at some point.

Another point I want to make is this section is talking about "Migration" - a term that mostly applies to existing applications. Most existing CRA apps likely already have their routing / data fetching implemented in some way (e.g. with older versions of React Router). Moving to something like Vite is the most proven and safe path to adopt modern tooling, without the risk of having to refactor their (currently working) routing and data fetching logic. I really wouldn't call this "unusual constraints" - rather, this is probably the most practical constraints: change the part that needs to be changed, keep the parts that was working.

src/content/blog/2025/02/14/sunsetting-create-react-app.md
yyx990803: We can work on an official one and later send a PR.

src/content/blog/2025/02/14/sunsetting-create-react-app.md
ryansolid: I was away for the weekend.. Only comment is "unusual constraints" still carries a tone of someone doing something wrong. But generally I think this will do it. It's tricky because there is a perceived history here, even if I don't think it's completely fair. Perceptions being as such makes the situation more delicate than would otherwise be merited.

src/content/blog/2025/02/14/sunsetting-create-react-app.md
rickhanlonii: I appreciate the feedback. Maybe there's some tweaking we can do to the language here. This specific sentence is just copy pasted from the intro of Build a React App from Scratch. Here we're just trying to explain that you can migrate to a build tool if you want. In the deep dive it says:

> Most apps would benefit from a framework, **but there are valid cases to build a React app from scratch**. A good rule of thumb is if your app needs routing, you would probably benefit from a framework. Just like Svelte has Sveltekit, Vue has Nuxt, and Solid has SolidStart, [React recommends using a framework](https://react.dev/blog/2025/02/14/sunsetting-create-react-app#why-we-recommend-frameworks) that fully integrates routing into features like data-fetching and code-splitting out of the box. This avoids the pain of needing to write your own complex configurations and essentially build a framework yourself. **However, you can always [build a React app from scratch](https://react.dev/learn/build-a-react-app-from-scratch) using a build tool like Vite, Parcel, or Rsbuild.**

The goal isn't to invalidate preferences, because the goal isn't to document preferences at all. The goal is to encourage users to use a setup that will allow them to build the best user experience possible, which means allowing them to integrate data fetching into routing easily, and to opt in to server features if they want instead of getting locked into a network-waterfall prone bloated CSR bundle that requires a lot of work to fix later (fwiw, this isn't a critique of Vite, since solving these problems are for frameworks and not a build tool).

As for the migration pain - you can use all of the same routing you're already using in the migration guides we provided. Both the Next guide and the React Router guide explain the steps to migrate _using your existing routing_ and you can incrementally migrate to the file based routing. I don't think it's fair to developers to frame that as less safe, or more work. There are plenty of stories of pain points migrating from Create React App to Vite (as well as successes).

Also, if we're going to compare usage, we should also include React Router (not just Remix), which has almost as many downloads as Next and Vite React combined. This is important context because it shows that most Vite users immediately install React Router. So most Vite apps are already using React Router (the library), and can choose React Router (the framework) to continue using Vite:

Screenshot 2025-02-18 at 1 55 12 PM

src/content/blog/2025/02/14/sunsetting-create-react-app.md
rickhanlonii: Yeah, I can see how that has the wrong tone. 

The intent is to cover use cases like a VSCode plugin, chrome extension, or game which are (let's be real) uncommon and don't need URL routing integrated into data-fetching and will never need server features because they're not deployed to the web. But framed in a way where folks who are new won't mistakenly pattern match themselves into this category and go down the wrong path and get locked in to CSR only.

Do you have suggestions on wording to clarify this and soften the tone?

src/content/blog/2025/02/14/sunsetting-create-react-app.md
rickhanlonii: Thanks! To be honest, not having clear and official guides (and I hadn't actually tested these, like i have the framework guides) were the only reason I didn't include this section in the original draft, and just linked to our new docs. That clearly sent the wrong message, and I needed to add this quickly, so it would be nice to get more official guides.

PR #7632, opened 2025-02-18
Fix broken header link

yaodingyd:
the header got updated but not the link


mattcarrollcode:
Nice catch! Thanks for the PR. Merging...

PR #7634, opened 2025-02-19
fix: Update examples removing unused 'index' argument

### LINKED ISSUE #7631: [Typo]: index argument of the map function is not used, so better to omit it as the second optional argument.
### Summary

In the 'Avoid duplication in state' part of this page, 'index' is used as the second argument in the map function, but never used.

### Page

https://react.dev/learn/choosing-the-state-structure

### Details

Remove the 'index' argument from the map function, as it is not used.

erasmoh:
Fixes #7631

### DIFF
--- a/src/content/learn/choosing-the-state-structure.md
+++ b/src/content/learn/choosing-the-state-structure.md
@@ -460,7 +460,7 @@ export default function Menu() {
     <>
       <h2>What's your travel snack?</h2> 
       <ul>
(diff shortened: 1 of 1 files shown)


erasmoh:
Hi @mattcarrollcode just when you have some time.

erasmoh:
Hi @poteto, if you have some time to check this one :) thanks.

PR #7637, opened 2025-02-20
fix: correct typo in hydrateRoot.md

hjunyoung:
Removed duplicated word ‘can’.


mattcarrollcode:

PR #7654, opened 2025-02-27
removing Provider as in React 19 Context can be used as a Provider

karthik-awebon:
https://react.dev/blog/2024/12/05/react-19#context-as-a-provider

As per this blog post, In React 19 you can render `<Context>` as a provider instead of `<Context.Provider>` and `<Context.Provider>` soon will be deprecated in future versions. So it is better to update the code on the page https://react.dev/learn/passing-data-deeply-with-context
@gaearon Please review this PR. Thanks

### DIFF
--- a/src/content/learn/passing-data-deeply-with-context.md
+++ b/src/content/learn/passing-data-deeply-with-context.md
@@ -468,15 +468,15 @@ import { LevelContext } from './LevelContext.js';
 export default function Section({ level, children }) {
   return (
     <section className="section">
-      <LevelContext.Provider value={level}>
+      <LevelContext value={level}>
         {children}
-      </LevelContext.Provider>
+      </LevelContext>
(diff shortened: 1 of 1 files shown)


gaearon (approved):
lgtm

locofocos:
I was reading this section on https://react.dev/learn/passing-data-deeply-with-context and I thought the docs were broken somehow. It was confusing, because the text says "Wrap them with a context provider" yet the code snippet doesn't show a provider.

Do you think it makes sense to reword the text somehow? Like
> Wrap them with a context ~provider~ to provide the LevelContext ...

Or would it be valuable to show both the old and new syntax, since it will be deprecated in "future versions"?

karthik-awebon:
@locofocos  The text 'context provider' looks good to me.

PR #7657, opened 2025-02-27
Typo correction

SandwichGouda:
Hello, 

I just corrected an obvious error in the docs. 

You can see it by yourself on the website : https://react.dev/learn/build-a-react-app-from-scratch

> Metro for React Native 
> If **you’d you’re** starting from scratch with React Native you’ll need to use [Metro](https://metrobundler.dev/), the JavaScript bundler for React Native. Metro supports bundling for platforms like iOS and Android, but lacks many features when compared to the tools here. We recommend starting with Vite, Parcel, or Rsbuild unless your project requires React Native support.

"If you’d you’re starting from scratch" should be "If you’re starting from scratch".


rickhanlonii (approved):
Thanks! Fixes https://github.com/reactjs/react.dev/issues/7708

PR #7672, opened 2025-03-08
docs: add tabbed terminal component, show npm/bun install

Electroid:
This adds support for a new component: `TabTerminalBlock`, which is similar to the `TerminalBlock` component, except it has support for multiple tabs.

The first place this is used is to show support for multiple package managers, such as npm and bun. Here's an example of the old component and the new component.

<table>
<tr>
<td align="center"><strong>Before</strong></td>
<td align="center"><strong>After</strong></td>
</tr>
<tr>
<td>

</td>
<td>

</td>
</tr>
</table>

Also, the code tabs is synced. When you select one, it switches the other code tab as well.

For further context, Bun is a npm-compatible package manager that is designed to be used with Node.js. While it's also a JavaScript runtime, these changes are limited to showing Bun as an install option alongside `npm install react`, not as a runtime option.

These changes also allow the future addition of other package managers that can be used to install React.


Electroid:
This should be ready for review! I updated the PR description to include a preview URL to Vercel.

PR #7701, opened 2025-03-31
Remove captureOwnerStack canary banner

jackpope:
This is now stable with the release of 19.1


rickhanlonii (approved):
let's also remove the canary icon from the sidebar and move it up so it's sorted alphabetically

Issue #7707, opened 2025-04-01
[Typo]: Section "My subscribe function gets called after every re-render" of useSyncExternalStore documentation is incorrect


hugmouse:
### Summary

In the given context the code that is used in the example will break due to javascript execution flow. `subscribe` is used before it being defined - it throws `Uncaught TypeError: subscribe is not a function`.

### Page

https://react.dev/reference/react/useSyncExternalStore#my-subscribe-function-gets-called-after-every-re-render

### Details

The following code will throw `Uncaught TypeError: subscribe is not a function`. Fix: declare `subscribe` before passing it to the useSyncExternalStore.

https://github.com/reactjs/react.dev/blob/b5f5134b1f0359237894bbb814b6b1d75a730650/src/content/reference/react/useSyncExternalStore.md?plain=1#L438-L447

rickhanlonii:
Nice catch, want to submit a PR?

Ideally, all 3 would move the subscribe before `uSES`. The middle example, moving it to the module, should go above the component (not for syntax reasons, but convention).

rickhanlonii:
Fixed in https://github.com/reactjs/react.dev/pull/7711, thanks again!

PR #7711, opened 2025-04-02
fix(reference): declare `subscribe` before passing it to the hook (issue #7707)

hugmouse:
Fixes the typo in the docs, related issue: https://github.com/reactjs/react.dev/issues/7707

## Context from the issue

The following code will throw `Uncaught TypeError: subscribe is not a function`. 

https://github.com/reactjs/react.dev/blob/b5f5134b1f0359237894bbb814b6b1d75a730650/src/content/reference/react/useSyncExternalStore.md?plain=1#L438-L447

Documentation page: https://react.dev/reference/react/useSyncExternalStore#my-subscribe-function-gets-called-after-every-re-render

## Fix

Just declaring `subscribe` before passing it to the hook.

## Screenshot

In case you don't want to check out the auto-generated preview :)

Screenshot of a section "My subscribe function gets called after every re-render" and related documentation of it

### DIFF
--- a/src/content/reference/react/useSyncExternalStore.md
+++ b/src/content/reference/react/useSyncExternalStore.md
@@ -405,14 +405,14 @@ If your store data is mutable, your `getSnapshot` function should return an immu
 
 This `subscribe` function is defined *inside* a component so it is different on every re-render:
 
-```js {4-7}
+```js {2-5}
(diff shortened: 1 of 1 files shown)


rickhanlonii (approved):
great thanks!

PR #7715, opened 2025-04-03
Add note about `useId` being allowed in non-async Server Components

hernan-yadiel:
I remember seeing this [PR](https://github.com/facebook/react/pull/31208) that mentions that `useId` is allowed in non-async Server Components. However, nowhere in the documentation mentions this fact.

### DIFF
--- a/src/content/reference/react/useId.md
+++ b/src/content/reference/react/useId.md
@@ -2,12 +2,6 @@
 title: useId
 ---
 
-<RSC>
-
-`useId` is allowed in non-async [Server Components](/reference/rsc/server-components).
-
-</RSC>
-
 <Intro>
 
 `useId` is a React Hook for generating unique IDs that can be passed to accessibility attributes.
@@ -51,6 +45,8 @@ function PasswordField() {
 * `useId` is a Hook, so you can only call it **at the top level of your component** or your own Hooks. You can't call it inside loops or conditions. If you need that, extract a new component and move the state into it.
 
 * `useId` **should not be used to generate keys** in a list. [Keys should be generated from your data.](/learn/rendering-lists#where-to-get-your-key)
+
+* `useId` currently cannot be used in [async Server Components](/reference/rsc/server-components#async-components-with-server-components).
 
 ---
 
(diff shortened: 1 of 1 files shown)


src/content/reference/react/useId.md
```diff
@@ -2,6 +2,12 @@
 title: useId
 ---
 
+<RSC>
+
+`useId` is allowed in non-async [Server Components](/reference/rsc/server-components).
+
+</RSC>
+
```
rickhanlonii, line 10: Can you move this to the caveats? And instead word it as something like "useId currently cannot be used in async server components". And there should be a troubleshooting item for the lint error with an explanation, but that can be added separately.

hernan-yadiel: Sure! No problem!

hernan-yadiel: I moved the note to caveats.

Where should I add the troubleshooting item you mentioned? That's on the react hooks eslint package, right?

src/content/reference/react/useId.md
```diff
@@ -46,6 +46,8 @@ function PasswordField() {
 
 * `useId` **should not be used to generate keys** in a list. [Keys should be generated from your data.](/learn/rendering-lists#where-to-get-your-key)
 
+* `useId` currently cannot be used in [async Server Components](/reference/rsc/server-components#async-components-with-server-components).
```
hernan-yadiel, line 49: Initially, I thought about wording it to highlight that `useId` is the only hook allowed in Server Components. But then I realized this wording assumes that hooks are disallowed in Server Components. Somehow I thought this was just another React Hooks Rule, or another Server Components rule, when actually that’s not exactly the case. I think this assumption is common among devs, maybe because all of them are not allowed in SCs (except from `useId` which it wasn't even documented).

My main concern is that if React later introduces more hooks usable in both Server and Client Components, explaining “some hooks are allowed on server components, others aren't” might get tricky.

I can't think of a clear solution, but perhaps labeling hooks differently could help clarify this distinction. For example, “Interactive Hooks” vs. “Non-Interactive Hooks” (or simply “Hooks”).

eps1lon (approved):
Thanks!

PR #7716, opened 2025-04-03
docs: improved clarity of useCallback explanation

Kendaka:
- Clarified the explanation of useCallback hook.
- Improved grammar and made the documentation more concise.


rickhanlonii:
Thanks but I think this type of wording change is more preference on tone, which different people will have an opinion on and submit PRs to tweak. Let's stick with what we have.

PR #7717, opened 2025-04-04
Add note about root.render microtask scheduling

### LINKED ISSUE #7317: [Suggestion] Document the usage of flushSync with createRoot.render for synchronous rendering
### Summary

Document the usage of the flushSync API with render to achieve synchronous initial rendering, effectively mimicking pre-Concurrent Mode behavior.

### Page

https://react.dev/reference/react-dom/client/createRoot

### Details

While integrating React Router (using RouterProvider), we ran into an issue. After moving most of our routes to use [loader](https://reactrouter.com/start/framework/data-loading) for the fetch-as-you-render pattern, we noticed that some routes with very simple or fast loaders were breaking the app.

The problem comes down to a timing issue (this was tricky to debug since it only happened in the production build and only when the console wasn’t open—though I’m not sure why, and it might be a red herring). React Router subscribes to its internal state in a useLayoutEffect ([relevant code](https://github.com/remix-run/react-router/blob/9aeff58b44707273d5829081eb5dcdbbd840bbfb/packages/react-router/lib/components.tsx#L323)), but if the loaders finish before this subscription happens, the root app component doesn’t get the update and can’t proceed with rendering.

While looking for a way to flush updates synchronously (shoutout to this [Stack Overflow post](https://stackoverflow.com/questions/71982225/react-18-how-to-force-synchronous-rendering)), I learned that the flushSync API can handle this (I never though about using this API outside of React, I mean before initial render...). 

Here’s a [sandbox example](https://codesandbox.io/p/sandbox/react-and-synchronous-react-render-jkpcz7?file=%2Fsrc%2Findex.js) showing how it behaves.

I think this behavior should be mentioned on [react.dev](https://react.dev/) (in the createRoot section?). It’s a key detail for anyone needing synchronous rendering in cases like this (similar to how React worked before Concurrent Mode).

I’m also not sure if React Router’s current approach is the best way to handle this. I’ll open an issue with them as well to suggest adding something about this edge case in their documentation.

Happy to help push this forward—just let me know the next steps!

jukben:
This PR adds documentation about the microtask scheduling behavior of `createRoot().render()`, as requested in #7317 and confirmed by the React team in facebook/react#32811.

The documentation clarifies that while rendering is synchronous once it starts, the `render()` call itself is scheduled in a microtask, similar to how `setState` works. This can be important when integrating with code that depends on precise effect timing (https://github.com/remix-run/react-router/issues/12459)

Fixes #7317
Closes https://github.com/facebook/react/issues/32811

### DIFF
--- a/src/content/reference/react-dom/client/createRoot.md
+++ b/src/content/reference/react-dom/client/createRoot.md
@@ -90,7 +90,7 @@ React will display `<App />` in the `root`, and take over managing the DOM insid
 
 * If you call `render` on the same root more than once, React will update the DOM as necessary to reflect the latest JSX you passed. React will decide which parts of the DOM can be reused and which need to be recreated by ["matching it up"](/learn/preserving-and-resetting-state) with the previously rendered tree. Calling `render` on the same root again is similar to calling the [`set` function](/reference/react/useState#setstate) on the root component: React avoids unnecessary DOM updates.
 
-* Although rendering is synchronous once it starts, calling `root.render(...)` schedules the render in a microtask. This means code after `root.render()` will run before any effects (`useLayoutEffect`, `useEffect`) are fired. This is usually fine and rarely needs adjustment. In rare cases where effect timing matters, you can wrap `root.render(...)` in [`flushSync`](https://react.dev/reference/react-dom/client/flushSync) to ensure the initial render runs fully synchronously.
+* Although rendering is synchronous once it starts, `root.render(...)` is not. This means code after `root.render()` may run before any effects (`useLayoutEffect`, `useEffect`) of that specific render are fired. This is usually fine and rarely needs adjustment. In rare cases where effect timing matters, you can wrap `root.render(...)` in [`flushSync`](https://react.dev/reference/react-dom/client/flushSync) to ensure the initial render runs fully synchronously.
 
 ---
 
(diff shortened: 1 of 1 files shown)


src/content/reference/react-dom/client/createRoot.md
```diff
@@ -90,6 +90,8 @@ React will display `<App />` in the `root`, and take over managing the DOM insid
 
 * If you call `render` on the same root more than once, React will update the DOM as necessary to reflect the latest JSX you passed. React will decide which parts of the DOM can be reused and which need to be recreated by ["matching it up"](/learn/preserving-and-resetting-state) with the previously rendered tree. Calling `render` on the same root again is similar to calling the [`set` function](/reference/react/useState#setstate) on the root component: React avoids unnecessary DOM updates.
 
+* Although rendering is synchronous once it starts, calling `root.render(...)` schedules the render in a microtask. This means code after `root.render()` will run before any effects (`useLayoutEffect`, `useEffect`) are fired. This is usually fine and rarely needs adjustment. In rare cases where effect timing matters, you can wrap `root.render(...)` in [`flushSync`](https://react.dev/reference/react-dom/client/flushSync) to ensure the initial render runs fully synchronously.
```
eps1lon, line 93:
```suggestion
* Although rendering is synchronous once it starts, `root.render(...)` is not. This means code after `root.render()` may run before any effects (`useLayoutEffect`, `useEffect`) of that specific render are fired. This is usually fine and rarely needs adjustment. In rare cases where effect timing matters, you can wrap `root.render(...)` in [`flushSync`](https://react.dev/reference/react-dom/client/flushSync) to ensure the initial render runs fully synchronously.
```

The exact timing of when `root.render()` starts is either an implementation detail or decided by the context in which it runs e.g. if it's called in `flushSync` it's not in the next microtask.

A small code example would help here e.g.

```
root.render(...)
// Will not include the HTML from this specific `root.render` call
console.log(document.body.innerHTML)
```

jukben: Good points, I agree, thanks. I will reflect it 👍

eps1lon (approved):
Thank you

PR #7719, opened 2025-04-04
Fix(ProductTable): Use continue for correct product filtering

### LINKED ISSUE #7698: [Typo]: Used "return" instead of "continue" in sandbox code
### Summary

From step 4 till the end of the page, mistakenly used "return" instead of "continue" in ProductTable component that results in not showing any product when the checkbox is checked.

Image

Image

Image

### Page

https://react.dev/learn/thinking-in-react#step-4-identify-where-your-state-should-live

### Details

In the ProductTable component, we handle filtering products by two if statements, the first one that filters text and the second one that filters checkbox. Consider that a product does not fit into our filters, then one of the if statements triggers and the function get returned. In this scenario the following code would not execute and it leads to ProductTable function not returning any JSX, therefore, there is no product to see in the UI.

Ajayss04:
Fix(ProductTable): Use continue for correct product filtering

Replaces `return` with `continue` within the `forEach` loop's filter conditions. Using `continue` correctly skips the current iteration for products that don't match the filters, preventing them from being added to the `rows` array. 
Closes #7698

### DIFF
--- a/src/content/learn/thinking-in-react.md
+++ b/src/content/learn/thinking-in-react.md
@@ -353,10 +353,10 @@ function ProductTable({ products, filterText, inStockOnly }) {
         filterText.toLowerCase()
       ) === -1
     ) {
-      return;
+      continue;
     }
     if (inStockOnly && !product.stocked) {
-      return;
+      continue;
     }
     if (product.category !== lastCategory) {
       rows.push(


rickhanlonii:
Inside a `forEach()`, `return` will continue to the next element, so this change isn't necessary. Will comment on the issue.

rickhanlonii:
Actually, TIL `continue` even errors when used inside `forEach()`. This is from your change:

Screenshot 2025-05-16 at 2 02 28 PM

PR #7724, opened 2025-04-07
fix(reference): missing parameter in hydrateRoot

Kafuu-Chinocya:
Fix missing parameter issue for <App /> in hydrateRoot

### DIFF
--- a/src/content/reference/react-dom/client/hydrateRoot.md
+++ b/src/content/reference/react-dom/client/hydrateRoot.md
@@ -380,10 +380,11 @@ By default, React will log all errors to the console. To implement your own erro
 
 ```js [[1, 6, "onCaughtError"], [2, 6, "error", 1], [3, 6, "errorInfo"], [4, 10, "componentStack", 15]]
 import { hydrateRoot } from "react-dom/client";
+import App from "./App.js";
 import { reportCaughtError } from "./reportError";
 
 const container = document.getElementById("root");
-const root = hydrateRoot(container, {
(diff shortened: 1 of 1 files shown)


src/content/reference/react-dom/client/hydrateRoot.md
```diff
@@ -380,10 +380,11 @@ By default, React will log all errors to the console. To implement your own erro
 
 ```js [[1, 6, "onCaughtError"], [2, 6, "error", 1], [3, 6, "errorInfo"], [4, 10, "componentStack", 15]]
 import { hydrateRoot } from "react-dom/client";
+import App from "./App.js";
 import { reportCaughtError } from "./reportError";
 
 const container = document.getElementById("root");
-const root = hydrateRoot(container, {
+const root = hydrateRoot(container, <App />, {
```
rickhanlonii, line 387: Nice catch but this causes a crash now:
Screenshot 2025-04-08 at 9 34 43 AM

This happens because the lines listed on 381 are now off by one. Can you update it so it still highlights like this:
Screenshot 2025-04-08 at 9 37 35 AM

rickhanlonii (changes requested):

src/content/reference/react-dom/client/hydrateRoot.md
Kafuu-Chinocya: Och! Thank for the catch, Let me fix it.

eps1lon (approved):
Thank you!

PR #7727, opened 2025-04-08
docs: Add note about using `use` hook in useContext documentation

vishal-tambi:
In order to increase developers' awareness of the use() hook's ability to read from context, particularly in the context of React Server Components (such as Next.js apps), this PR adds a note to the useContext documentation page.

📌 Why has this changed?
Only Client Components can use useContext().

Use() offers greater flexibility, async values, and support for Server Components.

This note assists developers in avoiding misunderstandings and provides guidance on which hook to use based on the type of component.

### DIFF
--- a/src/content/reference/react/useContext.md
+++ b/src/content/reference/react/useContext.md
@@ -951,6 +951,14 @@ ul, li { margin: 0; padding: 0; }
 
 If React can't find any providers of that particular <CodeStep step={1}>context</CodeStep> in the parent tree, the context value returned by `useContext()` will be equal to the <CodeStep step={3}>default value</CodeStep> that you specified when you [created that context](/reference/react/createContext):
 
+
+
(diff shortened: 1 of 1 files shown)


rickhanlonii:
Thanks but unfortunately this isn't accurate in a couple ways. 

First, `use(Context)` can be used without a RSC framework. Any app can use `use(Context)` today. 

Second, even in RSC frameworks, you can't `use(Context)` in Server Components, only Client Components. 

Closing because we're already working on a `use(Context)` note in #7728.

vishal-tambi:
Thanks a lot for letting me know that, I'm trying my best to update my knowledge.
Next time, I will try my best and ensure I am providing exact information .

rickhanlonii:
@vishal-tambi no worries, it's a common misunderstanding!

Issue #7723, opened 2025-04-06
[Suggestion]: Add a note to make developers more aware of `use` in the `useContext` page


adriancuadrado:
### Summary

Add a note saying this in [useContext](https://react.dev/reference/react/useContext):
> `use` is preferred over `useContext` because it is more flexible.

### Page

https://react.dev/reference/react/useContext

### Details

This is already said in [this article](https://react.dev/reference/react/use#reading-context-with-use):
> `use` is preferred over `useContext` because it is more flexible.

It would be nice to add this note in [useContext](https://react.dev/reference/react/useContext) too.

preetpdsilva-netizen:
Hey Maxwell, I'd be happy to get this new docs page setup. Is this issue still available?

PR #7728, opened 2025-04-08
Added a note about use in useContext page to make developers more aware of it.

karthik-awebon:
As per the suggestion explained in #7723 I had added a note in useContext page for the use(context) hook. 
@rickhanlonii can you please review this PR. 
Thanks

### DIFF
--- a/src/content/reference/react/useContext.md
+++ b/src/content/reference/react/useContext.md
@@ -82,6 +82,11 @@ function Form() {
 ```
 
 It doesn't matter how many layers of components there are between the provider and the `Button`. When a `Button` *anywhere* inside of `Form` calls `useContext(ThemeContext)`, it will receive `"dark"` as the value.
+<Note>
+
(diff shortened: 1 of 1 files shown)


src/content/reference/react/useContext.md
```diff
@@ -82,6 +82,11 @@ function Form() {
 ```
 
 It doesn't matter how many layers of components there are between the provider and the `Button`. When a `Button` *anywhere* inside of `Form` calls `useContext(ThemeContext)`, it will receive `"dark"` as the value.
+<Note>
+
+[`use(context)`](/reference/react/use) is preferred over `useContext` when you need to access context inside conditionals (like if) or loops (like for), as it is more flexible than `useContext`.
+
+</Note>
```
rickhanlonii, line 89: Let's move this to the top, after `<Intro>` and have it say:

```
Starting in React 19, you can use `use(Context)` instead of `useContext(Context)`, which allows you to read context in conditionals and loops. 
```

PR #7733, opened 2025-04-10
Add example of `useActionState` handling execution order

ajits01:
Code example comparing the execution order using `Transitions` vs `useActionState`


rickhanlonii:
Ideally this would be on the `useActionState` page, but until we rewrite that this seems better.

PR #7746, opened 2025-04-17
Server Functions docs fixes

kapantzak:
- Fix typo
- Clarify that a server function reference is created only when that function is used by a Client Component

### DIFF
--- a/src/content/reference/rsc/server-functions.md
+++ b/src/content/reference/rsc/server-functions.md
@@ -28,7 +28,7 @@ To support Server Functions as a bundler or framework, we recommend pinning to a
 
 </Note>
 
-When a Server Function is defined with the [`"use server"`](/reference/rsc/use-server) directive, your framework will automatically create a reference to the server function, and pass that reference to the Client Component. When that function is called on the client, React will send a request to the server to execute the function, and return the result.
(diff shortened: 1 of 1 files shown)


src/content/reference/rsc/server-functions.md
```diff
@@ -28,7 +28,7 @@ To support Server Functions as a bundler or framework, we recommend pinning to a
 
 </Note>
 
-When a Server Functions is defined with the [`"use server"`](/reference/rsc/use-server) directive, your framework will automatically create a reference to the server function, and pass that reference to the Client Component. When that function is called on the client, React will send a request to the server to execute the function, and return the result.
+When a Server Function is defined with the [`"use server"`](/reference/rsc/use-server) directive and is used by a Client Component, your framework will automatically create a reference to the Server Function, and pass that reference to the Client Component. When that function is called on the client, React will send a request to the server to execute the function, and return the result.
```
rickhanlonii, line 31: Doesn't the second part of the sentence say it passes the reference to a client component? It's not really accurate to say it's used by a Client Component because it can be used by a server component and passed to a Client Component.

kapantzak: Agree, I removed that part :)

PR #7747, opened 2025-04-17
Change 'Server Actions' to 'Server Functions' in React-19 docs

kapantzak:
Change `Server Actions` to `Server Functions` in the React 19 docs to align with the server-functions docs.


rickhanlonii:
We don't typically change the blog content this significantly after the fact. When 19 was released, we called them Server Actions and the docs include a note about the name change here: https://react.dev/reference/rsc/server-functions

PR #7754, opened 2025-04-21
Add key and improve clarity in people.map() example

LijuanTang94:
This PR improves the beginner-facing example in the "Rendering data from arrays" section.

Changes:
- Adds a `key` prop to each `<li>` element when mapping over the `people` array (to resolve console warning)

This change aligns the code with React best practices and helps new learners avoid common pitfalls when rendering lists.


rickhanlonii:
Yeah this is intentional to teach the need for the `key`.

PR #7765, opened 2025-04-22
Add a bold version of Source Code Pro

Nicolapps:
Currently, react.dev uses a single `.woff2` font file for the monospace typeface it uses (Source Code Pro). This requires the browser to create a bold weight from the regular weight it has access to (a.k.a. [faux bold](https://alistapart.com/article/say-no-to-faux-bold/)), which looks bad:

faux-bold-title
faux-bold

This PR adds the latest bold font file from the [Source Code Pro repo](https://github.com/adobe-fonts/source-code-pro). It also refreshes the regular font file from the same source. With this change, bold monospace text looks much better:

correct-bold-title
correct-bold


public/fonts/Source-Code-Pro-Regular.woff2
rickhanlonii, line 1: Why was this edited?

rickhanlonii:
Thanks but IIUC it's unsafe to add fonts from untrusted sources, so I downloaded the font and created a new PR with you as the co-author: https://github.com/reactjs/react.dev/pull/7815

Nicolapps:
Thank you!

PR #7767, opened 2025-04-22
Update "Deep Dive" in reusing-logic-with-custom-hooks.md with link and example of the use API

aurorascharff:
I noticed that the "Deep Dive" in `reusing-logic-with-custom-hooks.md` about whether React will provide a built-in data fetching solution in the future contained information about a non-existent `use` API, which now exists. I updated the "Deep Dive" with a link and example of `use` and removed the comment stating it's not yet available.

### DIFF
--- a/src/content/learn/reusing-logic-with-custom-hooks.md
+++ b/src/content/learn/reusing-logic-with-custom-hooks.md
@@ -1419,10 +1419,31 @@ Similar to a [design system,](https://uxdesign.cc/everything-you-need-to-know-ab
 
 #### Will React provide any built-in solution for data fetching? {/*will-react-provide-any-built-in-solution-for-data-fetching*/}
 
+Today, with the [`use`](/reference/react/use#streaming-data-from-server-to-client) API, data can be streamed from the server to the client by passing a Promise as a prop from a [Server Component](/reference/rsc/server-components) and resolving it in a [Client Component](/reference/rsc/use-client):
+
+```js {3,6,13}
+"use client";
+
+import { use, Suspense } from "react";
+
+function Message({ messagePromise }) {
+  const messageContent = use(messagePromise);
+  return <p>Here is the message: {messageContent}</p>;
+}
+
+export function MessageContainer({ messagePromise }) {
+  return (
+    <Suspense fallback={<p>⌛Downloading message...</p>}>
+      <Message messagePromise={messagePromise} />
+    </Suspense>
+  );
+}
+```
+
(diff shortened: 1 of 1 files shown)


src/content/learn/reusing-logic-with-custom-hooks.md
```diff
@@ -1419,10 +1419,31 @@ Similar to a [design system,](https://uxdesign.cc/everything-you-need-to-know-ab
 
 #### Will React provide any built-in solution for data fetching? {/*will-react-provide-any-built-in-solution-for-data-fetching*/}
 
+Today, with the [`use`](/reference/react/use#streaming-data-from-server-to-client) API, data can be streamed from the server to the client by passing a Promise as a prop from a [Server Component](/reference/rsc/server-components) and resolving it in a [Client Component](/reference/rsc/use-client):
```
rickhanlonii, line 1422: This kinda makes it sound like you can only use `use` with RSC

aurorascharff: I agree. The [docs](https://react.dev/reference/react/use#streaming-data-from-server-to-client) for "usage" only specify the RSC example. Would it be better to reference the more generic example [here](https://react.dev/reference/react/use#reference)?

src/content/learn/reusing-logic-with-custom-hooks.md
```diff
@@ -1419,10 +1419,29 @@ Similar to a [design system,](https://uxdesign.cc/everything-you-need-to-know-ab
 
 #### Will React provide any built-in solution for data fetching? {/*will-react-provide-any-built-in-solution-for-data-fetching*/}
 
+Today, with the [`use`](/reference/react/use#streaming-data-from-server-to-client) API, data can be streamed from the server to the client by passing a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) as a prop:
```
rickhanlonii, line 1422: It doesn't necessarily need to be streamed from the server, it just needs to be an existing cached promise. 

Maybe something like this:

```suggestion
Today, with the [`use`](/reference/react/use#streaming-data-from-server-to-client) API, data can be read in render by passing a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) to `use`:
```
With a link to the `use` reference docs?

aurorascharff:
@rickhanlonii I just committed your suggestion, thank you! Do you want anything else here? The `use` is already linked in the explanation.

rickhanlonii:
Thanks!

PR #7768, opened 2025-04-22
Fix ESLint plugin version.

cpojer:
`6.0.0` of the ESLint plugin was released accidentally as per (https://github.com/facebook/react/issues/32982#issuecomment-2820117164). However, when using a `^` for the version range, `6.0.0` takes precedence over `6.0.0-rc.1`.

This PR fixes the blog post to lock the version to `6.0.0-rc.1`.

If possible, please release `6.1.0-rc.1` of the plugin instead to ensure `6.0.0` will not accidentally be installed.


cpojer:
cc @poteto @josephsavona

poteto:
Thanks! Good call about rereleasing, I hadn't realized the caret range would pick up a deprecated package

PR #7771, opened 2025-04-23
Update Next.js deployment notes

eduardoboucas:
This PR updates the notes about deployment options on the [Next.js (App Router)](https://react.dev/learn/creating-a-react-app#nextjs-app-router) section with the following changes:

1. Replaces the mention of "serverless hosting" with "Docker containers" to reflect the three self-hosting options listed on https://nextjs.org/docs/app/guides/self-hosting. It is currently not true that you can deploy Next.js to any serverless hosting provider, as confirmed by the [Deployment Adapters API RFC](https://github.com/vercel/next.js/discussions/77740). This issue is described in more detail [here](https://eduardoboucas.com/posts/2025-03-25-you-should-know-this-before-choosing-nextjs/#fact-2-no-official-serverless-support).

2. Removes the reference to Vercel paid cloud services. There are other cloud providers also offering different paid services that apply to Next.js sites, so I don't think there's a reason to single out Vercel's in this case.

### DIFF
--- a/src/content/learn/creating-a-react-app.md
+++ b/src/content/learn/creating-a-react-app.md
@@ -32,7 +32,7 @@ This allows you to start with a client-only app, and if your needs change later,
 npx create-next-app@latest
 </TerminalBlock>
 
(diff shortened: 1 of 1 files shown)


gaearon:
How did the Vercel mention even get into this page? I didn’t write that when I was writing it originally.

rickhanlonii:
It definitely reads like a promotion which def wasn't the intent lol

leerob:
I like the new version better. Thank you!

PR #7782, opened 2025-04-27
fix: set wrap-content to true for better code readability

ad1992:
I have added wrap content to improve the code readability in the playgrounds

https://github.com/user-attachments/assets/eee1cdd1-99eb-48a5-8bf9-7873bc0696e3

### DIFF
--- a/src/components/MDX/Sandpack/CustomPreset.tsx
+++ b/src/components/MDX/Sandpack/CustomPreset.tsx
@@ -125,6 +125,7 @@ const Editor = memo(function Editor({
       showTabs={false}
       showRunButton={false}
       extensions={lintExtensions}
+      wrapContent={true}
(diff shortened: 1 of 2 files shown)


rickhanlonii:
I think this kinda makes it worse. Maybe ideally we would run this through prettier with a lower number of columns so the code is formatted for the smaller width

ad1992:
> I think this kinda makes it worse. Maybe ideally we would run this through prettier with a lower number of columns so the code is formatted for the smaller width

@rickhanlonii you mean for smaller width, the formatting might not look good since the content is wrapped ?
 
How about adding an optional checkbox `wrap text` which user can check to wrap the content instead? This way it will not be a mandatory change for everyone and still they can use it if they want to.

PR #7783, opened 2025-04-28
fix: update crossOrigin description to clarify requirement when using as="fetch"

### LINKED ISSUE #6617: [Typo]: `crossOrigin` and `as` values content
### Summary

Some document contains values ​​for `crossOrigin` and `as`, but descriptions are inconsistent.

This is the current state of the document:

image

image

image

So I thought `preinit` had too much content and submitted a [PR 6615](https://github.com/reactjs/react.dev/pull/6615).

### Page

https://react.dev/reference/react-dom/preinit

### Details

_No response_

vishal-rathod-07:
Update crossOrigin prop description to explicitly state that it's required when using `as="fetch"`.

The documentation lacked clarity in several places about the mandatory nature of crossOrigin when the as prop is set to "fetch". This change:
- Adds explicit requirement note in prop descriptions
- Ensures consistent documentation across all relevant files
- Helps developers avoid common misconfiguration issues

Also closes #6617


sophiebits (changes requested):
as="fetch" isn't valid for these so this doesn't make sense. If you want to preload a URL that you will fetch then I believe call `preload` with as="fetch".

sophiebits:
For preloadModule perhaps we could allow setting `as` to "audioworklet", "paintworklet", "serviceworker", "sharedworker", or "worker" per MDN https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/rel/modulepreload, but still "fetch" wouldn't be supported.

Issue #7751, opened 2025-04-19
[Suggestion]: Providing formal definition of same JSX or giving a clarity if there is no formal definition possible


krp12345:
### Summary

Formal Definition of Equality of two JSX values.

### Page

https://react.dev/learn/keeping-components-pure

### Details

[React documentation on "Keeping Components Pure"](https://react.dev/learn/keeping-components-pure) says that 

> React assumes that every component you write is a pure function. This means that React components you write must always return the same JSX given the same inputs.

It also mentions:

> React offers a “Strict Mode” in which it calls each component’s function twice during development. By calling the component functions twice, Strict Mode helps find components that break these rules.

However, the page does not provide a formal definition of what it means for two JSX values to be "equal."

we can easily say 4 === 4 or "foo" === "foo",
But, To give a simple example, how do we define the equality of two JSX values like: **_"<Child {childConfig}="object1">_** and **_<Child {childConfig}="object2">"_**

If object1 and object2 are deeply equal, does that imply that both JSX elements are considered equal as well? Or is equality based on referential identity of object1 and object2? 
I think, This distinction is important to understand what it truly means of two JSX values being same.

If such a definition exists elsewhere in the React documentation, it would be helpful to reference it explicitly from this page to clarify expectations around JSX equality and purity.

PR #7789, opened 2025-04-29
docs: clarify JSX equality

vib3-ch3ck:
#7751


rickhanlonii (changes requested):
This is not correct. JSX is never "equal" because it returns new objects every time. The purity aspect mentioned in the docs talk about it returning the "same JSX" which is more about when the component runs, the same thing is returned every time, so it doesn't depend on something outside that could change it. 

It's not really about JSX so much as the same output, and that doesn't mean you should "construct" and equal output (in that case you would need to know how to construct it to make it equal), it should execute the same code paths and return the same output each time.

PR #7792, opened 2025-04-29
Fix useOptimistic example

rickhanlonii:
This example was busted because the setState after await wasn't inside `startTransition`. Also converted it to use the action prop naming pattern, and put the form on top so it doesn't move when you submit.

### DIFF
--- a/src/content/reference/react/useOptimistic.md
+++ b/src/content/reference/react/useOptimistic.md
@@ -66,15 +66,15 @@ For example, when a user types a message into the form and hits the "Send" butto
 
 
 ```js src/App.js
-import { useOptimistic, useState, useRef } from "react";
+import { useOptimistic, useState, useRef, startTransition } from "react";
 import { deliverMessage } from "./actions.js";
 
-function Thread({ messages, sendMessage }) {
+function Thread({ messages, sendMessageAction }) {
   const formRef = useRef();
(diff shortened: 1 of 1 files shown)


sophiebits (approved):
if the form’s on top do you wanna prepend instead of append?

rickhanlonii:
@sophiebits good call, fixed

src/content/reference/react/useOptimistic.md
```diff
@@ -107,11 +108,15 @@ export default function App() {
...
+  function sendMessageAction(formData) {
+    startTransition(async () => {
+      const sentMessage = await deliverMessage(formData.get("message"));
+      startTransition(() => {
+        setMessages((messages) => [{ text: sentMessage }, ...messages]);
+      })
+    })
   }
-  return <Thread messages={messages} sendMessage={sendMessage} />;
+  return <Thread messages={messages} sendMessageAction={sendMessageAction} />;
```
sebmarkbage, line 119: This shouldn't really use the `Action` suffix convention if you do it this way because that implies that `Thread` wraps it in `startTransition` for you (which it could if you want to). That's what an Action is.

rickhanlonii: https://github.com/reactjs/react.dev/pull/7796

PR #7797, opened 2025-05-02
docs: update broken Next.js basePath URL

Parvezkhan0:


rickhanlonii:
You only removed whitespace? Closing since the link and formatting are correct.

PR #7800, opened 2025-05-06
Fix wrong documentation for effect events

gbhardwaj00:
`onConnected` was never listed as a dependency in the instrcutive code before this line, but this line mentions that we had to *remove* it. So it does not make full sense. The only dependency that was added was `theme` and we did remove that.

So I changed the documenation text to correct it.

### DIFF
--- a/src/content/learn/separating-events-from-effects.md
+++ b/src/content/learn/separating-events-from-effects.md
@@ -439,7 +439,7 @@ function ChatRoom({ roomId, theme }) {
   // ...
 ```
 
-This solves the problem. Note that you had to *remove* `onConnected` from the list of your Effect's dependencies. **Effect Events are not reactive and must be omitted from dependencies.**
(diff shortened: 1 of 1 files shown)


src/content/learn/separating-events-from-effects.md
```diff
@@ -439,7 +439,7 @@ function ChatRoom({ roomId, theme }) {
   // ...
 ```
 
-This solves the problem. Note that you had to *remove* `onConnected` from the list of your Effect's dependencies. **Effect Events are not reactive and must be omitted from dependencies.**
+This solves the problem. Note that you had to *remove* `theme` from the list of your Effect's dependencies, and you also don't need to *add* `onConnected` to it . **Effect Events are not reactive and must be omitted from dependencies.**
```
rickhanlonii, line 442:
```suggestion
This solves the problem. Note that you had to *remove* `theme` from the list of your Effect's dependencies, because it's no longer used in the Effect. You also don't need to *add* `onConnected` to it, because **Effect Events are not reactive and must be omitted from dependencies.**
```

rickhanlonii:
Thanks!

PR #7807, opened 2025-05-10
fix: docs give conflict answers about when effects will run

thind9xdev:


rickhanlonii:
Sorry, this isn't clearer IMO and has some repetition and typos. Keep in mind, small tweaks like this are hard for our translations to maintain, so there needs to be a clear fix needed.

PR #7821, opened 2025-05-20
Make banner visible on mobile

mattcarrollcode:


rickhanlonii:
Can you test different device widths and either hide or lower the font size so it doesn't wrap and take up too much space on smaller width devices?

PR #7833, opened 2025-05-31
fix: correct broken WAI-ARIA modal dialog link in createPortal reference

dimatitov:
While translating the createPortal reference page to Russian, I tested the external WAI-ARIA link and noticed it led to the general APG homepage (#dialog_modal anchor is missing or invalid). I’ve replaced it with the correct and direct link to the Modal Dialog pattern page: `https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal`

This ensures users can access the specific modal accessibility guidelines without confusion.


rickhanlonii (approved):
nice thanks

PR #7837, opened 2025-06-02
Remove `forwardRef` reference from API listing

kassens:
This API is now under "Legacy APIs" and should probably no longer be listed as a "modern API".


rickhanlonii (approved):
Nice catch

PR #7845, opened 2025-06-11
Fix tiny typo in handbook

irismessage:


rickhanlonii:
It's marginally better but not worth the translation churn IMO

PR #7849, opened 2025-06-15
Fixed broken placeholder links on manipulating dom with refs page

daniel-norris:


gaearon:
I think we should just host them ourselves and do a pass to remove any external image links from the docs. Would you mind taking a look at that?

I'd also prefer to keep cats personally. But if you do the above, I can take dogs.

gaearon:
One argument to keep cats is to avoid churning the translations.

daniel-norris:
> we should just host them

Sure ok, I can have a look at some options. Anything in particular you had in mind here? What about simply putting them in https://github.com/reactjs/react.dev/tree/main/public/images/docs/illustrations or similar? 

> One argument to keep cats is to avoid churning the translations.

That's fair enough.

PR #7865, opened 2025-07-11
docs: remove outdated caveat about useId in async Server Components

ohjooyeong:
The previous documentation stated that useId cannot be used in Server Components that are async functions.

However, as of React 19.1.0, useId now works correctly in async Server Components as well.

I have tested this behavior and confirmed that the caveat is no longer necessary.

This PR removes the outdated note to reflect the current behavior.

> Tested with:
> - next: 15.3.5
> - react: 19.1.0
> - react-dom: 19.1.0
>
> ```jsx
> // app/page.jsx
> import { useId } from "react";
>
> export default async function Page() {
> const id = useId();
> return (
> <div>
> <label htmlFor={id}>Name:</label>
> <input id={id} type="text" />
> <p>Generated id: <code>{id}</code></p>
> </div>
> );
> }
> ```
>
> This async Server Component renders without any error and the generated id works as expected.

### DIFF
--- a/src/content/reference/react/useId.md
+++ b/src/content/reference/react/useId.md
@@ -46,8 +46,6 @@ function PasswordField() {
 
 * `useId` **should not be used to generate keys** in a list. [Keys should be generated from your data.](/learn/rendering-lists#where-to-get-your-key)
 
-* `useId` currently cannot be used in [async Server Components](/reference/rsc/server-components#async-components-with-server-components).
-
 ---
 
 ## Usage {/*usage*/}


rickhanlonii:
It doesn't work if you do this:

```
// app/page.jsx
import { useId } from "react";

export default async function Page() {
  await something();
  const id = useId();
  return (...);
```

The error is in the linter only I believe.

rickhanlonii:
cc @eps1lon

PR #7866, opened 2025-07-11
Fix runtime error in a broken clock challenge caused by missing element check

soralvy:
### **What’s the issue?**
The Clock challenge component throws a runtime error because it tries to set the className on an element that does not yet exist in the DOM. Specifically, document.getElementById('time') returns null during the initial render, causing this error:

image

_This PR adds a simple existence check before accessing the element’s className, preventing the runtime error while preserving the original logic_


rickhanlonii:
@gaearon is the error intentional for the challenge? The text sounds like it shouldn't, but looking at the history I don't see how this would have ever worked without an error?

soralvy:
Any updates on that? @rickhanlonii @gaearon

PR #7868, opened 2025-07-11
[compiler] Rewrite React Compiler Docs

poteto:
We've received [feedback](https://bsky.app/profile/danabra.mov/post/3lr46ciujjs2r) that the compiler docs are difficult to understand and not prominent enough that people don't realize the compiler is a serious project and is near stable.

This PR rewrites the whole compiler doc section, giving it its own category as well as a standalone reference page.

Preview: https://react-dev-git-pr7868-fbopensource.vercel.app/

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/reactjs/react.dev/pull/7868).
* __->__ #7868
* #7883


src/content/reference/react/useMemo.md
```diff
@@ -12,6 +12,12 @@ const cachedValue = useMemo(calculateValue, dependencies)
 
 </Intro>
 
+<Note>
+
+[React Compiler](/learn/react-compiler) automatically memoizes values and functions, reducing the need for manual `useMemo` calls. You can use the compiler to handle memoization automatically.
```
gaearon, line 17: this is great, was just browsing the docs and surprised it wasn't there

src/content/learn/react-compiler-deep-dive-backup.md
```diff
@@ -0,0 +1,59 @@
...
+  );
+}
+```
+[_See this example in the React Compiler Playground_](https://playground.react.dev/#N4Igzg9grgTgxgUxALhAMygOzgFwJYSYAEAYjHgpgCYAyeYOAFMEWuZVWEQL4CURwADrEicQgyKEANnkwIAwtEw4iAXiJQwCMhWoB5TDLmKsTXgG5hRInjRFGbXZwB0UygHMcACzWr1ABn4hEWsYBBxYYgAeADkIHQ4uAHoAPksRbisiMIiYYkYs6yiqPAA3FMLrIiiwAAcAQ0wU4GlZBSUcbklDNqikusaKkKrgR0TnAFt62sYHdmp+VRT7SqrqhOo6Bnl6mCoiAGsEAE9VUfmqZzwqLrHqM7ubolTVol5eTOGigFkEMDB6u4EAAhKA4HCEZ5DNZ9ErlLIWYTcEDcIA)
+
+React Compiler automatically applies the equivalent of manual memoization, ensuring that only the relevant parts of an app re-render as state changes, which is sometimes referred to as "fine-grained reactivity". In the above example, React Compiler determines that the return value of `<FriendListCard />` can be reused even as `friends` changes, and can avoid recreating this JSX _and_ avoid re-rendering `<MessageButton>` as the count changes.
+
+#### Expensive calculations also get memoized {/*expensive-calculations-also-get-memoized*/}
+
+React Compiler can also automatically memoize for expensive calculations used during rendering:
```
acutmore:
```suggestion
React Compiler can also automatically memoize expensive calculations used during rendering:
```

src/content/learn/react-compiler/compatibility.md
```diff
@@ -0,0 +1,74 @@
...
+
+<YouWillLearn>
+
+* How to configure React Compiler for React 17 and 18
+
+</YouWillLearn>
+
+## Using React Compiler with React 17 or 18 {/*using-react-compiler-with-react-17-or-18*/}
+
+React Compiler works out of the box with React 19. For React 17 or 18, you need two things:
```
fabiendem: ❓ Question:
Could you please clarify the following:

> React Compiler works out of the box with React 19.

Does that mean that it's _enabled_ by default for anyone using React 19? (ie. if I use React 19 the compiler is running).
If so, can one disable it if they are using React 19?

"Works out of the box" can be interpreted in two ways:
- It works without extra setup, but you need to enable it
- It works without extra setup and it's enabled by default

The previous doc was saying:

> If you're starting a new project, you can enable the compiler on your entire codebase, which is the default behavior.

which was interpreted differently depending on who you ask

Thanks!

src/content/learn/react-compiler/getting-started.md
```diff
@@ -0,0 +1,256 @@
...
+
+Or with pnpm:
+
+<TerminalBlock>
+pnpm install -D babel-plugin-react-compiler@rc
+</TerminalBlock>
+
+## Basic Setup {/*basic-setup*/}
+
+React Compiler is designed to work out of the box without configuration. However, if you need to configure it (for example, to target React versions below 19), refer to the [compiler options reference](/reference/react/react-compiler).
```
fabiendem: ❓ Question:

Same question as above about React 19 and the compiler enablement

src/content/reference/react/react-compiler.md
```diff
@@ -0,0 +1,270 @@
...
+```js
+const ReactCompilerConfig = {
+  logger: {
+    logEvent(filename, event) {
+      console.log(`[Compiler] ${filename}: ${event.kind}`);
+    }
+  }
+};
+```
+
```
fabiendem: 💡 Suggestion:

This snippet helped us identify issues while adopting the Compiler
```
  logger: {
    logEvent: (filename, event) => {
      if (event.kind === "CompileSuccess") {
        console.log("✨ React Compiler successfully compiled: ", filename);
      }

      if (event.kind === "CompileError") {
        console.warn("⚠️ React Compiler failed to compile: " + filename);
        console.warn(filename, JSON.stringify(event, null, 2));
      }
    },
  },
```
This is useful in React Native, Metro will output the Compiler results while the JS bundle is being loaded into the app.

src/content/learn/react-compiler/compatibility.md
poteto: Good question! Since the compiler is a separate package, "It works without extra setup, but you need to enable it" is what I meant. I can make this clearer

src/content/learn/react-compiler-deep-dive-backup.md
poteto: Oops, didn't mean to commit this file

src/content/reference/react/react-compiler.md
```diff
@@ -0,0 +1,284 @@
...
+// Before compilation
+function MyComponent(props) {
+  return <div>{props.text}</div>;
+}
+
+// After compilation (simplified)
+import { isCompilerEnabled } from 'ReactCompilerFeatureFlag';
+import { c as _c } from 'react/compiler-runtime';
+
+const MyComponent = isCompilerEnabled()
```
GSTJ: If that's implemented at a component level and doesn't utilize dynamic imports, can we expect a big increase in bundle size?

As effectively, we are importing double the amount of code per optimized component?

I know it won't be exactly a 2x increase, as we aren't optimizing node_modules nor normal functions, but I wonder if this would negatively impact production in a significant way.

poteto: Yeah the number of lines per function does get doubled, but the bundle size increase isn't 1:1 with the number of lines since it depends on the compression algorithm. For us, we found that a React Native (Android) project saw an overall increase of ~10% in apk size.

There's some tradeoff math here to do, while bundle size may increase in the gating mode, you will likely see performance benefits in the `test` case that tradeoff against the bundle size. You'll need to run an experiment to really be sure. 

Of course the gating mode is optional too. The Incremental Adoption page covers a bit more about other strategies.

GSTJ: Got it, thanks @poteto!

src/content/learn/react-compiler/installation.md
```diff
@@ -0,0 +1,256 @@
...
+    }),
+  ],
+});
+```
+
+### Next.js {/*usage-with-nextjs*/}
+
+Please refer to the [Next.js docs](https://nextjs.org/docs/app/api-reference/next-config-js/reactCompiler) for more information.
+
+### Remix {/*usage-with-remix*/}
```
brookslybrand: Not sure if this is the appropriate place for this feedback, or if it should be a separate PR, but the Remix section should probably be switched to React Router

I just tested and the `vite-plugin-babel` works the exact same, just needs to be:

```
export default defineConfig({
  plugins: [
    reactRouter(),
    babel({
      filter: /\.[jt]sx?$/,
      babelConfig: {
        presets: ["@babel/preset-typescript"], // if you use TypeScript
        plugins: [
          ["babel-plugin-react-compiler", ReactCompilerConfig],
        ],
      },
    }),
  ],
});
```

src/content/learn/react-compiler/debugging.md
```diff
@@ -0,0 +1,93 @@
...
+**Runtime issues** occur when compiled code behaves differently than expected. Most of the time, if you encounter an issue with React Compiler, it's a runtime issue. This typically happens when your code violates the Rules of React in subtle ways that the compiler couldn't detect, and the compiler mistakenly compiled a component it should have skipped.
+
+When debugging runtime issues, focus your efforts on finding Rules of React violations in the affected components that were not detected by the ESLint rule. The compiler relies on your code following these rules, and when they're broken in ways it can't detect, that's when runtime problems occur.
+
+
+## Common Breaking Patterns {/*common-breaking-patterns*/}
+
+One of the main ways React Compiler can break your app is if your code was written to rely on memoization for correctness. This means your app depends on specific values being memoized to work properly. Since the compiler may memoize differently than your manual approach, this can lead to unexpected behavior like effects over-firing, infinite loops, or missing updates.
+
+Common scenarios where this occurs:
```
aleclarson, line 34: Another breaking pattern is when you reference a variable declaration inside a callback that was declared *after* the callback. This can happen accidentally when writing an especially complex component or hook. That callback won't be memoized by the compiler.

Would you consider this a common scenario?
https://playground.react.dev/#N4Igzg9grgTgxgUxALhAMygOzgFwJYSYAEAQgIYAmAFAJRHAA6xRchYORhAwgDZ5wBrIgF4itEQD56TIrJZsIPBADoeEAOZU0ECDRlEAvvv2tM7IgG1tEADREwCHADEdAXRFEoDgMo4yOBCoABj1MfRhHWGIAHgo8ADdOTF5+AWFgbj5BAwlgawNogHo4+IkmIzDMDGx8QiIAcR1qOkZmU3MrHTsHZzcPLwRff0CQkzYOTNSPcWEpVrl5M0UVNU1rUNkK8MiYGJKklMF0yezc-KKSsswDEAMgA

src/content/learn/react-compiler/incremental-adoption.md
```diff
@@ -0,0 +1,225 @@
...
+// babel.config.js
+module.exports = {
+  plugins: [],
+  overrides: [
+    {
+      test: './src/experimental/**/*.{js,jsx,ts,tsx}',
+      plugins: [
+        ['babel-plugin-react-compiler', {
+          compilationMode: 'annotation', // Only compile "use memo" components
+          panicThreshold: 'NONE' // More permissive for experimental code
```
henryqdineen:
```suggestion
          panicThreshold: 'none' // More permissive for experimental code
```

lowercased to match [the types](https://github.com/facebook/react/blob/dffacc7b8094576c19790fe8341996f743ba4a89/compiler/packages/babel-plugin-react-compiler/src/Entrypoint/Options.ts#L23-L36) and be more consistent with `src/content/reference/react-compiler/panicThreshold.md`

src/content/learn/react-compiler/incremental-adoption.md
```diff
@@ -0,0 +1,225 @@
...
+        }]
+      ]
+    },
+    {
+      test: './src/production/**/*.{js,jsx,ts,tsx}',
+      plugins: [
+        ['babel-plugin-react-compiler', {
+          panicThreshold: 'CRITICAL_ERRORS' // Stricter for production code
```
henryqdineen:
```suggestion
          panicThreshold: 'critical_errors' // Stricter for production code
```

src/content/learn/react-compiler/installation.md
poteto: oh yeah good catch, thanks!

src/content/learn/react-compiler/debugging.md
josephsavona: Yeah this is a fun case. Note that you can't memoize such callbacks yourself — adding a `useCallback()` in the "Bad" example in your playground link would be a TDZ violation when you added the dep on `foo`:

```js
function Bad() {
  const onClick = useCallback(() => {
    console.log(foo)
  }, [foo /* TDZ violation */ ]);
  
  const [foo, setFoo] = useState(0)

  return <div onClick={onClick}>{foo}</div>
}
```

We can look into adding linting against this pattern to suggest moving the callback. Note that the compiler will already reject this code if you try to manually memoize it, because the dependency changes later.

josephsavona: > Would you consider this a common scenario?

Since this pattern doesn't work with manual memoization, this pattern is most likely in code that isn't memoized already. Definitely worth adding validation to help identify opportunities in such code to memoize more.

src/content/learn/react-compiler/incremental-adoption.md
fabiendem, line 1: ❓ Question:

It may not belong to the "incremental adoption" section, but I wonder if it's worth answering the following question in the adoption plan or somewhere else:

There is a recurring question regarding the best practice for handling existing manual memoization APIs (useMemo, useCallback, memo) in a codebase after the React Compiler has been enabled. This uncertainty is common in discussions about adopting the compiler for large, legacy applications.

To address this, we could add a section to the official documentation, that answers the following:

FAQ: What is the recommended strategy for useMemo, useCallback, and memo in a large and legacy codebase after enabling the React Compiler?

Once the compiler is active, should teams begin a gradual process of removing these manual memoizations?
Our working assumption is that this is the intended outcome, as it aligns with the core purpose of the compiler.

Providing clear, official guidance on this topic will significantly reduce community uncertainty and help teams establish a consistent and effective adoption plan.

fabiendem, line 1: ❓ Question:

Is the use of `react-compiler-healthcheck` recommended at all?

poteto: Added: https://react-dev-git-pr7868-fbopensource.vercel.app/learn/react-compiler/introduction#what-should-i-do-about-usememo-usecallback-and-reactmemo

poteto: That package isn't being actively maintained right now so I'm hesitant to document it. You can run the eslint rule as an alternative to get the same information on which files have violations that were detected.

fabiendem: Sorry one more question, will `useCallback`, `useMemo`, and `memo` eventually be marked as deprecated?

eppisapiafsl:
👋 Have a question around the removal of `useCallback` declaration

1.  Could you confirm how the arrow function will behave after removing the `useCallback`? 

For performance improvement, it's recommended to use the [jsx-no-bind](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md) to memoize our function and prevent creating a new instance in every re-render

Do we have an equivalent of the rule for the React-compiler, or will it be `onPress={() => { some actions}}` automatically memoized?

src/content/learn/react-compiler/debugging.md
```diff
@@ -0,0 +1,93 @@
...
+function ProblematicComponent() {
+  "use no memo"; // Skip compilation for this component
+  // ... rest of component
+}
+```
+
+If the issue disappears, it's likely related to a Rules of React violation.
+
+You can also try removing manual memoization (useMemo, useCallback, memo) from the problematic component to verify that your app works correctly without any memoization. If the bug still occurs when all memoization is removed, you have a Rules of React violation that needs to be fixed.
+
```
eppisapiafsl, line 69: Adding my thoughts, I recommend separating the concepts of "transition" and "future vision" more clearly in the documents about the React Compiler.

Many developers are now removing these hooks preemptively, which can cause compatibility issues and potential regressions:

- If the callback is used in useEffect can trigger warnings from the exhaustive-deps-hooks rule.
- The compiler often indicates when code does not behave as expected. If developers remove memoization hooks without evaluating their impact, projects might experience unforeseen regressions. (This is on the developer side, not really an issue from the compiler)

From what I understand, the long-term goal is to remove the need for hooks such as useCallback, useMemo, and memo. However, unless I am mistaken, the current state does not require developers to remove these hooks immediately.

I know this is optional, but suggesting a formal transition plan will help developers and companies understand the work to do for a smooth transition. For example:

Step 1: Ensure code is fully compliant with the React Compiler.
Step 2: Once Meta officially announces support, remove implementations of useCallback, useMemo, and similar hooks (optionally, developers may do this earlier if they choose).

In theory, that info is already in the doc 😅, but I would centralize it in a single file

Could we clarify this distinction in the documentation? If my understanding is incorrect, I welcome input on the current expectations.

src/content/learn/react-compiler/incremental-adoption.md
poteto: We're still in the process of figuring out if we can safely deprecate manual memoization. It would be a nice DX win if we can make it happen. But to set expectations, it's better to assume that `useMemo`, `useCallback`, and `React.memo` will stay for the foreseeable future. However, with React Compiler most apps should be able to remove them, so it's possible that as the compiler gets adopted more broadly, these APIs will naturally become less commonplace. 

One reason that makes it difficult to just deprecate manual memoization is that apps today may rely on it for "correctness". That means not using memoization purely for performance, but to prevent infinite loops or overfiring effects. This is a bad reason to use manual memoization as it's likely hiding some latent bugs in your app. One way to test this in your app is by [codemodding away manual memoization](https://codemod.com/registry/react-19-remove-memoization) locally in dev. Your app will be slower, but it should not break.

There's a lot more nuances here that our team is currently exploring. When we finish our research on this topic, we will share our findings with the community.

poteto:
> 👋 Have a question around the removal of `useCallback` declaration
> 
> 1. Could you confirm how the arrow function will behave after removing the `useCallback`?
> 
> For performance improvement, it's recommended to use the [jsx-no-bind](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md) to memoize our function and prevent creating a new instance in every re-render
> 
> Do we have an equivalent of the rule for the React-compiler, or will it be `onPress={() => { some actions}}` automatically memoized?

https://playground.react.dev/ would be the best way to see for yourself. React Compiler does automatically memoize callbacks as well, including lambdas in JSX.

src/content/learn/react-compiler/debugging.md
poteto: This is mentioned in https://react-dev-git-pr7868-fbopensource.vercel.app/learn/react-compiler/introduction#what-should-i-do-about-usememo-usecallback-and-reactmemo. Let me know if that answers your question

eppisapiafsl: Nice! Thanks for sharing it @poteto ; it does 👍

PR #7869, opened 2025-07-11
Docs for partial-prerendering APIs

rickhanlonii:
The main section to focus on is in the `resume` docs: https://react-c6xxv66nl-fbopensource.vercel.app/reference/react-dom/server/resume#resuming-a-prerender

Resume-and-prerender APIs behave no different than resume with the exception that you can continue a prerender. So referring to resume/prerender docs for concrete examples makes more sense.

The Node.js APIs behave no different other than having a slightly different interface so we don't need to repeat usage examples. 

There's room for a more detailed learn series about partial prerendering that can be distilled from blog posts and talks going into infra details. But that is out of scope for reference docs.

### DIFF
--- a/src/content/reference/react-dom/server/index.md
+++ b/src/content/reference/react-dom/server/index.md
@@ -10,19 +10,20 @@ The `react-dom/server` APIs let you server-side render React components to HTML.
 
 ---
 
-## Server APIs for Node.js Streams {/*server-apis-for-nodejs-streams*/}
-
-These methods are only available in the environments with [Node.js Streams:](https://nodejs.org/api/stream.html)
-
-* [`renderToPipeableStream`](/reference/react-dom/server/renderToPipeableStream) renders a React tree to a pipeable [Node.js Stream.](https://nodejs.org/api/stream.html)
-
----
-
 ## Server APIs for Web Streams {/*server-apis-for-web-streams*/}
 
 These methods are only available in the environments with [Web Streams](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API), which includes browsers, Deno, and some modern edge runtimes:
 
 * [`renderToReadableStream`](/reference/react-dom/server/renderToReadableStream) renders a React tree to a [Readable Web Stream.](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream)
+* [`resume`](/reference/react-dom/server/renderToPipeableStream) resumes [`prerender`](/reference/react-dom/static/prerender) to a [Readable Web Stream](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream).
+---
+
+## Server APIs for Node.js Streams {/*server-apis-for-nodejs-streams*/}
+
+These methods are only available in the environments with [Node.js Streams:](https://nodejs.org/api/stream.html)
+
+* [`renderToPipeableStream`](/reference/react-dom/server/renderToPipeableStream) renders a React tree to a pipeable [Node.js Stream.](https://nodejs.org/api/stream.html)
+* [`resumeToPipeableStream`](/reference/react-dom/server/renderToPipeableStream) resumes [`prerenderToNodeStream`](/reference/react-dom/static/prerenderToNodeStream) to a pipeable [Node.js Stream.](https://nodejs.org/api/stream.html)
 
 ---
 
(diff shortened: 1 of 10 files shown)


src/content/reference/react-dom/server/resume.md
```diff
@@ -0,0 +1,240 @@
...
+
+## Reference {/*reference*/}
+
+### `resume(node, postponed, options?)` {/*resume*/}
+
+Call `resume` to resume rendering a pre-rendered React tree as HTML into a [Readable Web Stream.](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream)
+
+```js
+import { resume } from 'react-dom/server';
+import {getPostponedState} from 'storage';
```
gaearon: this looks like a real npm package, is this meant to be something like `readFile`?

rickhanlonii: Loading from the filesystem is probably not going to work right? Typically it would need to be like redis or s3 or something like that. 

What about `'./your-storge-layer'`

gaearon: that seems fine, my main point is just it’s hard to imagine what this is supposed to mean if the format isn’t clear (ie is this just JSON?)

i don’t see why it wouldn’t work via file system though

rickhanlonii: Yeah we'll clean it up to make it more clear cc @mattcarrollcode 

For the file system to work you'd need to serve both the prerender response and the resume request from the same server, which would be uncommon for prod usecases (you're probably scaled to multiple servers, or running vms that can be destroyed between requests, etc). And if you're serving the prerender result from a CDN, it's basically not possible/advisable because you'd need to sync the postpone state to your resume server somehow. So you probably need some kind of permanent storage layer.

samselikoff: Serving the prerender + request from the same server could make sense if you wanted to pre-render some dynamic components at build time, no? For example Remix could expose an API to pre-render some route segments that have dynamic `loaders()` (while deferring other segments' `loaders()` to execute at request-time).

What about something like this:

```js
import { resume } from 'react-dom/server';
import { storageClient } from './redis'; // or S3, a file, etc.

async function handler(request) {
  const postponed = await storageClient.get(request.url);
  const stream = await resume(<App />, postponed, {
    bootstrapScripts: ['/main.js']
  });

  return new Response(stream, {
    headers: { 'content-type': 'text/html' },
  });
}
```

Also `postponed` does have to be JSON, correct? Even though its shape is opaque?

eps1lon: > (ie is this just JSON?)

That's good feedback! We'll make it clear in the prerender docs that the postponed state is JSON-serializable.

src/content/reference/react-dom/server/index.md
```diff
@@ -10,19 +10,22 @@ The `react-dom/server` APIs let you server-side render React components to HTML.
...
+## Server APIs for Web Streams {/*server-apis-for-web-streams*/}
 
-These methods are only available in the environments with [Node.js Streams:](https://nodejs.org/api/stream.html)
+These methods are only available in the environments with [Web Streams](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API), which includes browsers, Deno, and some modern edge runtimes:
 
-* [`renderToPipeableStream`](/reference/react-dom/server/renderToPipeableStream) renders a React tree to a pipeable [Node.js Stream.](https://nodejs.org/api/stream.html)
+* [`renderToReadableStream`](/reference/react-dom/server/renderToReadableStream) renders a React tree to a [Readable Web Stream.](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream)
+* [`resume`](/reference/react-dom/server/renderToPipeableStream) resumes [`prerender`](/reference/react-dom/static/prerender) to a [Readable Web Stream](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream).
 
+Node.js also includes these methods for compatibility, but they are not recommended due to worse performance. Use the [dedicated Node.js APIs](#server-apis-for-nodejs-streams) instead.
```
rickhanlonii, line 23:
```suggestion

<Note>

Node.js also includes these methods for compatibility, but they are not recommended due to worse performance. Use the [dedicated Node.js APIs](#server-apis-for-nodejs-streams) instead.

</Note>
```

rickhanlonii: fixes this:

Screenshot 2025-10-01 at 8 22 30 AM

src/sidebarReference.json
```diff
@@ -319,9 +329,19 @@
           "title": "prerender",
           "path": "/reference/react-dom/static/prerender"
         },
+        {
+          "title": "resumeAndPrerender",
+          "path": "/reference/react-dom/static/resumeAndPrerender",
+          "version": "canary"
+        },
```
rickhanlonii, line 340:
```suggestion
```

src/sidebarReference.json
```diff
@@ -319,9 +329,19 @@
...
         {
           "title": "prerenderToNodeStream",
           "path": "/reference/react-dom/static/prerenderToNodeStream"
+        },
+        {
```
rickhanlonii:
```suggestion
        },
        {
          "title": "resumeAndPrerender",
          "path": "/reference/react-dom/static/resumeAndPrerender",
          "version": "canary"
        },
        {
```

Issue #7872, opened 2025-07-14
Bug Report: Blank Screen Issue in React Documentation


aakashpalasiya01:
### Summary

Description:
When browsing the React documentation, the page works fine initially. However, after spending some time scrolling or reading the content, the page goes blank. To restore the content, a manual reload is required. This issue has occurred more than 20 times.

Steps to Reproduce:

Open the React documentation page.

Scroll or read the content for some time.

The page turns blank after a while.

Refresh the page to restore the content.

Frequency: More than 20 times.

### Page

https://react.dev/learn/thinking-in-react

### Details

https://drive.google.com/file/d/1xGqxOGxiJCff_Z2jj4pNw6seeSD8zCYJ/view?usp=drive_link

Shiv10122003:
Hey please assign the issue to me I will look after it

rickhanlonii:
@Shiv10122003 we don't assign tasks, we'll review the first PR someone submits.

Shiv10122003:
Okk sir then I will come with a pr

Shiv10122003:
Sir please review

vickrant-dev:
Unable to reproduce the blank screen crash you have mentioned. The PR is not merged right? @Shiv10122003

Shiv10122003:
Right it is not merged

vickrant-dev:
I don't think a merge is required here, as I am unable to reproduce this issue. The page is working totally fine without any blank screen.

Shiv10122003:
Yaa but I have added error boundary if there is any other kinds of error . You may go through the code

vickrant-dev:
I did go through the code, but then is it necessary to have an error boundary when no problem exists. Are you able to reproduce this issue where the screen goes blank after using it for few minutes?

Shiv10122003:
So talk to organisers to close this issue .

Shiv10122003:
I will close my pr

vickrant-dev:
Organizers will proly take care of closing this. 
@rickhanlonii sry for direct mentioning but there is no blank screen issue as mentioned on ur side right? (I don't hv on my side) Commenting this to close if no issue exists. Thank you : )

rickhanlonii:
I think we need @aakashpalasiya01 to confirm whether they're still seeing it, I'm not able to reproduce. I'll leave this open in case it's still an issue.

aakashpalasiya01:
hiii @rickhanlonii 
I’m not facing this issue anymore.

PR #7873, opened 2025-07-16
Add global ErrorBoundary to prevent blank screen crashes (#7872)

Shiv10122003:
➕ Created src/components/ErrorBoundary.tsx

A class-based boundary using React's error lifecycle

Renders a fallback UI and logs errors via console.error

🌀 Wrapped <Component /> in _app.tsx with <ErrorBoundary>

Ensures all page-level and nested errors are caught gracefully

### DIFF
--- a/src/components/ErrorBoundary.tsx
+++ b/src/components/ErrorBoundary.tsx
@@ -1,0 +1,38 @@
+"use client";
+import { Component, ReactNode } from "react";
+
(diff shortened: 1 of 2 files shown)


rickhanlonii:
Is this AI? We're not even using the app router or RSCs. 

We don't need this because all the content is built at build time.

PR #7904, opened 2025-08-06
Fixes animation jank on hover

wesbos:
Small little fix to the community photos transition. transition-all caused the page to drop frames when painting when we really just want to transition the transform (scale) 

Video detailing the problem here: https://x.com/wesbos/status/1953183430306611280


gaearon:
ah thank you! i’ve been wondering what this is. at some point in the past i think it was fast so not sure how it regressed

PR #7905, opened 2025-08-07
Fix: box-shadow animation on community photos transition

ChezFre:
The change from https://github.com/reactjs/react.dev/pull/7904 broke the fade transition on the box-shadow.

The original code added a box-shadow on hover, triggering a repaint from the browser. With these changes the box-shadow is always there with an opacity of 0 and fades in on hover.

Because the hidden overflow I had to add another wrapping element to prevent the pseudo element's box-shadow from being cut off.

Also fixes another issue where `overflow-x: hidden` was combined with `overflow-y: visible`, which doesn't work causing the `box-shadow` to be cut off in Firefox.

### DIFF
--- a/src/components/Layout/HomeContent.js
+++ b/src/components/Layout/HomeContent.js
@@ -755,9 +755,7 @@ function CommunityGallery() {
   }, []);
 
   return (
-    <div
-      ref={ref}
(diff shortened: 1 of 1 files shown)


gaearon:
thanks!

PR #7908, opened 2025-08-08
docs: add JSX rule about function references as children

elkanatum:
## Summary
Adds documentation explaining the common mistake of passing function references as JSX children instead of calling them.

## Changes
- Added new section "4. Don't pass function references as children" to JSX rules in `writing-markup-with-jsx.md`
- Included code examples showing incorrect and correct usage
- Addresses issue facebook/react#34007

## Type of Change
- [x] Documentation update

This addresses a common beginner mistake where developers accidentally pass function references as JSX children, resulting in `[Function]` being displayed instead of meaningful content.

## Related Issue
Addresses facebook/react#34007

### DIFF
--- a/src/content/learn/writing-markup-with-jsx.md
+++ b/src/content/learn/writing-markup-with-jsx.md
@@ -222,6 +222,32 @@ For historical reasons, [`aria-*`](https://developer.mozilla.org/docs/Web/Access
 
 </Pitfall>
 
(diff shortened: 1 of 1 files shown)


rickhanlonii:
This isn't the right place for this. I think we could add an example or challenge to the next page JavaScript in JSX with Curly Braces instead. That page currently has a challenge using an object, but doesn't mention functions.

PR #7915, opened 2025-08-15
Fix usememo initial render wording

### LINKED ISSUE #7752: [Typo]: explanation of useMemo first-render return value, doesn't make logical sense
### Summary

Inconsistent explaining on what useMemo returns on first render.

### Page

https://react.dev/reference/react/useMemo#reference

### Details

From article, in order:
1. "On the initial render, useMemo returns the result of calling calculateValue _with no arguments._"
2. "On the initial render, the value you’ll get from useMemo will be the result of calling your calculation."

The first sentence, to me does not make sense, as calculateValue is called _with_ arguments, it's just that dependencies don't factor in yet on first render. 
Or my bad sorry:)

Raghuboi:
Fixes #7752

### DIFF
--- a/src/content/reference/react/useMemo.md
+++ b/src/content/reference/react/useMemo.md
@@ -50,7 +50,7 @@ function TodoList({ todos, tab }) {
 
 #### Returns {/*returns*/}
 
-On the initial render, `useMemo` returns the result of calling `calculateValue` with no arguments.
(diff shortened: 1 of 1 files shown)


rickhanlonii:
No, it's important to know that the `calculateValue` argument is never called with arguments.

Issue #73, opened 2017-10-08
Add ES5 example to Tutorial


bvaughn:
This issue was originally reported by @quadruplejump via facebook/react/issues/10147

### Text of original issue:

The tutorials, with the exception of the "advanced guides", assume ES6/JSX.

The React without ES6 tutorial https://facebook.github.io/react/docs/react-without-es6.html still uses JSX.

The React without JSX tutorial https://facebook.github.io/react/docs/react-without-jsx.html still uses ES6.

What would a simple ES5 React demo look like?

To compare, the Vue introduction https://vuejs.org/v2/guide/ starts with very simple examples that you can just drop in a webpage.

### Note from Dan

We can probably add an intro paragraph to both of these pages that links to HTML page with minimal example. 

### Note from Brian

Pending the outcome of #122, it might be nice to host the new minimal example in the repo and display it in CodeSandbox/Codepen.

M-ZubairAhmed:
Hi,
i have made this simple [gist](https://gist.github.com/M-ZubairAhmed/43a548bd14d068a263cd8327f48d36c8) to demonstrate simple react without jsx or es6 by following similar approach of 'getting started with Vue.js" guide. 
Is this something you are looking for?

quadruplejump:
@M-ZubairAhmed that would've been helpful months ago :D Do you need the code in the JS panel of the pen? render is being called twice with no obvious reason why

M-ZubairAhmed:
yes you are right I don't need one more render method. So does this solve the issue ? or should I include anything more.

isaacchien:
Is there anything else that needs to be done here? I'd be happy to finish it.

brettdewoody:
@bvaughn Is this issue still in need of being completed? While I think some would find a non-ES6/JSX demo useful, it's now 2 years on since this issue was raised and ES6 is much more standard.

davidbarratt:
It would be really helpful to see an ES5 example that uses functional components.

MiryamMann:
Hi, I'd like to contribute to this issue. Are there any remaining items that I can work on?

rickhanlonii:
Closing as outdated and it's 2025 so not sure and es5 example is useful.

PR #7922, opened 2025-08-17
Add ES5 React example for beginners (#73)

riyak145:
Add ES5 React example for beginners (#73)

Added a simple ES5 React example under examples/es5/es5-example.html for beginners.
It uses React.createElement and a plain ES5 function component to render "Hello, React ES5!".
This helps beginners see React without ES6 or JSX.

### DIFF
--- a/examples/es5/es5-example.html
+++ b/examples/es5/es5-example.html
@@ -1,0 +1,23 @@
+<!DOCTYPE html>
+<html>
+<head>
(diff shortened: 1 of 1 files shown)


rickhanlonii:
Is this AI generated? The page isn't linked to at all?

I closed the original issue and I don't think we need this anyway.

PR #7929, opened 2025-08-20
fix(docs): remove redundant section describing caveats of profiling

yeskunall:
Noticed that the caveat is rewritten as a pitfall later on in the docs. I chose to keep the latter as I felt it was redundant to keep both. LMK if there’s a reason both exist.

---

💖


rickhanlonii:
Let's keep both, this cannot be emphasized enough IMO

Issue #7913, opened 2025-08-14
[Typo]: Typographic bug: both quotes render as closing (”) around “Mirroring” on Choosing the State Structure page


1711naveen:
### Description: 
The word Mirroring is displayed with two closing curly quotes (”Mirroring”) instead of an opening curly quote followed by a closing curly quote.

### Expected:
“Mirroring” (U+201C opening curly quote, then U+201D closing curly quote)

### Actual:
”Mirroring” (U+201D closing curly quote on both sides)

### Page

https://react.dev/learn/choosing-the-state-structure

### 

<img width="930" height="411" alt="Image" src="https://github.com/user-attachments/assets/93ede237-42b6-497e-bd66-1db9f08d401f" />

Here’s a concise 30-word version:

On the “Choosing the State Structure” page, “Mirroring” appears with two closing curly quotes instead of opening and closing. Expected: `“Mirroring”`. Actual: `”Mirroring”`. Likely typography plugin misclassification.

rickhanlonii:
I think this is a bug in the markdown to HTML converter somewhere, or a missing plugin. We shouldn't need to use the opening and closing quotes any time we want to quote something in markdown, the parser should insert the right one.

PR #7931, opened 2025-08-20
Update choosing-the-state-structure.md

mimosapudical:
# [Typo] Fix Mirroring quotes on Choosing the State Structure page #7913

Corrects a typographic issue where “Mirroring” appeared with two closing curly quotes (”Mirroring”) instead of opening and closing quotes (“Mirroring”).


rickhanlonii (changes requested):
I think this is a bug in the markdown to HTML converter somewhere, or a missing plugin. We shouldn't need to use the opening and closing quotes any time we want to quote something in markdown, the parser should insert the right one.

PR #7937, opened 2025-08-22
Fix: Improve phrasing /error/index.md

WuMingDao:
image


rickhanlonii:
This is down to personal preference and not worth the translation churn.

PR #7943, opened 2025-08-22
feat(i18n): make DocsPageFooter navigation labels translatable

mrbadri:
Summary

This PR updates the DocsPageFooter component to support translation of navigation labels (“Previous” and “Next”).

Before
	•	Labels for navigation were hardcoded ({type}), preventing proper i18n.
	•	Could not override or translate them based on locale.

After
	•	Labels are now conditionally rendered:
	`{type === 'Previous' ? 'Previous' : 'Next'}`


rickhanlonii (approved):
It would be nice if all the translatable text in JS was moved to a single file so they can all be translated in one place. So this would be something like:

```js
import i18n from './i18n';


// ...
{type === 'Previous' ? i18n.previous : i18n.next}
```

There are libraries that do this, but this use case is simple enough I think we can just implement something manually.

Issue #7932, opened 2025-08-20
[Suggestion]: Change recommended React linting plugin


thernstig:
### Summary

Change recommended React linting config [eslint-config-react-app](https://www.npmjs.com/package/eslint-config-react-app) to [eslint-plugin-react](https://www.npmjs.com/package/eslint-plugin-react).

### Page

https://react.dev/learn/editor-setup

### Details

This text probably recommends the wrong config:

> Make sure that you’ve enabled all the [eslint-plugin-react-hooks](https://www.npmjs.com/package/eslint-plugin-react-hooks) rules for your project. They are essential and catch the most severe bugs early. The recommended [eslint-config-react-app](https://www.npmjs.com/package/eslint-config-react-app) preset already includes them.

If using the new ESLint flat config format, it is better to use the configs included with https://www.npmjs.com/package/eslint-plugin-react.

Essentially I think the paragraph should recommend both [eslint-plugin-react](https://www.npmjs.com/package/eslint-plugin-react) and [eslint-plugin-react-hooks](https://www.npmjs.com/package/eslint-plugin-react-hooks).

rickhanlonii:
cc @poteto, I'm not super familiar with the differences between the two.

Maybe instead of "recommending" one of the configs, we can just say:

> presets like `eslint-config-react-app` and `eslint-plugin-react` already include them.

thernstig:
https://www.npmjs.com/package/eslint-config-react-app is just outdated, and [eslint-plugin-react](https://www.npmjs.com/package/eslint-plugin-react) and [eslint-plugin-react-hooks](https://www.npmjs.com/package/eslint-plugin-react-hooks) are the ones to use. It aligns with that create-react-app is not to be used anymore, but rather a framework or similar. Next.js only has the two latter. Also eslint-config-react-app and the usage example is deprecated in the latest versions of ESLint and will even be removed in the next version.

rickhanlonii:
Yeah makes sense to not recommend the outdated plugin, but a lot of users are still using it so it's good to note that it supports eslint-plugin-react-hooks out of the box.

josephsavona:
We should definitely update this to remove eslint-config-react-app. The primary linter that we recommend is eslint-plugin-react-hooks — in the upcoming major version it includes multiple critical correctness checks from React Compiler that all developers should use. 

[...]

PR #7945, opened 2025-08-23
Update editor-setup.md

Dejumo:
Update editor setup docs to recommend eslint-plugin-react and eslint-plugin-react-hooks

#7932

### DIFF
--- a/src/content/learn/editor-setup.md
+++ b/src/content/learn/editor-setup.md
@@ -36,7 +36,7 @@ Code linters find problems in your code as you write, helping you fix them early
 * [Install ESLint with the recommended configuration for React](https://www.npmjs.com/package/eslint-config-react-app) (be sure you have [Node installed!](https://nodejs.org/en/download/current/))
 * [Integrate ESLint in VSCode with the official extension](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint)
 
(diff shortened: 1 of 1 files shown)


src/content/learn/editor-setup.md
```diff
@@ -36,7 +36,7 @@ Code linters find problems in your code as you write, helping you fix them early
...
 * [Integrate ESLint in VSCode with the official extension](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint)
 
-**Make sure that you've enabled all the [`eslint-plugin-react-hooks`](https://www.npmjs.com/package/eslint-plugin-react-hooks) rules for your project.** They are essential and catch the most severe bugs early. The recommended [`eslint-config-react-app`](https://www.npmjs.com/package/eslint-config-react-app) preset already includes them.
+**Make sure that you've enabled all the [`eslint-plugin-react-hooks`](https://www.npmjs.com/package/eslint-plugin-react-hooks) rules in your project.** They are essential and catch the most severe bugs early. If you’re setting up linting manually, we recommend using both [`eslint-plugin-react`](https://www.npmjs.com/package/eslint-plugin-react) and [`eslint-plugin-react-hooks`](https://www.npmjs.com/package/eslint-plugin-react-hooks). If you’re using the new ESLint flat config format, you can extend the configs included with those plugins.
```
rickhanlonii, line 39: See my comment in the issue - I don't think we should recommend one or the other. The goal here is to inform users that if they are using one of the config presets, it's already included.

Issue #7921, opened 2025-08-17
[Suggestion]: Add missing onChange handlers to input elements in server-functions.md documentation examples


rowleks:
### Summary

Add missing onChange handlers to input elements in the documentation examples for Server Functions, especially in server-functions.md. This will make the examples more accurate and prevent confusion for users learning about controlled inputs and. also make the code example valid

### Page

https://react.dev/reference/rsc/server-functions or directly in src/content/reference/rsc/server-functions.md

### Details

Some input elements in the Server Functions documentation (notably in the UpdateName component) currently lack onChange handlers. This makes the examples misleading, as users cannot update the value of these controlled inputs. Adding the appropriate onChange handlers will clarify correct usage and improve the learning experience.

Example code improvement:

Expected:
```js
<input type="text" name="name" value={name} onChange={e => setName(e.target.value)} />
```

Actual:
```js
<input type="text" name="name" />
```

Updating these examples will help reinforce best practices for controlled inputs in React.

rickhanlonii:
Mentioned in the PR, but the current text is correct. Controlled input are not necessary, and the example given uses an uncontrolled form, so the action receives the form data as an argument. See the docs on [Form `action` props](https://react.dev/reference/react-dom/components/form#props) for more info.

PR #7946, opened 2025-08-23
Update server-functions.md

Dejumo:
Fix: Add missing onChange to input in server-functions.md examples

#7921


rickhanlonii:
The current text is correct. Controlled input are not necessary, and the example given uses an uncontrolled form, so the action receives the form data as an argument. See the docs on [Form `action` props](https://react.dev/reference/react-dom/components/form#props) for more info.

PR #7951, opened 2025-08-26
Typographic bug: both quotes render as closing (”) around “Mirroring”…

AdiyaMitra:
Fix for bug #7913  : Typo on the "Choosing the State Structure" page.

This PR addresses the typographic bug where the opening and closing quotes around the word "Mirroring" were both rendering as closing quotes.

I have corrected the characters to ensure the opening quote is a proper typographical open quote (`“`) and the closing quote is a proper typographical close quote (`”`), as intended.

---

### Before the fix

The text rendered incorrectly as:
image

### After the fix

The text now renders correctly as:
image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,14934 @@
+{
+  "name": "react-dev",
+  "version": "1.0.0",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "react-dev",
+      "version": "1.0.0",
+      "hasInstallScript": true,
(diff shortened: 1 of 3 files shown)


rickhanlonii:
Dupe of https://github.com/reactjs/react.dev/pull/7931.

I commented on that PR that did a similar fix:

> I think this is a bug in the markdown to HTML converter somewhere, or a missing plugin. We shouldn't need to use the opening and closing quotes any time we want to quote something in markdown, the parser should insert the right one.

PR #7953, opened 2025-08-28
[compiler] Tweak intro section on manual memo guidance

poteto:
The previous sentence "If you are using React Compiler, useMemo, useCallback, and React.memo can be removed." was coming off a bit too strong and makes it incorrectly seem like the manual memos and compiler memos are 1:1.

Removing the sentence doesn't take anything away from this paragraph, so let's remove it to reduce confusion.


josephsavona (approved):
Looks great!

Issue #7848, opened 2025-06-15
[Bug]: "Reset state" on the playgrounds UX is confusing


NicolasRoelandt:
### Summary

If you want to reset the React state of a playground, but not the code, it's very unclear how to do so.
The only way I found is clicking on "Reset", which triggers a popup saying "Reset all your edits too?" and then pressing "Cancel". This will successfully reset the state but not the code. However there are two issues with that:
1. It's not clear at all that this is the proper way to do so. Should there be two buttons instead?
2. UX itself is counter-intuitive: 
     a. The question is "Reset all your edits too?", which is a yes/no question, but the options are "Ok" or "Cancel"
     b. Normally if you press "Cancel" on a confirmation modal, it should cancel the original action altogether, but here is applies the action partially (reset the state but not the edits)

 

### Page

https://react.dev/learn/separating-events-from-effects#challenges

vickrant-dev:
**Bug Confirmed**

I did notice the bug in the app and have worked on it and have submitted. Once it's approved, it will be all good to go.

PR #7954, opened 2025-08-28
Add reload button, rename reset to clear

### LINKED ISSUE #4776: Reset Button beta.react.org
Add a reset Button to the UI section in the Challenge some times we need only to reset the UI but not the code because the reset that exist in the challenge reset both the code and the UI you loss all the work you make is this any chance that you can add this feature;

- I like to add an invisible button reset to the right top corner of UI Screen and get Visible only when you hover or focus within it and thank you so much

rickhanlonii:
## Overview

Closes #4776 #7848

The "reset" button is weird. 

Screenshot 2025-08-28 at 3 13 42 PM

When you click it, it alerts with "Do you want to reset your edits too"
- Cancel - only reload
- Ok - clear edits and reload

That would be fine I guess if we just relabeled the buttons, but when you're making edits, it's actually annoying to have to keep clicking "cancel". 

So this PR separates the two into different buttons:
- Reload: reload the iframe "page" (no confirmation)
- Clear: clear all edits and reload (with confirmation)

https://github.com/user-attachments/assets/50840ee3-9dd5-4caa-b9f3-29ef38f2cab2

On mobile we'll now collapse to all icons:

Screenshot 2025-08-28 at 3 12 29 PM

### DIFF
--- a/next-env.d.ts
+++ b/next-env.d.ts
@@ -1,5 +1,6 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
+/// <reference types="next/navigation-types/compat/navigation" />
 
 // NOTE: This file should not be edited
-// see https://nextjs.org/docs/pages/api-reference/config/typescript for more information.
+// see https://nextjs.org/docs/app/api-reference/config/typescript for more information.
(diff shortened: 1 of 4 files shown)


next-env.d.ts
```diff
@@ -1,5 +1,6 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
+/// <reference types="next/navigation-types/compat/navigation" />
```
rickhanlonii, line 3: image

PR #7958, opened 2025-08-29
refactor: remove unused import in scripts

noritaka1166:
remove unused import in scripts

add [eslint-plugin-unused-imports](https://www.npmjs.com/package/eslint-plugin-unused-imports)


noritaka1166:
> Yeah nice catch. But each PR we do like this needs to open be sync'd to all the translation repos which creates churn for one-offs. Instead, can you add a linter for this and fix all of the failing cases in one PR?

@rickhanlonii 
I added `eslint-plugin-unused-imports` and ran `yarn run lint:fix`.

Issue #7734, opened 2025-04-11
[Typo]: Should server functions be wrapped in transitions or not?  (conflicting info in the docs)


gmoniava:
### Summary

[Here ](https://react.dev/reference/rsc/use-server#calling-a-server-function-outside-of-form) it says:

> When using a Server Function outside a [form](https://react.dev/reference/react-dom/components/form), call the Server Function in a [Transition](https://react.dev/reference/react/useTransition), which allows you to display a loading indicator, show [optimistic state updates](https://react.dev/reference/react/useOptimistic), and handle unexpected errors

Which seems to suggest we should **always** wrap server functions in transitions when using them outside forms. 

However [here ](https://react.dev/reference/rsc/server-functions#creating-a-server-function-from-a-server-component) is example from docs **without** wrapping them in transitions:

```
// Server Component
import Button from './Button';

function EmptyNote () {
  async function createNoteAction() {
    // Server Function
    'use server';
    
    await db.notes.create();
  }

  return <Button onClick={createNoteAction}/>;
}
```

```
// Client 
"use client";

export default function Button({onClick}) { 
  console.log(onClick); 
  // {$$typeof: Symbol.for("react.server.reference"), $$id: 'createNoteAction'}
  return <button onClick={() => onClick()}>Create Empty Note</button>
}
```

Can docs make it clearer which to use?

### Page

https://react.dev/reference/rsc/use-server#calling-a-server-function-outside-of-form

### Details

The two information seems to be conflicting.

gmoniava:
@poteto Sorry for direct tag - but would be nice if someone can have a look

rickhanlonii:
Yeah good question. The recommendation is to call it in a transition, but it's not necessary. There was a common misunderstanding that Server Functions needed to be actions (which is why we renamed the concept away from Server Action). 

The example in the docs that don't use a transition is showing that it doesn't need to be in a transition, but it's probably a bad example because that case should be a transition. That page includes examples with and without transitions to show both.

So I'd recommend that we:
- Update the language [here](https://react.dev/reference/rsc/use-server#calling-a-server-function-outside-of-form) to frame it more as a recommendation. It's also weird IMO to frame the two use cases as "in a form" and "outside a form". I would structure it as:
  - Calling a server function
    - server functions can be called directly
    - we recommend server functions that will update the UI to be called in an action
  - Calling a server function in an action
    - show using the action prop pattern to show a pending state automatically
  - Calling a server function in a form action
[...]

PR #7967, opened 2025-09-03
Fix handleClick compiler intro example

rickhanlonii:
Found by @gmoniava in https://github.com/reactjs/react.dev/issues/7734#issuecomment-3247727025

In the current version, passing an arrow function to the child breaks manual memoization, which means it's not an "equivalent" memoization like the text says. 

This is actually a subtle bug that the compiler prevents - the compiler output for this _does_ memoize correctly on `onClick` even with the function. That may be a good example to call out if we want to add to these docs.


src/content/learn/react-compiler/introduction.md
```diff
@@ -50,6 +50,21 @@ const ExpensiveComponent = memo(function ExpensiveComponent({ data, onClick }) {
...
+
+This manual memoization has a subtle bug that breaks memoization:
+
+```js [[2, 1, "() => handleClick(item)"]]
+<Item key={item.id} onClick={() => handleClick(item)} />
+```
+
+Even though `handleClick` is wrapped in `useCallback`, the arrow function `() => handleClick(item)` creates a new function every time the component renders. This means that `Item` will always receive a new `onClick` prop, breaking memoization.
+
+The React Compiler is able to optimize this correctly with or without the arrow function, ensuring that `Item` only re-renders when `props.onClick` changes.
```
josephsavona: nit: drop the "the", just React Compiler

josephsavona (approved):

poteto (approved):

src/content/learn/react-compiler/introduction.md
rickhanlonii:
```suggestion
React Compiler is able to optimize this correctly with or without the arrow function, ensuring that `Item` only re-renders when `props.onClick` changes.
```

PR #7968, opened 2025-09-03
Remove deprecated loremflickr placeholder images

### LINKED ISSUE #7966: [Bug]: photos in ref section of the tutorial are down (again)
### Summary

Previous issue: #6758

https://placekitten.com/g/200/200

We should just host the images ourselves at this point.

### Page

https://react.dev/learn/referencing-values-with-refs, https://react.dev/learn/manipulating-the-dom-with-refs

### Details

_No response_

poteto:
The previous examples used a service called loremflickr.com to link to  placeholder images. Since that service is now down, I updated them to reference placecats.com instead, which we already use in other examples as well. placecats doesn't have the same random cat feature so I roughly approximated it in the example code.

Closes #7966

### DIFF
--- a/src/content/learn/manipulating-the-dom-with-refs.md
+++ b/src/content/learn/manipulating-the-dom-with-refs.md
@@ -247,13 +247,13 @@ export default function CatFriends() {
       <nav>
         <button onClick={() => scrollToCat(catList[0])}>Neo</button>
         <button onClick={() => scrollToCat(catList[5])}>Millie</button>
-        <button onClick={() => scrollToCat(catList[9])}>Bella</button>
+        <button onClick={() => scrollToCat(catList[8])}>Bella</button>
       </nav>
       <div>
         <ul>
-          {catList.map((cat) => (
+          {catList.map((cat, index) => (
             <li
-              key={cat}
+              key={`${cat}?${index}`}
               ref={(node) => {
                 const map = getMap();
                 map.set(cat, node);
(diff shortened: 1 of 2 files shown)


src/content/learn/manipulating-the-dom-with-refs.md
```diff
@@ -275,9 +275,14 @@ export default function CatFriends() {
 function setupCatList() {
   const catList = [];
   for (let i = 0; i < 10; i++) {
-    catList.push("https://loremflickr.com/320/240/cat?lock=" + i);
+    if (i < 5) {
+      catList.push("https://placecats.com/neo/320/240");
```
rickhanlonii: We probably shouldn't break the key in the example. Instead, this can push an object like `{id: 1, src: 'https://...'}`, or you could add an index param to the URL like in the previous version (like `https://placecats.com/neo/320/240?id=1`).

rickhanlonii: Looks like we do both options in other examples

rickhanlonii (approved):
LGTM just a note on the key

src/content/learn/manipulating-the-dom-with-refs.md
poteto: yeah I noticed we were implicitly relying on the index previously so I thought to just keep it instead of fixing it. But you're right, let me fix that too

poteto: fixed to use an object with an id!

PR #7979, opened 2025-09-11
Update suspense notes

rickhanlonii:
These notes were not fully updated when `use(promise)` was released. Now that there's an official way to suspend, we don't need so many caveats and scary warnings to use a data fetching library. It's just:

> For Suspense you need to suspend, and here are the options

### DIFF
--- a/src/content/reference/react/Activity.md
+++ b/src/content/reference/react/Activity.md
@@ -870,17 +870,9 @@ Pre-rendering components with hidden Activity boundaries is a powerful way to re
 
 <Note>
 
-**Only Suspense-enabled data sources will be fetched during pre-rendering.** They include:
+**Activity does not fetch data inside an Effect during pre-render.** 
 
-- Data fetching with Suspense-enabled frameworks like [Relay](https://relay.dev/docs/guided-tour/rendering/loading-states/) and [Next.js](https://nextjs.org/docs/app/building-your-application/routing/loading-ui-and-streaming#streaming-with-suspense)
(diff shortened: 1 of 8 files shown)


src/content/reference/react/Activity.md
```diff
@@ -870,17 +870,9 @@ Pre-rendering components with hidden Activity boundaries is a powerful way to re
...
-- Data fetching with Suspense-enabled frameworks like [Relay](https://relay.dev/docs/guided-tour/rendering/loading-states/) and [Next.js](https://nextjs.org/docs/app/building-your-application/routing/loading-ui-and-streaming#streaming-with-suspense)
-- Lazy-loading component code with [`lazy`](/reference/react/lazy)
-- Reading the value of a cached Promise with [`use`](/reference/react/use)
-
-Activity **does not** detect data that is fetched inside an Effect.
-
-The exact way you would load data in the `Posts` component above depends on your framework. If you use a Suspense-enabled framework, you'll find the details in its data fetching documentation.
-
-Suspense-enabled data fetching without the use of an opinionated framework is not yet supported. The requirements for implementing a Suspense-enabled data source are unstable and undocumented. An official API for integrating data sources with Suspense will be released in a future version of React. 
+Since Effects are not mounted, only data fetched in render using Suspense will be pre-rendered.
```
slorber, line 875: But we can still warm up caches to reduce waterfalls right?

I mean calling React Query `queryClient.prefetchQuery()` while pre-rendering for example

PR #7980, opened 2025-09-11
docs: fix StrictMode cat scrolling example

kenkam:
### Steps to Reproduce

In the correct example in [StrictMode](./src/content/reference/react/StrictMode.md) _Fixing bugs found by re-running ref callbacks in development_ section, click on Millie to change cats, then click any number that is not 0.

### Expected

Should scroll to relevant `<li>`.

### Actual

Runtime error :crying_cat_face:


rickhanlonii (approved):

rickhanlonii:
Nice catch

PR #7986, opened 2025-09-15
Add new eslint rule reference docs

poteto:
Adds new docs for our new eslint rules.

### DIFF
--- a/src/content/reference/eslint-plugin-react-hooks/lints/error-boundaries.md
+++ b/src/content/reference/eslint-plugin-react-hooks/lints/error-boundaries.md
@@ -1,0 +1,72 @@
+---
+title: error-boundaries
+---
+
+<Intro>
+
+Validates usage of Error Boundaries instead of try/catch for errors in child components.
+
+</Intro>
+
+## Rule Details {/*rule-details*/}
+
+Try/catch blocks can't catch errors that happen during React's rendering process. Errors thrown in rendering methods or hooks bubble up through the component tree. Only [Error Boundaries](/reference/react/Component#catching-rendering-errors-with-an-error-boundary) can catch these errors.
+
+### Invalid {/*invalid*/}
+
+Examples of incorrect code for this rule:
+
+```js {expectedErrors: {'react-compiler': [4]}}
+// ❌ Try/catch won't catch render errors
+function Parent() {
+  try {
+    return <ChildComponent />; // If this throws, catch won't help
+  } catch (error) {
+    return <div>Error occurred</div>;
+  }
+}
+```
+
+### Valid {/*valid*/}
+
+Examples of correct code for this rule:
+
+```js
+// ✅ Using error boundary
+function Parent() {
+  return (
+    <ErrorBoundary>
+      <ChildComponent />
+    </ErrorBoundary>
+  );
+}
+```
+
+## Troubleshooting {/*troubleshooting*/}
+
+### Why is the linter telling me not to wrap `use` in `try`/`catch`? {/*why-is-the-linter-telling-me-not-to-wrap-use-in-trycatch*/}
+
+The `use` hook doesn't throw errors in the traditional sense, it suspends component execution. When `use` encounters a pending promise, it suspends the component and lets React show a fallback. Only Suspense and Error Boundaries can handle these cases. The linter warns against `try`/`catch` around `use` to prevent confusion as the `catch` block would never run.
+
+```js {expectedErrors: {'react-compiler': [5]}}
+// ❌ Try/catch around `use` hook
+function Component({promise}) {
+  try {
+    const data = use(promise); // Won't catch - `use` suspends, not throws
+    return <div>{data}</div>;
+  } catch (error) {
+    return <div>Failed to load</div>; // Unreachable
+  }
+}
+
+// ✅ Error boundary catches `use` errors
+function App() {
+  return (
+    <ErrorBoundary fallback={<div>Failed to load</div>}>
+      <Suspense fallback={<div>Loading...</div>}>
+        <DataComponent promise={fetchData()} />
+      </Suspense>
+    </ErrorBoundary>
+  );
+}
+```
--- a/src/content/reference/eslint-plugin-react-hooks/lints/preserve-manual-memoization.md
+++ b/src/content/reference/eslint-plugin-react-hooks/lints/preserve-manual-memoization.md
@@ -1,0 +1,93 @@
+---
+title: preserve-manual-memoization
+---
+
+<Intro>
+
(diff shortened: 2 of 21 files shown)


src/content/reference/eslint-plugin-react-hooks/lints/refs.md
```diff
@@ -0,0 +1,69 @@
...
+// ✅ Use state for UI values
+function Component() {
+  const [count, setCount] = useState(0);
+
+  return (
+    <button onClick={() => setCount(count + 1)}>
+      {count}
+    </button>
+  );
+}
```
josephsavona, line 92: We should mention the supported patterns for lazy ref initialization

josephsavona:
Wow this is incredible!!! Still going through all the content but looks great :-)

src/content/reference/eslint-plugin-react-hooks/lints/error-boundaries.md
```diff
@@ -0,0 +1,72 @@
...
+```js {expectedErrors: {'react-compiler': [5]}}
+// ❌ Try/catch around `use` hook
+function Component({promise}) {
+  try {
+    const data = use(promise); // Won't catch - `use` suspends, not throws
+    return <div>{data}</div>;
+  } catch (error) {
+    return <div>Failed to load</div>; // Unreachable
+  }
+}
```
bthall16, line 60: I understand the intention is for the implementation details of `use` to be unnecessary for React developers, but in the current implementation of `use` [it will throw](https://github.com/facebook/react/blob/16df13b84c40d92ca1bb2dee552f5a780e292d41/packages/react-reconciler/src/ReactFiberThenable.js#L273-L284) and the `catch` block will run. Someone could log to the console before returning `<div>Failed to load</div>` and see the `catch` block did indeed run, couldn't they? While it totally makes sense for developers to _not_ catch or otherwise manipulate what `use` throws, it seems like the docs stating `use` doesn't throw will cause confusion if and when they encounter it throwing.

EDIT: Maybe this is only relevant when using the compiler. I'm looking at it from the perspective of a non-compiled React component.

poteto: This will be deprecated soon: https://github.com/facebook/react/pull/34032

src/content/reference/eslint-plugin-react-hooks/lints/refs.md
poteto: added

src/content/reference/eslint-plugin-react-hooks/lints/preserve-manual-memoization.md
```diff
@@ -0,0 +1,99 @@
...
+    return a[sortBy] - b[sortBy];
+  });
+  
+  return <List items={sorted} />;
+}
+```
+
+<Note>
+
+React Compiler automatically memoizes values when beneficial. Your manual memoization won't hurt, but removing it reduces code complexity. The compiler is often better at determining what needs memoization than manual analysis.
```
josephsavona: Hmm, this is a bit more strongly in favor of removing manual memos than the other area of the docs that we recently updated. At minimum, we should suggest testing and also only removing if this rule doesn't trigger.

src/content/reference/eslint-plugin-react-hooks/lints/refs.md
```diff
@@ -0,0 +1,85 @@
...
+
+// ✅ Lazy initialization of ref value
```
josephsavona, line 94: Another thing to mention here and/or in 'immutability' is the pattern of naming ref values as "ref" or with "Ref" suffix. Maybe list all the ways we infer refs:
* Return of `useRef()`
* Both named as a ref ("ref" or "Ref" suffix) *and* some read/write of .current
* Passed as the `ref` property to JSX

src/content/reference/eslint-plugin-react-hooks/lints/set-state-in-render.md
```diff
@@ -0,0 +1,92 @@
...
+// ❌ Wrong approach
+function EditForm({itemId}) {
+  const [formData, setFormData] = useState({});
+
+  // This creates infinite renders
+  if (itemId !== lastItemId) {
+    setFormData({});
+  }
+
+  return <form>...</form>;
```
josephsavona: This is valid, but the pattern would involve storing lastItemId in another state and also resetting that state value in the conditional

src/content/reference/eslint-plugin-react-hooks/lints/use-memo.md
```diff
@@ -0,0 +1,88 @@
...
+
+Examples of incorrect code for this rule:
+
+```js {expectedErrors: {'react-compiler': [3]}}
+// ❌ No return value
+function Component({ data }) {
+  const processed = useMemo(() => {
+    data.forEach(item => console.log(item));
+    // Missing return!
+  }, [data]);
```
josephsavona, line 25: maybe add a case where the useMemo returns, but the result is not assigned anywhere? we detect that too right?

```js
useMemo(() = {
  return ...
}, [...]);
```

josephsavona (approved):
These docs are incredible!

src/content/reference/eslint-plugin-react-hooks/lints/use-memo.md
poteto: ~~oh hmm, let me check if i handled that~~ i did not handle that case, will follow up

Screenshot 2025-09-18 at 3 47 51 PM

src/content/reference/eslint-plugin-react-hooks/lints/preserve-manual-memoization.md
poteto: oops, meant to remove this

src/content/reference/eslint-plugin-react-hooks/lints/set-state-in-render.md
poteto: let me change up this example, it's not very good

src/content/reference/eslint-plugin-react-hooks/lints/use-memo.md
poteto: added the example

karlhorky:
Thanks for this PR @poteto ! Cool to see that the `react-compiler` rule was split up into multiple rules for clearer feedback.

I noticed that [`incompatible-library`](https://react.dev/reference/eslint-plugin-react-hooks/lints/incompatible-library) is not yet available in [`eslint-plugin-react-hooks@6.0.0-rc.2`](https://www.npmjs.com/package/eslint-plugin-react-hooks/v/6.0.0-rc.2) (latest RC at time of writing), opened this issue:

- https://github.com/facebook/react/issues/34632

poteto:
@karlhorky thanks, we're cutting a new RC soon

PR #7987, opened 2025-09-16
Fix sidebar overlapping main content by making width responsive

sunil5151:
### Problem
The sidebar on the React documentation had a fixed width, causing it to overlap the main article content on certain screen sizes.

### Solution
Changed the sidebar's CSS width from a fixed value to `auto`/responsive so that it no longer overlaps the main content.

### Result
- Sidebar now adjusts dynamically based on screen size.
- Main article content is fully visible.
- Layout is responsive and works on all screen widths.

Fixes: #<7976>


sunil5151:
PLease tell me what is issue , please i want to solve the problem but i just dont know about it

rickhanlonii (changes requested):
why did the lock file change?

PR #7989, opened 2025-09-17
fix compiler errors

poteto:
I went through the list of existing errors and tried to separate the expected errors from those that seem to be flagging unexpected issues.  In particular, our effects validations are flagging patterns that our own docs examples use. I added todos for these and will follow up later.

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/reactjs/react.dev/pull/7989).
* #7986
* __->__ #7989
* #7988

### DIFF
--- a/src/content/learn/describing-the-ui.md
+++ b/src/content/learn/describing-the-ui.md
@@ -474,7 +474,7 @@ By strictly only writing your components as pure functions, you can avoid an ent
 
 <Sandpack>
 
-```js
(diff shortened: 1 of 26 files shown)


aaronjordan:
Hey folks, I think something in this PR stack is causing [issue 7994](https://github.com/reactjs/react.dev/issues/7994).

PR #7999, opened 2025-09-19
Fix ViewTransition example

EugeneChoi4:
The first example shown in the `<ViewTransition>` docs actually did not work, since the `<ViewTransition>` element was not wrapping a DOM node. Updated the Pitfall section for more clarity as well.


gaearon:
cc @rickhanlonii

EugeneChoi4:
I think I'm missing a Vercel permission here, would I be able to get this?

PR #8001, opened 2025-09-21
add: Nagpur to the list of React meetups

AayushPaigwar:
Adding the React Community for Nagpur India


rickhanlonii:
We don't put bit.ly links on the site because you can just re-direct them later.

PR #8010, opened 2025-09-23
Add experimental docs for Fragment refs

jackpope:
Here's a first pass at documenting the Fragment refs feature which is available on experimental builds.

### DIFF
--- a/src/content/reference/react/Fragment.md
+++ b/src/content/reference/react/Fragment.md
@@ -4,7 +4,9 @@ title: <Fragment> (<>...</>)
 
 <Intro>
 
-`<Fragment>`, often used via `<>...</>` syntax, lets you group elements without a wrapper node.
+`<Fragment>`, often used via `<>...</>` syntax, lets you group elements without a wrapper node. 
+
+<Experimental> Fragments can also accept refs, which enable interacting with underlying DOM nodes without adding wrapper elements. See reference and usage below.</Experimental>
 
 ```js
 <>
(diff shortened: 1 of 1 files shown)


src/content/reference/react/Fragment.md
```diff
@@ -208,3 +239,92 @@ function PostBody({ body }) {
...
+  return (
+    <VisibilityObserverFragment onVisibilityChange={handleVisibilityChange}>
+      <SomeThirdPartyComponent />
+      <AnotherComponent />
+    </VisibilityObserverFragment>
+  );
+}
+```
+
+This pattern is an alternative to effect-based visibility logging, which is usually not what you want since you don't know of the element you're targeting is in the viewport.
```
josephsavona: nit: "not what you want" feels a bit too casual and imprecise for the docs

src/content/reference/react/Fragment.md
```diff
@@ -28,13 +30,42 @@ Wrap elements in `<Fragment>` to group them together in situations where you nee
 #### Props {/*props*/}
 
 - **optional** `key`: Fragments declared with the explicit `<Fragment>` syntax may have [keys.](/learn/rendering-lists#keeping-list-items-in-order-with-key)
+- <ExperimentalBadge />  **optional** `ref`: A ref object or callback function. React will forward the ref to a `FragmentInstance` object that provides methods for interacting with the DOM nodes wrapped by the Fragment.
```
eps1lon:
```suggestion
- <ExperimentalBadge />  **optional** `ref`: A ref object (e.g. from [`useRef`](/reference/react/useRef)) or [callback function](/reference/react-dom/components/common#ref-callback). React provides a `FragmentInstance` in the ref value that implements methods for interacting with the DOM nodes wrapped by the Fragment.
```

src/content/reference/react/Fragment.md
```diff
@@ -208,3 +239,92 @@ function PostBody({ body }) {
...
+  return (
+    <VisibilityObserverFragment onVisibilityChange={handleVisibilityChange}>
+      <SomeThirdPartyComponent />
+      <AnotherComponent />
+    </VisibilityObserverFragment>
+  );
+}
+```
+
+This pattern is an alternative to effect-based visibility logging, which is an anti-pattern in most cases. Relying on effects alone does not guarantee that the rendered component is observable by the user.
```
eps1lon:
```suggestion
This pattern is an alternative to effect-based visibility logging, which is an anti-pattern in most cases. Relying on Effects alone does not guarantee that the rendered Component is observable by the user.
```

src/content/reference/react/Fragment.md
```diff
@@ -208,3 +239,92 @@ function PostBody({ body }) {
...
+      {children}
+    </Fragment>
+  );
+}
+```
+---
+
+### <ExperimentalBadge /> Tracking visibility with Fragment refs {/*tracking-visibility-with-fragment-refs*/}
+
+Fragment refs are useful for visibility tracking and intersection observation. This enables you to monitor when content becomes visible without requiring the child components to expose refs:
```
eps1lon:
```suggestion
Fragment refs are useful for visibility tracking and intersection observation. This enables you to monitor when content becomes visible without requiring the child Components to expose refs:
```

eps1lon (commented):
Have we introduced the concept of "Host Components" somewhere else? This isn't just specific to DOM nodes but the Host Components of a specific renderer. The DOM is only one example. Native is another.

eps1lon (approved):
non-blocking review above.

gaearon:
We don't say "host" in docs. The common approach so far has been to just say DOM and imply that native has "its own DOM". Which kind of aligns with the DOM-like APIs that RN is adding but may be a bit confusing. I think it's fine though and can be called out separately.

src/content/reference/react/Fragment.md
```diff
@@ -28,13 +30,42 @@ Wrap elements in `<Fragment>` to group them together in situations where you nee
...
+- `removeEventListener(type, listener, options?)`: Removes an event listener from all first-level DOM children of the Fragment.
+- `dispatchEvent(event)`: Dispatches an event to a virtual child of the Fragment to call any added listeners and can bubble to the DOM parent.
+
+**Layout methods:**
+- `compareDocumentPosition(otherNode)`: Compares the document position of the Fragment with another node.
+  - If the Fragment has children, the native `compareDocumentPosition` value is returned. 
+  - Empty Fragments will attempt to compare positioning within the React tree and include `Node.DOCUMENT_POSITION_IMPLEMENTATION_SPECIFIC`.
+  - Elements that have a different relationship in the React tree and DOM tree due to portaling or other insertions are `Node.DOCUMENT_POSITION_IMPLEMENTATION_SPECIFIC`.
+- `getClientRects()`: Returns a flat array of `DOMRect` objects representing the bounding rectangles of all children.
+- `getRootNode()`: Returns the root node containing the Fragment's parent DOM node.
```
slorber, line 50: Dumb question but why can we easily get the root node, but not the parent node?

Wouldn't this be more flexible alternative to findDOMNode? `fragmentRef.current.parent.getRootNode()` ?

src/content/reference/react/Fragment.md
```diff
@@ -208,3 +239,92 @@ function PostBody({ body }) {
...
+
+### <ExperimentalBadge /> Focus management with Fragment refs {/*focus-management-with-fragment-refs*/}
+
+Fragment refs provide focus management methods that work across all DOM nodes within the Fragment:
+
+```js
+import { Fragment, useRef } from 'react';
+
+function FocusFragment({ children }) {
+  const fragmentRef = useRef(null);
```
slorber, line 320: looks like a typo, PR https://github.com/reactjs/react.dev/pull/8056

slorber (commented):
See 2 comments

Also, isn''t `scrollIntoView()` docs missing? (cf https://github.com/facebook/react/pull/34720)

PR #8012, opened 2025-09-23
[useCallback] Rephrase sentence for removing Effect dependencies

rammba:
If I understood the intention of the current sentence correctly, it should say that function should be moved outside of an Effect, because it's an explanation of `Removing Effect Dependencies` article. I just wrote that intention explicitly.

Note: This PR has a conflict with #8011.


rammba:
I just realized there is a [section](https://react.dev/learn/removing-effect-dependencies#move-dynamic-objects-and-functions-inside-your-effect) about moving data inside an Effect. This can be abandoned because I misunderstood the sentence.

PR #8014, opened 2025-09-24
docs: Update head change to <head> in /reference/react-dom/components/index.md

WuMingDao:
[change here](https://react.dev/reference/react-dom/components#resource-and-metadata-components)

I think this provides more clarity:

## before 
They are special in React because React can render them into the document head,

## after
They are special in React because React can render them into the document `<head>`,

## becase
image

this into `<head>`


rickhanlonii:
Thanks, but this isn't really worth the translation churn.

PR #8020, opened 2025-09-25
Fix fragment link back button issue

DipakHalkude:
- Change fragment links from <a> tags to NextLink components in MDX/Link.tsx
- Set history.scrollRestoration = 'auto' for all browsers in _app.tsx
- This prevents multiple new pages in backStack and maintains scroll position
- Similar to how Wikipedia handles fragment links

### DIFF
--- a/src/components/MDX/Link.tsx
+++ b/src/components/MDX/Link.tsx
@@ -44,9 +44,9 @@ function Link({
         </ExternalLink>
       ) : href.startsWith('#') ? (
         // eslint-disable-next-line jsx-a11y/anchor-has-content
-        <a className={cn(classes, className)} href={href} {...props}>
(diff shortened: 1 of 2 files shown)


DipakHalkude:
This PR fixes the browser "back" button issue when clicking on fragment links (links that start with # ) in the React documentation.

Problem: When users clicked on fragment links like [Link](#section) and then used the browser's back button, the page would not return to the previous scroll position.

Solution:

- Updated src/components/MDX/Link.tsx : Changed fragment links from <a> tags to <NextLink> components to use Next.js's client-side routing
- Updated src/pages/_app.tsx : Set history.scrollRestoration = 'auto' for all browsers (not just Safari) to enable proper scroll position restoration

PR #8023, opened 2025-09-26
Docs for `cacheSignal`

eps1lon:
Preview: https://react-do5kq6ru7-fbopensource.vercel.app/reference/react/cacheSignal

Mostly taken from https://github.com/facebook/react/pull/33557

### DIFF
--- a/src/content/reference/react/cacheSignal.md
+++ b/src/content/reference/react/cacheSignal.md
@@ -18,7 +18,7 @@ title: cacheSignal
 
 <Intro>
 
-`cacheSignal()` returns an [`AbortSignal`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) which aborts when the cache lifetime is done based on the same execution scope as a [`cache()`](/reference/react/cache)ed function 
+`cacheSignal` allows you to know when the `cache()` life time is over.
 
 ```js
 const signal = cacheSignal();


src/content/reference/react/cacheSignal.md
```diff
@@ -0,0 +1,121 @@
...
+
+**The `cacheSignal()` API is currently only available in React’s Canary and Experimental channels.** 
+
+[Learn more about React’s release channels here.](/community/versioning-policy#all-release-channels)
+
+</Canary>
+
+<RSC>
+
+`cacheSignal` is only for use with [React Server Components](/blog/2023/03/22/react-labs-what-we-have-been-working-on-march-2023#react-server-components).
```
sebmarkbage:
```suggestion
`cacheSignal` is currently only used with [React Server Components](/blog/2023/03/22/react-labs-what-we-have-been-working-on-march-2023#react-server-components). 
```

src/content/reference/react/cacheSignal.md
```diff
@@ -0,0 +1,121 @@
...
+
+This function does not accept any parameters.
+
+#### Returns {/*returns*/}
+
+`cacheSignal` returns an `AbortSignal` if called during rendering. Otherwise `cacheSignal()` returns `null`.
+
+#### Caveats {/*caveats*/}
+
+- `cacheSignal` is for use in [React Server Components](/reference/rsc/server-components) only. In Client Components, it will always return `null`.
```
sebmarkbage:
```suggestion
- `cacheSignal` is currently for use in [React Server Components](/reference/rsc/server-components) only. In Client Components, it will always return `null`. In the future it will also be used for Client Component when a client cache refreshes or invalidates. You should not assume it'll always be null on the client.
```

src/content/reference/react/cacheSignal.md
```diff
@@ -0,0 +1,121 @@
...
+
+<RSC>
+
+`cacheSignal` is currently only used with [React Server Components](/blog/2023/03/22/react-labs-what-we-have-been-working-on-march-2023#react-server-components). 
+
+</RSC>
+
+<Intro>
+
+`cacheSignal()` returns an [`AbortSignal`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) which aborts when the cache lifetime is done based on the same execution scope as a [`cache()`](/reference/react/cache)ed function
```
rickhanlonii: This needs to be explained simple enough to fit on one line in desktop. Something like:

```md
`cacheSignal` allows you to know when a `cache()` is cleared.
```

eps1lon:
```suggestion
`cacheSignal` allows you to know when the `cache()` life time is over.
```

src/content/reference/react/cacheSignal.md
```diff
@@ -0,0 +1,121 @@
...
+## Reference {/*reference*/}
+
+### `cacheSignal` {/*cachesignal*/}
+
+Call `cacheSignal` to get an `AbortSignal`.
+
+```js {3,7}
+import {cacheSignal} from 'react';
+async function Component() {
+  await fetch(url, { signal: cacheSignal() });
```
rickhanlonii, line 42: This example doesn't make sense because you have no idea that the fetch here is using `cache()`.

Can you use something like a DB example, like here? 

https://react-do5kq6ru7-fbopensource.vercel.app/reference/react/cache#preload-data

eps1lon: You don't need to know if the call is cached or not. Just that it may be cached by `cache()`.

PR #8025, opened 2025-09-26
Update useEffectEvent docs for canary

jackpope:
Note: there are still some code samples using the `experimental_useEffectEvent` import. Will update those as soon as we have a published version without the prefix.


src/content/learn/escape-hatches.md
```diff
@@ -471,7 +471,7 @@ This is not ideal. You want to re-connect to the chat only if the `roomId` has c
 
 ```js
 import { useState, useEffect } from 'react';
-import { experimental_useEffectEvent as useEffectEvent } from 'react';
+import { useEffectEvent as useEffectEvent } from 'react';
```
eps1lon:
```suggestion
import { useEffectEvent } from 'react';
```
and the rest

src/content/reference/react/useEffectEvent.md
```diff
@@ -0,0 +1,102 @@
...
+  const { items } = useContext(ShoppingCartContext);
+  const numberOfItems = items.length;
+
+  const onNavigate = useEffectEvent((visitedUrl) => {
+    logVisit(visitedUrl, numberOfItems);
+  });
+
+  useEffect(() => {
+    onNavigate(url);
+  }, [url]);
```
eps1lon, line 95: Isn't the value of `visitedUrl` the same as the `url` since `useEffectEvent` always closes over the latest values? So this feels like useless indirection. 

A better example would be using it as an event handler and passing something as an argument that's not from props/state.

jackpope: This shows a mix of reactive and nonreactive values. So `visitedUrl` and `url` are the same, but using the `url` in the effect allows us to log on each render that `url` changed, without worrying about logging when `numberOfItems` changes.

eps1lon (approved):
Still seeing a lot of references to `experimental_useEffectEvent` e.g `src/content/learn/removing-effect-dependencies.md`. The sandboxes can also be updated to `canary`. Makes it easier for me to ship everything as stable with a simple grep

jackpope:
@eps1lon I cleaned up the prefixes in https://github.com/reactjs/react.dev/pull/8025/commits/eec0fe61d84fcd6cbd58260acbbc5b12798f21c7

eps1lon:
Whoops. That was a pending comment that wasn't supposed to go into this approval.

src/content/reference/react/useEffectEvent.md
gaearon: It would be good to maybe explicitly say that in the text. Ie explain the resulting behavior.

PR #8027, opened 2025-09-30
Fix incorrect negative in use-memo intro

karlhorky:
From the examples, it appears that [the `react-hooks/use-memo` rule](https://react.dev/reference/eslint-plugin-react-hooks/lints/use-memo) validates for `useMemo()` **with** a return value, not **without**.

cc @poteto fix for small error in #7986 

### Alternatives considered

I first considered wording it as a double negative:

```
Validates against usage of the `useMemo` hook without a return value.
```

But that seemed more confusing, so I went with flipping to a positive.

### DIFF
--- a/src/content/reference/eslint-plugin-react-hooks/lints/use-memo.md
+++ b/src/content/reference/eslint-plugin-react-hooks/lints/use-memo.md
@@ -5,7 +5,7 @@ version: rc
 
 <Intro>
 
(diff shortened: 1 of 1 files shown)


poteto:
Thanks!

karlhorky:
Glad to help, thanks for the review and merge!

Issue #7012, opened 2024-07-03
Some recommended alternatives to useEffect are less declarative and concise


jedwards1211:
### Summary

You Might Not Need an Effect has this example:
```js
  // 🔴 Avoid: Adjusting state on prop change in an Effect
  useEffect(() => {
    setSelection(null);
  }, [items]);
```
I understand the technical reasons for discouraging `useEffect`, but it seems like the authors haven't considered this from a readability perspective.

The docs suggest this instead, which is way less concise, declarative, and IMO much less readable:
```js
  // Better: Adjust the state while rendering
  const [prevItems, setPrevItems] = useState(items);
  if (items !== prevItems) {
    setPrevItems(items);
    setSelection(null);
  }
```

### Page

https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

### Details

To have the better readability, I would use this custom hook in userland.  Since it's so easy for people to abuse `useEffect`, I would argue React should provide something like the following as a builtin.  But if not, the docs should at least recommend it:

```js
useImmediateEffect(() => {
  setSelection(null);
}, [items]);
```
```js
function useImmediateEffect(effect, deps) {
  const [finishLastEffect, setFinishLastEffect] = useState(() => {});
  const [prevDeps, setPrevDeps] = useState(deps);
  if (!shallowEqual(deps, prevDeps)) {
    setPrevDeps(deps);
    finishLastEffect();
    setFinishLastEffect(effect());
  }
}

```

Luccasoli:
I agree! And it helps to influence technically within the technical teams we work with!

I recently saw this article creating a custom hook similar to yours: https://www.jacobparis.com/content/use-reset-callback#extracting-the-logic-into-a-custom-hook

````jsx
function useResetCallback(initialValue, resetFn: () => any) {
  const [prevValue, setPrevValue] = useState(initialValue)
  if (prevValue !== initialValue) {
    resetFn()
    setPrevValue(initialValue)
  }
}
````

karlhorky:
Also relevant here are two new React Compiler ESLint rules:

- [`react-hooks/set-state-in-effect`](https://react.dev/reference/eslint-plugin-react-hooks/lints/set-state-in-effect)
- [`react-hooks/set-state-in-render`](https://react.dev/reference/eslint-plugin-react-hooks/lints/set-state-in-render)

[...]

PR #8029, opened 2025-10-01
Fix "Purity" hydration error, add random content example

karlhorky:
The last example on [the `react-hooks/purity` rule docs](https://react.dev/reference/eslint-plugin-react-hooks/lints/purity) causes a hydration error when SSRed:

CodeSandbox: https://codesandbox.io/p/devbox/date-now-with-hydration-error-ntchmm?workspaceId=ws_GfAuHrswXyA1DoeSwsjjjz

Screenshot 2025-10-01 at 10 47 10

The `useState()` docs mention that the initializer function should be pure:

> If you pass a function as initialState, it will be treated as an initializer function. It should be pure,

- https://react.dev/reference/react/useState#parameters

### Suggested solution

1. Move the `Math.random()` call from the `useState()` initializer function to `useEffect()`
2. Initializing the state variable with `0` to avoid TypeScript type errors
3. Use early return to render nothing if state variable is set to `0`

CodeSandbox: https://codesandbox.io/p/devbox/date-now-without-hydration-error-lq5t8c?workspaceId=ws_GfAuHrswXyA1DoeSwsjjjz

Screenshot 2025-10-01 at 10 58 07

As part of this work, I also added an additional example for random content, which seems like a fairly common use case:

4. Create a `message` state variable, initializing with `''`
5. Set a random value of the state variable in `useEffect()`
6. Use early return to render nothing if the state variable is set to `''`

CodeSandbox: https://codesandbox.io/p/devbox/math-random-without-hydration-error-sjdcmh?file=%2Fapp%2FRandomMessage.tsx%3A7%2C16-17%2C2&workspaceId=ws_GfAuHrswXyA1DoeSwsjjjz

Screenshot 2025-10-01 at 11 11 50

cc @poteto @josephsavona 

### Additional notes

If the change to the `Date.now()` example is accepted, then it probably should also be changed in the example on this page:

- Components and Hooks must be pure -> Components and Hooks must be idempotent https://react.dev/reference/rules/components-and-hooks-must-be-pure#components-and-hooks-must-be-idempotent

### Alternatives considered

A) Leaving out the early return - this would cause Flash of Unhydrated Content as the page loads, which is probably undesirable (could be changed to a loading message, but that seems out of scope)
B) Different approach: using `next/dynamic` with `ssr: false` (downside: this requires an additional Client Component in between, because `ssr: false` cannot be used in the `page.tsx`, because it's a Server Component)

### DIFF
--- a/src/content/reference/eslint-plugin-react-hooks/lints/purity.md
+++ b/src/content/reference/eslint-plugin-react-hooks/lints/purity.md
@@ -62,6 +62,41 @@ function Component() {
 
 ## Troubleshooting {/*troubleshooting*/}
 
+### I need to show random content {/*random-content*/}
+
+Calling `Math.random()` during render makes your component impure:
+
+```js {expectedErrors: {'react-compiler': [7]}}
+const messages = ['a', 'b', 'c'];
+
+// ❌ Wrong: Random message changes every render
+function RandomMessage() {
+  return (
+    <div>
+      Random message: {messages[Math.floor(messages.length * Math.random())]}
+    </div>
+  );
+}
+```
+
+Instead, move the impure function into an effect:
+
+```js
+const messages = ['a', 'b', 'c'];
+
+function RandomMessage() {
+  const [message, setMessage] = useState('');
+
+  useEffect(() => {
+    setMessage(messages[Math.floor(messages.length * Math.random())]);
+  }, []);
+
+  if (message === '') return;
+
+  return <div>Random message: {message}</div>;
+}
+```
+
 ### I need to show the current time {/*current-time*/}
 
 Calling `Date.now()` during render makes your component impure:
@@ -77,7 +112,7 @@ Instead, [move the impure function outside of render](/reference/rules/component
 
 ```js
 function Clock() {
-  const [time, setTime] = useState(() => Date.now());
+  const [time, setTime] = useState(0);
 
   useEffect(() => {
     const interval = setInterval(() => {
@@ -86,6 +121,8 @@ function Clock() {
 
     return () => clearInterval(interval);
   }, []);
+
+  if (time === 0) return;
 
   return <div>Current time: {time}</div>;
 }


src/content/reference/eslint-plugin-react-hooks/lints/purity.md
```diff
@@ -62,6 +62,41 @@ function Component() {
...
+Instead, move the impure function into an effect:
+
+```js
+const messages = ['a', 'b', 'c'];
+
+function RandomMessage() {
+  const [message, setMessage] = useState('');
+
+  useEffect(() => {
+    setMessage(messages[Math.floor(messages.length * Math.random())]);
```
karlhorky, line 89: [`react-hooks/set-state-in-effect`](https://react.dev/reference/eslint-plugin-react-hooks/lints/set-state-in-effect) is currently erroring out in [the latest workflow run](https://github.com/reactjs/react.dev/actions/runs/18157968651/job/51682092724?pr=8029) with:

```
/home/runner/work/react.dev/react.dev/src/content/reference/eslint-plugin-react-hooks/lints/purity.md
  91:5  error  Error: Calling setState synchronously within an effect can trigger cascading renders

Effects are intended to synchronize state between React and external systems such as manually updating the DOM, state management libraries, or other platform APIs. In general, the body of an effect should do one or both of the following:
* Update external systems with the latest state from React.
* Subscribe for updates from some external system, calling setState in a callback function when external state changes.

Calling setState synchronously within an effect body causes cascading renders that can hurt performance, and is not recommended. (https://react.dev/learn/you-might-not-need-an-effect)

/home/runner/work/react.dev/react.dev/src/content/reference/eslint-plugin-react-hooks/lints/purity.md#codeblock:7:5
   5 |
   6 |   useEffect(() => {
>  7 |     setMessage(messages[Math.floor(messages.length * Math.random())]);
     |     ^^^^^^^^^^ Avoid calling setState() directly within an effect
   8 |   }, []);
   9 |
  10 |   if (message === '') return;  local-rules/lint-markdown-code-blocks

✖ 1 problem (1 error, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.

ERROR: "lint" exited with 1.
Error: Process completed with exit code 1.
```

I could add an `eslint-disable-next-line` for `react-hooks/set-state-in-effect`, but it seems suboptimal to disable one lint rule to satisfy another.

Three other ways to avoid the reported lint problem, all kind of code-smelly:

```tsx
useEffect(() => {
  startTransition(() => {
    setMessage(messages[Math.floor(messages.length * Math.random())]);
  });
```

or:

```tsx
useEffect(() => {
  const id = requestAnimationFrame(() => setMessage(messages[Math.floor(messages.length * Math.random())]));
  return () => cancelAnimationFrame(id);
```

or:

```tsx
useEffect(() => {
  (async () => setMessage(messages[Math.floor(messages.length * Math.random())]))()
```

karlhorky:
Also, if we change focus to Server Components instead:

What's the recommendation for a Server Component (Next.js page) which shows random quotes on every page load?

Currently the example code below "works":

1. Arguably is correct behavior, because re-renders of the page (on page load) should cause a random quote to appear
2. Is based on the example code from [the Next.js `connection()` docs](https://nextjs.org/docs/app/api-reference/functions/connection) by @delbaoliveira (https://github.com/vercel/next.js/blob/canary/docs/01-app/03-api-reference/04-functions/connection.mdx updated in https://github.com/vercel/next.js/pull/70031/files#diff-f2532d57b916ccfb551fd27493164f75eb9a118d1bf0ce874c12f2bebb855838)

`app/page.tsx`

```tsx
import { connection } from 'next/server';

const quotes = [
  'Unlock diverse expertise',
  'Fresh perspectives, proven skills',
  'Future-ready talent',
];

export default async function Page() {
  await connection();
  return <div>{quotes[Math.floor(Math.random() * quotes.length)]}</div>;
}
```

This currently shows the `Cannot call impure function during render` error with `react-hooks/purity` on the `Math.random()` call:

Screenshot 2025-10-01 at 15 31 26<br /><br />

Could of course switch to a wrapper function `mathRandom()` which is not recognized by `react-hooks/purity`, but that seems like just tricking the lint rule - the function is still impure:

`app/page.tsx`

```tsx
import { connection } from 'next/server';

const quotes = [
  'Unlock diverse expertise',
  'Fresh perspectives, proven skills',
  'Future-ready talent',
];

function mathRandom() {
  return Math.random();
}

export default async function Page() {
  await connection();
  return <div>{quotes[Math.floor(mathRandom() * quotes.length)]}</div>;
}
```

---

What is the recommended pattern for Server Components? Should [the `react-hooks/purity` lint rule](https://react.dev/reference/eslint-plugin-react-hooks/lints/purity) be changed to add exceptions for A) usage of async component functions or B) usage of dynamic functions like [`connection()`](https://nextjs.org/docs/app/api-reference/functions/connection), [`headers()`](https://nextjs.org/docs/app/api-reference/functions/headers), etc?

Also, if the Next.js example doesn't show best practice for the React Compiler, should the Next.js docs file [`connection.mdx`](https://github.com/vercel/next.js/blob/canary/docs/01-app/03-api-reference/04-functions/connection.mdx) be changed?

src/content/reference/eslint-plugin-react-hooks/lints/purity.md
karlhorky: Another way around the `react-hooks/set-state-in-effect` problem that @gaearon made me think of in https://github.com/reactjs/react.dev/issues/7012#issuecomment-3357773451 and https://github.com/reactjs/react.dev/pull/8033, based on the approach of setting state during rendering mentioned in [Adjusting some state when a prop changes](https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes):

```diff
-  useEffect(() => {
-    setMessage(messages[Math.floor(messages.length * Math.random())]);
-  }, []);
+  if (message === '') {
+    setMessage(messages[Math.floor(messages.length * Math.random())]);
+  }
```

But then a problem appears for `react-hooks/purity` again, because the `Math.random()` is in render...

Issue #8034, opened 2025-10-01
[Bug]: <Activity /> doc page / 19.2.x examples crashing.


webbertakken:
### Summary

`<Activity />` doc [page](https://react.dev/reference/react/Activity) second example crashing.

Looks like the same might be true for other features that are using the new React 19.2.

Perhaps the tool wasn't correctly updated to use the latest react version in prod?

### Page

https://react.dev/reference/react/Activity

### Details

`<Activity />` doc [page](https://react.dev/reference/react/Activity) second example crashing, right after the phrase: `and check out the new behavior:`

<img width="1523" height="1117" alt="Image" src="https://github.com/user-attachments/assets/618ceaa0-a9b3-4bba-8967-609a6860669c" />

<details>

<summary><h2>Error in console:</h2></summary>

```console
Uncaught Error: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.

Check the render method of `App`.
    at createFiberFromTypeAndProps (786946de.sandpack-bundler-4bw.pages.dev/node_modules/react-dom/cjs/react-dom-client.development.js:11:107370)
    at createFiberFromElement (786946de.sandpack-bundler-4bw.pages.dev/node_modules/react-dom/cjs/react-dom-client.development.js:11:107754)
    at createChild (786946de.sandpack-bundler-4bw.pages.dev/node_modules/react-dom/cjs/react-dom-client.development.js:11:144463)
    at reconcileChildrenArray (786946de.sandpack-bundler-4bw.pages.dev/node_modules/react-dom/cjs/react-dom-client.development.js:11:151589)
    at reconcileChildFibersImpl (786946de.sandpack-bundler-4bw.pages.dev/node_modules/react-dom/cjs/react-dom-client.development.js:11:159233)
    at eval (786946de.sandpack-bundler-4bw.pages.dev/node_modules/react-dom/cjs/react-dom-client.development.js:11:162207)
    at reconcileChildren (786946de.sandpack-bundler-4bw.pages.dev/node_modules/react-dom/cjs/react-dom-client.development.js:11:231764)
    at updateFunctionComponent (786946de.sandpack-bundler-4bw.pages.dev/node_modules/react-dom/cjs/react-dom-client.development.js:11:239506)
[...]

PR #8035, opened 2025-10-01
Fix Activity sandboxes

rickhanlonii:
Hopefully fixes https://github.com/reactjs/react.dev/issues/8034


rickhanlonii:
well, that doesn't work

rickhanlonii:
This is a total hack, it looks like sandpack is resolving to the wrong versions. Like if you put `19.2.0` in the deps, it uses some random canary version. The hack is to do:

```js
import { unstable_Activity, Activity as ActivityStable} from 'react'; 
let Activity = ActivityStable ?? unstable_Activity;
```

Until we figure out how to fix sandpack.

CompuIves:
I'm looking at this now. Will try to repro it on my side and see what we need to fix to resolve 19.2 correctly!

PR #8037, opened 2025-10-01
Upgrade sandboxes to 19.2

gaearon:
Sandpack is just broken for some reason

### DIFF
--- a/src/components/MDX/Sandpack/template.ts
+++ b/src/components/MDX/Sandpack/template.ts
@@ -35,8 +35,8 @@ root.render(
           eject: 'react-scripts eject',
         },
         dependencies: {
-          react: '^19.1.0',
-          'react-dom': '^19.1.0',
+          react: '^19.2.0',
+          'react-dom': '^19.2.0',
           'react-scripts': '^5.0.0',
         },
       },
--- a/src/components/MDX/Sandpack/template.ts
+++ b/src/components/MDX/Sandpack/template.ts
@@ -34,11 +34,9 @@ root.render(
           test: 'react-scripts test --env=jsdom',
           eject: 'react-scripts eject',
         },
-        // pin these deps to avoid issues
-        // with sandpack's dependency resolution
         dependencies: {
-          react: '19.2.0',
-          'react-dom': '19.2.0',
+          react: '^19.2.0',
+          'react-dom': '^19.2.0',
           'react-scripts': '^5.0.0',
         },
       },
--- a/src/components/MDX/Sandpack/template.ts
+++ b/src/components/MDX/Sandpack/template.ts
@@ -34,9 +34,11 @@ root.render(
           test: 'react-scripts test --env=jsdom',
           eject: 'react-scripts eject',
         },
+        // pin these deps to avoid issues
+        // with sandpack's dependency resolution
         dependencies: {
-          react: '^19.2.0',
-          'react-dom': '^19.2.0',
+          react: '19.2.0',
+          'react-dom': '19.2.0',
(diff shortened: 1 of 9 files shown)


webbertakken:
@gaearon I'm not getting 15.2 to work locally either. Could it be a bug in the lib instead of sandpack?

Exploring the `react` package, I only see types for `Activity` and such, no implementation. Not sure if that's expected?

gaearon:
No, 19.2 itself should work fine: https://codesandbox.io/p/sandbox/36zsm3

webbertakken:
Yes, it works as expected: [sandbox fork](https://codesandbox.io/p/devbox/react-dev-forked-2yh62m?workspaceId=ws_5LsUS6sAWsXFYcZVyC7EHL)

gaearon:
Filed https://github.com/codesandbox/sandpack-cdn/issues/43 for possible root cause, idk what's wrong though

webbertakken:
I'm getting the same locally in my next app. Here's a sentry log of it.

The %o and %s represent a DOM Element and a string.

I'm using `next dev` without the `--turbo` flag, meaning it's using webpack under the hood.

Not sure if that helps 🤷🏼

image

gaearon:
Please create a reproducing project, this is not necessarily related. Next includes a vendor copy of React so it might not have caught up with 19.2 yet.

webbertakken:
Here's a minimal repo to reproduce the issue in Next latest: https://github.com/webbertakken/repro-react-19-2-element-type-undefined

If I had to guess, I'd say it might have something to do with how some bundlers handle the exported Symbols.

CompuIves:
Looking into this!

gaearon:
@webbertakken I'm hearing from Next people that this is expected because Next hasn't done a stable release with 19.2 yet. It should work on canary. Next vendors its own copy of React so the versions in `package.json` have no effect.

webbertakken:
Cool, thanks!

Looks like on the dev website it's also [working](https://react.dev/reference/react/Activity#:~:text=and%20check%20out%20the%20new%20behavior%3A) now.

CompuIves:
I found the issue on our side. Due to an upgrade to the npm couchdb instance (https://github.com/orgs/community/discussions/152515) our npm registry replicator for sandpack cdn (which resolves dependency versions) does not completely sync anymore. Because of this, we cannot resolve semvers properly. I am writing a fix now.

gaearon:
Fantastic, thank you so much!

CompuIves:
Wrote a fix, the CDN is syncing now, I will make sure that it downloads new metadata in parallel to go faster. I will have an update on how long it'll take soon.

CompuIves:
Okay, it works now! Manually synced `react`, `react-dom` and `scheduler`. We should be up to date with npm in about 1 hour. Apologies for this problem!!

Both existing (live examples) and the 19.2.0 references (this PR) now work.

rickhanlonii:
Thanks @CompuIves!!!

Ok this is good to go, I needed to revert some of the `package.json` removals because they had deps on `toastify`, but I tested all the sandboxes and they're working. [Here's a diff of only my changes](https://github.com/reactjs/react.dev/compare/f16a0cb036189d48dbb47d8d43526e65c203eb37..81dccddca5f830b2229b1cd435b3043cef5631b5).

rickhanlonii:
I did find one (existing) issue on the escape hatches page: https://github.com/reactjs/react.dev/issues/8045

gaearon:
@CompuIves We may be seeing something like this again, e.g. on https://react.dev/reference/react/ViewTransition. Do you think the issue could be not fully fixed?

PR #8054, opened 2025-10-05
Document React's profiling builds

eps1lon:
Only document that they exist on when you need them. Specifics are build tool and framework dependent. We could come up with a list of build tools/frameworks we want to reference but how that list is chose and curated  may be controversial so I left it vague for now.

The main motivation was that the perf track docs said you only need to use `<Profiler>` in prod builds to opt-into Component tracks. But you need both: a profiling build and wrapping a subtree in `<Profiler>`.


src/content/reference/react/Profiler.md
```diff
@@ -37,7 +37,7 @@ Wrap a component tree in a `<Profiler>` to measure its rendering performance.
 
 #### Caveats {/*caveats*/}
 
-* Profiling adds some additional overhead, so **it is disabled in the production build by default.** To opt into production profiling, you need to enable a [special production build with profiling enabled.](https://fb.me/react-profiling)
```
eps1lon, line 40: That gist included docs for Webpack 4 and CRA 3. Both are either outdated or deprecated so I just kept a vague ("use bundler aliases or framework").

PR #8055, opened 2025-10-06
fix(Sandpack): prevent phantom browser history entries on scroll (#8047)

### LINKED ISSUE #8047: [Bug]: Browser back button not working as expected after scrolling down the page
### Summary

After slowly scrolling all the way down a page like https://react.dev/learn/separating-events-from-effects, I have to hit the browser back button several times before I'm actually taken to the previous page in my history.

### Page

https://react.dev/learn/separating-events-from-effects

### Details

I've often noticed that once I've read to the bottom of a page in the React docs, I have to hit my browser's back button several times before I'm actually taken to the previous page in my browser history. I've noticed this at least on Android Chrome and macOS Firefox.

As I scroll down a page like https://react.dev/learn/separating-events-from-effects, certain elements seem to create "phantom" browser-history entries when they enter the viewport. Here is a video grab demonstrating what I mean:

https://github.com/user-attachments/assets/f67aba96-52f7-4f4c-a744-1122bdf3e91a

In this clip, at first there are no previous browser-history entries available to go back to (the back button is dimmed). However, once the "Extracting non-reactive logic out of Effects" heading comes into view, an entry is created. If I start at the top of the page and scroll all the way to the bottom (slowly), I have to hit the back button four times before the browser navigates to my previous page; the first three times, nothing appears to happen at all.

This may or may not be related to #8021 and/or #787.

DeveloperViraj:
Fixes #8047

This PR resolves an issue where scrolling down documentation pages (such as
“Separating Events from Effects”) caused multiple phantom browser history entries,
requiring multiple back button presses to navigate away.

- Ensures Sandpack iframe bundler registers only once per mount.
- Prevents repeated history pushes from iframe reloads.
- Verified locally on Chrome and Firefox.

### DIFF
--- a/src/components/MDX/Sandpack/Preview.tsx
+++ b/src/components/MDX/Sandpack/Preview.tsx
@@ -97,18 +97,17 @@ export function Preview({
 
   const sandpackIdle = sandpack.status === 'idle';
 
-useEffect(function createBundler() {
-  const iframeElement = iframeRef.current!;
-  if (!iframeElement.dataset.registered) {
-    registerBundler(iframeElement, clientId);
-    iframeElement.dataset.registered = 'true';
-  }
-  return () => {
-    unregisterBundler(clientId);
-    iframeElement.dataset.registered = '';
-  };
-}, []);
-
+  useEffect(function createBundler() {
+    const iframeElement = iframeRef.current!;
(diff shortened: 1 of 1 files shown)


carnoxen:
I think this issue is related to codesandbox/sandpack#992, not react.dev site...

PR #8065, opened 2025-10-09
React Compiler v1

poteto:
Updates our blog posts and docs for React Compiler 1.0


src/content/blog/2024/10/21/react-compiler-beta-release.md
```diff
@@ -12,9 +12,9 @@ October 21, 2024 by [Lauren Tan](https://twitter.com/potetotes).
 
 <Note>
 
-### React Compiler is now in RC! {/*react-compiler-is-now-in-rc*/}
+### React Compiler is now stable! {/*react-compiler-is-now-in-rc*/}
 
-Please see the [RC blog post](/blog/2025/04/21/react-compiler-rc) for details.
+Please see the [stable release blog post](/blog/2025/10/08/react-compiler-1) for details.
```
josephsavona: I love that we’re updating our old blog posts to help developers avoid outdated information. 

That said it’s really important to not break links, and not rewrite history. So here and in other places, let’s keep the existing text and add an update about the stable release. Strikethrough text for what was there works too, but let’s not rewrite so aggressively.

poteto: @josephsavona I added a redirect in vercel.json, and I also added a deadlink checking script a while back to make sure we have no deadlinks. We did the same thing (rewrite the RC post) for [React 19 RC to React 19](https://github.com/reactjs/react.dev/pull/7321). I can do a separate one instead but thought we'd keep the tradition

josephsavona: Ahh ok, that's cool

src/content/blog/2025/10/08/react-compiler-1.md
```diff
@@ -0,0 +1,147 @@
...
+<TerminalBlock>
+{`pnpm add --save-dev --save-exact babel-plugin-react-compiler@latest`}
+</TerminalBlock>
+
+yarn
+<TerminalBlock>
+{`yarn add --dev --exact babel-plugin-react-compiler@latest`}
+</TerminalBlock>
+
+As part of the stable release, we've been making React Compiler easier to add to your projects and added optimizations to how the compiler generates memoization. React Compiler now supports optional chains and array indices as dependencies. We're exploring how to infer even more dependencies like equality checks and string interpolation. These improvements ultimately result in fewer re-renders and more responsive UIs, while letting you keep writing idiomatic declarative code.
```
josephsavona: > We're exploring how to infer even more dependencies like equality checks and string interpolation. These improvements ultimately result in fewer re-renders and more responsive UIs, while letting you keep writing idiomatic declarative code.

I am confident we can pull this off based on @mofeiZ's explorations, but i think we can leave this off for now. Let's talk about it when it's ready. It's easy for developers to misread this and think these optimizations are ready now (think about how everyone thought that 19 included the compiler even though we didn't say that)

src/content/blog/2025/10/07/react-compiler-1.md
```diff
@@ -0,0 +1,147 @@
...
+
+In other words, changing memoization may under rare circumstances cause unexpected behavior. For this reason, we recommend following the Rules of React and employing continuous end-to-end testing of your app so you can upgrade the compiler with confidence and identify any rules of React violations that might cause issues.
+
+If you don't have good test coverage, we recommend pinning the compiler to an exact version (eg `1.0.0`) rather than a SemVer range (eg `^1.0.0`). You can do this by passing the `--save-exact` (npm/pnpm) or `--exact` flags (yarn) when upgrading the compiler. You should then do any upgrades of the compiler manually, taking care to check that your app still works as expected.
+
```
josephsavona, line 191: should we include an answer to "can i delete my useMemo/useCallback" here? the answer can be in the docs, but it might be worth linking to it from the blog post to get more attention to how the compiler treats these hooks now

src/content/blog/2025/10/08/react-compiler-1.md
poteto: done

src/content/blog/2025/10/07/react-compiler-1.md
poteto: done

src/content/blog/2025/10/07/react-compiler-1.md
josephsavona, line 1: update to today's date?

poteto: i made it 10/07 to match the keynote

src/content/blog/2025/10/07/react-compiler-1.md
```diff
@@ -0,0 +1,190 @@
...
+Although this first version of the new React Compiler was eventually rewritten, the first prototype gave us increased confidence that this was a tractable problem, and the learnings that an alternative compiler architecture could precisely give us the memoization characteristics we wanted. [Joe Savona](https://x.com/en_JS), [Sathya Gunasekaran](https://x.com/_gsathya), [Mofei Zhang](https://x.com/zmofei), and [Lauren Tan](https://x.com/potetotes) worked through our first rewrite, moving the compiler's architecture into a Control Flow Graph (CFG) based High-Level Intermediate Representation (HIR). This paved the way for much more precise analysis and even an inference-based type system within React Compiler. Since then, many significant portions of the compiler have been rewritten, with each rewrite informed by our learnings from the previous attempt. And we have received significant help and contributions from many members of the [React team](/community/team) along the way.
```
josephsavona: super nit, instead of "inference-based type system" maybe "more precise analysis and even type inference within React Compiler"

src/content/blog/2025/10/07/react-compiler-1.md
```diff
@@ -0,0 +1,190 @@
...
+React Compiler adds automatic memoization more precisely and granularly than is possible with [`useMemo`](/reference/react/useMemo), [`useCallback`](/reference/react/useCallback), and [`React.memo`](/reference/react/memo). If you choose to keep manual memoization, React Compiler will analyze them and determine if your manual memoization matches its automatically inferred memoization. If there isn't a match, the compiler will choose to bail out of optimizing that component.
+
+This is done out of caution as a common anti-pattern with manual memoization is using it for correctness.  This means your app depends on specific values being memoized to work properly. For example, in order to prevent an infinite loop, you may have memoized some values to stop a `useEffect` call from firing. This breaks the Rules of React, but since it can potentially be dangerous for the compiler to automatically remove manual memoization, the compiler will just bail out instead. You should manually remove your handwritten memoization and verify that your app still works as expected.
```
josephsavona: This is all technically correct but I think a lot of readers may not come away with a clear understanding of how to think about these hooks in a compiler world.

An alternative framing is something like:

By default, React Compiler will memoize your code based on its analysis and heuristics. In most cases, this memoization will be as precise, or moreso, than what you may have written — and as noted above, the compiler can memoize even in cases where useMemo/useCallback cannot be used, such as after an early return.

However, in some cases developers may need more control over memoization. The useMemo and useCallback hooks can continue to be used with React Compiler as an escape hatch to provide control over which values are memoized. A common use-case for this is if a memoized value is used as an effect dependency, in order to ensure that an effect does not fire repeatedly even when its dependencies do not meaningfully change.

For new code, we recommend relying on the compiler for memoization and using useMemo/useCallback where needed to achieve precise control.

For existing code, we recommend either leaving existing memoization in place (removing it can change compilation output) or carefully testing before removing the memoization.

poteto: love it, updated

PR #8068, opened 2025-10-09
Fix grammatical error in ViewTransition documentation

Caisere:
## What
Fixed a grammatical error in the ViewTransition component documentation.

## Changes
- **Before:** `<ViewTransition> only activates if it is placed is before any DOM node`
- **After:** `<ViewTransition> only activates if it is placed before any DOM node`

## Why
The extra "is" was grammatically incorrect and made the sentence confusing for readers.

This is a simple documentation fix that improves clarity without changing any technical meaning.


Caisere:
I've signed the CLA but the check hasn't updated. Can someone please check the status?

PR #8078, opened 2025-10-15
fix: Breaking up a sentence to make it easier to understand

moijes12:
The line `Trees are a relationship model between items and UI is often represented using tree structures` was a *slightly* confusing to me. I twice read it as `Trees are a relationship model between items and UI` which was incorrect. Therefore, I think it would be better to break the sentence into two which would convey make it easier to read and convey the correct meaning. My apologies if this goes against the convention used in Meta docs.


eps1lon (approved):
understandable, thanks!

PR #8102, opened 2025-10-27
Docs/make drink pure

### LINKED ISSUE #8100: [Suggestion]: Make the Conditional Rendering example a pure function component
### Summary

In the Conditional Rendering [Challenges](https://react.dev/learn/conditional-rendering#refactor-a-series-of---to-if-and-variables), the second solution uses an external object:

```js
const drinks = {
  tea: { part: 'leaf', caffeine: '15–70 mg/cup', age: '4,000+ years' },
  coffee: { part: 'bean', caffeine: '80–185 mg/cup', age: '1,000+ years' }
};
function Drink({ name }) {
  const info = drinks[name];
  // ...
}
```

This makes the component not entirely pure because it depends on external data.

### Page

https://react.dev/learn/conditional-rendering#refactor-a-series-of---to-if-and-variables

### Details

The second solution uses an external object.
This makes the component not entirely pure because it depends on external data.

```js
const drinks = {
  tea: { part: 'leaf', caffeine: '15–70 mg/cup', age: '4,000+ years' },
  coffee: { part: 'bean', caffeine: '80–185 mg/cup', age: '1,000+ years' }
};
function Drink({ name }) {
  const info = drinks[name];
  // ...
}
export default function DrinkList() {
  return (
    <div>
      <Drink name="tea" />
      <Drink name="coffee" />
    </div>
  );
}
```

### Suggestion

It would be helpful to either:

Mention the [“Keeping Components Pure”](https://react.dev/learn/keeping-components-pure)
section for readers to understand the implications, or

Modify the example to define `drinks` inside the function so that it becomes a pure function component.

If you agree, I would be happy to submit a PR for this improvement.

DeveloperViraj:
Fixes #8100

This change moves the `drinks` data inside the `Drink` component so the component depends only on its `name` prop, making it a pure function component. This avoids a dependency on external data and better illustrates purity when teaching conditional rendering.

- Moved `drinks` object inside `Drink`
- Kept example structure intact and behavior identical
- Verified build and local preview

Let me know if you'd prefer the data to live in the parent component and pass `info` down via props, happy to update to that pattern if it's preferred.

### DIFF
--- a/src/content/learn/conditional-rendering.md
+++ b/src/content/learn/conditional-rendering.md
@@ -675,26 +675,30 @@ There are multiple ways you could go about this, but here is one starting point:
 
 ```js
 function Drink({ name }) {
-  let part, caffeine, age;
-  if (name === 'tea') {
-    part = 'leaf';
-    caffeine = '15–70 mg/cup';
-    age = '4,000+ years';
-  } else if (name === 'coffee') {
-    part = 'bean';
-    caffeine = '80–185 mg/cup';
-    age = '1,000+ years';
-  }
+  const drinks = {
+    tea: {
+      part: 'leaf',
+      caffeine: '15–70 mg/cup',
+      age: '4,000+ years'
+    },
+    coffee: {
+      part: 'bean',
+      caffeine: '80–185 mg/cup',
+      age: '1,000+ years'
+    }
+  };
+
+  const info = drinks[name];
   return (
     <section>
       <h1>{name}</h1>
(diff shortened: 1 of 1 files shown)


rickhanlonii:
Sorry if this isn't AI, but it smells like it. 

No. The description doesn't match the change (your drinks are defined inside the component, not outside). And the next part of the solution does exactly this - the point is to show the different options.

PR #8103, opened 2025-10-27
Docs/clarify effect challenge 2

### LINKED ISSUE #8099: Confusing explanation in Challenge 2 of React Learn ("You Might Not Need an Effect")
### Summary

The explanation in Challenge 2 (Cache a calculation without Effects) of You Might Not Need an Effect is confusing and somewhat misleading. The challenge suggests that replacing the useEffect with useMemo (or extracting a NewTodo component) reduces the number of calls to getVisibleTodos(), but this isn’t accurate. The original code already avoids unnecessary recomputations.

### Page

https://react.dev/learn/you-might-not-need-an-effect

### Details

In the original implementation, getVisibleTodos() is only called when either todos or showActive change. Typing in the input only updates the text state, which isn’t part of the dependency array—so getVisibleTodos() is not called on every keystroke.

The proposed useMemo solution behaves the same way: it recalculates getVisibleTodos() when todos or showActive change, and not when text changes. As a result, it doesn’t reduce the number of calls compared to the original implementation. Console logs confirm this.

The second suggested approach (extracting NewTodo into a separate component) gives the impression that the initial code was calling getVisibleTodos() on every input change, but that’s incorrect. The original code was already optimized since text wasn’t part of the effect’s dependencies.

The documentation even says:

“This approach satisfies the requirements too. When you type into the input, only the text state variable updates. Since the text state variable is in the child NewTodo component, the parent TodoList component won’t get re-rendered. This is why getVisibleTodos() doesn’t get called when you type. (It would still be called if the TodoList re-renders for another reason.)”

However, this explanation is misleading because getVisibleTodos() wasn’t being called at every input in the first place.

In short:

The useMemo version doesn’t reduce calls to getVisibleTodos(), it's exactly the same.

The second version’s explanation implies a problem that didn’t exist in the original code.

DeveloperViraj:
Fixes #8099

Clarifies that `getVisibleTodos()` is only called when `todos` or `showActive` change. Typing into the input (which updates `text`) does not trigger it because `text` is local to the `NewTodo` component. Extracting `NewTodo` improves clarity, not runtime behavior.

### DIFF
--- a/src/content/learn/you-might-not-need-an-effect.md
+++ b/src/content/learn/you-might-not-need-an-effect.md
@@ -1272,7 +1272,9 @@ input { margin-top: 10px; }
 
 </Sandpack>
 
-This approach satisfies the requirements too. When you type into the input, only the `text` state variable updates. Since the `text` state variable is in the child `NewTodo` component, the parent `TodoList` component won't get re-rendered. This is why `getVisibleTodos()` doesn't get called when you type. (It would still be called if the `TodoList` re-renders for another reason.)
+This approach works because `getVisibleTodos()` is only called when its dependencies (`todos` or `showActive`) change. In the original implementation, typing in the input doesn’t trigger additional calls because the `text` state is local to the `NewTodo` component and not part of the dependencies. 
+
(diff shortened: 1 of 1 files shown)


rickhanlonii:
Is this ai slop, the description doesn't match the changes.

Issue #6991, opened 2024-06-24
[Bug]: Error Boundaries Caveats are not added in the new docs


robertwt7:
### Summary

In the [previous docs](https://legacy.reactjs.org/docs/error-boundaries.html), error boundaries caveats are clearly defined. In the [new one](https://react.dev/reference/react/Component#catching-rendering-errors-with-an-error-boundary) however, it's not.

If we want to direct all new react eng to the new docs we should be adding this details to the new one as well

> Error boundaries do not catch errors for:
Event handlers ([learn more](https://legacy.reactjs.org/docs/error-boundaries.html#how-about-event-handlers))
Asynchronous code (e.g. setTimeout or requestAnimationFrame callbacks)
Server side rendering
Errors thrown in the error boundary itself (rather than its children)

### Page

https://react.dev/reference/react/Component#catching-rendering-errors-with-an-error-boundary

ShakedAdi:
Hi! I'm reviewing this issue and noticed that canveats about Error Boundaries apear to already be included in the current docs.

<img width="840" height="317" alt="Image" src="https://github.com/user-attachments/assets/7a4214d0-3255-4876-8a0f-377fe8a595f7" />

tell me if I misunderstood?

robertwt7:
yeah it seems like its there now, I think when i reported it last time it wasn't there. we can close this thankyou

PR #8108, opened 2025-10-29
Add note on error boundary limitations

JayCeeKay1991:
Issue #6991 

Adding a note about the limitations of Error Boundaries according to the legacy docs.

Screenshot 2025-10-29 at 16 31 34

### DIFF
--- a/src/content/reference/react/Component.md
+++ b/src/content/reference/react/Component.md
@@ -1275,9 +1275,9 @@ By default, if your application throws an error during rendering, React will rem
 Error boundaries do not catch errors for:
 
 - Event handlers [(learn more)](/learn/responding-to-events)
-- Asynchronous code (e.g. `setTimeout` or `requestAnimationFrame` callbacks)
 - [Server side rendering](/reference/react-dom/server) 
 - Errors thrown in the error boundary itself (rather than its children)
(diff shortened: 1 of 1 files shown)


src/content/reference/react/Component.md
```diff
@@ -1271,6 +1271,16 @@ We recommend defining components as functions instead of classes. [See how to mi
 
 By default, if your application throws an error during rendering, React will remove its UI from the screen. To prevent this, you can wrap a part of your UI into an *Error Boundary*. An Error Boundary is a special component that lets you display some fallback UI instead of the part that crashed--for example, an error message.
 
+<Note>
+Error boundaries do not catch errors for:
+
+- Event handlers [(learn more)](/learn/responding-to-events)
+- Asynchronous code (e.g. `setTimeout` or `requestAnimationFrame` callbacks)
```
eps1lon: Can we caveat that with `startTransition` from `useTransition` in those scenarios behaving different? Erroring in a Transition that is bound to the Hook version will trigger Error boundaries.

JayCeeKay1991: Thanks for your feedback, I added the exception.

eps1lon (approved):
Thanks!

PR #8109, opened 2025-10-29
Fix incorrect condition in "Chains of computations" example

### LINKED ISSUE #8097: [Typo]: Incorrect condition in Chains of computations section
### Summary

The logic in the “fixed” example from the “Chains of computations” section of the You Might Not Need an Effect page (the one replacing the chain of effects) seems to be incorrect — the condition should use goldCardCount < 3 instead of goldCardCount <= 3.

### Page

https://react.dev/learn/you-might-not-need-an-effect#chains-of-computations

### Details

In the “Chains of computations” section of the You Might Not Need an Effect page, the “fixed” example uses the condition goldCardCount <= 3.
This causes the player to pick a gold card five times instead of four (as in the original example) to advance to the next round, effectively changing the game logic.
Replacing the condition with goldCardCount < 3 should fix the issue.

PaulyBearCoding:
Fixes #8097

## Problem

In the "Chains of computations" section, the refactored example uses an incorrect condition that changes the game logic from the original:

```js
// Current (incorrect)
if (goldCardCount <= 3) {
  setGoldCardCount(goldCardCount + 1);
} else {
  // advance round
}
```

This allows **5 gold cards** before advancing the round, which differs from the original Effect-based example that advances after **4 gold cards**.

## Solution

Changed the condition from `<=` to `<`:

```js
// Fixed
if (goldCardCount < 3) {
  setGoldCardCount(goldCardCount + 1);
} else {
  // advance round
}
```

## Why This Fix is Correct

The original "bad" example increments first, then checks:
```js
useEffect(() => {
  setGoldCardCount(c => c + 1);  // 0→1, 1→2, 2→3, 3→4
}, [card]);

useEffect(() => {
  if (goldCardCount > 3) {  // Triggers when count reaches 4
    setRound(r => r + 1);
    setGoldCardCount(0);
  }
}, [goldCardCount]);
```

This advances after 4 cards (when count reaches 4).

The fixed example checks before incrementing:
- With `<= 3`: Allows counts 0,1,2,3 to increment → 4 values → advances on 5th card ❌
- With `< 3`: Allows counts 0,1,2 to increment → 3 values → advances on 4th card ✅

## Testing

Verified the fix by:
1. Tracing execution logic step-by-step
2. Building the docs site locally and confirming correct rendering
3. Checking that only the intended line was modified

## Changes

- File: `src/content/learn/you-might-not-need-an-effect.md`
- Line: 440
- Change: `goldCardCount <= 3` → `goldCardCount < 3`

### DIFF
--- a/src/content/learn/you-might-not-need-an-effect.md
+++ b/src/content/learn/you-might-not-need-an-effect.md
@@ -437,7 +437,7 @@ function Game() {
     // ✅ Calculate all the next state in the event handler
     setCard(nextCard);
     if (nextCard.gold) {
-      if (goldCardCount <= 3) {
+      if (goldCardCount < 3) {
         setGoldCardCount(goldCardCount + 1);
       } else {
         setGoldCardCount(0);


rickhanlonii:
I don't see the 5th card:

```js
if (nextCard.gold) {
  if (goldCardCount <= 3) {
    // doesn't enter if there are 4 cards
    setGoldCardCount(goldCardCount + 1);
  } else {
    // ends at only 4 cards
    setGoldCardCount(0);
    setRound(round + 1);
    if (round === 5) {
      alert('Good game!');
    }
  }
}
```

piowin:
> I don't see the 5th card:
> 
> ```js
> if (nextCard.gold) {
>   if (goldCardCount <= 3) {
>     // doesn't enter if there are 4 cards
>     setGoldCardCount(goldCardCount + 1);
>   } else {
>     // ends at only 4 cards
>     setGoldCardCount(0);
>     setRound(round + 1);
>     if (round === 5) {
>       alert('Good game!');
>     }
>   }
> }
> ```

@rickhanlonii, the fact that the program doesn't enter the first block when `goldCardCount` is equal to 4 doesn't mean that picking a gold card four times advances the game to the next round in this case - that is the issue. When `goldCardCount` is equal to 3, `setGoldCardCount(3+1)` is executed which advances `goldCardCount` to 4, but doesn't advances `round`, therefore the user has to pick a fifth gold card so the else block is finally executed and `setRound(round + 1)` is called.

1. `goldCardCount` is equal to 0
2. the user picks a gold card (first time)
3. `setGoldCardCount(0 + 1)` is executed and then `goldCardCount` is equal to 1
4. the user picks a gold card (second time)
5. `setGoldCardCount(1 + 1)` is executed and then `goldCardCount`  is equal to 2
6. the user picks a gold card (third time)
7. `setGoldCardCount(2 + 1)` is executed and then `goldCardCount`  is equal to 3
8. the user picks a gold card (fourth time)
9. `setGoldCardCount(3 + 1)` is executed and then `goldCardCount` is equal to 4
10. the user picks a gold card (fifth time)
11. the else block is now executed as `goldCardCount <= 3` is finally `false`
12. `setGoldCardCount(0)` and `setRound(round + 1)` are executed

rickhanlonii:
I see, so it takes 5 clicks with a gold card next.

I was wrong, I confirmed your right in this sandbox: 
https://codesandbox.io/p/devbox/amazing-star-mlmv44?utm_source=dotnew

PaulyBearCoding:
@rickhanlonii Thank you for taking the time to review this and for being open to reconsidering! I really appreciate you verifying the behavior in the sandbox.

@piowin Thank you so much for the detailed explanation! Your step-by-step breakdown really helped clarify the issue. I'm grateful for your support.

Thanks to both of you for helping get this fix merged! 🙏

PR #8125, opened 2025-11-04
Docs for Promise subclasses

eps1lon:
Turning [posts](https://bsky.app/profile/sebmarkbage.calyptus.eu/post/3lku7b7xjmk2w) into [docs](https://react-pb6c6sbhj-fbopensource.vercel.app/reference/react/use#avoiding-fallbacks-by-passing-promise-subclasses)

I.e. how React leverages `promise.status`. Targeted at Suspense-enabled libraries.

[Preview](https://react-pb6c6sbhj-fbopensource.vercel.app/reference/react/use#avoiding-fallbacks-by-passing-promise-subclasses)

### DIFF
--- a/src/content/reference/react/use.md
+++ b/src/content/reference/react/use.md
@@ -456,8 +456,9 @@ Calling `use` conditionally depending on whether a Promise is settled or not is 
 import { Suspense, use, useState } from "react";
 
 function preloadUser(id) {
-  // This is not a real implementation of getting the Promise for the user.
-  // A real implementation would probably call `fetch` or another data fetching method.
+  // This is not a real implementation of getting the
+  // Promise for the user. A real implementation would
+  // probably call `fetch` or another data fetching method.
   // The actual implementation should cache the Promise.
   const promise = Promise.resolve(`User #${id}`);
 
@@ -492,7 +493,7 @@ export default function App() {
       >
         Load User #2
       </button>
-      <Suspense key={userId} fallback={<p>Loading user...</p>}>
+      <Suspense key={userId} fallback={<p>Loading</p>}>
         {userUsable ? (
           <UserDetails userUsable={userUsable} />
         ) : (
@@ -529,8 +530,9 @@ class PromiseWithStatus extends Promise {
       resolve = _resolve;
       reject = _reject;
     });
-    // Setting the `status` field allows React to synchronously read
-    // the value if the Promise is already settled by the time the Promise is
+    // Setting the `status` field allows React to
+    // synchronously read the value if the Promise
+    // is already settled by the time the Promise is
     // passed to `use`.
     executor(
       (value) => {
@@ -548,11 +550,13 @@ class PromiseWithStatus extends Promise {
 }
 
 function preloadUser(id) {
-  // This is not a real implementation of getting the Promise for the user.
-  // A real implementation would probably call `fetch` or another data fetching method.
+  // This is not a real implementation of getting the
+  // Promise for the user. A real implementation would
+  // probably call `fetch` or another data fetching method.
   // The actual implementation should cache the Promise.
-  // The important part is that we are using the PromiseWithStatus subclass here.
-  // Check out the next step if you're not controlling the Promise creation
+  // The important part is that we are using the
+  // PromiseWithStatus subclass here. Check out the next
+  // step if you're not controlling the Promise creation
   // (e.g. when `fetch` is used).
   const promise = PromiseWithStatus.resolve(`User #${id}`);
 
@@ -573,8 +577,9 @@ export default function App() {
     <div>
       <button
         onClick={() => {
-          // flushSync is only used for illustration purposes.
-          // A real app would probably use startTransition instead.
+          // flushSync is only used for illustration
+          // purposes. A real app would probably use
+          // startTransition instead.
           flushSync(() => {
             setUser(preloadUser(1));
             setUserId(1);
@@ -591,7 +596,7 @@ export default function App() {
       >
         Load User #2
       </button>
-      <Suspense key={userId} fallback={<p>Loading user...</p>}>
+      <Suspense key={userId} fallback={<p>Loading</p>}>
         {userUsable ? (
           <UserDetails userUsable={userUsable} />
         ) : (
@@ -618,13 +623,15 @@ import { flushSync } from "react-dom";
 
 function preloadUser(id) {
   const value = `User #${id}`;
-  // This is not a real implementation of getting the Promise for the user.
-  // A real implementation would probably call `fetch` or another data fetching method.
+  // This is not a real implementation of getting the
+  // Promise for the user. A real implementation would
+  // probably call `fetch` or another data fetching method.
   // The actual implementation should cache the Promise.
   const promise = Promise.resolve(value);
 
   // We don't need to create a custom subclass.
-  // We can just set the necessary fields directly on the Promise.
+  // We can just set the necessary fields directly on the
+  // Promise.
   promise.status = "pending";
   promise.then(
     (value) => {
@@ -637,10 +644,11 @@ function preloadUser(id) {
     }
   );
 
-  // Setting the status in `.then` is too late if we want to create an already
-  // settled Promise. We only included setting the fields in `.then` for
-  // illustration purposes. Since our demo wants an already resolved Promise, 
-  // we set the necessary fields synchronously.
+  // Setting the status in `.then` is too late if we want
+  // to create an already settled Promise. We only included
+  // setting the fields in `.then` for illustration
+  // purposes. Since our demo wants an already resolved
+  // Promise, we set the necessary fields synchronously.
   promise.status = "fulfilled";
   promise.value = value;
   return promise;
@@ -660,8 +668,9 @@ export default function App() {
     <div>
       <button
         onClick={() => {
-          // flushSync is only used for illustration purposes.
-          // A real app would probably use startTransition instead.
+          // flushSync is only used for illustration
+          // purposes. A real app would probably use
+          // startTransition instead.
           flushSync(() => {
             setUser(preloadUser(1));
             setUserId(1);
@@ -680,7 +689,7 @@ export default function App() {
       >
         Load User #2
       </button>
-      <Suspense key={userId} fallback={<p>Loading user...</p>}>
+      <Suspense key={userId} fallback={<p>Loading</p>}>
         {userUsable ? (
           <UserDetails userUsable={userUsable} />
         ) : (


src/content/reference/react/use.md
```diff
@@ -436,6 +436,88 @@ To use the Promise's <CodeStep step={1}>`catch`</CodeStep> method, call <CodeSte
 
 ---
 
+### Avoiding fallbacks by passing Promise subclasses {/*avoiding-fallbacks-by-passing-promise-subclasses*/}
+
+React will read the `status` field on Promise subclasses to synchronously read the value without waiting for a microtask. If the Promise is already settled (resolved or rejected), React can read the value immediately without suspending and showing a fallback if the update was not part of a Transition (e.g. [`ReactDOM.flushSync()`](/reference/react-dom/flushSync)).
```
rickhanlonii: Can we add some code here for a high-level view of what this looks like?

```js
const dataPromise = {
  then: () => {},
  status: 'fulfilled',
  value: `some data'
}
```

src/content/reference/react/use.md
```diff
@@ -436,6 +436,88 @@ To use the Promise's <CodeStep step={1}>`catch`</CodeStep> method, call <CodeSte
...
+
+React will set the `status` field itself if the passed Promise does not have this field set. Suspense-enabled libraries should set the `status` field on the Promises they create to avoid unnecessary fallbacks. Calling `use` conditionally depending on whether a Promise is settled or not is discouraged. `use` should be called unconditionally so that React DevTools can show that the Component may suspend on data.
+
+<Sandpack>
```
rickhanlonii, line 453: Can we do something like this instead:

Screenshot 2025-11-04 at 9 31 49 AM


And structure it so that the most relevant code (the promise creation) is what's visible first? 

Like this you should see this:

## With a Promise subclass

```js
function getUser(id) {
  return {
    status: 'fulfilled',
    value: `User #${id}`,
    then: () => {}
  }
}


function UserDetails() {
  const user = use(getUser());
  return <p>Hello, {user}!</p>;
}
```

## Without a Promise subclass

```js
function getUser(id) {
  return Promise.resolve(`User #${id}`);
}


function UserDetails() {
  const user = use(getUser());
  return <p>Hello, {user}!</p>;
}
```

src/content/reference/react/use.md
```diff
@@ -436,6 +436,88 @@ To use the Promise's <CodeStep step={1}>`catch`</CodeStep> method, call <CodeSte
 
 ---
 
+### Avoiding fallbacks by passing Promise subclasses {/*avoiding-fallbacks-by-passing-promise-subclasses*/}
+
+React will read the `status` field on Promise subclasses to synchronously read the value without waiting for a microtask. If the Promise is already settled (resolved or rejected), React can read the value immediately without suspending and showing a fallback if the update was not part of a Transition (e.g. [`ReactDOM.flushSync()`](/reference/react-dom/flushSync)).
+
+React will set the `status` field itself if the passed Promise does not have this field set. Suspense-enabled libraries should set the `status` field on the Promises they create to avoid unnecessary fallbacks. Calling `use` conditionally depending on whether a Promise is settled or not is discouraged. `use` should be called unconditionally so that React DevTools can show that the Component may suspend on data.
```
Ephem: I think this is great content and very helpful! 

It's a pretty advanced topic though so maybe point out the intended audience early? Maybe lead with:

> If you are implementing a Suspense-enabled library, you can help React avoid unnecessarily suspending when you know the promise has already settled, by using `status` and `value` or `reason` fields.

I think "subclass" and "microtask" is likely language that will trip people up, but if it's clear this is an advanced topic I think that's fine.

Maybe be a bit more explicit about the rejected/reason case as well?

> Calling `use` conditionally depending on whether a Promise is settled or not is discouraged. `use` should be called unconditionally so that React DevTools can show that the Component may suspend on data.

I think this should be its own paragraph as it's a separate point.

src/content/reference/react/use.md
eps1lon: I don't think we want to show how to create a custom thenable.

sebmarkbage: Yea while you can use a custom Thenable, it's probably recommended to just use a native Promise for this.

You can make it a sub-class like such:

```js
class FulfilledPromise extends Promise {
  status = 'fulfilled';
  value = null;
  constructor(data) {
     super(resolve => resolve(data));
     this.value = data;
  }
}
const promise = new FulfilledPromise(data);
```

This makes it more palatable for sticklers that thinks adding expandos to native objects is bad. It shows that the pattern is encouraged by the platform. If it wasn't, then why can I sub-class natives? If can't add fields, what am I supposed to do with those sub-classes?

But in practice when you're not adding any methods to the prototype this is really just the same thing as:

```js
const promise = Promise.resolve(data);
promise.status = 'fulfilled';
promise.value = data;
```

Which is smaller code and faster, so why not do that? It also doesn't risk getting compiled which would break native subclassing.

So maybe document one stickler version and one faster version (and hint that thenables might be even faster).

src/content/reference/react/use.md
eps1lon: done

src/content/reference/react/use.md
eps1lon: done

src/content/reference/react/use.md
eps1lon: done

src/content/reference/react/use.md
```diff
@@ -436,6 +436,277 @@ To use the Promise's <CodeStep step={1}>`catch`</CodeStep> method, call <CodeSte
 
 ---
 
+### Avoiding fallbacks by passing Promise subclasses {/*avoiding-fallbacks-by-passing-promise-subclasses*/}
+
+If you are implementing a Suspense-enabled library, you can help React avoid unnecessarily suspending when you know the Promise has already settled, by using `status` and `value` or `reason` fields.
+
+React will read the `status` field on the Promise to synchronously read the value without having to wait for a microtask. If the Promise is already settled (resolved or rejected), React can read the value immediately without suspending and showing a fallback if the update was not part of a Transition (e.g. [`ReactDOM.flushSync()`](/reference/react-dom/flushSync)).
+
+React will set the `status` field itself if the passed Promise does not have this field set. Suspense-enabled libraries should set the `status` field on the Promises they create to avoid unnecessary fallbacks.
```
Ephem, line 445: I think this is good!

Maybe "... if the update was not part of a Transition (e.g. [`ReactDOM.flushSync()`](/reference/react-dom/flushSync))" is too much detail here? At least for me it made the sentence slightly harder to parse and didn't seem that important in this specific context, but not a big deal.

I wondered what it would look like if the docs tried to explain the basics more and noodled on a separate approach, I'm not at all sure this is better or captures everything correctly, so read it as an exploration:

```suggestion
In JavaScript, Promise callbacks always run in a "microtask", even if the Promise has already settled. This means that if you call `use(alreadySettledPromise)`, React can not synchronously resolve its values and has to unnecessarily suspend.

By adding `status: 'fulfilled' | 'rejected'` and either `value` or `reason` to the Promise when it settles, your library makes it possible for React to access these values synchronously and avoid suspending. It is recommended for Suspense-enabled libraries to do this.

React will set these fields itself if the passed Promise does not have them set.
```

src/content/reference/react/use.md
Ephem: I like this!

Many examples in the docs that compares approaches like this tend to point out what to look for, like "Note how loading the users shows a loading state", but I don't think that's necessary here.

src/content/reference/react/use.md
```diff
@@ -436,6 +436,277 @@ To use the Promise's <CodeStep step={1}>`catch`</CodeStep> method, call <CodeSte
...
+  const [userUsable, setUser] = useState(null);
+
+  return (
+    <div>
+      <button
+        onClick={() => {
+          // flushSync is only used for illustration
+          // purposes. A real app would probably use
+          // startTransition instead.
+          flushSync(() => {
```
Ephem, line 583: NIT: Is `flushSync` necessary for this to happen, or is it there to illustrate the difference to a `transition` clearer?

If it's not necessary, it might distract from the point more than help, at least for me it made me wonder if this was a _necessary_ condition?

I do think it's good to point out this does not happen with transitions though and not sure how to do that differently. 😄

src/content/reference/react/use.md
rickhanlonii: I like this, but I also like including examples of when this can happen. But I wouldn't use flush sync for that, I would use examples like "such as when the user clicks the back button, or when they continue interacting with the page while a transition is in progress"

Ephem: I like that. The preload case might be another one that's easy to understand and helps explain things.

src/content/reference/react/use.md
rickhanlonii: There's actually a much easier and common way for this to happen:

```js
export default function App() {
  const [userId, setUserId] = useState(null);
  const [userUsable, setUser] = useState(null);

  return (
    <div>
      <button
        onClick={() => {
          // ⚠️ this startTransition causes a sync update
          // to App to update the `isPending` value
          startTransition(() => {
            setUser(preloadUser(1));
            setUserId(1);
          });
        }}
      >
        Load User #1
      </button>
      {/* ... */}
      <span>{isPending && 'loading...'}</span>
      <Suspense key={userId} fallback={<p>Loading</p>}>
        {/* 
            ⚠️  Without proper memoization, the sync update 
            flows into this boundary, causing it to suspend
            if the data is not available syncronously
        */}
        {userUsable ? (
          <UserDetails userUsable={userUsable} />
        ) : (
          <p>No user selected</p>
        )}
      </Suspense>
    </div>
  );
}
```

https://codesandbox.io/p/sandbox/ymm9s8?file=%2Fsrc%2FApp.js

Ephem: I realize now that you might both trying to show practical examples of how this situation might happen in real code? I was focused on just the practical aspects of what's _necessary_ for it to trigger.

I was focused on the fact that even the first example triggers the suspend with just:

```tsx
      <button
        onClick={() => {
          setUser(preloadUser(2));
          setUserId(2);
        }}
      >
```

rickhanlonii: Yeah that's also true, good point, the flushSync isn't necessary in any of the examples to trigger this.

However, not using a transition here would be an anti-pattern since it would force the fallback if the data wasn't synchronously available (i.e. preloading didn't finish), so I think using a transition is still good here. 

https://react.dev/reference/react/Suspense#preventing-already-revealed-content-from-hiding

Ephem: > not using a transition here would be an anti-pattern 

Oh yeah, totally onboard with that.

> so I think using a transition is still good here

I don't have a strong opinion on that, in my mind it all depends on what the purpose of the examples are, which I'm not a good judge of.

* Tightly focused on explaining the `use()` behavior
  * Maybe better to keep them as minimal as possible to avoid distraction (but still point out in a comment this is an anti-pattern)
* Broader, try to explain common situations/why it's important etc
  * Makes sense to show a more "complete" example

Personally, I already know there are a lot of situations this might happen in, so I lean towards minimal, but the broader examples might be more helpful to others!

rickhanlonii: One point that's in favor of not having it is if you're using the Compiler, then the code works because everything is memoized correctly. But I still don't like the idea of synchronously setting the value your'e suspending on - you just shouldn't do that. 

Instead, a better example would be if there was something in `UserDetails` that was idiomatically a sync update that needed to re-render, like a text input. Then that update wouldn't be in a transition, and you'd need to handle it correctly so it doesn't just suspend the text input you're typing in on every key press.

rickhanlonii: Though in that case the Compiler also solves it :)

Ephem: The simplest case I can think of thats also common is likely just loading the page with prefetched data? Not sure if there is special behavior or if it makes for a good example though.

Ephem: If we are talking cases, wouldnt a transition where a new suspense boundary reveals with the use inside also have this problem? Fallback show if status is not set, never shows if it is?

rickhanlonii: Correct. And actually I just noticed the key so this example is good as is (with or without the transition) so I think we should just leave the transition in?

src/content/reference/react/use.md
```diff
@@ -436,6 +436,277 @@ To use the Promise's <CodeStep step={1}>`catch`</CodeStep> method, call <CodeSte
...
+
+function UserDetails({ userUsable }) {
+  const user = use(userUsable);
+  return <p>Hello, {user}!</p>;
+}
+
+export default function App() {
+  const [userId, setUserId] = useState(null);
+  // The initial
+  const [userUsable, setUser] = useState(null);
```
debel27, line 476: Not critical, but I have a few suggestions on this bit:
  - the comment is not very clear. I suggest to either remove it or adapt it.
  - when reading through the JSX, I was looking for a `setUserUsable` call. I realized later that the state and the setter functions were not named consistently. I suggest we make them consistent.
  - I assume you are using the "Usable" suffix to emphasize the state is not a raw promise? Should the "Promise" suffix be discouraged when naming variables provided to `use()`?

src/content/reference/react/use.md
eps1lon: > I like this, but I also like including examples of when this can happen. But I wouldn't use flush sync for that, I would use examples like "such as when the user clicks the back button, or when they continue interacting with the page while a transition is in progress"

That's not trivial to repro though. We have to make a concession between showing the observed behavior and when this actually applies. The back button wouldn't work in a docs demo. I don't know what the "when they continue interacting with the page while a transition is in progress" case is. Would you mind following up with a better example and we merge this so that we can link to something in the meantime=

src/content/reference/react/use.md
eps1lon: I don't think it helps showing a more involved example. This is good enough for me so that we have something to link to. Feel free to merge or include an example that helps understand the status field.

src/content/reference/react/use.md
Ephem: I agree! How I read this was that it was not about the examples, but about the text. Reading what Ricky wrote again I'm not sure?

I agree there is a concession to be made and that the observed behavior is most important. It's more important to know this _can_ happen (and you can look up how in other docs) as well as how to prevent it, rather than the exact situations it does.

That said, I don't think it hurts to add a sentence with an example of when a promise might already be resolved either to just give some feel for why it's important to handle this, maybe something like:

```
One case where a Promise might already have resolved is if the resource has been preloaded ahead of time, another is when the user triggers back in the browser and the Promise is cached.
```

I don't think it's terribly important though.

(Sidenote: Explaining all these intricacies also has value, but this section is short on space to do that. Some of the more involved things might be better for a "guide"-format, like on Suspenseful library integration?)

src/content/reference/react/use.md
```diff
@@ -436,6 +436,277 @@ To use the Promise's <CodeStep step={1}>`catch`</CodeStep> method, call <CodeSte
...
+<Sandpack>
+
+```js src/App.js active
+import { Suspense, use, useState } from "react";
+
+function preloadUser(id) {
+  // This is not a real implementation of getting the
+  // Promise for the user. A real implementation would
+  // probably call `fetch` or another data fetching method.
+  // The actual implementation should cache the Promise.
```
debel27, line 462: I think the docs do not stress on this requirement enough. Earlier, the "caveats" section says we should "_prefer_" creating Promises in Server components, which suggests that doing it in Client components is still OK (just less efficient). If caching Promises is in fact a hard requirement, mentioning it only here makes it burried in the fine print. I didn't fully process that comment until React started logging errors during development.

Once I discovered it, it was not obvious _why_ React requires cached Promises. After reading the docs for Promise subclasses, I (mistakenly) assumed that `use` did not care about Promise stability, as long as the `status`/`value`/`reason` properties evolved consistently accross renders (similar to how [`getSnapshot`](https://react.dev/reference/react/useSyncExternalStore#im-getting-an-error-the-result-of-getsnapshot-should-be-cached) works). I thought the following code was fine because `user` itself was stable and remained accessible from the `value` field:

```jsx
function Example() {
  const userUsable = getUser(1)
  const user = use(userUsable);
  return <p>Hello, {user}!</p>;
}

function getUser(id) {
  return PromiseWithStatus.resolve(`User #${id}`);
}
```

My guess is that React encourages caching Promises because there is no guarantee that a new Promise will keep the `status`/`value`/`reason` properties consistent with the previous Promise instance, let alone preserve them. Without this guarantee, React cannot detect if the new Promise still captures the same data (though it is not clear why React needs to know about this). 

Finally, I'm not sure how to interpret the error message logged by React:

> "_Creating promises inside a Client Component or hook is not yet supported, except via a Suspense-compatible library or framework._"

 - "_not yet supported_" -> so it will be supported in the future? How?
 - "_except via Suspense-compatible library_" -> what do those libraries do to make it work? If the "Promise subclasses" docs are addressed at library maintainers, it could be useful to explore this topic.

This is the thought process of someone who struggled trying `use` for the first time. I hope it helps :)

src/content/reference/react/use.md
debel27: Regarding the third point, I noticed the TypeScript/Flow definitions use `usable` as well:

```typescript
export function use<T>(usable: Usable<T>): T;
```

What do you think about aligning the docs with that terminology (e.g. `use(resource)` -> `use(usable)`)?

debel27 (commented):
I did the exercise of implementing a Suspense-enabled library based on the new docs. 

This was my very first attempt at using the `use` hook. I think there is an opportunity for the docs to organize information a bit differently. Here are a few suggestions.

## The rules of `use`

Addressing the rules of `use` straight on would provide a clear mental model for beginners.

For instance, there could be a summary of the rules in a "Usage with Promises" section:

  1. Promises must be cached (required).
  2. Promises should be augmented with the `status`/`value`/`reason` properties (recommended).
  3. `use` must not be called conditionally based on the `status` property (required)

The rules can then be followed by "deep dive" cards, explaining why they matter.

## Guidelines for library authors

Based on the rules of `use`, the docs could provide general guidelines for library authors to avoid pitfalls. Listing some examples below.

(these guidelines could also be put in the "Troubleshoting" section).

### Async functions should not return cached promises

Async functions wrap their return value with a new Promise, which won't have the `status`/`value`/`reason` properties.

```javascript
async function getUser(id) {
  return PromiseWithStatus.resolve(`User #${id}`);
}

const userUsable = getUser(1);
//    ^^^^^^^^^^ userUsable.status is `undefined`
```

### Immediately-settled promises must be Promise subclasses

Immediately-settled promises must be created as Promise subclasses from the start. This is the only opportunity we have to make their value synchronously available when `use` is called on the same call stack.

This implies that Suspense-enabled libraries must own the Promises they store in the cache. Otherwise, they will not be able to create a Promise subclass straight away: 

```typescript
import { fetchWithCache } from 'non-suspense-enabled-library';

function getUser(id) {
  const userPromise = fetchWithCache(`https://example.com/users/${id}`);
  // Problem: userPromise is not a Promise subclass.
  // Wrapping it here won't allow to synchronously
  // read its value in case it is already resolved.
  return PromiseWithStatus.resolve(promise);
}
```

PR #8128, opened 2025-11-05
fix(learn): Clear form error on new input in City Quiz examples

isaacbernat:
**What is the current behavior?**

In the "City quiz" examples found on both the `learn/managing-state` and `learn/reacting-to-input-with-state` pages, submitting a wrong answer causes an error message to appear. This error message then persists indefinitely, even when the user starts typing a new answer or resubmits the form, creating a confusing user experience.

Current_behavior

**What is the new behavior?**

This PR fixes the UX bug in both examples by clearing the `error` state whenever a new user action (`onChange` or `onSubmit`) is initiated. This ensures the UI provides timely and relevant feedback, which better reflects modern form handling best practices.

new_behavior

**How to test:**

1.  Navigate to the "City quiz" example on the `learn/managing-state` page.
2.  Submit an incorrect answer (e.g., "test"). The error message appears.
3.  Start typing a new answer in the text area.
4.  **Expected:** The error message immediately disappears.
5.  Repeat the test on the `learn/reacting-to-input-with-state` page to confirm the fix in both locations.

### DIFF
--- a/src/content/learn/managing-state.md
+++ b/src/content/learn/managing-state.md
@@ -35,6 +35,7 @@ export default function Form() {
   const [answer, setAnswer] = useState('');
   const [error, setError] = useState(null);
   const [status, setStatus] = useState('typing');
+  const showError = status === 'typing' && error && answer === error.failedAnswer;
 
   if (status === 'success') {
     return <h1>That's right!</h1>
@@ -49,6 +50,7 @@ export default function Form() {
     } catch (err) {
       setStatus('typing');
       setError(err);
+      setError({ message: err.message, failedAnswer: answer });
     }
   }
 
@@ -75,7 +77,7 @@ export default function Form() {
         }>
           Submit
         </button>
-        {error !== null &&
+        {showError &&
           <p className="Error">
             {error.message}
           </p>
--- a/src/content/learn/reacting-to-input-with-state.md
+++ b/src/content/learn/reacting-to-input-with-state.md
@@ -57,6 +57,7 @@ async function handleFormSubmit(e) {
 }
 
 function handleTextareaChange() {
+  hide(errorMessage);
   if (textarea.value.length === 0) {
     disable(button);
   } else {
@@ -414,6 +415,7 @@ export default function Form() {
   const [answer, setAnswer] = useState('');
   const [error, setError] = useState(null);
   const [status, setStatus] = useState('typing');
+  const showError = status === 'typing' && error && answer === error.failedAnswer;
 
   if (status === 'success') {
     return <h1>That's right!</h1>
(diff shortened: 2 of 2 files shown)


rickhanlonii:
I'm skeptical that this is much better, but in order to change it, we would need to:
a) update the vanilla example at the beginning to support this behavior
b) if this is the way the UX should be, the you shouldn't be able to "forget" to clear the error in every place (which is a point explained in this guide).

isaacbernat:
Thanks for the quick feedback @rickhanlonii . My primary goal with this PR is to fix what I believe is a UX bug: the error message persists incorrectly. This may confuse users and doesn't reflect modern form handling.

I agree with you that a `useEffect` would be the most robust, production-grade solution. However, `useEffect` hasn't been introduced at this stage of the tutorial, so using it here might break the learning flow for a beginner.

Therefore, I believe the best path forward is a simple, targeted fix that uses only the concepts already taught. I intent to:
a) Apply the equivalent fix to the initial vanilla JS example for behavioral consistency (good catch!).
b) Apply the `setError(null)` fix to the final React examples (as already committed in the PR).

This approach solves the UX bug while respecting the tutorial's step-by-step structure. I'll push these changes shortly unless you have further concerns.

Thanks again for the insights and raising great points about the implementation.

rickhanlonii:
I think if you can't address point b) I wouldn't spend any more time on this

isaacbernat:
@rickhanlonii Thanks for pushing for a better solution on point (b)! This led me to a fully declarative implementation which respects the tutorial's learning flow (no useEffect needed).

It uses a "smarter" error state that stores the failed answer. The decision to display the error is then derived on every render, checking if the current input matches the failed answer and the form isn't actively submitting.

This systemically solves the issue:
- The error disappears the moment the user types or resubmits.
- It requires no manual `setError(null)` calls, eliminating the anti-pattern you pointed out.
- It uses only the concepts introduced to the reader at this stage of the docs.

I've pushed up a single, amended commit that implements this improved logic consistently across all relevant examples.

This was a great learning experience. Let me know what you think.

isaacbernat:
@rickhanlonii , just checking in on this. I've rebased the branch on the latest main to resolve any potential conflicts. Let me know if anything else is needed from my side. Thanks!

isaacbernat:
@rickhanlonii
Hope you had a great holiday break! Just giving this a final, gentle bump to bring it back onto the radar.
I've just rebased the branch on the latest main so it's fresh and ready to go. However, I completely understand that priorities shift, so if this fix is no longer relevant or on the roadmap, please feel free to close this PR. No hard feelings at all!
Thanks again for the guidance on this one.

isaacbernat:
Hi @rickhanlonii , this is just a friendly bump since I think the fix is still relevant and haven't heard back. I've rebased the branch again on the latest main to resolve any potential conflicts. Please, tell me if you need something more from my side. Thanks!

PR #6523, opened 2023-12-28
fix wrong quotes around inline codes

2wheeh:
related PR: 
- #6522

~~The root cause is fixing on `remark-smartypants`: [here](https://github.com/silvenon/remark-smartypants/pull/80) 
But It's hard to tell when it gets merged so in the meanwhile this typo fix can be helpful.~~

**Edit**: The codes seemed to be referred to `silvenon/remark-smartypants` but we are not depending on that package. They resolved same issue recently so I utilize [their fixes](https://github.com/silvenon/remark-smartypants/pull/80) in this PR.

after: 

image

image

PR #8146, opened 2025-11-15
fix: Stop SmartyPants from altering TerminalBlock commands (like `--save-dev` to `—save-dev`)

### LINKED ISSUE #8140: [Bug]: SmartyPants replaces `--` into an em-dash in <TerminalBlock>
### Summary

In https://react.dev/learn/typescript#adding-typescript-to-an-existing-react-project :

Image

As you can see, `--` has been converted to `—` (a non-ASCII em-dash). This happens during the site build step, so when someone copies this and pastes it into a terminal, they get an error.

The root cause is this file:

https://github.com/reactjs/react.dev/blob/b440d6698f6e21d56a78b10f625bd23191183588/plugins/remark-smartypants.js

It applies SmartyPants to all `text` AST nodes in Markdown. Since `<TerminalBlock>` is [implemented using only simple `<div>`s](https://github.com/reactjs/react.dev/blob/b440d6698f6e21d56a78b10f625bd23191183588/src/components/MDX/TerminalBlock.tsx) and the plugin does not skip contents aside from `<style>` and `<script>`, the plugin converts `--save-dev` to `—save-dev` in `<TerminalBlock>`.

I think the fix would involve two steps:

1. Use `<pre>` or `<code>` appropriately to render the contents of `<TerminalBlock>`
2. In the plugin above, skip applying SmartyPants not only when the parent tag is `<script>`/`<styule>` but also when it's `<code>` or `<pre>`

smikitky:
This fixes #8140.

With this change, contents in `<TerminalBlock>` will no longer be processed by SmartyPants. Previously, `--` was converted to en-dashes, so `--save-dev` looked like `—save-dev` (with an en-dash) unless each snippet was manually escaped using a hacky and nonintuitive syntax. This PR will prevent this type of bugs once and for all for contributors.

Before (current problem in https://react.dev/learn/typescript):

image

After:

image

### Change summary

- Edited the `remark-smartypants.js` plugin so that it no longer transforms `<TerminalBlock>` content. This is done by tagging its subtrees with `data.skipSmartyPants` and bailing out before running the typography pass.
- Render TerminalBlock output with `<pre><code>` instead of relying on generic `<div>` wrappers. This was not strictly necessary to fix the current bug, but I believe it's desirable.
- Clean up every existing TerminalBlock usage by stripping the hacky ``{`…`}`` workaround
- Bump `DISK_CACHE_BREAKER`

### DIFF
--- a/src/utils/compileMDX.ts
+++ b/src/utils/compileMDX.ts
@@ -10,7 +10,7 @@ import {MDXComponents} from 'components/MDX/MDXComponents';
 
 // ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 // ~~~~ IMPORTANT: BUMP THIS IF YOU CHANGE ANY CODE BELOW ~~~
-const DISK_CACHE_BREAKER = 10;
+const DISK_CACHE_BREAKER = 11;
 // ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 export default async function compileMDX(


smikitky:
@rickhanlonii Looks like you were the last person to review the code in this area (in #6523). Could you take a look at this as well?

PR #8162, opened 2025-11-26
Specify that Effects run on commit, not render

eps1lon:
[preview: useEffect reference docs ](https://react-5wbg1cutd-fbopensource.vercel.app/reference/react/useEffect#parameters)

I had two instances recently where people were confused by useEffect not running after each render (https://github.com/facebook/react/issues/35221 and https://github.com/facebook/react/issues/34730). I think the docs aren't super clear about this and you need a bit of a React brain to infer the actual behavior.

In the learn section where we introduce the render and commit phase, we don't talk about Effects: https://react.dev/learn/render-and-commit. That seems fine to me since Effects are an advanced concept. 

However, the `useEffect` reference docs, say that the setup runs after "render" not after "commit": https://react.dev/reference/react/useEffect#parameters

I went through all mentions of `render` in the Effect reference docs and replaced it with "commit" when it talked about the timing of the Effect execution.


src/content/reference/react/useEffect.md
eps1lon, line 1771: I kept this as "re-render" instead of commit since I suspect people troubleshooting this particular issue search more often with "render" than commit.

PR #8171, opened 2025-12-03
fix: use correct function name for addTransitionType

FelixTraxler:
The docs for [addTransitionType](https://react.dev/reference/react/addTransitionType) say that:

Reference
`addTransitionType `
...
Returns
`startTransition` does not return anything

but it should say

`addTransitionType ` does not return anything


eps1lon (approved):
thanks!

PR #8238, opened 2026-01-13
Clarify when Effects are needed for data fetching (#8226)

### LINKED ISSUE #8226: [Mistake]: Wrong example in fetching data section on You Might not need a Effect Page
### Summary

In the [Fetching Data Section](https://react.dev/learn/you-might-not-need-an-effect#fetching-data) of [You might not need an effect](https://react.dev/learn/you-might-not-need-an-effect) is solving for different scenario with different problem altogether which will a create a confusion whether it requires a effect or not

### Page

https://react.dev/learn/you-might-not-need-an-effect#fetching-data

### Details

The [Fetching Data Section](https://react.dev/learn/you-might-not-need-an-effect#fetching-data) of [You might not need an effect](https://react.dev/learn/you-might-not-need-an-effect) is solving a different scenario with a different problem altogether.

In this code block:
```js
function SearchResults({ query }) {
  const [results, setResults] = useState([]);
  const [page, setPage] = useState(1);
  useEffect(() => {
    // 🔴 Avoid: Fetching without cleanup logic
    fetchResults(query, page).then(json => {
      setResults(json);
    });
  }, [query, page]);
  function handleNextPageClick() {
    setPage(page + 1);
  }
  // ...
}
```

The documentation mentions that:
> This might seem like a contradiction with the earlier examples where you needed to put the logic into the event handlers! However, consider that it's not the typing event that's the main reason to fetch. Search inputs are often prepopulated from the URL, and the user might navigate Back and Forward without touching the input.
> 
> It doesn't matter where page and query come from. While this component is visible, you want to keep results [synchronized](https://react.dev/learn/synchronizing-with-effects) with data from the network for the current page and query. This is why it's an Effect.

This encourages people to add `page` as a dependency in the `useEffect` and also categorize the page change event as an effect altogether. But page changes are user events (wrt to the context like the results are shown in a table), and the documentation should encourage people to fetch data inside the change handler rather than using an effect (similar to how we encourage handling [POST request on a form](https://react.dev/learn/you-might-not-need-an-effect#sending-a-post-request)).

Regarding the URL scenario: let's say the user goes to page 3, navigates to a different page, and comes back again. The URL page param should be read inside the effect callback like this:

```js
function SearchResults({ query }) {
  const [results, setResults] = useState([]);
  const [page, setPage] = useState(new URLSearchParams(window.location.search).get("page"));
  
  useEffect(() => {
    const urlSearchParams = new URLSearchParams(window.location.search);
    const currentPage = urlSearchParams.get("page");
    fetchResults(query, currentPage).then(json => {
      setResults(json);
    });
  }, [query]);
  
  function handleNextPageClick() {
    setPage(page + 1);
    fetchResults(query, page + 1).then(json => {
      setResults(json);
    });
  }
  // ...
}
```

This way, `page` can be removed as a dependency altogether, and the effect only runs whenever the `query` changes while always getting the latest page param value. When the page is changed by the user, the fetch should be invoked inside `handleNextPageClick` since page change is an event.

The earlier example will result in more confusion, and people will add things that should be inside the event handler into the effect with dependencies, making [Event handlers sharing logic](https://react.dev/learn/you-might-not-need-an-effect#sharing-logic-between-event-handlers) obsolete too. This also make the people aware on avoiding the "setState triggers effect" pattern for user events and follows the same principle on removing effects and using events properly as the article intended.

kanikaa88:
This PR improves the explanation in the "Fetching data" section of "You Might Not Need an Effect" to clarify when an Effect is needed versus when it's not.

The previous text was confusing because it said "You don't need to move this fetch to an event handler" which seemed contradictory given the page's title. This change:

- Clearly states that this IS a case where an Effect IS needed (for synchronization with external system)
- Explains the key difference: Effects are needed when synchronizing with external systems, regardless of how the state changes
- Adds a contrast example showing when you would use an event handler instead

This makes it clearer for readers to understand when Effects are appropriate for data fetching.

Closes #8226

### DIFF
--- a/src/content/reference/react/Component.md
+++ b/src/content/reference/react/Component.md
@@ -1019,6 +1019,12 @@ Implementing `static getDerivedStateFromProps` in a class component is equivalen
 
 </Note>
 
+<Note>
+
+**Note:** The `static getDerivedStateFromProps` method can also be used with components created via the `createClass` API (using the [`create-react-class`](https://www.npmjs.com/package/create-react-class) package). This allows legacy codebases to adopt the newer lifecycle pattern.
+
+</Note>
+
 ---
 
 ## Usage {/*usage*/}


src/content/reference/react/Component.md
```diff
@@ -1019,6 +1019,12 @@ Implementing `static getDerivedStateFromProps` in a class component is equivalen
 
 </Note>
 
+<Note>
+
+**Note:** The `static getDerivedStateFromProps` method can also be used with components created via the `createClass` API (using the [`create-react-class`](https://www.npmjs.com/package/create-react-class) package). This allows legacy codebases to adopt the newer lifecycle pattern.
```
awot-b, line 1024: The mention of legacy create-react-class and getDerivedStateFromProps feels out of scope for modern Hooks-based guidance. To directly address #8226, we should instead focus the example on the core distinction: using an Effect for Initial Sync versus an Event Handler for Pagination. This keeps the mental model clean and relevant.

PR #8243, opened 2026-01-17
docs: explain stale state in request counter example (fixes #8230)

### LINKED ISSUE #8230: It may allow beginners to taste the sweetness of not following the rules
### Summary

https://react.dev/learn/queueing-a-series-of-state-updates

The first challenge in this chapter can still work normally even if it is written like this, because it relies on click behavior
```js
async function handleClick() {
    setPending(pending + 1);
    await delay(3000);
    setPending(pending=>pending - 1);
    setCompleted(completed=>completed + 1);
  }
```

However, if it is changed to this, it will not work. At this point, we must change setPending(pending + 1) to setPending(pending => pending + 1)
```js
async function handleClick() {
    setTimeout(()=> {setPending(pending + 1);}, 1000)
    await delay(3000);
    setPending(pending=>pending - 1);
    setCompleted(completed=>completed + 1);
  }
```

### Page

https://react.dev/learn/queueing-a-series-of-state-updates

### Details

Perhaps some annotations could be added

2024si96556-Kishori:

### DIFF
--- a/src/content/learn/queueing-a-series-of-state-updates.md
+++ b/src/content/learn/queueing-a-series-of-state-updates.md
@@ -2,22 +2,22 @@
 title: Queueing a Series of State Updates
 ---
 
-`<Intro>`{=html}
+<Intro>
 
 Setting a state variable will queue another render. But sometimes you
 might want to perform multiple operations on the value before queueing
 the next render. To do this, it helps to understand how React batches
 state updates.
 
-`</Intro>`{=html}
+</Intro>
 
-`<YouWillLearn>`{=html}
+<YouWillLearn>
 
 -   What "batching" is and how React uses it to process multiple state
     updates
 -   How to apply several updates to the same state variable in a row
 
-`</YouWillLearn>`{=html}
+</YouWillLearn>
 
 ## React batches state updates {/*react-batches-state-updates*/}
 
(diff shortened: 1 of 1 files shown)


kanikaa88 (commented):
Thanks for the PR — a few small fixes please to avoid rendering/Sandpack breakage:

Please revert wrapping MDX components in backticks with {=html}. Examples: <Sandpack>, <Intro>, <YouWillLearn>, <Note>, <Recap>, <Solution>, <Hint>, <Challenges>. They should remain real MDX components.
Fix the mangled code fences (e.g. processQueue): keep the fence metadata on its own line and the code underneath (don’t put the fence and code on the same line).
Remove the stray backslash before the heading (change \### What happens... to ### What happens...).
Don’t wrap <Note> and similar components inside code fences — render them as components or use the repo’s admonition pattern.
Prefer pipe-style Markdown tables over preformatted text, and run the repo formatter (prettier/markdownlint) to normalize line-wrapping and list indentation.
After making these changes, please run the local/CI site preview to verify Sandpack examples and MDX components render correctly.
Thanks — happy to re-review once those are addressed.

2024si96556-Kishori:
Thanks for the detailed feedback!

I’ve fixed the MDX component rendering, corrected the Sandpack code fences, removed the stray backslash in the heading, converted the tables to pipe-style Markdown, and verified the page renders correctly.

Could you please take another look when convenient? Thanks again!

Issue #7737, opened 2025-04-14
[Suggestion]: Can you add a paragraph saying useOptimistic updater function should be called in transition?


gmoniava:
### Summary

Here:

`  const [optimisticState, addOptimistic] = useOptimistic(state, updateFn);`

When you call `updateFn ` without transition react gives warning:

> Warning: An optimistic state update occurred outside a transition or action. To fix, move the update to an action, or wrap with startTransition.

### Page

https://react.dev/reference/react/useOptimistic

### Details

Can you also add this information (which is in the warning) in the docs so that it is not only as a warning in the console?
The example in the docs may further confuse people because `formAction` is not wrapped in transition (but apparently react does that for you when it is passed as `action` prop).
Because I see people using it without transitions.

gmoniava:
@poteto  Sorry for direct tag - but would be nice if someone can have a look

rickhanlonii:
Yeah we need to rewrite all of the `useOptimistic` docs.

MaxwellCohen:
A clear warning ought to be added, but in addition, notes/sections should be added throughout the page

some ideas: 

- Pifall warning section with the console warning
- In the returns section 
https://github.com/reactjs/react.dev/blob/2da4f7fbd90ddc09835c9f85d61fd5644a271abc/src/content/reference/react/useOptimistic.md?plain=1#L53
Add a few words like the suggested bold text here 
`addOptimistic`: `addOptimistic` is the dispatching function to call when you have an optimistic update in **a transition or form action**. It takes one argument, `optimisticValue`, of any type and will call the `updateFn` with `state` and `optimisticValue`.
- linking to the resulting documentation of the Async React Working Group.
-  code example(s) **explicitly** calling that addOptimistic needs to be called in a transition.
 

I am very excited about the useOptimistic and other Async React work. This is a shift in thinking that will have lots of little paper cuts, so it is important to have doc/warnings/tools be explicit about how to use them.  I hope this is helpful.

rickhanlonii:
Thanks @MaxwellCohen, I think I address all that feedback and the original suggest in https://github.com/reactjs/react.dev/issues/7737!

gmoniava:
@rickhanlonii Ricky I want to use an opportunity and ask you is anyone trying to address this: https://github.com/reactjs/react.dev/issues/7984? don't you think it raises valid points?

Issue #7379, opened 2024-12-20
[Mistake]: The demo of useOptimistic may occur an incorrect rendering  


felismargarita:
### Summary

useOptimistic demo merges state via [...state, { text: newMessage, sending:true }], which may render a same message twice in the page at the same time(although it will render correct page immediately after the incorrect one)

### Page

https://react.dev/reference/react/useOptimistic

### Details

merge message logic in this demo has to handle the situation: 
state(array of the messages) already includes the sending message when the setMessages is called to make the page re-rendering.

karthik-awebon:
Hello @felismargarita the existing code effectively demonstrates the use of the UseOptimistic Hook with a simple example. Adding unnecessary complexity to handle duplicate messages detracts from the core purpose of the demo.

Additionally, your PR does not function as intended. The use of time as a unique identifier for messages is flawed, as identical messages can still appear multiple times if their timestamps differ, as evidenced in the screenshot.
image
![image](https://github.com/user-attachments/assets/fc50db22-80c5-4ca9-8d4e-d520e195e7d4)

rickhanlonii:
Should be fixed in https://github.com/reactjs/react.dev/pull/8264

Issue #7098, opened 2024-08-08
[Omission]: useOptimistic is documented as having a required second parameter 


denis-sokolov:
### Summary

useOptimistic’s second parameter is allegedly optional ([source](https://github.com/facebook/react/issues/30638#issuecomment-2276652326)), but the docs imply that it is required.

### Page

https://react.dev/reference/react/useOptimistic#parameters

### Details

Perhaps in the Reference section, Parameters, under updateFn, it could be mentioned that the function is optional, and that the returned function addOptimistic will instead be a setter (setOptimistic).

AkashJana18:
Hi @denis-sokolov ,
I wanted to follow up on this pull request and see if there's anything else you need from me or any additional changes you'd like to see. Please let me know if you have any feedback or if there's a timeline for review.
Thanks for your time and consideration!

Issue #6926, opened 2024-06-01
[Mistake]: Unused key in useOptimistic code sample


lnicepei:
### Summary

The code example for `useOptimistic` hook contains unused `key` parameter

### Page

https://19.react.dev/reference/react/useOptimistic#usage

### Details

In the [`Optimistically updating forms` section](https://19.react.dev/reference/react/useOptimistic#usage), inside the code sample there is [an unused `key` parameter](https://codesandbox.io/p/sandbox/nostalgic-cdn-hsvs2d?file=%2Fsrc%2FApp.js%3A40%2C1-40%2C53&utm_medium=sandpack), which, apparently, should have been used [on the line 25](https://codesandbox.io/p/sandbox/nostalgic-cdn-hsvs2d?file=%2Fsrc%2FApp.js%3A25%2C1-25%2C26&utm_medium=sandpack):

```ts
<div key={index}>
```

Wellitsabhi:
It's not a mistake, 'key' parameter is used in reconciliation to optimize render performance . And it's right way of writing  code when using  ```.map```

PR #8264, opened 2026-01-26
Revamp useOptimistic docs

rickhanlonii:
### [Preview](https://react-dev-git-fork-rickhanlonii-rh-use-opti-c506a1-fbopensource.vercel.app?_vercel_share=YHkrW0qM96Xt84w4gIUI81Yy9AqsbjCA)

Fixes https://github.com/reactjs/react.dev/issues/7737
Fixes https://github.com/reactjs/react.dev/issues/7379
Fixes https://github.com/reactjs/react.dev/issues/7098
Fixes https://github.com/reactjs/react.dev/issues/6926

### DIFF
--- a/src/content/reference/react/useOptimistic.md
+++ b/src/content/reference/react/useOptimistic.md
@@ -83,7 +83,7 @@ function handleClick() {
 
 #### How optimistic state works {/*how-optimistic-state-works*/}
 
-`useOptimistic` lets you show a temporary value while a Transition is running:
+`useOptimistic` lets you show a temporary value while a Action is in progress:
 
 ```js
 const [value, setValue] = useState('a');
@@ -91,27 +91,38 @@ const [optimistic, setOptimistic] = useOptimistic(value);
 
 startTransition(async () => {
   setOptimistic('b');
-  setValue('b');
+  const newValue = await saveChanges('b');
+  setValue(newValue);
 });
 ```
 
-When the setter is called inside an Action, `useOptimistic` will trigger a re-render to show that value while the Action is in progress. Otherwise, the `value` passed to `useOptimistic` is returned.
+When the setter is called inside an Action, `useOptimistic` will trigger a re-render to show that state while the Action is in progress. Otherwise, the `value` passed to `useOptimistic` is returned.
 
 This state is called the "optimistic" because it is used to immediately present the user with the result of performing an Action, even though the Action actually takes time to complete.
 
-When the Action completes, `useOptimistic` returns the current `value` you passed in.
-
 **How the update flows**
 
-1. **Immediate optimistic update**: When `setOptimistic('b')` is called, `optimistic` immediately becomes `'b'` and React renders with this state.
+1. **Update immediately**: When `setOptimistic('b')` is called, React immediately renders with `'b'`.
 
-2. **Transition scheduled**: `setValue('b')` schedules an update to the real state, but this update won't commit until the Action completes.
+2. **(Optional) await in Action**: If you await in the Action, React continues showing `'b'`.
 
-3. **Optimistic state persists during async work**: If the Transition suspends or uses `await`, the optimistic `'b'` continues showing while React waits.
+3. **Transition scheduled**: `setValue(newValue)` schedules an update to the real state.
 
-4. **Single render commit**: When all async work finishes, the new state (`'b'`) and optimistic state (`'b'`) commit together in a single render.
+4. **(Optional) wait for Suspense**: If `newValue` suspends, React continues showing `'b'`.
+
+5. **Single render commit**: Finally, the `newValue` is commits for `value` and `optimistic`.
 
 There's no extra render to "clear" the optimistic state. The optimistic and real state converge in the same render when the Transition completes.
+
+<Note>
+
+#### Optimistic state is temporary {/*optimistic-state-is-temporary*/}
+
+Optimistic state is only renders while an Action is in progress, otherwise `value` is rendered.
+
+If `saveChanges` returned `'c'`, then both `value` and `optimistic` will be `'c'`, not `'b'`.
+
+</Note>
 
 **How the final state is determined**
 


src/content/reference/react/useOptimistic.md
```diff
@@ -18,114 +18,1029 @@ title: useOptimistic
...
+const [value, setValue] = useState('a');
+const [optimistic, setOptimistic] = useOptimistic(value);
+
+startTransition(async () => {
+  setOptimistic('b');
+  setValue('b');
+});
+```
+
+When the setter is called inside a transition, `useOptimistic` will render that value while the transition is in progress. Otherwise, the `value` passed to `useOptimistic` is rendered.
```
gaearon: I think the key thing to communicate here is that this "setting" itself is temporary. This is very different from `useState` API where setting is permanent.

Maybe you can flip it around and say explicitly that calling `setOptimistic` sets the value *for the duration of that transition*, "shadowing" the real value.

MaxwellCohen: I agree there is some confusion about when and why to use useOptimistic. Overall, these docs are much better than before.

Could creating a chart of some type help explain this better? 

This is my attempt at a flowchart of this (with the help of LLMs). I feel something is missing or could be better explained.
```mermaid
flowchart LR
    A["`Component renders<br/>(optimistic value = value passed to useOptimistic)`"]
    A --> T_start
    
    subgraph Transition["`Transition`"]
        T_start["`Transition or Action starts`"]
        T_start --> C["`setOptimistic called with an optimistic value`"]
        C --> D["`Component renders<br/>(optimistic value = value calcuated from setOptimistic)`"]
        D --> E["`Async action runs<br/>(submit data, wait for response)`"]
        E -->|Success| G["`State is updated from the async action`"]
        E -->|Failure| H["`no value passed to useOptimistic is updated`"]
        G --> T_end["`Transition or action ends`"]
        H --> T_end
    end
    
    T_end --> O["`Component renders<br/>(optimistic value = value passed to useOptimistic)`"]
```

gaearon: I think I'd shorten this entire thing and the flows, it makes it sound more complicated than it is. Really all we're doing here is letting you shadow a value for the duration of a transition.

rickhanlonii: Maybe this deep dive is too high up? In the first usage example "Adding optimistic state to a component" it explains that it "lets you temporarily change the state during an Action or Transition" with the setter. In this deep dive my goal is to explain:

- the optimistic state is shown immediately
- it's shown during async actions and transitions
- it automatically "reverts" (but avoiding the "revert" jargon) 
- the "revert" happens in the same paint as the transition, not additionally

I could probably delete the "How the final state is determined" and everything after it. But maybe the "revert on delete" is good.

gaearon: oh I didn't realize it's a deep dive. i was looking at GH rendering of the preview. since the actual preview doesn't work and is behind some kind of access control

gaearon: lemme just check it out locally

src/content/reference/react/useOptimistic.md
```diff
@@ -18,114 +18,1029 @@ title: useOptimistic
...
+  const [optimisticAge, setOptimisticAge] = useOptimistic(age);
+  const [optimisticName, setOptimisticName] = useOptimistic(name);
+  const [optimisticTodos, setOptimisticTodos] = useOptimistic(todos, reducer);
+  // ...
+```
+
+`useOptimistic` returns an array with exactly two items:
+
+1. The <CodeStep step={2}>current state</CodeStep> of the optimistic value, returning the <CodeStep step={1}>state</CodeStep> provided.
+2. The <CodeStep step={3}>set function</CodeStep> that lets you temporarily change the state during an Action or Transition.
```
gaearon: wrong code step for reducer

src/content/reference/react/useOptimistic.md
```diff
@@ -18,114 +18,1029 @@ title: useOptimistic
...
+
+The `set` function returned by `useOptimistic` lets you update the state to a different value inside an Action or Transition and trigger an immediate re-render. You can pass the next state directly, or a function that calculates it from the previous state:
+
+```js
+const [optimisticLike, setOptimisticLike] = useOptimistic(name);
+
+function handleClick() {
+  startTransition(async () => {
+    setOptimisticLike(true);
+    setOptimisticLike(liked => !liked);
```
gaearon: showing both modifying same variable in one code snippet is confusing, let's pick one to show here

src/content/reference/react/useOptimistic.md
```diff
@@ -18,114 +18,1029 @@ title: useOptimistic
...
+`useOptimistic` returns an array with exactly two values:
+
+1. `optimisticState`: The resulting optimistic state. It is equal to `state` unless an Action or Transition is pending, in which case it is equal to the value returned by `reducer` (or the value passed to the set function if no `reducer` was provided).
+2. The [`set` function](#setoptimistic) that lets you update the optimistic state to a different value inside an Action or Transition.
+
+---
+
+### `set` functions, like `setOptimistic(optimisticValue)` {/*setoptimistic*/}
+
+The `set` function returned by `useOptimistic` lets you update the state to a different value inside an Action or Transition and trigger an immediate re-render. You can pass the next state directly, or a function that calculates it from the previous state:
```
gaearon: i feel like the bigger point to emphasize here is "for the duration of a transition" rather than "trigger a re-render". i'm not sure in general that we need to be thinking in terms of describing what triggers a re-render as much as what's displayed on the screen.

src/content/reference/react/useOptimistic.md
```diff
@@ -18,114 +18,1029 @@ title: useOptimistic
...
+```js [[3, 3, "setOptimisticAge"]]
+function handleClick(e) {
+  startTransition(() => {
+    setOptimisticAge(42);
+    onAgeChange(42);
+  });
+}
+```
+
+Or call it inside an action:
```
gaearon: if "or" is needed, that's a separation subsection in Usage imo. we're trying to describe each canonical scenario

src/content/reference/react/useOptimistic.md
```diff
@@ -18,114 +18,1029 @@ title: useOptimistic
...
+  const [optimisticName, setOptimisticName] = useOptimistic(name);
+  const [optimisticTodos, setOptimisticTodos] = useOptimistic(todos, reducer);
+  // ...
+```
+
+`useOptimistic` returns an array with exactly two items:
+
+1. The <CodeStep step={2}>current state</CodeStep> of the optimistic value, returning the <CodeStep step={1}>state</CodeStep> provided.
+2. The <CodeStep step={3}>set function</CodeStep> that lets you temporarily change the state during an Action or Transition.
+   * If a <CodeStep step={1}>reducer</CodeStep> is provided, it will run before rendering the temporary state.
```
gaearon: similarly, this feels like a separate section to me.

src/content/reference/react/useOptimistic.md
```diff
@@ -18,114 +18,1029 @@ title: useOptimistic
...
+2. The [`set` function](#setoptimistic) that lets you update the optimistic state to a different value inside an Action or Transition.
+
+---
+
+### `set` functions, like `setOptimistic(optimisticValue)` {/*setoptimistic*/}
+
+The `set` function returned by `useOptimistic` lets you update the state to a different value inside an Action or Transition and trigger an immediate re-render. You can pass the next state directly, or a function that calculates it from the previous state:
+
+```js
+const [optimisticLike, setOptimisticLike] = useOptimistic(name);
```
brenelz: This should probably be `useOptimistic(false);` instead of `useOptimistic(name);`?

src/content/reference/react/useOptimistic.md
gaearon: i see, you already have a section like this below. so imo it's fine to now emphasize this here. you can do another step-by-step later

src/content/reference/react/useOptimistic.md
```diff
@@ -18,114 +18,1029 @@ title: useOptimistic
...
+import { updateName } from './actions.js';
+
+export default function EditName({ name, setName }) {
+  const [optimisticName, setOptimisticName] = useOptimistic(name);
+
+  async function submitAction(formData) {
+    const newName = formData.get('name');
+    setOptimisticName(newName);
+    const updatedName = await updateName(newName);
+    setName(updatedName);
```
gaearon: does this setName need to be wrapped in startTransition due to lack of async context?

gaearon:
added some initial nits. examples add a lot of clarity, so great work on those. i'd just slightly trim everything and focus it down so each part explains one aspect

src/content/reference/react/useOptimistic.md
```diff
@@ -18,114 +18,1029 @@ title: useOptimistic
...
+* `nextState`: The value that you want the optimistic state to be. If you provided a `reducer` to `useOptimistic`, this value will be passed as the second argument to your reducer. It can be a value of any type.
+    * If you pass a function as `nextState`, it will be treated as an _updater function_. It must be pure, should take the pending state as its only argument, and should return the next state. React will put your updater function in a queue and re-render your component. During the next render, React will calculate the next state by applying the queued updaters to the previous state similar to [`useState` updaters](/reference/react/useState#setstate-parameters)
+
+#### Returns {/*setoptimistic-returns*/}
+
+`set` functions do not have a return value.
+
+#### Caveats {/*setoptimistic-caveats*/}
+
+* The `set` function must be called inside a [Transition](/reference/react/useTransition) using [`startTransition`](/reference/react/startTransition) or inside an [Action](/reference/react/useTransition#perform-non-blocking-updates-with-actions). If you call the setter outside an Action or Transition, the optimistic value will briefly appear and then immediately revert.
```
brenelz: does the `set` function warn if called outside a transition? I haven't really tried it

src/content/reference/react/useOptimistic.md
MaxwellCohen: For form actions `<form action={submitAction}>`, a startTransition is not needed  https://react.dev/reference/react-dom/components/form because of the action props. A "flag comment" about startTranstion not need could be helpful

src/content/reference/react/useOptimistic.md
rickhanlonii: The confusion I've seen is that people think they need to wrap `setOptimistic` in `startTransition` even if they're in an Action callback. There are Action specific examples below, like "Using optimistic state in Action props", but for this first example I wanted to mention both so if you only read this one you can see both forms. Maybe I should just hoist the Action prop example up to right after this example?

btw, this doc uses "Action or Transition" everywhere which I hate. What are your thoughts on Action vs Transition. Part of me wants to just say Action throughout this doc, since a function inside a Transition is an Action, but I'm worried that "Action" is not a well-understood enough concept to go all in that way.

rickhanlonii: If this doc was Async React first, it would start with "Using optimistic state in Actions", and then have "Implementing Action props with Optimistic state". You really should only need to add your own startTransition when you're implementing an `action` prop.

src/content/reference/react/useOptimistic.md
rickhanlonii: @MaxwellCohen a diagram would be good, but I'm going to punt to later since they're a pain to make

src/content/reference/react/useOptimistic.md
rickhanlonii: I stole this from the useState docs fwiw

Screenshot 2026-01-27 at 2 26 27 PM

But that uses different state variables, I'll update this to use different ones.

gaearon: right yea with different it's fine

src/content/reference/react/useOptimistic.md
gaearon: generally from a code example i want to see enough code that i can copy paste it and it works. the "action" concept is indeed a bit vague because it isn't anchored to any api, and there's also no Learn page for it. so it's hard to guess by looking what makes it an action

gaearon: if you want to show only actions i think that's fine but then you have to show the `action` prop implementation too so that i can see the whole pattern ... in action hahaah

src/content/reference/react/useOptimistic.md
gaearon: i mean the async gap — i don't think react has any way to tie setStates after await back to the cause. asynccontext will solve this but it isn't there yet

src/content/reference/react/useOptimistic.md
rickhanlonii: > it's fine to now emphasize this here

do you mean keep this or remove it?

src/content/reference/react/useOptimistic.md
rickhanlonii: nice catch

src/content/reference/react/useOptimistic.md
rickhanlonii: Good call, I updated this to say "React will show a warning" with a link to Troubleshooting below with the actual warning.

src/content/reference/react/useOptimistic.md
rickhanlonii: Yeah, moved it to the parent though so this can focus on the setOptimistic

src/content/reference/react/useOptimistic.md
rickhanlonii: Updated: hoisted up the action prop example, and moved the "Or call it inside an action:" to a Note that points to the next example.

src/content/reference/react/useOptimistic.md
rickhanlonii: Updated!

src/content/reference/react/useOptimistic.md
```diff
@@ -18,114 +18,1044 @@ title: useOptimistic
...
+const [value, setValue] = useState('a');
+const [optimistic, setOptimistic] = useOptimistic(value);
+
+startTransition(async () => {
+  setOptimistic('b');
+  setValue('b');
+});
+```
+
+When the setter is called inside a transition, `useOptimistic` will render that value while the transition is in progress. Otherwise, the `value` passed to `useOptimistic` is rendered.
```
stephan-noel: **nit**:  Is it consistent usage to say a hook _renders_ a value?  It _returns_ a value, and that value usually results in it being rendered for state-related hooks like this one, but the usage felt off to me.

Later usage also uses _return_: 

> When the action or transition completes, `useOptimistic` returns the current `value` you passed in.

Too pedantic?

src/content/reference/react/useOptimistic.md
```diff
@@ -18,114 +18,1044 @@ title: useOptimistic
...
+
+1. **Immediate optimistic update**: When `setOptimistic('b')` is called, `optimistic` immediately becomes `'b'` and React renders with this value.
+
+2. **Transition scheduled**: `setValue('b')` schedules an update to the real state, but this update won't commit until the transition completes.
+
+3. **Optimistic state persists during async work**: If the transition suspends or uses `await`, the optimistic `'b'` continues showing while React waits.
+
+4. **Single render commit**: When all async work finishes, the new state (`'b'`) and optimistic state (`'b'`) commit together in a single render.
+
+There's no extra render to "clear" the optimistic value—the optimistic and real state converge in the same render when the transition completes.
```
stephan-noel: **nit**: "-" punctuation.  Seems tight without spaces, or maybe a comma would do just as well.  I'm no grammar expert though.

src/content/reference/react/useOptimistic.md
```diff
@@ -18,114 +18,1044 @@ title: useOptimistic
...
+
+- **Hardcoded values** like `useOptimistic(false)`: After the transition, `state` is still `false`, so the UI shows `false`. This is useful for pending states where you always start from `false`.
+
+- **Props or state passed in** like `useOptimistic(isLiked)`: If the parent updates `isLiked` during the transition, the new value is used after the transition ends. This is how the UI reflects the result of the action.
+
+- **Reducer pattern** like `useOptimistic(items, fn)`: If `items` changes while the transition is pending, React re-runs your `reducer` with the new `items` to recalculate what to show. This keeps your optimistic additions on top of the latest data.
+
+**What happens when the action fails**
+
+If the Action throws an error, the Transition still ends, and React renders with whatever `state` currently is. Since the parent typically only updates `state` on success, a failure means `state` hasn't changed—so the UI shows what it showed before the optimistic update. You can catch the error to show a message to the user.
```
stephan-noel: **nit**:  How are capitalization of action and transition being determined?  It's capitalized here but above and elsewhere it's not.

src/content/reference/react/useOptimistic.md
```diff
@@ -18,114 +18,1044 @@ title: useOptimistic
...
+
+</DeepDive>
+
 [See more examples below.](#usage)
 
 #### Parameters {/*parameters*/}
 
-* `state`: the value to be returned initially and whenever no action is pending.
-* `updateFn(currentState, optimisticValue)`: a function that takes the current state and the optimistic value passed to `addOptimistic` and returns the resulting optimistic state. It must be a pure function. `updateFn` takes in two parameters. The `currentState` and the `optimisticValue`. The return value will be the merged value of the `currentState` and `optimisticValue`.
+* `state`: The value returned when there are no pending Actions or Transitions.
```
stephan-noel: Probably my lack of knowledge, but seeing how the old docs focuses only on a pending action, and now "Actions or Transitions" is being called out, it's not clear to me the need for that distinction. Especially since an action is usually is just a function passed to startTransition.

In other words, what non-action transitions would be relevant for useOptimistic?

src/content/reference/react/useOptimistic.md
```diff
@@ -18,114 +18,1044 @@ title: useOptimistic
...
+    * If you pass a function as `nextState`, it will be treated as an _updater function_. It must be pure, should take the pending state as its only argument, and should return the next state. React will put your updater function in a queue and re-render your component. During the next render, React will calculate the next state by applying the queued updaters to the previous state similar to [`useState` updaters](/reference/react/useState#setstate-parameters)
+
+#### Returns {/*setoptimistic-returns*/}
+
+`set` functions do not have a return value.
+
+#### Caveats {/*setoptimistic-caveats*/}
+
+* The `set` function must be called inside a [Transition](/reference/react/useTransition) using [`startTransition`](/reference/react/startTransition) or inside an [Action](/reference/react/useTransition#perform-non-blocking-updates-with-actions). If you call the setter outside an Action or Transition, [React will show a warning](#an-optimistic-state-update-occurred-outside-a-transition-or-action) and the optimistic value will briefly render.
```
stephan-noel: Similar to an above comment I made, the explicit mention of "Action or Transition" makes me feel as if I'm missing something, the argument to startTransition is named "action".  Is there an some way of calling a setter in a transition that isn't an "action"?

**nit**: 

An "of" might read a bit better in "If you call the setter outside an Action or Transition"

outside **of** an Action

src/content/reference/react/useOptimistic.md
```diff
@@ -18,114 +18,1044 @@ title: useOptimistic
...
-    // updateFn
-    (currentState, optimisticValue) => {
-      // merge and return new state
-      // with optimistic value
-    }
-  );
+function MyComponent({name, todos}) {
+  const [optimisticAge, setOptimisticAge] = useOptimistic(28);
+  const [optimisticName, setOptimisticName] = useOptimistic(name);
+  const [optimisticTodos, setOptimisticTodos] = useOptimistic(todos, todoReducer);
```
hernan-yadiel, line 31: I found this example confusing. For me, this example only made sense after reading the **Deep Dive** section where it's mentioned that the `state` parameter can be either hardcoded values, props or state.

What if we just show a minimal example like:
```typescript
const [optimisticName, setOptimisticName] = useOptimistic(name);
```

Then in the **Parameters** section, we could mention that you can pass a hardcoded value or a state/prop to the `state` parameter.

src/content/reference/react/useOptimistic.md
```diff
@@ -18,114 +18,1044 @@ title: useOptimistic
...
 }
 ```
 
+<DeepDive>
+
+#### How optimistic state works {/*how-optimistic-state-works*/}
```
hernan-yadiel: When I first saw the Deep Dive accordion, I immediately opened it. Would it make sense to move it after the **Caveats** section?? I think it fits better once the user has read the API.

src/content/reference/react/useOptimistic.md
rickhanlonii: This is how we do it in other docs - this section is intended to be a quick reference, and guide you to look at the usages below. However, I think the Deep Dive here is making this confusing, and we don't include that in this section anywhere else so I'll move some things around.

src/content/reference/react/useOptimistic.md
rickhanlonii: Good call, I'll update it to "will trigger a re-render` like `useState` does.

src/content/reference/react/useOptimistic.md
rickhanlonii: Oops you found a claude `-` lol. Will fix.

src/content/reference/react/useOptimistic.md
```diff
@@ -18,114 +18,1044 @@ title: useOptimistic
...
+</DeepDive>
+
 [See more examples below.](#usage)
 
 #### Parameters {/*parameters*/}
 
-* `state`: the value to be returned initially and whenever no action is pending.
-* `updateFn(currentState, optimisticValue)`: a function that takes the current state and the optimistic value passed to `addOptimistic` and returns the resulting optimistic state. It must be a pure function. `updateFn` takes in two parameters. The `currentState` and the `optimisticValue`. The return value will be the merged value of the `currentState` and `optimisticValue`.
+* `state`: The value returned when there are no pending Actions or Transitions.
+* **optional** `reducer(currentState, optimisticValue)`: The reducer function that specifies how the optimistic state gets updated. It must be pure, should take the current state and the optimistic value as arguments, and should return the next optimistic state.
```
hernan-yadiel: When I read "state" the first thing that comes to my mind is either a prop or the value from `useState`. Then I think of the more abstract concept of "state".

II wonder if it would make _**more**_ sense to name this parameter `value` instead of `state`. The same would apply to the first parameter of the reducer function.

src/content/reference/react/useOptimistic.md
gaearon: *not emphasize haha

src/content/reference/react/useOptimistic.md
rickhanlonii: Good catch, I'll update it. It should be capitalized when referring to the React concept.

src/content/reference/react/useOptimistic.md
rickhanlonii: Navigations mostly. Tbh, I struggle with what to do here. See my comment here: https://github.com/reactjs/react.dev/pull/8264#discussion_r2733472946

src/content/reference/react/useOptimistic.md
rickhanlonii: You're right - I'm including both because I worry that people are not familiar with the idea that Actions are just the callback inside Transitions, so it can either be an `action` prop, or wrapped in `startTransition`. Not sure how to clarify it enough for devs with less experience with Transitions/Actions.

src/content/reference/react/useOptimistic.md
rickhanlonii: Yeah, I actually had the same idea to address your first comment.

src/content/reference/react/useOptimistic.md
rickhanlonii: Not sure. State makes it clear that it re-renders. Fun fact, `useOptimistic` is short for `useOptimisticState`.

src/content/reference/react/useOptimistic.md
hernan-yadiel: On the `useTransition` docs the parameter that the `startTransition` function take is named as an "action". But notice that it's **action**, not **Actions** (with an upper case A). I believe they're not the same thing. 

I think the only place where **Actions** were defined is in these two blog posts:

- https://react.dev/blog/2024/02/15/react-labs-what-we-have-been-working-on-february-2024
- https://react.dev/blog/2024/12/05/react-19

Probably the `action` term is a bit overloaded or ambiguous.

src/content/reference/react/useOptimistic.md
```diff
@@ -7,7 +7,7 @@ title: useOptimistic
 `useOptimistic` is a React Hook that lets you optimistically update the UI.
 
 ```js
-  const [optimisticState, addOptimistic] = useOptimistic(state, updateFn);
+const [optimisticState, addOptimistic] = useOptimistic(state, reducer?);
```
aurorascharff: The naming of "addOptimistic" is a bit strange to me, especially now that the updateFn has been renamed to reducer. 

In the examples below, you use the "setOptimistic" and "optimisticTodos" naming. I assume the appropriate name depends on what you are doing in updateFn/reducer, so would it make sense to use a more generic naming here in the signature? We are not necessarily adding anything. Maybe just `setOptimistic`?

src/content/reference/react/useOptimistic.md
aurorascharff: > btw, this doc uses "Action or Transition" everywhere which I hate. What are your thoughts on Action vs Transition. Part of me wants to just say Action throughout this doc, since a function inside a Transition is an Action, but I'm worried that "Action" is not a well-understood enough concept to go all in that way.

The concept of Action is new, but it is defined in https://react.dev/reference/react/useTransition#starttransition. Maybe you can just link here and remove the "or Transition"?

src/content/reference/react/useOptimistic.md
aurorascharff: It's also defined here; I believe they are the same thing. They're just not capitalized in the code itself. https://react.dev/reference/react/useTransition#starttransition

src/content/reference/react/useOptimistic.md
rickhanlonii: Nice catch, I thought I fixed this everywhere. Just pushed a fix!

The `add` is what you might would do for a list, like `addOptimisticTodo` in the list example. It's kinda up to people how they want to name things though, and as a result of this doc starting with `add` I've see people kinda assuming that this function is called `add` by convention, which adds to the confusion for this API.

src/content/reference/react/useOptimistic.md
```diff
@@ -18,114 +18,1043 @@ title: useOptimistic
...
+The `set` function returned by `useOptimistic` lets you update the state for the duration of a Transition. You can pass the next state directly, or a function that calculates it from the previous state:
+
+```js
+const [optimisticLike, setOptimisticLike] = useOptimistic(false);
+const [optimisticSubs, setOptimisticSubs] = useOptimistic(subs);
+
+function handleClick() {
+  startTransition(async () => {
+    setOptimisticLike(true);
+    setOptimisticSubs(a => a + 1);
```
aurorascharff, line 63: Is using the updater function here equivalent to defining `a => a + 1` in the `reducer` argument of `useOptimistic`?

rickhanlonii: Yeah, in fact just like useState, using an updater function like this actually uses a reducer under the hook (just a basic reducer that calls the updater function.

I added a deep dive below to explain when it makes sense to use updater vs reducer:

Screenshot 2026-01-27 at 7 40 19 PM

It's the same as when to use `useState` vs `useReducer`. We just decided to put it in one API instead of having `useOptimisticState` and `useOptimisticReducer` separate.

aurorascharff: Oh, sweet!

src/content/reference/react/useOptimistic.md
```diff
@@ -18,114 +18,1043 @@ title: useOptimistic
...
+    onAgeChange(42);
+  });
+}
+```
+
+React will render the optimistic state first, then complete the Transition with the new value.
+
+<Note>
+
+When using [Action props](/reference/react/useTransition#exposing-action-props-from-components), you can call the set function without `startTransition`:
```
aurorascharff, line 187: This separation is nice, it clarifies the confusion about "Action vs Transition". Maybe it can be explained more like this in other parts of the docs.

src/content/reference/react/useOptimistic.md
```diff
@@ -18,114 +18,1043 @@ title: useOptimistic
...
+import Button from './Button';
+import { submitForm } from './actions.js';
+
+export default function App() {
+  const [count, setCount] = useState(0);
+  return (
+    <div>
+      <Button action={async () => {
+        await submitForm();
+        setCount(c => c + 1);
```
aurorascharff: Here, the `setCount` is called after the async call, even though it doesn't use the return value of the `await submitForm`. Is it still in sync even without the [additional startTransition](https://react.dev/reference/react/useTransition#react-doesnt-treat-my-state-update-after-await-as-a-transition)?

src/content/reference/react/useOptimistic.md
```diff
@@ -18,114 +18,1043 @@ title: useOptimistic
...
+  const [optimisticState, updateOptimistic] = useOptimistic(
+    { isFollowing: user.isFollowing, followerCount: user.followerCount },
+    (current, isFollowing) => ({
+      isFollowing,
+      followerCount: current.followerCount + (isFollowing ? 1 : -1)
+    })
   );
 
+  function handleClick() {
+    const newFollowState = !optimisticState.isFollowing;
```
aurorascharff, line 505: Can this logic also be handled inside the reducer function?

src/content/reference/react/useOptimistic.md
```diff
@@ -18,114 +18,1043 @@ title: useOptimistic
...
+  }
+
+  // ...
+}
+
+// ✅ Also Correct: calling from an Action prop
+function MyComponent({ items }) {
+  const [isPending, setPending] = useOptimistic(false);
+
+  function action() {
```
aurorascharff, line 1024: The comment above MyComponentstates "Action prop", but this is not a prop. Is it assumed that the function is being passed to an action prop somewhere? Maybe we can use a better "Action prop" example here.

src/content/reference/react/useOptimistic.md
```diff
@@ -18,114 +18,1043 @@ title: useOptimistic
...
+
+
+1. **Check if `optimisticValue === value`**
+
+```js
+const [optimistic, setOptimistic] = useOptimistic(value);
+const isPending = optimistic !== value;
+```
+
+If the values are not equal, there's a Transition in progress. This is the same as wrapping with an additional `useTransition`, so an additional wrapper is not necessary.
```
aurorascharff: Since "wrapping with an additional `useTransition`" is a way to track the pending state, we actually have three different ways to track the pending state instead of two.

rickhanlonii: In the paragraph after this, I explain that this check is the same as useTransition

src/content/reference/react/useOptimistic.md
rickhanlonii: You want to pass it in here because you're posting the value to the `followAction` endpoint. If you did it in the reducer, then if there was an update while this was in progress, it would toggle it back, causing it to flicker when the post finishes.

src/content/reference/react/useOptimistic.md
rickhanlonii: Maybe I misunderstand - the comment above this one is:

```
// ✅ Correct: calling inside startTransition
```

and the next one is about action?

src/content/reference/react/useOptimistic.md
rickhanlonii: In the paragraph between 1 and 2 we say:

> This is the same as wrapping with an additional `useTransition`, so an additional wrapper is not necessary.

Is there a use case you've found for needing the useTransition?

src/content/reference/react/useOptimistic.md
rickhanlonii: Where are you thinking?

src/content/reference/react/useOptimistic.md
rickhanlonii: Nice catch, fixed!

src/content/reference/react/useOptimistic.md
aurorascharff: Not really, I might have done it because I forgot about the two other, better options.

src/content/reference/react/useOptimistic.md
aurorascharff: I was referring to this one:
`// ✅ Also correct: calling from an Action prop`

src/content/reference/react/useOptimistic.md
aurorascharff: The other sections that are saying "Action or Transition" where there were some comments about confusion. Related to https://github.com/reactjs/react.dev/pull/8264#discussion_r2733472946

src/content/reference/react/useOptimistic.md
rickhanlonii: Ahh, yeah ok I removed "prop", though I was thinking the `//...` was implying you'd pass the action as a prop.

src/content/reference/react/useOptimistic.md
```diff
@@ -18,114 +18,1055 @@ title: useOptimistic
...
+    setOptimisticLike(true);
+    setOptimisticSubs(a => a + 1);
+    await saveChanges();
+  });
+}
+```
+
+#### Parameters {/*setoptimistic-parameters*/}
+
+* `nextState`: The value that you want the optimistic state to be. If you provided a `reducer` to `useOptimistic`, this value will be passed as the second argument to your reducer. It can be a value of any type.
```
stephan-noel: Above and elsewhere it's referred to as `optimisticValue`, but here it's named `nextState`.  Is it better to stick to one?  Maybe `optimisticValue`?

Same applies to below for the argument as function explainer.

src/content/reference/react/useOptimistic.md
```diff
@@ -18,114 +18,1055 @@ title: useOptimistic
...
+    * If you pass a function as `nextState`, it will be treated as an _updater function_. It must be pure, should take the pending state as its only argument, and should return the next state. React will put your updater function in a queue and re-render your component. During the next render, React will calculate the next state by applying the queued updaters to the previous state similar to [`useState` updaters](/reference/react/useState#setstate-parameters).
```
stephan-noel: **nit**: Besides the naming of `optimisticValue` instead of `nextState`, when mentioning "next state" as separate words, is it more appropriate to say "optimistic state" instead?

For example, "should return the next state" makes sense for the useState updater function, but does "should return the optimistic state" make more sense here?  Not sure if the fact of transitions being interruptible has any bearing here.  Maybe that train of thought would get in the weeds with "next render" though.

src/content/reference/react/useOptimistic.md
```diff
@@ -18,114 +18,1055 @@ title: useOptimistic
...
+`useOptimistic` lets you show a temporary value while a Transition is running:
+
+```js
+const [value, setValue] = useState('a');
+const [optimistic, setOptimistic] = useOptimistic(value);
+
+startTransition(async () => {
+  setOptimistic('b');
+  setValue('b');
+});
```
stephan-noel, line 96: I think the simplicity does help digesting here. It just feels like something is missing with the transition being async without any corresponding `await`, especially with regards to point 4 below.

src/content/reference/react/useOptimistic.md
```diff
@@ -18,114 +18,1055 @@ title: useOptimistic
...
+function MyComponent({age, name, todos}) {
+  const [optimisticAge, setOptimisticAge] = useOptimistic(age);
+  const [optimisticName, setOptimisticName] = useOptimistic(name);
+  const [optimisticTodos, setOptimisticTodos] = useOptimistic(todos, reducer);
+  // ...
+```
+
+`useOptimistic` returns an array with exactly two items:
+
+1. The <CodeStep step={2}>current state</CodeStep> of the optimistic value, returning the <CodeStep step={1}>state</CodeStep> provided.
```
stephan-noel: "The current state of the optimistic value" felt a bit hard to parse.  It feels like saying the optimistic value is changing state, whereas before `optimisticValue` referred to the parameter of the set function.

Is this portion of the section like a summary of the signature so people don't have to scroll back up to the top?

src/content/reference/react/useOptimistic.md
```diff
@@ -18,114 +18,1055 @@ title: useOptimistic
...
+import { updateName } from './actions.js';
+
+export default function EditName({ name, onSubmit }) {
+  const [optimisticName, setOptimisticName] = useOptimistic(name);
+
+  async function submitAction(formData) {
+    const newName = formData.get('name');
+    setOptimisticName(newName);
+    const updatedName = await updateName(newName);
+    onSubmit(updatedName);
```
stephan-noel: Just sharing how I was thinking about this example and questions I had.

I guess technically onSubmit could also be an action prop since it's also called inside of an action?  I was trying to understand if leaving it as onSubmit was intentional to communicate that following the action prop naming convention is optional.  On the other hand, if you don't communicate somehow that it's inside an action, I thought consumers could be confused about why the updates from their set functions get entangled?  In this case, maybe it's fine to diverge since we usually expect submits to be async?

Update 1 hour later: 

Just realized it's called inside of an action but because of the lack of async context thing, it's not really in a transition anymore, for now at least.  That was pretty confusing 😄 .   "Half-action prop".

Maybe worth calling this out explicitly?  Feels like an action prop gotcha.  Could also help keep track of what needs to change once async context lands.

src/content/reference/react/useOptimistic.md
```diff
@@ -18,114 +18,1055 @@ title: useOptimistic
...
+
+</DeepDive>
+
+---
+
+### Adding pending state to a component {/*adding-pending-state-to-a-component*/}
+
+To show a pending state while an Action is running, use `useOptimistic` with a boolean.
+
+Here's a button that shows "Submitting..." while the Action is pending using the [Action prop pattern](/reference/react/useTransition#exposing-action-props-from-components):
```
stephan-noel: I feel like it would be even clearer to explicitly mention that `useTransition` could also be used here instead and that this example is just illustrating that `useOptimistic` could also be used for this purpose and it's just as "blessed".

A question that pops into my head is when to use which approach.  Makes me thing that an example which is harder to replace with useTransition might help address both of these.   If there is no such situation in which one is preferable, maybe mentioning that could be helpful.

src/content/reference/react/useOptimistic.md
```diff
@@ -18,114 +18,1055 @@ title: useOptimistic
...
+});
+dispatch(action);
+```
+
+**Use updaters** for calculations where the setter call naturally describes the update. This is similar to using `setState(prev => ...)` with `useState`.
+
+**Use reducers** when you need to pass data to the update (like which item to add) or when handling multiple types of updates with a single hook.
+
+Both patterns ensure React can update your optimistic value if the state changes while your Transition is pending.
+
```
stephan-noel, line 448: Maybe this deep dive is better placed after the later examples that illustrate when to use a reducer?  Maybe it's even is better as a kind of summary of heuristics based on the previous examples.

At least when reading top to bottom, I started reading the deep dive and couldn't make sense of it until the later examples.

src/content/reference/react/useOptimistic.md
```diff
@@ -18,114 +18,1055 @@ title: useOptimistic
...
+  await new Promise((res) => setTimeout(res, 1000));
+  // In a real app, this would save to the server
+  return { ...todo, pending: false };
+}
+```
 
 </Sandpack>
+
+The `reducer` receives the current list of todos and the new todo to add. This is important because if the `todos` prop changes while your add is pending (for example, another user added a todo), React will update your optimistic value by re-running the reducer with the updated list. This ensures your new todo is added to the latest list, not an outdated copy.
+
```
stephan-noel, line 648: I feel like here I finally understood why you need to use a reducer, because the updater function, or set function with optimisticValue would not be re-run when the state argument changes.  It feels like it's buried being here a bit.  Maybe including this explanation in the deep dive/summary as well would be useful.

As a side note, it does make me wonder about highlighting that reducer is optional, supporting only reducers which always work correctly instead of a kind of hybrid set function/reducer as well for convenience does feel like it's a wider/heavier/more error-prone api surface area for little gain.  That ship has probably sailed though.

src/content/reference/react/useOptimistic.md
```diff
@@ -18,114 +18,1055 @@ title: useOptimistic
...
+}
+```
+
+</Sandpack>
+
+The reducer handles three `action` types (`add`, `remove`, `update_quantity`) and returns the new optimistic state for each. Each `action` sets a `pending: true` flag so you can show visual feedback while the [Server Function](/reference/rsc/server-functions) runs.
+
+---
+
+### Optimistic delete with error recovery {/*optimistic-delete-with-error-recovery*/}
```
stephan-noel, line 829: Taking from the WG comment you made:

> Handling errors / mismatched: If an action errors, or the value resolves to a different value than the optimistic one (like the value from the server has changed from a different user or tab) the optimistic value will automatically commit the newest actual value passed in.

The reducer part is more for updates to the state parameter during a transition, but in this comment you meant a mismatch in when the server resolves to a different value right?  Did you still want to include the mismatch case in here similar to how the error case is covered?

src/content/reference/react/useOptimistic.md
stephan-noel: Just to share my experience, trying to learn these concepts, in the hope that it might help (probably out of scope though).  The transition/action terminology felt like a lot of mental weight when first learning these ideas.  One reason is probably because it's only covered in the reference section and might get a more conceptual treatment once it makes its way to the Learn section.  Another reason (mentioned above) is probably how similar they are and how it's used as synonyms in various scenarios (ie, Do I "call this function in a transition" or "in an action", Could `useActionState` have been named `useTransitionState`?  Would `<form transition={}>` have made just as much sense?).  When do you use which?

src/content/reference/react/useOptimistic.md
```diff
@@ -18,114 +18,1055 @@ title: useOptimistic
...
+// 🚩 Incorrect: outside a Transition
+function handleClick() {
+  setOptimistic(newValue);  // Warning!
+  // ...
+}
+
+// ✅ Correct: inside a Transition
+function handleClick() {
+  startTransition(async () => {
+    setOptimistic(newValue);
```
vadymstebakov: What’s the difference between calling an optimistic setter in an async Transition and a non-async Transition?

src/content/reference/react/useOptimistic.md
rickhanlonii: Yeah, good call. I'm going to update it to be:

```
const [optimisticState, setOptimistic] = useOptimistic(value);
```

The optimistic state is the current state returned from the hook, which you can update with the setter. The value is the what you're passing in to create optimistic state for.

src/content/reference/react/useOptimistic.md
rickhanlonii: Yeah, will update.

src/content/reference/react/useOptimistic.md
rickhanlonii: Yeah, in many/most of the reference doc, the first usage example is like a "how to use it" with the concepts and highlighting the concepts in code. So it's important to get this right. I took a pass at re-wring this based on your feedback, what do you think?

src/content/reference/react/useOptimistic.md
rickhanlonii: Yeah, it's not an action. I did that intentionally to avoid the additional `startTransition` since the example says you "don't need startTransition", but that's just super confusing. 

I updated to use `startTransition` and an action prop, but I think I'll eventually re-write this to use Suspense. Then there's no `await updateName` so we can remove the `startTransition` and explain how the optimistic state waits for Suspense at the same time.

src/content/reference/react/useOptimistic.md
rickhanlonii: Added

src/content/reference/react/useOptimistic.md
rickhanlonii: Agreed - and I actually deleted the updater only example.

src/content/reference/react/useOptimistic.md
rickhanlonii: For simple cases of just wrapping a primitive a reducer is overkill. I'm intentionally not leading with the reducer examples because people complain that `useOptimistic` code [looks more complicated than it is](https://www.columkelly.com/blog/use-optimistic), because using a reducer by default leads you into this pitfall where you use a reducer around all items instead of the primitive that you're actually updating. 

I want to follow up with a pitfall around this. For example, in that article was originally wrapping all of the todos in useOptimistic just to toggle the completed state on one item. Then you need to manage per-item pending states and the whole shebang, which can be hard to get right and show too much stuff pending.

Now it shows just wrapping the `todo.completed` close to the action, which is the easier and IMO better way to do it. Use optimistic state close to the place you're actually doing the action (i.e. the checkbox) not where the data is defined (todos in the parent).

src/content/reference/react/useOptimistic.md
rickhanlonii: I made some updates so the examples use values from the server for the "real" update, I think that probably covers that case.

src/content/reference/react/useOptimistic.md
rickhanlonii: Fixed. Implied here was that `setValue` would cause something to suspend (so conceptually the await is in render) but I agree implying that is confusing and I'll just add the await.

I also updated this deep dive a bit.

src/content/reference/react/useOptimistic.md
rickhanlonii: Actually, I added some more to the deep dive to cover this.

Screenshot 2026-01-30 at 2 09 32 PM

src/content/reference/react/useOptimistic.md
rickhanlonii: Updated to focus on Actions.

src/content/reference/react/useOptimistic.md
rickhanlonii: Wen't all in on "optimistic state", and the param you pass is the "value"

gaearon:
IMG_3548
IMG_3549

This feels confusing. The section starts with “you can call the optimistic setter directly without startTransition” but the example does have startTransition. I know it’s for an unrelated reason but is it possible to disambiguate or simplify the example? Or to explain what this is about.

src/content/reference/react/useOptimistic.md
stephan-noel: Way smoother to read now, nice!

stephan-noel:
> IMG_3548 IMG_3549
> 
> This feels confusing. The section starts with “you can call the optimistic setter directly without startTransition” but the example does have startTransition. I know it’s for an unrelated reason but is it possible to disambiguate or simplify the example? Or to explain what this is about.

True.  Related thread here https://github.com/reactjs/react.dev/pull/8264#discussion_r2747501457.

PR #8265, opened 2026-01-26
init claude config

rickhanlonii:
Adds a simple CLAUDE.md file, and 4 skills:
- `docs-components`: for opinions on using the MDX components like <CodeStep>
- `docs-sandpack`: for opinions when creating code sandboxes
- `docs-writer-learn`: for opinions on writing Learn docs
- `docs-writer-reference`: for opinions on writing Reference docs

Also adds a `docs-reviewer` agent that uses all of the context give a review, with a checklist of what to check. Using an agent for this allows isolating the context for everything into that agent, so it's not loaded into the main context window for every session. The idea is that you can write the docs using the skills, and then ask `docs-reviewer` to give it a full review before submitting.

### DIFF
--- a/.claude/agents/docs-reviewer.md
+++ b/.claude/agents/docs-reviewer.md
@@ -1,0 +1,77 @@
+---
+name: docs-reviewer
+description: "Use after editing docs to review changes. Orchestrates docs-writer-learn, docs-writer-reference, docs-components, and docs-sandpack skills to validate structure, components, and style"
+model: opus
+color: cyan
+---
+
+# React Documentation Reviewer Agent
+
+You are an expert reviewer for React documentation. Your role is to validate documentation changes for consistency, correctness, and adherence to established patterns.
+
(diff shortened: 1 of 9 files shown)


.claude/agents/docs-reviewer.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Key Validation Points
+
+These are quick checks - see the skills for full details:
+
+### All Documents
+- All `##`, `###`, `####` headings have explicit IDs: `{/*lowercase-with-hyphens*/}`
+- React terms capitalized: Hook, Effect, State, Context, Ref, Component
+- Uses "you" to address the reader
+- No time estimates ("quick", "simple", "easy")
```
josephsavona, line 71: Should we provide suggestions? i wonder if an extended example would help here. I mention because as a human, i'm not sure exactly what this is referring to and what the fix would be

PR #8272, opened 2026-01-28
feat: Add Accept header content negotiation for markdown

icyJoseph:
Serve markdown when requested via Accept header.

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -21,6 +21,20 @@ const nextConfig = {
   },
   async rewrites() {
     return [
+      // Serve markdown when Accept header prefers text/markdown
+      // Useful for LLM agents - https://www.skeptrune.com/posts/use-the-accept-header-to-serve-markdown-instead-of-html-to-llms/
+      {
+        source: '/:path*',
+        has: [
+          {
+            type: 'header',
+            key: 'accept',
+            value: '(.*text/markdown.*)',
+          },
+        ],
+        destination: '/api/md/:path*',
+      },
+      // Explicit .md extension also serves markdown
       {
         source: '/:path*.md',
         destination: '/api/md/:path*',


petermekhaeil:
TIL: I'm impressed with Next.js that you can do this with a few lines In `next.config.js`

next.config.js
```diff
@@ -21,6 +21,20 @@ const nextConfig = {
   },
   async rewrites() {
     return [
+      // Serve markdown when Accept header prefers text/markdown
+      // Useful for LLM agents - https://www.skeptrune.com/posts/use-the-accept-header-to-serve-markdown-instead-of-html-to-llms/
+      {
+        source: '/:path*',
```
icyJoseph, line 27: mm I think this need to be narrowed down a bit - doing a further review

icyJoseph: `GET /llms.txt` w/ `Accept: text/markdown` would map to `/api/md/llms.txt` and 404. 

It is correct, because there's no markdown `llms.txt`. Since it is a `txt` file, Agents might not include an Accept header at all 🤔

icyJoseph: Nope, this works fine because of how rewrites are resolved.

rickhanlonii: I don't think it works but I can't test in the preview URL because of protections. Let's merge and if it doesn't work we can fix after.

rickhanlonii:
@icyJoseph I don't think that works

```
curl -H -L "Accept: text/markdown" https://react.dev/reference/react
```

returns the HTML

icyJoseph:
@rickhanlonii mmm let me take a stab at it again

Edit: I think the command should've been:

```
curl -L -H "Accept: text/markdown" https://react.dev/reference/react
```

That being said, in the live site I still see HTML too. Locally this returns markdown:

```
curl -L -H "Accept: text/markdown" http://localhost:3000/reference/react
```

Has this been promoted to a production release?

icyJoseph:
Fixed here: https://github.com/reactjs/react.dev/pull/8276 🙏

PR #8279, opened 2026-01-30
useEffectEvent revamp

rickhanlonii:
### [Preview](https://react-dev-git-fork-rickhanlonii-rh-uee-revamp-fbopensource.vercel.app/reference/react/useEffectEvent?_vercel_share=LBGvDegXAouO7PjMIPcaTpMLwAxKr5Ja)

Used the new claude skills to revamp the useEffectEvent docs, then did a human pass.

If you're interested, the first commit is claude's attempt, and the second commit is my edits. 

Here's the PR with the claude improvements this used: https://github.com/reactjs/react.dev/pull/8280

### DIFF
--- a/src/content/reference/react/useEffectEvent.md
+++ b/src/content/reference/react/useEffectEvent.md
@@ -32,13 +32,13 @@ function ChatRoom({ roomId, theme }) {
 }
 ```
 
-Effect Events are a part of your Effect logic, but they behave more like an event handler. They always “see” the latest values of your props and state without re-synchronizing your Effect, so they're excluded from Effect dependencies. See [Separating Events from Effects](/learn/separating-events-from-effects#extracting-non-reactive-logic-out-of-effects) to learn more. 
+Effect Events are a part of your Effect logic, but they behave more like an event handler. They always “see” the latest values from render (like props and state) without re-synchronizing your Effect, so they're excluded from Effect dependencies. See [Separating Events from Effects](/learn/separating-events-from-effects#extracting-non-reactive-logic-out-of-effects) to learn more. 
 
 [See more examples below.](#usage)
 
 #### Parameters {/*parameters*/}
 
-* `callback`: A function containing the logic for your Effect Event. The function can accept any number of arguments and return any value. When you call the returned Effect Event function, the `callback` always accesses the latest values from props and state at the time of the call. 
+* `callback`: A function containing the logic for your Effect Event. The function can accept any number of arguments and return any value. When you call the returned Effect Event function, the `callback` always accesses the latest committed values from render at the time of the call. 
 
 #### Returns {/*returns*/}
 
@@ -134,9 +134,9 @@ See [Separating Events from Effects](/learn/separating-events-from-effects) to l
 
 ### Using a timer with latest values {/*using-a-timer-with-latest-values*/}
 
-When you use `setInterval` or `setTimeout` in an Effect, you often want to read the latest state values without restarting the timer whenever those values change.
+When you use `setInterval` or `setTimeout` in an Effect, you often want to read the latest values from render without restarting the timer whenever those values change.
 
-This counter increments by the current `increment` value every second. The `onTick` Effect Event reads the latest `increment` without causing the interval to restart:
+This counter increments `count` by the current `increment` value every second. The `onTick` Effect Event reads the latest `count` and `increment` without causing the interval to restart:
 
 <Sandpack>
 
@@ -148,7 +148,7 @@ export default function Timer() {
   const [increment, setIncrement] = useState(1);
 
   const onTick = useEffectEvent(() => {
-    setCount(c => c + increment);
+    setCount(count + increment);
   });
 
   useEffect(() => {
@@ -192,9 +192,9 @@ Try changing the increment value while the timer is running. The counter immedia
 
 ---
 
-### Using an event listener with latest state {/*using-an-event-listener-with-latest-state*/}
+### Using an event listener with latest values {/*using-an-event-listener-with-latest-values*/}
 
-When you set up an event listener in an Effect, you often need to read the latest state in the callback. Without `useEffectEvent`, you would need to include the state in your dependencies, causing the listener to be removed and re-added on every change.
+When you set up an event listener in an Effect, you often need to read the latest values from render in the callback. Without `useEffectEvent`, you would need to include the values in your dependencies, causing the listener to be removed and re-added on every change.
 
 This example shows a dot that follows the cursor, but only when "Can move" is checked. The `onMove` Effect Event always reads the latest `canMove` value without re-running the Effect:
 


MaxwellCohen:
I do not know if it is this page or the [You Might Not Need an Effect](https://react.dev/learn/you-might-not-need-an-effect) page, but there should be somewhere that explains where you would and would not use useEffectEvent more explicitly.

gaearon:
Here’s the original doc that spells out the motivation more clearly: https://react.dev/learn/separating-events-from-effects#declaring-an-effect-event

I agree that this reference page should ideally convey this (but shorter). Reading it now as it is, I don’t have a clear sense of what problem it solves. I do get this sense from the original doc so the challenge is how to compress it.

src/content/reference/react/useEffectEvent.md
```diff
@@ -4,91 +4,551 @@ title: useEffectEvent
...
 import { useEffectEvent, useEffect } from 'react';
 
 function ChatRoom({ roomId, theme }) {
   const onConnected = useEffectEvent(() => {
     showNotification('Connected!', theme);
   });
+}
+```
+
+Effect Events are functions you can call inside Effects, such as `useEffect`.
```
gaearon: i can call any functions inside Effects so this is neither correct nor useful. in particular, this makes it seem like you’re supposed to extract any callable functions into Effect Events, which is not what we want to say at all. 

what i think we want to give here is conceptual grounding. it’s a way to for effect to notify the rest of the program that something happened — like an event handler. and it avoids that event handler’s content influencing how often the effect fires.

src/content/reference/react/useEffectEvent.md
```diff
@@ -4,91 +4,551 @@ title: useEffectEvent
...
+useEffect(() => {
+  onSomething();
+}, [onSomething]); // ESLint will warn about this
+```
+
+This is a deliberate design choice. Effect Events are meant to be called only from within Effects in the same component. Since you can only call them locally and cannot pass them to other components or include them in dependency arrays, a stable identity would serve no purpose, and would actually mask bugs.
+
+The non-stable identity acts as a runtime assertion: if your code incorrectly depends on the function identity, you'll see the Effect re-running on every render, making the bug obvious.
+
+This design reinforces the rule that Effect Events are "escape hatches" for reading the latest values, not general purpose callbacks to be passed around.
```
gaearon: i wouldn’t say they’re escape hatches in this context (refs are). it’s more that conceptually their code belongs *to a particular effect* and you want to always be able to see both parts next to each other.

rickhanlonii: Updated:

> This design reinforces that Effect Events conceptually belong to a particular effect, and are not a general purpose API to opt-out of reactivity.

src/content/reference/react/useEffectEvent.md
rickhanlonii: Updated to:

> Effect Events are a part of your Effect logic, but they behave more like an event handler. They always “see” the latest values of your props and state without re-synchronizing your Effect, so they're excluded from Effect dependencies. See [Separating Events from Effects](/learn/separating-events-from-effects#extracting-non-reactive-logic-out-of-effects) to learn more.

src/content/reference/react/useEffectEvent.md
```diff
@@ -4,91 +4,551 @@ title: useEffectEvent
...
+}
+```
+
+Effect Events are a part of your Effect logic, but they behave more like an event handler. They always “see” the latest values of your props and state without re-synchronizing your Effect, so they're excluded from Effect dependencies. See [Separating Events from Effects](/learn/separating-events-from-effects#extracting-non-reactive-logic-out-of-effects) to learn more. 
+
+[See more examples below.](#usage)
+
+#### Parameters {/*parameters*/}
+
+* `callback`: A function containing the logic for your Effect Event. The function can accept any number of arguments and return any value. When you call the returned Effect Event function, the `callback` always accesses the latest values from props and state at the time of the call.
```
stephan-noel: The function can accept any number of arguments _and return any value_.

The return value will be ignored right?  Maybe just mentioning the return value will be ignored?

src/content/reference/react/useEffectEvent.md
```diff
@@ -4,91 +4,551 @@ title: useEffectEvent
...
+  const onMove = useEffectEvent(e => {
+    if (canMove) {
+      setPosition({ x: e.clientX, y: e.clientY });
+    }
+  });
 
   useEffect(() => {
-    const connection = createConnection(serverUrl, roomId);
+    window.addEventListener('pointermove', onMove);
+    return () => window.removeEventListener('pointermove', onMove);
```
stephan-noel: I'm trying to reconcile the effect cleanup with the statement near the beginning of

> Effect Event functions do not have a stable identity. Their identity intentionally changes on every render.

onMove during the cleanup (unmount) will be the same function that it was on first render/mount since the effect only executes once?

src/content/reference/react/useEffectEvent.md
```diff
@@ -4,91 +4,551 @@ title: useEffectEvent
...
+
+```js
+import { useState, useEffect, useEffectEvent } from 'react';
+
+export default function Timer() {
+  const [count, setCount] = useState(0);
+  const [increment, setIncrement] = useState(1);
+
+  const onTick = useEffectEvent(() => {
+    setCount(c => c + increment);
```
stephan-noel: I'm wondering if people will skim this and read that `useEffectEvent` is for "reading latest values" without noticing that the main reason for `setCount` here is that we are reading a "latest value" that is unrelated to the state we are setting (increment).  Otherwise, the updater function already gives you the latest value of count.

This also makes me wonder if guidance on the latest values use case would be helpful, or at least highlight alternatives.  For example, with the `useState` set function, you can use the updater function instead.  `useReducer` is also an alternative to get the latest values and would have also been a valid alternative here.

I guess it's a general question about to what degree should we prefer alternatives?  Since `useEffectEvent` can only be called inside of a `useEffect`, and `useEffect` is still considered an escape hatch, I'm guessing that would imply that `useEffectEvent` is also an escape hatch and alternatives should be preferred?

src/content/reference/react/useEffectEvent.md
stephan-noel: **nit**:  I guess it's not only props and state, but anything from the surrounding closure?  At least that's what I understood by looking at `packages\react-reconciler\src\ReactFiberHooks.js` and `packages\react-reconciler\src\ReactFiberCommitWork.js` in the main repo.  Maybe most people won't read into it that much.

src/content/reference/react/useEffectEvent.md
```diff
@@ -4,91 +4,551 @@ title: useEffectEvent
...
+
+If a value should cause your Effect to re-run, keep it as a dependency. Only use Effect Events for logic that genuinely should not re-trigger your Effect.
+
+See [Separating Events from Effects](/learn/separating-events-from-effects) to learn more.
+
+</Pitfall>
+
+---
+
+### Using a timer with latest values {/*using-a-timer-with-latest-values*/}
```
stephan-noel, line 135: **nit**:  One small thing is that one section uses "state" and the other uses "values" (ie, "Using a timer with latest values" vs "Using an event listener with latest state" and both of the values are coming from `useState`.

Below you say "latest state values".  I guess all are right and it doesn't always have to be consistent but the difference was noticeable especially when looking at the sidebar.

src/content/reference/react/useEffectEvent.md
rickhanlonii: Fixed

src/content/reference/react/useEffectEvent.md
rickhanlonii: Agreed, fixed!

PR #8284, opened 2026-02-03
Rewrite useActionState

rickhanlonii:
### [Preview](https://react-dev-git-fork-rickhanlonii-rh-uas-revamp-fbopensource.vercel.app/reference/react/useActionState?_vercel_share=KttelzEwAJyjqfJ1Mseo2GGSPxiysiZJ)

- First commit: claude
- Second commit: my edits

I need to do some more passes, but it's ready to review.

cc @samselikoff @gaearon @stephan-noel @aurorascharff @brenelz @MaxwellCohen @hernan-yadiel 

## Goals

- Client action first (with a mention of form actions / server functions)
- explain queuing actions (aka, so you can reduce them
- explain how to "fix" queing (aka optimistic state, or cancelling)
- sandbox based usage examples

the usage examples build up from:
- 1 action
- 2 actions
- 2 actions with pending states (via action props)
- 2 actions with pending and optimistic states
- 2 actions with a <form>

## Terms
_Edit: I went with `dispatchAction`._

I struggled with what to call the returned function and the reducer in the signature

```
const [_, dispatchAction, _] = useActionState(reducerAction, _);
```

I landed on `dispatchAction` because:
- it should have `dispatch` because it dispatches like `useReducer`
- it should use the "action" name, since it's called in a transition, so not just `dispatch`
- it's wordy, but shortening would make it unclear

I landed on `reducerAction` because:
- it has a reducer signature with the first arg
- it's an "action" so the returned state is updated in a transition
- it is a reducer inside an action, so it can do side effects

One wierd naming thing is this:

```
dispatchAction({type: 'Add'})
```

What do you call the argument passed to the action? useReducer calls it an "action", so that would mean it's

> call `action` with the `action` as the only argument.

So I called it `actionPayload`. It's the payload passed to the action.

## TODO

- [ ]  Usage example for how it can be used for error ui. 

## Followups

- Example of using `useState` and `useTransition` directly in useTransition docs.
- Full permalink example

### DIFF
--- a/src/content/reference/react/useActionState.md
+++ b/src/content/reference/react/useActionState.md
@@ -56,7 +56,7 @@ function MyCart({initialState}) {
 #### Caveats {/*caveats*/}
 
 * `useActionState` is a Hook, so you can only call it **at the top level of your component** or your own Hooks. You can't call it inside loops or conditions. If you need that, extract a new component and move the state into it.
-* React queues and executes multiple calls to `dispatchAction` sequentially, allowing each `reducerAction` to use the result of the previous Action.
+* React queues and executes multiple calls to `dispatchAction` sequentially. Each call to `reducerAction` receives the result of the previous call.
 * The `dispatchAction` function has a stable identity, so you will often see it omitted from Effect dependencies, but including it will not cause the Effect to fire. If the linter lets you omit a dependency without errors, it is safe to do. [Learn more about removing Effect dependencies.](/learn/removing-effect-dependencies#move-dynamic-objects-and-functions-inside-your-effect)
 * When using the `permalink` option, ensure the same form component is rendered on the destination page (including the same `reducerAction` and `permalink`) so React knows how to pass the state through. Once the page becomes interactive, this parameter has no effect.
 * When using Server Functions, `initialState` needs to be [serializable](/reference/rsc/use-server#serializable-parameters-and-return-values) (values like plain objects, arrays, strings, and numbers).
@@ -65,9 +65,9 @@ function MyCart({initialState}) {
 
 <Note>
 
-`dispatchAction` must be called within a Transition. 
+`dispatchAction` must be called from an Action. 
 
-You can wrap it in [`startTransition`](/reference/react/startTransition), or pass it to an [Action prop](/reference/react/useTransition#exposing-action-props-from-components).
+You can wrap it in [`startTransition`](/reference/react/startTransition), or pass it to an [Action prop](/reference/react/useTransition#exposing-action-props-from-components). Calls outside that scope won’t be treated as part of the Transition and [log an error](#async-function-outside-transition) on development mode.
 
 </Note>
 
@@ -86,11 +86,11 @@ async function reducerAction(previousState, actionPayload) {
 }
 ```
 
-Each time you call `dispatchAction`, React calls the `reducerAction` with the `actionPayload`. The reducer will perform side effects such as posting data, and return the new state. If `dispatchAction` is called multiple times, React queues and executes them in order so the result of the previous call is available for the current call.
+Each time you call `dispatchAction`, React calls the `reducerAction` with the `actionPayload`. The reducer will perform side effects such as posting data, and return the new state. If `dispatchAction` is called multiple times, React queues and executes them in order so the result of the previous call is passed as `previousState` for the current call.
 
 #### Parameters {/*reduceraction-parameters*/}
 
-* `previousState`: The current state. Initially this is equal to the `initialState`. After the first call to `dispatchAction`, it's equal to the last state returned.
+* `previousState`: The last state. Initially this is equal to the `initialState`. After the first call to `dispatchAction`, it's equal to the last state returned.
 
 * **optional** `actionPayload`: The argument passed to `dispatchAction`. It can be a value of any type. Similar to `useReducer` conventions, it is usually an object with a `type` property identifying it and, optionally, other properties with additional information.
 
@@ -104,6 +104,7 @@ Each time you call `dispatchAction`, React calls the `reducerAction` with the `a
 * `reducerAction` is not invoked twice in `<StrictMode>` since `reducerAction` is designed to allow side effects.
 * The return type of `reducerAction` must match the type of `initialState`. If TypeScript infers a mismatch, you may need to explicitly annotate your state type.
 * If you set state after `await` in the `reducerAction` you currently need to wrap the state update in an additional `startTransition`. See the [startTransition](/reference/react/useTransition#react-doesnt-treat-my-state-update-after-await-as-a-transition) docs for more info.
+* When using Server Functions, `actionPayload` needs to be [serializable](/reference/rsc/use-server#serializable-parameters-and-return-values) (values like plain objects, arrays, strings, and numbers).
 
 <DeepDive>
 
@@ -611,7 +612,7 @@ hr {
 
 ### Using with Action props {/*using-with-action-props*/}
 
-When you pass the `dispatchAction` function to a component that exposes an [Action prop](/reference/react/useTransition#exposing-action-props-from-components), you don't need to call `startTransition` or `useOptmisitc` yourself.
+When you pass the `dispatchAction` function to a component that exposes an [Action prop](/reference/react/useTransition#exposing-action-props-from-components), you don't need to call `startTransition` or `useOptimistic` yourself.
 
 This example shows using the `increaseAction` and `decreaseAction` props of a QuantityStepper component:
 
@@ -679,7 +680,7 @@ export default function QuantityStepper({value, increaseAction, decreaseAction})
 
   function handleDecrease() {
     startTransition(async () => {
-      setOptimisticValue(c => c + 1);
+      setOptimisticValue(c => Math.max(0, c - 1));
       await decreaseAction();
     });
   }
@@ -879,7 +880,7 @@ export default function QuantityStepper({value, increaseAction, decreaseAction})
 
   function handleDecrease() {
     startTransition(async () => {
-      setOptimisticValue(c => c + 1);
+      setOptimisticValue(c => Math.max(0, c - 1));
       await decreaseAction();
     });
   }
@@ -1408,7 +1409,7 @@ When `dispatchAction` is passed to an Action prop, React automatically wraps it 
 
 When you use `useActionState`, the `reducerAction` receives an extra argument as its first argument: the previous or initial state. The submitted form data is therefore its second argument instead of its first.
 
-```js
+```js {2,7}
 // Without useActionState
 function action(formData) {
   const name = formData.get('name');


src/content/reference/react/useActionState.md
```diff
@@ -4,269 +4,1320 @@ title: useActionState
...
+}
+```
+
+Each time you call `action`, React calls the `reducerAction` to perform side effects and compute the result of that Action. If the `action` is called multiple times, React queues and executes them in order so the result of the previous call is available for current call.
+
+#### Parameters {/*reduceraction-parameters*/}
+
+* `previousState`: The current state of the Action. Initially this is equal to the `initialState`. After the first call to `action`, it's equal to the last state returned.
+
+* `update`: The argument passed to `action`. It can be a value of any type. Similar to `useReducer` conventions, it is usually an object with a `type` property identifying it and, optionally, other properties with additional information.
```
stephan-noel: Should you also mention that when used for progressive enhancement with the `<form>` action prop, it should be `FormData`?  You cover it later in the "Using with `<form>` action props" section but maybe good to be explicit here as well?  

I haven't checked yet but I guess the same would hold for the [formAction attributes of button and input mentioned here](https://react.dev/reference/react-dom/components/form#props).

src/content/reference/react/useActionState.md
```diff
@@ -4,269 +4,1320 @@ title: useActionState
...
+* When using the `permalink` option, ensure the same form component is rendered on the destination page (including the same `reducerAction` and `permalink`) so React knows how to pass the state through. Once the page becomes interactive, this parameter has no effect.
+* When using Server Functions, `initialState` needs to be serializable (values like plain objects, arrays, strings, and numbers).
+* If `action` throws an error, React cancels all queued actions and shows the nearest [Error Boundary](/reference/react/Component#catching-rendering-errors-with-an-error-boundary).
+
+<Pitfall>
+
+When calling the `action` function, you must wrap the call in [`startTransition`](/reference/react/startTransition). If you call `action` without `startTransition`, the `isPending` flag will not update correctly, and React will show a warning in development.
```
stephan-noel: **nit**:  Not sure if it makes sense to also mention the alternative of passing it to an action prop.  I know you have an example later down in the "Using with Action props" section, but seems more complete to also mention it here.

src/content/reference/react/useActionState.md
```diff
@@ -4,269 +4,1320 @@ title: useActionState
...
+}
+```
+
+`useActionState` returns an array with exactly three items:
+
+1. The <CodeStep step={1}>current state</CodeStep>, initially set to the initial state you provided.
+2. The <CodeStep step={2}>`action` function</CodeStep> that lets you trigger the `reducerAction`.
+3. The <CodeStep step={3}>pending state</CodeStep> that tells you whether `action` is in progress.
+
+To trigger the Action, call the <CodeStep step={2}>`action` function</CodeStep> inside an [Action](/reference/react/useTransition#functions-called-in-starttransition-are-called-actions). React will call your `reducerAction` with the previous state and argument passed to `action`, and return the new <CodeStep step={1}>state</CodeStep>.
```
stephan-noel: **nit**:
>  call the action inside of an Action

This feels a bit off, especially since the reason the return value is called `action` is because it's called inside of a transition. You could also pass `action` directly to `startTransition`, as  `startTransition(action)` and that would be acceptable right?

src/content/reference/react/useActionState.md
```diff
@@ -4,269 +4,1320 @@ title: useActionState
...
+<Sandpack>
+
+```js src/App.js
+import { useActionState, startTransition } from 'react';
+import { addToCart } from './api';
+import Total from './Total';
+
+async function addTicket(prevCount) {
+  return await addToCart(prevCount);
+}
```
stephan-noel: This wrapping of `addToCart` made me wonder if it's just coincidental or if there is some intention behind it?    I tried in a separate NextJS project and the result was the same as passing `addToCart` directly to `useActionState`.


> Goals
> Client action first (with a mention of form actions / server functions)

Was the wrapping because of this that you mentioned in the PR description?  The terminology of "client action" or why you might need one, isn't used in the doc.  May also be good to clarify that client actions will lose progressive enhancement?

src/content/reference/react/useActionState.md
```diff
@@ -4,269 +4,1320 @@ title: useActionState
...
-* `initialState`: The value you want the state to be initially. It can be any serializable value. This argument is ignored after the action is first invoked.
-* **optional** `permalink`: A string containing the unique page URL that this form modifies. For use on pages with dynamic content (eg: feeds) in conjunction with progressive enhancement: if `fn` is a [server function](/reference/rsc/server-functions) and the form is submitted before the JavaScript bundle loads, the browser will navigate to the specified permalink URL, rather than the current page's URL. Ensure that the same form component is rendered on the destination page (including the same action `fn` and `permalink`) so that React knows how to pass the state through. Once the form has been hydrated, this parameter has no effect.
-
-{/* TODO T164397693: link to serializable values docs once it exists */}
+* `reducerAction`: The function to be called when the Action is triggered. When called, it receives the previous state (initially the `initialState` you provided, then its previous return value) as its first argument, followed by the arguments passed to the `action`.
```
hernan-yadiel: hey! I saw in the PR description you mention landing on `actionReducer` as the name for the first parameter, but in the examples it’s consistently `reducerAction`. I read `reducerAction` as "a reducer that runs as an action", while `actionReducer` reads more like "a reducer for actions".

src/content/reference/react/useActionState.md
```diff
@@ -4,269 +4,1320 @@ title: useActionState
...
-function StatefulForm({}) {
-  const [state, formAction] = useActionState(increment, 0);
-  return (
-    <form>
-      {state}
-      <button formAction={formAction}>Increment</button>
-    </form>
-  )
+function MyComponent() {
+  const [state, action, isPending] = useActionState(reducerAction, {quantity: 1});
```
hernan-yadiel:
```suggestion
  const [state, action, isPending] = useActionState(reducerAction, initialState);
```
so it matches what the reader is reading below.

src/content/reference/react/useActionState.md
```diff
@@ -4,269 +4,1320 @@ title: useActionState
 
 <Intro>
 
-`useActionState` is a Hook that allows you to update state based on the result of a form action.
+`useActionState` is a React Hook that lets you track the state of an [Action](/reference/react/useTransition#functions-called-in-starttransition-are-called-actions).
 
 ```js
-const [state, formAction, isPending] = useActionState(fn, initialState, permalink?);
+const [state, action, isPending] = useActionState(reducerAction, initialState, permalink?);
```
MaxwellCohen: The term `action` here is a little vague because reducers take an action, and it is called in Action; it seems like React is overloading the term `action`. Since 'useActionState' is like `useReducer` + Actions + side effects, using the `useReducer` reducer/dispatch language might be clearer. 

ie
```JS 
const [state, actionDispatch, isPending] = useActionState(actionReducer, initialState, permalink?)
```

src/content/reference/react/useActionState.md
```diff
@@ -4,269 +4,1320 @@ title: useActionState
...
+}
+```
+
+</Sandpack>
+
+In this example, when the user clicks the stepper arrows, the button submits the form and `useActionState` calls `updateCart` with the form data. The action uses `useOptimistic` to immediately show the new quantity while the server confirms the update.
+
+<RSC>
+
+When used with a [Server Function](/reference/rsc/server-functions), `useActionState` allows the server's response to be shown before hydration (when React attaches to server-rendered HTML) completes. You can also use the optional `permalink` parameter for progressive enhancement (allowing the form to work before JavaScript loads) on pages with dynamic content. This is typically handled by your framework for you.
```
stephan-noel: > You can also use the optional `permalink` parameter for progressive enhancement 

I read this as you need the permalink and that's what enables progressive enhandement.  What it is exactly that allows progressive enhancement has always been kind of fuzzy for me when reading the older docs.  It seemed like you needed useActionState for progressive enhancement but it seems like different parts of the progressive enhancement story are handled by different things.

src/content/reference/react/useActionState.md
```diff
@@ -4,269 +4,1320 @@ title: useActionState
...
-Call `useActionState` at the top level of your component to create component state that is updated [when a form action is invoked](/reference/react-dom/components/form). You pass `useActionState` an existing form action function as well as an initial state, and it returns a new action that you use in your form, along with the latest form state and whether the Action is still pending. The latest form state is also passed to the function that you provided.
+Call `useActionState` at the top level of your component to create state for the result of an Action.
 
 ```js
-import { useActionState } from "react";
+import { useActionState } from 'react';
 
-async function increment(previousState, formData) {
-  return previousState + 1;
+function reducerAction(state, action) {
```
hernan-yadiel: Prob. a typo.

Below you used `update`, and as you mentioned on the PR descrition, you are not convinced, me neither. `update` sounds more like a function name. What do you think about `actionPayload` (or `payload`)?

MaxwellCohen: Should the `async` be removed from reducerAction?

src/content/reference/react/useActionState.md
```diff
@@ -4,269 +4,1320 @@ title: useActionState
...
 ```
+
 </Sandpack>
 
-<Solution />
+Try clicking increase or decrease multiple times, and notice that the total updates within 1 second no matter how many times you click. This works because we're using an AbortController to "complete" the previous Action so the next Action can proceed.
+
+<Pitfall>
+
+Aborting an Action isn't always safe, which is why `useActionState` doesn't do it by default.
```
stephan-noel: Are there some examples of when or why it wouldn't be safe?  Otherwise I feel a bit left in the dark.

src/content/reference/react/useActionState.md
```diff
@@ -4,269 +4,1320 @@ title: useActionState
...
 
-`useActionState` returns an array with the following values:
+`useActionState` returns an array with exactly three values:
 
-1. The current state. During the first render, it will match the `initialState` you have passed. After the action is invoked, it will match the value returned by the action.
-2. A new action that you can pass as the `action` prop to your `form` component or `formAction` prop to any `button` component within the form. The action can also be called manually within [`startTransition`](/reference/react/startTransition).
-3. The `isPending` flag that tells you whether there is a pending Transition.
+1. The current state. During the first render, it will match the `initialState` you passed. After the action is invoked, it will match the value returned by the `reducerAction`.
+2. An `action` function that you call inside [Actions](/reference/react/useTransition#functions-called-in-starttransition-are-called-actions).
+3. The `isPending` flag that tells you whether there is a pending Action.
```
hernan-yadiel: I think this could read a little "global", like "any Action anywhere". My understanding is that `isPending` is scoped to the specific `action` returned by *that* `useActionState` call. Maybe something like "whether *this* action is pending" or "whether the `action` returned by this hook call is pending" would be a bit harder to misread.

src/content/reference/react/useActionState.md
```diff
@@ -4,269 +4,1320 @@ title: useActionState
 
 <Intro>
 
-`useActionState` is a Hook that allows you to update state based on the result of a form action.
+`useActionState` is a React Hook that lets you track the state of an [Action](/reference/react/useTransition#functions-called-in-starttransition-are-called-actions).
```
MaxwellCohen: This line feels a little vague in explaining what makes useActionState special compared to other hooks. UseActionState allows us to have state updates with side effects and async code

```
`useActionState` is a React Hook that manages state updates with side effects specifically within [Action](/reference/react/useTransition#functions-called-in-starttransition-are-called-actions)
```

src/content/reference/react/useActionState.md
```diff
@@ -4,269 +4,1320 @@ title: useActionState
...
+* `useActionState` is a Hook, so it must be called **at the top level of your component** or your own Hooks. You can't call it inside loops or conditions. If you need that, extract a new component and move the state into it.
+* React queues and executes multiple calls to `action` sequentially, allowing each `reducerAction` to use the result of the previous Action.
+* The `action` function has a stable identity, so you will often see it omitted from Effect dependencies, but including it will not cause the Effect to fire. If the linter lets you omit a dependency without errors, it is safe to do. [Learn more about removing Effect dependencies.](/learn/removing-effect-dependencies#move-dynamic-objects-and-functions-inside-your-effect)
+* When using the `permalink` option, ensure the same form component is rendered on the destination page (including the same `reducerAction` and `permalink`) so React knows how to pass the state through. Once the page becomes interactive, this parameter has no effect.
```
hernan-yadiel, line 61: Is there an example where `permalink` is used outside of the  `<form action={...}>` case?

I wasn’t totally sure if `permalink` is strictly a form/progressive-enhancement thing, or if it can apply more broadly. If it’s not form-only, I was leaning toward dropping "form" in "ensure the same form component is rendered…"

src/content/reference/react/useActionState.md
```diff
@@ -4,269 +4,1320 @@ title: useActionState
...
+* When using Server Functions, `initialState` needs to be serializable (values like plain objects, arrays, strings, and numbers).
```
hernan-yadiel: I think it could be helpful to link to the page that spells out what’s serializable and what's not:
[serializable parameters and return values](https://react-dev-git-fork-rickhanlonii-rh-uas-revamp-fbopensource.vercel.app/reference/rsc/use-server#serializable-parameters-and-return-values)

src/content/reference/react/useActionState.md
```diff
@@ -4,269 +4,1320 @@ title: useActionState
...
+
+* `previousState`: The current state of the Action. Initially this is equal to the `initialState`. After the first call to `action`, it's equal to the last state returned.
+
+* `update`: The argument passed to `action`. It can be a value of any type. Similar to `useReducer` conventions, it is usually an object with a `type` property identifying it and, optionally, other properties with additional information.
+
+#### Returns {/*reduceraction-returns*/}
+
+`reducerAction` returns the new state, and triggers a re-render with that state.
+
+#### Caveats {/*reduceraction-caveats*/}
```
hernan-yadiel, line 101: One thing I personally struggled with the first time (and many times) I used this hook with TypeScript is that `reducerAction` return type must match the type of `initialState`.

Would it make sense to add a short note in this caveats section about keeping the return type consistent with `initialState`? I think it would save people some head-scratching🫣

src/content/reference/react/useActionState.md
```diff
@@ -4,269 +4,1320 @@ title: useActionState
...
+
+**This is intentional in the design of `useActionState`.** 
+
+We have to wait for the previous result of `addTicket` in order to pass the `prevCount` to the next call to `addTicket`. That means React has to wait for the previous Action to finish before calling the next Action. 
+
+You can typically solve this by [using with useOptimistic](/reference/react/useActionState#using-with-useoptimistic) but for more complex cases you may want to consider [cancelling queued actions](#cancelling-queued-actions) or not using `useActionState`.
+
+</DeepDive>
+
+### Using multiple Action types {/*using-multiple-action-types*/}
```
MaxwellCohen, line 278: Is this the reducer patern, if so should it be in the title?

src/content/reference/react/useActionState.md
hernan-yadiel:
```suggestion
* **optional** `update`: The argument passed to `action`. It can be a value of any type. Similar to `useReducer` conventions, it is usually an object with a `type` property identifying it and, optionally, other properties with additional information.
```

Never used it without this second argument, but it makes total sense

src/content/reference/react/useActionState.md
rickhanlonii: Nice catch! Updated the descripton.

src/content/reference/react/useActionState.md
hernan-yadiel: also, I noticed that in @types/react it is named `payload`:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/c0e4c41164885d5d2d40845848b4d64782d7ca61/types/react/index.d.ts#L1958-L1967

I think I like `actionPayload` better since it kinda follows the pattern of being more explicit of the other parameter names (`reducerAction`, `initialState`).

src/content/reference/react/useActionState.md
```diff
@@ -4,269 +4,1320 @@ title: useActionState
...
+
+#### How is `useActionState` different from `useReducer`? {/*useactionstate-vs-usereducer*/}
+
+You might notice this example looks a lot like `useReducer`, but they serve different purposes:
+
+- **Use `useReducer`** to manage state of your UI. The reducer must be pure.
+
+- **Use `useActionState`** to manage state of your Actions. The reducer can perform side effects.
+
+You can think of `useActionState` as `useReducer` for side effects from user Actions. Since it computes the next Action to take based on the previous Action, it has to [order the calls sequentially](/reference/react/useActionState#how-useactionstate-queuing-works). If you want to perform Action in parallel, use `useState` and `useTransition` directly.
```
hernan-yadiel: It would be really nice to have an example we can point to here. Maybe we could add one to the useTransition docs (and link it from here), or include a small snippet right here that shows what the “run these in parallel with useState + useTransition” pattern looks like

src/content/reference/react/useActionState.md
```diff
@@ -4,269 +4,1320 @@ title: useActionState
...
+```
+
+</Sandpack>
+
+
+When the stepper arrow is clicked, `setOptimisticCount` immediately updates the quantity, and `action()` queues the `updateCart`. We show a pending indicator on both the quantity and total to give the user feedback that their update is still being applied.
+
+---
+
+### Using with `<form>` action props {/*use-with-a-form*/}
```
hernan-yadiel: I think it would be really helpful to show how it can be used for error ui. The troubleshooting section hints at “return error state instead of throwing”, but I could see people missing how to structure it.

Maybe we could add a small example where `initialState` is something like `{ error: null, count: 0 }`, and the reducer returns `{ error: '...' }` when the server call fails.

src/content/reference/react/useActionState.md
```diff
@@ -4,269 +4,1320 @@ title: useActionState
...
+- It *reduces* the previous state into a new state, like `useReducer`.
+- It's called inside a Transition and can perform side effects, like an Action.
+
+Conceptually, `useActionState` is like `useReducer`, but you can do side effects in the reducer.
+
+</DeepDive>
 
 ---
 
 ## Usage {/*usage*/}
```
hernan-yadiel, line 124: I've seen a pattern in a few places where people use `.bind` to "pre-fill" an argument before passing the reducer function to `useActionState`. Something like:

```js
function updateCart(userId, prevState, payload) {
  // ...
}

function Checkout({ userId }) {
  const updateCartForUser = updateCart.bind(null, userId);
  const [state, action] = useActionState(updateCartForUser, { error: null });
  // ..
}
```

Is this pattern encouraged?

I can see how it can get a bit weird with TypeScript.

src/content/reference/react/useActionState.md
aurorascharff: What is the reason behind this pitfall? If the dispatch function returned is an Action called "action", it's counterintuitive that we need to wrap it again in another startTransition. As @MaxwellCohen said above, something like `[state, actionDispatch, isPending] = useActionState(actionReducer, initialState, permalink?)` would make more sense since we still need to either wrap the dispatch in `useTransition`or pass it to a form `action` or other action prop.

samselikoff:
> I struggled with what to call the returned function and the reducer in the signature

>     const [_, action, _] = useActionState(reducerAction);

Kinda feels like we need a new term. I really like [the existing docs](https://react.dev/learn/extracting-state-logic-into-a-reducer) on reducers and would eventually love to see a similar build-up of the logic for action reducers. Show how you can do it on your own, then show how useActionState basically provides sugar over that.

What about asyncReducer?

```
const [state, asyncDispatch, isPending] = useActionState(asyncReducer, initialState, permalink?);
```

and

```
async function yourAsyncReducer(state, action) {
  // await any async functions, then return next state for React to set
}
```

If we eventually have a learn page around this stuff, we can explain the differences between "async reducers" and "reducers". asyncDispatch feels like a strong enough convention to hint that it must be called within a transition. And asyncReducer hints that "this is no ordinary reducer". This reducer can have side effects.

samselikoff:
In retrospect, and in light of your recent "Async React" branding (which I think is fantastic), maybe the hook could have been called useAsyncReducer ^_^

samselikoff:
I wonder if "preserving a form's inputs after a failed submission" warrants its own section in Usage. It's such a common one and not entirely obvious how to use defaultValue to pull it off (given React 19 resets forms):

```jsx
import { login } from "./actions";

function Form() {
  const [state, asyncDispatch, isPending] = useActionState(
    async (prev, formData) => {
      const { name, password } = Object.fromEntries(formData);

      try {
        await login(name, password);
        return { status: "success" };
      } catch (error) {
        return { status: "error", error: error.toString(), formData };
      }
    },
    { status: "init" },
  );

  return (
    <form action={asyncDispatch}>
      <input
        type="email"
        name="email"
        defaultValue={state.formData?.get("email") ?? ""}
      />
      {state.status === "error" && <p>{state.error}</p>}
      <input
        type="password"
        name="password"
        defaultValue={state.formData?.get("password") ?? ""}
      />
    </form>
  );
}
```

src/content/reference/react/useActionState.md
stephan-noel: Also, not sure if out of scope, but [that page](https://react.dev/reference/rsc/use-server) is missing the fact that that client references can be passed.

Image

samselikoff:
Another one that's not obvious is that you can mix async and sync code branches in the asyncReducer. Nice for resetting state or anything else that doesn't involve a side effect.

```jsx
async function updateCart(state, formData) {
  const type = formData.get("type");
  switch (type) {
    case "ADD": {
      return await addToCart(state.prevCount);
    }
    case "REMOVE": {
      return await removeFromCart(state.prevCount);
    }
    case "RESET": {
      return state.initialCount; // no async calls
    }
    default: {
      throw Error("Unknown action: " + type);
    }
  }
}
```

I've seen tons of folks in comments feeling like they're stuck with whatever state was returned from the previous server function, when they can just add a branch to reset a form all in the client.

src/content/reference/react/useActionState.md
MaxwellCohen: I want this pitfall text because I received the "useOptimistic" version of this message, and the docs didn't mention the warning, prompting me to reach out to @rickhanlonii. My theory is that the lack of documentation about async React is making it harder for developers and LLMS to use these cool new React tools.


The warning text (please include the exact text so that when someone Googles the error in 6 months can find the answer)

```
An async function with useActionState was called outside of a transition. This is likely not what you intended (for example, isPending will not update correctly). Either call the returned function inside startTransition, or pass it to an `action` or `formAction` prop.
```

rickhanlonii: > If the dispatch function returned is an Action called "action", it's counterintuitive that we need to wrap it again in another startTransition.

Isn't this the way _prop_ callbacks work? If you create a prop is named with `action`, the component calling the prop needs to wrap it in `startTransition`. So you have to call anything named `action` inside a transition, and then inside the function you know it's executed in a transition.

rickhanlonii:
@samselikoff re: naming - the actions don't need to be async or could be a mix of sync and async. For example, you could have `useActionState` where the "action" side effect is to call `showNotification` (a synchronous API)

src/content/reference/react/useActionState.md
```diff
@@ -4,269 +4,1320 @@ title: useActionState
...
-
-{/* TODO T164397693: link to serializable values docs once it exists */}
+* `reducerAction`: The function to be called when the Action is triggered. When called, it receives the previous state (initially the `initialState` you provided, then its previous return value) as its first argument, followed by the arguments passed to the `action`.
+* `initialState`: The value you want the state to be initially. React ignores this argument after invoking the action for the first time.
+* **optional** `permalink`: A string containing the unique page URL that this form modifies.
```
stephan-noel, line 44: I kinda always wanted an example showing permalink usage.  Maybe kinda tricky with the codesandboxes?

src/content/reference/react/useActionState.md
MaxwellCohen: I am trying to understand 'action' in the context of this page because we have several concepts that are called action:
- Components that take a callback that will be run in a transtion is called action or action props, i.e., form component. Forms will pass formdata to the callback.
- StartTransition takes a callback called action and schedules it to run in a transition. StartTransition will not provide any extra data to the callback. 
- The second return value of useActionState is 'action', which is like an async reducer that can take 0 to 1 values.

The pending state comes from 
- Components - need to expose pending options or use useActionState
- StartTransition - pending state is exposed in the useTransition hook and not in the StartTransition api function
- useActionState - pending state is the 3rd value in the array

If you want to add optimistic updates 
- Components have the setOptimistic in the action props callback function.
- StartTranstion have the setOptimistic in the action callback function. 
- for useActionState's 'action' wrap useActionState's 'action' and setOptimistic in a new function, and pass it to a transition or component.

It seems to me that, in some cases, useActionState's 'action' is the same as components or StartTransition's action, but in several common cases, useActionState's 'action'  is part of an action.  Thus adding confusion about what is being referenced or the developer's expectations in using useActionState's 'action.'

src/content/reference/react/useActionState.md
```diff
@@ -4,269 +4,1320 @@ title: useActionState
...
+  return await addToCart(prevCount);
+}
+
+export default function Checkout() {
+  const [count, action, isPending] = useActionState(addTicket, 0);
+
+  function handleClick() {
+    startTransition(() => {
+      action();
+    });
```
stephan-noel, line 166: **nit**: 
```suggestion
    startTransition(action);
```

Similar to [what I mentioned here](https://github.com/reactjs/react.dev/pull/8284/changes#r2757000829).  Kind of drives the point home that the argument to `startTransition` is an action (as is part of the result of `useActionState`), and that wrapping isn't necessary.

src/content/reference/react/useActionState.md
stephan-noel: Just adding my 2 cents here.  

I also first thought maybe `payload` would do, but then I looked at the [Redux docs](https://redux.js.org/tutorials/fundamentals/part-3-state-actions-reducers#writing-reducers) and that `payload` is a sibling attribute to `type` rather than the entire thing and so thought that might be somewhat confusing.  In contrast, [the useReducer docs](https://react.dev/reference/react/useReducer#writing-the-reducer-function) don't use the `payload` convention and have extra properties as direct siblings to `type` (simpler IMO).

Then I thought, let me look at web standards like those Remix guys.  The idea of reducers has some terminology in common with   [CustomEvent](https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent#options), it has a `type` that's a string, but then has `options` and `options.detail`.  Maybe just `detail`?  `actionDetail`?  Not sure it's better than `update`, but when I read `update` it do tend to read it more as a verb (and therefore function).

src/content/reference/react/useActionState.md
```diff
@@ -4,269 +4,1320 @@ title: useActionState
...
+function action(prevState, formData) {
+  const name = formData.get('name');
+}
+```
+
+---
+
+### My `isPending` flag is not updating {/*ispending-not-updating*/}
+
+If you're calling the action manually (not through a form's `action` prop), make sure you wrap the call in [`startTransition`](/reference/react/startTransition):
```
stephan-noel: **nit**:  Can be any action prop, not just a form's.

src/content/reference/react/useActionState.md
```diff
@@ -4,269 +4,1320 @@ title: useActionState
...
+import { useActionState, startTransition } from 'react';
+import { addToCart } from './api';
+import Total from './Total';
+
+async function addTicket(prevCount) {
+  return await addToCart(prevCount);
+}
+
+export default function Checkout() {
+  const [count, action, isPending] = useActionState(addTicket, 0);
```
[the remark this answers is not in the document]
stephan-noel: Kinda related on when you don't need `useActionState` in other cases, I think people may wonder, why not invoke the action directly and manage `isPending` and the return value with `useState()` for some cases.  I think you covered it in your talk, and maybe something for the Learn section, but felt relevant.

src/content/reference/react/useActionState.md
```diff
@@ -4,269 +4,1320 @@ title: useActionState
...
+<Pitfall>
+
+Aborting an Action isn't always safe, which is why `useActionState` doesn't do it by default.
 
-</Recipes>
+</Pitfall>
+
+---
 
 ## Troubleshooting {/*troubleshooting*/}
```
stephan-noel, line 1381: Not sure if here is a good spot, but I think people tend to lean towards `useEffect` when using the state return value of `useActionState`.  For example, [this long thread on showing a toast](https://x.com/sebastienlorber/status/1897431285267620059?s=20).  In the end, couldn't come up with a solution that worked with progressive enhancement without it.  I think the crux of the issue might be trying to mix uncontrolled apis with it.

Maybe something more for the Learn section, not sure.

samselikoff:
> @samselikoff re: naming - the actions don't need to be async or could be a mix of sync and async. For example, you could have `useActionState` where the "action" side effect is to call `showNotification` (a synchronous API)

Yep I mentioned that above, but the fact that they *can* be async (and normal reducers cannot) seems to be a pretty crucial differentiator!

src/content/reference/react/useActionState.md
stephan-noel: If I were to summarize the main use cases of `useActionState` as 1) queing updates to resolve out of order updates and 2) progressive enhancement for showing feedback and navigating (only for forms), would that be correct?

I feel like the _why_ behind it is too buried here and a summary of it (with some more focus on the problem it solves) near the beginning would be helpful.

I'm also curious, why was this hook created?  Like what was the main problem it was trying to solve and which features are just coincidental or add-ons?  I know we started with useFormState and then tried to generalize, but now the overlapping use cases and motivation make it all kinda fuzzy to me especially as usage with forms isn't emphasized as much.

PS: reading https://github.com/facebook/react/pull/28491 now to see if it clicks.

stefanprobst:
> I wonder if "preserving a form's inputs after a failed submission" warrants its own section in Usage.

also the "Save draft" example here: https://react.dev/reference/react-dom/components/form#handling-multiple-submission-types should be updated to not clear the input when saving the draft

src/content/reference/react/useActionState.md
rickhanlonii: Updated to:

> `useActionState` is a React Hook that updates state with side effects using Actions

src/content/reference/react/useActionState.md
rickhanlonii: Renamed to `dispatchAction`

src/content/reference/react/useActionState.md
rickhanlonii: Yeah I'm going to go with `actionPayload` to match the types. Makes sense as it's the payload for the action.

rickhanlonii: Removed

src/content/reference/react/useActionState.md
rickhanlonii: Updated to something slightly different. In this part of the reference I don't want to things like `quantity` I want more conceptual names like `initialState`

src/content/reference/react/useActionState.md
rickhanlonii: Yeah this is tricky if you don't have a side server side router to handle the URL changes without javascipt. Let's add it in the future.

src/content/reference/react/useActionState.md
rickhanlonii: Updated - but note that we currently over-entangle transitions, so if there's a different pending action that starts after this one is dispatched, then it will stay true until all pending actions (from this hook or otherwise) settle. 

Added an additional caveat for this.

src/content/reference/react/useActionState.md
rickhanlonii: Same as https://github.com/reactjs/react.dev/pull/8284/changes#r2761460149. This is hard, and most users can't do this without a framework since they don't have server-side route support anyway. Let's follow up with it.

src/content/reference/react/useActionState.md
rickhanlonii: Nice, good idea. @stephan-noel can you submit an issue for that?

src/content/reference/react/useActionState.md
rickhanlonii: Updated

rickhanlonii: I think calling it `dispatchAction` helps. You have to dispatch an action inside of an action. And an Action is just something called inside startTransition.

src/content/reference/react/useActionState.md
rickhanlonii: nice catch, done!

rickhanlonii: I don't think so. It's in the form specific example, and in the form action docs. I'm trying to remove confusion that the "Actions" concept is not tied to form actions specifically.

src/content/reference/react/useActionState.md
rickhanlonii: Done.

rickhanlonii: It's funny because claude initially included that and I removed it,.

src/content/reference/react/useActionState.md
rickhanlonii: yeah claude added that and i remove it

you don't need to do bind here, you can just inline the reducer. The reason you'd need this it with form actions and progressive enhancement, which is documented on the form action docs.

src/content/reference/react/useActionState.md
rickhanlonii: Agreed, updated!

src/content/reference/react/useActionState.md
rickhanlonii: Yeah, I don't want to import an action from `./api` like it's a server function. The idea is that there's the addToCartAction, and then there's the addToCart API POST.

I think I solved this by in-lining the action, and then the next examples call different API methods, which make the separation more clear.

src/content/reference/react/useActionState.md
rickhanlonii: I kinda like it, but I also don't think people should write this because when you want to add optimistic state or call another action you have to refactor.

src/content/reference/react/useActionState.md
rickhanlonii: All of the patterns are the reducer pattern - since the `reducerAction` _always_ receives the previous state.

src/content/reference/react/useActionState.md
rickhanlonii: Added

src/content/reference/react/useActionState.md
rickhanlonii: nice catch

rickhanlonii:
@samselikoff re: form action reset - I think we should add that to the [form action docs here](https://react.dev/reference/react-dom/components/form#handle-form-submission-with-a-server-function).

The form action docs need a revamp of their own

src/content/reference/react/useActionState.md
rickhanlonii: I don't really understand what the issue there is - you can't show a toast if there's no JS. So the flow is: permalink before JS (no toast), then "upgrade" with JS (the useActionState in the tweet). Could you post a question in the Async React working group explaining the use case?

gaearon:
One thing that's maybe worth calling out after the first demo is that we'll add optimistic handling in another demo (linked). It feels very broken that there's no immediate feedback.

src/content/reference/react/useActionState.md
```diff
@@ -4,269 +4,1325 @@ title: useActionState
...
+#### Returns {/*reduceraction-returns*/}
+
+`reducerAction` returns the new state, and triggers a Transition to re-render with that state.
+
+#### Caveats {/*reduceraction-caveats*/}
+
+* `reducerAction` can be sync or async. It can perform sync actions like showing a notification, or async actions like posting updates to a server. 
+* `reducerAction` is not invoked twice in `<StrictMode>` since `reducerAction` is designed to allow side effects.
+* The return type of `reducerAction` must match the type of `initialState`. If TypeScript infers a mismatch, you may need to explicitly annotate your state type.
+* If you set state after `await` in the `reducerAction` you currently need to wrap the state update in an additional `startTransition`. See the [startTransition](/reference/react/useTransition#react-doesnt-treat-my-state-update-after-await-as-a-transition) docs for more info.
```
MaxwellCohen, line 106: [Suggestion] Add something about when `reducerAction`  returns a promise to this caveat, because `reducerAction` can return a promise or a value that does not have to be an async function.  Since reducerAction can return both promises and the value directly is an amazing feature of this hook.


 
```If you set state after `await` in the `reducerAction` or  `reducerAction` returns a promise You need to wrap the state update in an additional `startTransition`. See the [startTransition](/reference/react/useTransition#react-doesnt-treat-my-state-update-after-await-as-a-transition) docs for more info.```


I can see not adding the last line and going with the stance that reducerAction must be called in a transition, as it could cause confusion vs developers saving a little bit of code

edit: While listening to Ryan Carniato's stream Sync to Async (https://www.youtube.com/watch?v=drLX0yTKP04), I think that having the strict useTransition message is better for overall useablity so dont't have a line like ' If `reducerAction` is sync (returns a value directly), `startTransition` is not needed. "

File: src/content/reference/react/useActionState.md (hunk 18 above)
Comment (rickhanlonii) [reply src/content/reference/react/useActionState.md]: Added an example in the "Handling Errors" usage.

**Comment by rickhanlonii** [comment]
@gaearon yeah good idea - but there is feedback right? the quantity and the total show a spinner (though really it should be on the button)

**Comment by gaearon** [comment]
Ah yeah, I just meant it feels very strange for a number to not increase

**Comment by gaearon** [comment]
Btw I'm also maybe team "async reducer"..

**Comment by gaearon** [comment]
Or maybe "action reducer"

**Comment by rickhanlonii** [comment]
Yeah I can see that, especially in the second example more than the first. 

The naming
 - async reducer is maybe nice, but when you start explaining it there's no continuity of terms.
 
 I also like how:
   - `useReducer`: `dispatch` -> `reducer`. 
   - `useActionState`: `dispatchAction` -> `reducerAction`. 
     - Since it's an Action, you can call `useOptimistic` in it, perform (async) side effects, etc.
     
In other words, since Action means "side effect that is maybe async", it is a async reducer but that's only a subset.

**Comment by rickhanlonii** [comment]
With the "async reducer" naming - what is the Action you're tracking the state of for `useActionState`? It's not the `dispatch`. The Action is the reducer, or there are multiple Actions in the reducer.

File: src/content/reference/react/useActionState.md [hunk 26]
```diff
@@ -4,269 +4,1509 @@ title: useActionState
 
 <Intro>
 
-`useActionState` is a Hook that allows you to update state based on the result of a form action.
+`useActionState` is a React Hook that lets you update state with side effects using [Actions](/reference/react/useTransition#functions-called-in-starttransition-are-called-actions).
```
stephan-noel, line 7: **nit**:  The "with side effects" threw me a bit until I realized (I think) that it's referring to the actionReducer being able to contain side effects?

Not sure of what's the best wording here, but yea, just wanted to leave my thoughts.

src/content/reference/react/useActionState.md
```diff
@@ -4,269 +4,1509 @@ title: useActionState
...
+`useActionState` returns an array with exactly three values:
+
+1. The current state. During the first render, it will match the `initialState` you passed. After `dispatchAction` is invoked, it will match the value returned by the `reducerAction`.
+2. A `dispatchAction` function that you call inside [Actions](/reference/react/useTransition#functions-called-in-starttransition-are-called-actions).
+3. The `isPending` flag that tells you if any dispatched Actions for this Hook are pending.
+
+#### Caveats {/*caveats*/}
+
+* `useActionState` is a Hook, so you can only call it **at the top level of your component** or your own Hooks. You can't call it inside loops or conditions. If you need that, extract a new component and move the state into it.
+* React queues and executes multiple calls to `dispatchAction` sequentially, allowing each `reducerAction` to use the result of the previous Action.
```
stephan-noel: **nit**:  "allowing each call to `reducerAction` to use..." or "allowing each `reducerAction` invocation to use..."?

"each `reducerAction` to use" seems like it's implying there are multiple reducerActions.

src/content/reference/react/useActionState.md
```diff
@@ -4,269 +4,1509 @@ title: useActionState
...
+Unlike reducers in `useReducer`, the `reducerAction` can be async and perform side effects:
+
+```js
+async function reducerAction(previousState, actionPayload) {
+  const newState = await post(actionPayload);
+  return newState;
+}
+```
+
+Each time you call `dispatchAction`, React calls the `reducerAction` with the `actionPayload`. The reducer will perform side effects such as posting data, and return the new state. If `dispatchAction` is called multiple times, React queues and executes them in order so the result of the previous call is available for the current call.
```
stephan-noel: **nit**:  "available **to** the current call" or "available for the current call **to use**".

It just read a bit off.

src/content/reference/react/useActionState.md
```diff
@@ -4,269 +4,1509 @@ title: useActionState
...
+
+#### Parameters {/*reduceraction-parameters*/}
+
+* `previousState`: The current state. Initially this is equal to the `initialState`. After the first call to `dispatchAction`, it's equal to the last state returned.
```
stephan-noel: It did make me pause to see it called `previousState` and then immediately read "The current state".  I get it, but did I have to think.  In the [new useOptimistic docs](https://react.dev/reference/react/useOptimistic#useoptimistic) it says `currentState` for the optional reducer, but I know "previous" also tends to be used.  Just kinda confusing to read it like this.

src/content/reference/react/useActionState.md
```diff
@@ -4,269 +4,1509 @@ title: useActionState
...
+
+* **optional** `actionPayload`: The argument passed to `dispatchAction`. It can be a value of any type. Similar to `useReducer` conventions, it is usually an object with a `type` property identifying it and, optionally, other properties with additional information.
```
stephan-noel, line 95: You mention `initialState` needs to be serializable when using Server Functions, but the same also holds for `actionPayload` right?  Should that be mentioned in the caveats section below?

rickhanlonii:
Ok @gaearon I just pushed a change I'm pretty excited about, this feels clean af now.

I updated all the pending states, so the feedback is better. More importantly, I moved the `useOptimistic` example up, and added `useOptimistic` to the "Using with Action props" so that the design component adds the optimistc state. 

Which lets me explain this:

> Since `<QuantityStepper>` has built-in support for transitions, pending state, and optimistically updating the count, you just need to tell the Action _what_ to change, and _how_ to change it is handled for you.

So the first three examples kinda build it up, and just when you think it's getting complicated, the Action prop example simplifies it all.

rickhanlonii:
The abortable section could be a re-usable hook btw:

```js
function useAbortableActionState(reducerAction, initialState) {
  const abortRef = useRef(null);
  const [state, dispatchAction, isPending] = useActionState(reducerAction, initialState);

  async function wrappedAction(payload) {
    if (abortRef.current) {
      abortRef.current.abort();
    }
    abortRef.current = new AbortController();
    await dispatchAction({ type: 'ADD', signal: abortRef.current.signal });
  }

  return [state, wrappedAction, isPending];
}
```

MaxwellCohen:
We are missing a troubleshooting section for the following console errors (it might be implied, but not exact text). 


- "An async function with useActionState was called outside of a transition. This is likely not what you intended (for example, isPending will not update correctly). Either call the returned function inside startTransition, or pass it to an `action` or `formAction` prop."
- "Cannot update form state while rendering."

src/content/reference/react/useActionState.md
```diff
@@ -4,269 +4,1493 @@ title: useActionState
...
+
+
+`setOptimisticCount` immediately updates the quantity, and `dispatchAction()` queues the `updateCartAction`. A pending indicator appears on both the quantity and total to give the user feedback that their update is still being applied.
+
+---
+
+
+### Using with Action props {/*using-with-action-props*/}
+
+When you pass the `dispatchAction` function to a component that exposes an [Action prop](/reference/react/useTransition#exposing-action-props-from-components), you don't need to call `startTransition` or `useOptmisitc` yourself.
```
icyJoseph:
```suggestion
When you pass the `dispatchAction` function to a component that exposes an [Action prop](/reference/react/useTransition#exposing-action-props-from-components), you don't need to call `startTransition` or `useOptimistic` yourself.
```

src/content/reference/react/useActionState.md
```diff
@@ -4,269 +4,1493 @@ title: useActionState
...
+* When using Server Functions, `initialState` needs to be [serializable](/reference/rsc/use-server#serializable-parameters-and-return-values) (values like plain objects, arrays, strings, and numbers).
+* If `dispatchAction` throws an error, React cancels all queued actions and shows the nearest [Error Boundary](/reference/react/Component#catching-rendering-errors-with-an-error-boundary).
+* If there are multiple ongoing Actions, React batches them together. This is a limitation that may be removed in a future release.
+
 <Note>
 
-In earlier React Canary versions, this API was part of React DOM and called `useFormState`.
+`dispatchAction` must be called within a Transition. 
+
+You can wrap it in [`startTransition`](/reference/react/startTransition), or pass it to an [Action prop](/reference/react/useTransition#exposing-action-props-from-components).
```
hernan-yadiel: Above, in the Returns section it was mentioned that:

> A `dispatchAction` function that you call inside Actions

I found a bit confusing that in this Note it is mentioned that a "`dispatchAction` must be called within a Transition."

I suggest something like:

```suggestion
`dispatchAction` must be called from an Action (i.e. inside startTransition), or passed to an Action prop.

Calls outside that scope won’t be treated as part of the Transition and log an error on development mode.
```

src/content/reference/react/useActionState.md
```diff
@@ -4,269 +4,1493 @@ title: useActionState
...
+    startTransition(async () => {
+      setOptimisticValue(c => c + 1);
+      await increaseAction();
     });
-    return "Couldn't add to cart: the item is sold out.";
   }
+
+  function handleDecrease() {
+    startTransition(async () => {
+      setOptimisticValue(c => c + 1);
```
stephan-noel: Small bug.

```suggestion
      setOptimisticValue(c => c - 1);
```

Also, I guess https://github.com/reactjs/react.dev/pull/8284/changes#r2776729030 is relevant here as well.

src/content/reference/react/useActionState.md
```diff
@@ -4,269 +4,1493 @@ title: useActionState
...
+---
 
 ## Troubleshooting {/*troubleshooting*/}
 
-### My action can no longer read the submitted form data {/*my-action-can-no-longer-read-the-submitted-form-data*/}
+### My Action can no longer read the submitted form data {/*action-cant-read-form-data*/}
+
+When you use `useActionState`, the `reducerAction` receives an extra argument as its first argument: the previous or initial state. The submitted form data is therefore its second argument instead of its first.
+
+```js
```
stephan-noel: Small thing, not sure.

```suggestion
```js {2,7}
```

src/content/reference/react/useActionState.md
```diff
@@ -4,269 +4,1493 @@ title: useActionState
...
+  function handleAdd() {
+    startTransition(() => {
+      setOptimisticCount(c => c + 1);
+      dispatchAction({ type: 'ADD' });
+    });
+  }
+
+  function handleRemove() {
+    startTransition(() => {
+      setOptimisticCount(c => c - 1);
```
stephan-noel, line 476: **nit**:  Maybe not worth the additional complexity, but this allows negative tickets whereas removeFromCart doesn't and will revert to 0.

src/content/reference/react/useActionState.md
rickhanlonii: Agreed but I don't want to block these docs on it.

src/content/reference/react/useActionState.md
rickhanlonii: Nice catch

src/content/reference/react/useActionState.md
rickhanlonii: Nice catch

rickhanlonii:
Ok I think this is good to go if someone want's to stamp it

samselikoff:
My first reaction on re-reading was that `reducerAction` should be called `actionReducer`... isn't the thing you're defining a reducer? Isn't the reducer the noun?

rickhanlonii:
Yeah I agree @samselikoff, but the convention is that Action is a suffix

samselikoff:
But only on the public API side, right? Like if I define or expose an `updateAction`. Whereas this is an argument, not a function called by the user.

src/content/reference/react/useActionState.md
```diff
@@ -4,269 +4,1562 @@ title: useActionState
...
+    startTransition(() => {
+      dispatchAction(null); // Pass null to trigger reset
+    });
+  }
+
+  // ...
+}
+```
+
+Alternatively, you can add a `key` prop to the component using `useActionState` to force it to remount with fresh state, or a `<form>` `action` prop, which resets automatically after submission.
```
icyJoseph, line 1478: > or a `<form>` `action` prop, which resets automatically after submission.

What did we mean here, in terms of resetting useActionState? Is it for the uncontrolled inputs that get reset after submit?

rickhanlonii: when a `<form>` is submitted with the `action` prop, React resets the form inputs

PR #8293, opened 2026-02-08
[errors] Add more info to "Invalid Hook Call" error page

rickhanlonii:
## [Preview](https://react-dev-git-fork-rickhanlonii-rh-377-fbopensource.vercel.app/errors/321?_vercel_share=Z327uYcnSvu7eHaPiGofl1NK3jrORMoI)

_Depends on https://github.com/reactjs/react.dev/pull/8291_

We already have the infra to add a `errors/{code}.md` pages that gives more info about specific errors like why they happen and common causes/solutions. This will be especially useful for Agents that hit errors, so I'm going to use claude to start start creating these pages.

Screenshot 2026-02-08 at 5 41 55 PM

### DIFF
--- a/src/content/errors/321.md
+++ b/src/content/errors/321.md
@@ -1,0 +1,127 @@
+<Intro>
+
+This page explains this React error and common ways to fix it.
+
+</Intro>
+
+The full text of the error is:
+
+<ErrorDecoder />
+
+<Note>
+
+In the minified production build of React, full error messages are replaced with short codes to reduce bundle size. We recommend using the development build when debugging, as it includes additional warnings and debug information.
+
+</Note>
+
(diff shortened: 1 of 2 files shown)


src/content/errors/321.md
```diff
@@ -0,0 +1,127 @@
...
+
+```bash
+npm ls react
+```
+
+If you see more than one entry for `react`, you have duplicate copies. You can also add a temporary log to confirm. Add this at the top of your component file:
+
+```js
+import React from 'react';
+console.log(React === window.React); // false means duplicates
```
icyJoseph, line 103: I think here we need a caveat, or further instruction, as in, you first need to identify where a different React version might be used. And assign that React version to `window.React`.

PR #8298, opened 2026-02-10
[RFC] Add <PerformanceTracks>

rickhanlonii:
_This is mostly vibe coded so I need to review and edit this myself, but I wanted to open an RFC to see if this is something that is valuable to add._

## Overview

Adds a `<PeformanceTracks>` component that allows you to render a docs-branded diagram of the React Performance Tracks.  Also includes a `convertTrace` script that converts an exported performance trace to a copyable `<PerformanceTracks>` snippet for that trace:

```bash
node scripts/convertTrace.js trace.json
```

```js
<PerformanceTracks
  data={{
    duration: 1000,
    tracks: {
      scheduler: {
        blocking: [
          {
            type: "event",
            name: "click",
            start: 0,
            duration: 76,
          },
          // ...
        ],
        // ...
      },
      // ...
    },
  }}
/>
```

## Features
- Ensures the entries expand enough to show the label so "Event: click" is not collapsed to "Ev..."
- Uses a log(2) scale to reduce the length of entries (like long Suspense entries)
- Scrollable, but attempts to scale down if it's close to not needing to scroll
- Can add annotations to call out parts of the trace 
  - I think these can be improved a lot

### Dark Mode
Screenshot 2026-02-10 at 1 06 33 PM

### Light Mode

Screenshot 2026-02-10 at 1 08 22 PM

## Video

https://github.com/user-attachments/assets/c2012bdf-cc18-42b9-aa07-63dd08a4c1d1

# TODO

- [ ] Actually review and clean up the code

### DIFF
--- a/scripts/convertTrace.js
+++ b/scripts/convertTrace.js
@@ -1,0 +1,537 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+'use strict';
+
+const ID_MAP = {
+  'Scheduler \u269B': 'scheduler',
+  Scheduler: 'scheduler',
+  'Components \u269B': 'components',
+  Components: 'components',
+  Blocking: 'blocking',
+  Transition: 'transition',
+  Suspense: 'suspense',
+  Idle: 'idle',
+};
+
+function toId(name) {
+  return (
+    ID_MAP[name] ||
+    name
+      .toLowerCase()
+      .replace(/[^a-z0-9]+/g, '-')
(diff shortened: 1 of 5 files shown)


MaxwellCohen:
I really like the idea of this component. Having a clearer understanding of the computation tasks React executes is very helpful for building better apps with React.
Some feedback 
1) Leverage `React.lazy` or next `dynamic` to lazy load, and add treeshake to this component for the overall React.dev performance, since it won't be on every page (there might be other components that should be lazy loaded too). The first load sizes seem a little large overall. 
2) The mobile and small screen view is less than ideal
image

3) The accessibility tree in Chrome makes it look like this component would not be very useful for screen readers (just listing out React render steps).

PR #8300, opened 2026-02-11
Add RSC Sandboxes

rickhanlonii:
## Overview 

Adds browser-only `<SandpackRSC>` component for RSC sandboxes.

This uses the browser version of `react-server-dom-webpack` to run RSCs in a web worker, so there's no custom sandpack bundler. In the worker, we use `sucrase` to transpile JSX and then send it back to the client.

It currently does not support opening it in an external sandbox, but we can set that up with a new custom template (might need help from the sandpack folks).


rickhanlonii:
cc @lubieowoce @gaearon

gaearon:
you might find https://github.com/gaearon/rscexplorer helpful, I'm relatively sure it works correctly but it has a bunch of stuff you don't need

rickhanlonii:
thanks @gaearon yeah i gave that to claude as a reference as well as @lubieowoce's branch

.claude/skills/docs-rsc-sandpack/SKILL.md
```diff
@@ -0,0 +1,407 @@
...
+
+## File Directives
+
+Files are classified by the directive at the top of the file:
+
+| Directive | Where it runs | Rules |
+|-----------|--------------|-------|
+| (none) | Web Worker (server) | Default. Can be `async`. Can import other server files. Cannot use hooks, event handlers, or browser APIs. |
+| `'use client'` | Sandpack iframe (browser) | Must be first statement. Can use hooks, event handlers, browser APIs. Cannot be `async`. Cannot import server files. |
+| `'use server'` | Web Worker (server) | Marks server actions. Can be module-level (all exports are actions) or function-level. Callable from client via props or form `action`. |
```
eps1lon:
```suggestion
| `'use server'` | Web Worker (server) | Marks server functions. Can be module-level (all exports are server functions) or function-level. Callable from client via props or form `action`. |
```

.claude/skills/docs-rsc-sandpack/SKILL.md
```diff
@@ -0,0 +1,407 @@
...
+
+---
+
+## Constraints and Pitfalls
+
+### What Server Components Cannot Do
+
+| Cannot | Reason |
+|--------|--------|
+| Use hooks (`useState`, `useEffect`, etc.) | Server components are stateless, run once |
```
eps1lon: This is misleading. Server Component's can't use Client Hooks. They can use some Hooks to enable authoring shared Components (e.g. `useMemo`). Full list is in https://github.com/facebook/react/blob/ab18f33d46171ed1963ae1ac955c5110bb1eb199/packages/react/src/ReactServer.js#L61-L64

.claude/skills/docs-rsc-sandpack/SKILL.md
```diff
@@ -0,0 +1,407 @@
...
+| Attach event handlers (`onClick`, `onChange`) | No interactivity on server |
+| Access browser APIs (`window`, `document`, `localStorage`) | Runs in Web Worker, not browser |
+| Import client components without `'use client'` | Client code must be explicitly marked |
+| Use `useContext` | Context is a client concept |
+
+### What Client Components Cannot Do
+
+| Cannot | Reason |
+|--------|--------|
+| Be `async` | Client components render synchronously |
```
eps1lon: Should we include that Client Components can use `React.use` to unwrap Promises? Ideally the agent would be able to reason that out but mentioning `use` makes it more implicit.

.claude/skills/docs-rsc-sandpack/SKILL.md
```diff
@@ -0,0 +1,407 @@
...
+| Import server-only files directly | Would execute server code in browser |
+| Use `'use server'` at module level | That marks a server actions file |
```
eps1lon:
```suggestion
| Use `'use server'` at module level | That marks a server functions file |
```

.claude/skills/docs-rsc-sandpack/SKILL.md
```diff
@@ -0,0 +1,407 @@
...
+
+### Other Constraints
+
+| Constraint | Detail |
+|------------|--------|
+| FormData serialization | FormData is not structurally cloneable for `postMessage`; serialized as `[key, value]` entries and reconstructed in Worker |
+| Module resolution | Custom `resolvePath()` — only relative paths (`./`, `../`) with `.js`, `.jsx`, `.ts`, `.tsx` extensions. No `node_modules` resolution. |
+| CSS imports | `require('*.css')` returns empty object. No CSS bundling in the Worker. |
+| Circular dependencies | Partially supported — partially populated exports returned during circular `require()` |
+| React version | Client: `^19.2.1` via Sandpack CDN. Server: pinned to pre-bundled version in `worker-bundle.source.js`. Must stay in sync. |
```
eps1lon: I'm not sure we can have Client and Server version drift. Can't we pin the Client version as well?

.claude/skills/docs-rsc-sandpack/SKILL.md
```diff
@@ -0,0 +1,407 @@
...
+| `src/components/MDX/Sandpack/sandpack-rsc/RscFileBridge.tsx`     | Monitors Sandpack; posts raw files to iframe |
+| `src/components/MDX/Sandpack/sandpack-rsc/sandpack-rsc-setup.ts` | Loads raw source files, assembles infrastructure file map |
+| `.../sandbox-code/src/worker-server.source.js`                   | Worker runtime: module system, Sucrase compilation, `renderToReadableStream()` |
+| `.../sandbox-code/src/rsc-client.source.js`                      | Client bridge: Worker creation, file classification, Flight stream consumption |
+| `.../sandbox-code/src/webpack-shim.source.js`                    | Minimal `__webpack_require__` / `__webpack_module_cache__` shim |
+| `.../sandbox-code/src/worker-bundle.source.js`                   | Pre-bundled IIFE (generated): React server + RSDW/server + Sucrase |
+| `scripts/buildRscWorker.mjs`                                     | esbuild script: bundles worker-server.source.js into worker-bundle.source.js |
+| `next.config.js` (lines 49-60)                                   | Webpack `raw-loader` rule for `.source.js` files |
```
eps1lon: This is outdated I believe. There are no changes to `next.config.js` in this PR anymore.

.claude/skills/docs-rsc-sandpack/SKILL.md
```diff
@@ -0,0 +1,407 @@
...
+
+## Build System
+
+### Rebuilding the Worker Bundle
+
+After modifying `worker-server.source.js` or `webpack-shim.source.js`:
+
+```bash
+node scripts/buildRscWorker.mjs
+```
```
eps1lon: Can we do this in CI to validate this has been done? Pretty easy to forget otherwise.

src/components/MDX/Sandpack/sandpack-rsc/sandbox-code/src/rsc-server.js
```diff
@@ -0,0 +1,541 @@
...
+    return resultPromise.then(function () {
+      var App = deployed.module.default || deployed.module;
+      return RSDWServer.renderToReadableStream(
+        {root: React.createElement(App), returnValue: resultPromise},
+        createModuleMap(),
+        {
+          onError: function (err) {
+            var msg = err && err.message ? err.message : String(err);
+            var stack = err && err.stack ? err.stack : '';
+            console.error('[RSC Server Error]', msg, stack);
```
eps1lon:
```suggestion
            console.error('[RSC Server Error]', err);
```

That way browsers can display the error correctly (including other properties like `cause`) while ensuring ignore-listing works.

There are more places that needlessly dissect the error instead of logging it as-is.

eps1lon (commented):
Super exciting to get these going.

/learn/rsc-sandbox-test produces a 404 when deployed. Works locally though.

> I think something fucky is happening

What's that referring to?

> does this stream? can it?

It does the way you structured it now. Each Flight chunk is sent separately with postMessage which is the streaming way on the Server. It wouldn't stream if you'd accumulate all chunks before sending them. The Client puts the chunks into a stream as well that's passed to the Client for decoding before all chunks have. So the Client is streaming as well.

>  Editing sandboxes can get into a broken state

Seems fine from my random testing locally.

rickhanlonii:
@eps1lon thanks! rsc-sandbox-test is gated to DEV only. 

The something fucky part is old, i don't remember what it was though

.claude/skills/docs-rsc-sandpack/SKILL.md
rickhanlonii: Oh yeah this skill is way out of date, i'll review and update it all

.claude/skills/docs-rsc-sandpack/SKILL.md
rickhanlonii: it will automatically run on `yarn build`

rickhanlonii: We could add a CI check for it, but not sure if it's worth running a CI check on every diff for somthing that will in all likelyhood almost never need to run again?

src/components/MDX/Sandpack/sandpack-rsc/sandbox-code/src/rsc-server.js
rickhanlonii: Updated on the server, on the client we need to access digest because the server is actually running the prod build. I tried to get it to run the client version, but I couldn't get it to work. 

Screenshot 2026-02-28 at 11 11 09 AM

rickhanlonii:
OK feedback updates done, I think this is ready for final feedback

.claude/skills/docs-rsc-sandpack/SKILL.md
eps1lon: worker-server.source is a product of rsc-server? There's plenty of things we might change there. Definitely for when we showcase console replaying.

eps1lon:
> thanks! rsc-sandbox-test is gated to DEV only.

Gotcha, so we need an actual page where we demo it to test deploy behavior?

eps1lon (approved):
🚀

rickhanlonii:
Yeah, I confirmed it works in a preview deploy on an earlier commit before I made it dev-only.

I'll add it to an actual docs page next and we can confirm it works in that preview deploy too.

merlinaudio:
this is amazing. thank you for all your great work.

PR #8305, opened 2026-02-13
Rewrite `use()` docs

rickhanlonii:
# [Preview](https://react-dev-git-fork-rickhanlonii-rewrite-use-docs-fbopensource.vercel.app/reference/react/use?_vercel_share=JpVpWNp0fJza55SvUULA3TJWg0X6GclL)

cc @samselikoff @gaearon @stephan-noel @aurorascharff @brenelz @MaxwellCohen @hernan-yadiel

### DIFF
--- a/src/content/reference/react/use.md
+++ b/src/content/reference/react/use.md
@@ -18,45 +18,69 @@ const value = use(resource);
 
 ## Reference {/*reference*/}
 
-### `use(resource)` {/*use*/}
+### `use(context)` {/*use-context*/}
 
-Call `use` in your component to read the value of a resource like a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) or [context](/learn/passing-data-deeply-with-context).
+Call `use` with a [context](/learn/passing-data-deeply-with-context) to read its value. Unlike [`useContext`](/reference/react/useContext), `use` can be called within loops and conditional statements like `if`.
+
+```js
+import { use } from 'react';
+
+function Button() {
+  const theme = use(ThemeContext);
+  // ...
+```
+
+[See more examples below.](#usage-context)
+
+#### Parameters {/*context-parameters*/}
+
+* `context`: A [context](/learn/passing-data-deeply-with-context) created with [`createContext`](/reference/react/createContext).
+
+#### Returns {/*context-returns*/}
+
+The context value for the passed context, determined by the closest context provider above the calling component. If there is no provider, the returned value is the `defaultValue` passed to [`createContext`](/reference/react/createContext).
+
+#### Caveats {/*context-caveats*/}
+
+* `use` must be called inside a Component or a Hook.
+* Reading context with `use` is not supported in [Server Components](/reference/rsc/server-components).
+
+---
+
+### `use(promise)` {/*use-promise*/}
+
+Call `use` with a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) to read its resolved value. The component calling `use` *suspends* while the Promise is pending.
 
 ```js
 import { use } from 'react';
 
 function MessageComponent({ messagePromise }) {
   const message = use(messagePromise);
-  const theme = use(ThemeContext);
   // ...
 ```
 
-Unlike React Hooks, `use` can be called within loops and conditional statements like `if`. Like React Hooks, the function that calls `use` must be a Component or Hook.
+If the component that calls `use` is wrapped in a [`Suspense`](/reference/react/Suspense) boundary, the fallback will be displayed while the Promise is pending. Once the Promise is resolved, the Suspense fallback is replaced by the rendered components using the data returned by `use`. If the Promise is rejected, the fallback of the nearest [Error Boundary](/reference/react/Component#catching-rendering-errors-with-an-error-boundary) will be displayed.
 
-When called with a Promise, `use` integrates with [`Suspense`](/reference/react/Suspense) and [Error Boundaries](/reference/react/Component#catching-rendering-errors-with-an-error-boundary). The component calling `use` *suspends* while the Promise passed to `use` is pending. If the component that calls `use` is wrapped in a Suspense boundary, the fallback will be displayed. Once the Promise is resolved, the Suspense fallback is replaced by the rendered components using the data returned by `use`. If the Promise passed to `use` is rejected, the fallback of the nearest Error Boundary will be displayed.
+[See more examples below.](#usage-promises)
 
-[See more examples below.](#usage)
+#### Parameters {/*promise-parameters*/}
 
-#### Parameters {/*parameters*/}
+* `promise`: A [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) whose resolved value you want to read. The Promise must be [cached](#caching-promises-for-client-components) so that the same instance is reused across re-renders.
 
-* `resource`: This is the source of the data you want to read a value from. A resource can be a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) or a [context](/learn/passing-data-deeply-with-context).
+#### Returns {/*promise-returns*/}
 
-#### Returns {/*returns*/}
+The resolved value of the Promise.
 
-The `use` API returns the value that was read from the resource like the resolved value of a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) or [context](/learn/passing-data-deeply-with-context).
-
-#### Caveats {/*caveats*/}
+#### Caveats {/*promise-caveats*/}
 
 * `use` must be called inside a Component or a Hook.
-* `use` cannot be called inside a try-catch block. Instead, wrap your component in an [Error Boundary](#displaying-an-error-with-an-error-boundary), or use the Promise's [`.catch` method](#providing-an-alternative-value-with-promise-catch) to provide an alternative value.
-* When fetching data in a [Server Component](/reference/rsc/server-components), prefer `async` and `await` over `use`. `async` and `await` pick up rendering from the point where `await` was invoked, whereas `use` re-renders the component after the data is resolved.
-* Prefer creating Promises in [Server Components](/reference/rsc/server-components) and passing them to [Client Components](/reference/rsc/use-client) over creating Promises in Client Components. Promises created in Client Components are recreated on every render. Promises passed from a Server Component to a Client Component are stable across re-renders. [See this example.](#streaming-data-from-server-to-client)
-* Promises passed to `use` must be stable across renders. Creating a new Promise inside a component on every render will cause React to display the Suspense fallback on every re-render. [See caching Promises below.](#caching-promises-for-client-components)
-* Reading context with `use` is not supported in [Server Components](/reference/rsc/server-components).
+* `use` cannot be called inside a try-catch block. Instead, wrap your component in an [Error Boundary](#displaying-an-error-with-an-error-boundary) to catch the error and display a fallback.
+* Promises passed to `use` must be cached so the same Promise instance is reused across re-renders. [See caching Promises below.](#caching-promises-for-client-components)
+* When passing a Promise from a Server Component to a Client Component, its resolved value must be serializable to pass between server and client. Data types like functions aren't serializable and cannot be the resolved value of such a Promise.
 
 ---
 
-## Usage {/*usage*/}
+## Usage (Context) {/*usage-context*/}
 
 ### Reading context with `use` {/*reading-context-with-use*/}
 
@@ -198,6 +222,8 @@ function Button({ show, children }) {
 </Sandpack>
 
 ---
+
+## Usage (Promises) {/*usage-promises*/}
 
 ### Reading a Promise with `use` {/*reading-a-promise-with-use*/}
 
@@ -409,327 +435,76 @@ export async function fetchAlbums() {
 
 <Pitfall>
 
-##### Promises passed to `use` must be stable across renders {/*promises-must-be-stable*/}
+##### Promises passed to `use` must be cached {/*promises-must-cached*/}
 
-Promises created inside a component are recreated on every render. This causes React to show the Suspense fallback repeatedly and prevents content from appearing. Instead, pass a Promise from a Suspense-compatible data source. These include Suspense-enabled frameworks and client-side caching solutions such as a [`fetchData` wrapper](#caching-promises-for-client-components), or pass a Promise from a Server Component.
+Promises created inside a component are recreated on every render. This causes React to show the Suspense fallback repeatedly and prevents content from appearing. Instead, pass a Promise from a cache, a Suspense-enabled framework, or a Server Component.
 
 ```js
 function Albums() {
-  // 🔴 Creating a new Promise on every render
-  const albums = use(fetchAlbums());
+  // 🔴 fetch creates a new Promise on every render.
+  const albums = use(fetch('/albums'));
   // ...
 }
 ```
 
 ```js
-// ✅ Cache the Promise so it's stable across renders
+// ✅ fetchData reads the promise from a cache.
 const albums = use(fetchData('/albums'));
 ```
+
+Ideally, Promises are created before rendering — such as in an event handler, a route loader, or a Server Component — and passed to the component that calls `use`. Fetching lazily in render delays network requests and can create waterfalls.
 
 </Pitfall>
 
 ---
 
-### Streaming data from server to client {/*streaming-data-from-server-to-client*/}
+### Caching Promises for Client Components {/*caching-promises-for-client-components*/}
 
-Data can be streamed from the server to the client by passing a Promise as a prop from a <CodeStep step={1}>Server Component</CodeStep> to a <CodeStep step={2}>Client Component</CodeStep>.
+Promises passed to `use` in Client Components must be cached so the same Promise instance is reused across re-renders. If a new Promise is created directly in render, React will display the Suspense fallback on every re-render.
 
-```js [[1, 4, "App"], [2, 2, "Message"], [3, 7, "Suspense"], [4, 8, "messagePromise", 30], [4, 5, "messagePromise"]]
-import { fetchMessage } from './lib.js';
-import { Message } from './message.js';
+```js
+// ✅ Cache the Promise so the same one is reused across renders
+let cache = new Map();
 
-export default function App() {
-  const messagePromise = fetchMessage();
-  return (
-    <Suspense fallback={<p>waiting for message...</p>}>
-      <Message messagePromise={messagePromise} />
-    </Suspense>
-  );
-}
-```
-
-The <CodeStep step={2}>Client Component</CodeStep> then takes <CodeStep step={4}>the Promise it received as a prop</CodeStep> and passes it to the <CodeStep step={5}>`use`</CodeStep> API. This allows the <CodeStep step={2}>Client Component</CodeStep> to read the value from <CodeStep step={4}>the Promise</CodeStep> that was initially created by the Server Component.
-
-```js [[2, 6, "Message"], [4, 6, "messagePromise"], [4, 7, "messagePromise"], [5, 7, "use"]]
-// message.js
-'use client';
-
-import { use } from 'react';
-
-export function Message({ messagePromise }) {
-  const messageContent = use(messagePromise);
-  return <p>Here is the message: {messageContent}</p>;
-}
-```
-Because <CodeStep step={2}>`Message`</CodeStep> is wrapped in <CodeStep step={3}>[`Suspense`](/reference/react/Suspense)</CodeStep>, the fallback will be displayed until the Promise is resolved. When the Promise is resolved, the value will be read by the <CodeStep step={5}>`use`</CodeStep> API and the <CodeStep step={2}>`Message`</CodeStep> component will replace the Suspense fallback.
-
-<Sandpack>
-
-```js src/message.js active
-"use client";
-
-import { use, Suspense } from "react";
-
-function Message({ messagePromise }) {
-  const messageContent = use(messagePromise);
-  return <p>Here is the message: {messageContent}</p>;
-}
-
-export function MessageContainer({ messagePromise }) {
-  return (
-    <Suspense fallback={<p>⌛Downloading message...</p>}>
-      <Message messagePromise={messagePromise} />
-    </Suspense>
-  );
-}
-```
-
-```js src/App.js hidden
-import { useState } from "react";
-import { MessageContainer } from "./message.js";
-
-function fetchMessage() {
-  return new Promise((resolve) => setTimeout(resolve, 1000, "⚛️"));
-}
-
-export default function App() {
-  const [messagePromise, setMessagePromise] = useState(null);
-  const [show, setShow] = useState(false);
-  function download() {
-    setMessagePromise(fetchMessage());
-    setShow(true);
+export function fetchData(url) {
+  if (!cache.has(url)) {
+    cache.set(url, getData(url));
   }
-
-  if (show) {
-    return <MessageContainer messagePromise={messagePromise} />;
-  } else {
-    return <button onClick={download}>Download message</button>;
-  }
+  return cache.get(url);
 }
 ```
 
-```js src/index.js hidden
-import React, { StrictMode } from 'react';
-import { createRoot } from 'react-dom/client';
-import './styles.css';
-
-// TODO: update this example to use
-// the Codesandbox Server Component
-// demo environment once it is created
-import App from './App';
-
-const root = createRoot(document.getElementById('root'));
-root.render(
-  <StrictMode>
-    <App />
-  </StrictMode>
-);
-```
-
-</Sandpack>
+The `fetchData` function returns the same Promise each time it's called with the same URL. When `use` receives the same Promise on a re-render, it reads the already-resolved value synchronously without suspending.
 
 <Note>
 
-When passing a Promise from a Server Component to a Client Component, its resolved value must be serializable to pass between server and client. Data types like functions aren't serializable and cannot be the resolved value of such a Promise.
+The way you cache Promises depends on the framework you use with Suspense. Frameworks typically provide built-in caching mechanisms. If you don't use a framework, you can use a simple module-level cache like the one above, or a library that supports Suspense-compatible caching.
 
 </Note>
 
-
-<DeepDive>
-
-#### Should I resolve a Promise in a Server or Client Component? {/*resolve-promise-in-server-or-client-component*/}
-
-A Promise can be passed from a Server Component to a Client Component and resolved in the Client Component with the `use` API. You can also resolve the Promise in a Server Component with `await` and pass the required data to the Client Component as a prop.
-
-```js
-export default async function App() {
-  const messageContent = await fetchMessage();
-  return <Message messageContent={messageContent} />
-}
-```
-
-But using `await` in a [Server Component](/reference/rsc/server-components) will block its rendering until the `await` statement is finished. Passing a Promise from a Server Component to a Client Component prevents the Promise from blocking the rendering of the Server Component.
-
-</DeepDive>
-
----
-
-### Displaying an error with an Error Boundary {/*displaying-an-error-with-an-error-boundary*/}
-
-If the Promise passed to `use` is rejected, the error propagates to the nearest [Error Boundary](/reference/react/Component#catching-rendering-errors-with-an-error-boundary). Wrap the component that calls `use` in an Error Boundary to display a fallback when the Promise is rejected.
-
-<Sandpack>
-
-```js src/message.js active
-"use client";
-
-import { use, Suspense } from "react";
-import { ErrorBoundary } from "react-error-boundary";
-
-export function MessageContainer({ messagePromise }) {
-  return (
-    <ErrorBoundary fallback={<p>⚠️Something went wrong</p>}>
-      <Suspense fallback={<p>⌛Downloading message...</p>}>
-        <Message messagePromise={messagePromise} />
-      </Suspense>
-    </ErrorBoundary>
-  );
-}
-
-function Message({ messagePromise }) {
-  const content = use(messagePromise);
-  return <p>Here is the message: {content}</p>;
-}
-```
-
-```js src/App.js hidden
-import { useState } from "react";
-import { MessageContainer } from "./message.js";
-
-function fetchMessage() {
-  return new Promise((resolve, reject) => setTimeout(reject, 1000));
-}
-
-export default function App() {
-  const [messagePromise, setMessagePromise] = useState(null);
-  const [show, setShow] = useState(false);
-  function download() {
-    setMessagePromise(fetchMessage());
-    setShow(true);
-  }
-
-  if (show) {
-    return <MessageContainer messagePromise={messagePromise} />;
-  } else {
-    return <button onClick={download}>Download message</button>;
-  }
-}
-```
-
-```js src/index.js hidden
-import React, { StrictMode } from 'react';
-import { createRoot } from 'react-dom/client';
-import './styles.css';
-
-// TODO: update this example to use
-// the Codesandbox Server Component
-// demo environment once it is created
-import App from './App';
-
-const root = createRoot(document.getElementById('root'));
-root.render(
-  <StrictMode>
-    <App />
-  </StrictMode>
-);
-```
-
-```json package.json hidden
-{
-  "dependencies": {
-    "react": "19.0.0",
-    "react-dom": "19.0.0",
-    "react-scripts": "^5.0.0",
-    "react-error-boundary": "4.0.3"
-  },
-  "main": "/index.js"
-}
-```
-</Sandpack>
-
----
-
-### Providing an alternative value with `Promise.catch` {/*providing-an-alternative-value-with-promise-catch*/}
-
-If you'd like to provide an alternative value when the Promise passed to `use` is rejected, you can use the Promise's <CodeStep step={1}>[`catch`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/catch)</CodeStep> method.
-
-```js [[1, 6, "catch"],[2, 7, "return"]]
-import { Message } from './message.js';
-
-export default function App() {
-  const messagePromise = new Promise((resolve, reject) => {
-    reject();
-  }).catch(() => {
-    return "no new message found.";
-  });
-
-  return (
-    <Suspense fallback={<p>waiting for message...</p>}>
-      <Message messagePromise={messagePromise} />
-    </Suspense>
-  );
-}
-```
-
-To use the Promise's <CodeStep step={1}>`catch`</CodeStep> method, call <CodeStep step={1}>`catch`</CodeStep> on the Promise object. <CodeStep step={1}>`catch`</CodeStep> takes a single argument: a function that takes an error message as an argument. Whatever is <CodeStep step={2}>returned</CodeStep> by the function passed to <CodeStep step={1}>`catch`</CodeStep> will be used as the resolved value of the Promise.
-
----
-
-### Showing stale content with `useTransition` {/*showing-stale-content-with-usetransition*/}
-
-When data changes (for example, when navigating between pages), you can combine `use` with [`useTransition`](/reference/react/useTransition) to keep displaying the previous content while new data loads. The `isPending` flag from `useTransition` lets you show a loading indicator while keeping old content visible.
-
-```js
-function App({ albumsPromise }) {
-  const albums = use(albumsPromise);
-  const [isPending, startTransition] = useTransition();
-  // ...
-}
-```
-
-When an update is wrapped in `startTransition`, React keeps showing the current content instead of switching to the Suspense fallback. The component re-renders with the new data once the Promise resolves.
-
 <Sandpack>
 
 ```js src/App.js active
-import { use, Suspense, useState, useTransition } from 'react';
+import { use, Suspense, useState } from 'react';
 import { fetchData } from './data.js';
 
 export default function App() {
-  const [artistId, setArtistId] = useState('the-beatles');
-  const [isPending, startTransition] = useTransition();
+  const [count, setCount] = useState(0);
   return (
     <>
-      <div>
-        <button
-          onClick={() => {
-            startTransition(() => {
-              setArtistId('the-beatles');
-            });
-          }}
-        >
-          The Beatles
-        </button>
-        <button
-          onClick={() => {
-            startTransition(() => {
-              setArtistId('led-zeppelin');
-            });
-          }}
-        >
-          Led Zeppelin
-        </button>
-        <button
-          onClick={() => {
-            startTransition(() => {
-              setArtistId('pink-floyd');
-            });
-          }}
-        >
-          Pink Floyd
-        </button>
-      </div>
-      <div style={{ opacity: isPending ? 0.6 : 1 }}>
-        <Suspense fallback={<p>Loading...</p>}>
-          <Albums artistId={artistId} />
-        </Suspense>
-      </div>
+      <button onClick={() => setCount(count + 1)}>
+        Re-render
+      </button>
+      <p>Render count: {count}</p>
+      <Suspense fallback={<p>Loading...</p>}>
+        <Albums />
+      </Suspense>
     </>
   );
 }
 
-function Albums({ artistId }) {
-  const albums = use(fetchData('/' + artistId + '/albums'));
+function Albums() {
+  const albums = use(fetchData('/albums'));
   return (
     <ul>
       {albums.map(album => (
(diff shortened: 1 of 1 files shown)


src/content/reference/react/use.md
```diff
@@ -316,85 +1060,76 @@ But using `await` in a [Server Component](/reference/rsc/server-components) will
...
-    <ErrorBoundary fallback={<p>⚠️Something went wrong</p>}>
-      <Suspense fallback={<p>⌛Downloading message...</p>}>
-        <Message messagePromise={messagePromise} />
-      </Suspense>
-    </ErrorBoundary>
+    <>
+      <button onClick={handleRefresh} disabled={isPending}>
+        {isPending ? 'Refreshing...' : 'Refresh'}
+      </button>
+      <ErrorBoundary fallback={<p>⚠️Something went wrong</p>}>
```
MaxwellCohen: This ErrorBoundary does not seem to be working 

Image

src/content/reference/react/use.md
```diff
@@ -194,11 +221,734 @@ function Button({ show, children }) {
...
+    return await getAlbums('pink-floyd');
+  } else {
+    throw Error('Not implemented');
+  }
+}
+
+async function getAlbums(artistId) {
+  // Add a fake delay to make waiting noticeable.
+  await new Promise(resolve => {
+    setTimeout(resolve, 80);
```
aurorascharff: Is this delay a bit too short? The loading state is very brief in the sandbox. I keep trying to see the loading state in the other two tabs, but the promise resolves too quickly for the difference between hovering/not hovering to be clear.
Trying it in a forked sandbox, 1second seems better.

src/content/reference/react/use.md
```diff
@@ -316,85 +1060,76 @@ But using `await` in a [Server Component](/reference/rsc/server-components) will
...
+
+export default function App() {
+  const [albumsPromise, setAlbumsPromise] = useState(
+    () => fetchData('/the-beatles/albums')
+  );
+  const [isPending, startTransition] = useTransition();
+
+  function handleRefresh() {
+    startTransition(() => {
+      setAlbumsPromise(fetchData('/the-beatles/albums'));
```
aurorascharff: Adding some artificial delay in `fetchData` would make the `isPending` part of the code and the `<Suspense fallback={<p>Loading...</p>}>` more useful.

However, without async/await, any potential delay on the refresh will not actually disable the button. I added some delay in a forked sandbox, and had to add async await in front of `fetchData` for it to correctly disable the refresh button. But I guess that would make the error boundary useless?

As Maxwell said, seems like there's a few issues with this example.

src/content/reference/react/use.md
```diff
@@ -194,11 +221,734 @@ function Button({ show, children }) {
...
+<Pitfall>
+
+##### Promises passed to `use` must be cached {/*promises-must-cached*/}
+
+Promises created inside a component are recreated on every render. This causes React to show the Suspense fallback repeatedly and prevents content from appearing. Instead, pass a Promise from a cache, a Suspense-enabled framework, or a Server Component.
+
+```js
+function Albums() {
+  // 🔴 fetch creates a new Promise on every render.
+  const albums = use(fetch('/albums'));
```
MaxwellCohen, line 495: Adding more examples of ways to recreate promises between renders would be helpful, such as some ways you can create a new promise between renders 
1) calling a function that creates a new uncached promise. ie fetch  
2) adding a `.then`, `.catch`, `.finally` on a cached promise 
3) calling an async function 
4) calling functions from the Promise object i.e. Promise.resolve

src/content/reference/react/use.md
aurorascharff: This framing reads a bit like we should always prefer use() over awaiting in a Server Component. It might help to clarify that use() doesn’t inherently avoid blocking and that both approaches still depend on Suspense, and the reason we prefer use() is because we need to use a client component.

src/content/reference/react/use.md
```diff
@@ -194,11 +221,734 @@ function Button({ show, children }) {
...
+}
+```
+
+</Sandpack>
+
+<Solution />
+
+#### Fetching data with `useEffect` {/*fetching-data-with-useeffect*/}
+
+Without `use`, a common approach is to fetch data in an Effect and update state when the data arrives. This requires managing loading and error states manually, and the component renders empty on first paint before the Effect fires.
```
brenelz: would it make sense to mention this isn't recommended and you should use `use`. Maybe thats obvious

src/content/reference/react/use.md
```diff
@@ -194,11 +221,734 @@ function Button({ show, children }) {
...
+
+</Sandpack>
+
+<Solution />
+
+</Recipes>
+
+<Pitfall>
+
+##### Promises passed to `use` must be cached {/*promises-must-cached*/}
```
brenelz, line 463: I really like this section as I think its a common pitfall

src/content/reference/react/use.md
```diff
@@ -194,11 +221,734 @@ function Button({ show, children }) {
...
+
+let cache = new Map();
+
+export function fetchData(url) {
+  if (!cache.has(url)) {
+    const promise = getData(url);
+    // Set status fields so React can read the value
+    // synchronously if the Promise resolves before
+    // `use` is called (e.g. when preloading on hover).
+    promise.status = 'pending';
```
brenelz, line 921: interesting hadn't known it was preferred too set the status fields like this

src/content/reference/react/use.md
```diff
@@ -410,53 +1145,64 @@ root.render(
...
-      <Message messagePromise={messagePromise} />
-    </Suspense>
-  );
-}
+You are calling `use` inside a try-catch block. `use` throws internally to integrate with Suspense, so it cannot be wrapped in try-catch. Instead, wrap the component that calls `use` in an [Error Boundary](#displaying-an-error-with-an-error-boundary) to handle errors.
+
+```jsx
+function Albums({ albumsPromise }) {
+  try {
+    // ❌ Don't wrap `use` in try-catch
```
brenelz, line 1288: does this trip quite a few people up? Its interesting because its pretty obvious when you know `use` throws internally. Good callout forsure

brenelz:
Overall looks great. Nice work!

src/content/reference/react/use.md
```diff
@@ -18,53 +18,80 @@ const value = use(resource);
...
+[See more examples below.](#usage-promises)
 
-The `use` API returns the value that was read from the resource like the resolved value of a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) or [context](/learn/passing-data-deeply-with-context).
+#### Parameters {/*promise-parameters*/}
 
-#### Caveats {/*caveats*/}
+* `promise`: A [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) whose resolved value you want to read. The Promise must be [cached](#caching-promises-for-client-components) so that the same instance is reused across re-renders.
 
-* The `use` API must be called inside a Component or a Hook.
-* When fetching data in a [Server Component](/reference/rsc/server-components), prefer `async` and `await` over `use`. `async` and `await` pick up rendering from the point where `await` was invoked, whereas `use` re-renders the component after the data is resolved.
```
hernan-yadiel, line 51: I'm curious why this caveat was removed? Is it no longer the preferred way?

I know it could depend of the use case, but I did follow this suggestion as it helps shipping smaller initial bundles, instead of `use(dataPromise)` where the component where it was called  from is still part of the initial bundle.

src/content/reference/react/use.md
```diff
@@ -18,53 +18,80 @@ const value = use(resource);
...
+#### Returns {/*promise-returns*/}
+
+The resolved value of the Promise.
+
+#### Caveats {/*promise-caveats*/}
+
+* `use` must be called inside a Component or a Hook.
+* `use` cannot be called inside a try-catch block. Instead, wrap your component in an [Error Boundary](#displaying-an-error-with-an-error-boundary) to catch the error and display a fallback.
+* Promises passed to `use` must be cached so the same Promise instance is reused across re-renders. [See caching Promises below.](#caching-promises-for-client-components)
+* When passing a Promise from a Server Component to a Client Component, its resolved value must be serializable to pass between server and client. Data types like functions aren't serializable and cannot be the resolved value of such a Promise.
```
hernan-yadiel: the `'use client'` docs has a section with a list of serializable props, we could link to this page

```suggestion
* When passing a Promise from a Server Component to a Client Component, its resolved value must be [serializable](https://react.dev/reference/rsc/use-client#serializable-types) to pass between server and client. Data types like functions aren't serializable and cannot be the resolved value of such a Promise.
```

hernan-yadiel: > [...] to pass between server and client.

nit: for me this phrase feels redundant here, and it reads a bit weird when I first read it. It also subtly shifts from React terminology to infra terminology, which could be confusing since RSC don't necessarily imply a physical server. I would just drop that phrase:

```suggestion
* When passing a Promise from a Server Component to a Client Component, its resolved value must be serializable. Data types like functions aren't serializable and cannot be the resolved value of such a Promise.
```

src/content/reference/react/use.md
hernan-yadiel: Agree! To add to this, the "RSC blocks rendering" framing can read as if it blocks the entire app, when  that's not really the case. Maybe we can mention that if we move the async work to a child async Server Component wrapped in `<Suspense>`, it enables SSR streaming and avoids blocking.

I think this Deep Dive (or somewhere else in these use() docs) would benefit from mentioning the async SC + Suspense pattern and clarifying when each approach is preferable. Also, it's worth noting that neither the Suspense docs nor the RSC docs currently cover this "stremaing SSR" pattern either. Could be a good follow-up to add streaming examples there as well :)

hernan-yadiel: One thing I'm not fully clear on: beyond the bundle size tradeoff, are there other meaningful tradeoffs between the two approaches? Might be worth clarifying that in the Deep Dive as welll

MaxwellCohen: The big trade-off is when you will have information on the page (i.e., loading states). When a developer uses await for data loading, the rendering of that vDOM/DOM branch is blocked until the promise resolves. If this is set too high on the page, you will see a blank page until all data is ready (as with traditional PHP or Django webpages). When the developer uses the`use` hook, the user gets the loading section of the vDOM/DOM branch up to the first suspense boundary, and the rest of the page can render. Giving a much better user experience.

hernan-yadiel: @MaxwellCohen You're right in that it blocks rendering until the promise resolves, unless you wrap it in a `<Suspense />`. You can leverage this in such a way that it only blocks rendering for the UI that strictly depends on the data we're fetching, the rest is immediately sent to the browser.
See this snippet: https://gist.github.com/hernan-yadiel/2b9567df4536120b4c3931846978ada2

MaxwellCohen: Once #8300  is merged, it would be amazing to have a comparison example of both approaches.  My rule of thumb is to `await` for first paint critical path data (Metadata, above the fold data, key product data, etc.) and use `use` for secondary data (below the fold data, recommendations)

src/content/reference/react/use.md
```diff
@@ -18,53 +18,80 @@ const value = use(resource);
...
 
 ---
 
-## Usage {/*usage*/}
+## Usage (Context) {/*usage-context*/}
 
 ### Reading context with `use` {/*reading-context-with-use*/}
 
-When a [context](/learn/passing-data-deeply-with-context) is passed to `use`, it works similarly to [`useContext`](/reference/react/useContext). While `useContext` must be called at the top level of your component, `use` can be called inside conditionals like `if` and loops like `for`. `use` is preferred over `useContext` because it is more flexible.
+When a [context](/learn/passing-data-deeply-with-context) is passed to `use`, it works similarly to [`useContext`](/reference/react/useContext). While `useContext` must be called at the top level of your component, `use` can be called inside conditionals like `if` and loops like `for`.
```
stephan-noel, line 87: Now that there are 2 ways to read context, `use` and `useContext`, it seems that there is really no reason to use `useContext` anymore.  Not sure if it's worth clarifying any future plans or lack thereof.

src/content/reference/react/use.md
```diff
@@ -194,11 +221,734 @@ function Button({ show, children }) {
...
+<Sandpack>
+
+```js src/App.js active
+import { useState, useEffect } from 'react';
+import { fetchAlbums } from './data.js';
 
-```js [[1, 4, "App"], [2, 2, "Message"], [3, 7, "Suspense"], [4, 8, "messagePromise", 30], [4, 5, "messagePromise"]]
+export default function App() {
+  const [albums, setAlbums] = useState(null);
+  const [isLoading, setIsLoading] = useState(true);
```
stephan-noel, line 393: > This requires managing loading and error states manually, **and the component renders empty on first paint before the Effect fires**.

Emphasis mine.  The component doesn't render empty on first paint because the default state is set to true.  Just confused me a little bit when I saw it not rendering empty.

src/content/reference/react/use.md
```diff
@@ -18,53 +18,80 @@ const value = use(resource);
...
-* The `use` API must be called inside a Component or a Hook.
-* When fetching data in a [Server Component](/reference/rsc/server-components), prefer `async` and `await` over `use`. `async` and `await` pick up rendering from the point where `await` was invoked, whereas `use` re-renders the component after the data is resolved.
-* Prefer creating Promises in [Server Components](/reference/rsc/server-components) and passing them to [Client Components](/reference/rsc/use-client) over creating Promises in Client Components. Promises created in Client Components are recreated on every render. Promises passed from a Server Component to a Client Component are stable across re-renders. [See this example](#streaming-data-from-server-to-client).
+#### Returns {/*promise-returns*/}
+
+The resolved value of the Promise.
+
+#### Caveats {/*promise-caveats*/}
+
+* `use` must be called inside a Component or a Hook.
```
stephan-noel, line 76: **nit**: I know it's implied in the title and by virtue of being in the APIs section instead of the hooks section, but I wonder if it's worth explicitly calling out that `use` is **not** a hook despite it starting with the "use" naming convention.

Just by habit and this restriction being similar to one of the the "rules of hooks" might be confusing.

src/content/reference/react/use.md
```diff
@@ -194,11 +221,734 @@ function Button({ show, children }) {
...
+
+Promises passed to `use` in Client Components must be cached so the same Promise instance is reused across re-renders. If a new Promise is created directly in render, React will display the Suspense fallback on every re-render.
+
+```js
+// ✅ Cache the Promise so the same one is reused across renders
+let cache = new Map();
+
+export function fetchData(url) {
+  if (!cache.has(url)) {
+    cache.set(url, getData(url));
```
stephan-noel, line 531: I'm trying to reconcile this with the Rules of React.  My first thought is that this is "okay" because it falls under [lazy initialization](https://react.dev/reference/rules/components-and-hooks-must-be-pure#lazy-initialization).  Is that how you think about it?  

In the example given on that page, it says "Good: if it doesn't affect other components" and this will affect other components in some sense, but I guess it affects them in a way that doesn't really matter.

src/content/reference/react/use.md
stephan-noel: TIL.  I definitely believed it was the `await` itself that was holding up the initial SSR response and that the lack of `await` (regardless of `Suspense` usage) is what allows us to get around this.  Surprised to find that `use()` also blocks the initial SSR unless you use `Suspense`.

src/content/reference/react/use.md
```diff
@@ -194,11 +221,734 @@ function Button({ show, children }) {
...
+
+```css
+button { margin-bottom: 10px; }
+```
+
+</Sandpack>
+
+<Note>
+
+Frameworks that support Suspense typically provide their own caching and invalidation mechanisms. Building a custom cache like the one above is useful for understanding the pattern, but in practice you should use your framework's data fetching solution.
```
stephan-noel: If you wanted to use React without a framework and just use `use()` either via an underlying library or building your own, would that also be considered valid usage?  I think I recall talks of trying to replace `useEffect` in react-query with use.

It's just framework made me think meta-framework as in only NextJS, RR7, Tanstack Start, etc.

stephan-noel (commented):
Just left some questions small nits, looks great overall.

src/content/reference/react/use.md
MaxwellCohen: This nerd sniped me into creating a quick demo comparison of async and use. The bottom line is that Suspense + use/async is magical when you isolate slow data within Suspense boundaries.

demo link: https://streaming-test-three.vercel.app/comparison 
repo: https://github.com/maxwellCohen/streaming-test/

src/content/reference/react/use.md
MaxwellCohen: The way I think about `use` for promises is that it is like an useEffect, where you pass in a promise dependency. As the promise changes status, `use` will run various actions (throwing an error if not resolved) and rerender when resolved. Just like you cannot put a newly created function/object in an useEffect depency array, you cannot pass a new promise into `use`


I am very open to other ways of thinking about this.

stephan-noel: I think I'm understanding what you mean.  Here I'm more focusing on the fact that we do `cache.set` during render as well as `getData` and so it's a side-effect (impure).  Technically `setState` (conditionally and isolated to the component itself) during render and other one-time side-effects are also allowed despite being impure as mentioned in the linked docs.  

My first instinct was that originally the side-effect during render part of it was also a reason why we used to fetch in `useEffect`, and so I was just looking for some clarification around that (perhaps mistaken) conception and how to think about why these side effects during render are "okay".

src/content/reference/react/use.md
aurorascharff: fixed https://github.com/reactjs/react.dev/pull/8305/changes/ceab6dadbfd754ef3c83a02b8a3cd182d501ed65

src/content/reference/react/use.md
aurorascharff: fixed https://github.com/reactjs/react.dev/pull/8305/changes/ceab6dadbfd754ef3c83a02b8a3cd182d501ed65

src/content/reference/react/use.md
aurorascharff: fixed https://github.com/reactjs/react.dev/pull/8305/changes/ceab6dadbfd754ef3c83a02b8a3cd182d501ed65

src/content/reference/react/use.md
aurorascharff: Fixed! https://github.com/reactjs/react.dev/pull/8305/changes/ceab6dadbfd754ef3c83a02b8a3cd182d501ed65

src/content/reference/react/use.md
aurorascharff: I tried to adress all this here! Let me know if its more clear. https://github.com/reactjs/react.dev/pull/8305/changes/ceab6dadbfd754ef3c83a02b8a3cd182d501ed65

src/content/reference/react/use.md
aurorascharff: Added a link to useEffect docs here, is that helpful? https://github.com/reactjs/react.dev/pull/8305/changes/ceab6dadbfd754ef3c83a02b8a3cd182d501ed65

src/content/reference/react/use.md
aurorascharff: I think this should be adressed here! https://github.com/reactjs/react.dev/pull/8305/changes/ceab6dadbfd754ef3c83a02b8a3cd182d501ed65
have a look

src/content/reference/react/use.md
aurorascharff: Fixed! https://github.com/reactjs/react.dev/pull/8305/changes/ceab6dadbfd754ef3c83a02b8a3cd182d501ed65

src/content/reference/react/use.md
```diff
@@ -410,53 +1211,64 @@ root.render(
 ```
 </Sandpack>
 
-#### Providing an alternative value with `Promise.catch` {/*providing-an-alternative-value-with-promise-catch*/}
+---
 
-If you'd like to provide an alternative value when the Promise passed to `use` is rejected you can use the Promise's <CodeStep step={1}>[`catch`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/catch)</CodeStep> method.
+---
```
MaxwellCohen: This is an extra line that I do not think is needed

src/content/reference/react/use.md
```diff
@@ -194,11 +222,745 @@ function Button({ show, children }) {
...
+        <li key={album.id}>
+          {album.title} ({album.year})
+        </li>
+      ))}
+    </ul>
+  );
+}
+```
+
+The component that calls <CodeStep step={1}>`use`</CodeStep> must be wrapped in a [`Suspense`](/reference/react/Suspense) boundary. While the Promise is pending, the Suspense fallback is displayed. Once the Promise resolves, React reads the value with `use` and replaces the fallback with the rendered component.
```
MaxwellCohen: Do we need a deep dive into how `use(promise)` and `Suspense` work, or more details about how use under the hood throws an error, or a little more details like this: 

While the Promise is pending, `use` throws an internal error, and Suspense catches the error and displays a fallback. Once the Promise resolves, React replaces the fallback with the freshly rendered component, reading the resolved promise value with `use`.

src/content/reference/react/use.md
```diff
@@ -194,11 +222,745 @@ function Button({ show, children }) {
...
+
+<Solution />
+
+</Recipes>
+
+<Pitfall>
+
+##### Promises passed to `use` must be cached {/*promises-must-cached*/}
+
+Promises created inside a component are recreated on every render. This causes React to show the Suspense fallback repeatedly and prevents content from appearing. Instead, pass a Promise from a cache, a Suspense-enabled framework, or a Server Component.
```
MaxwellCohen: We might want to add more of an explanation that when components re-render, React will not preserve any previous state 

``` 
Promises created inside a component are recreated on every render because React does not preserve any state for renders that were suspended before they could mount for the first time. When the component has loaded, React will retry rendering the suspended tree from scratch. Recreating promises within the component causes React to repeatedly render the Suspense fallback, preventing content from appearing. Using internal component caching, such as useState factory functions, useMemo, or the React compiler, will cause promises to be recreated. Instead, pass a Promise from an external cache, a Suspense-enabled framework, or a Server Component.
```

aurorascharff: Added a quick link! Hows that? https://github.com/reactjs/react.dev/pull/8305/changes/92208c21624e244d5608a27586def0b6c8cbe6a6

src/content/reference/react/use.md
aurorascharff: Thanks for the suggestion! I think we can leave this one as-is. The internals (use throwing, Suspense catching) are already covered in the Troubleshooting section about try/catch, where it's more relevant to the reader's problem.

src/content/reference/react/use.md
aurorascharff: Thank you! fixed in https://github.com/reactjs/react.dev/pull/8305/changes/92208c21624e244d5608a27586def0b6c8cbe6a6

src/content/reference/react/use.md
rickhanlonii: It's covered in "Should I resolve a Promise in a Server or Client Component" below.

src/content/reference/react/use.md
```diff
@@ -194,11 +222,745 @@ function Button({ show, children }) {
...
+
+Common ways a Promise can be unintentionally recreated on every render:
+
+```js
+function Albums() {
+  // 🔴 `fetch` creates a new Promise on every render.
+  const albums = use(fetch('/albums'));
+
+  // 🔴 Calling an async function creates a new Promise on every render.
+  const albums = use(getAlbums());
```
rickhanlonii: @aurorascharff this is confusing because it's not clear how `getAlbums` and `fetchData` are different.

src/content/reference/react/use.md
rickhanlonii: Looks good for now, but note that this is subject to change in the future (no need to mention that though).

src/content/reference/react/use.md
aurorascharff: Hows this? https://github.com/reactjs/react.dev/pull/8305/commits/3cf1bf539b636a255e769341657b570ac9bda2cc

src/content/reference/react/use.md
```diff
@@ -222,6 +222,33 @@ function Button({ show, children }) {
 
 </Sandpack>
 
+### Reading a Promise from context {/*reading-a-promise-from-context*/}
```
MaxwellCohen, line 224: Great thinking on adding this section!

src/content/reference/react/use.md
```diff
@@ -18,46 +18,74 @@ const value = use(resource);
...
 
-* The `use` API must be called inside a Component or a Hook.
-* When fetching data in a [Server Component](/reference/rsc/server-components), prefer `async` and `await` over `use`. `async` and `await` pick up rendering from the point where `await` was invoked, whereas `use` re-renders the component after the data is resolved.
-* Prefer creating Promises in [Server Components](/reference/rsc/server-components) and passing them to [Client Components](/reference/rsc/use-client) over creating Promises in Client Components. Promises created in Client Components are recreated on every render. Promises passed from a Server Component to a Client Component are stable across re-renders. [See this example](#streaming-data-from-server-to-client).
+The resolved value of the Promise.
+
+#### Caveats {/*promise-caveats*/}
+
+* `use` must be called inside a Component or a Hook.
+* Despite its name, `use` is not a Hook. Unlike Hooks, it can be called inside loops and conditional statements like `if`.
```
stephan-noel: **nit**:  Small asymmetry now is that this is in the caveats section here, but for use(context) it's not in the caveats section, but rather the 2nd sentence in the main body. Really small, just mentioning to be exhaustive.

src/content/reference/react/use.md
```diff
@@ -18,46 +18,74 @@ const value = use(resource);
...
+* `use` cannot be called inside a try-catch block. Instead, wrap your component in an [Error Boundary](#displaying-an-error-with-an-error-boundary) to catch the error and display a fallback.
+* Promises passed to `use` must be cached so the same Promise instance is reused across re-renders. [See caching Promises below.](#caching-promises-for-client-components)
+* When passing a Promise from a Server Component to a Client Component, its resolved value must be [serializable](/reference/rsc/use-client#serializable-types). Data types like functions aren't serializable and cannot be the resolved value of such a Promise.
```
stephan-noel: > Data types like functions aren't serializable and cannot be the resolved value of such a Promise.

Doesn't seem fully accurate as client references can be passed.  My take would be to just leave this sentence out and let the linked page take care of that nuance.

src/content/reference/react/use.md
```diff
@@ -194,11 +222,798 @@ function Button({ show, children }) {
...
+  const user = use(userPromise);
+  return <h1>{user.name}</h1>;
+}
+```
+
+Wrap the components that read the Promise in [`<Suspense>`](/reference/react/Suspense) so only that subtree suspends while the Promise is pending. See [Usage (Promises)](#usage-promises) below for more on reading Promises with `use`.
+
+<Pitfall>
+
+Passing a Promise through context is convenient, but comes with trade-offs. Prefer passing the Promise as a [prop](#reading-a-promise-with-use) from a [Server Component](/reference/rsc/server-components) when possible.
```
stephan-noel: Should we also mention reading directly from the cache (ie, the fetchData function using the cache) here as well (and possibly link to that section)? Another approach would be to mention that passing promises via context is more for server-initiated fetches.

In terms of deciding when to use each approach, I'd think client-initiated fetches should probably prefer reading directly from the cache similar to the RSC guidance of fetching directly in the component where the data is used.  Mostly trying to prevent people seeing this approach and then unnecessarily using it with client-initiated fetches.

src/content/reference/react/use.md
```diff
@@ -194,11 +222,798 @@ function Button({ show, children }) {
...
+
+* **Double unwrap.** Reading the value requires two calls: `use(context)` to get the Promise, then `use(promise)` to get its resolved value. The context value itself is not awaited.
+* **Revalidation scope.** Because the Promise is created in the component that provides it, revalidating the data requires re-rendering that component. Hoisting data into a component high in the tree means everything below it re-renders to update the data.
+* **One Promise per context.** Each piece of data needs its own context. Reaching for more contexts to share more data is usually a sign that the data should be fetched closer to where it's used.
```
stephan-noel: The "One Promise per context" feels a bit more prescriptive than usual.  I wonder if linking to `/learn/passing-data-deeply-with-context#before-you-use-context` would be preferable here.

src/content/reference/react/use.md
```diff
@@ -194,11 +222,798 @@ function Button({ show, children }) {
...
+```
+
+Wrap the components that read the Promise in [`<Suspense>`](/reference/react/Suspense) so only that subtree suspends while the Promise is pending. See [Usage (Promises)](#usage-promises) below for more on reading Promises with `use`.
+
+<Pitfall>
+
+Passing a Promise through context is convenient, but comes with trade-offs. Prefer passing the Promise as a [prop](#reading-a-promise-with-use) from a [Server Component](/reference/rsc/server-components) when possible.
+
+* **Double unwrap.** Reading the value requires two calls: `use(context)` to get the Promise, then `use(promise)` to get its resolved value. The context value itself is not awaited.
+* **Revalidation scope.** Because the Promise is created in the component that provides it, revalidating the data requires re-rendering that component. Hoisting data into a component high in the tree means everything below it re-renders to update the data.
```
stephan-noel: I wonder if this is a problem with "[lift content up](https://overreacted.io/before-you-memo/)" and the React compiler.  Even if you pass a promise to a provider rendered in a server component, it only affects the consumers that would read from the associated context.

src/content/reference/react/use.md
```diff
@@ -194,11 +222,798 @@ function Button({ show, children }) {
...
+        <li key={album.id}>
+          {album.title} ({album.year})
+        </li>
+      ))}
+    </ul>
+  );
+}
+```
+
+The component that calls <CodeStep step={1}>`use`</CodeStep> must be wrapped in a [`Suspense`](/reference/react/Suspense) boundary. While the Promise is pending, the Suspense fallback is displayed. Once the Promise resolves, React reads the value with `use` and replaces the fallback with the rendered component.
```
stephan-noel: > The component that calls use must be wrapped in a [Suspense](http://localhost:3000/reference/react/Suspense) boundary

This uses `must` as if an error would occur, but I deleted Suspense and it still worked.  It wasn't pretty but there wasn't any error either.

src/content/reference/react/use.md
```diff
@@ -194,11 +222,798 @@ function Button({ show, children }) {
...
+[React doesn't preserve state for renders that suspended before mounting](/reference/react/Suspense#caveats). After each suspension, React retries rendering from scratch, so any Promise created during render is recreated.
+
+Common ways a Promise can be unintentionally recreated during render:
+
+```js
+function Albums() {
+  // 🔴 `fetch` creates a new Promise on every render.
+  const albums = use(fetch('/albums'));
+
+  // 🔴 Inline `async` function calls create a new Promise on every render.
```
stephan-noel: **nit**: I don't think it's the fact of the call being inline (whether inline is interpreted as the function is defined inline, or invoked inline as the argument to use).  It's that the function returns a promise which isn't cached.

Maybe "Uncached `async` function calls create a new Promise on every render."?

src/content/reference/react/use.md
```diff
@@ -292,109 +1107,147 @@ root.render(
...
+```js
+// Server Component
+export default function App() {
+  const messagePromise = fetchMessage();
+  return <Message messagePromise={messagePromise} />;
+}
+```
 
-### Dealing with rejected Promises {/*dealing-with-rejected-promises*/}
+In both cases, the component that reads the Promise suspends. Make sure there's a [`<Suspense>`](/reference/react/Suspense) boundary above it so React can display a fallback there and keep rendering the rest of the page. Without a boundary above the suspending component, the suspension bubbles up the tree and blocks the surrounding UI until the Promise resolves.
```
stephan-noel: Nice improvement 👍

src/content/reference/react/use.md
```diff
@@ -292,109 +1107,147 @@ root.render(
...
 
-In some cases a Promise passed to `use` could be rejected. You can handle rejected Promises by either:
+Passing the Promise down is useful when a Client Component needs the data and you want to reveal as much of the surrounding UI as possible while the Promise is pending. Because the Client Component is the component that suspends, you can place the `<Suspense>` boundary close to it and let the rest of the page render right away. See [Displaying a fallback while content is loading](/reference/react/Suspense#displaying-a-fallback-while-content-is-loading) for more on boundary placement.
 
-1. [Displaying an error to users with an Error Boundary.](#displaying-an-error-to-users-with-error-boundary)
-2. [Providing an alternative value with `Promise.catch`](#providing-an-alternative-value-with-promise-catch)
+When the surrounding UI is small or depends on the same data, prefer `await` on the server. If a Server Component above already awaits the data, pass the resolved value down instead of creating a new Promise to call `use`.
```
stephan-noel: One thing that's not clear to me here is _why_ you should prefer `await` on the server in this case.  Maybe just stating the advantage would be better?

One thing I thought that was different between these approaches was the time it takes for the promise to reach the client from the server.  If you suspend on the server, you'd start streaming the fallback, however if suspend on the client and your async call is fast enough, by the time your promise reaches the client with the rest of the response, it could be resolved already and so you may not show the user an unnecessary fallback.  I could be completely wrong here.

stephan-noel (commented):
Nice updates.  Overall looks good to me, just left a few more comments in the interest of being exhaustive.

src/content/reference/react/use.md
aurorascharff: Fixed! https://github.com/reactjs/react.dev/pull/8305/changes/c46d999a1247842c2d1a581dd586159be00a8df6 thanks

src/content/reference/react/use.md
aurorascharff: https://github.com/reactjs/react.dev/pull/8305/changes/c46d999a1247842c2d1a581dd586159be00a8df6
Fixed here, thanks!

src/content/reference/react/use.md
aurorascharff: Does this make it clearer? I'm still not sure about this section! https://github.com/reactjs/react.dev/pull/8305/changes/c46d999a1247842c2d1a581dd586159be00a8df6

src/content/reference/react/use.md
aurorascharff: Done! https://github.com/reactjs/react.dev/pull/8305/changes/c46d999a1247842c2d1a581dd586159be00a8df6 thanks

src/content/reference/react/use.md
aurorascharff: Nice catch, fixed! https://github.com/reactjs/react.dev/pull/8305/changes/c46d999a1247842c2d1a581dd586159be00a8df6

src/content/reference/react/use.md
stephan-noel: I think it's clearer and the added code block and comment with `// Will suspend until the data is available.` was a nice addition.  

Long wall of text incoming.

Thought it's clearer it's still a bit difficult to parse the recommendations.  I think I'm seeing it as there are 2 different things.  1) what the actual behavioral differences are and 2) what the implications of those behavioral differences are which may lead to picking one approach over the other.  I'm not fully sure of 1) myself so 2) isn't clear.  Ideally I'd hope the section could clarify both 1) and 2).

In terms me trying to make sense of the actual behavior differences, I see 4 options.  a) await in RSC without Suspense, b) use() in Client Component without Suspense c) await in an RSC with Suspense d), use() in a Client Component  with Suspense.  

> In both cases, wrap the component that reads the Promise in a Suspense boundary so React can display a fallback while the Promise is pending.

I think the code examples help clarify the behavior of  a) and b), but in the end we recommend to always choose either c) or d), but the difference between c) and d) isn't made clear.  If I'm using Suspense, what does it matter if we use `use()` in a Client Component or `await` in a Server Component?  Both unblock the initial server render (at least in NextJS).  I think there is some difference between c) and d) (both in NextJS and framework-independent), however subtle, and that's what would be useful to clarify IMO.

One structure that might help is: 
- first start clarifying that both use() and await() without Suspense , will suspend, like you already do, but do so without making recommendations yet (because later down you mention to always use Suspense).
- Then show code examples with Suspense and clarify the behavioral differences of use() vs await with Suspense. 
- Based on those behavioral differences, provide recommendations.

This is just how it would make sense to me, but I also get it would make a pretty long deep dive as well.

One final nitpick about this part is that " In both cases, wrap the component that reads the Promise in a Suspense boundary" can be interpreted literally and lead to loading states that are too fine-grained (spinners everywhere) instead of [revealing things together at once](https://react.dev/reference/react/Suspense#revealing-content-together-at-once).


> In general, prefer resolving the Promise with await in a Server Component. The Server Component suspends until the Promise resolves, and the Client Component receives the resolved value as a prop

I still can't tell why this would be recommended here.  It seems that the only advantage is you'd be passing the data instead of a promise with the data.  IMHO mentioning the behavior differences/tradeoffs may be enough if there's no clear winner. 


> Resolve the Promise with use in a Client Component for data that isn't important for the surrounding UI to render, like below-the-fold content, or content revealed by a hover or click (popovers, tooltips). The Server Component starts the Promise but doesn't await it, so it returns immediately.

What makes `use()` well-suited for this case?  For below the fold content you can use Suspense with await in an RSC as well.  Maybe you are referring to client-initiated fetches with `use()` here?  I do think server-initiated fetches can be too eager for things that are hidden that users may never interact with.  If it were that, maybe it would warrant a change in the title of the deep dive since it's not just where to resolve a Promise (and assume it's initiated on the server) but where to create the Promise as well.

> The Server Component starts the Promise but doesn't await it, so it returns immediately. The Client Component suspends when it calls use:
> 
> ...
> This way the Client Component suspends without blocking the rest of the page from rendering.

Without Suspense (the associated code example doesn't use Suspense) it does block the rest of the page.  This is what you mentioned in [this thread](https://github.com/reactjs/react.dev/pull/8305#discussion_r2814590686) and what I was testing in NextJS. Not sure if it is framework-specific though.

I can comment directly on the commit next time if you prefer, just though trying to keep a thread going with everything might be easier.

src/content/reference/react/use.md
```diff
@@ -222,6 +222,33 @@ function Button({ show, children }) {
 
 </Sandpack>
 
+### Reading a Promise from context {/*reading-a-promise-from-context*/}
+
+You can pass a Promise through context to share data without prop drilling. Pass the Promise as the context value, then call `use(context)` to read the Promise and `use(promise)` to read its resolved value:
```
rickhanlonii: > You can pass a Promise through context to share data without prop drilling. 

Hm, this makes it sound like you have to pass a promise to share data, but you can also just await the promise and put the response in context. When should you await vs no?

src/content/reference/react/use.md
```diff
@@ -222,6 +222,33 @@ function Button({ show, children }) {
...
+function Profile() {
+  const userPromise = use(UserContext);
+  const user = use(userPromise);
+  return <h1>{user.name}</h1>;
+}
+```
+
+Wrap the components that read the Promise in [`<Suspense>`](/reference/react/Suspense) so only that subtree suspends while the Promise is pending. See [Usage (Promises)](#usage-promises) below for more on reading Promises with `use`.
+
+<Pitfall>
```
rickhanlonii: These are more caveats than pitfalls.

src/content/reference/react/use.md
```diff
@@ -222,6 +222,33 @@ function Button({ show, children }) {
...
+
+Passing a Promise through context is convenient, but comes with trade-offs. Prefer passing the Promise as a [prop](#reading-a-promise-with-use) from a [Server Component](/reference/rsc/server-components) when possible.
+
+* **Double unwrap.** Reading the value requires two calls: `use(context)` to get the Promise, then `use(promise)` to get its resolved value. The context value itself is not awaited.
```
rickhanlonii: This seems framed as a problem (for example, maybe i assume this means reading multiple contexts are slow). But it's really just something to not forget to do.

src/content/reference/react/use.md
```diff
@@ -222,6 +222,33 @@ function Button({ show, children }) {
...
+* **Revalidation scope.** Because the Promise is created in the component that provides it, revalidating the data requires re-rendering that component. Hoisting data into a component high in the tree means everything below it re-renders to update the data.
```
rickhanlonii: Also React/RSCs don't really have "revalidation" mentioned anywhere else in the docs. Other frameworks might not even have it or call it "revalidation".

src/content/reference/react/use.md
rickhanlonii: +1

src/content/reference/react/use.md
aurorascharff: Updated: https://github.com/reactjs/react.dev/pull/8305/commits/e7be13d2c36ed066fcf93b242df2ec7f908140c9

src/content/reference/react/use.md
aurorascharff: Tried to adress it here: https://github.com/reactjs/react.dev/pull/8305/commits/e7be13d2c36ed066fcf93b242df2ec7f908140c9

src/content/reference/react/use.md
aurorascharff: Adressed: https://github.com/reactjs/react.dev/pull/8305/commits/e7be13d2c36ed066fcf93b242df2ec7f908140c9

src/content/reference/react/use.md
aurorascharff: Tried to address here: https://github.com/reactjs/react.dev/pull/8305/commits/e7be13d2c36ed066fcf93b242df2ec7f908140c9

src/content/reference/react/use.md
aurorascharff: Tried to adress it better: https://github.com/reactjs/react.dev/pull/8305/commits/e7be13d2c36ed066fcf93b242df2ec7f908140c9

src/content/reference/react/use.md
MaxwellCohen: [nit] use() in React 19 can only be used for promises or context. The "like" seems to imply you can use it with other data resources.    

So something like this would be more accurate:
```
`use` is a React API that lets you read the value of a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) resource or [context](/learn/passing-data-deeply-with-context) resource.
```


Ther e
https://github.com/react/react/blob/b91823e21434ac665450e67ddc6f816710255938/packages/react-server/src/ReactFizzHooks.js#L748-L763

aurorascharff: https://github.com/rickhanlonii/react.dev/commit/5d3e81af Adressed! good catch. Afaik in the future there might be more "resources", probably why it was framed like that!

src/content/reference/react/use.md
```diff
@@ -4,7 +4,7 @@ title: use
 
 <Intro>
 
-`use` is a React API that lets you read the value of a resource like a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) or [context](/learn/passing-data-deeply-with-context).
+`use` is a React API that lets you read the value of a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) or [context](/learn/passing-data-deeply-with-context) resource.
```
rickhanlonii:
```suggestion
`use` is a React API that lets you read the value of a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) or [context](/learn/passing-data-deeply-with-context).
```

src/content/reference/react/use.md
```diff
@@ -1144,9 +1138,11 @@ export function Message({ messagePromise }) {
 }
 ```
 
-If a Server Component above already awaits the data, pass the resolved value down as a prop instead of creating a new Promise to call `use`.
+Prefer `await` in a Server Component when possible, since it keeps the data fetching on the server. If a Server Component above already awaits the data, pass the resolved value down as a prop instead of creating a new Promise to call `use`.
+
+Use `use` in a Client Component to push the data read deeper in the tree, so more of the surrounding UI can render while the Promise is pending. A common case is interactive content like popovers and tooltips, where the data is needed only after a hover or click. Client Components can't `await`, so they rely on `use` to suspend on a Promise.
```
rickhanlonii:
```suggestion
You can also pass promise as a prop to a Client Component without awaiting it, and then read it with `use(promise)` to suspend deeper in the tree. This allows more of the surrounding UI to complete while the Promise is pending. A common case is interactive content like popovers and tooltips, where the data is needed only after a hover or click. Client Components can't `await`, so they rely on `use` to suspend on a Promise.
```

src/content/reference/react/use.md
rickhanlonii: No but it's one of those things where when you hit it, people get mad that it's not in the docs

src/content/reference/react/use.md
rickhanlonii: Yeah it's like a lazy init pattern. Technically, setting the initial state of a hook, or setting the initial value of a ref, or sync'ing with set state in render are all side effects. But they are all still pure. But ideally, you're prewarming the cache (like setting the promise in a route loader, or passing the promise from a server component).

src/content/reference/react/use.md
rickhanlonii: I don't really understand the framing of "valid usage" tbh. I'm not really here to be the React police and tell people what's valid and not valid to do. The note here is just saying if you're using a framework, favor their patterns. But you're free to create your own patterns or libraries, or to even ignore that advice and roll your own inside a framework.

src/content/reference/react/use.md
rickhanlonii: Looks good!

src/content/reference/react/use.md
rickhanlonii: I think this was a different thing, and maybe more specific to particular frameworks. 

The issue I was flagging with this pattern is that if you have nested routes like:

```
- users
 - page.js  <- context here
 - [id] 
   - page.js <- read here
```

then you would more often need to call `revalidate('/users')` instead of the more scoped `revalidate('/users/[id]')`, since the data the lower page.js is reading from is hoisted higher in the tree and only re-fetched when the higher page.js is refetched.

rickhanlonii: * **Refetching scope for RSCs**: When this pattern is used with Server Components, refetching the promise requires refetching the server component that sets the promise in context. Avoid setting the promise in context high in the tree, since this would require refetching large parts of the app unnecessarily.

aurorascharff (approved):

rickhanlonii:
Thanks all for the feedback, and thanks to @aurorascharff especially for getting this finished! 

Let's merge and iterate as needed.

PR #8308, opened 2026-02-16
Added missing prop to useActionState example

JakeSaterlay:
In the first example of `useActionState` the `<Total />` component deconstructs an `isPending` prop, however it is not passed to it in the example. This means that the `Updating...` text is never shown.

https://react.dev/reference/react/useActionState#adding-state-to-an-action

### DIFF
--- a/src/content/reference/react/useActionState.md
+++ b/src/content/reference/react/useActionState.md
@@ -177,7 +177,7 @@ export default function Checkout() {
         <button onClick={handleClick}>Add Ticket{isPending ? ' 🌀' : '  '}</button>
       </div>
       <hr />
-      <Total quantity={count} isPending={isPending} />
+      <Total quantity={count} />
     </div>
   );
 }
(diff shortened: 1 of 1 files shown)


rickhanlonii:
Yeah that's intentional, this came up in review. Since there's no optimistic state for the count, having the total be pending is weird. Instead it's just a button that's pending until the API responds, and the other examples after this add more features and a better UX. But you can remove the isPending from Total for this example.

JakeSaterlay:
> Yeah that's intentional, this came up in review. Since there's no optimistic state for the count, having the total be pending is weird. Instead it's just a button that's pending until the API responds, and the other examples after this add more features and a better UX. But you can remove the isPending from Total for this example.

@rickhanlonii gotcha, thanks for the explanation, have removed `isPending` from `<Total />`

rickhanlonii:
Thanks!

PR #8309, opened 2026-02-17
Expand FragmentInstance reference docs

jackpope:
[Preview](https://react-dev-git-fork-jackpope-fr-update-fbopensource.vercel.app/reference/react/Fragment)

  - Rewrites FragmentInstance reference with per-method API docs
  - Adds scrollIntoView reference and usage example
  - Documents reactFragments DOM handle and global IntersectionObserver caching pattern
  - Fixes bugs and improves all Fragment ref usage examples

### DIFF
--- a/src/content/reference/react/Fragment.md
+++ b/src/content/reference/react/Fragment.md
@@ -46,26 +46,15 @@ Wrap elements in `<Fragment>` to group them together in situations where you nee
 
 When you pass a `ref` to a Fragment, React provides a `FragmentInstance` object. It implements methods for interacting with the first-level DOM children wrapped by the Fragment.
 
-```js
-import { Fragment, useRef } from 'react';
+* [`addEventListener`](#addeventlistener) and [`removeEventListener`](#removeeventlistener) manage event listeners across all first-level DOM children.
+* [`dispatchEvent`](#dispatchevent) dispatches an event on the Fragment, which can bubble to the DOM parent.
+* [`focus`](#focus), [`focusLast`](#focuslast), and [`blur`](#blur) manage focus across all nested children depth-first.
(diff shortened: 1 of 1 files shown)


src/content/reference/react/Fragment.md
```diff
@@ -242,87 +459,216 @@ function PostBody({ body }) {
...
+
+</DeepDive>
+
+---
+
+### <CanaryBadge /> Managing focus across a group of elements {/*managing-focus-across-elements*/}
+
+Fragment `ref`s provide `focus`, `focusLast`, and `blur` methods that operate across all DOM nodes within the Fragment:
+
+```js {5-6}
```
eps1lon: Can we move this, and the other demos to Sandpacks?

src/content/reference/react/Fragment.md
```diff
@@ -30,41 +30,258 @@ Wrap elements in `<Fragment>` to group them together in situations where you nee
...
+
+---
+
+#### `observeUsing(observer)` {/*observeusing*/}
+
+Starts observing all first-level DOM children of the Fragment with the provided observer.
+
+```js
+const observer = new IntersectionObserver(callback, options);
+fragmentRef.current.observeUsing(observer);
```
slorber, line 174: Unrelated to this docs PR, but not sure where else to ask.

Is there a convenient way to get a list of refs from a fragment ref?
If not, this looks like something devs will find workarounds for 😄. cc @rickhanlonii might be worth reviewing this use case (and this whole post overall)

Image

https://barvian.me/react-exit-animations

jackpope: > Is there a convenient way to get a list of refs from a fragment ref?

The current design approach is to avoid coupling with specific children or parents. When the FragmentInstance handles coordination, things are more composable. You can refactor around Fragments without breaking dependent code.

Workarounds like this are possible but they put the responsibility on the user to handle edge cases. And we intentionally don't promote that pattern through a first-class API.

rickhanlonii (approved):

PR #8326, opened 2026-02-27
[wip] Add optimisticKey docs

rickhanlonii:
WIP - this is only the claude output, which is pretty good but I need to review / verify it.

### DIFF
--- a/src/content/reference/react/experimental_optimisticKey.md
+++ b/src/content/reference/react/experimental_optimisticKey.md
@@ -1,0 +1,258 @@
+---
+title: experimental_optimisticKey
+version: experimental
+---
+
+<Experimental>
+
+**This API is experimental and is not available in a stable version of React yet.**
+
+You can try it by upgrading React packages to the most recent experimental version:
+
+- `react@experimental`
+- `react-dom@experimental`
+- `eslint-plugin-react-hooks@experimental`
+
+Experimental versions of React may contain bugs. Don't use them in production.
+
+</Experimental>
+
+<Intro>
+
+`optimisticKey` lets you create temporary keys for new items.
+
+```js
+import { optimisticKey } from 'react';
+
+<Item key={optimisticKey} item={optimisticItem} />
+```
+
+</Intro>
+
+<InlineToc />
+
+---
+
+## Reference {/*reference*/}
+
+### `optimisticKey` {/*optimistickey*/}
+
+`optimisticKey` is a special value you can use as a `key` prop for elements that represent optimistic items -- items shown immediately before the server confirms them. When the transition completes and the real item renders in the same position, React transfers the component state from the optimistic element to the real one, instead of destroying and recreating it.
+
+```js
+import { useOptimistic, optimisticKey } from 'react';
+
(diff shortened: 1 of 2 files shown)


src/content/reference/react/experimental_optimisticKey.md
```diff
@@ -0,0 +1,258 @@
...
+  function handleAddTodo(text) {
+    startTransition(async () => {
+      addOptimisticTodo({ text });
+      await addTodoAction({ text });
+    });
+  }
+
+  return (
+    <div>
+      <button onClick={() => handleAddTodo('New todo')}>
```
MaxwellCohen, line 141: It would be nice to have a button to add a todo without using the optimisticKey, so it is easier to see the benefit of this api.

src/content/reference/react/experimental_optimisticKey.md
```diff
@@ -0,0 +1,258 @@
+---
+title: experimental_optimisticKey
+version: experimental
+---
+
+<Experimental>
+
+**This API is experimental and is not available in a stable version of React yet.**
```
MaxwellCohen, line 8: What feedback does the React team want regarding optimisticKey (and other experimental APIs)? 

Some APIs are mostly complete and just need testing for the next release, or for web standards to reach the point where they can be used (like <ViewTransitions>). Then there are some APIs that edge cases and additional use cases might need to be figured out about (like fragment references). 

For optimisticKey it seems like this is an API that needs to be tested before being released in React 19.3 or 20

samselikoff:
Awesome first pass! Two thoughts:

1. Animation is always an area where changing keys have bit me in the past, so might be great to include it as a usage example. For example if you have a list of messages (think iMessage) and you animate the new message from Composer up to the list, you give it a key. When the persisted message comes back it should match the optimistic key, otherwise you might get funky behavior when animating. It's a nice use case that's not necessarily about internal state, but more about the identity that key represents. (I'm more familiar with the problem when using libs like Motion but I imagine we could show a similar example with ViewTransitions, where you'd expect an `update` view transition to smoothly translate the message up, but because in the optimisticKey-less version you swap keys, you actually end up triggering an exit/enter instead.)
2. You should only be able to assign optimisticKey if you're in a transition, right? Do we error if you don't? If so maybe we can highlight this under caveats or troubleshooting.

src/content/reference/react/experimental_optimisticKey.md
```diff
@@ -0,0 +1,258 @@
...
+    "react-scripts": "latest"
+  }
+}
+```
+
+</Sandpack>
+
+<Note>
+
+Compare this to the [`useOptimistic`](/reference/react/useOptimistic#optimistically-adding-to-a-list) example, which uses `crypto.randomUUID()` as the key. That approach works for showing optimistic items, but loses component state when the real item replaces it. Use `optimisticKey` when your list items have internal state you want to preserve.
```
eps1lon, line 180: We should update the useOptimistic example to mention the experimental API and then only use optimisticKey once it lands in Canary.

Even now, I'd prefer if we have the comparison side-by-side like suggested in https://github.com/reactjs/react.dev/pull/8326/changes#r2866143024

PR #8338, opened 2026-03-06
Migrate docs site to App Router on Next canary

feedthejim:
## Summary
- migrate the docs site from the Pages Router to the App Router
- upgrade to `next@canary`, enable Cache Components, and rely on Turbopack as the default build path
- replace the old MDX pipeline, move special endpoints to route handlers, and make the Sandpack/RSC path Turbopack-safe

## What changed
- add App Router entrypoints for docs pages, errors, `llms.txt`, `rss.xml`, markdown API, `404`, `500`, `not-found`, and `error`
- add a shared content layer with `use cache` page-data helpers, cache tagging, and dev invalidation wiring for markdown/sidebar edits
- replace the old `compileMDX` / `prepareMDX` path with a server-first `generateMDX` pipeline and dedicated MDX plugins for TOC extraction, code metadata, max-width wrapping, and stable `data-mdx-name` annotations
- keep the existing client shell but refactor router integrations to `next/navigation` / `useLinkStatus`, plus explicit client islands for TOC, language list, challenges, examples, terminal copy UI, analytics, and error decoder behavior
- move root document concerns into the App Router layout, including theme bootstrapping, uwu mode, scroll restoration, Algolia preconnects, and analytics hooks
- replace build-time RSS generation with a route handler and move linting to flat-config ESLint
- remove the old Pages Router files once parity was restored

## Sandpack / tooling notes
- generate raw RSC sources during the prebuild step instead of relying on `raw-loader`
- swap the browser lint path to `eslint-linter-browserify` and use a browser-safe alias for `eslint-plugin-react-hooks`
- add a small dev websocket watcher so content edits trigger cache invalidation and `router.refresh()` under Turbopack

## Validation
- `yarn build`
- `yarn lint` (passes; existing `no-img-element` warnings remain)
- `yarn deadlinks`
- `yarn test:eslint-local-rules`


src/components/SafariScrollHandler.tsx
```diff
@@ -0,0 +1,21 @@
...
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import {useEffect} from 'react';
+
+export function SafariScrollHandler() {
+  useEffect(() => {
+    const isSafari = /^((?!chrome|android).)*safari/i.test(navigator.userAgent);
+    if (isSafari) {
+      history.scrollRestoration = 'auto';
```
gaearon, line 16: is this hack still needed and does it still work?

iirc this was to make Back gesture show previous content

src/utils/generateMDX.tsx
```diff
@@ -0,0 +1,150 @@
...
+      mdx,
+      path,
+      params,
+      mdxComponentNames: Object.keys(MDXComponents),
+      DISK_CACHE_BREAKER,
+      lockfile: fs.readFileSync(LOCKFILE_PATH, 'utf8'),
+    })
+  );
+
+  const cachedResult = await readFromCache(store, hash);
```
gaearon, line 58: do we need this caching stuff now? feels like it's just reimplementing with `'use cache'` gives us

MaxwellCohen:
Thank you for doing this upgrade!
I did some quick performance testing on the useActionState page (this pr build on my machine vs [react.dev](https://react.dev/reference/react/useActionState)) and saw a slight performance downgrade
LCP: .34s to .48s on local  vs .2s to .3s on prod
INP: 138ms vs 144ms  on local vs 112ms to 128ms on prod 

I am also seeing a lot more JS being loaded, 2.9mb on local vs 1.9mb on prod, while both versions use about 980kb of JS. This suggests we have some tree-shaking/bundling issues. With Turbopack, I am not sure how to fix it because the magic comments do not appear to be supported yet. I could be missing something.

localhost JS coverage report from Chrome
image

Production JS coverage report from Chrome
image

PR #8341, opened 2026-03-07
Add Copy Page button

gaearon:
I really like that Next.js docs have a button like this so I figured maybe we can add one too?

I usually just use this to feed something specific into Claude when it struggles with an API or a concept.

https://github.com/user-attachments/assets/53954aef-98d9-4144-8984-3aefdfe7202d

### DIFF
--- a/src/components/PageHeading.tsx
+++ b/src/components/PageHeading.tsx
@@ -14,8 +14,12 @@ import Tag from 'components/Tag';
 import {H1} from './MDX/Heading';
 import type {RouteTag, RouteItem} from './Layout/getRouteMeta';
 import * as React from 'react';
+import {useState, useEffect} from 'react';
+import {useRouter} from 'next/router';
 import {IconCanary} from './Icon/IconCanary';
 import {IconExperimental} from './Icon/IconExperimental';
+import {IconCopy} from './Icon/IconCopy';
+import {Button} from './Button';
 
 interface PageHeadingProps {
   title: string;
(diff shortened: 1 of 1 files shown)


src/components/PageHeading.tsx
```diff
@@ -27,6 +31,44 @@ interface PageHeadingProps {
...
+  const [copied, setCopied] = useState(false);
+
+  useEffect(() => {
+    if (!copied) return;
+    const timer = setTimeout(() => setCopied(false), 2000);
+    return () => clearTimeout(timer);
+  }, [copied]);
+
+  async function handleCopy() {
+    const cleanPath = asPath.split(/[?#]/)[0];
```
jeremenichelli, line 45: Doesn't `usePathname` already provides you what you are parsing here? 🤔 

Reference: https://nextjs.org/docs/app/api-reference/functions/use-pathname

gaearon: that's from App Router which we aren't yet using

MaxwellCohen:
I was going to suggest using `useTransition`  to manage the copied state notification. i.e. something like this

``` jsx
  const [copied, startTransition] = useTransition();

  async function handleCopy() {
    const cleanPath = asPath.split(/[?#]/)[0];
    try {
      const res = await fetch(cleanPath + '.md');
      if (!res.ok) return;
      const text = await res.text();
      await navigator.clipboard.writeText(text);
      startTransition(async() => {
        await new Promise((resolve) => setTimeout(resolve, 2000));
      });
    } catch {
      // Silently fail
    }
  }
```

If it is too late, feel free to disregard.

PR #8368, opened 2026-03-20
feat: migrate sandboxes to WebContainer

SamVerschueren:
This PR is still in draft but aims to replace Sandpack with WebContainer. With this PR we also switch from `create-react-app` to `Vite`.

There's still some things we need to resolve but wanted to build this in the open.

I used AI to generate [a sandpack compatible package](https://github.com/SamVerschueren/webcontainer-react) which uses WebContainer API under the hood. It's still in development but wanted to iterate fast.

Here's a list of things this PR does
- Move away from imgur and bringing all images into the repository. This is needed because `react.dev` now needs to [cross-origin isolated (because WebContainer requires it)](https://blog.stackblitz.com/posts/cross-browser-with-coop-coep/) and imgur doesn't serve the proper headers so all those images are broken.


SamVerschueren:
@rickhanlonii I can't see why Vercel fails to deploy this PR. Would it be possible for me to look at the logs myself? Otherwise it might be annoying to constantly ping someone that can look for me 🙈 .

PR #8405, opened 2026-04-12
Clarify form submission handling and examples

nikhilkumar5177:
Updated the form submission section to clarify how to prevent unwanted form resets and provided an example using controlled inputs with React state.


nikhilkumar5177:
Signed the CLA. Thanks!

PR #8406, opened 2026-04-12
Add guidance on preventing form reset in React

nikhilkumar5177:
## Summary
Adds documentation explaining how to prevent automatic form reset when using the `action` prop.

## Changes
- Added a new section: "Preventing unwanted form reset"
- Explained default reset behavior of uncontrolled inputs
- Included example using controlled inputs with `onSubmit`

## Why
Forms using the `action` prop reset uncontrolled fields after submission. This behavior may be unexpected and can cause confusion. This update clarifies how developers can preserve form state.

## Notes
- No breaking changes
- Follows existing documentation style


nikhilkumar5177:
Looks like the Vercel preview failed. Let me know if I should fix anything on my side.

PR #8439, opened 2026-05-16
replacing <img> with <Image/>

### LINKED ISSUE #4382: Add next/image component as MDX component and convert img tags to Image component
There are a lot of images inside markdown that are either `img` tags or markdown images. Need to use next/image component instead of img tags and markdown images.

Riya61018:
fixes  #4382

### DIFF
--- a/src/content/learn/index.md
+++ b/src/content/learn/index.md
@@ -456,7 +456,7 @@ function MyButton() {
 
 ```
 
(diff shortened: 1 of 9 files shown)


MaxwellCohen:
I have some concerns about this PR. The <Image /> component is Next.js (and some other meta frameworks) specific component not part of React, so React documentation should not be updated to tell using the <Image/>. There is a need to update the code of react.dev to leverage <Image /> component where it is not leveraged, so that this repo can be a good demo of best practices around Next and React.

Riya61018:
Should I close this pr?

MaxwellCohen:
I would, this PR does not fix the issue with the mdx component

PR #8465, opened 2026-06-01
docs: explain how to preserve form state after submission

### LINKED ISSUE #8397: [Suggestion]: improve <Form> documentation to explain how to keep form state after submission
### Summary

https://github.com/facebook/react/issues/29034 raises concerns about actions resetting forms. There are many use cases where developers would want to prevent the form from resetting. Since React's behavior is not expected to change, the docs should explain to users how to control form resets and preserve the form's previous state. 

### Page

https://react.dev/reference/react-dom/components/form

### Details

I do not know the best way to provide this information. Some suggestions include:
1. Add a deep dive on when to use onSubmit vs action props.  
2. Include an example of opting out of automatic resets using an example based on @eps1lon comment here:  https://github.com/facebook/react/issues/29034#issuecomment-2143595195. The issue has several other workarounds that may or may not be endorsed by the Rect team.  
3. Update the [Handling multiple submission types ](https://react.dev/reference/react-dom/components/form#handling-multiple-submission-types) example so that the "Save draft" button will keep the text of the form.

rylansedlacek:
Form page now explains how to preserve form state after submission when using forms.

In the issue it was mentioned that uncontrolled fields are reset after successful actions, but there is no explanation on how to opt out of that behavior.

This change adds a new section demonstrating how to use `onSubmit` together with `startTransition` to preserve form values while still using transitions.

Resolves #8397

### DIFF
--- a/src/content/reference/react-dom/components/form.md
+++ b/src/content/reference/react-dom/components/form.md
@@ -71,6 +71,51 @@ export default function Search() {
 
 </Sandpack>
 
+### Keep form values after submission {/*keep-form-values-after-submission*/}
+
+When a function is passed to the `action` prop, React resets all uncontrolled field elements in the form after the action succeeds.
+
+If you want to preserve the current form values after submission, use `onSubmit` instead of `action`. You can prevent the default form submission behavior and call your action manually inside a Transition.
+
+<Sandpack>
+
+```js src/App.js
+import { useTransition } from "react";
+
+export default function Search() {
+  const [isPending, startTransition] = useTransition();
+
+  async function search(formData) {
+    const query = formData.get("query");
+    alert(`You searched for '${query}'`);
+  }
+
+  function handleSubmit(event) {
+    event.preventDefault();
+
+    const formData = new FormData(event.target);
+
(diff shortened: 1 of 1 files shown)


src/content/reference/react-dom/components/form.md
```diff
@@ -71,6 +71,51 @@ export default function Search() {
...
+        Search
+      </button>
+    </form>
+  );
+}
+```
+
+</Sandpack>
+
+Using `onSubmit` preserves the browser's default behavior of leaving form values unchanged after submission while still allowing the submission logic to run in a Transition.
```
MaxwellCohen, line 116: This line is incorrect because HTML forms, by default, reset the page/form. We can see it at the `event.preventDefault();` in line 94

MaxwellCohen (commented):
Thank you very much for taking an attempt at the issue.

I noticed an inaccuracy with the help text. Forms by default reset, because they refresh the pages or go to another page on submission. 

Also, a missing example/section of using actions and useActionState to keep state like this:  https://codesandbox.io/p/sandbox/react-dev-forked-3ptf98

PR #8489, opened 2026-06-21
docs: clarify Client Components are pre-rendered during SSR

### LINKED ISSUE #7223: [Typo]: Client Components running on the server
### Summary

The ["use client"](https://react.dev/reference/rsc/use-client) docs state:

> During render, the framework will server-render the root component and continue through the [render tree](https://react.dev/learn/understanding-your-ui-as-a-tree#the-render-tree), opting-out of evaluating any code imported from client-marked code.
> 
> The server-rendered portion of the render tree is then sent to the client. The client, with its client code downloaded, then completes rendering the rest of the tree.

This implies that Client Components are skipped during SSR, but I don’t think that’s true.

### Page

https://react.dev/reference/rsc/use-client

### Details

My understanding is that Client Components will render on the server during SSR, and hydrate on the client. This is necessary so that the server-generated HTML is complete; if Client Components were truly skipped during SSR, our HTML would be full of holes that would only be filled in after hydration.

Another way to verify this is to add a `console.log()` to a Client Component, within a component function. The log will be printed in our server terminal, showing that the code is indeed evaluated on the server.

It’s possible I'm misunderstanding what is meant by "render" in this context, but it really makes it sound like the server skips any files that are marked with "use client". And unless this changed very recently, it is demonstrably untrue.

LuxologyGG:
## Summary

Closes #7223

The `use client` reference incorrectly implied that Client Components are skipped during server rendering. The page also conflated component type (logical Client/Server boundaries) with rendering (where and when code runs).

This PR separates those two concepts, defines Client Components as code sent to the client (optionally prerendered to HTML ahead of time), and defines Server Components as server-only code never sent to the client as executable code.

## Before / After

**Before:** "opting-out of evaluating any code imported from client-marked code" and Client Components defined as "rendered on the client."

**After:** Component type is determined by module boundaries; rendering is determined by the framework. Client Components are sent to the client as executable code and may be prerendered to HTML ahead of time.

## Verification

Documentation-only change.

### DIFF
--- a/src/content/reference/rsc/use-client.md
+++ b/src/content/reference/rsc/use-client.md
@@ -151,9 +151,9 @@ In the module dependency tree of this example app, the `'use client'` directive 
 `'use client'` segments the module dependency tree of the React Server Components app, marking `InspirationGenerator.js` and all of its dependencies as client-rendered.
 </Diagram>
 
-During render, the framework will server-render the root component and continue through the [render tree](/learn/understanding-your-ui-as-a-tree#the-render-tree), opting-out of evaluating any code imported from client-marked code.
+During render, the framework will server-render the root component and continue through the [render tree](/learn/understanding-your-ui-as-a-tree#the-render-tree). Client Components are also pre-rendered on the server to HTML. Their module source code is bundled for the client separately from Server Components.
 
-The server-rendered portion of the render tree is then sent to the client. The client, with its client code downloaded, then completes rendering the rest of the tree.
+The rendered HTML is sent to the client. The client downloads the client bundles and hydrates Client Components, attaching event handlers and other client-only behavior.
 
 <Diagram name="use_client_render_tree" height={250} width={500} alt="A tree graph where each node represents a component and its children as child components. The top-level node is labelled 'App' and it has two child components 'InspirationGenerator' and 'FancyText'. 'InspirationGenerator' has two child components, 'FancyText' and 'Copyright'. Both 'InspirationGenerator' and its child component 'FancyText' are marked to be client-rendered.">
 The render tree for the React Server Components app. `InspirationGenerator` and its child component `FancyText` are components exported from client-marked code and considered Client Components.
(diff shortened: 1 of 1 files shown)


gaearon:
I think we may need to do a more significant rewrite/rethinking of these pages. The model they're teaching is simplified and kind of incorrect, but we need to think about how to convey them well. The problem is that two worlds (Client and Server) have a very clean split; however, they don't have to map 1:1 to the physical "client" and "server". That's the key thing that we need to communicate somewhere.

LuxologyGG:
Thank you for the feedback.

You're right that the current Client/Server Component model on this page is simplified. Issue #7223 specifically flagged the sentence implying Client Components are skipped during SSR, which contradicts how frameworks actually pre-render them to HTML. This PR only targets that inaccuracy: it clarifies pre-rendering on the server and hydration on the client, without attempting to address the broader 1:1 mapping between logical Client/Server boundaries and physical client/server runtimes.

If a larger rewrite is the preferred path, I'm happy to close this PR and wait for direction on how you'd like that conceptual split communicated. Alternatively, if this targeted correction is still useful while a broader rethink is planned, I can keep it scoped to the SSR wording fix.

Let me know which direction you'd prefer.

gaearon:
Do you have any suggestions for how we could do this properly?

The current description isn't correct for HTML-generating frameworks on the first render (but it is correct for next render, or for frameworks that skip HTML generation, since it's up to the framework to decide whether to do it or not). So strictly speaking your change wouldn't be fully correct either.

It also makes the flow sound a lot more complicated when it's really more like two different things. One is the React Client and React Server trees. Another is how they get turned into HTML or other output formats.

I don't have a specific idea for how to solve this at the moment. But I have a hunch that a proper explanation would still look more "symmetrical".

LuxologyGG:
Thank you for the follow-up — you are right that a single sentence about pre-rendering does not capture the full picture.

Here is one approach that might keep the explanation symmetrical:

**Separate the two concepts explicitly:**

1. **Component type (logical):** Whether a component usage is a Client Component or a Server Component — determined by the module boundary (`'use client'` directive and its transitive dependencies). This is a property of the React tree, not the physical machine.

2. **Rendering phase (physical):** Where and when that tree node is evaluated and turned into output (HTML, RSC payload, etc.) — determined by the framework.

**A symmetrical framing for the definitions:**

- **Server Components** run only on the server. They can access server resources and are never sent to the client as executable code.
- **Client Components** are defined by the module boundary, but *where* they run depends on the framework and render pass:
  - On the **initial request**, frameworks that generate HTML (Next.js, etc.) typically evaluate Client Components on the server to produce HTML for the first paint.
  - On **subsequent client-side navigations**, those same Client Components run on the client.
  - Frameworks that skip HTML generation (e.g., client-only SPAs) evaluate them on the client from the start.

This avoids implying a 1:1 mapping between "Client Component" and "runs on the client," while still being accurate for both first render and later renders.

**For the render-tree walkthrough**, the description could be restructured around two parallel axes rather than a linear "server then client" flow — something like:

> During the initial render, the framework walks the React tree. Server Components are evaluated on the server. Client Component boundaries are also encountered on the server (for HTML-generating frameworks) or deferred to the client (for others). The framework decides per render pass where each boundary is evaluated.

I do not have a polished draft of the full page rewrite, but I wanted to share this framing in case it is useful for the broader rethink. Happy to close this PR and contribute to a larger rewrite if that is the preferred path — just let me know.

gaearon:
I'd be curious to look at a version that follow this approach, though can't promise merging. (I realize the current situation is also not good, but I think we want to nail this one instead of iterating over many versions.)

Another framing relevant to yours is that Client components are components sent as code to the client. They may also additionally be prerendered to HTML ahead of time (we don't have to say the word "server" here).

LuxologyGG:
Thank you for the additional framing. I pushed an updated version that follows the symmetrical approach:

1. **Component type** — determined by module boundaries (`'use client'` and transitive dependencies), not by where code physically runs.
2. **Rendering** — determined by the framework and render pass. Server Components are evaluated on the server. Client Components are sent to the client as executable code and may also be prerendered to HTML ahead of time.

The definitions now use your framing: Client Components are code sent to the client, optionally prerendered to HTML, without implying a 1:1 mapping between logical Client boundaries and physical client runtime.

Happy to iterate further if this direction is closer to what you had in mind.

PR #8494, opened 2026-06-23
Increase dark-mode button contrast

aurorascharff:
Bumps button contrast to match the treatment on [reactnative.dev](https://reactnative.dev/), in both dark and light mode.

In dark mode the primary button's text (`dark:text-secondary`, #404756) and the secondary button's border (#404756) were both low-contrast against the teal and dark backgrounds. This raises the primary text to near-black (`gray-95`, #16181D) and lifts the secondary border to #4E5769, the same value React Native uses (rgb(78, 86, 104)). Primary text contrast goes from about 4.6:1 to about 8.9:1.

In light mode the buttons already nearly matched React Native; the only gap was the secondary border, which was a touch lighter (#D9DBE3) than React Native's (#BCC1CD). This nudges it to #BCC1CD so the two are consistent.

Token changes:

- ButtonLink.tsx: primary `dark:text-secondary` to `dark:text-gray-95`
- tailwind.config.js: `secondary-button-stroke-dark` #404756 to #4E5769 (dark), and `secondary-button-stroke` #D9DBE3 to #BCC1CD (light)

## Screenshots

| Before | After | React Native (reference) |
| --- | --- | --- |
| CleanShot 2026-06-23 at 13 05 19@2x | 
image
 | CleanShot 2026-06-23 at 13 05 47@2x |

React Native reference: https://reactnative.dev/

## Notes

- Affects both dark and light mode (verified both).
- These are shared tokens, so every primary/secondary button is affected — flagging for design sign-off.

### DIFF
--- a/src/components/ButtonLink.tsx
+++ b/src/components/ButtonLink.tsx
@@ -33,7 +33,7 @@ function ButtonLink({
     className,
     'active:scale-[.98] transition-transform inline-flex font-bold items-center outline-none focus:outline-none focus-visible:outline focus-visible:outline-link focus:outline-offset-2 focus-visible:dark:focus:outline-link-dark leading-snug',
     {
-      'bg-link text-white dark:bg-brand-dark dark:text-secondary hover:bg-opacity-80':
+      'bg-link text-white dark:bg-brand-dark dark:text-gray-90 hover:bg-opacity-80':
         type === 'primary',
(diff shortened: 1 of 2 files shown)


rickhanlonii:
Maybe there's a better third option, but the darker text seems worse, like there's too much contrast. 

How about `'gray-80': '#343A46'`

Screenshot 2026-06-23 at 11 43 25 AM

rickhanlonii:
gray-90 also seems fine. i think the issue with 95 is that theres no cyan tint to it 

Screenshot 2026-06-23 at 11 45 15 AM

rickhanlonii (approved):

PR #8499, opened 2026-06-25
fix: update use.md pitfall to make conditional calling clearer

MaxwellCohen:
This PR helps clarify that use should be called based on Promise readiness/status, based on what @gaearon pointed out on [bluesky](https://bsky.app/profile/danabra.mov/post/3mp2ybeswjk2k) that  "nit: this seems misleading out of context. it shouldn't say always use unconditionally. it's just promise readiness that you shouldn't check. it can be conditional on other stuff (data, props, etc)"

CC @rickhanlonii @aurorascharff

### DIFF
--- a/src/content/reference/react/use.md
+++ b/src/content/reference/react/use.md
@@ -662,7 +662,8 @@ This cache pattern is the foundation for [re-fetching data](#re-fetching-data-in
 
 <Pitfall>
 
-Don't skip calling `use` based on whether a Promise is already settled. 
+Don't skip calling `use` based on whether a Promise is already settled.
+
 Unlike other hooks, `use` can be called inside conditions and loops — but it must always be called for the Promise itself. Never read `promise.status` or `promise.value` directly to bypass `use`; always pass the Promise to `use` and let React handle it. 
 
 
--- a/src/content/reference/react/use.md
+++ b/src/content/reference/react/use.md
@@ -664,7 +664,7 @@ This cache pattern is the foundation for [re-fetching data](#re-fetching-data-in
 
 Don't skip calling `use` based on whether a Promise is already settled.
 
-Unlike other hooks, `use` can be called inside conditions and loops — but it must always be called for the Promise itself. Never read `promise.status` or `promise.value` directly to bypass `use`; always pass the Promise to `use` and let React handle it. 
+Unlike other hooks, `use` can be called inside conditions and loops — but it must always be called for the Promise itself. Never read `promise.status` or `promise.value` directly to bypass `use`; always pass the Promise to `use` and let React handle it.
 
 
 ```js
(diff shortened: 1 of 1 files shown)


src/content/reference/react/use.md
```diff
@@ -662,18 +662,18 @@ This cache pattern is the foundation for [re-fetching data](#re-fetching-data-in
 
 <Pitfall>
 
-Don't conditionally call `use` based on whether a Promise is settled. Always call `use` unconditionally and let React handle reading the `status` field. This ensures React DevTools can show that the component may suspend on data.
+Don't skip calling `use` based on whether a Promise is already settled. Unlike other hooks, `use` can be called inside conditions and loops — but it must always be called for the Promise itself. Never read `promise.status` or `promise.value` directly to bypass `use`; always pass the Promise to `use` and let React handle it. Bypassing `use` this way can corrupt React's internal state tracking, causing stale renders, missed updates, or incorrect Suspense behavior. This also ensures React DevTools can show that the component may suspend on data.
```
rickhanlonii:
```suggestion
Don't skip calling `use` based on whether a Promise is already settled. 

Unlike other hooks, `use` can be called inside conditions and loops — but it must always be called for the Promise itself. Never read `promise.status` or `promise.value` directly to bypass `use`; always pass the Promise to `use` and let React handle it. 

Bypassing `use` this way can corrupt React's internal state tracking, and breaks React DevTools suspense data tracking:
```

src/content/reference/react/use.md
```diff
@@ -662,18 +662,22 @@ This cache pattern is the foundation for [re-fetching data](#re-fetching-data-in
...
+// 🔴 Don't bypass `use` by reading promise status directly
 if (promise.status === 'fulfilled') {
   return promise.value;
 }
 const value = use(promise);
 ```
 
 ```js
-// ✅ Always call `use` unconditionally
+// ✅ Always pass the promise to `use` and let React track promise readiness
```
rickhanlonii:
```suggestion
// ✅ Pass the promise to `use` and let React track the promise
```

src/content/reference/react/use.md
```diff
@@ -662,18 +662,22 @@ This cache pattern is the foundation for [re-fetching data](#re-fetching-data-in
 
 <Pitfall>
 
-Don't conditionally call `use` based on whether a Promise is settled. Always call `use` unconditionally and let React handle reading the `status` field. This ensures React DevTools can show that the component may suspend on data.
+Don't skip calling `use` based on whether a Promise is already settled. 
+
+Unlike other hooks, `use` can be called inside conditions and loops — but it must always be called for the Promise itself. Never read `promise.status` or `promise.value` directly to bypass `use`; always pass the Promise to `use` and let React handle it. 
+
+Bypassing `use` this way can corrupt React's internal state tracking, and breaks React DevTools suspense data tracking:
```
rickhanlonii:
```suggestion
```

src/content/reference/react/use.md
```diff
@@ -662,21 +662,26 @@ This cache pattern is the foundation for [re-fetching data](#re-fetching-data-in
 
 <Pitfall>
 
-Don't conditionally call `use` based on whether a Promise is settled. Always call `use` unconditionally and let React handle reading the `status` field. This ensures React DevTools can show that the component may suspend on data.
+Don't skip calling `use` based on whether a Promise is already settled. 
+
```
rickhanlonii, line 666:
```suggestion
```

Issue #8490, opened 2026-06-22
[Suggestion]: Suspense is missing a caveat about 300ms delay caused by FALLBACK_THROTTLE_MS


MaxwellCohen:
### Summary

There is a built-in delay of about 300 ms, controlled by FALLBACK_THROTTLE_MS, before Suspense content appears on the page if the promise is unresolved, but this is not documented.

### Page

https://react.dev/reference/react/Suspense

### Details

There have been several issues opened in the React repo regarding a 300ms delay in Suspense.  
- https://github.com/react/react/issues/36569
- https://github.com/react/react/issues/30408
- https://github.com/react/react/issues/31819

This is a design decision that should be called out so users can expect it as a caveat.

ShivaKoushik33:
****Hi, I’d like to work on this documentation change and open a PR to add the suggested caveat about the Suspense fallback throttle. Could a maintainer please assign this issue to me? Thanks.****

MaxwellCohen:
Resolved with #8505

PR #8505, opened 2026-06-27
[Suspense] Document what activates a boundary, with live examples

aurorascharff:
**Why:**

The Suspense reference said *only Suspense-enabled data sources* activate a boundary. That's outdated (many things activate Suspense now), and the term "Suspense-enabled framework" was used across several pages without ever being defined. The note also claimed framework-less data fetching was unsupported and undocumented, which the recent `use()` docs have since made untrue. The same stale note was copied into the SSR API pages too.

**What:**

- Replaces the old note with a linkable "What activates a Suspense boundary" section listing what actually activates one: `lazy`, `use`, stylesheets, fonts, streamed HTML, images, and deferred CPU work, with links to an example for each where possible.
- Adds a definition of "Suspense-enabled framework" and points framework-less fetching at `use()` with a cached Promise.
- Drops Next.js from the "Suspense-enabled frameworks" line, since it relies on Server Components and `use()` rather than a custom integration.
- Adds a side-by-side example contrasting a boundary that activates (reads a Promise with `use`) with one that doesn't (fetches in an Effect).
- Adds live examples for the demoable triggers: streaming SSR (re-runnable behind a button), a stylesheet with `precedence` blocking the boundary, fonts and images holding a `<ViewTransition>` reveal, and resetting boundaries with a `key` on navigation. The resource loads are real and cache-busted, so every run shows the waiting.
- Adds an "Animating from Suspense content" section sharing its demo with the `<ViewTransition>` page, and reorders the Usage sections so Canary-badged content comes last.
- Adds a caveat about React waiting at least 300ms before replacing a fallback with content, so boundaries that resolve close together reveal at once. Links to the React 19.2 post for the details.
- Slims the duplicated note on the SSR API pages (`renderToPipeableStream`, `renderToReadableStream`, `prerender`, `prerenderToNodeStream`) and `Activity` down to a short pointer to the Suspense section.


src/content/reference/react/Suspense.md
```diff
@@ -203,17 +203,25 @@ async function getAlbums() {
 
 </Sandpack>
 
-<Note>
+---
+
+### What activates a Suspense boundary {/*what-activates-a-suspense-boundary*/}
```
MaxwellCohen: Very excited to see these being called out and updated!

My understanding is that a "suspense enabled framework" is a framework that wraps these features that activates a suspense boundaries . If that is true could we specifically call it out what suspense enabled means?

aurorascharff: Maybe this? https://github.com/reactjs/react.dev/pull/8505/changes/021f94dc847fae905d80f8d9a85f843767ddb41e

src/content/reference/react/Suspense.md
```diff
@@ -203,17 +203,25 @@ async function getAlbums() {
...
+- <ExperimentalBadge /> Performing CPU-bound render work inside a `<Suspense>` boundary marked with the `defer` prop.
+
+<Note>
 
 Suspense **does not** detect when data is fetched inside an Effect or event handler.
 
-The exact way you would load data in the `Albums` component above depends on your framework. If you use a Suspense-enabled framework, you'll find the details in its data fetching documentation.
+The exact way you would load data in the `Albums` component above depends on your framework. If you use a framework with built-in Suspense support, you'll find the details in its data fetching documentation.
 
 Suspense-enabled data fetching without the use of an opinionated framework is not yet supported. The requirements for implementing a Suspense-enabled data source are unstable and undocumented. An official API for integrating data sources with Suspense will be released in a future version of React.
```
MaxwellCohen: This line is out of date because of yours and @rickhanlonii great work on the use() docs.

aurorascharff: Nice catch!!! Updating!

src/content/reference/react/Suspense.md
MaxwellCohen, line 38: I opened issue #8490 about the need for a caveat regarding React's batching of Suspense content, which has resulted in several issues in the React repo that can be solved with documentation. If we can add it to this pr, it would be great. I am thinking something like this:


```suggestion
- React waits at least ~300ms before revealing a `Suspense` fallback, and applies the same throttle across different boundaries. This reduces flicker when nested boundaries resolve in quick succession, which is common during [streaming server rendering](/reference/react-dom/server). Content that becomes ready within that window is held and revealed together once the window closes. A boundary that resolves before its fallback is committed—for example, synchronously, or within roughly 10ms—may skip the fallback entirely.

```

src/content/reference/react/Suspense.md
```diff
@@ -203,19 +204,27 @@ async function getAlbums() {
 
 </Sandpack>
 
-<Note>
+---
+
+### What activates a Suspense boundary {/*what-activates-a-suspense-boundary*/}
 
-**Only Suspense-enabled data sources will activate the Suspense component.** They include:
+A Suspense boundary displays its `fallback` while its content is loading, and it can also wait on other resources before revealing that content. The following activate a boundary:
```
eps1lon: "activate a boundary" is a bit ambiguous. We shouldn't necessarily say "shows a fallback" because that depends on the kind of update (Transition, sync etc.).


Wdyt about "The following block content from being revealed"?

src/content/reference/react/Suspense.md
```diff
@@ -203,19 +204,27 @@ async function getAlbums() {
...
-- Lazy-loading component code with [`lazy`](/reference/react/lazy)
-- Reading the value of a cached Promise with [`use`](/reference/react/use)
+- Lazy-loading component code with [`lazy`](/reference/react/lazy).
+- Reading a Promise with [`use`](/reference/react/use), including data streamed from [Server Components](/reference/rsc/server-components) and integrations from frameworks like [Relay](https://relay.dev/docs/guided-tour/rendering/loading-states/).
+- Loading a stylesheet rendered with [`<link rel="stylesheet">` and a `precedence` prop.](/reference/react-dom/components/link#special-rendering-behavior) React blocks the boundary until the stylesheet loads, up to a timeout.
+- Loading fonts. React blocks a streamed boundary until [`document.fonts.ready`](https://developer.mozilla.org/en-US/docs/Web/API/FontFaceSet/ready) resolves, up to a timeout. Fonts also block a [`<ViewTransition>`](/reference/react/ViewTransition) update.
+- Streaming a large boundary's HTML during server rendering. React reveals the content as the HTML arrives.
```
eps1lon: The other examples are probably self-explanatory. This one could do with an example. Examples for all would be nice with a side-by-side with a Suspense implementation that wouldn't block (e.g. with just some simple DOM operations).

src/content/reference/react/Suspense.md
```diff
@@ -203,19 +204,27 @@ async function getAlbums() {
...
+- Loading an image, where the `src` blocks the boundary until the image loads. This behavior is not enabled by default. When enabled, an `onLoad` handler opts an image out, and images in a [`<ViewTransition>`](/reference/react/ViewTransition) update opt in automatically.
+- <ExperimentalBadge /> Performing CPU-bound render work inside a `<Suspense>` boundary marked with the `defer` prop.
+
+<Note>
 
-Suspense **does not** detect when data is fetched inside an Effect or event handler.
+#### What is a Suspense-enabled framework? {/*what-is-a-suspense-enabled-framework*/}
 
-The exact way you would load data in the `Albums` component above depends on your framework. If you use a Suspense-enabled framework, you'll find the details in its data fetching documentation.
+A *Suspense-enabled framework* integrates data fetching with Suspense, so that reading data in a component activates the nearest boundary. The exact way you load data in the `Albums` component above depends on your framework, and you'll find the details in its data fetching documentation.
```
eps1lon: Is this still needed? Frameworks would just `use` as well. This section usually gets people asking how they could become a "Suspense-enabled framework" which is just `use`.

src/content/reference/react/Suspense.md
aurorascharff: @eps1lon can you verify this number?

src/content/reference/react/Suspense.md
aurorascharff: hows this? https://github.com/reactjs/react.dev/pull/8505/changes/d34d3ce2af06de54ca8a3df2e1544075603a9d1f

src/content/reference/react/Suspense.md
aurorascharff: hows this? https://github.com/reactjs/react.dev/pull/8505/changes/d34d3ce2af06de54ca8a3df2e1544075603a9d1f 

i still prefer the heading to have the simpler version maybe

src/content/reference/react/Suspense.md
aurorascharff: https://github.com/reactjs/react.dev/pull/8505/changes/d34d3ce2af06de54ca8a3df2e1544075603a9d1f added here but happy to revert

eps1lon: React waits at least 300ms before replacing the fallback with new content. That's different to blocking display of a fallback which only happens during updates in Transitions.

The idea is that filling in content more often leads to a "popcorn" effect. Imagine a 10x10 grid where each cell pops in 50ms intervals instead of doing 2-3 reveals including multiple cells. 300ms is just a heuristic that allows batching reveal of multiple boundaries.

src/content/reference/react/Suspense.md
eps1lon: Yeah except for the 300ms thing: https://github.com/reactjs/react.dev/pull/8505#discussion_r3499909004

src/content/reference/react/Suspense.md
aurorascharff: How abou[t with effect as an exa](https://github.com/reactjs/react.dev/pull/8505/changes/2e01ae74b3136c0c3cc384569b9eb1fbe050694d)mple? It's a common confusion

CleanShot 2026-06-30 at 11 46 52@2x

src/content/reference/react/Suspense.md
aurorascharff: Simplified https://github.com/reactjs/react.dev/pull/8505/changes/16fa091c826ad7bd5b7f4a50311681ac2cdd6085 and added link https://github.com/reactjs/react.dev/pull/8505/changes/54958a759677506b6b8d55f0291b5e95324d4f83

not quite sure i understood what you wanted though

MaxwellCohen: I like your caveat much better than my suggestion (great job!). But I feel like it implies that suspense will always show a fallback. 

My understanding is that 
0ms to ~10 ms - no fallback; suspense just renders children
11ms to ~300ms - batch at call; all render all together after 300ms
~310ms+  render in 300ms windows (not sure about this one)

This is figuring out an ideal rendering schedule; it is very impressive work by the React team, but is a little hard to explain. I hope I am not derailing the goals of this pr.

aurorascharff: Would need @eps1lon here to confirm and define what docs should include

src/content/reference/react-dom/static/prerender.md
eps1lon: This is also out of date. We probably have this note in a couple of places. Inlining it everywhere feels too heavy if you just want to learn about SSR APIs. A link to the Suspense docs is probably fine?

aurorascharff: Hows this?https://github.com/reactjs/react.dev/pull/8505/changes/6f6185d1ba68986e7c495007ef30af32e4f98a10

src/content/reference/react/Suspense.md
eps1lon: The throttling heuristic always blocks new boundary reveals counting from the last reveal. It has no impact on whether we display a fallback or not which is decided independently (always show fallback on mount, stay on content on update).

The very first boundary React will always reveal asap. If the boundary suspended, React shows a fallback immediately. When the boundary unsuspends, React will not reveal the boundaries content before 300ms have elapsed since the last reveal.

Throttling is surfaced in the performance tracks added by React.

Consider this example (using https://codesandbox.io/p/sandbox/quiet-http-yplr4n):

```js
function App({ p1, p2, p3 }) {
  return (
    <main>
      <h1>document</h1>
      <p>shell </p>
      <Suspense fallback="Loading lvl 1">
        <p>boundary lvl</p>
        <Suspender promise={p1} />
        <Suspense fallback="Loading lvl 2">
          <p>boundary lvl2</p>
          <Suspender promise={p2} />
          <Suspense fallback="Loading lvl 3">
            <p>boundary lvl3</p>
            <Suspender promise={p3} />
          </Suspense>
        </Suspense>
      </Suspense>
    </main>
  );
}
```

All Promises start at the same time as we can commit shell (which doesn't suspend and is committed instantly).

p1 resolves at 100ms
p2 resolves at 450ms
p3 resolves at 850ms

When p1 resolves we committed the root boundary (what we commonly refer to as "the shell") 100ms earlier. Since we don't want to reveal more often than every 300ms, we throttle for 200ms and reveal lvl 1 at 300ms.
When p2 resolves at 450ms, lvl1 was revealed 150ms earlier. Since we don't want to reveal more often than every 300ms, we throttle for 150ms and reveal lvl 2 at 600ms.
When p3 resolves at 850ms, lv2 was revealed 250ms earlier. Since we don't want to reveal more often than every 300ms, we throttle for 50ms and reveal lvl 2 at 900ms.


https://github.com/user-attachments/assets/8c920bef-7fdf-4915-9234-88b44426adbd



The batching can be observed when you let two boundaries resolve within the same 300ms window e.g.
p1 resolves at 100ms
p2 resolves at 350ms
p3 resolves at 450ms

p1 is throttled for 200ms
p2 is throttled for 250ms
p3 is throttled for 150ms

p2 and p3 are revealed together.

https://github.com/user-attachments/assets/ad630adc-c3a8-4dc0-aea4-9ce4cd558820


Without throttling all would commit independently within a short timeframe.

Throttling is global so it doesn't just apply to parent-child reveals. Initially, React throttled to 500ms which got reduced to 300ms in React 19. In React 19 we also applied throttling more broadly. The initial version of Suspense revealed boundaries immediately if there were no more nested boundaries.

src/content/reference/react/Suspense.md
```diff
@@ -203,22 +205,221 @@ async function getAlbums() {
...
 
-- Data fetching with Suspense-enabled frameworks like [Relay](https://relay.dev/docs/guided-tour/rendering/loading-states/) and [Next.js](https://nextjs.org/docs/app/building-your-application/routing/loading-ui-and-streaming#streaming-with-suspense)
-- Lazy-loading component code with [`lazy`](/reference/react/lazy)
-- Reading the value of a cached Promise with [`use`](/reference/react/use)
+- Lazy-loading component code with [`lazy`](/reference/react/lazy).
+- Reading a Promise with [`use`](/reference/react/use), including data streamed from [Server Components](/reference/rsc/server-components) and integrations from frameworks like [Relay](https://relay.dev/docs/guided-tour/rendering/loading-states/).
+- Loading a stylesheet rendered with [`<link rel="stylesheet">` and a `precedence` prop.](/reference/react-dom/components/link#special-rendering-behavior) React blocks the boundary until the stylesheet loads, up to a timeout.
+- Loading fonts. React blocks a streamed boundary until [`document.fonts.ready`](https://developer.mozilla.org/en-US/docs/Web/API/FontFaceSet/ready) resolves, up to a timeout. Fonts also block a [`<ViewTransition>`](/reference/react/ViewTransition) update.
```
eps1lon: Is that clear that this only applies to boundaries that are revealed by new SSR content arriving?

src/content/reference/react/Suspense.md
eps1lon: That's for `use()` vs fetching in an Effect though. I was thinking about a boundary with a lot of text. There are sandboxes in the `prerender` docs (or `resume`. Can't remember but we do have one example for PPR in the docs) that show how to build sandboxes showing streaming SSR.

aurorascharff: I see. I must have misunderstood. I'm wondering if the example you are referring to is too low level for the suspense page, but would fit nicely in the prerender or resume docs. I also think keeping this current example is quite a nice way to show this common misconception. I've added now a link to those docs here, we can add the example on that page? https://github.com/reactjs/react.dev/pull/8505/changes/201b5e35d0e01a9b970988031ac3e68385ada275 possibly in a follow-up PR

eps1lon: It's any SSR API with support for streaming. The whole point for me was to explain Suspense in a single place. It isn't low-level because it explains how React thinks about Suspense.

aurorascharff: oh, gotcha, will add!

src/content/reference/react/Suspense.md
aurorascharff: clarified: https://github.com/reactjs/react.dev/pull/8505/changes/d2df82afcd0fafc81afdc3b4ed1ea96a4bc8a06a

src/content/reference/react/Suspense.md
aurorascharff: https://github.com/reactjs/react.dev/pull/8505/changes/d2df82afcd0fafc81afdc3b4ed1ea96a4bc8a06a

So I simplified my example. Still open to removing the whole use affecting if we don't think it's necessary, and then I added your example.

src/content/reference/react/Suspense.md
```diff
@@ -203,22 +205,248 @@ async function getAlbums() {
...
-- Reading the value of a cached Promise with [`use`](/reference/react/use)
+- Lazy-loading component code with [`lazy`](/reference/react/lazy).
+- Reading a Promise with [`use`](/reference/react/use), including data streamed from [Server Components](/reference/rsc/server-components) and integrations from frameworks like [Relay](https://relay.dev/docs/guided-tour/rendering/loading-states/).
+- Loading a stylesheet rendered with [`<link rel="stylesheet">` and a `precedence` prop.](/reference/react-dom/components/link#special-rendering-behavior) React blocks the boundary until the stylesheet loads, up to a timeout.
+- Loading fonts. When a boundary is revealed by streamed SSR content, React waits for [`document.fonts.ready`](https://developer.mozilla.org/en-US/docs/Web/API/FontFaceSet/ready) before showing it, up to a timeout, so text doesn't flash with a fallback font. Fonts also block a [`<ViewTransition>`](/reference/react/ViewTransition) update.
+- Streaming a large boundary's HTML during server rendering. React [reveals the content as the HTML arrives.](/reference/react-dom/server/renderToReadableStream#streaming-more-content-as-it-loads)
```
eps1lon: The link is not related to the feature being explained. The feature here is React showing fallbacks just because the HTML is large. No explicit `use` or similar is needed. Just the fact that the content in a boundary is large, makes it suspend e.g. like in https://github.com/react/react/blob/6ca51ab1bd0d726eb4beb8a32f7ad162109710ed/packages/react-dom/src/__tests__/ReactDOMFizzServer-test.js#L9852-L9878

src/content/reference/react/Suspense.md
```diff
@@ -203,22 +205,248 @@ async function getAlbums() {
...
+- Loading an image, where the `src` blocks the boundary until the image loads. This behavior is not enabled by default. When enabled, an `onLoad` handler opts an image out, and images in a [`<ViewTransition>`](/reference/react/ViewTransition) update opt in automatically.
+- <ExperimentalBadge /> Performing CPU-bound render work inside a `<Suspense>` boundary marked with the `defer` prop.
+
+<Note>
 
-Suspense **does not** detect when data is fetched inside an Effect or event handler.
+#### What is a Suspense-enabled framework? {/*what-is-a-suspense-enabled-framework*/}
 
-The exact way you would load data in the `Albums` component above depends on your framework. If you use a Suspense-enabled framework, you'll find the details in its data fetching documentation.
+A *Suspense-enabled framework* reads data with [`use`](/reference/react/use) under the hood, so that reading data in a component activates the nearest boundary. The exact way you load data in the `Albums` component above depends on your framework, and you'll find the details in its data fetching documentation.
```
MaxwellCohen: This line seems out of place here: "The exact way you load data in the Albums component above depends on your framework, and you’ll find the details in its data fetching documentation."


```suggestion
A *Suspense-enabled framework* reads data with [`use`](/reference/react/use) under the hood, so that reading data in a component activates the nearest boundary. The exact way you load your data depends on your framework; you'll find the details in its documentation.
```

eps1lon (approved):

aurorascharff:
@eps1lon thanks for approval. However, some of the changes didnt make fully sense to me, still. Have a look at this: https://github.com/reactjs/react.dev/pull/8505/changes/2dfe10fc197b1c64761345fb7e555fcd2da74bf9

Specifically, what was weird was that we removed the specifics about Next.js as per request, but it's still valuable to know that Next.js uses use(). And that Relay, which was still mentioned, does not. And then I didn't have a connection between the suspense enabled framework anywhere, we just put the note there, so I wanted to have a way to link to it.

src/content/reference/react/Suspense.md
```diff
@@ -203,22 +205,277 @@ async function getAlbums() {
...
+- Data fetching through a [Suspense-enabled framework](#suspense-enabled-frameworks) like [Relay](https://relay.dev/docs/guided-tour/rendering/loading-states/), which integrates its own data source with Suspense.
+- Loading a stylesheet rendered with [`<link rel="stylesheet">` and a `precedence` prop.](/reference/react-dom/components/link#special-rendering-behavior) React blocks the boundary until the stylesheet loads, up to a timeout.
+- Loading fonts. When a boundary is revealed by streamed SSR content, React waits for [`document.fonts.ready`](https://developer.mozilla.org/en-US/docs/Web/API/FontFaceSet/ready) before showing it, up to a timeout, so text doesn't flash with a fallback font. Fonts also block a [`<ViewTransition>`](/reference/react/ViewTransition) update.
+- Streaming a large boundary's HTML during server rendering. React reveals the content as the HTML arrives.
+- Loading an image, where the `src` blocks the boundary until the image loads. This behavior is not enabled by default. When enabled, an `onLoad` handler opts an image out, and images in a [`<ViewTransition>`](/reference/react/ViewTransition) update opt in automatically.
```
gaearon: >This behavior is not enabled by default. When enabled,

this is confusing. when is it enabled, if it's not enabled by default?

src/content/reference/react/Suspense.md
```diff
@@ -203,22 +205,277 @@ async function getAlbums() {
...
+- <ExperimentalBadge /> Performing CPU-bound render work inside a `<Suspense>` boundary marked with the `defer` prop.
 
-Suspense **does not** detect when data is fetched inside an Effect or event handler.
+<Note>
+
+#### Suspense-enabled frameworks {/*suspense-enabled-frameworks*/}
 
-The exact way you would load data in the `Albums` component above depends on your framework. If you use a Suspense-enabled framework, you'll find the details in its data fetching documentation.
+A *Suspense-enabled framework* integrates its data fetching with Suspense, so that reading data in a component activates the nearest boundary. Some frameworks build on Server Components and [`use`](/reference/react/use), like Next.js, while others provide a custom integration, like Relay. The exact way you load your data depends on your framework, and you'll find the details in its documentation.
```
gaearon: I think this is slightly misleading / magic. There's really two ways: `use` and the legacy way which we're [supposed to deprecate](https://github.com/react/react/pull/34032) but never did. So canonically it's really just `use`. Maybe we don't explain it as a separate thing?

gaearon: Or we could leave the frameworks note but imply they just do it with `use`.

aurorascharff:
Restructured so all the info about suspense triggers is not below "usages", based on feedback from Dan this is not normally how we structure the docs

src/content/reference/react/Suspense.md
aurorascharff: @eps1lon do you know about this?

eps1lon: If inside ViewTransition, every images is opted in. Adding `onLoad` opts back out.

So priority is: default -> is inside `ViewTransition`? -> has `onLoad`?

by default it's currently disabled.

src/content/reference/react/Suspense.md
```diff
@@ -29,13 +29,39 @@ title: <Suspense>
...
+- Loading a stylesheet rendered with [`<link rel="stylesheet">` and a `precedence` prop.](/reference/react-dom/components/link#special-rendering-behavior) React blocks the boundary until the stylesheet loads, up to a timeout. [See an example below.](#waiting-for-a-stylesheet-to-load)
+- Waiting for a large boundary's HTML to arrive during streaming server rendering. Sending HTML takes time, so a boundary with enough content activates even when nothing in it suspends. React reveals the content as the HTML arrives.
+- <CanaryBadge /> Loading fonts. Suspense doesn't wait for fonts by default, but a [`<ViewTransition>`](/reference/react/ViewTransition) update waits for new fonts to load, up to a timeout, so text doesn't flash with a fallback font. [See an example below.](#waiting-for-a-font-to-load)
+- <CanaryBadge /> Loading images. Suspense doesn't wait for images by default, but during a [`<ViewTransition>`](/reference/react/ViewTransition) update, React blocks the boundary until the image loads, up to a timeout. Adding an `onLoad` handler opts a specific image out. [See an example below.](#waiting-for-an-image-to-load)
```
hernan-yadiel, line 50: Heyy! To me, this seems more like a `<ViewTransition />` feature than a `<Suspense>` one, since, as far as I can tell, you can take advantage of these features without using `<Suspense>`, right?

eps1lon: We're experimenting with making this behavior the default. Though probably prudent to still mention this on `<ViewTransition>`. I wouldn't move it off of `<Suspense>` even if this never becomes the default because you might not be aware that an image is part of a Transition. You just see it in devtools suspending a boundary.

src/content/reference/react/Suspense.md
```diff
@@ -29,13 +29,39 @@ title: <Suspense>
...
+- <ExperimentalBadge /> Performing CPU-bound render work inside a `<Suspense>` boundary marked with the `defer` prop.
```
hernan-yadiel, line 51:
```suggestion
- <ExperimentalBadge /> Performing CPU-bound render work inside a `<Suspense defer>` boundary.
```
or
```suggestion
- <ExperimentalBadge /> Performing CPU-bound render work inside a Suspense boundary with the `defer` prop set.
```

Also, the `defer` prop is missing in the Props section.

eps1lon (approved):

src/content/reference/react/Suspense.md
aurorascharff: https://github.com/reactjs/react.dev/pull/8526/changes

Added here, @eps1lon can you verify?

eps1lon: Isn't that just a wording change?

aurorascharff: ah, i added the prop to the props section

PR #8512, opened 2026-06-29
Update form documentation for forms default reset behavior

### LINKED ISSUE #8397: [Suggestion]: improve <Form> documentation to explain how to keep form state after submission

MaxwellCohen:
## Summary

Documents React’s default form-reset behavior when using the function `action` / `formAction` props, and explains how to preserve field values when you don’t want that reset. Applying @rickhanlonii's feedback from #7795 and building off of work @aurorascharff recentlly did. 

- Adds default reset behavior in Caveats, the `action` prop section, and a new Troubleshooting entry (“Why does my form reset when I use an action?”). React resets **uncontrolled** fields after a successful action, matching browser `<form action="...">` behavior (including before JS loads).
- **Adds “Preserve form values after submission”** with a Sandpack example: call `e.preventDefault()` in `onSubmit`, then run the action manually inside `useTransition` while keeping the `action` prop for progressive enhancement.
- **Adds a DeepDive** for finer control: `requestFormReset` from `react-dom`, and returning submitted `FormData` from server actions to restore values via `defaultValue` (with `useActionState`).
- **Updates the multiple-submission-types example** to a draft/publish scenario that shows how controlled state and `key`/`defaultValue` keep textarea content after “Save draft.”

Closes #8397 

Alternative to #7795  and #8465

### DIFF
--- a/src/content/reference/react-dom/components/form.md
+++ b/src/content/reference/react-dom/components/form.md
@@ -95,7 +95,7 @@ When you pass a function to `action`, React:
 * Propagates any errors to the nearest error boundary.
 * Resets the form's uncontrolled fields when the function succeeds. To keep their values, see [Preserving form values after submission](#preserve-form-values-after-submission).
 
-Because the action runs in a Transition, you can also use [`useActionState`](/reference/react/useActionState) to manage form state and [`useOptimistic`](/reference/react/useOptimistic) for optimistic UI.
+Because the Action runs in a Transition, you can also use [`useFormStatus`](/reference/react-dom/hooks/useFormStatus) for submission status, [`useActionState`](/reference/react/useActionState) to manage form state, and [`useOptimistic`](/reference/react/useOptimistic) for optimistic UI.
 
 <Sandpack>
 
--- a/src/content/reference/react-dom/components/form.md
+++ b/src/content/reference/react-dom/components/form.md
@@ -202,7 +202,7 @@ Because you call the action manually from `onSubmit`, [`useFormStatus`](/referen
 
 <DeepDive>
 
-#### Resetting only some fields, or resetting on the server {/*resetting-only-some-fields*/}
+#### Reset only some fields, or reset on the server {/*resetting-only-some-fields*/}
 
 The `onSubmit` approach keeps every uncontrolled field. When you need finer control, two other patterns are available:
 
@@ -276,7 +276,7 @@ export async function submitForm(query) {
 
 To learn more about the `useFormStatus` Hook see the [reference documentation](/reference/react-dom/hooks/useFormStatus).
 
-### Optimistically updating form data {/*optimistically-updating-form-data*/}
+### Optimistically update form data {/*optimistically-updating-form-data*/}
 The `useOptimistic` Hook provides a way to optimistically update the user interface before a background operation, like a network request, completes. In the context of forms, this technique helps to make apps feel more responsive. When a user submits a form, instead of waiting for the server's response to reflect the changes, the interface is immediately updated with the expected outcome.
 
 For example, when a user types a message into the form and hits the "Send" button, the `useOptimistic` Hook allows the message to immediately appear in the list with a "Sending..." label, even before the message is actually sent to a server. This "optimistic" approach gives the impression of speed and responsiveness. The form then attempts to truly send the message in the background. Once the server confirms the message has been received, the "Sending..." label is removed.
@@ -346,7 +346,7 @@ export async function deliverMessage(message) {
 [//]: # 'Uncomment the next line, and delete this line after the `useOptimistic` reference documentation page is published'
 [//]: # 'To learn more about the `useOptimistic` Hook see the [reference documentation](/reference/react/useOptimistic).'
 
-### Handling form submission errors {/*handling-form-submission-errors*/}
+### Handle form submission errors {/*handling-form-submission-errors*/}
 
 In some cases the function called by a `<form>`'s `action` prop throws an error. You can handle these errors by wrapping `<form>` in an Error Boundary. If the function called by a `<form>`'s `action` prop throws an error, the fallback for the error boundary will be displayed.
 
@@ -446,7 +446,7 @@ export async function signUpNewUser(newEmail) {
 
 Learn more about updating state from a form action with the [`useActionState`](/reference/react/useActionState) docs
 
-### Handling multiple submission types {/*handling-multiple-submission-types*/}
+### Handle multiple submission types {/*handling-multiple-submission-types*/}
 
 A form can have more than one submit button, each running a different action. Set the `formAction` prop on a `<button>` to override the `<form>`'s `action` when that button submits the form.
 
--- a/src/content/reference/react-dom/components/form.md
+++ b/src/content/reference/react-dom/components/form.md
@@ -472,7 +472,6 @@ export default function ArticleForm() {
     const content = payload.data.get('content');
     switch (payload.type) {
       case 'save':
-
         alert(`Your draft of '${content}' was saved!`);
         // Keep the submitted content as the current draft
         return payload.data;
@@ -481,23 +480,22 @@ export default function ArticleForm() {
         // reset the form
         return new FormData();
     }
-  });
+  }, new FormData());
 
   function publish(formData) {
     dispatchFormState({
       type: 'publish',
       data: formData
-    })
+    });
   }
 
   function save(formData) {
     dispatchFormState({
       type: 'save',
       data: formData
-    })
+    });
   }
 
-  // const savedContent = draft.get('content') || '';
   return (
     <form action={publish}>
       <textarea
--- a/src/content/reference/react-dom/components/form.md
+++ b/src/content/reference/react-dom/components/form.md
@@ -43,13 +43,13 @@ To create interactive controls for submitting information, render the [built-in 
 #### Caveats {/*caveats*/}
 
 * When a function is passed to `action` or `formAction` the HTTP method will be POST regardless of the value of the `method` prop.
-* When a function is passed to `action` or `formAction`, React resets all [uncontrolled](/reference/react-dom/components/input#reading-the-input-values-when-submitting-a-form) field elements after the action succeeds. See [Preserve form values after submission](#preserve-form-values-after-submission).
+* When a function is passed to `action` or `formAction`, React resets all [uncontrolled](/reference/react-dom/components/input#reading-the-input-values-when-submitting-a-form) field elements after the action succeeds. See [Preserving form values after submission](#preserve-form-values-after-submission).
 
 ---
 
 ## Usage {/*usage*/}
 
-### Handle form submission with an event handler {/*handle-form-submission-with-an-event-handler*/}
+### Handling form submission with an event handler {/*handle-form-submission-with-an-event-handler*/}
 
 Pass a function to the `onSubmit` event handler to run code when the form is submitted. By default, the browser sends the form data to the current URL and refreshes the page. Call [`e.preventDefault()`](https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault) in your handler to override that behavior. Read the submitted data with [`new FormData(e.target)`](https://developer.mozilla.org/en-US/docs/Web/API/FormData).
 
@@ -82,9 +82,9 @@ Reading form data with `onSubmit` works in every version of React and gives you 
 
 </Note>
 
-### Handle form submission with an action prop {/*handle-form-submission-with-an-action-prop*/}
+### Handling form submission with an action prop {/*handle-form-submission-with-an-action-prop*/}
 
-Pass a function to the `action` prop of form to run the function when the form is submitted. [`formData`](https://developer.mozilla.org/en-US/docs/Web/API/FormData) will be passed to the function as an argument so you can access the data submitted by the form. This differs from the conventional [HTML action](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#action), which only accepts URLs. Unlike `onSubmit`, an `action` runs in a [Transition](/reference/react/useTransition) and calling `e.preventDefault()` isn't needed. After the `action` function succeeds, React resets all [uncontrolled](/reference/react-dom/components/input#reading-the-input-values-when-submitting-a-form) field elements in the form. To keep their values, see [Preserve form values after submission](#preserve-form-values-after-submission).
+Pass a function to the `action` prop of form to run the function when the form is submitted. [`formData`](https://developer.mozilla.org/en-US/docs/Web/API/FormData) will be passed to the function as an argument so you can access the data submitted by the form. This differs from the conventional [HTML action](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#action), which only accepts URLs. Unlike `onSubmit`, an `action` runs in a [Transition](/reference/react/useTransition) and calling `e.preventDefault()` isn't needed. After the `action` function succeeds, React resets all [uncontrolled](/reference/react-dom/components/input#reading-the-input-values-when-submitting-a-form) field elements in the form. To keep their values, see [Preserving form values after submission](#preserve-form-values-after-submission).
 
 <Sandpack>
 
@@ -106,7 +106,7 @@ export default function Search() {
 </Sandpack>
 
 
-### Handle form submission with a Server Function {/*handle-form-submission-with-a-server-function*/}
+### Handling form submission with a Server Function {/*handle-form-submission-with-a-server-function*/}
 
 Render a `<form>` with an input and submit button. Pass a Server Function (a function marked with [`'use server'`](/reference/rsc/use-server)) to the `action` prop of form to run the function when the form is submitted.
 
@@ -153,7 +153,8 @@ function AddToCart({productId}) {
 ```
 
 When `<form>` is rendered by a [Server Component](/reference/rsc/use-client), and a [Server Function](/reference/rsc/server-functions) is passed to the `<form>`'s `action` prop, the form is [progressively enhanced](https://developer.mozilla.org/en-US/docs/Glossary/Progressive_Enhancement).
-### Preserve form values after submission {/*preserve-form-values-after-submission*/}
+
+### Preserving form values after submission {/*preserve-form-values-after-submission*/}
 
 The browser clears a form's input state on submit. A URL `action` follows this same behavior. React does the same when `action` is a function, so your form works consistently before and after JavaScript loads.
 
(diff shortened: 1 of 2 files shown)


aurorascharff:
Going to dig a bit into the current behavior of this, so might take some time to review! Thank you!

src/content/reference/react-dom/components/form.md
```diff
@@ -86,9 +82,22 @@ Reading form data with `onSubmit` works in every version of React and gives you
...
 
-Pass a function to the `action` prop of form to run the function when the form is submitted. [`formData`](https://developer.mozilla.org/en-US/docs/Web/API/FormData) will be passed to the function as an argument so you can access the data submitted by the form. This differs from the conventional [HTML action](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#action), which only accepts URLs. Unlike `onSubmit`, an `action` runs in a [Transition](/reference/react/useTransition) and calling `e.preventDefault()` isn't needed. After the `action` function succeeds, all uncontrolled field elements in the form are reset.
+This extends the [HTML `action` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#action), which only accepts a URL, to also accept a function. Because the form stays a native HTML form, it can be submitted before JavaScript loads; with a [Server Function](/reference/rsc/server-functions), it works without JavaScript enabled.
```
aurorascharff: `ecause the form stays a native HTML form, it can be submitted before JavaScript loads; with a [Server Function](/reference/rsc/server-functions), it works without JavaScript enabled`

this part is a bit unclear and hard to parse!

src/content/reference/react-dom/components/form.md
```diff
@@ -86,9 +82,22 @@ Reading form data with `onSubmit` works in every version of React and gives you
...
+
+When you pass a function to `action`, React:
+
+* Runs the submission function in a [Transition](/reference/react/useTransition).
+* Tracks pending state so child components can read it with [`useFormStatus`](/reference/react-dom/hooks/useFormStatus).
+* Sends thrown errors to the nearest error boundary.
```
aurorascharff: Not sure if "sends" is the right terminology for this 🤔

src/content/reference/react-dom/components/form.md
```diff
@@ -86,9 +82,22 @@ Reading form data with `onSubmit` works in every version of React and gives you
...
+* Works with [`useActionState`](/reference/react/useActionState) and [`useOptimistic`](/reference/react/useOptimistic) for form state and optimistic UI.
```
aurorascharff: This bullet seems out of place in its framing, since the intro to the list is `When you pass a function to `action`, React:` but here "works" is not answering that question!

src/content/reference/react-dom/components/form.md
```diff
@@ -157,7 +167,87 @@ function AddToCart({productId}) {
 
 When `<form>` is rendered by a [Server Component](/reference/rsc/use-client), and a [Server Function](/reference/rsc/server-functions) is passed to the `<form>`'s `action` prop, the form is [progressively enhanced](https://developer.mozilla.org/en-US/docs/Glossary/Progressive_Enhancement).
 
-### Display a pending state during form submission {/*display-a-pending-state-during-form-submission*/}
+### Preserving form values after submission {/*preserve-form-values-after-submission*/}
+
+The browser clears a form's input state on submit. A URL `action` follows this same behavior. React does the same when `action` is a function, so your form works consistently before and after JavaScript loads.
```
aurorascharff: this paragraph reads a little strange, and the behavior was explained better above!

src/content/reference/react-dom/components/form.md
```diff
@@ -157,7 +167,87 @@ function AddToCart({productId}) {
...
+
+```js src/api.js hidden
+export async function submitForm(formData) {
+  await new Promise((res) => setTimeout(res, 1000));
+}
+```
+
+</Sandpack>
+
+Because you call the action manually from `onSubmit`, [`useFormStatus`](/reference/react-dom/hooks/useFormStatus) won't report its pending state. Read `isPending` from the same [`useTransition`](/reference/react/useTransition) call instead.
```
aurorascharff: Hm, i thought useFormStatus worked regardless, also with onSubmit?

src/content/reference/react-dom/components/form.md
```diff
@@ -372,28 +460,45 @@ Learn more about updating state from a form action with the [`useActionState`](/
...
 
-When a user taps a specific button, the form is submitted, and a corresponding action, defined by that button's attributes and action, is executed. For instance, a form might submit an article for review by default but have a separate button with `formAction` set to save the article as a draft.
+When a button without `formAction` submits the form, React calls the form's `action`. When a button with `formAction` submits the form, React calls that button's action instead. For example, the form below publishes an article by default, but its **Save draft** button stores the current content without publishing it.
+
+In this example the draft is held in state, so the saved content stays in the textarea after you submit it. In a real app you would persist the draft on the server. Pass a [Server Function](/reference/rsc/server-functions) (a function marked with [`'use server'`](/reference/rsc/use-server)) to `formAction` to save the draft from the server, optionally combined with [`useActionState`](/reference/react/useActionState) to track its pending state and result.
```
aurorascharff: (a function marked with [`'use server'`](/reference/rsc/use-server))

is this a common pattern in the docs?

src/content/reference/react-dom/components/form.md
```diff
@@ -157,7 +167,87 @@ function AddToCart({productId}) {
...
+* **Reset to server-provided values.** When an action validates input on the server, return the submitted `FormData` and pass it to each field's `defaultValue`. React restores those values instead of clearing them, and the form keeps working before JavaScript loads:
+
+```js
+import { useActionState } from "react";
+import { submitForm } from "./actions.js";
+
+function EditForm() {
+  // The action returns { submitted: formData, error } on failure
+  const [state, formAction] = useActionState(submitForm, {});
+  const values = state.submitted ?? new FormData();
```
aurorascharff: Why do we need a seperate variable for values? Is this pattern better than returning the values from the useActionState function?

https://aurorascharff.no/posts/handling-form-validation-errors-and-resets-with-useactionstate/
https://www.robinwieruch.de/react-server-action-reset-form/

src/content/reference/react-dom/components/form.md
MaxwellCohen: Good point! I will reword!

see b3417281599b83c801989235ed0d119b89be8d05

src/content/reference/react-dom/components/form.md
MaxwellCohen: Great point. When I read it 3 times last week, it seemed so clear ;) now it is not 


udpate in b3417281599b83c801989235ed0d119b89be8d05

src/content/reference/react-dom/components/form.md
MaxwellCohen: yep, I got just using onSubmit without action confused

src/content/reference/react-dom/components/form.md
MaxwellCohen: Good point, I forgot about optional chaining (one of my favorite JS features when working on making the examples)

src/content/reference/react-dom/components/form.md
MaxwellCohen: is b3417281599b83c801989235ed0d119b89be8d05 better

src/content/reference/react-dom/components/form.md
MaxwellCohen: is 4f39d4b6c85a2261449552f6c67c86cd47c671fc better?

src/content/reference/react-dom/components/form.md
MaxwellCohen: Most of the examples are called out in the code update it in code, so moving to code and using useActionState 
947c926e596f3154c80723312b8698af998b81bd

aurorascharff: To be clear, sorry if this was too biref, i was wondering about the usage of the () inline linking of in that way!!

aurorascharff: But i think its better the way you did it regardless

src/content/reference/react-dom/components/form.md
aurorascharff: I feel you on that 😂

src/content/reference/react-dom/components/form.md
```diff
@@ -42,27 +42,23 @@ To create interactive controls for submitting information, render the [built-in
...
 
-This example reads the submitted values with [`new FormData(e.target)`](https://developer.mozilla.org/en-US/docs/Web/API/FormData), which collects every field by its `name`. This keeps the inputs [uncontrolled](/reference/react-dom/components/input#reading-the-input-values-when-submitting-a-form). If you instead [control an input with state](/reference/react-dom/components/input#controlling-an-input-with-a-state-variable), read from that state on submit rather than from `FormData`.
 
 <Sandpack>
 
 ```js src/App.js
 export default function Search() {
   function handleSubmit(e) {
-    // Prevent the browser from reloading the page
-    e.preventDefault();
```
hernan-yadiel, line 63: Why was this `e.preventDefault()` removed? 🤔

src/content/reference/react-dom/components/form.md
```diff
@@ -157,7 +165,86 @@ function AddToCart({productId}) {
...
 
-### Display a pending state during form submission {/*display-a-pending-state-during-form-submission*/}
+### Preserving form values after submission {/*preserve-form-values-after-submission*/}
+
+By default, the browser clears a form's input state after submission. Forms with a URL `action` follow this behavior, and React mirrors it when `action` is a function-ensuring your form behaves consistently both before and after JavaScript loads.
+
+When you pass a function to `action` or `formAction`, React resets the form's [uncontrolled fields](/reference/react-dom/components/input#reading-the-input-values-when-submitting-a-form) after the action succeeds. This reset only affects uncontrolled fields—[inputs controlled with state](/reference/react-dom/components/input#controlling-an-input-with-a-state-variable) are never cleared.
+
+To preserve the values of uncontrolled fields after submission, add an `onSubmit` handler that calls `e.preventDefault()` and runs the action inside a [Transition](/reference/react/useTransition). Keep the `action` prop on the form so it continues to work before JavaScript loads.
```
hernan-yadiel: I just tried this and it calls both: the `submitForm` action and the `handleSubmit` function.

If we want the form to _progressively enhance_ and at the same time preserving values after submission, I believe the right answer is `useActionState`.

```jsx
"use client";

import { useActionState } from "react";
import { someServerFunction } from "./some-server-function";

export default function EditForm() {
  const [state, dispatchAction, isPending] = useActionState(someServerFunction, {
    title: "My draft",
  });

  return (
    <form action={dispatchAction}>
      <input name="title" defaultValue={state.title} />
      <button type="submit" disabled={isPending}>
        {isPending ? "Saving..." : "Save"}
      </button>
    </form>
  );
}
```

src/content/reference/react-dom/components/form.md
MaxwellCohen: I removed `e.preventDefault()` to show that, by default, forms reset their values. My goal is to progressively disclose the behavior of forms. React extends HTML form behavior with the onSubmit and action props, not replaces it. I think this is the brilliance of React 19 form APIs.

React is at its best when it takes something that was hard, such as updating content on the page or building reusable components, and makes it easy (sometimes too easy with useEffect 😉 ). I see the form APIs as the next step of this.

If it is unclear, or better to have it, I am open to suggestions.

src/content/reference/react-dom/components/form.md
MaxwellCohen: Thank you very much for looking into this.

`handleSubmit` includes `submitForm` so both functions will be called in onSubmit. 
This comment made me think we should show multiple examples of resetting a form rather than just one. So I added a Recipe with 3 examples of state-keeping. 2cb13706116422e98e0c4dfa177dc1faade6577c

src/content/reference/react-dom/components/form.md
```diff
@@ -157,7 +165,178 @@ function AddToCart({productId}) {
...
+}
+```
+
+</Sandpack>
+
+<Solution />
+
+#### With `onSubmit` and `useTransition` {/*with-onsubmit-and-usetransition*/}
+
+If you also pass a function to the `action` prop—for example, to support [progressive enhancement](/reference/rsc/server-functions)—add an `onSubmit` handler that calls `e.preventDefault()` and runs the Action inside a [Transition](/reference/react/useTransition).
```
hernan-yadiel: > If you also pass a function to the `action` prop-for example, to support progressive enhacement.

This could be a bit misleading, since progressive enhancement is possible only when the function is a **Server Function**, as you  previously mentioned:

> When the action prop is a Server Function, the form is progressively enhanced: submission works before JavaScript loads, and even with JavaScript disabled.

src/content/reference/react-dom/components/form.md
```diff
@@ -157,7 +165,178 @@ function AddToCart({productId}) {
...
+Handle submission in an `onSubmit` handler and call `e.preventDefault()` to stop the browser from navigating and refreshing the page.
+
+<Sandpack>
+
+```js src/App.js
+import { useState } from "react";
+import { submitForm } from "./api.js";
+
+export default function EditForm() {
+  const [isPending, setIsPending] = useState(false);
```
hernan-yadiel: I think this is a pattern we should stop recommending now that we have better primitives, so I would just drop this example and leave the other two which are "Async React compliant". What do you think?

MaxwellCohen: Very good point. I removed this example in a3cd24079a5197fa10f9b5cbfe800f88cc20e348

src/content/reference/react-dom/components/form.md
MaxwellCohen: I updated the text with a3cd24079a5197fa10f9b5cbfe800f88cc20e348

src/content/reference/react-dom/components/form.md
```diff
@@ -42,27 +42,23 @@ To create interactive controls for submitting information, render the [built-in
...
 
 ---
 
 ## Usage {/*usage*/}
 
-### Handle form submission with an event handler {/*handle-form-submission-with-an-event-handler*/}
+### Handling form submission with an event handler {/*handle-form-submission-with-an-event-handler*/}
 
-Pass a function to the `onSubmit` event handler to run code when the form is submitted. By default, the browser sends the form data to the current URL and refreshes the page, so call [`e.preventDefault()`](https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault) to override that behavior.
+Pass a function to the `onSubmit` event handler to run code when the form is submitted. By default, the browser sends the form data to the current URL and refreshes the page. Call [`e.preventDefault()`](https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault) in your handler to override that behavior.
```
brenelz: maybe change to `in your event handler` instead of just `handler` thought its probably pretty obvious

src/content/reference/react-dom/components/form.md
brenelz: hmm yeah i wasn't sure why this was removed either. I don't really have strong feelings either way tho

src/content/reference/react-dom/components/form.md
```diff
@@ -368,38 +374,189 @@ export async function signUpNewUser(newEmail) {
 
 </Sandpack>
 
-Learn more about updating state from a form action with the [`useActionState`](/reference/react/useActionState) docs
+Learn more about updating state from a form Action with the [`useActionState`](/reference/react/useActionState) docs
+
+### Preserving form values after submission {/*preserve-form-values-after-submission*/}
+
+By default, the browser clears a form's input state after submission. Forms with a URL `action` follow this behavior, and React mirrors it when `action` is a function-ensuring your form behaves consistently both before and after JavaScript loads.
```
brenelz: This should probably be 2 seperate words `function-ensuring`?

src/content/reference/react-dom/components/form.md
```diff
@@ -368,38 +374,189 @@ export async function signUpNewUser(newEmail) {
...
+}
+```
+
+</Sandpack>
+
+<Solution />
+
+#### `onSubmit` and `action` {/*with-onsubmit-and-usetransition*/}
+
+Adding an `onSubmit` handler that calls `e.preventDefault()` will run instead of the `action` prop.
```
brenelz: so if you don't `preventDefault` it will run the `action` prop function? `onSubmit` doesn't always override?

src/content/reference/react-dom/components/form.md
```diff
@@ -368,38 +374,189 @@ export async function signUpNewUser(newEmail) {
...
+
+</Recipes>
+
+<DeepDive>
+
+#### Resetting only some fields, or resetting on the server {/*resetting-only-some-fields*/}
+
+The `onSubmit` approach above keeps every uncontrolled field. When you need finer control, two other patterns are available:
+
+* **Reset from your own Action API.** If you build an Action-based API and still want the form to reset after the Action runs, call the `requestFormReset` API from `react-dom` with the form element inside the Transition.
```
brenelz: did not know this api existed

MaxwellCohen: make this issue to get it documented https://github.com/reactjs/react.dev/issues/8529

src/content/reference/react-dom/components/form.md
MaxwellCohen: That is correct; you can use onSubmit as a logging check, so you would not want to override the default browser behavior. I added a comment to better explain the behavior.

 cfb442226d0ba1b132c4fba7bb8a8667a46bbfff

src/content/reference/react-dom/components/form.md
MaxwellCohen: cleaned up in cfb442226d0ba1b132c4fba7bb8a8667a46bbfff

src/content/reference/react-dom/components/form.md
MaxwellCohen: upated in cfb442226d0ba1b132c4fba7bb8a8667a46bbfff

PR #8521, opened 2026-07-08
[use] Rework await vs use DeepDive to fix inaccurate framing

aurorascharff:
## Summary

Rewrites the "Should I resolve a Promise in a Server or Client Component?" DeepDive on the [`use` reference page](https://react.dev/reference/react/use#resolve-promise-in-server-or-client-component).

The previous text said to prefer `await` in a Server Component "since it keeps the data fetching on the server". That's not accurate: when the Promise is created in a Server Component, the fetch runs on the server whether you `await` it or pass it down, passing an unawaited Promise streams the resolved value to the client, it doesn't move the fetch (raised by @samselikoff). The section also taught passing a Promise down as something specific to `use`, when the same technique works with `await` in a deeper Server Component.

## New framing (from discussion with @gaearon)

- `await` unwraps a Promise on the server, `use` unwraps it on the client.
- Unwrapping where you create the Promise is usually simplest, but blocks everything below.
- Passing the Promise down and unwrapping it deeper is a general technique to unblock the surrounding UI, not a `use` feature. Wrapped in `<Suspense>`, only that part of the tree waits.
- Three examples: blocking `await`, pass down + `await` deeper (unblocking on the server), pass down + `use` deeper (unblocking on the client).
- The closing teaches the only real rule: Client Components can't `await` during render, so they unwrap with `use`. The popover/tooltip case stays attached to the client example.

### DIFF
--- a/src/content/reference/react/use.md
+++ b/src/content/reference/react/use.md
@@ -1113,29 +1113,43 @@ root.render(
 
 #### Should I resolve a Promise in a Server or Client Component? {/*resolve-promise-in-server-or-client-component*/}
 
-A Promise can be resolved with `await` in a Server Component, or passed as a prop to a Client Component and resolved there with `use`.
+If you have a Promise, at some point you need to unwrap it to read its value. You unwrap it with `await` on the server, and with `use` on the client.
 
-Using `await` in a Server Component suspends the Server Component itself, and the Client Component receives the resolved value as a prop:
+Usually, the simplest option is to `await` the Promise where you create it. The Server Component suspends until the data is ready, and everything below it waits too:
 
 ```js
 // Server Component
 export default async function App() {
-  // Will suspend the Server Component.
   const messageContent = await fetchMessage();
   return <Message messageContent={messageContent} />;
 }
 ```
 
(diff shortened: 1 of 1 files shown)


rickhanlonii (approved):

src/content/reference/react/use.md
```diff
@@ -1149,9 +1149,9 @@ export function Message({ messagePromise }) {
 }
 ```
 
-Prefer `await` in a Server Component when possible, since it keeps the data fetching on the server. If a Server Component above already awaits the data, pass the resolved value down as a prop instead of creating a new Promise to call `use`.
+Prefer `await` in a Server Component when possible. If a Server Component above already awaits the data, pass the resolved value down as a prop instead of creating a new Promise to call `use`.
```
gaearon: >Prefer `await` in a Server Component when possible. If a Server Component above already awaits the data, pass the resolved value down as a prop instead of creating a new Promise to call `use`.

This still feels a bit too strong to me. It sounds as if `use` is strictly worse but it's not. It's just that if you need the actual Promise on the server, you `await` where you need it, and if you need it on the client, you `use` that Promise instead.

gaearon (approved):

PR #8536, opened 2026-07-16
Clarify reducer state updater convention

rickvian:
## Summary
- use prevState from setState argument instead of direct state value

## Why
In Past, i actually have production design system component issue caused by someone updates state using existing state directly instead of using prevState from the parameter passed,
i do not want this approach creates bad habit.

The revised lesson teaches new readers the safe convention first: whenever new state is calculated from previous state, pass an updater function and call its argument `prevState`. It introduces batching only after the implementation is understood, as the reason the convention matters.

### DIFF
--- a/src/content/learn/extracting-state-logic-into-a-reducer.md
+++ b/src/content/learn/extracting-state-logic-into-a-reducer.md
@@ -2439,7 +2439,7 @@ textarea {
 
 <Solution>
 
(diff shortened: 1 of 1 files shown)


rickvian:
kindly help review @aurorascharff @MaxwellCohen

rickvian:
Closing this stale PR; the change has been recreated in the current branch.

PR #8538, opened 2026-07-19
Fix Server Function examples for Actions, forms, and useActionState

### LINKED ISSUE #8537: [Mistakes]: Server Functions examples use incompatible action signatures
### Summary

The `updateName` Server Function is defined as `updateName(name)`, but the page later uses it as a direct form action and with `useActionState`. These usages pass different arguments, so the examples cannot work with the shown function.

### Page

https://react.dev/reference/rsc/server-functions

### Details

The page defines:

  ```js
  export async function updateName(name) {
    if (!name) {
      return {error: 'Name is required'};
    }
    await db.users.updateName(name);
  }
```
  There are several inconsistencies/mistakes:

  - In “Server Functions with Actions”, the input does not update the `name` state, so `updateName(name)` always receives `''`.

  - On success, `updateName` returns `undefined`, but the client destructures `{error}` from its result.

  - When passed to `<form action={updateName}>`, React calls `updateName` with `FormData`, not a string.

  - When passed to `useActionState`, React calls it with `(previousState, FormData)`, and it must consistently return the next state.

  The examples should define a Server Function with the correct parameters and return value for each usage.

### PR DESCRIPTION
**Pull request by smith558**
## Summary

Fixes https://github.com/reactjs/react.dev/issues/8537.

The Server Functions reference reused one `updateName` function across examples that invoke it with different arguments. This PR gives each usage a compatible Server Function and completes the corresponding client examples:

- A client Action passes the current name and handles the Server Function's returned error.
- A Server Function passed to the `<form>` `action` prop receives the submitted `FormData`.
- A Server Function passed to `useActionState` receives the previous state and submitted `FormData`, then returns the next state.
- The `permalink` example explains the browser navigation that occurs when a form is submitted before JavaScript loads.

The PR also clarifies Server Function references, form reset behavior, Action terminology, pending state, and the boundary between server and client files.

### DIFF
--- a/src/content/reference/rsc/server-functions.md
+++ b/src/content/reference/rsc/server-functions.md
@@ -42,7 +42,7 @@ Server Components can define Server Functions with the `"use server"` directive:
 // Server Component
 import Button from './Button';
 
-function EmptyNote () {
+function EmptyNote() {
   async function createNoteAction() {
     // Server Function
     'use server';
@@ -50,24 +50,21 @@ function EmptyNote () {
     await db.notes.create();
   }
 
-  return <Button onClick={createNoteAction}/>;
+  return <Button onClick={createNoteAction} />;
 }
 ```
 
 When React renders the `EmptyNote` Server Component, it will create a reference to the `createNoteAction` function, and pass that reference to the `Button` Client Component. When the button is clicked, React will send a request to the server to execute the `createNoteAction` function with the reference provided:
 
-```js {5}
+```js {4}
 "use client";
 
 export default function Button({onClick}) {
-  console.log(onClick);
-  // {$$typeof: Symbol.for("react.server.reference"), $$id: 'createNoteAction'}
-  return <button onClick={() => onClick()}>Create Empty Note</button>
+  return <button onClick={() => onClick()}>Create Empty Note</button>;
 }
 ```
 
 For more, see the docs for [`"use server"`](/reference/rsc/use-server).
-
 
 ### Importing Server Functions from Client Components {/*importing-server-functions-from-client-components*/}
 
@@ -79,19 +76,17 @@ Client Components can import Server Functions from files that use the `"use serv
 export async function createNote() {
   await db.notes.create();
 }
-
 ```
 
 When the bundler builds the `EmptyNote` Client Component, it will create a reference to the `createNote` function in the bundle. When the `button` is clicked, React will send a request to the server to execute the `createNote` function using the reference provided:
 
-```js [[1, 2, "createNote"], [1, 5, "createNote"], [1, 7, "createNote"]]
+```js [[1, 3, "createNote"], [1, 6, "createNote"]]
 "use client";
+
 import {createNote} from './actions';
 
 function EmptyNote() {
-  console.log(createNote);
-  // {$$typeof: Symbol.for("react.server.reference"), $$id: 'createNote'}
-  <button onClick={() => createNote()} />
+  return <button onClick={() => createNote()}>Create Empty Note</button>;
 }
 ```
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by MaxwellCohen** [comment]
Hi, 

I am getting a client-side error on the [server functions preview](https://react-8a6t40u3j-react-foundation.vercel.app/reference/rsc/server-functions). This is usually caused by a misalignment of line numbers for function highlights.


Screenshot_20260719_082731_Chrome



Edit: This has been fixed!

**Comment by smith558** [comment]
@MaxwellCohen thanks, will fix

**Comment by smith558** [comment]
@MaxwellCohen fixed

**Comment by smith558** [comment]
@MaxwellCohen @rickhanlonii @aurorascharff 🙂 please review

File: src/content/reference/rsc/server-functions.md [hunk 1]
```diff
@@ -123,38 +122,55 @@ function UpdateName() {
...
-For more, see the docs for [Calling a Server Function outside of `<form>`](/reference/rsc/use-server#calling-a-server-function-outside-of-form)
+For more, see the docs for [Calling a Server Function outside of `<form>`](/reference/rsc/use-server#calling-a-server-function-outside-of-form).
 
-### Server Functions with Form Actions {/*using-server-functions-with-form-actions*/}
+### Using a Server Function as a Form Action {/*using-server-functions-with-form-actions*/}
 
 Server Functions work with the new Form features in React 19.
 
-You can pass a Server Function to a Form to automatically submit the form to the server:
+You can pass a Server Function to a Form to automatically submit the form to the server. React passes the submitted `FormData` to the Server Function as its first argument:
```
MaxwellCohen, line 161: we should add a link to MDN form data here so that users can see what is included in the form data
[FormData](https://developer.mozilla.org/en-US/docs/Web/API/FormData) link so users know what functions are available. 


```suggestion
You can pass a Server Function to a Form to automatically submit the form to the server. React passes the submitted `[FormData](https://developer.mozilla.org/en-US/docs/Web/API/FormData)` to the Server Function as its first argument:
```

src/content/reference/rsc/server-functions.md
```diff
@@ -219,6 +250,6 @@ function UpdateName() {
 }
 ```
 
-When the <CodeStep step={2}>permalink</CodeStep> is provided to `useActionState`, React will redirect to the provided URL if the form is submitted before the JavaScript bundle loads.
+When the <CodeStep step={2}>permalink</CodeStep> is provided to `useActionState`, the browser will navigate to the provided URL if the form is submitted before the JavaScript bundle loads. At the destination, render `useActionState` with the same Server Function and permalink so React can pass the returned state to it.
```
MaxwellCohen, line 253: ❤️  I really like pointing out that this is browser behavior, not React behavior.

src/content/reference/rsc/server-functions.md
```diff
@@ -166,50 +182,65 @@ function UpdateName() {
...
 
-When the Form submission succeeds, React will automatically reset the form. You can add `useActionState` to access the pending state, last response, or to support progressive enhancement.
+When the Form Action succeeds, React will automatically reset the form's uncontrolled fields. Server Function forms can be submitted before the JavaScript bundle loads. You can add `useActionState` to access the pending state and last response.
 
 For more, see the docs for [Server Functions in Forms](/reference/rsc/use-server#server-functions-in-forms).
 
-### Server Functions with `useActionState` {/*server-functions-with-use-action-state*/}
+### Calling a Server Function with `useActionState` {/*server-functions-with-use-action-state*/}
+
+You can call Server Functions with `useActionState` for the common case where you just need access to the action pending state and last returned response. The Server Function receives the previous state as its first argument and the submitted `FormData` as its second argument. Its return value becomes the next state:
```
MaxwellCohen, line 195:
```suggestion
You can call Server Functions with `useActionState` for the common case where you need access to the Action's pending state and the last returned response. The Server Function receives the previous state as its first argument and the submitted `[FormData](https://developer.mozilla.org/en-US/docs/Web/API/FormData)` as its second argument. Its return value becomes the next state:
```

PR #8539, opened 2026-07-19
fix: update JSX syntax to self-closing tags (img) in writing-markup-w…

rusicsemenov:
fix: update JSX syntax to self-closing tags (img) in writing-markup-with-jsx.md

### DIFF
--- a/src/content/learn/writing-markup-with-jsx.md
+++ b/src/content/learn/writing-markup-with-jsx.md
@@ -74,7 +74,7 @@ Suppose that you have some (perfectly valid) HTML:
   src="https://react.dev/images/docs/scientists/yXOvdOSs.jpg"
   alt="Hedy Lamarr"
   class="photo"
->
+/>
 <ul>
     <li>Invent new traffic lights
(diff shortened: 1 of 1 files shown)


smith558:
@rusicsemenov Please close the PR. These are NOT mistakes; these are intentional. Have a look at the whole article.

rusicsemenov:
@smith558  Hi, 
`` is a void element, so it can't have a closing tag;  it should be self-close it with /> (or just leave it unclosed in **plain HTML**).

yes, maybe for first part we can skip it, nut in the App components (JSX) ing tag should be closed

image

smith558:
@rusicsemenov  I think you misunderstood. The section in the article is about examples of INCORRECT tags. So this is on purpose. You are right that the snippet is not correct. However, in this case, that is on purpose! :)

Have a look at the paragraphs above and below in the article. :)
image

rusicsemenov:
@smith558 Maybe you are right and they are special made more than one mistake, and later show fill for all of them. Yes in this case I can close PR.

PR #8545, opened 2026-07-21
Fix/writing markup with jsx

rusicsemenov:
docs: add example for using refs to access DOM values directly in referencing-values-with-refs.md


rusicsemenov:
Wrong branch name with another commit

PR #8572, opened 2026-08-03
Fix scroll restore when using browser back after hash links

### LINKED ISSUE #787: The browser "back" button doesn't work
The browser "back" button doesn't work after clicking a fragment link (`/page` => `/page#something`).

URL changes back from `/page#something` to `/page`, but it has no effect on the scroll position.

moltresIn:
## Summary
- Fixes #787: after clicking an in-page `#` fragment link, browser Back updates the URL but does not restore scroll when `history.scrollRestoration` is `manual` (Chrome/Firefox).
- Saves `scrollY` on the current history entry via `replaceState` before hash navigation, then restores it on `popstate` when the hash is cleared.
- Leaves Safari’s existing `scrollRestoration = 'auto'` path unchanged so Next.js page navigations stay on manual restoration elsewhere.

## Test plan
- [x] `yarn check-all`
- [x] Local repro on `/learn/describing-the-ui`: scroll mid-page → click heading anchor → Back → scroll returns to prior position (verified with Playwright against `yarn dev`)
- [ ] Reviewer: try Chrome/Firefox on the Vercel preview with the same steps
- [ ] Reviewer: confirm normal sidebar page navigations still scroll as before
- [ ] Reviewer: quick Safari check that back-swipe / auto restoration still feels fine

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/pages/_app.tsx
+++ b/src/pages/_app.tsx
@@ -41,10 +41,74 @@ export default function MyApp({Component, pageProps}: AppProps) {
       // However, we *also* don't want Safari grey screen during the back swipe gesture.
       // Seems like it doesn't hurt to enable auto restore *and* Next.js logic at the same time.
       history.scrollRestoration = 'auto';
-    } else {
-      // For other browsers, let Next.js set scrollRestoration to 'manual'.
-      // It seems to work better for Chrome and Firefox which don't animate the back swipe.
+      return;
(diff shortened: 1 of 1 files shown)


moltresIn:
Closing — opened by mistake; not ready for review.

PR #8578, opened 2026-08-05
Document attaching multiple refs to one element in useRef

### LINKED ISSUE #7251: [Suggestion]: Documentation for `useRef` should tell us how to deal with multiple refs on a single element
### Summary

Sometimes you need to add more than one ref to a single element. For example:

* When using `forwardRef` you need to pass the forwarded ref to whatever element, but you also need your own ref attached to the same element.
* When using `useController` of `react-hook-form`, you get a ref you need to pass to the input element for focus management, but you also need your own ref for doing other things with that same input element.

This is an issue that is not even mentioned on the current documentation of `useRef`. I've been using [gregberge/react-merge-refs](https://github.com/gregberge/react-merge-refs) to deal with this, but after I added the recommended `eslint-plugin-react-compiler`, this method now generates a warning:

> Ref values (the `current` property) may not be accessed during render.

If you're not allowed to access ref values during render to merge them, then how are you supposed to attach multiple refs to a single element now?

The documentation needs to tell us how to do this properly, and in a way that doesn't break the compiler or generate warnings in your eslint plugins.

### Page

https://react.dev/reference/react/useRef

### Details

Example component needing this:

```tsx
export default forwardRef(
  function BaseCheckbox(
    { label, indeterminate = false, onChange, onCheckedChange, ...props },
    ref
  ) {
    const inputId = useId();
    const inputRef = useRef(null);

    useLayoutEffect(() => {
      if (inputRef.current)
        inputRef.current.indeterminate = indeterminate;
    }, [indeterminate]);

    return (
      <div className="checkbox">
        <input
          type="checkbox"
          id={inputId}
          /**
           * Here we have two refs we need to attach to this input, but there's no
           * documented way to do that and the undocumented way is not allowed
           * according to the new rules from `eslint-plugin-react-compiler`.
           */ 
          ref={mergeRefs(inputRef, ref)}
          onChange={(e) => {
            onChange?.(e);
            onCheckedChange?.(e.currentTarget.checked);
          }}
          aria-checked={
            indeterminate ? 'mixed' : props.checked ? 'true' : 'false'
          }
          {...props}
        />
        <label htmlFor={inputId}>
          {label}
        </label>
      </div>
    );
  }
);
```

smg99:
Fixes #7251

## Summary
- Add a new `useRef` troubleshooting section: "How can I attach more than one ref to a single element?"
- Show a compiler-safe pattern using a ref callback to assign the same DOM node to multiple refs
- Cover both object refs and callback refs in the example

## Why
Issue #7251 asks for docs guidance on combining multiple refs on one element (for example local + forwarded refs) without relying on patterns that read `ref.current` during render.

## Validation
- Ran `yarn lint`
- Ran `yarn lint-heading-ids`

### DIFF
--- a/src/content/reference/react/useRef.md
+++ b/src/content/reference/react/useRef.md
@@ -592,3 +592,36 @@ export default MyInput;
 Then the parent component can get a ref to it.
 
 Read more about [accessing another component's DOM nodes.](/learn/manipulating-the-dom-with-refs#accessing-another-components-dom-nodes)
+
+### How can I attach more than one ref to a single element? {/*how-can-i-attach-more-than-one-ref-to-a-single-element*/}
+
+Sometimes you need one element to update multiple refs. For example, you might keep a local ref and also receive a ref from a parent.
+
+Use a ref callback that assigns the same DOM node to each ref:
+
+```js
+import { useRef } from 'react';
+
+function assignRef(ref, value) {
+  if (typeof ref === 'function') {
+    ref(value);
+  } else if (ref != null) {
+    ref.current = value;
+  }
+}
+
+function MyInput({ ref: forwardedRef }) {
+  const localRef = useRef(null);
+
+  return (
+    <input
+      ref={node => {
+        assignRef(localRef, node);
+        assignRef(forwardedRef, node);
+      }}
+    />
+  );
+}
+```
+
+This pattern avoids reading `ref.current` during render and works with both object refs and callback refs.


smg99:
Requesting review for this docs addition.

This PR adds a `useRef` troubleshooting section showing how to attach multiple refs to one element with a ref callback, including object refs and callback refs. The intent is to provide a compiler-safe pattern without reading `ref.current` during render.

Happy to adjust placement or wording.

PR #8579, opened 2026-08-05
Clarify ref callback cleanup behavior in common ref docs

### LINKED ISSUE #7811: [Suggestion]: clarify behaviour of ref callbacks
### Summary

The described behaviour for ref callbacks with `null` as an argument is ambiguous.

### Page

https://react.dev/reference/react-dom/components/common#ref-callback

### Details

From https://react.dev/reference/react-dom/components/common#ref-callback:

> React will also call your ref callback whenever you pass a different ref callback. In the above example, (node) => { ... } is a different function on every render. **When your component re-renders, the previous function will be called with null as the argument, and the next function will be called with the DOM node**.

If I understand correctly, the emphasises sentenced is not correct when the ref callback returns a clean up function. This needs to be clarified.

Reduced test case to confirm this behaviour: https://stackblitz.com/edit/vitejs-vite-2r9rn6m2?file=src%2FApp.tsx&terminal=dev

smg99:
Fixes #7811

## Summary
- Clarify the `ref` callback behavior when callbacks change between renders
- Explicitly describe the cleanup-function path as the primary behavior
- Clarify that passing `null` to the old callback is legacy fallback behavior when no cleanup is returned

## Why
Issue #7811 points out ambiguity in the current wording, especially around when React calls a previous callback with `null`. This change aligns wording with current React 19 cleanup semantics while preserving backwards-compatibility notes.

## Validation
- Ran `yarn lint`
- Ran `yarn lint-heading-ids`

### DIFF
--- a/src/content/reference/react-dom/components/common.md
+++ b/src/content/reference/react-dom/components/common.md
@@ -257,9 +257,9 @@ Instead of a ref object (like the one returned by [`useRef`](/reference/react/us
 
 [See an example of using the `ref` callback.](/learn/manipulating-the-dom-with-refs#how-to-manage-a-list-of-refs-using-a-ref-callback)
 
-When the `<div>` DOM node is added to the screen, React will call your `ref` callback with the DOM `node` as the argument. When that `<div>` DOM node is removed, React will call your the cleanup function returned from the callback.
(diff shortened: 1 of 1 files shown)


smg99:
Requesting review for this wording clarification.

This updates the `ref` callback section to emphasize the cleanup-function path first, and clarifies that `null` is the legacy fallback when no cleanup function is returned.

Happy to refine wording further if maintainers prefer a different phrasing.

PR #8582, opened 2026-08-08
Document browser-only rendering

gnoff:
Add the Canary `browser` API reference, including optional lazy reasons, server bailout reporting, fatal and abort behavior, navigation entries, and `onBrowserBailout` options for every streaming, resume, and prerender API that supports it.

Preview: https://react-57rtx199n-react-foundation.vercel.app/reference/react-dom/browser


src/content/reference/react-dom/browser.md
```diff
@@ -0,0 +1,179 @@
...
+
+During server rendering, React includes the `Loading editor...` fallback in the HTML. When the app renders in the browser, `use(browser())` continues immediately and React renders the `Editor` instead.
+
+---
+
+### Conditionally rendering in the browser {/*conditionally-rendering-in-the-browser*/}
+
+Like other calls to [`use`](/reference/react/use), `use(browser())` can be called conditionally, including inside a custom Hook. For example, you can wrap a Suspense-enabled data-fetching library's `useQuery` to render initial data on the server, but defer to the browser when that data is missing:
+
+```js {3}
```
MaxwellCohen, line 211: Does this example need a  "use client" here?

```suggestion
```js {4}
'use client';
```

gnoff: `react-dom` is a "client" library in the sense that it is for browser (react-dom/client) and render-for-browser-as-html (react-dom/server). You can use this in frameworks that don't even support RSC so I don't think the examples should imply that this is in any associated with RSC even if it is true that in an RSC framework you can't use this API in Server Components.

It's also not the case that you need "use client" on every file that is part of the client bundle, you only need this in files that you want to import into the server as references

MaxwellCohen: > you can't use this API in Server Components

This is my main concern: `use(browser())` must be a client component; this page does not explicitly call it out, although it is implied. 

When working with a traditional client-focused framework (like Next.js Pages Router, TanStack Start),` use(browser())` works great. 

There is a different story for server component frameworks (like Next.js App Router and Waku), where components default to being server components and do not execute code in the browser, so components that have `use(browser())` and no other hooks will be treated as server components by default. Thus, they will not render and will error out unless they are client components. `use(browser())` does not force a client component, like having a hook in it. While `"use client"` is not needed in every file, I would argue that if you have code that must be in the browser, it should have a `"use client"` so developers know that code is going into the client and don't have hidden dependencies that could break people's apps. 

If this example does not use "use client", there should be some explicit callout of how to use `use(browser())` in server component frameworks.

aurorascharff: Added a note: https://github.com/reactjs/react.dev/pull/8582/changes/385d3068055a8b093954e38cdcd6310d11165bb7

PR #8593, opened 2026-08-14
Fix act() event example using button before declaration

fbellinder:
## Summary
- The [Dispatching events in tests](https://react.dev/reference/react/act#dispatching-events-in-tests) example called `button.dispatchEvent(...)` before `const button = ...`, which would throw a `ReferenceError` (temporal dead zone).
- Move the `querySelector` calls above the click `act()`, remove leftover `it.only`, and update highlight line numbers.
- Related open PR that touches the same snippet for async style only: https://github.com/reactjs/react.dev/pull/8142

## Test plan
- [ ] Confirm the example on `/reference/react/act` renders with the highlighted lines on the dispatch `act()` block
- [ ] Mentally/run the snippet: `button` is defined before `dispatchEvent`

### DIFF
--- a/src/content/reference/react/act.md
+++ b/src/content/reference/react/act.md
@@ -117,26 +117,27 @@ Using `act` ensures that all updates have been applied before we make assertions
 
 To test events, wrap the event dispatch inside `act()`:
 
-```js {14,16}
+```js {16,17}
 import {act} from 'react';
 import ReactDOMClient from 'react-dom/client';
 import Counter from './Counter';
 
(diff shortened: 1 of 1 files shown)


deepusnath:
Confirmed against main, the second example really does call button.dispatchEvent() before const button is declared, so copying it as-is throws a TDZ ReferenceError. Good catch.

Worth calling out that this also removes an it.only that had crept into the snippet,  in a copy-pasted docs example that would silently skip every other test in the user's file, so that part is arguably the more valuable half of the fix.

PR #8597, opened 2026-08-15
Add cache, cacheSignal, and captureOwnerStack to the APIs overview

see-stack:
The Built-in React APIs page lists all modern React APIs, but it was missing `cache`, `cacheSignal`, and `captureOwnerStack`, each of which has a dedicated reference page. This adds all three.

**Before**: missing `cache`, `cacheSignal`, and `captureOwnerStack`
Screenshot 2026-08-15 at 6 48 53 AM

**After**: with `cache`, `cacheSignal`, and `captureOwnerStack` added
Screenshot 2026-08-15 at 6 49 01 AM


eps1lon (approved):
Thank you

PR #8598, opened 2026-08-15
Fix resume and resumeToPipeableStream links

see-stack:
The `resume` and `resumeToPipeableStream` entries in the Server React DOM APIs page linked to `renderToPipeableStream` instead of their own reference pages. This fixes both links.

**Before**: `resume` redirected to `renderToPipeableStream`
Screenshot 2026-08-15 at 7 11 13 AM
Screenshot 2026-08-15 at 7 10 21 AM

**After**: `resume` redirects to its own page
Screenshot 2026-08-15 at 7 10 25 AM


deepusnath (commented):
Verified both: server/index.md L18 and L33 each pointed at renderToPipeableStream rather than the API the bullet describes, and resume.md / resumeToPipeableStream.md both exist as targets. Straightforward correct fix.

PR #8599, opened 2026-08-15
Fix wrong link targets in react-dom docs

see-stack:
Three links pointed to the wrong page:
- `resume` linked to `renderToReadableStream` (should be `resume`)
- `react-dom/server` linked to `createRoot` (should be `server`)
- `renderToPipeableStream` linked to `renderToReadableStream` (should be `renderToPipeableStream`)


deepusnath (commented):
Verified.  hydrateRoot.md:126 says "generated by react-dom/server" while linking to client/createRoot, so the link text and destination disagree. /reference/react-dom/server exists as a target. Correct fix.

PR #8601, opened 2026-08-15
Fix broken links to the React 19 Upgrade Guide

see-stack:
Three links still used the old `/blog/2024/04/25/react-19` path, which 404s since the page was renamed to `react-19-upgrade-guide`. Updated them to the current path.


deepusnath:
These links aren't currently broken — vercel.json (L248) permanently redirects /blog/2024/04/25/react-19 → /blog/2024/12/05/react-19, and the fragment is preserved, so they land on the right headings today.

The new target is the problem: react-19-upgrade-guide.md doesn't define {/*error-handling*/} or {/*ref-as-a-prop*/} — those headings live in blog/2024/12/05/react-19.md (L727 and L396). So after this change the pages load but drop the reader at the top of the wrong post, which a link checker won't flag.

If the goal is to skip the redirect hop, the repo already has a convention for that — useImperativeHandle.md:49 links straight to /blog/2024/12/05/react-19#ref-as-a-prop. Retargeting to that path instead would work.

see-stack:
Thanks for catching this!

see-stack:
I updated the references to point to the correct React 19 blog anchors. The changes are now limited to the affected docs links. Appreciate the review!

deepusnath:
Re-checked the updated diff — all three links now resolve to /blog/2024/12/05/react-19, and both anchors exist there ({/*ref-as-a-prop*/} L396, {/*error-handling*/} L727). Matches the convention in useImperativeHandle.md. Looks right to me.

PR #8603, opened 2026-08-15
Fix API Reference link to point to /reference/react

see-stack:
The `/reference` root has no landing page (only `/reference/react`, `/reference/react-dom`, etc.), so the link 404'd. Point it at the React API reference landing.


deepusnath (commented):
Verified. There's no src/content/reference/index.md, so /reference really is a 404, and /reference/react/index.md exists. Correct.

Issue #8594, opened 2026-08-14
Accessibility on Quick Start Page


samilieberman:
On the [Quick Start page](https://react.dev/learn), there is an accessibility issue that should be fixed to `Ensure the order of headings is semantically correct`. The axe chrome extension can be used to validate the fix. More info below.

<img width="1727" height="646" alt="Image" src="https://github.com/user-attachments/assets/af298d2a-48fd-4883-baa2-5a3abab8ee2a" />

<img width="603" height="589" alt="Image" src="https://github.com/user-attachments/assets/9c0c7c1a-346c-420d-93fa-533de8d76ade" />

taauntik:
Checking ^

PR #8607, opened 2026-08-17
Accessibility on Quick Start Page

taauntik:
RE: https://github.com/reactjs/react.dev/issues/8594

Updated the `SimpleCallout.tsx` to fix it. Replaced `<H3>` to `<H2>`

### DIFF
--- a/src/components/MDX/SimpleCallout.tsx
+++ b/src/components/MDX/SimpleCallout.tsx
@@ -11,7 +11,7 @@
 
 import * as React from 'react';
 import cn from 'classnames';
-import {H3} from './Heading';
+import {H2} from './Heading';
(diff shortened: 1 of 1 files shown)


deepusnath (commented):
Verified this is safe beyond Quick Start — <YouWillLearn>/SimpleCallout is used on 43 pages, and in every one it precedes the first ## section, so bumping H3→H2 doesn't introduce a skip or reorder anywhere else in the docs. Nice targeted fix.

Side note, not blocking: I didn't see an automated heading-hierarchy/axe check in CI (site_lint.yml), so this class of regression won't be caught automatically next time — might be worth a follow-up issue.

PR #8612, opened 2026-08-19
Clarify browser-only rendering guidance

aurorascharff:
## Summary

- mention the Canary `browser` value in the `use` introduction and Built-in React APIs overview
- give `use(browser())` its own usage group instead of nesting it under Promise usage
- distinguish browser-only rendering from hydration's two-pass client/server content pattern
- link the `useLayoutEffect` server-rendering guidance to the Canary browser-only API

## Why

The `browser` API reference and examples landed in #8582, but nearby documentation still described `use` as accepting only Promises and context or presented older browser-only workarounds without mentioning the dedicated Canary API. This follow-up connects those existing entry points while preserving the guidance needed by stable React releases.

## Validation

- `yarn lint-heading-ids`
- ESLint on all changed Markdown files
- `yarn deadlinks` (all 1,900 links valid)
- `yarn build` (all 816 static pages and 219 OG images generated)


src/content/reference/react-dom/client/hydrateRoot.md
```diff
@@ -319,6 +319,10 @@ export default function App() {
 
 This way the initial render pass will render the same content as the server, avoiding mismatches, but an additional pass will happen synchronously right after hydration.
 
+Use this approach when you need to render different content on the server and in the browser.
```
gaearon: this is a bit confusing. maybe "when you want the client-rendered content to be different from the initial server-rendered HTML" or something like that?

src/content/reference/react/apis.md
```diff
@@ -25,13 +25,21 @@ In addition to [Hooks](/reference/react/hooks) and [Components](/reference/react
 
 *Resources* can be accessed by a component without having them as part of their state. For example, a component can read a message from a Promise or read styling information from a context.
 
-To read a value from a resource, use this API:
+You can pass these types of resources to [`use`](/reference/react/use):
+
+* A [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) to read its resolved value.
+* A [context](/learn/passing-data-deeply-with-context) to read its value.
+* <CanaryBadge /> The value returned by [`browser`](/reference/react-dom/browser) to mark a component as browser-only during server rendering.
```
gaearon, line 32: Is it considered special/opaque or is it just a Promise? Question to @gnoff maybe.

src/content/reference/react/apis.md
```diff
@@ -25,13 +25,21 @@ In addition to [Hooks](/reference/react/hooks) and [Components](/reference/react
...
 
-* [`use`](/reference/react/use) lets you read the value of a resource like a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) or [context](/learn/passing-data-deeply-with-context).
 ```js
 function MessageComponent({ messagePromise }) {
   const message = use(messagePromise);
   const theme = use(ThemeContext);
   // ...
 }
+
+function BrowserOnlyComponent() {
```
gaearon: This feels weirdly split from the above example. I think we can just `use(browser())` as a third line in the previous one?

src/content/reference/react/use.md
```diff
@@ -4,7 +4,7 @@ title: use
 
 <Intro>
 
-`use` is a React API that lets you read the value of a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) or [context](/learn/passing-data-deeply-with-context).
+`use` is a React API that lets you read the value of a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) or [context](/learn/passing-data-deeply-with-context). You can also pass `use` the value returned by [`browser`](/reference/react-dom/browser) to mark a component as browser-only during server rendering.
```
gaearon: Same question, do we consider it opaque.

src/content/reference/react-dom/client/hydrateRoot.md
```diff
@@ -319,6 +319,10 @@ export default function App() {
 
 This way the initial render pass will render the same content as the server, avoiding mismatches, but an additional pass will happen synchronously right after hydration.
 
+Use this approach when you want the client-rendered content to be different from the initial server-rendered HTML.
+
+<Canary>If a component should render only in the browser, call [`use(browser())`](/reference/react/use#use-browser) instead of waiting for an Effect.</Canary>
```
gaearon, line 326: Note we should probably delete this entire subsection once this ships in stable?

src/content/reference/react/use.md
```diff
@@ -4,7 +4,7 @@ title: use
 
 <Intro>
 
-`use` is a React API that lets you read the value of a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) or [context](/learn/passing-data-deeply-with-context).
+`use` is a React API that lets you read the value of a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) or [context](/learn/passing-data-deeply-with-context). You can also call `use` with the opaque value returned by [`browser`](/reference/react-dom/browser) to mark a component as browser-only during server rendering.
```
gaearon: This feels a bit disproportionate:

Image

Any ideas for how to compress it so that it doesn't take 80% of the explanation for a rare case?

gaearon (approved):

PR #8628, opened 2026-08-30
Fix mobile in-app browser navigation

### LINKED ISSUE #656: Mobile in-app browser navigation glitching
Opened up an article inside the Reddit in-app browser and noticed some funky glitching in the navigation. 

**Issue**
On open, the nav starts half-scrolled, and then on scroll you can see the bottom of the text going behind the navigation. 

On mobile Chrome, Safari, and Firefox, the same thing happens- where if the user scrolls inside the nav, the content behind it scrolls- you just can't see it in these browsers due to correct screen positioning of the nav. The in-app browser seems to force the nav to have some kind of margin at the bottom.

bug-f

This definitely could also be a gatsby problem, reddit mobile problem, in-app browsers as a whole problem, or this particular site problem. I also never throw out the possibility of it being a problem with my own device. I'll poke around and see what I can find.

**Sidenote**: I also have a CSS ~hack~ fix that prevents background content from scrolling when the user scrolls in the nav. I'll open a new issue for that if that's wanted.

lava-m312:
## Description

Fixes mobile in-app browser navigation positioning and scroll behavior.

## Changes

- Use `100dvh` for the mobile navigation height.
- Use `fixed` positioning for the open mobile navigation.
- Keep the navigation positioned at the top of the viewport.

## Testing

- Tested mobile navigation using Chrome DevTools with Pixel 7 emulation.
- Confirmed the navigation scrolls correctly.
- Confirmed the page behind the navigation does not scroll when reaching the navigation boundary.
- Confirmed the navigation opens from the top of the viewport.
- `yarn lint` passes.
- `yarn tsc` passes.

Fixes #656

### DIFF
--- a/src/components/Layout/TopNav/TopNav.tsx
+++ b/src/components/Layout/TopNav/TopNav.tsx
@@ -247,7 +247,7 @@ export default function TopNav({
       <div
         className={cn(
           isMenuOpen
-            ? 'h-screen sticky top-0 lg:bottom-0 lg:h-screen flex flex-col shadow-nav dark:shadow-nav-dark z-20'
+            ? 'h-[100dvh] fixed inset-x-0 top-0 lg:bottom-0 lg:h-[100dvh] flex flex-col shadow-nav dark:shadow-nav-dark z-20'
             : 'z-40 sticky top-0'
         )}>
         <nav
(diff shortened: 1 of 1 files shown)


lava-m312 (commented):
check on this and give review sonner

Issue #8523, opened 2026-07-09
[Bug]: server function page is broken


kamol-ayyub:
### Summary

The Server Functions documentation page is broken and crashes when loaded.

### Page

https://react.dev/reference/rsc/server-functions

### Details

When I visit the Server Functions documentation page (https://react.dev/reference/rsc/server-functions), the page crashes instead of loading correctly.

I've attached screenshots showing the issue.

<img width="907" height="429" alt="Image" src="https://github.com/user-attachments/assets/600a2b55-668f-4538-9099-28f82ac6f061" />

<img width="1640" height="333" alt="Image" src="https://github.com/user-attachments/assets/58e2094d-3ed9-4294-81af-9b3c20155c45" />

ParinitaDutta:
I can reproduce this issue as well.

**Environment:**
OS: Windows 10
Browsers: Google Chrome and Microsoft Edge
Location: India

**I also tried:**
Incognito mode
Hard refresh
Disabling cache in DevTools

None of these resolved the issue.

**Console output:**
Minified React error #419
The server could not finish this Suspense boundary, likely due to an error during server rendering. Switched to client rendering.
Error: Could not find: 'submitAction'

Other pages on react.dev work normally. Only https://react.dev/reference/rsc/server-functions is affected.

PR #8645, opened 2026-09-08
Validate inline code highlight metadata in CI

aurorascharff:
## Summary

Invalid inline code highlight metadata currently throws from `CodeBlock` at runtime, which can leave an otherwise valid documentation page showing a client-side application error. This happened after #8520 moved `<form action={submitAction}>` from line 23 to line 25 without updating its highlight metadata. The broken Server Functions page was reported in #8523 and fixed separately in #8531.

This PR extends the existing Markdown ESLint rule to validate inline code highlights before deployment. It checks that:

- the metadata is valid JSON with the expected tuple structure;
- the referenced line exists in the code block;
- the highlighted text exists on that line; and
- repeated text specifies a `fromIndex`.

The validator uses the normalized metadata produced by the Markdown parser, matching what MDX passes to `CodeBlock` and supporting highlights that contain escaped quotes. Regression fixtures cover valid metadata, missing text, and out-of-bounds lines. The local-rule tests now run as part of `ci-check`.

## Testing

- `yarn test:eslint-local-rules`
- `yarn eslint "src/content/**/*.md"`
- `yarn ci-check`

To verify the regression, I temporarily changed the Server Functions example's valid `submitAction` highlight from line 25 to line 24. `ci-check` failed with:

```text
server-functions.md:115 — Could not find 'submitAction' on highlighted line 24
```

After restoring line 25, the complete `ci-check` passed.

### DIFF
--- a/eslint-local-rules/__tests__/fixtures/src/content/missing-highlight.md
+++ b/eslint-local-rules/__tests__/fixtures/src/content/missing-highlight.md
@@ -1,0 +1,3 @@
+```js [[1, 1, "submitAction"]]
+function UpdateName() {}
+```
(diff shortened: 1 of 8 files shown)


eps1lon (approved):
Immediate PTSD when thinking about how much time I spent getting these highlight tupples right.

Issue #8231, opened 2026-01-12
[Mistake]:  Learn > Escape Hatches > Lifecycle of Reactive Effects


jaem0629:
### Summary

If you look at `ChatRoom.js` in an example of the **4. Fix a connection switch code** in **Try out some challenge**, it looks like the image below.
The `useState` is not used in the code, so it is correct to remove it.

<img width="1045" height="439" alt="Image" src="https://github.com/user-attachments/assets/a0ecbf58-a63c-4bb3-aaec-4f36a005326c" />

### Page

https://react.dev/learn/lifecycle-of-reactive-effects

PR #8647, opened 2026-09-08
Fix unused useState import in lifecycle effects example

devakichuu22-design:
Removed the unused `useState` import from the `ChatRoom.js` example in the Lifecycle of Reactive Effects documentation.

This fixes the unused import mentioned in issue #8231.


SapanMozammel (commented):
LGTM! Clean fix, removes the unused useState import from the example snippet.

PR #8648, opened 2026-09-09
Clarify error handling for standalone startTransition

aurorascharff:
## Summary

- document the pending-state, component-association, and error-handling differences between the standalone `startTransition` function and the function returned by `useTransition`
- clarify that errors from a standalone Action are reported as uncaught, while errors from an Action passed to the hook-returned function can be handled by the component's nearest Error Boundary
- note that the standalone function is not a Hook and can also be called outside components

## Validation

- `yarn eslint src/content/reference/react/startTransition.md src/content/reference/react/useTransition.md`
- `yarn lint-heading-ids`
- `yarn tsc`
- `git diff --check`

### DIFF
--- a/src/content/reference/react/startTransition.md
+++ b/src/content/reference/react/startTransition.md
@@ -92,14 +92,7 @@ With a Transition, your UI stays responsive in the middle of a re-render. For ex
 
 <Note>
 
-`startTransition` is very similar to
-[`useTransition`](/reference/react/useTransition), except that it does not
-provide the `isPending` flag to track whether a Transition is ongoing. The
-standalone function is also not associated with a component, so if the function
-passed to it throws an error or returns a rejected Promise, React reports the
-error as uncaught. You can call `startTransition` when `useTransition` is not
-available. For example, `startTransition` works outside components, such as from
-a data library.
+`startTransition` is very similar to [`useTransition`](/reference/react/useTransition), except that it does not provide the `isPending` flag to track whether a Transition is ongoing. The standalone function is also not associated with a component, so if the function passed to it throws an error or returns a rejected Promise, React reports the error with [`reportError`](https://developer.mozilla.org/en-US/docs/Web/API/Window/reportError). You can call `startTransition` when `useTransition` is not available. For example, `startTransition` works outside components, such as from a data library.
 
 [Learn about Transitions and see examples on the `useTransition` page.](/reference/react/useTransition)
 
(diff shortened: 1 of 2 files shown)


src/content/reference/react/useTransition.md
```diff
@@ -1738,7 +1738,11 @@ This is a JavaScript limitation due to React losing the scope of the async conte
 
 ### I want to call `useTransition` from outside a component {/*i-want-to-call-usetransition-from-outside-a-component*/}
 
-You can't call `useTransition` outside a component because it's a Hook. In this case, use the standalone [`startTransition`](/reference/react/startTransition) method instead. It works the same way, but it doesn't provide the `isPending` indicator.
+You can't call `useTransition` outside a component because it's a Hook. In this
+case, the standalone [`startTransition`](/reference/react/startTransition)
+function can mark state updates as Transitions. It does not provide the
+`isPending` flag. Because the standalone function is not associated with a
+component, an Error Boundary cannot handle errors from its Action.
```
eps1lon:
```suggestion
component, an Error Boundary cannot handle errors from its Transition.
```

src/content/reference/react/startTransition.md
```diff
@@ -92,7 +92,14 @@ With a Transition, your UI stays responsive in the middle of a re-render. For ex
 
 <Note>
 
-`startTransition` is very similar to [`useTransition`](/reference/react/useTransition), except that it does not provide the `isPending` flag to track whether a Transition is ongoing. You can call `startTransition` when `useTransition` is not available. For example, `startTransition` works outside components, such as from a data library.
+`startTransition` is very similar to
+[`useTransition`](/reference/react/useTransition), except that it does not
+provide the `isPending` flag to track whether a Transition is ongoing. The
+standalone function is also not associated with a component, so if the function
+passed to it throws an error or returns a rejected Promise, React reports the
```
eps1lon: Sounds like it goes to `onUncaughtError`. Can you check if that's the case. Other candidates are `reportError` (which the browser users if a thrown value isn't caught by try-catch). I suspect React really doesn't do anything here and we defer to the platform default.

aurorascharff: > Standalone startTransition calls reportGlobalError for synchronous errors and uses it as the rejection handler for returned Promises. reportGlobalError delegates to the global reportError when available, with fallbacks for environments where it isn’t available. It doesn’t go through the root’s onUncaughtError path.

eps1lon: makes sense. I think we can say that it goes through [reportError](https://developer.mozilla.org/en-US/docs/Web/API/Window/reportError) then to avoid confusion with `onUncaughtError`.

It makes implicitly sense because `onUncaughtError` is only triggered for errors you could have handled with an error boundary.

eps1lon (approved):

PR #8652, opened 2026-09-11
Reword `browser()` Server Component caveat

hernan-yadiel:
The current wording feels a bit odd, since apps aren't typically referred to as "React Server Components apps." The caveat is clearer and more direct without that part.


SapanMozammel (commented):
Nice editorial cleanup, removes redundant phrasing while keeping the caveat clear.

eps1lon (approved):
thanks

PR #8658, opened 2026-09-15
Add an `<img>` API reference

aurorascharff:
Adds an API reference for the built-in browser `` component.

The new reference:

- documents image-specific props and caveats
- explains automatic image preload hints during server rendering, including opt-out behavior and how frameworks and component libraries may provide different defaults
- documents how images participate in synchronized View Transition reveals

It also adds `` to the React DOM components index and sidebar.

The automatic preloading behavior was introduced in https://github.com/react/react/pull/27191 and discussed in https://github.com/react/react/issues/34217.


src/content/reference/react-dom/components/img.md
```diff
@@ -0,0 +1,257 @@
...
+* `referrerPolicy`: a string. Specifies the [referrer information](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#referrerpolicy) to send when fetching the image.
+* `sizes`: a string. Specifies the image sizes for different page layouts. Used with `srcSet`.
+* `src`: a string. Specifies the URL of the image.
+* `srcSet`: a string. Specifies one or more candidate image sources for the browser to choose from.
+* `useMap`: a string. Associates the image with a [client-side image map](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/map).
+* `width`: a number or string. Specifies the rendered width of the image.
+
+#### Caveats {/*caveats*/}
+
+* Do not pass an empty string to `src`. React warns in development and omits the attribute. To render no image, omit the `<img>` or pass `null` to `src`.
```
eps1lon, line 52: worth explaining why? With an empty attribute the browser would make a request to the current page.

eps1lon (approved):

src/content/reference/react-dom/components/img.md
```diff
@@ -0,0 +1,257 @@
...
+  );
+}
+
+export default function App() {
+  const [src, setSrc] = useState(null);
+  return (
+    <>
+      <button
+        onClick={() => {
+          startTransition(() => {
```
slorber, line 167: Hey, it's not clear to me whether startTransition is required here

The text above doesn't mention it, may be worth clarifying

src/content/reference/react-dom/components/img.md
```diff
@@ -0,0 +1,257 @@
...
+import { freshImageUrl } from './image.js';
+
+export default function VanillaProfile() {
+  const ref = useRef(null);
+  function show() {
+    ref.current.innerHTML = `<div class="card">
+      <img src="${freshImageUrl()}" alt="Jack Pope" width="80" height="80" />
+      <p>Jack Pope</p>
+    </div>`;
+  }
```
slorber, line 198: I'm a bit confused by this example

One is using `<ViewTransition><Suspense>`, the other is using non-idionmatic React code with imperative mutations.

If the goal is to show the image behavior within `<ViewTransition>`, shouldn't this be compared against idiomatic React code (setState) that is not within a `<ViewTransition>`?

